PR #15, opened 2014-12-03
Documentation

syclops:


lib/packet/packet_base.py
```diff
@@ -50,12 +53,14 @@ class PacketBase(object):
...
-        payload: a packet (subclass of PacketBase) or bytes literal
-            representing the packet payload.
+    :ivar parsed: whether or not the packet has been parsed.
+    :vartype parsed: bool
+    :ivar raw: the raw bytes of the packet contents.
+    :vartype raw: bytes
+    :ivar hdr: the packet header.
+    :vartype hdr: :class:`HeaderBase`
+    :ivar payload: the packet payload
+    :vartype payload: :class:`HeaderBase` or bytes
```
pszal: `HeaderBase`?

btw. it would be good to have e.g. wiki page with commenting scheme you propose. Then we could just follow this style in a new code.

syclops: Is there something wrong with HeaderBase? That's the name of the class,
and all subclasses will have that type, so it seems correct to me...

pszal: Why payload in this line is of HeaderBase class? (should be subclass of PacketBase as in corresponding removed line).

syclops: Ah yes, you're right. That was a typo on my part - I'll fix and resubmit
the pull request.

-Steve

PR #20, opened 2015-01-26
IP address as an input, setup and run files split into multiple ones.

LorenzoBaesso:


topology/ADConfigurations.json
```diff
@@ -0,0 +1,122 @@
...
+        }
+    },
+    "2-26": {
+        "level": "LEAF",
+        "links": {
+            "2-23": "PARENT",
+            "2-24": "PARENT"
+        }
+    }
+}
```
pszal, line 123: Actually I like this format, but why do I need to set LEAF/INTERMEDIATE ? If links are only parent then it is LEAF. CORE/NON-CORE is enough.

LorenzoBaesso, line 123: Discussed. path_servers added to intermediate ADs (affects topology files and config files) and tested.

lib/crypto/certificates.py
```diff
@@ -26,11 +28,14 @@
 
 def load_root_certificates(path):
```
pszal: why do we have that? shouldn't be load_trc(...) ?

lib/crypto/trcs.py
```diff
@@ -1,19 +1,21 @@
-"""
```
pszal, line 1: why trcs.py not trc.py ?

topology/generator.py
```diff
@@ -20,39 +41,99 @@
...
+BS_RANGE = '1'
+CS_RANGE = '21'
+PS_RANGE = '41'
+ER_RANGE = '61'
 
+default_subnet = "127.0.0.0/8"
 
-def increment_address(ip_address):
+
+def increment_address(ip_address, increment=1):
```
pszal: does it detect broadcast/net addr  ?

topology/generator.py
```diff
@@ -20,39 +41,99 @@
...
+        if "subnet" in AD_configs[isd_ad_id]:
+            first_byte = AD_configs[isd_ad_id]["subnet"].split('.')[0]
+        else:
+            first_byte = default_subnet.split('.')[0]
+        (isd_id, ad_id) = isd_ad_id.split(ISD_AD_ID_DIVISOR)
+        ip_address_loc = '.'.join([first_byte, isd_id, ad_id, ER_RANGE])
+        ip_address_pub = \
+            '.'.join([first_byte, isd_id, ad_id, str(int(ER_RANGE) + 1)])
+        for link in AD_configs[isd_ad_id]["links"]:
+            er_ip_addresses[(isd_ad_id, link)] = \
```
pszal, line 113: AFAIR we try to avoid "\", especially when you have parentheses

topology/generator.py
```diff
@@ -276,65 +410,43 @@ def write_trc_files(CoreADs_tuples):
...
     Main function.
     """
-    if not os.path.isfile(ADTOISD_FILE):
-        print("ADToISD file missing.")
-        sys.exit()
-
-    if not os.path.isfile(ADRELATIONSHIPS_FILE):
-        print("ADRelationships file missing.")
+    if not os.path.isfile(ADCONFIGURATIONS_FILE):
+        print(ADCONFIGURATIONS_FILE + " file missing.")
```
pszal: switch print() to smth from logging

README.md
```diff
@@ -14,7 +14,7 @@ Necessary steps in order to run SCION:
```
pszal, line 14: It would be helpful to have in README description of certificates/keys/trcs's names, directories structure, etc...

lib/crypto/certificates.py
LorenzoBaesso: It is meant, for now, for the verification signing_cert vs signed_cert. In the next branch I will adapt it to TRCs.

PR #21, opened 2015-01-29
Routing for inter isd

pszal:
Initially tested InterISD routing.
In next commits I'm going to clean router.py and write comprehesnvie endhost test, therefore don't focus on TODOs there. Also, I think that we should change UpPath-CorePath-DownPath to e.g. seg1-seg2-seg3.

### DIFF
--- a/infrastructure/path_server.py
+++ b/infrastructure/path_server.py
@@ -90,7 +90,7 @@ class PathServer(SCIONElement):
         path_request.hdr.path.reverse()
         path_request = PathSegmentRequest(path_request.pack()) #PSz: this is
         # a hack, as path_request with <up-path> only reverses to <down-path>
-        # only, and then reversed packet fails with .get_timestamp()
+        # only, and then reversed packet fails with .get_current_iof()
         # FIXME: change .reverse() when only one path segment exists
         path = path_request.hdr.path
         path_reply = PathSegmentRecords.from_values(dst, path_request.info,
--- a/lib/packet/path.py
+++ b/lib/packet/path.py
@@ -539,10 +539,10 @@ class PathCombinator(object):
     @staticmethod
     def _build_core_path(up_segment, core_segment, down_segment):
         """
-        Joins up_, core_ and down_path into core fullpath. core_path can be
-        'None' in case of a intra-ISD core_path of length 0.
-        Returns object of CorePath class. core_path (if exists) has to have
-        down-path orientation - to discuss.
+        Joins up_, core_ and down_segment into core fullpath. core_segment can
+        be 'None' in case of a intra-ISD core_segment of length 0.
+        Returns object of CorePath class. core_segment (if exists) has to have
+        down-segment orientation.
         """
         if (not up_segment or not down_segment or
             not up_segment.ads or not down_segment.ads):
@@ -563,8 +563,7 @@ class PathCombinator(object):
 
         full_path = CorePath()
         full_path.up_segment_info = up_segment.iof
-        full_path.up_segment_info.up_flag = True #PSz: should we set flags, or
-        # whoever calls the function should care about that?
+        full_path.up_segment_info.up_flag = True
         for block in reversed(up_segment.ads):
             full_path.up_segment_hops.append(copy.deepcopy(block.pcbm.hof))
         full_path.up_segment_hops[-1].info = OpaqueFieldType.LAST_OF
--- a/lib/packet/pcb.py
+++ b/lib/packet/pcb.py
@@ -349,10 +349,9 @@ class PathSegment(Marking):
         Returns the list of HopOpaqueFields in the path.
         """
         hofs = []
-        iof = self.iof
         if reverse_direction:
             ads = list(reversed(self.ads))
-            iof.up_flag = self.iof.up_flag ^ True
+            self.iof.up_flag = self.iof.up_flag ^ True
         else:
             ads = self.ads
         for ad_marking in ads:
@@ -413,7 +412,7 @@ class PathSegment(Marking):
             pcb.iof = InfoOpaqueField(raw[0:8])
             pcb.rotf = ROTField(raw[8:16])
             raw = raw[16:]
-            for _ in range(0, pcb.iof.hops):
+            for _ in range(pcb.iof.hops):
                 pcbm = PCBMarking(raw[:PCBMarking.LEN])
                 ad_marking = ADMarking(raw[:pcbm.ssf.sig_len +
                     pcbm.ssf.block_size])
--- a/lib/packet/scion.py
+++ b/lib/packet/scion.py
@@ -125,8 +125,8 @@ class SCIONCommonHdr(HeaderBase):
         self.src_addr_len = 0  # Length of the src address.
         self.dst_addr_len = 0  # Length of the dst address.
         self.total_len = 0  # Total length of the packet.
-        self.timestamp = 0  # Offset inside the packet to the timestamp.
-        self.current_of = 0  # Index of the current opaque field.
+        self.curr_iof_p = 0  # Pointer inside the packet to the current IOF.
+        self.curr_of_p = 0  # Pointer to the current opaque field.
         self.next_hdr = 0  # Type of the next hdr field (IP protocol numbers).
         self.hdr_len = 0  # Header length including the path.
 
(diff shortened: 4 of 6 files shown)


infrastructure/path_server.py
```diff
@@ -236,9 +238,9 @@ def _propagate_down_path_segments(self, path_segments, path_info):
         for router in self.topology.routers[NeighborType.ROUTING]:
             if router.interface.neighbor_isd == self.topology.isd_id:
                 cpaths = self.core_segments(src_isd=self.topology.isd_id,
-                                            src_ad=self.topology.ad_id,
-                                            dst_isd=router.interface.neighbor_isd,
-                                            dst_ad=router.interface.neighbor_ad)
+                    src_ad=self.topology.ad_id,
```
shitz, line 241: I actually prefer aligning all arguments with the argument on the baseline. Could you revert this?

pszal, line 241: then 3rd line is above 80 chars

lib/packet/opaque_field.py
```diff
@@ -26,16 +26,19 @@ class OpaqueFieldType(object):
...
-    TDC_XOVR = 0x80
-    NON_TDC_XOVR = 0xc0
-    INPATH_XOVR = 0xe0
-    INTRATD_PEER = 0xf0
-    INTERTD_PEER = 0xf8
-    PEER_XOVR = 0x10
-    ROT_OF = 0xff
-    LAST_OF = 0x20  # Indicates last hop OF on the half-path (TODO revise)
+    # Types for HopOpaqueFields (7 MSB bits).
+    NORMAL_OF = 0b0000000
```
shitz, line 32: I think these values are very counter-intuitive. Some of them serve as flags (like the LAST_OF) and some of them are types (like INTRATD_PEER). If you interpret LAST_OF as a flag, then INTRATD_PEER would have this flag set all the time. I think we could come up with a more intuitive and cleaner way for this.

infrastructure/path_server.py
shitz, line 241: Nevermind then. Disregard my comment.

lib/packet/opaque_field.py
```diff
@@ -220,18 +229,20 @@ def pack(self):
         """
         Returns InfoOpaqueFIeld as 8 byte binary string.
         """
+        info = (self.info << 1) + self.up_flag
```
shitz, line 232: Shouldn't this be `info |= self.up_flag`? Otherwise types will get messed up when you shift everything to the left.

lib/packet/path.py
```diff
@@ -532,9 +539,10 @@ class PathCombinator(object):
     @staticmethod
     def _build_core_path(up_segment, core_segment, down_segment):
         """
-        Joins up_, core_ and down_segment into core fullpath. core_segment can
-         be 'None' in case of a intra-ISD core_segment of length 0.
-        Returns object of CorePath class.
+        Joins up_, core_ and down_path into core fullpath. core_path can be
```
shitz: You should use 'segment' instead of 'path' here for clariry.

lib/packet/opaque_field.py
syclops, line 32: Agreed. Another problem is that this file contains code to check that a certain bit (e.g. the crossover bit) is set. However, among the four crossover field types, there don't seem to be any set bits in common. (I know that PEER_XOVR is a hop OF rather than an info OF, but we should at least indicate any special bits and make sure we design OF types with those special bits in mind.)

lib/packet/path.py
```diff
@@ -555,18 +563,22 @@ def _build_core_path(up_segment, core_segment, down_segment):
 
         full_path = CorePath()
         full_path.up_segment_info = up_segment.iof
+        full_path.up_segment_info.up_flag = True #PSz: should we set flags, or
```
shitz: I think we can set the flag here. No need to place additional burden on the caller.

lib/packet/pcb.py
```diff
@@ -349,8 +349,10 @@ def get_path(self, reverse_direction=False):
         Returns the list of HopOpaqueFields in the path.
         """
         hofs = []
+        iof = self.iof
         if reverse_direction:
             ads = list(reversed(self.ads))
+            iof.up_flag = self.iof.up_flag ^ True
```
shitz: Use `^=` and I don't think you need to assign `self.iof` to a temp variable, since it's already short enough.

lib/packet/pcb.py
```diff
@@ -411,7 +413,7 @@ def deserialize(raw):
             pcb.iof = InfoOpaqueField(raw[0:8])
             pcb.rotf = ROTField(raw[8:16])
             raw = raw[16:]
-            for i in range(0, pcb.iof.hops):
+            for _ in range(0, pcb.iof.hops):
```
shitz: Drop the 0 in `range(0, pcb.iof.hops)` as it is the default value anyway.

lib/packet/scion.py
```diff
@@ -362,17 +362,13 @@ def increase_of(self, number):
...
-        """
-        self.common_hdr.type = 0
-
-    def set_downpath(self):
+    def set_downpath(self): #FIXME probably not needed
         """
         Sets down path flag.
         """
-        self.common_hdr.type = 1
+        timestamp = self.get_timestamp()
```
shitz: Ok why is the up_flag in the timestamp? This is either wrong or else really confusing.

lib/packet/scion.py
```diff
@@ -382,9 +378,22 @@ def is_on_up_path(self):
...
         the common header.
         """
-        return not self.common_hdr.type & 0x1
+        timestamp = self.get_timestamp()
+        if timestamp is not None:
+            return timestamp.up_flag
+        else:
+            True  # FIXME for now True for EmptyPath.
+
+    def is_curr_of_last(self):
```
shitz: So this returns the very last OF of the path in the packet header? I suggest renaming it to `is_last_path_of()` or something like that.

lib/packet/scion.py
```diff
@@ -382,9 +378,22 @@ def is_on_up_path(self):
...
+        """
+        Returs 'True' if the current opaque field is the last opaque field,
+        'False' otherwise.
+        """
+        offset = (SCIONCommonHdr.LEN + self.common_hdr.src_addr_len +
+                  self.common_hdr.dst_addr_len)
+        return self.common_hdr.current_of + offset == self.common_hdr.hdr_len
 
-    def reverse_direction(self):
+    def reverse_direction(self): #TO REMOVE?
```
shitz: Can be removed if not needed.

lib/packet/opaque_field.py
pszal, line 32: I also agree. The problem is that for IOFs we need both, type (that determines path's type) and direction flag, while for HOFs we need only flags. Currently logic for info field is following:
First MSB is IOF/HOF indicator, hence:
0xxxxxxx - HOF
1xxxxxxx - IOF

Then for HOF next bits are flags (continue, LAST_OF, PEER_XOVR, ...), while for IOF next bits dispatch type (TDC, NON_TDC, ...) **and** last bit denotes direction (up/down). Hence IOF is a combination of constants and flags.

I would like to hear a good suggestion how to simplify that, otherwise I will adjust implementation to this logic (introducing explicit flags to the classes, etc...)

lib/packet/opaque_field.py
pszal, line 232: see my previous comment

lib/packet/path.py
pszal: The potential issue is that _build_core_path() can build only UP-DOWN-DOWN path, but I think it is fine.

lib/packet/opaque_field.py
LorenzoBaesso, line 32: What about:
0xxx yyyy - HOF
1xxx yyyy - IOF

where the x bits are used only for flags and the y bits only for types. This wouldn't bring big advantages but perhaps it's a bit cleaner.

lib/packet/scion.py
pszal: good point. I just did some renaming.

lib/packet/opaque_field.py
pszal, line 32: It is a bit cleaner, but it does not change too much. E.g. HOF has only flags, hence yyyy in HOF does not make much sense.

Anyway, I'm going to merge the pull request, and let's fix this naming in another one.

PR #26, opened 2015-02-22
Added Certs and TRC support

LorenzoBaesso:


infrastructure/beacon_server.py
```diff
@@ -120,10 +119,14 @@ def _create_ad_marking(self, ingress_if, egress_if):
             peer_marking = \
                 PeerMarking.from_values(router_peer.interface.neighbor_ad,
                                         hof, spf)
-            pcbm.ssf.block_size += peer_marking.LEN
+            data_to_sign += str(peer_marking)
             peer_markings.append(peer_marking)
-
-        return ADMarking.from_values(pcbm, peer_markings)
+        sig_key_file = get_sig_key_file_path(self.topology.isd_id,
```
shitz: You should really load the signing key at the startup of the server instead of always reading it from the file each time you register a segment.

infrastructure/beacon_server.py
```diff
@@ -139,14 +142,17 @@ def handle_request(self, packet, sender, from_local_socket=True):
...
             logging.warning("Type not supported")
-        # TODO add ROT support etc..
 
     def run(self):
-        threading.Thread(target=self.handle_pcbs_propagation).start()
-        threading.Thread(target=self.register_segments).start()
-        SCIONElement.run(self)
+        """
+        """
+        pass
```
shitz: If you don't use it anymore, then delete it from this class. It's already defined in `SCIONElement`.

infrastructure/beacon_server.py
```diff
@@ -320,18 +329,167 @@ def register_segments(self):
...
+        else:
+            logging.info("Beacon never seen before.")
+            cert_isd = beacon.pcb.ads[-1].pcbm.spcbf.isd_id
+            cert_ad = beacon.pcb.ads[-1].pcbm.ad_id
+            trc_version = beacon.pcb.rotf.rot_version
+            if self._check_certs_trc(cert_isd, cert_ad, trc_version,
+                beacon.pcb.rotf.if_id):
+                if self._verify_beacon(beacon.pcb):
+                    self.registered_beacons.append(beacon.pcb)
+                    self.beacons.append(beacon.pcb)
```
shitz: Why do we still need `self.beacons`? Isn't a beacon either registered and therefore in `self.registered_beacons` or unregistered and therefore in `self.unregistered_beacons`?

infrastructure/beacon_server.py
```diff
@@ -320,18 +329,167 @@ def register_segments(self):
...
             time.sleep(self.config.registration_time)
 
+    def process_pcb(self, beacon):
+        """
+        Receives beacon and appends it to beacon list.
+        """
+        assert isinstance(beacon, PathConstructionBeacon)
+        logging.info("PCB received")
+        if self._is_beacon_registered(beacon.pcb):
+            logging.info("Beacon already seen before.")
```
shitz: These are not really useful comments. If you only use them for debugging purposes then use `logging.debug(...)`.

infrastructure/beacon_server.py
```diff
@@ -320,18 +329,167 @@ def register_segments(self):
...
+
+    def _check_certs_trc(self, cert_isd, cert_ad, trc_version, if_id):
+        """
+        """
+        trc_file = get_trc_file_path(cert_isd, trc_version)
+        if os.path.exists(trc_file):
+            trc = TRC(trc_file)
+            cert_file = get_cert_file_path(cert_isd, cert_ad, 0)
+            issuer = 'ISD:' + str(cert_isd) + '-AD:' + str(cert_ad)
+            while os.path.exists(cert_file):
```
shitz: What is this loop supposed to do? Find the 'last' cert_file (w/e last should mean)?

infrastructure/beacon_server.py
```diff
@@ -320,18 +329,167 @@ def register_segments(self):
...
+                cert_file = get_cert_file_path(cert_isd, cert_ad, 0)
+            if issuer in trc.core_ads:
+                return True
+            else:
+                cert_isd = int(issuer[4:].split('-AD:')[0])
+                cert_ad = int(issuer[4:].split('-AD:')[1])
+                new_cert_req = CertRequest.from_values(PT.CERT_REQ_LOCAL,
+                    self.addr, if_id, self.topology.isd_id, self.topology.ad_id,
+                    cert_isd, cert_ad, 0)
+                dst_addr = self.topology.certificate_servers[0].addr
```
shitz: Can it be that there is no cert_server? In this case the server would crash. Add error handling here.

infrastructure/beacon_server.py
```diff
@@ -320,18 +329,167 @@ def register_segments(self):
...
+                self.send(new_cert_req, dst_addr)
+                logging.info("New Certificate request sent.")
+                return False
+        else:
+            new_trc_req = TrcRequest.from_values(PT.TRC_REQ_LOCAL, self.addr,
+                if_id, self.topology.isd_id, self.topology.ad_id, cert_isd,
+                trc_version)
+            dst_addr = self.topology.certificate_servers[0].addr
+            self.send(new_trc_req, dst_addr)
+            logging.info("New TRC request sent.")
```
shitz: These are all debug comments.

infrastructure/beacon_server.py
```diff
@@ -320,18 +329,167 @@ def register_segments(self):
...
+        trc_version = pcb.rotf.rot_version
+        chain_list = []
+        for ad in pcb.ads[1:]:
+            cert = Certificate(get_cert_file_path(ad.pcbm.spcbf.isd_id,
+                ad.pcbm.ad_id, 0))
+            chain_list.insert(0, cert)
+        signature = last_ad.sig
+        subject = 'ISD:' + str(isd_id) + '-AD:' + str(ad_id)
+        chain = CertificateChain.from_values(chain_list)
+        trc = TRC(get_trc_file_path(pcb.ads[0].pcbm.spcbf.isd_id, trc_version))
```
shitz: What if the TRC doesn't exist? Is it guaranteed to exist at this point?

infrastructure/beacon_server.py
```diff
@@ -320,18 +329,167 @@ def register_segments(self):
...
+            trc_version)
+
+    def handle_unverified_beacons(self):
+        """
+        """
+        while True:
+            while self.unverified_beacons:
+                unverified_beacons = []
+                for pcb in self.unverified_beacons:
+                    cert_isd = pcb.ads[-1].pcbm.spcbf.isd_id
```
shitz: `pcb.ads[-1].pcbm` should be replaced by `pcb.get_last_ad()` (look at pcb.py).

infrastructure/beacon_server.py
```diff
@@ -320,18 +329,167 @@ def register_segments(self):
...
+                        if self._verify_beacon(pcb):
+                            self.registered_beacons.append(pcb)
+                            self.beacons.append(pcb)
+                            logging.info("Registered valid beacon.")
+                        else:
+                            logging.info("Invalid beacon.")
+                    else:
+                        logging.info("Certificate(s) or TRC missing.")
+                        unverified_beacons.append(beacon.pcb)
+                self.unverified_beacons = unverified_beacons
```
shitz: This is a dangerous operation. What if in the meanwhile other beacons got added to `self.unverified_beacons`? Then you would lose them. A better way to this would be something like this:

```
while True:
   while self.unverified_beacons:
       pcb = self.unverified_beacons.popleft()
       if verify(pcb):
            good
       else:
            self.unverified_beacons.append(pcb)
```

For this to work `self.unverified_beacons` needs to be a `deque`. You can look at how `self.beacons` and `self.reg_queue` are used.

infrastructure/router.py
```diff
@@ -501,6 +510,8 @@ def handle_request(self, packet, sender, from_local_socket=True):
             self.process_ifid_reply(packet, next_hop)
         elif ptype == PT.BEACON:
             self.process_pcb(packet, next_hop, from_local_ad)
+        elif ptype in [PT.CERT_REQ, PT.CERT_REP, PT.TRC_REQ, PT.TRC_REP]:
+            self.simple_forward(spkt, next_hop, from_local_ad)
```
shitz: `simple_forward` doesn't seem like a good name for this method. It's too generic.

infrastructure/beacon_server.py
LorenzoBaesso: We don't have self.unregistered_beacons. After a beacon arrives, it becomes unverified (not yet in self.beacons) and then it gets added to self.beacons for registration (therefore added to self.registered_beacons) and propagation.

infrastructure/beacon_server.py
LorenzoBaesso: The purpose of this function is to make sure that all the certs from the one linked to the beacon signature up to the TRC file are stored. This particular loop is to check this chain of certificates.

infrastructure/beacon_server.py
LorenzoBaesso: There must always be a cert server.

lib/packet/pcb.py
```diff
@@ -212,7 +212,7 @@ def __init__(self, raw=None):
         Marking.__init__(self)
         self.pcbm = None
         self.pms = []
-        self.sig = b''
+        self.sig = ''
```
shitz: Why do you change sig to be a string instead of bytes?

infrastructure/beacon_server.py
LorenzoBaesso: Yes, it is guaranteed. That is the purpose of _check_certs_trc and the reason why they are called in  this order.

infrastructure/beacon_server.py
LorenzoBaesso: That's right. However I'm thinking of a better name for that, given that get_last_ad = pcb.ad[x].pcbm is not very accurate. What about get_last_ads_pcbm?

lib/packet/scion.py
```diff
@@ -651,125 +659,108 @@ def from_values(cls, src, dst, path, cert_isd, cert_ad, cert_version, cert):
...
 
-    def pack(self):
-        self.payload = bitstring.pack("uintbe:16, uintbe:32, uintbe:64, " +
-            "uintbe:" + str(self.cert_len * 8), self.cert_isd,
-            self.cert_version, self.cert_ad, self.cert).bytes
-        return SCIONPacket.pack(self)
-
 
-class RotRequest(SCIONPacket):
+class TrcRequest(SCIONPacket):
```
shitz: `TRCRequest`

lib/packet/scion.py
```diff
@@ -651,125 +659,108 @@ def from_values(cls, src, dst, path, cert_isd, cert_ad, cert_version, cert):
...
-    def pack(self):
-        self.payload = bitstring.pack("uintbe:16, uintbe:32", self.rot_isd,
-            self.rot_version).bytes
-        if self.hdr.common_hdr.type != PacketType.ROT_REQ:
-            self.payload += self.path.pack()
-        return SCIONPacket.pack(self)
-
 
-class RotReply(SCIONPacket):
+class TrcReply(SCIONPacket):
```
shitz: `TRCReply`

lib/packet/scion.py
```diff
@@ -651,125 +659,108 @@ def from_values(cls, src, dst, path, cert_isd, cert_ad, cert_version, cert):
...
 
     def __init__(self, raw=None):
         SCIONPacket.__init__(self)
-        self.rot_isd = 0
-        self.rot_version = 0
-        self.rot = b''
-        self.rot_len = 0
+        self.trc_isd = 0
+        self.trc_version = 0
+        self.trc = ''
```
shitz: I really think we should use bytes on a packet level.

infrastructure/router.py
LorenzoBaesso: relay_packet?

lib/packet/pcb.py
LorenzoBaesso: TO be consistent with other classes like Certificate and TRC. Plus normally our classes' variables are not bytes.

infrastructure/beacon_server.py
shitz: It would make more sense if it would go like this:
Beacon arrives -> gets put in `self.unverified_beacons`
Beacon gets verified -> gets put in `self.unregistered_beacons`
Beacon gets registered -> gets put in `self.registered_beacons`

Btw. I don't think every beacon should get registered. If a previously registered path-segment is still valid for the segment that is contained in the beacon, then the BS shouldn't register it and instead just propagate it.

lib/packet/scion.py
LorenzoBaesso: Here it would maybe make more sense but I think I use it somewhere else where a str is more convenient. I'll keep it in mind though.

infrastructure/beacon_server.py
shitz: So then add a comment in the code, since it might be not clear to people reading this particular code segment.

infrastructure/beacon_server.py
shitz: Then we can get rid of that accessor altogether imho. So far it was a short-cut (as in less actual text to write) to access the last ads pcbm, since that was the common access pattern.

infrastructure/beacon_server.py
LorenzoBaesso: A beacon is unverified until it's registered (or discarded if the verification fails), so the first two containers are the same. I agree on skipping the registrations for registered beacons though.

infrastructure/beacon_server.py
LorenzoBaesso: Yeah I'll put it in the doc section. :)

infrastructure/router.py
shitz: This is also generic. It seems this function is specifically used to forward stuff to/from the cert_server, so this fact should be reflected in the method name. Otherwise you have to make it more generic that it also works for other types of servers.

lib/packet/pcb.py
shitz: I think we use bytes throughout the packet library. Also signatures are semantically bytes and not strings. Anyway, discuss this with Pawel.

infrastructure/beacon_server.py
LorenzoBaesso: Agreed. I'll remove the accessor.

infrastructure/router.py
LorenzoBaesso: right..relay_cert_server_packet?

infrastructure/beacon_server.py
LorenzoBaesso: Actually I think I will leave it because get_last_ad and get_first_ad are safer than accessing directly ads[x].pcbm. I will use it in my code too. Although the name is still wrong imho.

infrastructure/beacon_server.py
LorenzoBaesso: True that. However between the while self.unverified_beacons: and self._verify_beacon(pcb): there's an if that changes things. Meaning that some beacons, according to that if, might need to stay in the unverified_beacons queue. For those, I could create a temporary queue and AFTER the while self.unverified_beacons loop, append the temporary queue to the unverified_beacons queue. Would you say it's a good solution?

LorenzoBaesso: Sorry for spamming, but just to clarify. It would become:

```
while True:
            temp_queue = deque()
            while self.unverified_beacons:
                pcb = self.unverified_beacons.popleft()
                cert_isd = pcb.get_last_ad().spcbf.isd_id
                cert_ad = pcb.get_last_ad().ad_id
                trc_version = pcb.rotf.rot_version
                if self._check_certs_trc(cert_isd, cert_ad, trc_version,
                    pcb.rotf.if_id):
                    if self._verify_beacon(pcb):
                        self.registered_beacons.append(pcb)
                        self.beacons.append(pcb)
                        logging.info("Registered valid beacon.")
                    else:
                        logging.info("Invalid beacon.")
                else:
                    logging.debug("Certificate(s) or TRC missing.")
                    temp_queue.append(beacon.pcb)
            self.unverified_beacons.extend(temp_queue)
            time.sleep(BeaconServer.TIME_INTERVAL)
```

shitz: Why a temporary queue? Couldn't you just append it to `self.unverified_beacons` again? Since you always popleft and append-right, it should work just fine.

infrastructure/beacon_server.py
shitz: I would be fine with `get_last_pcbm`.

infrastructure/beacon_server.py
LorenzoBaesso: There's a synchronization issue. _check_certs_trc will send CertReq and TRCReq until the chain is complete. Therefore appending the unverified beacons inside the loop will just keep the CPU busy.

shitz: But it doesn't make a difference to the code you suggested. When you have processed everything in `self.unverified_beacons` you will add the ones that couldn't be verified back and start the loop again. You won't get into sleep, but that sleep seems rather arbitrary to me anyway. It could very well be that in the meantime you received the necessary certs and trc files to verify the beacon and you are wasting time by sleeping on this thread.

Actually you should keep track of the trc/cert request you already sent out, since now it could be that you send multiple requests for the same files (bc _check_certs_trc will request them no matter what).

infrastructure/beacon_server.py
```diff
@@ -320,18 +326,173 @@ def register_segments(self):
...
+    def _is_beacon_registered(self, pcb):
+        """
+        """
+        assert isinstance(pcb, PathSegment)
+        for reg_pcb in self.registered_beacons:
+            if reg_pcb.compare_hops(pcb):
+                return True
+        return False
+
+    def _check_certs_trc(self, isd_id, cert_ad, trc_version, if_id):
```
pszal: probably we will use 3 following methods also in other places (than BS)

infrastructure/beacon_server.py
```diff
@@ -320,18 +326,173 @@ def register_segments(self):
...
+            str(last_ad.pcbm.spcbf))
+        for peer_marking in last_ad.pms:
+            data_to_verify += str(peer_marking)
+        return verify(data_to_verify, signature, subject, chain, trc,
+            trc_version)
+
+    def handle_unverified_beacons(self):
+        """
+        """
+        while True:
```
pszal: this while seems to take 100% CPU. Use sleep()

pszal: ... also isn't easier to handle these unverified beacons after receiving cert/trc (at the end on process_trc/cert_rep()) ?

infrastructure/cert_server.py
```diff
@@ -52,133 +44,109 @@ def process_cert_request(self, cert_req):
...
+                cert_req.ingress_if, cert_req.src_isd, cert_req.src_ad,
+                cert_req.cert_isd, cert_req.cert_ad, cert_req.cert_version)
+            dst_addr = self.ifid2addr[cert_req.ingress_if]
             self.send(new_cert_req, dst_addr)
+            logging.info("New certificate request sent.")
         else:
-            logging.info('Certificate %s:%s found, sending it back to ' +
-                'requester(%s)', cert_isd, cert_ad, src_addr)
+            logging.info('Certificate file found.')
             with open(cert_file, 'r') as file_handler:
```
pszal: we should have get_cert_from_file() that includes get_cert_file_path() and this open()+read().
We will use this a lot. Also don't we have a TrustStorage class or smth similar to cache certs and TRCs ?

infrastructure/router.py
```diff
@@ -1,32 +1,27 @@
...
 from infrastructure.scion_elem import SCIONElement, SCION_UDP_PORT
 from lib.packet.host_addr import IPv4HostAddr
-from lib.packet.opaque_field import OpaqueField
-from lib.packet.opaque_field import OpaqueFieldType as OFT
+from lib.packet.opaque_field import OpaqueField, OpaqueFieldType as OFT
 from lib.packet.pcb import PathConstructionBeacon
-from lib.packet.scion import PacketType as PT
-from lib.packet.scion import SCIONPacket, IFIDRequest, IFIDReply, get_type
+from lib.packet.scion import (PacketType as PT, SCIONPacket, IFIDRequest,
+    IFIDReply, get_type, CertRequest, CertReply, TRCRequest, TRCReply)
```
pszal: remove unused imports

infrastructure/cert_server.py
```diff
@@ -1,49 +1,41 @@
...
 from infrastructure.scion_elem import SCIONElement
 from lib.packet.path import EmptyPath
 import sys
 import os
 import logging
-
-
-ISD_PATH = '../topology/ISD'
-CERTS_PATH = '/certificates/'
+import base64
```
pszal: where it is used?

lib/crypto/certificate.py
```diff
@@ -16,7 +16,8 @@
 ===========================================
 """
 
-from lib.crypto.asymcrypto import *
+from lib.crypto.asymcrypto import sign
+from lib.crypto.nacl import crypto_sign_ed25519_open
```
pszal, line 20: hmm. is crypto_sign_ed25519_open() an inverse of sign() ?

lib/crypto/trc.py
```diff
@@ -69,27 +71,31 @@ def __init__(self, trc_file=None):
...
             self.parse(trc_file)
 
-    def get_trc_dict(self):
+    def get_trc_dict(self, with_signatures=False):
         """
         Return the TRC information.
 
+        :param with_signatures: True or False whether the returned data should
+                                contain the signatures section or not.
+        :type with_signatures: bool
```
pszal: what is meaning of with_signatures?

lib/packet/pcb.py
pszal: We use bytes by default. Also the output of signing, key generation, should give bytes.
I just noticed that generate_*_keypair() returns ASCII what is awkward . Is there any specific reason to do so?

LorenzoBaesso: Keys and signatures are stored into files and used in higher level classes. If these functions return bytes, then I will have to apply encode/decode everywhere else. My idea was that sig/keys could be strings for easier usage in the classes/files/etc and these crypto functions internally would work with bytes. Should I change every sig/key in every class into bytes then?

infrastructure/beacon_server.py
LorenzoBaesso: We can't do that only if a cert/TRC rep was recived, if this is what you mean. Otherwise unverified beacons for which we have the certs will not be processed.

lib/packet/pcb.py
pszal: Look at the generate_signature_keypair(). It should return bytes. For files use base64.

infrastructure/cert_server.py
LorenzoBaesso: I will add the get_cert_from_file(). I will add a TrustStorage when we'll have certs/TRCs updates. For now let's keep the files solution. Plus they are read only once per path.

lib/crypto/certificate.py
LorenzoBaesso, line 20: Sort of, it returns the plaintext (in bytes) of the message with signature (if the verification succeeds).

lib/crypto/trc.py
LorenzoBaesso: Sometimes I need the cert dictionary with the signatures section, sometimes without.

lib/packet/pcb.py
shitz: That has bothered me for a long time as well. It just doesn't make sense to me to handle everything using ascii strings. I agree with Pawel that, in code, signatures/hashes/MACs etc should be bytes and if you need to store them to a file then use some encoding (e.g., base64).

infrastructure/cert_server.py
LorenzoBaesso: Actually we could have a str = read_file(file_path) and write_file(file_path, str) for general purpose. Shall I place these two in scion.py, or where's best?

shitz: Not in scion.py please. scion.py should only have packet related stuff in it (more specifically scion-packet related stuff). If you want to introduce these functions the put them in `lib.util.py`.

LorenzoBaesso: Ah right! In there I found:

#FIXME to remove, as we don't need that anymore with python's SCIONDaemon
def get_paths(dst_isd, src_ad, dst_ad):

...shall I remove it or does anyone still need that?

pszal: yes, please remove that.

infrastructure/beacon_server.py
```diff
@@ -320,18 +328,197 @@ def register_segments(self):
...
+                logging.debug("Certificate(s) or TRC missing.")
+                self.unverified_beacons.append(beacon.pcb)
+
+    def _is_beacon_registered(self, pcb):
+        """
+        Return True or False whether a beacon was previously registered.
+        """
+        assert isinstance(pcb, PathSegment)
+        for reg_pcb in self.registered_beacons:
+            if reg_pcb.compare_hops(pcb):
```
shitz: I don't think this should be based on comparing AD-level hops. If some ADs have multiple links between them, then this would evaluate to true even though the beacons are for different paths. I introduced `get_hops_hash()` that returns a hash over all the interface revocation tokens of a segment. I think you should use that here.

infrastructure/beacon_server.py
```diff
@@ -320,18 +328,197 @@ def register_segments(self):
...
+        Return True or False whether the necessary Certificate and TRC files are
+        found.
+        """
+        trc_file = get_trc_file_path(self.topology.isd_id, self.topology.ad_id,
+            isd_id, trc_version)
+        if os.path.exists(trc_file):
+            trc = TRC(trc_file)
+            cert_file = get_cert_file_path(self.topology.isd_id,
+                self.topology.ad_id, isd_id, cert_ad, cert_version)
+            my_cert_file = get_cert_file_path(self.topology.isd_id,
```
shitz: Please don't use variable names that begin with `my_`. That is fine for some programming exercises but not for production code...

infrastructure/beacon_server.py
```diff
@@ -320,18 +328,197 @@ def register_segments(self):
...
+                now = int(time.time())
+                if cert_tuple not in self.requested_certs :
+                    new_cert_req = CertRequest.from_values(PT.CERT_REQ_LOCAL,
+                        self.addr, if_id, self.topology.isd_id,
+                        self.topology.ad_id, isd_id, cert_ad, cert_version)
+                    dst_addr = self.topology.certificate_servers[0].addr
+                    self.send(new_cert_req, dst_addr)
+                    self.requested_certs[cert_tuple] = now
+                    return False
+                if (now - self.requested_certs[cert_tuple] >
```
shitz: You do two times exactly the same thing. Change the first if condition to 
`if cert_tuple not in self.requested_certs or now - self.requested_certs[cert_tuple] > TIME_OUT`.

infrastructure/beacon_server.py
```diff
@@ -320,18 +328,197 @@ def register_segments(self):
...
+            now = int(time.time())
+            if trc_tuple not in self.requested_trcs:
+                new_trc_req = TRCRequest.from_values(PT.TRC_REQ_LOCAL,
+                    self.addr, if_id, self.topology.isd_id, self.topology.ad_id,
+                    isd_id, trc_version)
+                dst_addr = self.topology.certificate_servers[0].addr
+                self.send(new_trc_req, dst_addr)
+                self.requested_trcs[trc_tuple] = now
+                return False
+            if (now - self.requested_trcs[trc_tuple] >
```
shitz: same here about the if condition

infrastructure/beacon_server.py
```diff
@@ -320,18 +328,197 @@ def register_segments(self):
...
 
+    def process_pcb(self, beacon):
+        """
+        Receives beacon and appends it to beacon list.
+        """
+        assert isinstance(beacon, PathConstructionBeacon)
+        logging.info("PCB received")
+        if self._is_beacon_registered(beacon.pcb):
+            logging.debug("Beacon already seen before.")
+            self.beacons.append(beacon.pcb)
```
shitz, line 504: Why do you do this? What's the difference between the `beacons` and `registered_beacons` lists?

infrastructure/beacon_server.py
```diff
@@ -276,10 +279,16 @@ class LocalBeaconServer(BeaconServer):
...
+    REQUESTS_TIMEOUT = 10
+
     def __init__(self, addr, topo_file, config_file):
         BeaconServer.__init__(self, addr, topo_file, config_file)
         # Sanity check that we should indeed be a local beacon server.
         assert not self.topology.is_core_ad, "This shouldn't be a local BS!"
+        self.unverified_beacons = deque()
+        self.registered_beacons = []
+        self.requested_certs = {}
+        self.requested_trcs = {}
```
shitz, line 336: Just for me to understand: Why don't core beacon-servers have to do the same for core-segment construction?

infrastructure/beacon_server.py
```diff
@@ -320,18 +328,197 @@ def register_segments(self):
...
+        Process the Certificate reply.
+        """
+        assert isinstance(cert_rep, CertReply)
+        logging.info("Certificate reply received.")
+        cert_isd = cert_rep.cert_isd
+        cert_ad = cert_rep.cert_ad
+        cert_version = cert_rep.cert_version
+        cert_file = get_cert_file_path(self.topology.isd_id,
+            self.topology.ad_id, cert_isd, cert_ad, cert_version)
+        write_file(cert_file, cert_rep.cert.decode('ascii'))
```
shitz: ascii? Shouldn't we use utf-8?

infrastructure/beacon_server.py
```diff
@@ -320,18 +328,197 @@ def register_segments(self):
...
+        self.handle_unverified_beacons()
+
+    def _verify_beacon(self, pcb):
+        """
+        Once the necessary certificate and TRC files have been found, verify the
+        beacons.
+        """
+        assert isinstance(pcb, PathSegment)
+        last_ad = pcb.ads[-1]
+        cert_isd = last_ad.pcbm.spcbf.isd_id
```
shitz: Maybe use `pcb.get_last_pcbm()` here as well?

infrastructure/beacon_server.py
```diff
@@ -320,18 +328,197 @@ def register_segments(self):
...
+        cert_file = get_cert_file_path(self.topology.isd_id,
+            self.topology.ad_id, cert_isd, cert_ad, cert_version)
+        if os.path.exists(cert_file):
+            chain = CertificateChain(cert_file)
+        else:
+            chain = CertificateChain.from_values([])
+        trc_file = get_trc_file_path(self.topology.isd_id, self.topology.ad_id,
+            cert_isd, trc_version)
+        trc = TRC(trc_file)
+        data_to_verify = (str(cert_ad) + str(last_ad.pcbm.hof) +
```
shitz: This will break if `__str__` of any of these classes will be changed in the future (that could very well happen). Verification should be done on the byte representation of these objects (as returned by `pack()`) I think.

infrastructure/beacon_server.py
```diff
@@ -320,18 +328,197 @@ def register_segments(self):
...
+        return verify(data_to_verify, last_ad.sig, subject, chain, trc,
+            trc_version)
+
+    def handle_unverified_beacons(self):
+        """
+        Handle beacons which are waiting to be verified.
+        """
+        for index in range(len(self.unverified_beacons)):
+            pcb = self.unverified_beacons.popleft()
+            cert_isd = pcb.get_last_pcbm().spcbf.isd_id
```
shitz: This is code duplication. You do the exact same thing already in `process_pcb`.

infrastructure/beacon_server.py
```diff
@@ -320,18 +328,197 @@ def register_segments(self):
...
+        for peer_marking in last_ad.pms:
+            data_to_verify += str(peer_marking)
+        return verify(data_to_verify, last_ad.sig, subject, chain, trc,
+            trc_version)
+
+    def handle_unverified_beacons(self):
+        """
+        Handle beacons which are waiting to be verified.
+        """
+        for index in range(len(self.unverified_beacons)):
```
shitz: If you don't need `index` you can just write `for _ in ...`

lib/crypto/certificate.py
```diff
@@ -16,35 +16,75 @@
...
-        for name in files:
-            if name.endswith((".crt")):
-                cert = Certificate(path + name)
-                roots[cert.subject + '-V:' + str(cert.version)] = cert
-    return roots
+    sig = crypto_sign_ed25519(msg, signing_key)[:64]
+    return sig
+
+
+def verify(msg, sig, subject, chain, trc, trc_version):
```
shitz: Honestly, I'm not a big fan of moving those functions here... From what I understand is that `asymcrypto.py` should be the API for asymmetric crypto, i.e., encrypt/decrypt and sign/verify. As you moved sign/verify here, this is no longer true, which also means that our crypto library as it is now has some conceptual flaws. It should be possible that `aymcrypto.py` is at the very bottom of the hierarchy (importing everything that is need for these 4 operations). As far as I can see you just moved them here because you had some cyclic imports and this was the most straightforward way to break it. In my opinion this is the wrong way to do it...

lib/crypto/certificate.py
```diff
@@ -362,7 +412,249 @@ def __str__(self):
...
+                base64.b64encode(cert_dict['subject_enc_key']).decode('ascii')
+            cert_dict['signature'] = \
+                base64.b64encode(cert_dict['signature']).decode('ascii')
+            chain_dict[index] = cert_dict
             index += 1
         chain_str = json.dumps(chain_dict, sort_keys=True, indent=4)
         return chain_str
+
+
+class TRC(object):
```
shitz, line 409: Is there a reason why you moved `TRC` to `certificates.py` and deleted `trc.py`. I preferred it the other way.

infrastructure/beacon_server.py
LorenzoBaesso: I'll add it to the issue previously mentioned. I can't use that now because it was introduced later.

shitz: That's not true. You need to merge the current master branch in your branch anyway and it exists already in the current master branch.

infrastructure/beacon_server.py
LorenzoBaesso: True. Here I had a doubt though. Is the first condition always checked as first in python? Otherwise there would be an index error...

infrastructure/beacon_server.py
LorenzoBaesso, line 504: beacons is a temporary buffer used for propagation and registration..it can be emptied

infrastructure/beacon_server.py
shitz: Yes it's guaranteed, see https://docs.python.org/2/library/stdtypes.html#boolean-operations-and-or-not

infrastructure/beacon_server.py
LorenzoBaesso, line 336: Underneath ADs rely on trustworthy core ADs. Core ADs are supposed to be trustworthy by default, at least in principle.

infrastructure/beacon_server.py
LorenzoBaesso: Not sure, did we decide to use a particular encoding scheme?

shitz: Actually, I think we said that we'll used base64 for writing it to files.

infrastructure/beacon_server.py
LorenzoBaesso: Yes that's what I meant but I'd rather merge, solve conflicts (push) and later modify things but ok..

shitz: You are still doing it in this pull-request. By adding it to an issue you indicate that you will do it in a  later pull-request.

LorenzoBaesso: That was my intent. :)

infrastructure/beacon_server.py
LorenzoBaesso: For signatures and public keys, but not for certificates..

infrastructure/beacon_server.py
shitz: Yea well since it is a flaw in this pull-request and you are not dependent on anything that hasn't been added to the master branch yet, it really doesn't make sense to postpone it...

infrastructure/beacon_server.py
LorenzoBaesso: process_pcb is for just arrived beacons, while handle_unverified_beacons is for "old" beacons.

lib/crypto/certificate.py
LorenzoBaesso: That verify is actually very connected to the verification of signatures whose corresponding public keys are in the certificates. So, in my opinion, it makes more sense to have it in certificate.py than in asymcrypto.py. Honestly I think it was wrong to move it there in the first place because it's not a general verify(msg, sig, key).

infrastructure/beacon_server.py
shitz: That's not the point. You do the exact same steps here as in `process_pcb` so why don't you create a function that does these steps and call it from `process_pcb` and here.

lib/crypto/certificate.py
LorenzoBaesso: I would suggest to keep this verify here (rename it perhaps), move the sign back in asymcrypto (that would make sense) and open an issue for tenma marking that a general verify function is needed. Sounds good?

shitz: That makes sense. I would rename this verify to something more specific, since we already have verify in the classes and will have one in asymcrypto.

infrastructure/beacon_server.py
LorenzoBaesso: Yes but for 10 lines of code (in which 5 are either else or logging) I need to create a new function, name it in a way that makes sense with what I would do in it(checking the existence of the chain, checking the verification, etc). Not to mention that in the future these two segments of code might be different. I don't know, I just think it's too pushed here...

lib/crypto/certificate.py
LorenzoBaesso, line 409: certificates, certificate chains and TRCs are all related and somehow they're all certificates.

infrastructure/beacon_server.py
shitz: I strongly disagree. 
1) It's about 15 lines of code, we have many methods that have less than that and are used less than this code segment. It increases readability of the code.
2) It's unlikely that they will ever differ from each other since you are doing the exact same thing, namely trying to verify a beacon (this would also make for a good method name). If the process of verifying a beacon changes then both code segments would need to be changed.
3) It's bad coding practice to have the exact same lines of code at several places in the code. If you can save some lines of code, it will make the quality of code higher in general, especially if you keep doing this consistently. It will add up.

infrastructure/beacon_server.py
pszal, line 336: That's good point @shitz. We should have similar mechanism for core segments

infrastructure/beacon_server.py
LorenzoBaesso: Very good points. I'll do as suggested. 

---- shitz wrote ----

infrastructure/beacon_server.py
pszal: Hmm. For signature? Is PCB's signature in base64? It should be in bytes.
Base64 we use only (CMIIW) for certificate and TRC fields.

LorenzoBaesso: PCB's signatures are in bytes. Certificates and TRCs signatures are stored in base64 encoding to files. This piece of code is about writing a certificate to a file.

LorenzoBaesso: For completeness of the answer, certificates and TRCs are also in bytes (in the reply packets).

PR #28, opened 2015-02-23
Interface and Path-Segment IDs

shitz:
Added unique identifiers to interfaces and path-segments. These will be subsequently used in path revocation.

### DIFF
--- a/lib/packet/pcb.py
+++ b/lib/packet/pcb.py
@@ -64,7 +64,8 @@ class PCBMarking(Marking):
     """
     Packs all fields for a specific PCB marking, which includes: the Autonomous
     Domain's ID, the SupportSignatureField, the HopOpaqueField, the
-    SupportPCBField, and the hashes for the interfaces included in the HOF.
+    SupportPCBField, and the revocation tokens for the interfaces
+    included in the HOF.
     """
     LEN = 32 + 2 * 32
 
@@ -74,8 +75,8 @@ class PCBMarking(Marking):
         self.ssf = None
         self.hof = None
         self.spcbf = None
-        self.ingress_hash = 32 * b"\x00"
-        self.egress_hash = 32 * b"\x00"
+        self.ig_rev_token = 32 * b"\x00"
+        self.eg_rev_token = 32 * b"\x00"
         if raw is not None:
             self.parse(raw)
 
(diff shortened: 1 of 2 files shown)


lib/packet/pcb.py
```diff
@@ -82,7 +84,7 @@ def parse(self, raw):
         Populates fields from a raw bytes block.
         """
         assert isinstance(raw, bytes)
-        self.raw = raw
+        self.raw = raw[:]
```
LorenzoBaesso, line 88: Out of curiosity, what's the difference in this case?

topology/ADConfigurations.json
```diff
@@ -2,9 +2,9 @@
     "default_subnet": "127.0.0.0/8",
     "1-11": {
         "level": "CORE",
-        "beacon_servers": 2,
-        "certificate_servers": 3,
-        "path_servers": 2,
+        "beacon_servers": 1,
```
LorenzoBaesso, line 5: In the next branch I'll implement the beacon propagation logic in case of multiple beacons. So I will need to change this again. Is there a particular reason why it should be set to 1? (Same for the other servers)

lib/packet/pcb.py
shitz, line 88: It makes a copy of `raw`. `self.raw` should always contain the entire byte-string. `raw` gets modified throughout the parsing code, so we either have to create a copy of `raw` first or we don't modify it and instead work with offsets.

topology/ADConfigurations.json
shitz, line 5: Actually, I think we should really talk about this. We haven't decided yet on how to handle multiple instances of servers and it might very well be that if we use something like ZooKeeper or OpenReplica for path-servers, we also use it for beacon-servers. So for now everything will be developed for the one-instance-per-server case as far as i'm aware.

shitz, line 5: And besides, as long as the logic for multiple servers isn't implemented (which it isn't at the moment), the default configuration (in the master branch) shouldn't use multiple instances of servers.

pszal, line 5: No comments from my side, only a suggestion: "interface hash" may be
confusing for unfamiliar readers, e.g. "revocation token" sounds better
to me.

Best,
Pawel

shitz, line 5: That's a good point. I will change the name and proceed to merge this pull request.

PR #29, opened 2015-02-23
Created log files.

LorenzoBaesso:


infrastructure/beacon_server.py
```diff
@@ -352,6 +355,17 @@ def main():
...
         sys.exit()
+
+    isd_id = str(beacon_server.topology.isd_id)
+    ad_id = str(beacon_server.topology.ad_id)
+    ip_addr = str(beacon_server.addr)
+    log_file = '/'.join(['../logs', str(datetime.date.today()), 'ISD' + isd_id,
+        'AD' + ad_id, 'bs' + isd_id + '-' + ad_id + '-' + ip_addr + '.log'])
+    if not os.path.exists(os.path.dirname(log_file)):
+        os.makedirs(os.path.dirname(log_file))
+    logging.getLogger('').handlers = []
```
shitz, line 366: Why is this needed?

infrastructure/beacon_server.py
```diff
@@ -352,6 +355,17 @@ def main():
...
+    logging.basicConfig(filename=log_file, filemode='w', level=logging.DEBUG)
```
shitz, line 367: `filemode` should be 'a' (append) otherwise the logs will always be overwritten.

infrastructure/beacon_server.py
LorenzoBaesso, line 366: Clearing the handlers is needed when the logging facility has already been configured (see above). And we can't remove the first basicConfig because logging is used in the server inits, and therefore automatically configured with the default settings.

infrastructure/beacon_server.py
LorenzoBaesso, line 367: That's exactly what I think it's best to have because after a few seconds we run SCION there's already a number of MBs more because of the logs. Plus I think it's confusing in the debugging scenario to have logs from previous runs.

shitz, line 367: I agree with the debugging scenario, but otherwise this is not really like logs work...

PR #31, opened 2015-03-02
Expiring caches and new expiration of hops implemented

shitz:
- Implemented new expiration of path-segments based on creation timestamp + hop expiration fields
- Path Servers now evict expired path segments from their caches

### DIFF
--- a/infrastructure/beacon_server.py
+++ b/infrastructure/beacon_server.py
@@ -20,7 +20,6 @@ from _collections import deque
 import copy
 import datetime
 from infrastructure.scion_elem import SCIONElement
-from lib.defines import SCION_SECOND, MAX_SEGMENT_TTL
 from lib.packet.host_addr import IPv4HostAddr
 from lib.packet.opaque_field import (OpaqueFieldType as OFT, InfoOpaqueField,
     SupportSignatureField, HopOpaqueField, SupportPCBField, SupportPeerField,
@@ -30,7 +29,6 @@ from lib.packet.pcb import (PathSegment, ADMarking, PCBMarking, PeerMarking,
     PathSegmentType as PST)
 from lib.packet.scion import SCIONPacket, get_type, PacketType as PT
 import logging
-import os
 import sys
 import threading
 import time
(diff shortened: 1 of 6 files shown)


infrastructure/beacon_server.py
```diff
@@ -17,7 +17,10 @@
 """
 
 from _collections import deque
+import copy
+import datetime
 from infrastructure.scion_elem import SCIONElement
+from lib.defines import SCION_SECOND, MAX_SEGMENT_TTL
```
pszal: Why we need SCION_SECONDs ? It was introduced for SIBRA's expiration time. With absolute timestamp + exp time in HOFs we don't need that.

infrastructure/beacon_server.py
```diff
@@ -49,8 +50,9 @@ class BeaconServer(SCIONElement):
         reg_queue: A FIFO queue containing paths for registration with path
             servers.
     """
-    DELTA = 24 * 60 * 60  # Amount of real time a PCB packet is valid for.
-    TIME_INTERVAL = 4  # SCION second
+    # Amount of real time a HOF is valid realtive to MAX_TTL.
+    HOF_EXP_TIME = int(((6 * 60 * 60) / MAX_SEGMENT_TTL) * (2 ** 8)) - 1
```
pszal: Hmm, I'm not sure why we need this constant.
Also typo in above line "relative".

lib/packet/opaque_field.py
```diff
@@ -171,7 +179,7 @@ class InfoOpaqueField(OpaqueField):
     Class for the info opaque field.
 
     The info opaque field contains type info of the path-segment (1 byte),
-    an expiration timestamp (2 bytes), the ISD ID (2 byte), # hops for this
+    a creation timestamp (2 bytes), the ISD ID (2 byte), # hops for this
```
pszal: timestamp should be 4B (we should use 2B which are currently "reserved").

infrastructure/beacon_server.py
shitz: This basically says that a HOF is valid for 6 hours after creation. We should definitely make this configurable, but for now it is just a constant.

infrastructure/beacon_server.py
shitz: Well I don't know why we use SCION_SECONDS. It increases the range for the timestamp while giving up accuracy. If you say we can remove it then I'm fine with that. Have you ever thought about what to do when the timestamp reaches it's maximum capacity?

lib/packet/opaque_field.py
shitz: That makes sense, but do we have to run it by Adrian first?

infrastructure/beacon_server.py
pszal: hmm, actually I have not. This will happen in ~100yrs, but we can optimize starting SCION Epoch in 2015 (it will give 45yrs more ;-)

shitz: Hmm now with a 4B timestamp this will happen in 2038 if we use the UNIX epoch ;)

PR #36, opened 2015-03-05
Fidelity

syclops:
Path fidelity is not yet working, but we will probably need to make major changes to path_store.py in order to change how this works (probably involving moving PathSegmentInfo functionality into PathStore to provide more information required for evaluating path fidelity).

However, I've made major changes to the way candidate paths are stored in PathStore, namely, using SortedCollection, which is an external code recipe based on the bisect module. This allows us to keep all candidate paths sorted in a list without writing that functionality in our own code.

### DIFF
--- a/lib/path_store.py
+++ b/lib/path_store.py
@@ -96,22 +96,20 @@ class PathStoreRecord(object):
     :vartype id: bytes
     :ivar fidelity: the fidelity of the path record.
     :vartype fidelity: float
-    :ivar peer_links: the number of peering links in the path segment.
-    :vartype peer_links: int
-    :ivar hops_length: the length of the path segment in hops.
-    :vartype hops_length: int
-    :ivar disjointness: the disjointness of the path segment compared to the
-       other paths in the PathStore database.
-    :vartype disjointness: int
-    :ivar last_sent_time: the Unix time at which the path segment was last
-       sent.
+    :ivar peer_links: the normalized number of peer links in the path segment.
+    :vartype peer_links: float
+    :ivar hops_length: the normalized length of the path segment.
+    :vartype hops_length: float
+    :ivar disjointness: the normalized disjointness of the path segment compared
+                        to the other paths in the PathStore.
+    :vartype disjointness: float
+    :ivar last_sent_time: the Unix time at which the path segment was last sent.
     :vartype last_sent_time: int
-    :ivar last_seen_time: the Unix time at which the path segment was last seen
-       by the path server.
+    :ivar last_seen_time: the Unix time at which the path segment was last seen.
     :vartype last_seen_time: int
-    :ivar delay_time: the time in seconds between the PCB's creation and the
-       time it was last seen by the path server.
-    :vartype delay_time: int
+    :ivar delay_time: the normalized time in seconds between the PCB's creation
+                      and the time it was last seen by the path server.
+    :vartype delay_time: float
     :ivar expiration_time: the Unix time at which the path segment expires.
     :vartype expiration_time: int
     :ivar guaranteed_bandwidth: the path segment's guaranteed bandwidth.


infrastructure/scion_elem.py
```diff
@@ -51,7 +52,8 @@ class SCIONElement(object):
     :vartype addr: :class:`lib.packet.host_addr.HostAddr`
     """
 
-    def __init__(self, addr, topo_file, config_file=None, trc_file=None):
+    def __init__(self, addr, topo_file, config_file=None, trc_file=None,
+        path_policy_file=None):
```
pszal: I'm not sure why **path_policy/_file** belongs to SCIONElem, especially when SCIONElem does not have any instance of PathStore. I can imagine that we may want to use PathStore at endhost, however currently PathStore is intended only for beacon servers, right?

LorenzoBaesso: Yes, for now we're using/testing the PathStore functionality at the Beacon Servers.

infrastructure/beacon_server.py
```diff
@@ -67,8 +68,9 @@ class BeaconServer(SCIONElement):
     BEACONS_NO = 5
     REGISTERED_PATHS = 100
 
-    def __init__(self, addr, topo_file, config_file):
-        SCIONElement.__init__(self, addr, topo_file, config_file)
+    def __init__(self, addr, topo_file, config_file, path_policy_file):
+        SCIONElement.__init__(self, addr, topo_file, config_file=config_file,
+                              path_policy_file=path_policy_file)
```
shitz: `path_policy_file` shouldn't be a part of `SCIONElement`, since only the beacon server uses it.

infrastructure/beacon_server.py
```diff
@@ -452,15 +446,65 @@ def _check_certs_trc(self, isd_id, cert_ad, cert_version, trc_version,
...
         """
-        info = PathSegmentInfo.from_values(PST.UP,
-                                           self.topology.isd_id,
-                                           self.topology.isd_id,
-                                           pcb.get_first_pcbm().ad_id,
-                                           self.topology.ad_id)
+        # Store path
+        path_store_record = PathStoreRecord(pcb)
+        self.up_segments.add_record(path_store_record)
+        logging.debug(self.up_segments)
```
shitz: This shouldn't make it into master.

infrastructure/beacon_server.py
LorenzoBaesso: But this might not be case when the endhost is going to make use of PathStore; for example, the end2end test uses SCIONDaemon, and this calls SCIONElement.**init**(self, addr, topo_file). The path_policy_file could be another argument in that init. If we decide not to do so, then the same is valid for trc_file (because it's used only by the cert server). So, in the latter case, I will just pass the needed file to the cert/beacon server and do the parsing after the SCIONElement init. Shall I go for the second option then?

lib/path_store.py
```diff
@@ -1,218 +1,182 @@
...
+        return path_policy_str
 
 
-class PathSegmentInfo(object):
+class PathStoreRecord(object):
     """
-    Stores general information about a path.
+    Path record that gets stored in the the PathStore.
+
+    :ivar pcb: the PCB representing the record.
```
shitz, line 93: I thought the doc format was `:param X: <desc> :type: type(X)`?

syclops, line 93: For functions, yes. For data members of classes, we use the above, which represents an instance variable.

infrastructure/beacon_server.py
syclops: Er...any reason why not?

lib/path_store.py
```diff
@@ -1,218 +1,182 @@
...
-
-    def get_size(self):
-        """
-        Returns the path size.
-        """
-        return random.randint(0, 3)
+        now = time.time()
+        self.fidelity += (path_policy.property_weights['PeerLinks'] *
+                          self.peer_links)
+        self.fidelity += (10.0 * path_policy.property_weights['HopsLength'] /
```
shitz: Is there a specific reason why you multiply by 10 here? In general do these calculations make sense? Can't it be that, e.g., `guaranteed_bandwidth` get way larger than the number of hops and there has a much bigger impact on the fidelity, even though the weight might be smaller than the one for # of hops?

topology/generator.py
```diff
@@ -394,6 +401,49 @@ def write_conf_files(AD_configs):
...
+    Generate the AD path policies and store them into files.
+
+    :param AD_configs: the configurations of all SCION ADs.
+    :type AD_configs: dict
+    """
+    for isd_ad_id in AD_configs:
+        (isd_id, ad_id) = isd_ad_id.split(ISD_AD_ID_DIVISOR)
+        file_name = 'ISD:' + isd_id + '-AD:' + ad_id + '-V:' + '0'
+        path_pol_file = 'ISD' + isd_id + PATH_POL_DIR + file_name + '.json'
+        path_pol_dict = {'BestSetSize': 5,
```
shitz: Not so sure if it makes sense to define the policy files in the generator. I suggest you create a default policy file that just gets copied by the generator to the appropriate locations (and named appropriately). If people want to create their own configs, they could then adjust them after generation. This only clogs the generator code...

infrastructure/beacon_server.py
shitz: Just dumping a datastructure to the log whenever something gets registered is fine for immediate debugging, but is way too much output for anyone that doesn't care/develop on this part of the code.

infrastructure/beacon_server.py
shitz: Yes. SCIONElement should only contain what's common between all the classes that inherit from it, otherwise we should move that parts to the subclasses that need them.

lib/path_store.py
LorenzoBaesso: I set 10 because given the actually topology, the result of that operation would have been in line with the other addends. The main difficulty in this formula if to compare apples and oranges in a way that it makes sense and driven by the weights, as you pointed out. For the bandwidth, for now, we don't need to worry because weights are 0 and bandwidth values as well. The timing properties (lastseen, lastsent, delay and expiration) are casted within 0-1, so that's fine. Disjointness needs a fix (coming soon). Peer links and hops length are still arbitrary and for those we need to establish a MAX_PEER_LINKS and MAX_HOPS_LENGTH, to be within 0-1. Any values suggested?

LorenzoBaesso: Maybe, 20 for MAX_HOPS_LENGTH and 60 for MAX_PEER_LINKS? Could they be good bounds (for a single segment)?
Edit: Nevermind, I'll set these values dynamically according to the paths in the PathStore.

lib/path_store.py
```diff
@@ -220,118 +183,94 @@ class PathStore(object):
...
+        hl_count = []
+        for candidate in self.candidates:
+            candidate.hops_length = candidate.pcb.get_n_hops()
+            hl_count.append(candidate.hops_length)
+        max_hops_length = max(hl_count)
+        for candidate in self.candidates:
+            candidate.hops_length /= max_hops_length
 
-    def _check_unwanted_ads(self, path):
+    def _update_all_disjointness(self):
```
pszal, line 234: Before disjointness was defined as a filter, and now it is unclear to me whether if it really helps in selecting a disjoint paths. Can you describe a little bit why this is defined in a such way?

LorenzoBaesso, line 234: The workflow is now the following: the segment arrives, gets filtered, gets verified and then stored. Filtering paths means looking at each one of them (independently from the others) and discarding them, if some properties are not within the ranges. Properties instead could be either dependent or independent (from other paths). Disjointness is therefore a property. It is computed in a way such that the most disjoint paths (less ADs in common with ALL the other paths in the PathStore) will have higher fidelity (the higher, the better) and less disjoint paths will have lower fidelity. It also guarantees that similar paths get a similar fidelity.

lib/path_store.py
shitz: I think you should normalize ALL values that are used for calculating path fidelity to [0,1]. That way you'd prevent that parameters, that can get higher absolute values than others, would influence the fidelity disproportionately. What do you guys think?

LorenzoBaesso: I agree. This change is in the commit “Addressed Sam’s comments.”

pszal: I think it does not matter as everything can be adjusted with the
weights. However normalization by a non-constant value causes that
path's properties (and fidelity as a consequence) are relative. E.g. to
compare a property of path_record1 (from PathStore1) with the same
property of path_record2 (from PathStore2) we need compute it again. But
maybe it is not a case.

lib/path_store.py
```diff
@@ -1,218 +1,181 @@
...
-        """
-        return random.randint(0, 3)
-
-    def get_size(self):
-        """
-        Returns the path size.
-        """
-        return random.randint(0, 3)
+        now = time.time()
+        self.fidelity += (path_policy.property_weights['PeerLinks'] *
```
shitz, line 146: You don't normalize everything. You should normalize `peer_links`, `hops_length`, `disjointness`, `delay_time`, and all the `*_bandwidth`.

lib/path_store.py
```diff
@@ -1,218 +1,181 @@
...
+
+    :ivar pcb: the PCB representing the record.
+    :vartype pcb: :class:`lib.packet.pcb.PathSegment`
+    :ivar id: the path segment identifier stored in the record's PCB.
+    :vartype id: bytes
+    :ivar fidelity: the fidelity of the path record.
+    :vartype fidelity: float
+    :ivar peer_links: the number of peering links in the path segment.
+    :vartype peer_links: int
+    :ivar hops_length: the length of the path segment in hops.
```
pszal: now it is relative value

lib/path_store.py
```diff
@@ -1,218 +1,181 @@
...
+    :vartype hops_length: int
```
pszal: float (the same applies for others)

lib/path_store.py
LorenzoBaesso, line 146: _update_all_peer_links, _update_all_hops_length, _update_all_disjointness and _update_all_delay_time do that. The bandwidths for now are left to 0, as we are not sure about their progress.

lib/path_store.py
LorenzoBaesso: But in the end shall we keep everything (property values and fidelity) in int or float? What do you suggest guys?

pszal: IMHO currently it does not matter, and it can stay normalized. However if you are going to merge/compare/etc... records from different PathStores then it shouldn't be normalized (it may be helpful when we decide to replicate these records among beacon servers...)

PR #39, opened 2015-03-10
Hash-Chains and Path Management packet type

shitz:
- Implemented hash-chains for revocation tokens
- Created a new packet type: PathMgmtPacket. This type is used for requesting and delivering path segments, register a lease of a path segment with an authoritative path server and revoking a path-segment, interface, and hops. 
  The last two items are not yet implemented in this pull-request, but will be in a subsequent. I decided to split up these requests in order to keep the manageable.

### DIFF
--- a/infrastructure/beacon_server.py
+++ b/infrastructure/beacon_server.py
@@ -17,9 +17,6 @@
 """
 
 from _collections import deque
-import base64
-import copy
-import datetime
 from infrastructure.scion_elem import SCIONElement
 from lib.crypto.asymcrypto import sign
 from lib.crypto.certificate import verify_sig_chain_trc, CertificateChain, TRC
@@ -37,6 +34,9 @@ from lib.packet.scion import (SCIONPacket, get_type, PacketType as PT,
 from lib.util import (read_file, write_file, get_cert_file_path,
     get_sig_key_file_path, get_trc_file_path)
 from lib.util import init_logging
+import base64
+import copy
+import datetime
 import logging
 import os
 import sys
@@ -56,6 +56,10 @@ class BeaconServer(SCIONElement):
             propagation.
         reg_queue: A FIFO queue containing paths for registration with path
             servers.
+        if2rev_tokens: Contains the currently used revocation token
+            hash-chain for each interface.
+        seg2rev_tokens: Contains the currently used revocation token
+            hash-chain for a path-segment.
     """
     # Amount of time units a HOF is valid (time unit is EXP_TIME_UNIT).
     HOF_EXP_TIME = 63
@@ -71,10 +75,8 @@ class BeaconServer(SCIONElement):
                                              self.topology.ad_id, 0)
         self.signing_key = read_file(sig_key_file)
         self.signing_key = base64.b64decode(self.signing_key)
-        self.if2rev_tokens = {}  # Contains the currently used revocation token
-                                 # hash-chain for each interface.
-        self.seg2rev_tokens = {}  # Contains the currently used revocation
-                                  # token hash-chain for a path-segment.
+        self.if2rev_tokens = {}
+        self.seg2rev_tokens = {}
 
     def _get_if_rev_token(self, if_id):
         """
--- a/infrastructure/path_server.py
+++ b/infrastructure/path_server.py
@@ -16,8 +16,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-import copy
-import datetime
 from infrastructure.scion_elem import SCIONElement
 from lib.packet.host_addr import IPv4HostAddr
 from lib.packet.path_mgmt import (PathSegmentRecords, PathSegmentInfo,
@@ -26,6 +24,8 @@ from lib.packet.scion import PacketType as PT
 from lib.packet.scion import SCIONPacket, get_type
 from lib.path_db import PathSegmentDB
 from lib.util import update_dict, init_logging
+import copy
+import datetime
 import logging
 import sys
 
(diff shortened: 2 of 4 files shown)


infrastructure/beacon_server.py
```diff
@@ -17,31 +17,34 @@
...
 import sys
 import threading
 import time
-import os
-import copy
-import base64
-import datetime
+
+from Crypto import Random
+from Crypto.Hash import SHA256
```
LorenzoBaesso, line 48: Question about styling. I see sometimes you sort the imports alphabetically, right? But why not always? I mean, Why not having all imports after all from *...Just a suggestion..

infrastructure/beacon_server.py
```diff
@@ -68,40 +71,41 @@ def __init__(self, addr, topo_file, config_file):
                                              self.topology.ad_id, 0)
         self.signing_key = read_file(sig_key_file)
         self.signing_key = base64.b64decode(self.signing_key)
-        self.if2rev_tokens = {}  # Contains the currently used revocation tokens
-                                 # for each interface.
+        self.if2rev_tokens = {}  # Contains the currently used revocation token
+                                 # hash-chain for each interface.
```
LorenzoBaesso: Please move the comments to the documentation of the class.

infrastructure/path_server.py
```diff
@@ -413,17 +413,18 @@ def _request_paths_from_core(self, ptype, dst_isd, dst_ad,
...
-    def handle_path_request(self, path_request):
+    def handle_path_request(self, pkt):
         """
         Handles all types of path request.
         """
-        assert isinstance(path_request, PathSegmentRequest)
-        dst_isd = path_request.info.dst_isd
-        dst_ad = path_request.info.dst_ad
-        ptype = path_request.info.type
+        info = pkt.payload
```
LorenzoBaesso: Maybe info is a bit too general..

infrastructure/beacon_server.py
shitz, line 48: That is something my IDE does automatically and it conforms to the PEP-8 styleguide for Python. However, I don't have a strong opinion on that and what you suggested would also make sense.

infrastructure/beacon_server.py
shitz: Good point.

lib/packet/scion.py
```diff
@@ -44,8 +45,7 @@ class PacketType(object):
     TRC_REP = 8  # Root of Trust file reply from parent AD
     TO_LOCAL_ADDR = 100  # Threshold to distinguish local control packets
     BEACON = 101  # PathSegment type
-    PATH_REQ = 108  # Path request to TDC/lPS
-    PATH_REC = 109  # Path record (downpath reg. reply from TDC/lPS)
+    PATH_MGMT = 108  # Path management packet to CPS/lPS
```
LorenzoBaesso, line 48: Could you add please in the class description something about why some values should start from 0, while others from 100? (if that's the case)

lib/packet/scion.py
```diff
@@ -70,9 +70,9 @@ class IDSize(object):
 
 TYPES_SRC = {
         PacketType.CERT_REP: 33611786,
+        PacketType.PATH_MGMT: 67166218,
```
LorenzoBaesso, line 73: Question for my understanding; are these numbers purely random?

infrastructure/path_server.py
shitz: It's always a trade-off between brevity and expressiveness. I changed it to `segment_info`.

lib/packet/scion.py
shitz, line 48: Actually I have no idea why this is the case. You should ask Pawel about this. It seems that this was an early design decision, but from what I understand, most of these constant aren't need anymore anyway, since we use the src address of a packet to determine its type.

lib/packet/scion.py
shitz, line 73: No. These are 'multicast' IP-addresses, but written as an integer.

infrastructure/beacon_server.py
LorenzoBaesso, line 48: Oh I see. What's important then is that we decide which way to go because otherwise, I know it's a minor thing but still, at every pull request we'll see all these "changed" lines, while it's actually stuff moved around..Out of curiosity, is it something that can be turned off in your IDE? If not, we'll adapt to PEP-8..

shitz, line 48: I agree. I can definitely turn it off. I can also tell it to sort 'from X import Y' imports before 'import X' statements. So maybe we should adapt what you suggested and have the 'from X import Y' statements first and sorted alphabetically, and then have all the 'import X' statements, also sorted alphabetically. Sounds good?

LorenzoBaesso, line 48: Yes, perfect. Thanks

lib/crypto/hash_chain.py
```diff
@@ -0,0 +1,91 @@
...
+        self._init_chain()
+
+    def _init_chain(self):
+        """
+        Initializes the hash chain.
+        """
+        prev_ele = self._start_ele
+        self.entries.append(self._start_ele)
+        for _ in range(self._length - 1):
+            self.entries.append(self._hash_func.new(prev_ele).digest())
```
pszal, line 47: Does it work? prev_ele is not updated within this loop.

lib/packet/scion.py
pszal, line 48: Yes, that's legacy. Will fix it in a future cleanup pull request.

lib/crypto/hash_chain.py
shitz, line 47: no it doesn't. good catch

Issue #37, opened 2015-03-09
Use Supervisord for process management


tonyo:
I'm working on a testbed management system for SCION for my Master's thesis (including secure updates and connecting a new AD), and at the moment I'm looking for a convenient way to manage servers/processes for the current deployment model (one PC -- one or more ADs). It would be nice to have a module to keep track of all running server instances, be able to start/stop/restart them all at once, as well as having an API to do it from some other module.

The current solution with 'screen' is not flexible enough, especially when stderr and stdout are being written to log files. My suggestion is to use Supervisord (http://supervisord.org/). This lightweight tool provides a flexible way to manage processes (check status, start, restart), processes can be restarted in groups (might be useful to restart all the servers for a given AD), it also has a nice XML-RPC API and a simple web interface. 

My idea is to have Supervisord running for each machine where SCION is deployed, and to have another daemon, which connects to the centralized network management interface ("AD manager", written in Django, this is a work in progress). This daemon will be responsible for receiving commands from the AD manager via some secure protocol and performing updates and/or transmitting updated topology files if necessary, and it will proxy start/stop/restart commands to the Supervisord instance.

Would be great to get some feedback on this. I really don't see any reason to stick to 'screen' any longer, so I believe it's a good idea to employ something powerful and commonly used.

PR #41, opened 2015-03-17
Add initial support for Supervisord

tonyo:
As a follow-up to #37, here's the experimental integration with 'supervisord' as a process manager. This PR does NOT replace the original `scion.sh`, it only adds a `supervisor/`directory with the `supervisor.sh` script, which is basically a wrapper around `supervisord` and `supervisorctl`. This allows to try the new approach to the process management, having the old and the new scripts side by side. It is not recommended to mix two approaches, though, since it will lead to duplicated instances of the processes.

Advantages of 'supervisord' were listed in #37, so straight to the new workflow (there's also README.md in the `supervisor/` dir):

  0) Install dependencies (Python 2.7 is required for the current version of 'supervisor', but they are planning to release a new version soon with the Python 3.x support):

```
  $ sudo apt-get install python-pip
  $ sudo pip2 install supervisor==3.1.3 supervisor-quick
```

  1) Create 'supervisor' configuration files ('generator.py' is updated to generate 'supervisor' config files)

  ./scion.sh topology

  2) Check status

  ./supervisor/supervisor.sh status

Output example:

```
ad1-10:bs1-10-1                  STOPPED   Not started
ad1-10:cs1-10-1                  STOPPED   Not started
ad1-10:er1-10er1-19              STOPPED   Not started
ad1-10:ps1-10-1                  STOPPED   Not started
ad1-11:bs1-11-1                  STOPPED   Not started
ad1-11:cs1-11-1                  STOPPED   Not started
...
ad2-25:cs2-25-1                  STOPPED   Not started
ad2-25:er2-25er2-23              STOPPED   Not started
ad2-25:ps2-25-1                  STOPPED   Not started
ad2-26:bs2-26-1                  STOPPED   Not started
ad2-26:cs2-26-1                  STOPPED   Not started
ad2-26:er2-26er2-23              STOPPED   Not started
ad2-26:er2-26er2-24              STOPPED   Not started
ad2-26:ps2-26-1                  STOPPED   Not started
```

  4) Manage processes

  # start all processes
  ./supervisor/supervisor.sh start all

  # stop all processes
  ./supervisor/supervisor.sh stop all

  # restart all processes in AD1-11
  ./supervisor/supervisor.sh restart ad1-11:*

  # restart only the certificate server in AD2-26
  ./supervisor/supervisor.sh restart ad2-26:cs2-26-1

---

'Supervisor' also starts a local XML-RPC server on the port 9001, which allows to connect and control processes over API (more info: http://supervisord.org/api.html).

On a later stage, `supervisor.sh` can be called right from 'scion.sh', proxying control actions (such as run/stop).

Please tell me what you think of it.

### DIFF
--- a/topology/generator.py
+++ b/topology/generator.py
@@ -29,6 +29,7 @@ import json
 import logging
 import shutil
 import os
+import sys
 import struct
 import socket
 import base64
@@ -47,6 +48,7 @@ ENC_KEYS_DIR = '/encryption_keys/'
 SETUP_DIR = '/setup/'
 RUN_DIR = '/run/'
 PATH_POL_DIR = '/path_policies/'
+SUPERVISOR_DIR = '/supervisor/'
 
 CORE_AD = 'CORE'
 INTERMEDIATE_AD = 'INTERMEDIATE'
@@ -150,6 +152,7 @@ def create_directories(AD_configs):
         setup_path = 'ISD' + isd_id + SETUP_DIR
         run_path = 'ISD' + isd_id + RUN_DIR
         path_pol_path = 'ISD' + isd_id + PATH_POL_DIR
+        supervisor_path = 'ISD' + isd_id + SUPERVISOR_DIR
         if not os.path.exists(cert_path):
             os.makedirs(cert_path)
         if not os.path.exists(conf_path):
@@ -166,6 +169,8 @@ def create_directories(AD_configs):
             os.makedirs(run_path)
         if not os.path.exists(path_pol_path):
             os.makedirs(path_pol_path)
+        if not os.path.exists(supervisor_path):
+            os.makedirs(supervisor_path)
 
 
 def write_keys_certs(AD_configs):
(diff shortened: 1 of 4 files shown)


topology/generator.py
```diff
@@ -275,22 +281,37 @@ def write_topo_files(AD_configs, er_ip_addresses):
                      'CertificateServers': {},
                      'PathServers': {},
                      'EdgeRouters': {}}
-        with open(setup_file, 'a') as setup_fh, open(run_file, 'a') as run_fh:
+        group_programs = []
+        with open(setup_file, 'a') as setup_fh,\
```
shitz, line 293: small nit - use parenthesis for multiline statements instead of '\'.

tonyo, line 293: But is it really possible to wrap this line in parentheses? 
`with (open() as A, open() as B): ...` does not work.

shitz, line 293: Hmm that's strange. What if you enclose the `with` as well? Just try quickly if you get it to work somehow and otherwise it's no big deal.

tonyo, line 293: No, doesn't work. There are some tricks with `contextlib` and 'ExitStack', but they look even uglier, so let's leave it as is.

PR #42, opened 2015-03-23
Scion addr (small, but important change)

pszal:
Initial commit for scion addresses. 
It replaced IP addresses in SCIONPackets by SCION Addresses (ISD, AD, IP).
With this change all request packets that contains requester ISD,AD in a payload, are redundant.

In the following commits I'll:
- revise SCION typing 
- resize (ISD, AD) pair to 4B. I think to set 1B for ISD and 3 for AD
- replace SCIONElem.topology's addresses by SCIONAddr  (not sure this is a good idea, as addresses from self.topology are always local)
- also noticed that python3 comes with nice https://docs.python.org/3/library/ipaddress.html , that can replace our IPv{4,6}HostAddr classes

Let me know what do you think about these changes.

### DIFF
--- a/lib/packet/host_addr.py
+++ b/lib/packet/host_addr.py
@@ -143,6 +143,8 @@ class SCIONAddr(object):
     """
     Class for complete SCION addresses.
     """
+    ISD_AD_LEN = 10  # Size of (isd_id, ad_id) pair in bytes.
+
     def __init__(self, raw=None):
         self.isd_id = None
         self.ad_id = None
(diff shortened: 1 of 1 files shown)


lib/defines.py
```diff
@@ -25,4 +25,6 @@
 """
 
 MAX_SEGMENT_TTL = 24 * 60 * 60  # Max TTL of a PathSegment in realtime seconds.
-EXP_TIME_UNIT = MAX_SEGMENT_TTL / 2 ** 8  # Time unit for HOF expiration.
\ No newline at end of file
+EXP_TIME_UNIT = MAX_SEGMENT_TTL / 2 ** 8  # Time unit for HOF expiration.
+ISD_LEN = 2  # Size of ISD ID in bytes.
```
shitz: I would suggest making those constants class variables of SCIONAddr. defines.py will quickly become cluttered if we start putting everything in there.

lib/packet/scion.py
```diff
@@ -99,9 +99,10 @@ def get_addr_from_type(ptype):
     :rtype: :class:`IPv4HostAddr`
     """
     if ptype in TYPES_SRC:
-        return IPv4HostAddr(TYPES_SRC[ptype])
+        addr = IPv4HostAddr(TYPES_SRC[ptype])
     else:
-        return IPv4HostAddr(TYPES_DST[ptype])
+        addr = IPv4HostAddr(TYPES_DST[ptype])
+    return SCIONAddr.from_values(0, 0, addr)
```
shitz, line 105: Shouldn't this still return a local address instead of (0, 0, local_address). You have to fill in the isd and ad part anyway, so way not let this return a local address. It makes it much less 'hacky'.

endhost/gateway.py
```diff
@@ -123,7 +124,7 @@ def init_routing(self):
         device.
         """
         for i in self.scion_hosts.keys():
```
LorenzoBaesso: What about ip instead of i?

lib/packet/host_addr.py
```diff
@@ -23,8 +23,12 @@
     Fill in the docstring.
 """
 
+from lib.defines import ISD_LEN, AD_LEN
+from bitstring import BitArray
+import bitstring
 import socket
 import struct
+import logging
```
LorenzoBaesso: Minor thing: as a convention, we opted for ordering the imports alphabetically. That is:
from _collections ...
from lib. ...
from topology. ...
from Crypto import ..
from Crypto.Hash import ...
import bitstring
import logging

lib/packet/host_addr.py
```diff
@@ -134,3 +138,54 @@ def __init__(self, addr=None):
...
         if addr is not None:
             self.addr = addr
+
+
+class SCIONAddr(object):
+    """
+    Class for complete SCION addresses.
+    """
+    def __init__(self, raw=None):
+        self.isd = None
```
LorenzoBaesso: int

lib/packet/host_addr.py
```diff
@@ -134,3 +138,54 @@ def __init__(self, addr=None):
...
             self.addr = addr
+
+
+class SCIONAddr(object):
+    """
+    Class for complete SCION addresses.
+    """
+    def __init__(self, raw=None):
+        self.isd = None 
+        self.ad = None
```
LorenzoBaesso: int

lib/defines.py
pszal: I decided to remove and hardcode that, as this is done in the rest of the code.

lib/packet/scion.py
pszal, line 105: Yes, that's temporary hack. In a following pull request I'll revise types and replace that function by set_type() or something similar.

lib/packet/host_addr.py
```diff
@@ -134,3 +137,52 @@ def __init__(self, addr=None):
...
+        if raw:
+            self.parse(raw)
+
+    @classmethod
+    def from_values(cls, isd_id, ad_id, host_addr):
+        addr = SCIONAddr()
+        addr.isd_id = isd_id
+        addr.ad_id = ad_id
+        addr.host_addr = host_addr
+        addr.addr_len = 10 + addr.host_addr.addr_len
```
kormat: While i'm obviously new and don't know the context, seeing the repeated magical value '10' is a bit suspicious. If this is a constant, maybe define it as such, so that it's clear?

PR #43, opened 2015-03-23
Fixed naming for certificate chains.

LorenzoBaesso:
Minor fix about the names used for the certificate chain objects.

New: now CSs and BSs store (also) in memory the certificate chains and TRCs they receive. If a certificate chain or TRC is not found in memory, then the file is searched and, if found, the information gets stored in memory. It can be useful to speed up the cert chains/TRCs look-ups, but even more if a server needs to be rebooted and it has local copies of the required files (this will prevent useless requests).

### DIFF
--- a/test/certificates_test.py
+++ b/test/certificates_test.py
@@ -17,7 +17,7 @@
 """
 
 from infrastructure.cert_server import CertServer
-from infrastructure.scion_elem import SCION_UDP_PORT
+from infrastructure.scion_elem import SCION_UDP_PORT, BUFLEN
 from lib.crypto.asymcrypto import sign
 from lib.crypto.certificate import verify_sig_chain_trc, CertificateChain, TRC
 from lib.packet.host_addr import IPv4HostAddr
@@ -25,10 +25,12 @@ from lib.packet.scion import (SCIONPacket, get_type, PacketType as PT,
     CertChainRequest, CertChainReply, TRCRequest, TRCReply)
 from lib.topology import Topology
 from lib.util import (get_cert_chain_file_path, get_trc_file_path, read_file,
-    get_sig_key_file_path)
+    get_sig_key_file_path, write_file)
 import base64
-import socket
 import logging
+import os
+import select
+import socket
 import unittest
 
 
@@ -60,7 +62,7 @@ class TestCertificates(unittest.TestCase):
         sig = sign(msg, sig_priv13)
         chain = CertificateChain.from_values([])
         print('Sig test 2:', verify_sig_chain_trc(msg, sig, 'ISD:1-AD:13',
-            cert10, trc, 0))
+            cert10, trc, 0), '\n')
 
         topology = Topology("../topology/ISD1/topologies/ISD:1-AD:10.json")
         src_addr = IPv4HostAddr("127.0.0.1")
@@ -74,9 +76,21 @@ class TestCertificates(unittest.TestCase):
             topology.ad_id, 1, 0).pack()
         sock.sendto(msg, (str(dst_addr), SCION_UDP_PORT))
 
-        data, addr = sock.recvfrom(1024)
+        temp_file = './temp.txt'
+        timeout = 5
+
+        ready = select.select([sock], [], [], timeout)
+        if not ready[0]:
+            print("Error: no TRC reply was received!")
+            sock.close()
+            return
+
+        data, _ = sock.recvfrom(BUFLEN)
         print("Received TRC reply from local CS.")
         trc_reply = TRCReply(data)
+        write_file(temp_file, trc_reply.trc.decode('utf-8'))
+        trc = TRC(temp_file)
+        assert trc.verify()
 
         print("Sending cert chain request (ISD:1-AD:16-V:0) to local CS.")
         msg = CertChainRequest.from_values(PT.CERT_CHAIN_REQ_LOCAL, src_addr,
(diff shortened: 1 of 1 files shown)


infrastructure/beacon_server.py
```diff
@@ -400,51 +405,59 @@ def _try_to_verify_beacon(self, pcb):
...
-                    new_cert_req = CertRequest.from_values(PT.CERT_REQ_LOCAL,
-                        self.addr, if_id, self.topology.isd_id,
-                        self.topology.ad_id, isd_id, cert_ad, cert_version)
-                    dst_addr = self.topology.certificate_servers[0].addr
-                    self.send(new_cert_req, dst_addr)
-                    self.requested_certs[cert_tuple] = now
-                    return False
-        else:
+        trc = self.trcs.get((isd_id, trc_version))
+        if not trc:
```
shitz, line 416: Can you add some comments throughout the code. For example here you could write `# Try loading file from disk` and below `# Requesting TRC file from cert server`. These simple comments make it much easier to parse the code.

infrastructure/cert_server.py
```diff
@@ -100,27 +114,29 @@ def process_trc_request(self, trc_req):
...
-        src_addr = trc_req.hdr.src_addr
-        ptype = get_type(trc_req)
-        trc_file = get_trc_file_path(self.topology.isd_id, self.topology.ad_id,
-            trc_req.trc_isd, trc_req.trc_version)
-        if not os.path.exists(trc_file):
-            logging.info('TRC file not found.')
-            self.trc_requests.setdefault((trc_req.trc_isd, trc_req.trc_version),
-                []).append(src_addr)
+        trc = self.trcs.get((trc_req.isd_id, trc_req.version), '')
+        if trc == '':
```
shitz: `if not trc` 
It's considered bad coding style to compare to empty strings or 0 etc.

infrastructure/cert_server.py
```diff
@@ -36,63 +36,77 @@ class CertServer(SCIONElement):
...
+            cert_chain_file = get_cert_chain_file_path(self.topology.isd_id,
+                self.topology.ad_id, cert_chain_req.isd_id,
+                cert_chain_req.ad_id, cert_chain_req.version)
+            if os.path.exists(cert_chain_file):
+                cert_chain = read_file(cert_chain_file).encode('utf-8')
+                self.cert_chains[(cert_chain_req.isd_id, cert_chain_req.ad_id,
+                                  cert_chain_req.version)] = cert_chain
+        if cert_chain == '':
+            logging.info('Certificate chain not found.')
+            self.cert_chain_requests.setdefault((cert_chain_req.isd_id,
```
shitz: Instead of `setdefault` you could also define `self.cert_chain_requests = defaultdict([])`. The default value for a non existing key is then automatically set to an empty list. Do whatever you prefer.

infrastructure/cert_server.py
```diff
@@ -100,27 +114,29 @@ def process_trc_request(self, trc_req):
...
-                []).append(src_addr)
+        trc = self.trcs.get((trc_req.isd_id, trc_req.version), '')
+        if trc == '':
+            trc_file = get_trc_file_path(self.topology.isd_id,
+                self.topology.ad_id, trc_req.isd_id, trc_req.version)
+            if os.path.exists(trc_file):
+                trc = read_file(trc_file).encode('utf-8')
+                self.trcs[(trc_req.isd_id, trc_req.version)] = trc
+        if trc == '':
+            logging.info('TRC not found.')
```
shitz: These look more like debug outputs to me.

doc/addressing.txt
```diff
@@ -26,8 +26,8 @@ IFID_REQ ***    UNICAST                 10.224.0.10(UNICAST)            ER->ER
 IFID_REP ***    10.224.0.10 (UNICAST)   UNICAST                         ER->ER
 ROT_REQ         UNICAST                 10.224.0.8 (ANYCAST)            CS->CCS
 ROT_REP         10.224.0.8 (ANYCAST)    UNICAST                         CCS->CS
-CERT_REQ_LOCAL  UNICAST                 10.224.0.9 (ANYCAST)            BS->CS
-CERT_REP_LOCAL  10.224.0.9 (ANYCAST)    UNICAST                         CS->BS
+CERT_CHAIN_REQ_LOCAL  UNICAST                 10.224.0.9 (ANYCAST)            BS->CS
+CERT_CHAIN_REP_LOCAL  10.224.0.9 (ANYCAST)    UNICAST                         CS->BS
 DATA            UNICAST                 UNICAST                         ER->EH 
 DATA            UNICAST                 UNICAST                         EH->ER
```
pszal, line 33: General remark: probably with SCION addresses we won't need *_LOCAL requests/replies anymore.

infrastructure/beacon_server.py
```diff
@@ -348,8 +347,14 @@ def __init__(self, addr, topo_file, config_file, path_policy_file):
...
         self.down_segments = PathStore(self.path_policy)
-        self.requested_certs = {}
-        self.requested_trcs = {}
+        self.cert_chain_requests = {}
+        self.trc_requests = {}
+        self.cert_chains = {}
+        self.trcs = {}
+        cert_chain_file = get_cert_chain_file_path(self.topology.isd_id,
+            self.topology.ad_id, self.topology.isd_id, self.topology.ad_id,
+            self.config.cert_chain_version)
```
pszal, line 355: Why certificate version is in config file (shouldn't it be a path to certificate)?

lib/config.py
```diff
@@ -92,3 +98,5 @@ def parse(self, config_file):
         self.registers_paths = config['RegisterPath']
         self.pcb_queue_size = config['PCBQueueSize']
         self.path_server_queue_size = config['PSQueueSize']
+        self.sig_key_version = config['SigKeyVersion']
```
pszal: what's that? Cannot it be identified through cert version?

LorenzoBaesso: Depends; if a certificate expires, can it be signed with the old key pair or a new key pair must be generated as well?

infrastructure/beacon_server.py
LorenzoBaesso, line 355: ok, I'll add it as another argument of the beacon server

pszal, line 355: No, I think it is even worse. Leave it as it is now.

lib/config.py
pszal: It does not matter. Why this is a case? My question is when we may need this `sig_key_version` ?

infrastructure/beacon_server.py
LorenzoBaesso, line 355: ok, although note that the same (avoiding files as arguments) could be done also for topology/configuration/TRC files as the directories are hardcoded.

lib/config.py
LorenzoBaesso: When we want to keep track of the key pairs used.

pszal: Where do we plan to have such feature? That sounds like part of certificate management policy, which is current undefined, isn't it?

LorenzoBaesso: I see, alright. I will remove the version from the key files.

PR #44, opened 2015-03-27
Revocation v2

shitz:
Segment leases implemented and tested.

Segment, interface and hop revocation implemented. Segment and interface revocations are tested, hop revocations not, however, they are almost the same as interface revocations, so I'm positive that they should mostly work as well.

I didn't implement appending revocations to beacons yet, since this pull request is already large enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Necessary steps in order to run SCION:
 
 0. Install required packages with dependencies:
 
-    sudo apt-get install python3 python3-pip screen
+    sudo apt-get install python3 python-dev python3-dev python3-pip screen
     sudo pip3 install bitstring python-pytun pydblite
 
 1. Compile the crypto library:
--- a/topology/generator.py
+++ b/topology/generator.py
@@ -296,6 +296,8 @@ def write_topo_files(AD_configs, er_ip_addresses):
         with open(setup_file, 'a') as setup_fh,\
              open(run_file, 'a') as run_fh,\
              open(supervisor_file, 'a') as supervisor_fh:
+            # Create per-AD screen session.
+            run_fh.write('screen -d -m -S %s\n' % isd_ad_id)
             # Write Beacon Servers
             ip_address = '.'.join([first_byte, isd_id, ad_id, BS_RANGE])
             supervisor_common = ['autostart=false\n', 'redirect_stderr=True\n',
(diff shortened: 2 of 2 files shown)


infrastructure/beacon_server.py
```diff
@@ -623,6 +689,57 @@ def handle_unverified_beacons(self):
...
+        elif ptype == PT.BEACON:
+            self.process_pcb(PathConstructionBeacon(packet))
+        elif ptype == PT.CERT_CHAIN_REP:
+            self.process_cert_chain_rep(CertChainReply(packet))
+        elif ptype == PT.TRC_REP:
+            self.process_trc_rep(TRCReply(packet))
+        else:
+            logging.warning("Type not supported")
+
+#     def _test_revocation(self):
```
LorenzoBaesso: Is this still needed? Could it be moved to a unit test?

infrastructure/beacon_server.py
```diff
@@ -615,6 +610,77 @@ def process_trc_rep(self, trc_rep):
...
+        """
+        Sends out revocation to the local PS and a CPS and down_stream BS.
+        """
+        assert isinstance(rev_info, RevocationInfo)
+        # Build segment revocations for local path server.
+        rev_infos = []
+        to_remove = []
+        if rev_info.rev_type == RT.DOWN_SEGMENT:
+            if not self.down_segments.get_segment(rev_info.seg_id):
+                logging.error("Segment to revoke does not exist.")
```
pszal: Is not a warning() more appropriate ?

infrastructure/path_server.py
```diff
@@ -138,19 +167,126 @@ class CorePathServer(PathServer):
...
+                self._nentries += len(entries)
+
     def __init__(self, addr, topo_file, config_file):
         PathServer.__init__(self, addr, topo_file, config_file)
         # Sanity check that we should indeed be a core path server.
         assert self.topology.is_core_ad, "This shouldn't be a core PS!"
 
-    def _handle_up_segment_record(self, records):
-        PathServer._handle_up_segment_record(self, records)
+        self.leases = self.LeasesDict()
```
pszal, line 261: Why not implement this with ExpiringDict (do we need both LeasesDict and ExpiringDict) ?

lib/packet/packet_base.py
```diff
@@ -129,6 +129,15 @@ def __str__(self):
     def __repr__(self):
         return self.__str__()
 
+    def __hash__(self):
```
pszal, line 132: Does/Should it work with packets created via .from_values() ?

infrastructure/beacon_server.py
shitz: Unfortunately,  it can't be moved to a unit test.  For now we still need to trigger revocation manually. I left this in there so you could guys could test it if you want, but I 'll remove it before i merge

infrastructure/beacon_server.py
shitz: Agreed.

infrastructure/path_server.py
shitz, line 261: Expiring dict works differently. You can only set an expiration time on creation and it has to be the same for all entries. Besides it also does some other things that expiring dict doesn't provide and that are quite specific to leases.

lib/packet/packet_base.py
shitz, line 132: No,  since raw is None if you create it with from_values, but in that case we could call self.pack(), so it probably should be return hash(self.pack())

PR #48, opened 2015-04-02
Add basic zookeeper support to the beacon server

kormat:
How this works:
- Incoming PCBs are written to a /pcbs directory in ZK.
- A ZK counter is incremented for each new PCB, which triggers all BSes to process the queue.
- A lock is used to make sure only one BS processes the queue at a time.
- Any unverified PCBs are left in the queue. 
- A new TRC or cert chain reply also causes a BS to reprocess the queue.
- The first BS to successfully verify a PCB removes it from the queue, and propagates it downstream.

### DIFF
--- a/infrastructure/beacon_server.py
+++ b/infrastructure/beacon_server.py
@@ -49,7 +49,6 @@ from Crypto.Hash import SHA256
 import copy
 import logging
 
-
 class BeaconServer(SCIONElement):
     """
     The SCION PathConstructionBeacon Server.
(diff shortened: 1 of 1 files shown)


infrastructure/beacon_server.py
```diff
@@ -400,11 +467,13 @@ def _try_to_verify_beacon(self, pcb):
             if self._verify_beacon(pcb):
                 self.beacons.append(pcb)
                 logging.info("Registered valid beacon.")
+                return "verified"
```
shitz: I think we should really have (integer) constants for `verified`, `invalid`, and `unverified` instead of using strings.

infrastructure/beacon_server.py
```diff
@@ -384,6 +420,37 @@ def _verify_beacon(self, pcb):
...
+        """
+        while True:
+            # Block until there are pcbs to be processed
+            self._pcb_semaphore.acquire()
+            logging.debug("zk_process_pcbs() semaphore acquired")
+            # Acquire lock
+            with self._zk_pcb_lock:
+                logging.debug("zk_process_pcbs() got lock")
+                entries = self._zk.get_children(self._zk_pcbs_path)
+                logging.debug("zk_process_pcbs() processing %d entries" % len(entries))
```
shitz: Make sure code lines do not exceed 80 chars.

infrastructure/beacon_server.py
kormat: Done.

infrastructure/beacon_server.py
kormat: Fixed.

PR #49, opened 2015-04-05
Added core segments verification

LorenzoBaesso:


infrastructure/cert_server.py
```diff
@@ -145,10 +145,16 @@ def process_trc_request(self, trc_req):
             if get_type(trc_req) == PT.TRC_REQ_LOCAL:
                 dst_addr = trc_req.hdr.src_addr.host_addr
             else:
+                # Case in which the requester is a child AD
```
shitz: You can combine these cases with `for router in child_edge_routers + routing_edge_routers`.

infrastructure/beacon_server.py
```diff
@@ -223,6 +344,38 @@ def __init__(self, addr, topo_file, config_file, path_policy_file):
...
+                LocalBeaconServer.REQUESTS_TIMEOUT)):
+                new_trc_req = TRCRequest.from_values(PT.TRC_REQ_LOCAL,
+                    self.addr, if_id, self.topology.isd_id, self.topology.ad_id,
+                    isd_id, trc_version)
+                dst_addr = self.topology.certificate_servers[0].addr
+                self.send(new_trc_req, dst_addr)
+                self.trc_requests[trc_tuple] = now
+                return False
+        else:
+            return True
```
kormat: There seems to be a large duplication of code between CoreBeaconServer._check_certs_trc() and LocalBeaconServer._check_certs_trc(). The common code should probably stay in BeaconServer._check_certs_trc(), and have both child classes call that.

PR #52, opened 2015-04-08
Small fixes

pszal:
- removed old, unused, code
- fixed scion common header (type field replaced by version field)
- unified and simplified packet types
- introduced correct SCIONAddresses for all packets
- replaced *HostAddr classes by Python's IPv4/6Address
- noticed inconsistencies and strange behaviour of cert/trc logic, will revise that soon

### DIFF
--- a/infrastructure/path_server.py
+++ b/infrastructure/path_server.py
@@ -457,7 +457,7 @@ class CorePathServer(PathServer):
                     self.send(request, next_hop)
                     logging.info("Down-Segment request for different ISD. "
                                  "Forwarding request to CPS in (%d, %d).",
-                                 *dst_isd_ad)
+                                 dst_isd_ad)
                 # If no core_path was available, add request to waiting targets.
                 else:
                     self.waiting_targets.add((dst_isd, dst_ad, segment_info))


lib/packet/host_addr.py
```diff
@@ -25,120 +25,12 @@
...
-    Class for SCION host addresses.
-    """
-    def __init__(self, addr=None):
-        HostAddr.__init__(self)
-        self.addr_len = AddressLengths.HOST_ADDR_SCION
-        if addr is not None:
-            self.addr = addr
-
-
 class SCIONAddr(object):
```
shitz: I think we should rename this file to scion_addr.py now that HostAddr is gone.

lib/packet/path_mgmt.py
```diff
@@ -460,18 +460,21 @@ def from_values(cls, type, payload, path, src_addr=None, dst_addr=None):
         :type lib.packet.packet_base.PayloadBase
         :param path: the path of the packet
         :type lib.packet.path.PathBase
-        :param src_addr: source address
-        :type lib.packet.host_addr.HostAddr
-        :param dst_addr: destination address
-        :type lib.packet.host_addr.HostAddr
+        :param src_addr: source address (isd_id, ad_id tuple for response)
```
shitz: Maybe this would actually be a nice use case for a named tuple, since we probably need the (isd, ad) pair in many other places as well.

infrastructure/path_server.py
```diff
@@ -310,7 +310,8 @@ def _handle_down_segment_record(self, pkt):
         if paths_to_propagate:
             records = PathSegmentRecords.from_values(records.info,
                                                      paths_to_propagate)
-            pkt = PathMgmtPacket.from_values(PMT.RECORDS, records, None)
+            pkt = PathMgmtPacket.from_values(PMT.RECORDS, records, None,
+                                             self.addr, (0, 0))
```
LorenzoBaesso: Can these made default values in the PathMgmtPacket constructor?

infrastructure/path_server.py
```diff
@@ -440,17 +444,18 @@ def handle_path_request(self, pkt):
...
-                                                         path, self.addr)
+                                                         segment_info, path,
+                                                         self.addr, dst_isd_ad)
                     request.hdr.set_downpath()
                     self.send(request, next_hop)
                     logging.info("Down-Segment request for different ISD. "
                                  "Forwarding request to CPS in (%d, %d).",
-                                 cpaths[0].get_last_pcbm().spcbf.isd_id,
-                                 cpaths[0].get_last_pcbm().ad_id)
+                                 *dst_isd_ad)
```
LorenzoBaesso: Out of curiosity, how does this work? why not dst_isd_ad only?

lib/packet/scion.py
```diff
@@ -17,92 +17,36 @@
...
+    BEACON = IPv4Address("10.224.0.1")  # Path Construction Beacon
+    PATH_MGMT = IPv4Address("10.224.0.2")  # Path management packet from/to PS
+    TRC_REQ = IPv4Address("10.224.0.3")  # TRC file request to parent AD
+    TRC_REQ_LOCAL = IPv4Address("10.224.0.4")  # TRC file request to lCS
+    TRC_REP = IPv4Address("10.224.0.5")  # TRC file reply from parent AD
+    CERT_CHAIN_REQ = IPv4Address("10.224.0.6")  # cert chain request to parent AD
+    CERT_CHAIN_REQ_LOCAL = IPv4Address("10.224.0.7")  # local cert chain request
+    CERT_CHAIN_REP = IPv4Address("10.224.0.8")  # cert chain reply from lCS 
+    IFID_REQ = IPv4Address("10.224.0.9")  # IF ID request to the peer router
+    IFID_REP = IPv4Address("10.224.0.10")  # IF ID reply from the peer router
```
LorenzoBaesso, line 46: Is this completely independant from the topology settings?

infrastructure/path_server.py
pszal: I don't think so. This is a special case, and in the fact pkt with (dst_isd=0, dst_ad=0) is not sent, as method _propagate_to_core_ads() replaces (dst_id, dst_ad) by correct values.

infrastructure/path_server.py
pszal: Yes, without \* is less confusing. But you can look at https://docs.python.org/2/tutorial/controlflow.html#unpacking-argument-lists

lib/packet/scion.py
pszal, line 46: Yes, these are constants that determine type of a packet.

Issue #50, opened 2015-04-07
Beacon server high availability


kormat:
In order to remove Beacon Servers as [SPOFs](http://en.wikipedia.org/wiki/Single_point_of_failure), we're moving to a distributed architecture. The current proposal is based on [Zookeeper](https://zookeeper.apache.org/), with the [Kazoo](https://kazoo.readthedocs.org/en/latest/) python client library.

All Beacon Servers (BS) in an AD connect to a Zookeeper (ZK) cluster. They register as part of a [Party](https://kazoo.readthedocs.org/en/latest/api/recipe/party.html); this is used to create multiple DNS entries for the Beacon service. The BSes try to grab the master lock, whichever BS succeeds is the master until failure/shutdown.

When an Edge Router (ER) receives a new PCB, it connects to any BS in the DNS record. If that connection times out, the ER tries the next record, and so on. The receiving BS drops the new PCB into the shared ZK "incoming" directory. All BSes have watches set on this directory, and load any new PCBs into their PathStores.

Once per propagation period, the master BS will select k PCBs from Pathstore (using freshness criteria to make sure nothing stale is selected) and propagate them. It will also move all PCBS from the ZK "incoming" directory to the "recent" directory, clearing the "recent" dir beforehand. Any newly-started BS will load all PCBs from both "incoming" and "recent" directories to prime their PathStores.

PR #61, opened 2015-04-16
Add clustering of the beacon servers within an AD

kormat:
Uses Zookeeper, via the Kazoo library.

Incoming PCBs are written directly to a ZK shared path. All BSes in the
AD watch that path, and read new PCBS. All BSes compete for a master
lock. The successful BS master is responsible for propagating PCBS
downstream, and for registering up/down paths.

Any freshly started BS (or one that has reconnected after a connection
drop) will read all current PCBS, as well as those from the previous
propagation period before becoming eligible for the master lock.

All connected BSes will be in the ZK party for that AD, allowing
advertising of the working BSes within AD.

This is for issue #50 (Beacon server high availability).

### DIFF
--- a/infrastructure/beacon_server.py
+++ b/infrastructure/beacon_server.py
@@ -35,8 +35,10 @@ from lib.path_store import PathPolicy, PathStoreRecord, PathStore
 from lib.util import (read_file, write_file, get_cert_chain_file_path,
     get_sig_key_file_path, get_trc_file_path, init_logging, log_exception,
     thread_safety_net)
+from lib.zookeeper import (Zookeeper, ZkConnectionLoss, ZkNoNodeError)
 from Crypto import Random
 from Crypto.Hash import SHA256
+from external.stacktracer import trace_start
 import base64
 import copy
 import datetime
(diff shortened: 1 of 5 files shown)


infrastructure/beacon_server.py
```diff
@@ -133,9 +155,21 @@ def handle_pcbs_propagation(self):
         """
         pass
 
-    def process_pcb(self, beacon):
+    def store_pcb(self, beacon):
```
pszal, line 159: Maybe it makes sense to pre-validate (through _check_filters()) a beacon before putting it to ZK.

infrastructure/beacon_server.py
```diff
@@ -133,9 +155,21 @@ def handle_pcbs_propagation(self):
...
+        """
+        Receives beacon and stores it for processing.
+        """
+        assert isinstance(beacon, PathConstructionBeacon)
+        try:
```
pszal, line 166: General remark:
- we use 4 spaces for indentation
- and tw=80

infrastructure/beacon_server.py
kormat, line 159: I can do that, sure. The reason i didn't was that the existing code only checks filters in LocalBeaconServer, but not in CoreBeaconServer, for reasons that are unclear to me

pszal, line 159: Right, let's leave it for now. I'll talkt to Lorenzo on that.

infrastructure/beacon_server.py
kormat, line 166: Oops. I had some inconsistent options set in vim. Fixed.

infrastructure/beacon_server.py
kormat, line 159: Done.

kormat, line 159: Heh, ok. I can remove it again, too. :)

PR #67, opened 2015-04-21
Move scion dependancies from docs to scion.sh

kormat:
This will allow automation of testing.

@pszalach

### DIFF
--- a/scion.sh
+++ b/scion.sh
@@ -1,7 +1,40 @@
 #!/usr/bin/env bash
 
-
 # BEGIN subcommand functions
+
+PKG_DEPS="python python3 python-dev python3-dev python3-pip screen zookeeperd"
+PIP3_DEPS="bitstring python-pytun pydblite pygments pycrypto kazoo"
+
+cmd_deps() {
+    if [ ! -e /etc/debian_version ]; then
+        deps_debian
+    else
+        echo "As this is not a debian-based OS, please install the equivalents of these packages:"
+        echo "    $PKG_DEPS"
+    fi
+    echo "Installing necessary packages from pip3"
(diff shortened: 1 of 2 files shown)


scion.sh
```diff
@@ -1,7 +1,40 @@
 #!/usr/bin/env bash
 
-
 # BEGIN subcommand functions
+
+PKG_DEPS="python python3 python-dev python3-dev python3-pip screen zookeeperd"
```
shitz, line 5: I think you also need to add 'python-pip' to this list (at least I had to install it on my freshly set up Kubuntu).

scion.sh
```diff
@@ -1,7 +1,40 @@
...
+
+cmd_deps() {
+    if [ ! -e /etc/debian_version ]; then
+        deps_debian
+    else
+        echo "As this is not a debian-based OS, please install the equivalents of these packages:"
+        echo "    $PKG_DEPS"
+    fi
+    echo "Installing necessary packages from pip3"
+    pip3 install -q --user $PIP3_DEPS
```
syclops, line 16: It's possible to use pip along with a requirements file rather than incorporating this in scion.sh itself. Use `pip install -r <requirements.txt>` to accomplish this. See [here](https://devcenter.heroku.com/articles/python-pip) for more details.

scion.sh
kormat, line 5: Ah hah. This is exactly what i'm doing this to catch. Added to next PR :)

scion.sh
kormat, line 16: Ah, good to know. I think this way is fine for now, but if we needed a lot of exact versions, -r would indeed be nicer.

PR #80, opened 2015-04-24
Major zookeeper support update; Switch from events to polling.

kormat:
Using ZK watches meant events were being constantly generated, this
change switches over to polling the incoming ZK shared path every half
second.

Also:
- Add a lot more error handling, especially for corner-cases.
- Make the log output a lot less chatty
- Added time checks around some of the heavy-weight ZK operations
- Show a warning if we're failing to keep our propagation/registration
  intervals

I'm still not satisfied with the performance of this. I'm seeing reading of shared PCBs sometimes taking up to 10s, and propagation failing to keep inside its interval. I have an idea to fix this, but requires some more work.

(For #50)

@pszalach @shitz

### DIFF
--- a/lib/zookeeper.py
+++ b/lib/zookeeper.py
@@ -230,7 +230,7 @@ class Zookeeper(object):
             self._party = self._zk.Party(party_path, self._srv_id)
         try:
             self._party.join()
-        except ConnectionLoss:
+        except (ConnectionLoss, SessionExpiredError):
             raise ZkConnectionLoss
         logging.debug("Joined party, members are: %s", list(self._party))
 
@@ -250,7 +250,7 @@ class Zookeeper(object):
         try:
             self._zk.exists(path)
             self._zk.ChildrenWatch(path, func=func, allow_session_lost=False)
-        except ConnectionLoss:
+        except (ConnectionLoss, SessionExpiredError):
             raise ZkConnectionLoss
 
     def get_lock(self, timeout=60.0):
@@ -372,7 +372,7 @@ class Zookeeper(object):
         path = os.path.join(self._prefix, path)
         try:
             entries = self._zk.get_children(path)
-        except ConnectionLoss:
+        except (ConnectionLoss, SessionExpiredError):
             raise ZkConnectionLoss
         return entries
 


infrastructure/beacon_server.py
```diff
@@ -434,76 +413,102 @@ def read_shared_pcbs(self):
...
                 continue
+            self._state_synced.set()
 
+    def _read_shared_entries(self, path, name):
+        """
+        Read all entries from the specified path and return them
+        """
+        desc = "Fetching list of %s PCBs from shared path" % name
+        entries = self.zk.get_shared_entries(
+                path,
```
pszal, line 441: general remark:
4 space indentation or align to a parenthesis

infrastructure/beacon_server.py
```diff
@@ -434,76 +413,102 @@ def read_shared_pcbs(self):
...
                     raw = self.zk.get_shared_item(path, entry)
                 except ZkConnectionLoss:
                     logging.warning("Unable to retrieve PCB from shared path: "
                                     "no connection to ZK")
-                    return
+                    break
                 except ZkNoNodeError:
-                    logging.info("Unable to retrieve PCB from shared path: "
+                    logging.debug("Unable to retrieve PCB from shared path: "
                                  "no such entry (%s/%s)" % (path, entry))
```
pszal, line 485: one space more

infrastructure/beacon_server.py
```diff
@@ -587,11 +600,17 @@ def register_segments(self):
                          "register_segments")
             return
         while True:
-            logging.debug("register_segments() waiting for lock")
+            lock = self.zk.have_lock()
+            if not lock:
+                logging.debug("register_segements: waiting for lock")
```
pszal, line 605: typo: segments

infrastructure/beacon_server.py
```diff
@@ -885,26 +912,31 @@ def handle_pcbs_propagation(self):
         Main loop to propagate received beacons.
         """
         # TODO: define function that dispaches the pcbs among the interfaces
```
pszal, line 914: Older comment but typo: dispatches,
btw. I'm not sure what this comment is about ? Basing on PCB determine which ER sent that? I thought we have a function for that.

infrastructure/beacon_server.py
kormat, line 441: Damnit, yes. More generally i'm strongly in favour of us using the [Pep8](https://www.python.org/dev/peps/pep-0008/#indentation) style guide.

Anyway, fixed, and in a few other places too.

infrastructure/beacon_server.py
kormat, line 485: Fixed

infrastructure/beacon_server.py
kormat, line 605: Fixed

PR #92, opened 2015-04-29
Switch ZK to using a shared cache

kormat:
Store all PCBs under their path segment hash in ZK. This is much more efficient
than storing all PCBs seperately, and then doing mass operations on them, as
well as greatly simplifying the logic.

~~Not yet done: pruning old PCBs.~~
Pruning old PCBs added.

For #50 

@pszalach

### DIFF
--- a/lib/zookeeper.py
+++ b/lib/zookeeper.py
@@ -19,6 +19,7 @@
 import os.path
 import logging
 import threading
+import time
 
 from kazoo.client import (KazooClient, KazooState, KazooRetry)
 from kazoo.handlers.threading import TimeoutError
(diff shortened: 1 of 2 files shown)


lib/zookeeper.py
```diff
@@ -236,25 +237,6 @@ def join_party(self):
...
-        if not self.is_connected():
-            raise ZkConnectionLoss
-        path = os.path.join(self._prefix, path)
-        try:
-            self._zk.exists(path)
-            self._zk.ChildrenWatch(path, func=func, allow_session_lost=False)
-        except (ConnectionLoss, SessionExpiredError):
-            raise ZkConnectionLoss
-
     def get_lock(self, timeout=60.0):
```
pszal, line 240: timeout is unused

PR #94, opened 2015-04-29
Refactoring of generator.py

tonyo:
This PR contains major refactorings of `generator.py`. The main idea is to decompose it and make as reusable as possible, because it would be convenient to use the same code to generate configurations on-demand while connecting a new AD or updating existing configurations. One of the possible approaches (which is implemented in this PR) is to establish a pipeline for config files, which takes as input some limited set of files (topology and configuration files), and then generate 'supervisor' configs, setup files, and everything else based on this set. Configuration files (`ISDx/configurations/ISD:x-AD:y.conf`) are not heavily used at the moment, so only topology files can be currently taken as input.

A possible scenario: a topology of some AD was modified on the centralized web panel (e.g. IP address changes, or a router was added while connecting a new AD). The new topology file is then pushed to the AD, where new supervisor/setup files are generated from the topology.

I also took the liberty of removing everything related to 'run' files (scripts which were used to run everything in 'screen'), since they are not used anymore, thanks to 'supervisor'.

@LorenzoBaesso, @kormat you might want to take a look.

### DIFF
--- a/topology/generator.py
+++ b/topology/generator.py
@@ -64,537 +64,591 @@ CS_RANGE = '21'
 PS_RANGE = '41'
 ER_RANGE = '61'
 
-default_subnet = "127.0.0.0/8"
-out_dir = TOPOLOGY_PATH
+DEFAULT_SUBNET = "127.0.0.0/8"
 
 
-def _get_subnet_params(ad_config):
-    if "subnet" in ad_config:
-        subnet = ad_config["subnet"]
-    else:
-        subnet = default_subnet
-    first_byte = subnet.split('.')[0]
-    mask = subnet.split('/')[1]
-    return first_byte, mask
-
-
-def _path_dict(isd_id, ad_id):
+class ConfigGenerator():
     """
-    Return a dictionary with the computed paths for a given AD.
+    Configuration and/or topology generator.
     """
-    isd_name = 'ISD{}'.format(isd_id)
-    file_no_ext = 'ISD:{}-AD:{}'.format(isd_id, ad_id)
+    def __init__(self, out_dir=TOPOLOGY_PATH, subnet=DEFAULT_SUBNET):
+        if not os.path.isdir(out_dir):
+            logging.error(out_dir + " output directory missing")
+            sys.exit()
 
-    setup_file_abs = os.path.join(out_dir, isd_name,
-                                  SETUP_DIR, file_no_ext + '.sh')
-    supervisor_file_abs = os.path.join(out_dir, isd_name,
-                                       SUPERVISOR_DIR, file_no_ext + '.conf')
+        self.out_dir = out_dir
+        self.subnet = subnet
 
-    topo_path_tail = os.path.join(isd_name, TOPO_DIR, file_no_ext + '.json')
-    topo_file_abs = os.path.join(out_dir, topo_path_tail)
-    topo_file_rel = os.path.join("..", SCRIPTS_DIR, topo_path_tail)
+    def _get_subnet_params(self, ad_config=None):
+        """
+        Return the first byte and the mask of the subnet.
 
-    conf_path_tail = os.path.join(isd_name, CONF_DIR, file_no_ext + '.conf')
-    conf_file_abs = os.path.join(out_dir, conf_path_tail)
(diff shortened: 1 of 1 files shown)


topology/generator.py
```diff
@@ -88,37 +132,31 @@ def increment_address(ip_address, mask, increment=1):
     ip_address_int = ip2int(ip_address)
     ip_address_int += increment
     ip_address = int2ip(ip_address_int)
-    bytes = ip_address.split('.')
-    bits = ''.join([format(int(bytes[0]), '08b'), format(int(bytes[1]), '08b'),
-                    format(int(bytes[2]), '08b'), format(int(bytes[3]), '08b')])
+    ip_bytes = ip_address.split('.')
+    bits = ''.join([format(int(byte), '08b') for byte in ip_bytes])
```
kormat: We should just switch to using python's ipaddress module:

```
$ python3
Python 3.4.0 (default, Apr 11 2014, 13:05:11) 
[GCC 4.8.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ipaddress
>>> ipaddress.ip_address('192.168.0.1')+1
IPv4Address('192.168.0.2')
```

topology/generator.py
```diff
@@ -69,6 +69,50 @@
...
+
+
+def _path_dict(isd_id, ad_id):
+    """
+    Return a dictionary with the computed paths for a given AD.
+    """
+    isd_name = 'ISD{}'.format(isd_id)
+    file_no_ext = 'ISD:{}-AD:{}'.format(isd_id, ad_id)
+
+    setup_file_abs = os.path.join(out_dir, isd_name,
```
kormat: I know this isn't your fault, but it would be great if we could stop using global variables like out_dir

tonyo: Actually, it's my fault ;) Was added with my previous PR. 
I just can't think about a nicer way to have the output directory path from anywhere, without dragging it as a function argument. 
Do you have any suggestions?

kormat: The usual fix at that point is to turn a bunch of functions into a class.

topology/generator.py
tonyo: Right, I'll fix this.

topology/generator.py
```diff
@@ -190,13 +227,28 @@ def write_keys_certs(AD_configs):
         enc_priv = base64.b64encode(enc_priv).decode('utf-8')
         write_file(sig_key_file, sig_priv)
         write_file(enc_key_file, enc_priv)
+
+    # Generate keys for self-signing
+    self_sign_id = ISD_AD_ID_DIVISOR.join(['0', '0'])
+    (sig_pub, sig_priv) = generate_signature_keypair()
+    (enc_pub, enc_priv) = generate_cryptobox_keypair()
```
LorenzoBaesso: The (enc_pub, enc_priv) pair was introduced when the crypto library was created. Unless I'm missing something, I think, these keys are still not used. Shall we maybe decide in this pull request if we need to keep them or not?

pszal: Yes, we don't need that. Please remove.

tonyo: Could someone clarify:  do we want to remove all mentions of `enc_pub` and `enc_priv` from the generator? Does this mean that we simply pass an empty string to `Certificate.from_values` as the 'subject_enc_key'?

pszal: Just occurred to me that at some point, it may be needed by one project. So... just leave that as it is.

topology/generator.py
```diff
@@ -65,507 +64,591 @@
...
+
+        try:
+            ad_configs = json.loads(open(adconfigurations_file).read())
+        except (ValueError, KeyError, TypeError):
+            logging.error(adconfigurations_file + ": JSON format error.")
+            sys.exit()
+
+        if "default_subnet" in ad_configs:
+            self.subnet = ad_configs["default_subnet"]
+            del ad_configs["default_subnet"]
```
kormat, line 642: Minor simplification possible:
self.subnet = ad_configs.pop("default_subnet")

tonyo, line 642: Thanks, will fix this in the next PR.

PR #95, opened 2015-05-04
Bs core pcbs fix

pszal:
- introduced unique pathstore instances for core pcbs (per isd,ad pair)
- the same for core_segments
- moved check_filters() (with derivatives)  to PathStore (is PathPolicy more accurate?)
- fixed default policy that discarded few core beacons (currently, some pcbs are still discarded by a property range, TO FIX)
- get_pathstore_from_pcb() probably will be removed in a future, as we decide how many policies we need in which beacon server

### DIFF
--- a/infrastructure/beacon_server.py
+++ b/infrastructure/beacon_server.py
@@ -74,7 +74,7 @@ class BeaconServer(SCIONElement):
 
     def __init__(self, addr, topo_file, config_file, path_policy_file):
         SCIONElement.__init__(self, addr, topo_file, config_file=config_file)
-        self.path_policy_file = path_policy_file
+        self.path_policy = PathPolicy(path_policy_file)  # TODO: add 2 policies
         self.unverified_beacons = deque()
         self.trc_requests = {}
         self.trcs = {}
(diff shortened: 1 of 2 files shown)


infrastructure/beacon_server.py
```diff
@@ -498,8 +458,23 @@ def __init__(self, addr, topo_file, config_file, path_policy_file):
...
+        self.beacons = defaultdict(self._ps_factory)
+        self.core_segments = defaultdict(self._ps_factory)
+
+    def _ps_factory(self):
+        return PathStore(PathPolicy(self.path_policy_file))
+
+    def get_pathstore_from_pcb(self, pcb):
+        """
+        Return PathStore instance that should store the pcb, creates an instance
+        is necessary. Core Beacon Server has PathStore instance per (isd, ad).
```
LorenzoBaesso: typo: if necessary

infrastructure/beacon_server.py
```diff
@@ -498,8 +458,23 @@ def __init__(self, addr, topo_file, config_file, path_policy_file):
...
+        """
+        # TODO: discuss with Lorenzo whether we need separate policies for
+        # core_segments and beacons. Now processing logic is quite unified (one
+        # filters check etc...), however there are two containers.
```
LorenzoBaesso: I think we should have two different policies but possibly keep the same processing logic. Does it make sense?

infrastructure/beacon_server.py
pszal: Thanks, actually this sentence is outdated, as I switched to defaultdict

infrastructure/beacon_server.py
pszal: two for core BS? and three for local BS?
Then we need to pre-validate each PCB against every policy and discard a
PCB only if every policy drops that (e.g. pcb can be dropped for
downstream propagation, however for up/down path may be ok).

Issue #96, opened 2015-05-04
Issues with PathStore


pszal:
@LorenzoBaesso  I looked at path-store.py, it seems like incomplete and untested. 
E.g. get_candidates() is executed, however ... it is not implemented. 
store_selection() or remove_segments()  is not even executed. 
Does PathStore instance free memory ?

PR #98, opened 2015-05-06
Path store bug

LorenzoBaesso:
This pull requests is meant to solve the issue #96.

Now:
- Best paths history in path store has a max length
- Expired segments are removed before being stored in history or returned
- Update fidelity is called wherever is needed (that is, any time the candidates are modified, added or deleted).
- There's a new unit test for path store that checks insertion, deletion, and storage into history.

### DIFF
--- a/lib/path_store.py
+++ b/lib/path_store.py
@@ -78,30 +78,32 @@ class PathPolicy(object):
         Checks whether any of the path properties has a value outside the
         predefined min-max range.
         """
-        return (
-            (self.property_ranges['PeerLinks'][0]
-             <= pcb.get_n_peer_links() <=
-             self.property_ranges['PeerLinks'][1])
-            and
-            (self.property_ranges['HopsLength'][0]
-             <= pcb.get_n_hops() <=
-             self.property_ranges['HopsLength'][1])
-            and
-            (self.property_ranges['DelayTime'][0]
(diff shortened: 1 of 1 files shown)


lib/path_store.py
```diff
@@ -345,8 +346,9 @@ def get_last_selection(self, k=None):
...
             k = self.path_policy.best_set_size
         best_paths = []
-        for candidate in self.best_paths_history[0][:k]:
-            best_paths.append(candidate.pcb)
+        if self.best_paths_history[0]:
+            for candidate in self.best_paths_history[0][:k]:
+                best_paths.append(candidate.pcb)
         return best_paths
 
     def store_selection(self, k=None):
```
pszal: @LorenzoBaesso 
I'm still not sure about logic here.
store_selection() is not executed, why?
Can you point me to a design doc for pathstore?

test/path_store_test.py
```diff
@@ -0,0 +1,94 @@
...
+from lib.packet.pcb import (PathSegment, ADMarking, PCBMarking, PeerMarking,
+    PathConstructionBeacon)
+from lib.path_store import PathPolicy, PathStore
+from lib.util import read_file, get_sig_key_file_path, sleep_interval
+import base64
+import logging
+import threading
+import time
+import unittest
+
```
pszal, line 31: do you need all these imports ?

PR #106, opened 2015-05-11
Add static methods from_file and from_dict to Topology class

tonyo:
At the moment Topology objects can only be created from a file (file path), but this becomes inconvenient when we want to to build a Topology instance from an existing object (dict or str). In this case some ugly hacks can be used (for example, using a NamedTemporatyFile), but, well, they are ugly.

This PR adds two static methods (`from_file` and `from_dict`) to the Topology class to solve the described issue. 

If you like the idea, I'll also refactor Config and PathPolicy classes in the similar fashion.

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -73,14 +73,14 @@ class Config(object):
         """
         Create a Config instance from the file.
 
-        :param config_file: path to the config file
+        :param config_file: path to the configuration file
         :type config_file: str
         :returns: the newly created Config instance
         :rtype: :class: `Config`
         """
         try:
-            with open(config_file) as config_fh:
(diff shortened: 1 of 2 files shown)


lib/topology.py
```diff
@@ -180,23 +178,49 @@ def __init__(self, topology_file=None):
         self.child_edge_routers = []
         self.peer_edge_routers = []
         self.routing_edge_routers = []
-        if topology_file:
-            self.parse(topology_file)
 
-    def parse(self, topology_file):
+    @staticmethod
```
kormat: This should be @classmethod - that's the convention for alternative constructors.

PR #112, opened 2015-05-15
Remove IP addresses from commandlines

kormat:
Building on @rev112's work, the infrastructure servers now get an ID on
the commandline, and then look up their specific config in the topology
file. It gets rid of redundant information (IP on commandline and in
topology file), allows each server to know their own identity, and means
i can remove the depdancy on the special supervisor env var.

@pszalach :)

### DIFF
--- a/infrastructure/scion_elem.py
+++ b/infrastructure/scion_elem.py
@@ -25,7 +25,6 @@ Module docstring here.
 from lib.config import Config
 from lib.packet.scion_addr import SCIONAddr
 from lib.topology import Topology
-import ipaddress
 import logging
 import select
 import socket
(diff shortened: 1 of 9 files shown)


infrastructure/scion_elem.py
```diff
@@ -51,16 +51,27 @@ class SCIONElement(object):
     :vartype addr: :class:`lib.packet.scion_addr.SCIONAddr`
     """
 
-    def __init__(self, host_addr, topo_file, config_file=None):
+    def __init__(self, server_type, topo_file, config_file=None, server_id=None,
```
pszal, line 54: maybe  'elem_type' is more accurate than 'server_type'  as e.g. SCIONDaemon inherits from this class too.

kormat, line 54: I'm unclear on the distinction between a 'server' and an 'element'. The SCIONElement doc strings are very inconsistent, talking about servers and ServerBase (old name?) objects.

pszal, line 54: Right. This is old name, and we switched to SCIONElem as router and sciond inherit from that.

PR #113, opened 2015-05-15
Added Zookeeper to cert server.

LorenzoBaesso:
New incoming Certificate Chains and TRCs are now shared between certificate servers by using a Zookeeper (in the same fashion the ZK is used by beacon servers to share PCBs).

### DIFF
--- a/infrastructure/cert_server.py
+++ b/infrastructure/cert_server.py
@@ -27,11 +27,11 @@ from lib.log import (init_logging, log_exception)
 from lib.thread import thread_safety_net
 from lib.util import timed
 from lib.zookeeper import (Zookeeper, ZkConnectionLoss, ZkNoNodeError)
-from parse import *
 import collections
 import datetime
 import logging
 import os
+import parse
 import sys
 import threading
 import time
@@ -53,7 +53,8 @@ class CertServer(SCIONElement):
         self.trc_requests = collections.defaultdict(list)
         self.cert_chains = {}
         self.trcs = {}
-        self._latest_entry = 0
+        self._latest_entry_cert_chains = 0
+        self._latest_entry_trcs = 0
         # Set when we have connected and read the existing recent and incoming
         # cert chains and TRCs
         self._state_synced = threading.Event()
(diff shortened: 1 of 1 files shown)


infrastructure/cert_server.py
```diff
@@ -18,23 +18,34 @@
...
+from lib.crypto.certificate import CertificateChain, TRC
 from lib.packet.scion import (SCIONPacket, get_type, PacketType as PT,
     CertChainRequest, CertChainReply, TRCRequest, TRCReply)
 from lib.util import (read_file, write_file, get_cert_chain_file_path,
     get_trc_file_path, handle_signals)
 from lib.log import (init_logging, log_exception)
+from lib.thread import thread_safety_net
+from lib.util import timed
+from lib.zookeeper import (Zookeeper, ZkConnectionLoss, ZkNoNodeError)
+from parse import *
```
kormat: "import *" is bad, as it means we can't tell later which calls are using the parse module.

infrastructure/cert_server.py
```diff
@@ -42,6 +53,15 @@ def __init__(self, addr, topo_file, config_file, trc_file):
         self.trc_requests = collections.defaultdict(list)
         self.cert_chains = {}
         self.trcs = {}
+        self._latest_entry = 0
```
kormat: As you have two cache directories, you'll need a latest entry variable for each of them

infrastructure/cert_server.py
```diff
@@ -174,6 +214,159 @@ def process_trc_reply(self, trc_rep):
...
+
+    @timed(1.0)
+    def _process_cached_cert_chains(self, entries):
+        """
+        Retrieve new cert chains from the shared cache and send them for local
+        processing.
+        """
+        # TODO(lorenzo): move constant to proper place
+        chunk_size = 10
+        pcbs = []
```
kormat: This shouldn't be here :)

infrastructure/cert_server.py
```diff
@@ -174,6 +214,159 @@ def process_trc_reply(self, trc_rep):
...
+                                  (self.ZK_CERT_CHAIN_CACHE_PATH, entry))
+                    continue
+                isd_id, ad_id, version = self._get_cert_chain_identifiers(entry)
+                self.cert_chains[(isd_id, ad_id, version)] = raw
+                cert_chain_file = get_cert_chain_file_path(self.topology.isd_id,
+                                                           self.topology.ad_id,
+                                                           isd_id, ad_id,
+                                                           version)
+                write_file(cert_chain_file, raw.decode('utf-8'))
+        return len(self.cert_chains)
```
kormat: You should be returning the number of new cert chains processed, not the total number known (if i understand the code correctly)

infrastructure/cert_server.py
```diff
@@ -174,6 +214,159 @@ def process_trc_reply(self, trc_rep):
...
+
+    @timed(1.0)
+    def _process_cached_trcs(self, entries):
+        """
+        Retrieve new TRCs from the shared cache and send them for local
+        processing.
+        """
+        # TODO(lorenzo): move constant to proper place
+        chunk_size = 10
+        pcbs = []
```
kormat: Ditto

PR #121, opened 2015-05-22
SCION DNS server, to allow discovery of service instances.

kormat:
Based on work by @pszalach and @LorenzoBaesso.

DNS server polls ZK to find service instances, and exposes this
information over DNS for remote hosts to query.

Status:
- The DNS server works
- lib/dns.py is not finished yet
- edge_router needs updating to query DNS
- Unit tests not written yet

@pszalach


infrastructure/dns_server.py
```diff
@@ -0,0 +1,513 @@
...
+"""
+# Stdlib
+import binascii
+import datetime
+import logging
+import os
+import sys
+import threading
+import time
+from ipaddress import (ip_address)
```
pszal: parentheses ?

lib/dns.py
```diff
@@ -0,0 +1,42 @@
...
+====================================
+"""
+# Stdlib
+import socket
+
+# External packages
+from dnslib import DNSRecord, QTYPE
+
+
+def get_hosts_by_srvname(service_name, dns_ip, dns_port=53, timeout=2):
```
pszal: dns_port=SCION_DNS_PORT ?

infrastructure/dns_server.py
kormat: Fixed

lib/dns.py
kormat: Fixed, and added a comment saying this function isn't finished/tested/anything.

PR #124, opened 2015-05-26
unittest for /lib/packet/ext_hdr.py

prateshg:


test/ext_hdr_test.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2014 ETH Zurich
```
kormat: 2015

test/ext_hdr_test.py
```diff
@@ -0,0 +1,70 @@
...
+from lib.packet.ext_hdr import (
+    ExtensionHeader,
+    ICNExtHdr,
+)
+from test.testcommon import SCIONCommonTest
+class TestExtensionHeaders(SCIONCommonTest):
+    """
+    Unit tests for ext_hdr.py.
+    """
+    def test_extension_header(self):
```
kormat: It's much better to make multiple test classes. In this case, i'd make a class each for ExtensionHeader.__init__, ExtensionHeader.parse, ExtensionHeader.pack, and then do the same again for ICNExtHdr.

test/ext_hdr_test.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        Ensure that parsing a packed extension header results in same extension
+        header
+        """
```
kormat: This is a slightly higher level than a unit test. Typically unit tests work at the level of a single method. As __init__ is very simple, all you need to test for it is that the attributes are set, and that parse() is called if raw is set.

test/ext_hdr_test.py
```diff
@@ -0,0 +1,70 @@
...
+        eh = ExtensionHeader()
+        self.assertFalse(eh.parsed)
+        ehCopy = ExtensionHeader()
+        ehCopy.parse(eh.pack());
+        self.assertTrue(eh.next_ext == ehCopy.next_ext and
+                        eh.hdr_len == ehCopy.hdr_len and
+                        ehCopy.parsed)
```
kormat: Doing asserts and'd together like this means that if /any/ of those statements is wrong, it'll throw an error, but it won't give you any clue which part fails. It's better to make a separate assertion for each check you're doing.

test/ext_hdr_test.py
```diff
@@ -0,0 +1,70 @@
...
+"""
+# Stdlib
+import unittest
+
+# SCION
+from lib.packet.ext_hdr import (
+    ExtensionHeader,
+    ICNExtHdr,
+)
+from test.testcommon import SCIONCommonTest
```
kormat: Two blank lines before top-level classes or functions

test/ext_hdr_test.py
```diff
@@ -0,0 +1,70 @@
...
+class TestExtensionHeaders(SCIONCommonTest):
```
kormat: I would prefer if all new tests use the nosetest suite. This means that your classes should (for now) just inherit from `object`, and your unittest.main at the end changes to:
`nose.run(defaultTest=__name__)`
The assertions also change. Have a look at how `ntools` is used in https://github.com/netsec-ethz/scion/blob/master/test/lib_zookeeper_test.py

test/ext_hdr_test.py
```diff
@@ -0,0 +1,70 @@
...
+        header
+        """
+        eh = ExtensionHeader()
+        self.assertFalse(eh.parsed)
+        ehCopy = ExtensionHeader()
+        ehCopy.parse(eh.pack());
+        self.assertTrue(eh.next_ext == ehCopy.next_ext and
+                        eh.hdr_len == ehCopy.hdr_len and
+                        ehCopy.parsed)
+    def test_equality(self):
```
kormat: One blank line between methods

lib/packet/ext_hdr.py
```diff
@@ -93,7 +93,7 @@ def __init__(self, raw=None):
     def parse(self, raw):
         assert isinstance(raw, bytes)
         dlen = len(raw)
-        if dlen < ExtensionHeader.MIN_LEN:
+        if dlen < ICNExtHdr.MIN_LEN:
```
kormat: Good catch, though `self.MIN_LEN` would be better

test/ext_hdr_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class TestExtensionHeaderInit(object):
+    """
+    Unit tests for lib.packet.ext_hdr.ExtensionHeader.__init__
+    """
+    def test_basic(self):
+        ext_hdr = ExtensionHeader()
+        ntools.assert_true(ext_hdr.next_ext is not None)
+        ntools.assert_true(ext_hdr.hdr_len is not None)
```
kormat: This would be better:

```
ntools.eq_(ext_hdr.next_ext, 0)
ntools.eq_(ext_hdr.hdr_len, 0)
```

test/ext_hdr_test.py
```diff
@@ -0,0 +1,119 @@
...
+        ntools.assert_false(ext_hdr.parsed)
+        ext_hdr2 = ExtensionHeader(ext_hdr.pack())
```
kormat: The ideal is that each unit test function should test exactly one thing. The current `test_basic` tests that ExtensionHeader.__init__ initializes the object correctly. You need 1 more test for __init__, to make sure that it calls self.parse if is called with a raw packet to parse.

As you only want to check one thing in that test, however, you don't want it actually execute self.parse. All you want to know is that self.parse is called with the raw packet.

This is done by patching the ExtensionHeader object to replace the parse method temporarily. With that in mind, here's how i'd write it:

```
from unittools.mock import patch

@patch("lib.packet.ext_hdr.ExtensionHeader.parse")
def test_raw(self, parse):
  ext_hdr = ExtensionHeader("data")
  parse.assert_called_once_with("data")
```

You've already tested the rest of the functionality in __init__ with test_basic, so you don't need to repeat those checks.

test/ext_hdr_test.py
```diff
@@ -0,0 +1,119 @@
...
+        ntools.assert_true(ext_hdr2.parsed)
+
+    def test_equality(self):
+        ext_hdr1 = ExtensionHeader()
+        ext_hdr2 = ExtensionHeader()
+        ntools.assert_true(ext_hdr1.next_ext == ext_hdr2.next_ext)
+        ntools.assert_true(ext_hdr1.hdr_len == ext_hdr2.hdr_len)
+        ntools.assert_true(ext_hdr1.parsed == ext_hdr2.parsed)
```
kormat: This test doesn't add anything to the other tests, so it can be removed.

test/ext_hdr_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+class TestExtensionHeaderPack(object):
+    """
+    Unit tests for lib.packet.ext_hdr.ExtensionHeader.pack
+    """
+    def test_basic(self):
+        ext_hdr = ExtensionHeader()
+        ntools.assert_true(len(ext_hdr.pack()) >= ext_hdr.MIN_LEN)
```
kormat: As the packed format is known, i would prefer to see this tested like:

```
ext_hdr = ExtensionHeader()
ext_hdr.next_ext = 14
ext_hdr.hdr_len = 42
ntools.eq_(ext_hdr.pack(), bytes([14,42])
```

test/ext_hdr_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+class TestExtensionHeaderParse(object):
+    """
+    Unit tests for lib.packet.ext_hdr.ExtensionHeader.parse
+    """
```
kormat, line 60: There are 2 parts of parse() that need to be tested. 

The first is what happens if the raw packet is too short. This can be tested by, for example, calling `parse(b" ")`, and checking that parsed is still False.

The second is that a raw packet is parsed correctly (and the parsed flag set). For example:

```
ext_hdr.parse(bytes([14,42])
ntools.eq_(ext_hdr.next_ext, 14)
ntools.eq_(ext_hdr.hdr_len, 42)
ntools.assert_true(ext_hdr.parsed)
```

test/ext_hdr_test.py
```diff
@@ -0,0 +1,118 @@
...
+    def test_basic(self):
+        ext_hdr = ExtensionHeader()
+        ext_hdr.parse(bytes([14,42]))
+        ntools.eq_(ext_hdr.next_ext, 14)
+        ntools.eq_(ext_hdr.hdr_len, 42)
+        ntools.assert_true(ext_hdr.parsed)
+
+    def test_len(self):
+        ext_hdr = ExtensionHeader()
+        ext_hdr.parse(bytes.fromhex('f0'))
```
kormat: `bytes([14])` would be slightly better, as it's consistent with the previous calls

test/ext_hdr_test.py
```diff
@@ -0,0 +1,118 @@
...
+        ntools.assert_false(ext_hdr.parsed)
```
kormat, line 71: It just occurred to me - it would also be good to check that `ext_hdr.next_ext` and `ext_hdr.hdr_len` are both still 0 (i.e. that a failed parse doesn't change the object's state).

test/ext_hdr_test.py
```diff
@@ -0,0 +1,118 @@
...
+        iext_hdr = ICNExtHdr()
+        iext_hdr.parse(bytes([14,42,10,0,0,0,0,0]))
+        ntools.eq_(iext_hdr.next_ext, 14)
+        ntools.eq_(iext_hdr.hdr_len, 42)
+        ntools.eq_(iext_hdr.fwd_flag, 10)
+        ntools.assert_true(iext_hdr.parsed)
+
+    def test_len(self):
+        iext_hdr = ICNExtHdr()
+        iext_hdr.parse(bytes.fromhex('f0 f1 f0 f1 f0 f1 f0'))
```
kormat: same comment here about making the bytes() call format consistent

PR #133, opened 2015-05-28
Unit tests for lib.config

tonyo:
A couple of unit tests for `lib.config`.

### DIFF
--- a/test/config_test.py
+++ b/test/config_test.py
@@ -52,13 +52,6 @@ class BaseLibConfig(object):
         'cert_chain_version': 'CertChainVersion',
     }
 
-    def _compare_attributes(self, config, config_dict):
-        ntools.eq_(len(config.__dict__),
-                   len(self.ATTRS_TO_KEYS),
-                   "Unequal number of keys/attributes: is something missing?")
-        for attr, key in self.ATTRS_TO_KEYS.items():
-            ntools.eq_(getattr(config, attr), config_dict[key])
-
 
 class TestConfigInit(BaseLibConfig):
     """
@@ -117,6 +110,13 @@ class TestConfigParseDict(BaseLibConfig):
         cfg = Config()
         cfg.parse_dict(self.config_json)
         self._compare_attributes(cfg, self.config_json)
+
+    def _compare_attributes(self, config, config_dict):
+        ntools.eq_(len(config.__dict__),
+                   len(self.ATTRS_TO_KEYS),
+                   "Unequal number of keys/attributes: is something missing?")
+        for attr, key in self.ATTRS_TO_KEYS.items():
+            ntools.eq_(getattr(config, attr), config_dict[key])
 
 
 if __name__ == "__main__":


test/config_test.py
```diff
@@ -0,0 +1,97 @@
...
+        'registers_paths': 'RegisterPath',
+        'pcb_queue_size': 'PCBQueueSize',
+        'path_server_queue_size': 'PSQueueSize',
+        'cert_chain_version': 'CertChainVersion',
+    }
+
+    def _compare_attributes(self, config, config_dict):
+        ntools.eq_(len(config.__dict__),
+                   len(self.ATTRS_TO_KEYS),
+                   "Unequal number of keys/attributes: is something missing?")
```
kormat: This can possibly by simplified to:

ntools.assert_dict_equal(config.__dict__, self.ATTRS_TO_KEYS)

test/config_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+class TestConfigInit(BaseLibConfig):
+    """
+    Unit tests for lib.config.Config.__init__
+    """
+    def test_basic(self):
+        config = Config()
+        for attr in self.ATTRS_TO_KEYS.keys():
+            ntools.assert_true(hasattr(config, attr),
+                               "No attribute found: {}".format(attr))
```
kormat: This check isn't necessary, as the next line will throw an exception if something is missing, which is sufficient.

test/config_test.py
```diff
@@ -0,0 +1,97 @@
...
+            ntools.eq_(getattr(config, attr), 0)
+
+
+class TestConfigFromDict(BaseLibConfig):
```
kormat, line 95: It's preferrable to keep tests in the same order as the target classes. (from_file comes before from_dict in lib.config.Config)

test/config_test.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    Unit tests for lib.config.Config.from_dict
+    """
+    def test_basic(self):
+        config = Config.from_dict(self.config_json)
```
kormat: All you should need to test for from_dict() is that it calls config.parse_dict with the supplied argument, and returns a new config object. To do this, you'll need to create a mock:

```
@patch("lib.config.Config.parse_dict")
def test_basic(self, parse_dict):
  ntools.assert_is_instance(Config.from_dict("some data"), Config)
  parse_dict.assert_called_once_with("some data")
```

test/config_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @ntools.raises(KeyError)
+    def test_invalid_dict(self):
+        Config.from_dict({'a': 'b'})
+
+
+class TestConfigFromFile(BaseLibConfig):
+    """
+    Unit tests for lib.config.Config.from_file
+    """
```
kormat, line 108: Same thing here as from_dict(). You want to check that it calls open with the supplied string, calls json.load on the contents, and either calls from_dict() if the read succeeds, or returns None if it fails. I'll quickly sketch out how that should look:

```
@patch("lib.config.Config.from_dict")
@patch("lib.config.json.load")
@patch("lib.config.open")
def test_success(self, open_, load, from_dict):
   from_dict.return_value = "All ok"
   ntools.eq_(Config.from_file("config file"), "All ok")
   ntools.open_.assert_called_once_with("config file")
   load.assert_called_once_with(open.return_value)
```

kormat, line 108: For errors, you'll want to use a generator to test all of them.

```
@patch("lib.config.Config.from_dict")
@patch("lib.config.json.load")
@patch("lib.config.open")
def _check_error(self, excp, open_, load, from_dict):
  # Setup
   load.side_effect = excp  # Raises an exception when json.load() is called
   from_dict.side_effect = SCIONTestException("from_dict should not have been called")
  # Call
   ntools.eq_(Config.from_file("config file"), None)

def test_error(self):
  for excp in (ValueError, KeyError, TypeError):
    yield self._check_error, excp
```

test/config_test.py
```diff
@@ -0,0 +1,97 @@
...
+    def test_basic(self):
+        config = Config.from_file(self.config_path)
+        ntools.assert_true(isinstance(config, Config))
+        self._compare_attributes(config, self.config_json)
+
```
kormat, line 121: A separate test class for parse_dict is needed. Call `cfg=Config()` as setup, then call `cfg.parse_dict(config)` (where config is a dictionary populated with the required key/values, and finally do compare_attributes.

The idea is that you should not test the same thing multiple times, if possible. It means if that one thing breaks, lots of unrelated tests fail, and it's much harder to find the root cause.

test/config_test.py
tonyo: That's not really what I want. First, I compare their length, then I check that `config` object and `config_dict` dict has the same values for the corresponding attribute/key. For example I check that `config.master_of_gen_key == config_dict['MasterOFGKey']`.

kormat: Ahh, ok. Nevermind then :)

test/config_test.py
```diff
@@ -0,0 +1,123 @@
...
+
+class TestConfigParseDict(BaseLibConfig):
+    """
+    Unit tests for lib.config.Config.parse_dict
+    """
+    def test_basic(self):
+        cfg = Config()
+        cfg.parse_dict(self.config_json)
+        self._compare_attributes(cfg, self.config_json)
```
kormat, line 112: As this is now the only place that `_compare_attributes` is called from, it probably makes sense to just move it into this test class.

PR #134, opened 2015-05-28
added unit test for lib/packet/packet_base.py

pratyakshs:


test/lib_packet_base_test.py
```diff
@@ -0,0 +1,155 @@
...
+    Unit tests for lib.packet.packet_base.PacketBase.payload
+    """
+    def test_bytes(self):
+        """
+        Test for setting payload as bytes.
+        """
+        packet_base = PacketBase()
+        packet_base.payload = b't9gj646'
+        payload = packet_base.payload
+        ntools.eq_(payload, b't9gj646')
```
kormat: So, this is not ideal because you're not testing a single method, you're actually testing 3 at once. It's much better if you can isolate the tests. E.g., test the getter by:

```
packet_base._payload = "blah"
ntools.eq(packet_base.payload, "blah")
```

kormat: For the setter, you want to be sure it calls set_payload with the value, but for testing purposes you want to mock set_payload.

```
@patch("lib.packet.packet_base.PacketBase.set_payload")
def test_setter(self, set_payload):
  packet_base = PacketBase()
  packet_base.payload = "blah"
  set_payload.assert_called_once_with("blah")
```

kormat: The getter/setter can be in the same test class, but the set_payload() tests should be a seperate class. For that, you will want to check that it raises a TypeError if given the wrong type of payload. Use a test generator for that.

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,155 @@
...
+        payload = PacketBase()
+        packet_base = PacketBase()
+        packet_base.payload = payload
+        ntools.eq_(packet_base.payload, payload)
+
+class TestPacketBaseHdr(object):
+    """
+    Unit tests for lib.packet.packet_base.PacketBase.hdr
+    """
+    def test_basic(self):
```
kormat, line 143: Same comments here about testing the getter/setter/set_hdr methods seperately.

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        Test for setting hdr as HeaderBase instance.
+        """
+        packet_base = PacketBase()
+        header = HeaderBase()
+        packet_base.hdr = header
+        ntools.eq_(packet_base.hdr, header)
+
```
kormat, line 165: __len__ is also worth testing.

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    Unit tests for lib.packet.packet_base.PacketBase.__eq__
+    """
+    def test_basic(self):
+        packet_base1 = PacketBase()
+        packet_base2 = PacketBase()
+        raw = "rawstring"
+        packet_base1.raw = raw
+        packet_base2.raw = raw
+        ntools.eq_(packet_base1, packet_base2)
```
kormat, line 176: I would instead use a mock'd object for the comparison:

```
  packet_base = PacketBase()
  raw = "rawstring"
  packet_base.raw = raw
  other = MagicMock(spec_set=PacketBase)
  other.raw = raw
  ntools.eq_(packet_base, other)
```

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,155 @@
...
+class TestPayloadBaseInit(object):
+    """
+    Unit tests for lib.packet.packet_base.PayloadBase.__init__
+    """
+    def test_basic(self):
+        """
+        Tests proper member initialization.
+        """
+        payload = PayloadBase()
+        ntools.eq_(payload.raw, None)
```
kormat: This can be simplified slightly to: `ntools.assert_is_none(payload.raw)`

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,155 @@
...
+        ntools.assert_false(payload.parsed)
+
+class TestPayloadBaseParse(object):
+    """
+    Unit tests for lib.packet.packet_base.PayloadBase.parse
+    """
+    def test_basic(self):
+        payload = PayloadBase()
+        payload.parse("rawstring")
+        ntools.eq_(payload.raw, "rawstring")
```
kormat: parse() attempts to make a copy of its argument. To test that, change this test to pass in a mutable list instead, and check to make sure that payload.raw isn't the same object afterwards:

```
raw = [1,2,3,4]
payload.parse(raw)
ntools.eq_(payload.raw, raw)
ntools.assert_is_not(payload.raw, raw)
```

test/lib_packet_base_test.py
pratyakshs, line 165: @kormat In `PacketBase._hdr` is an instance of `HeaderBase` for which `__len__` returns `None`. And `PacketBase.__len__` calls `len(self._hdr)`, so this always gives an error. Should I change `HeaderBase.__len__` to return `0`?

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    Unit tests for lib.packet.packet_base.PayloadBase.__len__
+    """
+    def test_basic(self):
+        """
+        Tests len() on a PayloadBase instance.
+        """
+        payload = PayloadBase()
+        payload.raw = "rawstr"
+        ntools.eq_(len(payload), len("rawstr"))
```
kormat, line 224: ~~Needs another test to make sure it returns 0 if raw is unset~~

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,155 @@
...
+    def test_basic(self):
+        """
+        Tests equality of two PayloadBase instances.
+        """
+        payload1 = PayloadBase()
+        payload2 = PayloadBase()
+        raw = "randomstring"
+        payload1.raw = raw
+        payload2.raw = raw
+        ntools.eq_(payload1, payload2)
```
kormat, line 247: ~~Using a mock would be better, as above.~~

test/lib_packet_base_test.py
kormat, line 165: `PacketBase._hdr` will be a _subclass_ of `HeaderBase`, and the subclass should override `__len__`, so that looks correct to me as it is.

test/lib_packet_base_test.py
pratyakshs: @kormat How do I generate payloads with different types using a test generator?

test/lib_packet_base_test.py
pratyakshs, line 176: `PacketBase.__eq__` compares the `type` of the two operands. I can't get `type(other)` to return `lib.packet.packet_base.PacketBase`; it returns `unittest.mock.MagicMock`.

pratyakshs, line 176: Can change the `type()` comparison to an `isinstance()` check in `PacketBase.__eq__`, would that be fine?

shitz, line 176: In case of **eq** I'm against changing it to isinstance() since you test
for equality. Maybe using a mock object here isn't appropriate, although
@kormat knows probably more about it.

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,247 @@
...
+        ntools.assert_is_none(packet_base._payload)
+        ntools.assert_false(packet_base.parsed)
+        ntools.assert_is_none(packet_base.raw)
+
+
+class TestPacketBasePayload(object):
+    """
+    Unit tests for lib.packet.packet_base.PacketBase.payload
+    """
+    def test_getter_bytes(self):
```
kormat, line 64: As the `payload` getter/setter do not actually care about the type of data, testing just with `bytes` is plenty.

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,247 @@
...
+        ntools.eq_(packet_base._payload, payload)
+        payload = PayloadBase()
+        packet_base.set_payload(payload)
+        ntools.eq_(packet_base._payload, payload)
+        payload = b'data'
+        packet_base.set_payload(payload)
+        ntools.eq_(packet_base._payload, payload)
+        ntools.assert_raises(TypeError, packet_base.set_payload, 123)
+        ntools.assert_raises(TypeError, packet_base.set_payload, '123')
+        ntools.assert_raises(TypeError, packet_base.set_payload, 123.4)
```
kormat, line 118: This attempts to test many things in a single test method, which isn't good. There are 2 things you need to test:
- If `set_payload` is called with any of `PacketBase`, `PayloadBase` or `bytes` that it sets `_payload` to that value.
- If `set_payload` is called with anything else, it raises a `TypeError`

This is best done with 2 sets of generators. Here's an example for the success case:

```
def check_success(self, payload):
  # Setup
  packet_base = PacketBase()
  # Call
  packet_base.set_payload(payload)
  # Tests
  ntools.eq_(packet_base._payload, payload)

def test_success(self):
  for i in PacketBase(), PayloadBase(), b'test':
    yield, self.check_success, i
```

kormat, line 118: For the failure case, you don't need a generator, just a single test method that calls set_payload with an invalid type.

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,247 @@
...
+        header = HeaderBase()
+        packet_base._hdr = header
+        ntools.eq_(packet_base.hdr, header)
+
+    @patch("lib.packet.packet_base.PacketBase.set_hdr")
+    def test_setter(self, set_hdr):
+        packet_base = PacketBase()
+        header = HeaderBase()
+        packet_base.hdr = header
+        set_hdr.assert_called_once_with(header)
```
kormat, line 136: There's no need to use a HeaderBase() payload here, just a simple string would suffice.

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,247 @@
...
+    Unit tests for lib.packet.packet_base.PacketBase.set_hdr
+    """
+    def test_basic(self):
+        packet_base = PacketBase()
+        header = HeaderBase()
+        packet_base.set_hdr(header)
+        ntools.eq_(packet_base._hdr, header)
+        ntools.assert_raises(TypeError, packet_base.set_hdr, 123)
+        ntools.assert_raises(TypeError, packet_base.set_hdr, '123')
+        ntools.assert_raises(TypeError, packet_base.set_hdr, 123.4)
```
kormat, line 150: This is very similar to the above set_payload tests, but this time there's only a single type that's valid, so 2 test methods are all that's required to test this properly, no generator needed.

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,247 @@
...
+    """
+    Unit tests for lib.packet.packet_base.PacketBase.__len__
+    """
+    def test_basic(self):
+        packet_base = PacketBase()
+        header = ExtensionHeader(b'data')
+        payload = b'data2'
+        packet_base.hdr = header
+        packet_base.payload = payload
+        ntools.eq_(len(packet_base), len(header) + len(payload))
```
kormat, line 163: There's no need to use an `ExtensionHeader` object here, as `__len__` doesn't care.

test/lib_packet_base_test.py
kormat, line 176: Ah, good point. Yeah, i can't see any good way to use a mock here, so what you've submitted is fine.

kormat, line 176: With one caveat: you've only checked the success case. There are 2 other cases which are untested:
- If `__eq__` is called with an unequal `PacketBase` object
- If `__eq__` is called with an object of a different type.

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,247 @@
...
+        ntools.assert_false(payload.parsed)
+
+
+class TestPayloadBaseParse(object):
+    """
+    Unit tests for lib.packet.packet_base.PayloadBase.parse
+    """
+    def test_basic(self):
+        payload = PayloadBase()
+        raw = [1,2,3,4]
```
kormat, line 198: Formatting nit-pick: there should be spaces after each ,:
`[1, 2, 3, 4]`

test/lib_packet_base_test.py
```diff
@@ -0,0 +1,247 @@
...
+        ntools.assert_is_not(payload.raw, raw)
+
+
+class TestPayloadBasePack(object):
+    """
+    Unit tests for lib.packet.packet_base.PayloadBase.pack
+    """
+    def test_basic(self):
+        payload = PayloadBase()
+        payload.parse("rawstring")
```
kormat, line 210: This call to `parse` is unnecessary, just set `payload.raw` directly in this test.

test/lib_packet_base_test.py
kormat, line 247: Needs to test the unequal and wrong type cases, same as above.

test/lib_packet_base_test.py
pratyakshs, line 163: @kormat Should I use a `HeaderBase` object? But then `HeaderBase.__len__` returns `NoneType`.

kormat, line 163: I would just use a byte string.

```
header = b'data1'
payload = b'data2'
```

PR #135, opened 2015-05-28
unit test added for lib/packet/scion_addr.py

mskd12:


test/scion_addr_test.py
```diff
@@ -0,0 +1,175 @@
...
+        addr_len = SCIONAddr.ISD_AD_LEN + (IPV6LENGTH // 8)
+
+        addr = SCIONAddr.from_values(isd_id, ad_id, host_addr)
+        ntools.eq_(addr.isd_id, isd_id)
+        ntools.eq_(addr.ad_id, ad_id)
+        ntools.eq_(addr.host_addr, host_addr)
+        ntools.eq_(addr.addr_len, addr_len)
+
+
+class TestSCIONAddrPack(object):
```
kormat: Please reorder this so that the test classe are in the same order as the module source (i.e. parse() is defined before pack() in scion_addr.py)

PR #141, opened 2015-06-01
Fix to inconsistency issue between ZK and file system.

LorenzoBaesso:
Changes:
- Compacted some parts of the code
- Handled the case in which the data in the ZK is inconsistent with the local files.

Solution: if the file of a cert/TRC was not found (probably because ./scion.sh topology has been executed), then do not use whatever is in the ZK, as it might be outdated (just send out another cert/TRC request). If a file was found, overwrite whatever is in the ZK, as there might still be old data.

Right after the new topology files have been created, the verification of the old beacons may still fail. But once the new certs/TRC have been propagated (very quick), the new paths will be registered correctly.

### DIFF
--- a/infrastructure/cert_server.py
+++ b/infrastructure/cert_server.py
@@ -20,12 +20,10 @@ import collections
 import datetime
 import logging
 import os
+import re
 import sys
 import threading
 import time
-
-# External packages
-import parse
 
 # SCION
 from infrastructure.scion_elem import SCIONElement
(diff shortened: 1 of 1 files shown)


infrastructure/cert_server.py
```diff
@@ -187,6 +207,7 @@ def process_trc_request(self, trc_req):
             if os.path.exists(trc_file):
                 trc = read_file(trc_file).encode('utf-8')
                 self.trcs[(trc_req.isd_id, trc_req.version)] = trc
+                self._store_trc_in_zk(trc_file, trc)
         if not trc:
```
pszal, line 211: I'm not sure whether this logic is correct.
Shouldn't CS check whether TRC is in ZK? (the same for certificates)

LorenzoBaesso, line 211: The file system must have higher priority in order to avoid inconsistency.

pszal, line 211: So how then TRC can be updated?

LorenzoBaesso, line 211: If a new version is out, then a TRC req must be sent. Upon reply the new TRC gets stored locally, in memory and in the ZK.

LorenzoBaesso, line 211: Besides the fact that, if a new CS is started, the servers in the same AD would require new topology files (and to be restarted probably).The new CS will ignore the ZK's content once, ask for the file once and from then on query the ZK. I think it's a reasonable price to pay in order to solve the inconsistency problem.

Issue #147, opened 2015-06-09
Tests fail


tonyo:
Currently, some tests from `opaque_field_test.py` and `lib_packet_base_test.py` fail. 
@prateshg @pratyakshs

```
....................................................................E....................................................................................FF.EFF......................................................
======================================================================
ERROR: test.lib_packet_base_test.TestPacketBaseLen.test_basic
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/tonyo/scion_ethz/scion/test/lib_packet_base_test.py", line 159, in test_basic
    header = ExtensionHeader(b'data')
  File "/home/tonyo/scion_ethz/scion/lib/packet/ext_hdr.py", line 41, in __init__
    self.parse(raw)
  File "/home/tonyo/scion_ethz/scion/lib/packet/ext_hdr.py", line 50, in parse
    self.next_ext, self.hdr_len = struct.unpack("!BB", raw)
struct.error: unpack requires a bytes object of length 2

======================================================================
ERROR: test.opaque_field_test.TestHopOpaqueFieldPack.test_basic
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/tonyo/scion_ethz/scion/test/opaque_field_test.py", line 169, in test_basic
    ntools.eq_(hop_op_fld.pack(),bytes.fromhex('0e 2a 0a 0b 0c 0d 0e 0f'))
  File "/home/tonyo/scion_ethz/scion/lib/packet/opaque_field.py", line 162, in pack
    data += self.mac
TypeError: can't concat bytes to int

======================================================================
FAIL: test.opaque_field_test.TestHopOpaqueFieldFromValues.test_less_arg
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/tonyo/scion_ethz/scion/test/opaque_field_test.py", line 155, in test_less_arg
    ntools.eq_(hop_op_fld.mac, 0)
AssertionError: b'\x00\x00\x00' != 0

======================================================================
FAIL: test.opaque_field_test.TestHopOpaqueFieldInit.test_basic
----------------------------------------------------------------------
[...]

PR #149, opened 2015-06-09
fixed issue #147

pratyakshs:


test/lib_packet_base_test.py
```diff
@@ -156,7 +157,7 @@ class TestPacketBaseLen(object):
     """
     def test_basic(self):
         packet_base = PacketBase()
-        header = ExtensionHeader(b'data')
+        header = ExtensionHeader(b'da')
```
pratyakshs, line 160: the `ExtensionHeader.parse` expects the argument to be exactly 2 bytes. So, changed that.

README.md
```diff
@@ -70,5 +70,5 @@ In order to run the unit tests:
 
 1. `cd test/`
 
-2. `nosetests`
+2. `nosetests3`
```
tonyo: IIUC, `nosetests3` is a part of the `python3-nose` system package, which is not installed by running `./scion.sh deps`. Perhaps we could just give a hint that the Py3 version of `nosetests` should be used here?
The cons of system packages are that they are installed system-wide, updated not that often (in my case `nosetests3` has version 1.3.1, while the latest `nose` from PyPi is 1.3.7) and may vary across distributions.

pratyakshs: @rev112 You're right. Even in the other case though, `nosetests` is a part of `python-nose` package - which is not installed by `./scion.sh deps`. Would you suggest changing this line back to `python3 *.py`?

pszal: I'm confused. Why we cannot just add python3-nose to deps ?

tonyo: `nosetests` is installed together with the `nose` Python package which is listed in `PIP3_DEPS` in `scion.sh` (and hence it's the Py3 version), so there's no need to install `python3-nose`.

So, actually, I don't see any serious reason to change this line in the first place.

pszal: Ok, I see, so @pratyakshs please revert that or elaborate if you don't agree.

pratyakshs: @rev112 I booted from a LiveUSB, and ran `./scion.sh deps`. It doesn't seem to be installing the `nosetests` command. It's of course installing the `nose` module in python3 correctly. IMHO, reverting this line back to `nosetests` isn't correct.

tonyo: @pratyakshs I suspect that you have `nosetests` installed (with the `nose` module), but it's not in your $PATH directories. Check your `~/.local/bin`, if the script is there, just add it to your path.

pratyakshs: @rev112 Agreed, and reverted.

kormat: We _could_ use the python3-nose system package, but we also need 2 nose plugins that aren't packaged by ubuntu, and those would have to be pulled in from pip3 in any case (and i've no idea if they'd work with a system nose package).

I think @shitz is going to add a note to the README about adding `~/.local/bin` to your path.

PR #151, opened 2015-06-10
Topology generator

mskd12:
- This PR introduces the first version of [BRITE](http://www.cs.bu.edu/brite) to SCION topology converter. Using the BRITE output, we convert it into a SCION topology.
- BRITE can be downloaded from this [link](http://www.cs.bu.edu/brite/download.html)
- Instructions to quickly install BRITE and generate a `*.brite` output file can be found in  `topology/brite/README.md`. A sample BRITE output file with 250 nodes is also included - `topology/brite/isd1.brite`
- `topology_generator.py` can be found in `topology/brite` directory. 
- After generating a brite output file, type `python3 topology_generator.py {brite_file}` to create a `ADConfigurations.json` file


topology/brite/topology_generator.py
```diff
@@ -0,0 +1,219 @@
...
+    fd = open(topo_file, 'r')
+    parsing_nodes = False
+    parsing_edges = False
+
+    original_graph = nx.Graph()
+    num_outedges = list()
+    for line in fd:
+        values = line.split(" ")
+        if parsing_nodes:
+            if line == "\n":
```
shitz: The generated topology file tells you how many nodes and edges were generated. Use that as a counter instead of relying on \n.

topology/brite/topology_generator.py
```diff
@@ -0,0 +1,219 @@
...
+                dest_isd_ad_id = ISD_AD_ID_DIVISOR.join([str(ISD_NUM),
+                                                         values[2]])
+                original_graph.add_edge(source_isd_ad_id, dest_isd_ad_id)
+
+        if values[0] == "Nodes:":
+            parsing_nodes = True
+        elif values[0] == "Edges:":
+            parsing_edges = True
+
+    NUM_CORE_ADS = min(MAX_CORE_ADS, int(len(original_graph.nodes())/10))
```
shitz: whitespaces around `/`

topology/brite/topology_generator.py
```diff
@@ -0,0 +1,219 @@
...
+        core_ad_graph = graphs[0]
+        core_ads = core_ad_graph.nodes()
+        num_extra_nodes = NUM_CORE_ADS - len(core_ads)
+        neighbor_nodes = set()
+        for neighbor in [original_graph.neighbors(node) for node in core_ad_graph]:
+            for node in neighbor:
+                if node not in core_ad_graph.nodes():
+                    neighbor_nodes.add(node)
+
+        neighbor_nodes = sorted(neighbor_nodes,
```
shitz, line 97: Does this select the neighbor nodes with the highest degree? Please add a comment.

topology/brite/topology_generator.py
```diff
@@ -0,0 +1,219 @@
...
+    level = dict()
+    for node in core_ads:
+        level[node] = 1
+    visited = dict()
+    for core_ad in core_ads:
+        visited[core_ad] = 1
+    while queue:
+        node = queue.popleft()
+        for neighbor in original_graph.neighbors(node):
+            if neighbor in core_ads:
```
shitz, line 126: Just check `if visited[neighbor]:` since you already set it to 1 for all core ads. (Only works if you define `visited = defaultdict(lambda: 0)` otherwise you have to work with `.get`

Nevermind, I see that you do things for already visited nodes as well...

topology/README.md
```diff
@@ -0,0 +1,24 @@
+SCION Topology Generator
```
shitz: Please move this file to `topology/brite` as well. Otherwise it looks good to me.

PR #153, opened 2015-06-10
New documentation

LorenzoBaesso:
First part of the documentation. 

Major work/review needs to be done on the following folders:
- infrastructure
- lib/
- lib/crypto
- lib/packet

### DIFF
--- a/infrastructure/path_server.py
+++ b/infrastructure/path_server.py
@@ -55,14 +55,14 @@ class PathServer(SCIONElement):
 
     def __init__(self, server_id, topo_file, config_file):
         """
-        Constructor.
+        Initialize an instance of the class PathServer.
 
-        :param :
-        :type :
-        :param :
-        :type :
-        :param :
-        :type :
+        :param server_id:
+        :type server_id:
+        :param topo_file:
+        :type topo_file:
+        :param config_file:
+        :type config_file:
         """
         SCIONElement.__init__(self, "ps", topo_file, server_id=server_id,
                               config_file=config_file)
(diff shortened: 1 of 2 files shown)


infrastructure/beacon_server.py
```diff
@@ -100,6 +100,16 @@ class InterfaceState(object):
     IFID_TOUT = 3.5 * IFID_PKT_TOUT
 
     def __init__(self):
+        """
+        Constructor.
```
pszal: I'm not sure whether it helps to have these docstring without param names and where even number of params is incorrect (the same comment applies for almost all constructors below).
IMO it is better to have 10% of correct docstrings, than 100% of incorrect templates. 
What do you think?

LorenzoBaesso: Of course. Indeed it's not complete. But I need you to remove the classes that are not used anymore and after that I will pass again the docstrings.

ad_management/common.py
```diff
@@ -1,3 +1,20 @@
+# Copyright 2014 ETH Zurich
```
pszal, line 1: 2015 (the same for other new copyrights)

ad_management/secure_rpc_server.py
```diff
@@ -1,10 +1,23 @@
 #!/usr/bin/env python3
```
pszal, line 1: is this our code?

tonyo, line 1: It's my experimental stuff, I'll add copyrights sometime soon.

tonyo, line 1: Oh well, I mean, yes, it's based on something from SO, but it's not a copy-paste.

pszal, line 1: ok!

PR #154, opened 2015-06-11
Addr tuple resize

pszal:
- (isd, ad) pair resized to 4B
- redesigning pcb structure and parsing (removing weird classes from there, unified AD and Peer Markings, removed unused fields, but still some work needed)
- correct signing/verifying for PCB introduced

TODOs:
- revisit/do signing of up,downpaths
- adjust sciond's API to 4B isd,ad (discuss with Jason)
- ... doc

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,14 @@
+machine:
+  python:
+    version: 3.4.2
+
+dependencies:
+  cache_directories:
+    - "/var/cache/apt"
+  post:
+    - "./scion.sh init"
+    - "./scion.sh topology"
+
+test:
+  override:
+    - ./scion.sh test
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,14 @@
+Sphinx
+alabaster==0.7.4
+coverage
+dnslib
+kazoo
+nose
+nose-cov
+nose-descriptionfixer
+parse
+pycrypto
+pydblite
+pygments
+python-pytun
+sphinxcontrib-napoleon
--- a/scion.sh
+++ b/scion.sh
@@ -3,7 +3,6 @@
 # BEGIN subcommand functions
 
 PKG_DEPS="python python3 python-dev python-pip python3-dev python3-pip screen zookeeperd build-essential docker.io dnsutils"
-PIP3_DEPS="python-pytun pydblite pygments pycrypto kazoo alabaster==0.7.4 Sphinx sphinxcontrib-napoleon nose nose-descriptionfixer nose-cov coverage parse dnslib"
 
 cmd_deps() {
     # Treat all non-zero returns as fatal errors. Prevents issues like pip
@@ -16,7 +15,7 @@ cmd_deps() {
         echo "    $PKG_DEPS"
     fi
     echo "Installing necessary packages from pip3"
-    pip3 install --user $PIP3_DEPS
+    pip3 install --user -r requirements.txt
     echo "Installing supervisor packages from pip2"
     pip2 install --user supervisor==3.1.3
     pip2 install --user supervisor-quick


infrastructure/beacon_server.py
```diff
@@ -485,17 +476,31 @@ def _verify_beacon(self, pcb):
...
+        new_pcb = copy.deepcopy(pcb)
+        new_pcb.trcf.if_id = 0
+        new_pcb.ads[-1].sig = b''
+        new_pcb.ads[-1].sig_len = 0
+        return verify_sig_chain_trc(new_pcb.pack(), pcb.ads[-1].sig, subject,
                                     chain, trc, trc_version)
 
+    def _sign_beacon(self, pcb):
+        """
+        TODO: signature goes to the last ADMarking
```
shitz: What exactly is the TODO here? Also please still add a comment what a the method is supposed to do.

infrastructure/beacon_server.py
```diff
@@ -485,17 +476,31 @@ def _verify_beacon(self, pcb):
...
+        """
+        # if_id field is excluded from signature as it is changed by ingress ERs
+        if pcb.ads[-1].sig:
+            logging.error("PCB already signed.")
```
shitz: I wouldn't say this is an error.

lib/packet/pcb.py
```diff
@@ -111,34 +108,32 @@ def parse(self, raw):
...
             return
-        self.ad_id = struct.unpack("!Q", raw[:8])[0]
-        self.ssf = SupportSignatureField(raw[8:16])
-        self.hof = HopOpaqueField(raw[16:24])
-        self.spcbf = SupportPCBField(raw[24:32])
-        self.ig_rev_token = raw[32:64]
-        self.eg_rev_token = raw[64:96]
+        (self.isd_id, self.ad_id) = ISD_AD.from_raw(raw[:ISD_AD.LEN])
+        offset = ISD_AD.LEN
+        self.hof = HopOpaqueField(raw[offset:offset+HopOpaqueField.LEN])
```
shitz: whitespaces around `+` here and below

lib/packet/pcb.py
```diff
@@ -111,34 +108,32 @@ def parse(self, raw):
...
+        offset += HopOpaqueField.LEN
+        self.ig_rev_token = raw[offset:offset+32]
+        offset += 32
+        self.eg_rev_token = raw[offset:offset+32]
         self.parsed = True
 
     @classmethod
-    def from_values(cls, ad_id=0, ssf=None, hof=None, spcbf=None,
+    def from_values(cls, isd_id=0, ad_id=0, hof=None,
```
shitz: Should all of these parameters really have default values? Does it make sense to have a marking with these default values?

lib/packet/scion_addr.py
```diff
@@ -22,15 +22,29 @@
 from ipaddress import IPV4LENGTH, IPV6LENGTH, IPv4Address, IPv6Address
 
 
-ISD_AD = namedtuple('ISD_AD', ['isd', 'ad'])
+class ISD_AD(namedtuple('ISD_AD', 'isd ad')):
```
shitz, line 25: Can we come up with a good name for the (ISD, AD) tuple?

pszal, line 25: This is actually address of an AD, so maybe ADAddr ?
For the naming I would also replace ADMarking by ADBlock, and PCBMarking by HopMarking. Any opinions are welcome.

lib/packet/pcb.py
```diff
@@ -288,15 +191,18 @@ def parse(self, raw):
         assert isinstance(raw, bytes)
         self.raw = raw[:]
         dlen = len(raw)
-        if dlen < ADMarking.LEN:
+        if dlen < PCBMarking.LEN + 8:
```
kormat: Magic numbers like `8` are bad. Can this be a constant please?

lib/packet/scion_addr.py
kormat, line 25: `ISD_AD` seems ok to me as a name

lib/packet/scion_addr.py
```diff
@@ -22,15 +22,29 @@
...
+    """
+    TODO
+    """
```
kormat, line 33: It would be good to have a bit-level description here of the packed data. Something saying that the ISD id is the top 12 bits, and the AD id is the bottom 20 bits, for example.

test/opaque_field_test.py
```diff
@@ -440,281 +437,5 @@ def test_basic(self):
         trc_fld.reserved = 0x0f
         ntools.eq_(trc_fld.pack(), bytes.fromhex('0f 2a 0a 0b 0c 0d 0e 0f'))
 
-
-class TestTRCFieldEq(object):
```
prateshg, line 444: @pszalach this class shouldn't be deleted

PR #160, opened 2015-06-16
If revocation

shitz:
Use the interface timeout detection code by @pszalach to issue a revocation for the timed-out interface. I haven't put any thought to which BS should actually issue the revocation. If there are multiple, everyone of them would issue a revocation, however, this should only lead to some message overhead and not breakage.

Also added a toplevel **init**.py file to make Eclipse shut up about missing imports.

### DIFF
--- a/infrastructure/path_server.py
+++ b/infrastructure/path_server.py
@@ -801,13 +801,9 @@ class LocalPathServer(PathServer):
             return
         leases = []
         for pcb in records.pcbs:
-<<<<<<< HEAD
             assert pcb.segment_id != 32 * b"\x00", ("Trying to register a" +
                    " segment with ID 0:\n%s" % pcb)
-            src_isd = pcb.get_first_pcbm().spcbf.isd_id
-=======
             src_isd = pcb.get_first_pcbm().isd_id
->>>>>>> master
             src_ad = pcb.get_first_pcbm().ad_id
             dst_ad = pcb.get_last_pcbm().ad_id
             dst_isd = pcb.get_last_pcbm().isd_id


infrastructure/beacon_server.py
```diff
@@ -616,6 +635,35 @@ def _process_cached_pcbs(self, entries):
...
+
+    @thread_safety_net("_handle_if_timeouts")
+    def _handle_if_timeouts(self):
+        """
+        Periodically checks each interface state and issues an if revocation, if
+        no keep-alive message was received for IFID_TOUT.
+        """
+        while True:
+            start_time = time.time()
+            for (if_id, if_state) in self.ifid_state.items():
```
kormat, line 740: ~~This is not thread-safe.~~

infrastructure/beacon_server.py
```diff
@@ -616,6 +635,35 @@ def _process_cached_pcbs(self, entries):
...
+                # Check if interface has timed-out.
+                if if_state.is_expired():
+                    if_state.state = InterfaceState.REVOKED
```
kormat: This design isn't clean - we're changing the internal state of the state object directly. It probably makes more sense to change the state inside `is_expired`.

infrastructure/beacon_server.py
```diff
@@ -850,10 +898,72 @@ def register_core_segments(self):
...
             new_pcb.add_ad(ad_marking)
             new_pcb.segment_id = self._get_segment_rev_token(new_pcb)
+            assert new_pcb.segment_id != 32 * b"\x00", ("Trying to register a" +
+                   " segment with ID 0:\n%s" % new_pcb)
             self.register_core_segment(new_pcb)
             count += 1
         logging.info("Registered %d Core paths", count)
 
+    def _process_revocation(self, rev_info):
+        BeaconServer._process_revocation(self, rev_info)
```
kormat: It's cleaner to use `super()._process_revocation` instead

infrastructure/beacon_server.py
```diff
@@ -616,6 +635,35 @@ def _process_cached_pcbs(self, entries):
...
+                # Check if interface has timed-out.
+                if if_state.is_expired():
+                    if_state.state = InterfaceState.REVOKED
+                    logging.info("Issuing revocation for IF %d.", if_id)
+                    # Issue revocation
+                    assert if_id in self.if2rev_tokens
+                    chain = self.if2rev_tokens[if_id]
+                    rev_info = RevocationInfo.from_values(RT.INTERFACE,
+                        chain.current_element(), chain.next_element())
+                    self._process_revocation(rev_info)
```
kormat, line 749: This may not be thread-safe in the future, either.

infrastructure/beacon_server.py
```diff
@@ -850,10 +898,72 @@ def register_core_segments(self):
                                                  new_pcb.get_last_pcbm().hof)
             new_pcb.add_ad(ad_marking)
             new_pcb.segment_id = self._get_segment_rev_token(new_pcb)
+            assert new_pcb.segment_id != 32 * b"\x00", ("Trying to register a" +
+                   " segment with ID 0:\n%s" % new_pcb)
             self.register_core_segment(new_pcb)
             count += 1
         logging.info("Registered %d Core paths", count)
 
+    def _process_revocation(self, rev_info):
```
kormat: This method is so large that it's going to be ~impossible to test. I suggest splitting the different rev cases into smaller methods that get called by this.

infrastructure/beacon_server.py
```diff
@@ -97,31 +96,73 @@ class InterfaceState(object):
...
-        self.active_from = 0
-        self.active_until = 0
+        self.active_since = 0
+        self.last_updated = 0
+        self._state = self.INACTIVE
+        self._lock = threading.RLock()
 
-    def update(self):
+    @property
+    def state(self):
```
kormat: Does this need to be exposed as a property? Generally properties are used when you have a lot of existing code that expects your object to have a certain attribute, and access it directly.

shitz: Hmm the state by itself not, but the setter, since we want to set it to REVOKED from the outside as soon as we have issued a revocation for it.

infrastructure/beacon_server.py
```diff
@@ -97,31 +96,73 @@ class InterfaceState(object):
...
 
+    def update(self):
         """
-        return self.active_until + self.IFID_TOUT >= time.time()
+        Updates the state of the object.
+        """
+        with self._lock:
+            curr_time = time.time()
+            if (self.state != self.ACTIVE and
+                self.last_updated + self.IFID_TOUT < curr_time):
```
kormat: This line should be indented another step, so that it's distinct from the following lines.

kormat: Actually, i'm not sure of the purpose of the time check. If we already know the IF is not active, and we've gotten an update for it, then the time check is irrelevant, right?

kormat: Hurm. I'm unclear on how this works. If the state is REVOKED, what's supposed to happen?

infrastructure/beacon_server.py
```diff
@@ -97,31 +96,73 @@ class InterfaceState(object):
...
 
+    def is_active(self):
+        if self.state == self.ACTIVE:
+            if self.last_updated + self.IFID_TOUT >= time.time():
+                return True
+            else:
+                self.state = self.TIMED_OUT
+                return False
+        else:
+            return False
```
kormat, line 138: This can be simplified a bit, by using implied states:

```
        if self.state == self.ACTIVE:
            if self.last_updated + self.IFID_TOUT >= time.time():
                return True
            self.state = self.TIMED_OUT
            return False
        return False
```

I find it easier to read, at least.

infrastructure/beacon_server.py
```diff
@@ -97,31 +96,73 @@ class InterfaceState(object):
...
+            return True
+        elif (self.state == self.ACTIVE and
+              self.last_updated + self.IFID_TOUT < time.time()):
+            self.state = self.TIMED_OUT
+            return False
+        else:
+            return False
+
+    def is_revoked(self):
+        return (self.state == self.REVOKED)
```
kormat: No need for the ()'s

infrastructure/beacon_server.py
shitz: I think you are right about the time check.

If the interface is revoked and we get an update for that interface, we can set its state to ACTIVE again.

infrastructure/beacon_server.py
```diff
@@ -97,31 +96,73 @@ class InterfaceState(object):
...
+                self.state = self.ACTIVE
+                logging.debug('Interface activated')
+            self.last_updated = curr_time
 
+    def is_active(self):
+        if self.state == self.ACTIVE:
+            if self.last_updated + self.IFID_TOUT >= time.time():
+                return True
+            else:
+                self.state = self.TIMED_OUT
```
kormat: There are thread-safe issues here. You're reading 2 attributes, then changing 1. With multiple threads modifying this object, there's going to be subtle race conditions. To make it fully thread-safe, the entire method needs to be wrapped by a lock.

infrastructure/beacon_server.py
kormat: We can just have a 1-line method to do specifically that.

shitz: Fair enough.

infrastructure/beacon_server.py
```diff
@@ -616,6 +657,92 @@ def _process_cached_pcbs(self, entries):
...
+        """
+        Periodically checks each interface state and issues an if revocation, if
+        no keep-alive message was received for IFID_TOUT.
+        """
+        while True:
+            start_time = time.time()
+            for (if_id, if_state) in self.ifid_state.items():
+                # Check if interface has timed-out.
+                if if_state.is_expired():
+                    with if_state._lock:
```
kormat: Actually, i'm not sure this needs a lock. You now don't read any state from the object in this block, until you finally call set_revoked (which does its own locking).

infrastructure/beacon_server.py
```diff
@@ -850,10 +977,79 @@ def register_core_segments(self):
...
+        rev_infos = []
+        to_remove = []
+        path_store = None
+        for ps in self.core_segments.values():
+            if ps.get_segment(rev_info.seg_id):
+                path_store = ps
+                break
+            if path_store is None:
+                logging.warning("Segment to revoke does not exist.")
+                break
```
kormat: I'm not following the logic here. This if statement will always match, if i understand correctly. `path_store` is initialised to `None`, and the only place it gets assigned to also calls `break`.

kormat: So right now the rest of the for loop will never be reached

infrastructure/beacon_server.py
shitz: What if in the mean time the interface gets activated again? Then we would issue a revocation (which is correct) but update the state to revoked afterwards (which is not ok, since it is active).

infrastructure/beacon_server.py
shitz: Yeah this is a copy paste fail...

infrastructure/beacon_server.py
kormat: Ah hah, yep, that's a good point. :)

kormat: Thinking about this a bit more, there's no reason you can't do the check and state change with a single method call. You could change the call to be `revoke_if_expired`, or add a keyword arg to the method `revoke=True`. This avoids holding the lock for longer than necessary, removes a method call, and (most importantly for my taste :) means you're no longer interacting with the object's internal lock directly.

infrastructure/beacon_server.py
kormat: This bit looks good now.

infrastructure/beacon_server.py
shitz: Changed it to not acquire the lock and call `revoke_if_expired` after the revocation got issued.

kormat: Ok, that works.

infrastructure/beacon_server.py
kormat, line 749: I still have concerns here. Right now, it looks like `_process_revocation` is only called from the `_handle_if_timeouts` thread, but that's going to change in the future, right?

shitz, line 749: Yeah it might get called from other places in the future.

infrastructure/beacon_server.py
```diff
@@ -859,11 +986,80 @@ def register_core_segments(self):
                                                  new_pcb.get_last_pcbm().hof)
             new_pcb.add_ad(ad_marking)
             new_pcb.segment_id = self._get_segment_rev_token(new_pcb)
+            assert new_pcb.segment_id != 32 * b"\x00", ("Trying to register a" +
```
pszal, line 989: why exactly this case is asserted?

shitz, line 989: Because I encountered a weird case where the revoked segment ID was set to 32 \* \x00 after revoking an interface, i.e., a segment that contained this interface ID had a 0 segment ID, which mustn't happen. This prevents a BS to ever register a segment with an ID of 0.

pszal, line 989: I see... Maybe it is unrelated but remember that zookeeper keeps seen beacons even when you `./scion.sh stop`. To flush this I have to `stop zk; rm /var/lib/zookeeper/version-2/* ; start zk`

shitz, line 989: So far this hasn't lead to any problems. In any case, `_get_segment_rev_token` should never return a 32 \* \x00 segment ID.

infrastructure/path_server.py
```diff
@@ -792,7 +801,13 @@ def _handle_down_segment_record(self, pkt):
             return
         leases = []
         for pcb in records.pcbs:
+<<<<<<< HEAD
+            assert pcb.segment_id != 32 * b"\x00", ("Trying to register a" +
+                   " segment with ID 0:\n%s" % pcb)
+            src_isd = pcb.get_first_pcbm().spcbf.isd_id
+=======
```
pszal: resolve this conflict (spcbf is gone)

PR #166, opened 2015-06-16
Run other topos

syclops:
Modified topology/generator.py with the argparse module. This allows the user to specify an AD configurations file, path policy file, and output directory individually. Thus rather than having to either use all of the defaults or specify all three paths, the user can specify any or all of the paths, and the remaining paths will be set to their respective default values.

To test this functionality, I also added a tiny two-node ISD topology. In the future I can add alternate path policies, but I don't think that's as important at this stage.


topology/TinyConfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
pszal: what is a reason to have these test topologies in master? For automagic topology generation you can look at recent work on scion_simulator

topology/generator.py
```diff
@@ -16,6 +16,7 @@
 =============================================
```
pszal, line 16: is it consistent with scion.sh topology ?
Probably @LorenzoBaesso and @rev112 should look at this.

topology/TinyConfig.json
syclops: The test topologies are primarily for testing various path policies with a beacon server and evaluating fidelity, but they may be useful if we release the code and want to let others use our test topologies. If you think that this shouldn't be included in `master`, I'm happy to just merge it into `fidelity_test` instead.

The `simulator` branch, from what I can tell, currently only generates a two-node topology and runs a ping test. I didn't see any other code for automatic topology generation, although that's something I could probably implement as well.

mskd12: The automatic topology generation part is not just part of the `simulator` branch. It is also present in `master` branch. 
You can refer to #151 for its details.

syclops: Thanks, Krishna/@mskd96! I briefly looked through the code and it seems pretty nice, but I'll have to play around with it a bit more to see. This might be very helpful for the task I'm currently working on...

topology/generator.py
tonyo, line 16: The issue here is that when calling `./scion.sh topology` we enter `topology/` directory first, and only then call `generator.py`. It means that all relative paths given to `scion.sh topology` will be expanded incorrectly in `generator.py`.
A possible solution is to avoid changing directories in `cmd_topology()` ( in `scion.sh`).

syclops, line 16: I agree. That's a separate issue from the one I fixed, but it's something else that can (and probably should) be fixed. IMO the way to fix this would be to run `generator.py` from the root directory. Should I add this as well?

tonyo, line 16: If you don't mind. This shouldn't break anything, and you'll fix several issues at once, isn't it great? ;)

syclops, line 16: This bug is fixed.

topology/TinyConfig.json
syclops: @pszalach please let me know if the test topologies should be merged into `master`. If not, please do the following:
- merge the changes to `topology/generator.py`
- post a reply so I can merge the topologies into the `fidelity_test` branch.

pszal: I don't see a good reason to have these topologies in master, so please drop it from this PR.

syclops: Dropped.

PR #168, opened 2015-06-17
Add initial circle.yml for circleci.com

kormat:
Also moving the pip3 requiremnts into requirements.txt, which is better
supported by pip, and supported automatically by circle ci.

(This is still in testing, so please don't merge yet)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,14 @@
+machine:
+  python:
+    version: 3.4.2
+
+dependencies:
+  cache_directories:
+    - "/var/cache/apt"
+  post:
+    - "./scion.sh init"
+    - "./scion.sh topology"
+
+test:
+  override:
+    - ./scion.sh test
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,14 @@
+Sphinx
+alabaster==0.7.4
+coverage
+dnslib
+kazoo
+nose
+nose-cov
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+kazoo
+nose
+nose-cov
+nose-descriptionfixer
+parse
+pycrypto
+pydblite
+pygments
+python-pytun
+sphinxcontrib-napoleon
```
tonyo, line 14: Am I correct to assume that CircleCI runs `pip install -r requirements.txt` automatically?

circle.yml
```diff
@@ -0,0 +1,14 @@
+machine:
+  python:
+    version: 3.4.2
+
+dependencies:
+  cache_directories:
+    - "/var/cache/apt"
```
tonyo, line 7: What is the workflow when new system dependencies are added? Does someone have to start a build without cache?

requirements.txt
kormat, line 14: Correct.

circle.yml
kormat, line 7: No, circle-ci doesn't seem to do anything clever with the cache. It will just restore the old data. apt can then do an update and fetch new packages as needed. We're not actually using apt, just yet, but we'll see how it goes.

PR #757, opened 2016-06-03
Libssock integration test

aznair:
Adds SSP testing to integration tests and includes some changes to make it work:

Dispatcher
- Log file sizes updated to 10MB
- SSP packet recipient lookup goes through either port number (wildcard) or flow ID, not both
  - Any packet after initial connection establishment contains port number 0, uses flow ID only

Router
- Packet is <code>deliver()</code>ed if current isd_as is destination AND current hof is in the last segment of the path

Libssock
- Socket gets default IP address from OS, application can call <code>bind()</code> to specify own address
- Bug fixes, mostly dealing with deadlocks during connection termination

Proxy
- Destination address changed to 127.0.0.1 (should probably be a cmdline arg, not for this PR)

Integration test
- Socket creation is done in method, SSP test overrides to create SSP socket
- <code>get_path()</code> now optional, SSP test doesn't do it (done internally in socket)

---

This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/757)

### DIFF
--- a/endhost/ssp/ConnectionManager.cpp
+++ b/endhost/ssp/ConnectionManager.cpp
@@ -448,15 +448,9 @@ int SSPConnectionManager::waitForSendBuffer(int len, int windowSize, double time
     while (totalQueuedSize() + len > windowSize) {
         pthread_mutex_lock(&mSentMutex);
         if (timeout > 0.0) {
-            struct timespec ts;
-            struct timeval tv;
-            int secs = (int)timeout;
-            uint64_t ns = (timeout - secs) * 1000000000;
-            gettimeofday(&tv, NULL);
-            ts.tv_sec = tv.tv_sec + (int)timeout;
-            ts.tv_nsec = tv.tv_usec * 1000 + ns;
-            if (pthread_cond_timedwait(&mSentCond, &mSentMutex, &ts) == ETIMEDOUT) {
-                DEBUG("%p: timed out waiting for send buffer\n", this);
+            if (timedWait(&mSentCond, &mSentMutex, timeout) == ETIMEDOUT) {
+                DEBUG("%p: timeout waiting for send buffer\n", this);
+                pthread_mutex_unlock(&mSentMutex);
                 return -ETIMEDOUT;
             }
         } else {
--- a/endhost/ssp/SCIONProtocol.cpp
+++ b/endhost/ssp/SCIONProtocol.cpp
@@ -301,14 +301,7 @@ int SSPProtocol::recv(uint8_t *buf, size_t len, SCIONAddr *srcAddr, double timeo
             return -EWOULDBLOCK;
         }
         if (timeout > 0.0) {
-            struct timespec ts;
-            struct timeval tv;
-            int secs = (int)timeout;
-            uint64_t ns = (timeout - secs) * 1000000000;
-            gettimeofday(&tv, NULL);
-            ts.tv_sec = tv.tv_sec + (int)timeout;
-            ts.tv_nsec = tv.tv_usec * 1000 + ns;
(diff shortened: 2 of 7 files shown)


endhost/ssp/ConnectionManager.cpp
```diff
@@ -24,13 +26,47 @@ PathManager::PathManager(int sock, const char *sciond)
...
 }
 
 PathManager::~PathManager()
 {
     close(mDaemonSocket);
     pthread_mutex_destroy(&mPathMutex);
+    pthread_cond_destroy(&mPathCond);
+}
+
+void PathManager::getDefaultIP()
```
pszal, line 48: I though to implement something like that in TCP/SCION stack via a request to sciond. If a _correct_ sciond is known, the result would be more accurate.

aznair, line 48: The reason for not getting the ip via sciond (which is actually what libssock started out doing) is that in the integration test if an AS is used as src in one test and dst in the next, it would need to be stopped and restarted because it needs to switch from having the client ip (127.0.0.2) to the server ip (127.0.0.3).

In the future there should be a SCION version of <code>getifaddrs()</code> which an application can use to ask the kernel what SCION addresses are currently available on which interfaces of the system.

endhost/ssp/ConnectionManager.cpp
```diff
@@ -394,13 +443,28 @@ bool SSPConnectionManager::bufferFull(int window)
...
+        if (timeout > 0.0) {
+            struct timespec ts;
+            struct timeval tv;
+            int secs = (int)timeout;
+            uint64_t ns = (timeout - secs) * 1000000000;
+            gettimeofday(&tv, NULL);
+            ts.tv_sec = tv.tv_sec + (int)timeout;
+            ts.tv_nsec = tv.tv_usec * 1000 + ns;
+            if (pthread_cond_timedwait(&mSentCond, &mSentMutex, &ts) == ETIMEDOUT) {
+                DEBUG("%p: timed out waiting for send buffer\n", this);
```
pszal: why don't unlock `mSentMutex` here?

infrastructure/router/main.py
```diff
@@ -553,7 +553,10 @@ def _process_data(self, spkt, ingress, drop_on_error):
         hof = spkt.path.get_hof()
         if hof.verify_only:
             raise SCMPNonRoutingHOF
-        if spkt.addrs.dst.isd_as == self.addr.isd_as:
+        # FIXME(aznair): Remove second condition once PathCombinator is less
+        # stupid.
+        if (spkt.addrs.dst.isd_as == self.addr.isd_as and
+                spkt.path.is_on_last_segment()):
```
pszal, line 559: why? it's really interesting that it is needed

endhost/ssp/SCIONProtocol.cpp
```diff
@@ -254,39 +276,57 @@ int SSPProtocol::send(uint8_t *buf, size_t len, SCIONAddr *dstAddr)
...
     while (!mReadyToRead) {
-        DEBUG("no data to read yet\n");
+        DEBUG("%p: no data to read yet\n", this);
         if (!mBlocking) {
             pthread_mutex_unlock(&mReadMutex);
             DEBUG("non-blocking socket not ready to recv\n");
             return -EWOULDBLOCK;
         }
-        pthread_cond_wait(&mReadCond, &mReadMutex);
+        if (timeout > 0.0) {
```
pszal, line 303: a lot of common code with `waitForSendBuffer`

infrastructure/router/main.py
aznair, line 559: There's a discussion on this in Reviewable.

endhost/ssp/ConnectionManager.cpp
aznair: good catch :)

PR #808, opened 2016-07-07
Use a capnp object in zk to identify each service instance.

kormat:
This means that adding multiple addresses in the near future will be
easy.

---

This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/808)

### DIFF
--- a/test/lib/zk/zk_test.py
+++ b/test/lib/zk/zk_test.py
@@ -42,7 +42,7 @@ class BaseZookeeper(object):
     """
     Base class for lib.zk.zk.Zookeeper unit tests
     """
-    default_args = ["1-2", "srvtype", "srvid"]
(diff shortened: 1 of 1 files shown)


infrastructure/path_server/core.py
```diff
@@ -61,7 +60,7 @@ def _update_master(self):
         if not curr_master:
             logging.warning("_update_master(): current master is None.")
             return
-        if curr_master != self._master_id:
+        if not self._master_id or curr_master != self._master_id:
```
pszal, line 63: isn't that the same?

kormat, line 63: No, if `_master_id` is None, then the equality check will cause a type error in `Cerealizable.__eq__`.

PR #809, opened 2016-07-08
Fix master detection in cPSes.

kormat:
- The default ZkID comparison method fails on the list of addresses,
  even if the contents are identical, so for now switch to comparing
  string representations which are consistent.
- Use the zookeeper lock to determine if a cPS is already master.
- Make it clearer that only the master cPS propagates to other core
  ASes.
- Add assertions to make clear the expected state in various methods,
  and improved logging.
- Add handling for a possible but rare race-condition in _send_to_master

Also:
- explicitly print success/failure at the end of
  docker/integration_test.sh

---

This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/809)

### DIFF
--- a/infrastructure/path_server/core.py
+++ b/infrastructure/path_server/core.py
@@ -52,28 +52,30 @@ class CorePathServer(PathServer):
         Read master's address from shared lock, and if new master is elected
         sync it with segments.
         """
+        if self.zk.have_lock():
+            self._segs_to_master.clear()
+            return
         try:
             curr_master = self.zk.get_lock_holder()
         except ZkNoConnection:
             logging.warning("_update_master(): ZkNoConnection.")
             return
+        if curr_master and curr_master == self._master_id:
+            return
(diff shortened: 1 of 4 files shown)


infrastructure/path_server/core.py
```diff
@@ -52,36 +52,38 @@ def _update_master(self):
         Read master's address from shared lock, and if new master is elected
         sync it with segments.
         """
+        if self.zk.have_lock():
+            self._segs_to_master.clear()
```
pszal, line 56: why this?

infrastructure/path_server/core.py
```diff
@@ -52,36 +52,38 @@ def _update_master(self):
         Read master's address from shared lock, and if new master is elected
         sync it with segments.
         """
+        if self.zk.have_lock():
```
pszal, line 55: so now, basically that means `is_master` right?

PR #817, opened 2016-07-14
Add standard log library for go, including logging panics.

kormat:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/817)

### DIFF
--- a/go/lib/log/formatter.go
+++ b/go/lib/log/formatter.go
@@ -1,0 +1,44 @@
+// Copyright 2016 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


go/lib/log/formatter.go
```diff
@@ -0,0 +1,44 @@
...
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+const tsFmt = "2006-01-02T15:04:05.999"
+
+type UTCFormatter struct{}
+
+func (u *UTCFormatter) Format(e *log.Entry) ([]byte, error) {
```
ercanucan, line 30: Makes sense to return only []byte? It seems that in its current form this function can only return nil as error.

PR #823, opened 2016-07-18
Modify MPUDP's sendto() to be compatible with UNIX UDP socket API.

ercanucan:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/823)

### DIFF
--- a/endhost/scion_socket.py
+++ b/endhost/scion_socket.py
@@ -334,7 +334,7 @@ class ScionBaseSocket(object):
         """
         return self.sendto(msg)
 
-    def sendto(self, msg, dst=None, port=None):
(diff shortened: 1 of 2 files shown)


endhost/scion_socket.py
```diff
@@ -355,7 +355,8 @@ def sendto(self, msg, dst=None, port=None):
         if msg is None or len(msg) == 0:
             return 0
         ptr = None
-        if dst and port:
+        if dst_port:
+            dst, port = dst_port
             sa = addr_py2c(dst, port)
```
kormat: You don't need to expand this manually. `addr_py2c(*dst_port)` works just fine.

PR #864, opened 2016-08-25
Some additions based on initial attempt at integration

aznair:
- Updated instructions for pcap
- Added missing sample file/code
- Added port_id field to RouterPacket
- Misc cleanup

---

This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/864)


lib/libhsr/hsr_dpdk.c
```diff
@@ -825,13 +841,20 @@ void setup_kni()
         sprintf(cmd, "brctl addif br%d vEth%d", i, i);
```
kormat: Both these commands should be with `sudo`, for now at least.

PR #869, opened 2016-08-29
Libssocket: accept SCMP packets

aznair:
Only handles path errors, rest get ignored for now

---

This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/869)

### DIFF
--- a/endhost/dispatcher.c
+++ b/endhost/dispatcher.c
@@ -900,29 +900,35 @@ void deliver_scmp(uint8_t *buf, SCMPL4Header *scmp, int len, HostAddr *from)
         zlog_error(zc, "SCMP does not support SVC source.\n");
         return;
     }
-    L4Key key;
-    memset(&key, 0, sizeof(key));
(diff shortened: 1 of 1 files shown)


endhost/dispatcher.c
```diff
@@ -909,11 +910,18 @@ void deliver_scmp(uint8_t *buf, SCMPL4Header *scmp, int len, HostAddr *from)
     Entry *e;
     HASH_FIND(hh, udp_port_list, &key, sizeof(key), e);
     if (!e) {
-        zlog_error(zc, "SCMP entry for %s:%d not found\n",
+        zlog_info(zc, "SCMP entry for %s:%d (UDP) not found, try SSP",
```
kormat: I'm not sure i understand this. You should be able to determine the relevant l4 protocol from the scmp payload.

aznair: Good point.

PR #890, opened 2016-09-14
TCP fixes

pszalach:
TCP for infrastructure seems to be pretty stable, although still disabled by default.
- dropping `USE_TCP`s from BS and PS (they interfere)
- handling `connect()` error
- do not send/recv through an inactive socket
- close TCP connection in SIBRA

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/890)

### DIFF
--- a/infrastructure/beacon_server/base.py
+++ b/infrastructure/beacon_server/base.py
@@ -109,7 +109,6 @@ class BeaconServer(SCIONElement, metaclass=ABCMeta):
     ZK_REV_OBJ_MAX_AGE = HASHTREE_EPOCH_TIME
     # Interval to checked for timed out interfaces.
     IF_TIMEOUT_INTERVAL = 1
-    USE_TCP = True
 
     def __init__(self, server_id, conf_dir):
         """
--- a/infrastructure/beacon_server/base.py
+++ b/infrastructure/beacon_server/base.py
@@ -109,7 +109,7 @@ class BeaconServer(SCIONElement, metaclass=ABCMeta):
     ZK_REV_OBJ_MAX_AGE = HASHTREE_EPOCH_TIME
     # Interval to checked for timed out interfaces.
     IF_TIMEOUT_INTERVAL = 1
-    USE_TCP = False
+    USE_TCP = True
 
     def __init__(self, server_id, conf_dir):
         """
--- a/infrastructure/path_server/base.py
+++ b/infrastructure/path_server/base.py
@@ -63,7 +63,6 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
     ZK_SHARE_LIMIT = 10
     # Time to store revocations in zookeeper
     ZK_REV_OBJ_MAX_AGE = HASHTREE_EPOCH_TIME
-    USE_TCP = True
 
     def __init__(self, server_id, conf_dir):
         """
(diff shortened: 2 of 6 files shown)


pszalach:
@kormat's feedback will be addressed in new PR

pszalach:

---

*[infrastructure/scion_elem.py, line 484 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/890#-KRgvotE6GF--PSRMiVJ:-KRhQ_1Pl12cIE-iLVs4:b-34bs18) ([raw file](https://github.com/netsec-ethz/scion/blob/d413b6a82ba37974a68096a36a4a2f795f60c18c/infrastructure/scion_elem.py#L484)):*
<details><summary><i>Previously, kormat (Stephen Shirley) wrote…</i></summary>
> This seems to mask the error from the application, rather than reporting back that the message is being dropped...
</details>
Yes, we could return a bool from send_meta, or connect within a thread (to implement a retry mechanism).

---

*[infrastructure/path_server/base.py, line 270 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/890#-KRgw8rwtw43MR8Qt9GT:-KRhRd3Snn_U_rsm9BlI:bp3meeg) ([raw file](https://github.com/netsec-ethz/scion/blob/d413b6a82ba37974a68096a36a4a2f795f60c18c/infrastructure/path_server/base.py#L270)):*
<details><summary><i>Previously, kormat (Stephen Shirley) wrote…</i></summary>
> What does the FIXME mean?
</details>
TRC, cert, and signature checks

---

*[lib/defines.py, line 142 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/890#-KRgwi1Q6tNBEALGbS67:-KRhRp9eiaNqa6ur5iPX:bh9stfa) ([raw file](https://github.com/netsec-ethz/scion/blob/d413b6a82ba37974a68096a36a4a2f795f60c18c/lib/defines.py#L142)):*
<details><summary><i>Previously, kormat (Stephen Shirley) wrote…</i></summary>
> I can definitely see why we'd have problems with cpu consumption
</details>
This is my local setting, which gives similar performance and lower load average (compared to udp-only). I will test different values in circleci, before switching to TCP.

---

*[lib/socket.py, line 370 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/890#-KRgwrrDsf8EGFQ00qmp:-KRhSIGvE7-3M6w43UNk:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d413b6a82ba37974a68096a36a4a2f795f60c18c/lib/socket.py#L370)):*
<details><summary><i>Previously, kormat (Stephen Shirley) wrote…</i></summary>
> It's slightly better to set the local attribute first, and then use it to make decisions. It means the logic is less likely to get out of sync if things change. I.e.
> ```
> self.active = active
> if self.active:
>   ... set timeout ...
> ```
</details>
Done.

---

*[lib/socket.py, line 408 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/890#-KRgxgqKDQhUo5Motgiw:-KRhSfBMZ6NTd7-dGMkf:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d413b6a82ba37974a68096a36a4a2f795f60c18c/lib/socket.py#L408)):*
<details><summary><i>Previously, kormat (Stephen Shirley) wrote…</i></summary>
> It's easier to handle this case first:
> ```
> if not self.active:
>    logging.debug(...inactive socket)
>    return None...
> try:
>   read = ...
> ...
> ```
> Same applies for `send_msg` and `close`
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/890)*

PR #897, opened 2016-09-15
Finish go router support for traceroute extension.

kormat:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/897)


kormat:
@pszalach @klausman

PR #903, opened 2016-09-16
Fix SCMP/UDP checksum calculation.

kormat:
While the python code includes a zero'd checksum field in the checksum
calculation input, libscion's UDP and SCMP checksum functions did not.
It's unclear why this didn't cause obvious problems with UDP before, but
it completely broke SCMP checksum support in libscion.

Also:
- Change chk_add_chunk to keep an internal index, removing the need for
  specifying the index by the caller.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/903)

### DIFF
--- a/endhost/dispatcher.c
+++ b/endhost/dispatcher.c
@@ -868,8 +868,8 @@ void process_scmp(uint8_t *buf, SCMPL4Header *scmp, int len, HostAddr *from)
 {
     int calc_chk = scmp_checksum(buf);
     if (calc_chk != scmp->checksum) {
-        zlog_error(zc, "SCMP header checksum (%x) doesn't match computed checksum (%x)\n",
-                scmp->checksum, calc_chk);
+        zlog_error(zc, "SCMP header checksum (%x) doesn't match computed checksum (%x)",
+                ntohs(scmp->checksum), ntohs(calc_chk));
         return;
     }
     if (htons(scmp->class_) == SCMP_GENERAL_CLASS && htons(scmp->type) == SCMP_ECHO_REQUEST) {
@@ -885,7 +885,7 @@ void send_scmp_echo_reply(uint8_t *buf, SCMPL4Header *scmp, HostAddr *from)
     reverse_packet(buf);
     scmp->type = htons(SCMP_ECHO_REPLY);
     update_scmp_checksum(buf);
-    zlog_debug(zc, "send echo reply to %s:%d\n", addr_to_str(from->addr, from->addr_type, NULL), ntohs(from->port));
+    zlog_debug(zc, "send echo reply to %s:%d", addr_to_str(from->addr, from->addr_type, NULL), ntohs(from->port));
     send_data(buf, ntohs(sch->total_len), from);
 }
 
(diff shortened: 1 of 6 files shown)


kormat:
@klausman @pszalach @aznair

aznair:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/libscion/checksum.c, line 89 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/903#-KRnUe0yru_LUhlbb22U:-KRnUe0yru_LUhlbb22V:b8podg2) ([raw file](https://github.com/netsec-ethz/scion/blob/2d4b41027d80eccdc21f4ec0e53a45cc287e8fbb/lib/libscion/checksum.c#L89)):*
> ```C
>     if (in->idx >= in->total) {
>         fprintf(stderr, "ERROR chk_add_chunk: in->idx (%d) >= in->total (%d)\n", in->idx, in->total);
>         exit(1);
> ```

When might this happen, and is it really an exit condition?

---

*[lib/libscion/scmp.c, line 30 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/903#-KRnVuUjtePjoS7d_gm5:-KRnVuUkj51tfzXppFIh:bhdi2xo) ([raw file](https://github.com/netsec-ethz/scion/blob/2d4b41027d80eccdc21f4ec0e53a45cc287e8fbb/lib/libscion/scmp.c#L30)):*
> ```C
>     ptr = chk_add_chunk(input, ptr, 6);
>     // Use blank checksum field
>     chk_add_chunk(input, (uint8_t *)(&blank_sum), 2);
> ```

I must be missing something, how does adding a zero field change the sum at all?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/903)*

kormat:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/libscion/checksum.c, line 89 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/903#-KRnUe0yru_LUhlbb22U:-KRnYM1UK_i0pUkzxNa2:b52y5xu) ([raw file](https://github.com/netsec-ethz/scion/blob/2d4b41027d80eccdc21f4ec0e53a45cc287e8fbb/lib/libscion/checksum.c#L89)):*
<details><summary><i>Previously, aznair (Jason Lee) wrote…</i></summary>
> When might this happen, and is it really an exit condition?
</details>
It can only happen if the caller has a bug, where it declares it wants X chunks, but tries to use >X chunks. Without this check, it would be a straight segfault.

---

*[lib/libscion/scmp.c, line 30 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/903#-KRnVuUjtePjoS7d_gm5:-KRnYWaw4G03gF-W_3Vq:bbf3q5l) ([raw file](https://github.com/netsec-ethz/scion/blob/2d4b41027d80eccdc21f4ec0e53a45cc287e8fbb/lib/libscion/scmp.c#L30)):*
<details><summary><i>Previously, aznair (Jason Lee) wrote…</i></summary>
> I must be missing something, how does adding a zero field change the sum at all?
</details>
Yeah it's subtle alright - the thing is that the checksum algorithm effectively takes an array of bytes, but it sums them in 2-byte chunks. 2 zeros in the input can be split across chunks (if the any of the preceeding chunks have an odd length - like the protocol field), and thereby change the outcome.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/903)*

klausman:
:lgtm:

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/903#-:-KRnaHCBnP8JM6VFUZeg:bnfp4nl)*

PR #922, opened 2016-10-26
Pin all go dependencies to specific commit IDs.

kormat:
This prevents any compromises of upstream code from affecting us. The
downside is it requires careful maintenance of dependencies and
revisions.

This change removes the previous go dependency manager we were using
(https://github.com/rancher/trash) with govendor
(https://github.com/kardianos/govendor), as the former doesn't fulfill
all our requirements.

Also:
- Add a .keepme to logs/ so that it's already present in a fresh clone.
- Move go-related stuff from the top-level Makefile into go/
- Install `realpath` in circleci, as go/Makefile uses it, and it's not
  installed by default on ubuntu 14.04.
- Add target dependencies so that things like running go tests (or
  coverage) have the environement properly initalized.
- go/deps.txt and tools/godeps.py aren't needed any more, and are
  removed.

This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/922)


kormat:
@klausman

klausman:
Aside from some quoting an comments, :lgtm:

---

Reviewed 13 of 13 files at r1.

---

*[deps.sh, line 79 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/922#-KV0JDf5p89bkgaDmfXS:-KV0JDf5p89bkgaDmfXT:b-deixnu) ([raw file](https://github.com/netsec-ethz/scion/blob/e59c634dabb9bf6d3760ca1b43105bd7b7291865/deps.sh#L79)):*
> ```Shell
>             PROJECT=govendor
>             COMMIT=120a6099270fc9360236f4383430e2adda6181cc
>             GOPATH_BASE=${GOPATH%%:*}
> ```

Quote RHS, as it may contain spaces. Same for mkdir, cd and possibly git URL below.

---

*[Makefile, line 11 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/922#-KV0J_YnPbNn2lMRuA27:-KV0J_YnPbNn2lMRuA28:b-awqkek) ([raw file](https://github.com/netsec-ethz/scion/blob/e59c634dabb9bf6d3760ca1b43105bd7b7291865/Makefile#L11)):*
> ```Makefile
> 	if type -P go >/dev/null; then cd go && make clean; fi
> 
> go: libscion
> ```

Add a comment on why we're not using make -C here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/922#-:-KV0JtW2NqqjMHZG9_Qe:bjhoigt)*

kormat:
@perrig - as discussed on monday.

PR #927, opened 2016-10-28
Deprecate libssocket.

kormat:
libssocket (./endhost/ssp) has major threading/locking issues. These
prevent it from working at all on the new HP DL20 servers we're starting
to buy. Our best guess is that this is because the servers have
non-hyperthreaded CPUs, and this is somehow triggering otherwise latent
bugs in the thread handling.

Fixing this would involve a major amount of engineering time (and
probably a rewrite). So, this change stops building/testing libssocket
by default. It can still be used as a proof of concept, but is
definitely unsuitable for any kind of production traffic for the time
being.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/927)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ gohsr: libhsr
 	cd go && make hsr
 
 # Order is important
-clibs: libscion libfilter libssocket liblwip libtcpmw
+clibs: libscion libfilter liblwip libtcpmw
 
 libscion:
 	$(MAKE) -C lib/libscion install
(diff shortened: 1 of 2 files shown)


kormat:
@klausman 
FYI @ercanucan

klausman:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/927#-:-KVAH8n4xtfpeIOll2vX:bnfp4nl)*

PR #928, opened 2016-10-31
Use `golang` package instead of `golang-1.6`

kormat:
On ubuntu 16.04, the default golang version is 1.6, and installing the
`golang` package adds a symlink in `/usr/bin`, so there's no need to add
`/usr/lib/go-1.6/bin` to `$PATH` any more.

Also:
- Improve the golang version detection to allow golang >=1.6
- Generate an error if govendor detects any "outside" packages (i.e.
  third-party packages that are outside go/vendor/, or missing).

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/928)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ clean:
 	if type -P go >/dev/null; then cd go && make clean; fi
 
 go: libscion
-	# `make -C go` breaks if there are symlinks in $PWD
+	@# `make -C go` breaks if there are symlinks in $PWD
 	cd go && make
 
 gohsr: libhsr
--- a/README.md
+++ b/README.md
@@ -29,12 +29,11 @@ Necessary steps in order to run SCION:
 
 1. Make sure that you have a
    [Go workspace](https://golang.org/doc/code.html#GOPATH) setup, and that
-   `~/.local/bin`, `/usr/lib/go-1.6/bin` and `$GOPATH/bin` can be found in your
-   `$PATH` variable. For example:
+   `~/.local/bin`, and `$GOPATH/bin` can be found in your `$PATH` variable. For example:
 
     ```
     echo 'export GOPATH="$HOME/go"' >> ~/.profile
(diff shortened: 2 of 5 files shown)


kormat:
@klausman

klausman:
One question for clarification, otherwise :lgtm:

---

Reviewed 5 of 5 files at r1.

---

*[deps.sh, line 72 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/928#-KVQwBMYw5t7sVVapGvh:-KVQwBMZyWk263HPnX0z:b-t5mtpp) ([raw file](https://github.com/netsec-ethz/scion/blob/7146fc43fd7ab9d06fac68e284444908963eb0ae/deps.sh#L72)):*
> ```Shell
>         sudo DEBIAN_FRONTEND=noninteractive apt-get install $APTARGS --no-install-recommends golang git
>     fi
>     if ! go version | grep -vq ' go1\.[0-5]\.'; then
> ```

Are there versions of go 0.x that e need to worry about?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/928#-:-KVQwZUfet90dLJHsBmm:b-vrdio2)*

kormat:

---

*[deps.sh, line 72 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/928#-KVQwBMYw5t7sVVapGvh:-KVUCBHe-Mfww1L-Yak_:b-fs1gs3) ([raw file](https://github.com/netsec-ethz/scion/blob/7146fc43fd7ab9d06fac68e284444908963eb0ae/deps.sh#L72)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary>
> Are there versions of go 0.x that e need to worry about?
</details>
Good question, but the answer is "basically, no". Go 1.0 was released in 2012, and 1.7 is the current version.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/928)*

Issue #918, opened 2016-10-18
TCP pipe mode with callbacks


pszalach:
For efficiency reasons, Stephen suggested to use callbacks with the pipe mode.
The design is as follows:
https://gist.github.com/pszalach/dd9c337e65141d6047ff9b880b812afa

@kormat check are we on the same page
@klausman

klausman:
Notes after a quick pass over the gist:

- using select limits us to at most 1024 FDs (actually, no FDs with an ID > 1023, see the BUGS section of the select(2)); we should use poll(2)
- Locking is not explicitly mentioned, but required for this to work multithreaded.

pszalach:
Thanks @klausman,
1) I don't see that in my man page (2015-07-23)
2) yes, I left locking for an actual PR

kormat:
http://stackoverflow.com/a/3951845 goes into some detail why `poll` is better than `select`, http://stackoverflow.com/a/2033785 has some more. Select is limited to FD_SETSIZE fds, which is 1024 on linux.

kormat:
`Interacting with a state is thread-safe.` is definitely untrue. Both sides will be modifying `conn_ready`, the lwip side will be modifying `conn`.

Line 61: `if (state.fd...)` - `0` is a valid fd. It should instead be set to/checked against -1.
Line 63: `if (state.tcp_buf || state.conn_ready)` - is that meant to be `state.tcp_buf_len`?
Line 77: `if (!state.tcp_buf_len)` - that's checking the wrong buffer.

If `all_conn_states` is an array, what does `remove()` do?
If `struct conn_state` had an `fd_ready`-type field, then `send_to_tcp` could check and read from both sources in sequence (but recursion would need to be adjusted).

pszalach:
re `poll/select`:
I'm fine with `poll` (just my man page does not have the mentioned bug)

> `Interacting with a state is thread-safe.` is definitely untrue.

This is an assumption for this doc. Of course locking is necessary for that.

> Line 61...

Yes, I know that, but thought that my intention will be understood correctly here ;-)

> Line 63...
> Line 77...

Good catches!

> If `all_conn_states` is an array, what does `remove()` do?

I'm not sure whether connection states will be kept in an array or other data structure, but `remove(state)` removes `state` from this data structure.

> If `struct conn_state` had an `fd_ready`-type field..

Sure, can be done like that... and the goal is performance?

kormat:
> > If all_conn_states is an array, what does remove() do?
[...]

PR #938, opened 2016-11-23
Testing CircleCI with an implementation of #918 

pszalach:
Do not review, testing only.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/938)


pszalach:
This implements #918, however, I'm closing this PR as: a) I couldn't find any benefit over the current implementation (actually, it takes much more CPU), b) due to the faster router, the current approach started passing the CircleCI integration tests.

PR #943, opened 2016-11-30
Fix packet recycling with libhsr in the go router.

kormat:
readHSRInput was changed to keep a static set of packet buffers for
re-use recently, but the packets were still recycled, meaning they were
available for other parts of the router to reuse. This leads to an
interesting and wide variety of strange errors. This change resets the
packet buffers instead, which clears the contents.

Also:
- Fix a ~cosmetic with how hsr.AddrMs is updated that made debugging
  much harder than it needed to be. The old version takes a /copy/ of
  each element of AddrMs, and modifies the copy, leaving the original
  untouched. The new version fixes that.
- Remove the VSS ulimit, as it causes the hsr library to die fatally on
  startup without much useful information.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/943)

### DIFF
--- a/go/border/hsr/hsr.go
+++ b/go/border/hsr/hsr.go
@@ -93,9 +93,9 @@ func Init(zlog_cfg string, args []string, addrMs []AddrMeta) *common.Error {
 	AddrMs = addrMs
 	// Calculate the C address structure from the Go address structure.
 	cAddrs := make([]C.saddr_storage, len(AddrMs))
-	for i, addrM := range AddrMs {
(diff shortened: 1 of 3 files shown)


kormat:
@klausman

klausman:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/943#-:-KXq2jo6oV0ePvli4cMq:bnfp4nl)*

PR #944, opened 2016-12-01
Disable TCP for now.

kormat:
The infrastructure's use of TCP is fundamentally broken, as TCP
connections are made in the middle of the main processing threads. This
means that if any TCP connection attempt doesn't immediately succeed,
the thread is blocked. E.g. beacon servers trying to register downpaths
can block the entire main worker loop.

Also, TCP spews vast amount of noise into the logs, making it very
difficult to debug anything.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/944)

### DIFF
--- a/infrastructure/scion_elem.py
+++ b/infrastructure/scion_elem.py
@@ -107,7 +107,7 @@ class SCIONElement(object):
     """
     SERVICE_TYPE = None
     STARTUP_QUIET_PERIOD = STARTUP_QUIET_PERIOD
-    USE_TCP = True
+    USE_TCP = False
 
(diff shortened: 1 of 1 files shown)


klausman:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/944#-:-KXtUydRQmTr2gV3Gf3E:bnfp4nl)*

pszalach:
@kormat yes, I agree, `connect()` should be executed within a thread. :lgtm:

pszalach:
@kormat consequently, you should disable ZK support, which also timeouts in the main thread.

PR #955, opened 2017-01-04
Peer revocation

shitz:
Implemented correct handling of peer interface revocation.

So far, path segments were removed if they contained at least one revoked interface - regardless of whether this was an up/down hop or a peering interface. This was not only inefficient, but would also lead to problems when a local path server removed its only up segment due to a failed peering interface.

This PR implements peer interface revocations in the following way:
1) Beacon servers stop including revoked peering interfaces in newly announced beacons. This was already the case.
2) Path servers do not remove path segments when receiving a revocation for a peering interface. Instead, they include the revocation for peering interfaces in the path segment when they respond to a path segment.
3) The path combinator has been adjusted to check for revocations in the path segment when creating peering paths and ignores (correctly) revoked interfaces.

The revocation information is added as a PCB extension.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/955)


kormat:
My comments will seem a bit confused, as the code is not making sense to me in places.

---

Reviewed 13 of 15 files at r1.
Review status: 13 of 15 files reviewed at latest revision, 13 unresolved discussions.

---

*[docker/integration_test.sh, line 67 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_ilPBMGx38-tQPxFkf:-K_ilPBMGx38-tQPxFkg:b-53aicq) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/docker/integration_test.sh#L67)):*
> ```Shell
> 
> run Revocation "test/integration/revocation_test.sh\
>  ${REV_BRS:-as1-11:br1-11-3 as2-26:br2-26-2 as1-14:br1-14-3 as1-16:br1-16-2}"
> ```

As discussed on slack, this approach is actually a dead-end, as the BR names are non-deterministic, and in the not-too-distant future there will be multiple interfaces per router making this not work at all.

---

*[infrastructure/beacon_server/base.py, line 746 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jDNEsJEtD6JJoyO6D:-K_jDNEtnoH3TGyxdM36:b-lvz4hm) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/beacon_server/base.py#L746)):*
> ```Python
>         return to_remove
> 
>     def _verify_revocation_for_asm(self, rev_info, as_marking):
> ```

Unless i missed something else, this code is identical to that now in `sciond`, except this version iterates over all PCBMs, correct? If so it can be combined pretty simply in `scion_elem` as before, by adding an extra param to say whether to check all PCBMS or not. E.g.

```
for pcbm in as_marking.iter_pcbms():
    if rev_info.p.ifID in [pcbm.hof().ingress_if, pcbm.hof().egress_if]:
        return True
    if not check_all:
        break
```

---

*[infrastructure/beacon_server/base.py, line 760 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jCudPit0hqihuKtjx:-K_jCudPit0hqihuKtjy:b3ui2tt) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/beacon_server/base.py#L760)):*
> ```Python
>             return False
>         pcbm = as_marking.pcbm(0)
>         if rev_info.p.ifID in [pcbm.hof().ingress_if, pcbm.hof().egress_if]:
> ```

It occurs to me that if the interface ID is ever 0, then it will revoke all terminated segments. That needs to be guarded against carefully.

---

*[infrastructure/path_server/base.py, line 333 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jDzFY56H548StRQtm:-K_jDzFY56H548StRQtn:b8xt9r0) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/path_server/base.py#L333)):*
> ```Python
> 
>         :param seg: The PathSegment object.
>         :return: False, if the path segment contains a revoked hop interface.
> ```

This should clarify that it's only the upstream & downstream interfaces that are being checked for (right?).

---

*[infrastructure/path_server/core.py, line 258 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jK4DT4edFfxWXvoed:-K_jK4DT4edFfxWXvoee:bjlil8h) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/path_server/core.py#L258)):*
> ```Python
>             return False
> 
>         self._add_peer_revs(down_segs)
> ```

Why here? Shouldn't this already be added when the revocation is processed?

---

*[lib/packet/pcb.py, line 180 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jFl0lXwSM_2OWL1HV:-K_jFl0lXwSM_2OWL1HW:bohzr5l) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/pcb.py#L180)):*
> ```Python
>         self._calc_min_exp()
>         self.sibra_ext = None
>         self.rev_ext = None
> ```

This is never set...

---

*[lib/packet/pcb.py, line 247 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jFBjmGASw1oi6nhMG:-K_jFBjmGASw1oi6nhMH:b-um739z) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/pcb.py#L247)):*
> ```Python
>     def add_rev_ext(self, ext_p):  # pragma: no cover
>         self.p.exts.rev = ext_p.copy()
>         self.rev_infos = RevPCBExt(self.p.exts.rev)
> ```

This overwrites any existing revocation extensions, which is a bit surprising when the method name starts with `add`. Is this intended/desired?

---

*[lib/packet/pcb.py, line 339 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jFbjNraT-81Pa19uf:-K_jFbjNraT-81Pa19ug:b-6gfplk) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/pcb.py#L339)):*
> ```Python
>         """
>         result = {}
>         if self.rev_ext:
> ```

This is checking the wrong variable, it should be `self.p.exts.rev`. Same for the next line.

---

*[lib/packet/path_mgmt/rev_info.py, line 90 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jLtExchY2MRB7LbP0:-K_jLtExchY2MRB7LbP1:baorjhl) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/path_mgmt/rev_info.py#L90)):*
> ```Python
> 
> 
> class RevPCBExt(Cerealizable):  # pragma: no cover
> ```

Why does this exist? It contains no data, it's just a list of revocation infos.

---

*[lib/packet/path_mgmt/rev_info.py, line 100 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jGbWXFzSlUxMr5FHX:-K_jGbWXFzSlUxMr5FHY:b2m6kji) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/path_mgmt/rev_info.py#L100)):*
> ```Python
>     def __init__(self, p):
>         super().__init__(p)
>         self.rev_infos = p.revInfos
> ```

What is the point of this? Ah, it's also not used in this class.

---

*[lib/packet/path_mgmt/rev_info.py, line 106 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jGjQjeIX34bt5TRER:-K_jGjQjeIX34bt5TRES:bac1ae4) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/path_mgmt/rev_info.py#L106)):*
> ```Python
>         p.init("revInfos", len(rev_infos))
>         for i, rev_info in rev_infos:
> ```

Either `rev_infos` isn't a straight list of revocation info objects, or you've forgotten `enumerate`.

---

*[tools/pp.py, line 2 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jJeWCqqSuM-F_H-8n:-K_jJeWCqqSuM-F_H-8o:b-dr9lab) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/tools/pp.py#L2)):*
> ```Python
> #!/usr/bin/python3
> ```

Add a comment saying what this file is for. E.g. `Pretty-print a hex-encoded SCION packet`.

---

*[tools/pp.py, line 6 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jI7Hjy-3XsVkdd1J7:-K_jI7Hjy-3XsVkdd1J8:b-6nc0ew) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/tools/pp.py#L6)):*
> [4:]
Please don't hard-code this :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-:-K_jM8ag1G78XsA8AZ9e:b-4vfg80)*

kormat:
Review status: 13 of 15 files reviewed at latest revision, 15 unresolved discussions.

---

*[lib/packet/path.py, line 505 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jSIfes8nJ6RlGIR5C:-K_jSIff2NDtm9NjOLdK:b78p4tq) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/path.py#L505)):*
> ```Python
>         up_iof.peer = down_iof.peer = True
>         paths = []
>         for uph, dph, pm in cls._find_peer_hfs(up_segment.asm(up_index),
> ```

Move first arg to next line.

---

*[lib/packet/path.py, line 679 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jVxWUNJBscNMrfjdr:-K_jVxWUNJBscNMrfjds:b-n79id6) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/path.py#L679)):*
> ```Python
>                     # interface.
>                     up_rev = up_peer_rev_map.get((up_ia, up_peer.p.inIF),
>                                                  None)
> ```

The default return value for `get` is already `None`. Also this line fits within 80 chars :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955)*

shitz:
Yeah sorry, the code had some artifacts from multiple iterations and also quite some bugs that I found...

---

Review status: 2 of 13 files reviewed at latest revision, 15 unresolved discussions.

---

*[docker/integration_test.sh, line 67 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_ilPBMGx38-tQPxFkf:-K_n4OOb8VqSnLOCBWYc:bnv097s) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/docker/integration_test.sh#L67)):*

As discussed on slack, this approach is actually a dead-end, as the BR names are non-deterministic, and in the not-too-distant future there will be multiple interfaces per router making this not work at all.
</blockquote></details>

Will take care of this in a separate PR. For now, as long as no one changes the generator, we should be fine.

---

*[infrastructure/beacon_server/base.py, line 746 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jDNEsJEtD6JJoyO6D:-K_n6BUuafZlOd23P0mA:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/beacon_server/base.py#L746)):*

Unless i missed something else, this code is identical to that now in `sciond`, except this version iterates over all PCBMs, correct? If so it can be combined pretty simply in `scion_elem` as before, by adding an extra param to say whether to check all PCBMS or not. E.g.

```
for pcbm in as_marking.iter_pcbms():
    if rev_info.p.ifID in [pcbm.hof().ingress_if, pcbm.hof().egress_if]:
        return True
    if not check_all:
        break
```
</blockquote></details>

Done.

---

*[infrastructure/beacon_server/base.py, line 760 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jCudPit0hqihuKtjx:-K_n70v5fu4BMcUL8rux:b6zrrkh) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/beacon_server/base.py#L760)):*

It occurs to me that if the interface ID is ever 0, then it will revoke all terminated segments. That needs to be guarded against carefully.
</blockquote></details>

Good catch.

---

*[infrastructure/path_server/base.py, line 333 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jDzFY56H548StRQtm:-K_n7NU4CVGXJzrNJSYP:b-pjb8uw) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/path_server/base.py#L333)):*

This should clarify that it's only the upstream & downstream interfaces that are being checked for (right?).
</blockquote></details>

Was implied by '...contains a revoked hop interface...`. Made it more obvious.

---

*[infrastructure/path_server/core.py, line 258 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jK4DT4edFfxWXvoed:-K_n7etW4s7OzPt_81ri:b-2e9hzm) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/path_server/core.py#L258)):*

Why here? Shouldn't this already be added when the revocation is processed?
</blockquote></details>

I don't think so, since revocations are much shorter lived that path segments. If you add revocations to PCBs when it is processed, then you'd need to check here whether the revocation is still valid. I think adding it when the PCB makes more sense from a program flow point of view. What do you think?

---

*[lib/packet/path.py, line 505 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jSIfes8nJ6RlGIR5C:-K_n8_I1i-m4nTLcrTDY:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/path.py#L505)):*

Move first arg to next line.
</blockquote></details>

Done.

---

*[lib/packet/path.py, line 679 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jVxWUNJBscNMrfjdr:-K_n8k1zm5G0fbCIy0nG:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/path.py#L679)):*

The default return value for `get` is already `None`. Also this line fits within 80 chars :P
</blockquote></details>

Done.

---

*[lib/packet/pcb.py, line 180 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jFl0lXwSM_2OWL1HV:-K_nFIVnyTwgu6_-EDNi:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/pcb.py#L180)):*

This is never set...
</blockquote></details>

Done.

---

*[lib/packet/pcb.py, line 247 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jFBjmGASw1oi6nhMG:-K_nFOin1i9PqsPUv5r9:b-lpdyx2) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/pcb.py#L247)):*

This overwrites any existing revocation extensions, which is a bit surprising when the method name starts with `add`. Is this intended/desired?
</blockquote></details>

Changed to appending.

---

*[lib/packet/pcb.py, line 339 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jFbjNraT-81Pa19uf:-K_nGAvzafU4efH8pIyJ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/pcb.py#L339)):*

This is checking the wrong variable, it should be `self.p.exts.rev`. Same for the next line.
</blockquote></details>

Done.

---

*[lib/packet/path_mgmt/rev_info.py, line 90 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jLtExchY2MRB7LbP0:-K_nGJEV8DUkD2mDxANl:b-nfv707) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/path_mgmt/rev_info.py#L90)):*

Why does this exist? It contains no data, it's just a list of revocation infos.
</blockquote></details>

Changed to just using a list of revocation infos.

---

*[lib/packet/path_mgmt/rev_info.py, line 100 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jGbWXFzSlUxMr5FHX:-K_nHAL1Oy6q3vwSAUPZ:bss4a2b) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/path_mgmt/rev_info.py#L100)):*

What is the point of this? Ah, it's also not used in this class.
</blockquote></details>

Removed.

---

*[lib/packet/path_mgmt/rev_info.py, line 106 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jGjQjeIX34bt5TRER:-K_nH5VKcjZdFdtkWCJ5:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/lib/packet/path_mgmt/rev_info.py#L106)):*

Either `rev_infos` isn't a straight list of revocation info objects, or you've forgotten `enumerate`.
</blockquote></details>

Done.

---

*[tools/pp.py, line 2 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jJeWCqqSuM-F_H-8n:-K_nHOebpmkqLxPWyZ_z:bdv89gc) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/tools/pp.py#L2)):*

Add a comment saying what this file is for. E.g. `Pretty-print a hex-encoded SCION packet`.
</blockquote></details>

Not applicable anymore.

---

*[tools/pp.py, line 6 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jI7Hjy-3XsVkdd1J7:-K_nHCcxFv40im0TXqCg:bkpb6bl) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/tools/pp.py#L6)):*

> [4:]
Please don't hard-code this :P
</blockquote></details>

This got accidentally added. Reverted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-:-K_nHQQrdVro0I1Ta26N:bi9r1ky)*

kormat:
Reviewed 1 of 15 files at r1, 10 of 11 files at r2.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[infrastructure/scion_elem.py, line 735 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2G4hcVwYdvnj047RU:-Ka2G4hcVwYdvnj047RV:byz0mgw) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/infrastructure/scion_elem.py#L735)):*
> ```Python
>         if rev_info.p.ifID == 0:
>             logging.warning("Received revocation for ifID 0. Ignoring.\n%s" %
>                             rev_info.short_desc())
> ```

I think this is too late. The revocation should be dropped immediately on receiving if the ifid is 0.

---

*[infrastructure/path_server/base.py, line 333 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jDzFY56H548StRQtm:-Ka1rPhoHG3edELnCTDO:b-rpjrrm) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/path_server/base.py#L333)):*

Was implied by '...contains a revoked hop interface...`. Made it more obvious.
</blockquote></details>

I'm not familiar with `hop interface` implying not-peer. This looks good now.

---

*[infrastructure/path_server/base.py, line 166 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka1rAHdw2-WolEqmH8O:-Ka1rAHdw2-WolEqmH8P:b-pt3eu6) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/infrastructure/path_server/base.py#L166)):*
> ```Python
>         Add if revocation token to segment ID mappings.
>         """
>         def add_one_mapping(key):
> ```

Can be replaced with a single line: `set.htroot_if2set.setdefault(key, set()).add(key)`.

---

*[infrastructure/path_server/core.py, line 258 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jK4DT4edFfxWXvoed:-Ka2En0Nsm9HLysVHf0S:b-hpb178) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/path_server/core.py#L258)):*

I don't think so, since revocations are much shorter lived that path segments. If you add revocations to PCBs when it is processed, then you'd need to check here whether the revocation is still valid. I think adding it when the PCB makes more sense from a program flow point of view. What do you think?
</blockquote></details>

The downside to doing it JIT, as it were, is that you have to do this multiple times for every path request you get, which can get expensive. Hmm. One thing that would solve this is having a cache of PCBs with current revocations applied, which is expired every epoch. When a new revocation comes in, apply it to the cache. If a request comes in for one of the cached PCBs, just reply with it. If a request doesn't have a cached reply,  grab the PCB from the path store, apply current revs, cache it, and send the response. This should be very efficient in all cases.

---

*[lib/packet/path.py, line 682 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2Gh20M8_CgKBBm2FA:-Ka2Gh20M8_CgKBBm2FB:bwbxbfo) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/lib/packet/path.py#L682)):*
> ```Python
>                     if (cls._skip_peer(up_rev, up_asm.p.hashTreeRoot) or
>                             cls._skip_peer(down_rev, down_asm.p.hashTreeRoot)):
>                         logging.debug("Not using peer %s:%d <-> %s:%d due to"
> ```

Formatting.

---

*[lib/packet/pcb.py, line 190 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2HtJd6cpRWzl1EqEf:-Ka2HtJe2W-ZIZ5gYtju:bd2x0qz) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/lib/packet/pcb.py#L190)):*
> ```Python
>             p.exts.sibra = sibra_ext.p
>         if rev_infos:
>             p.exts.revInfos = [info.pack() for info in rev_infos]
> ```

This isn't legal, `p.exts.revInfos` hasn't been `init`'d.

---

*[proto/pcb.capnp, line 36 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2H7YMdxzFhTBihCdY:-Ka2H7YMdxzFhTBihCdZ:b-2m4xi1) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/proto/pcb.capnp#L36)):*
> ```
>     exts :group {
>         sibra @3 :Sibra.SibraPCBExt;
>         revInfos @4 :List(Data);
> ```

Why `Data`, and not just `RevInfo`?

---

*[test/lib/packet/path_test.py, line 949 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2JXV_VU6wqNGJ2QCc:-Ka2JXV_VU6wqNGJ2QCd:bwbxbfo) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/test/lib/packet/path_test.py#L949)):*
> ```Python
>         up_asm = create_mock_full({"isd_as()": "1-1", "iter_pcbms()": up_pcbms,
>                                    "p": p})
>         down_asm = create_mock_full({"isd_as()": "2-1",
> ```

Formatting.

---

*[test/lib/packet/path_test.py, line 968 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2K6bN5VkLpnCpUAbL:-Ka2K6bN5VkLpnCpUAbM:bwbxbfo) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/test/lib/packet/path_test.py#L968)):*
> ```Python
>         # Tests
>         ntools.eq_(peers, [(up_pcbms[0].hof(), down_pcbms[0].hof(), 500)])
>         skip_peer.assert_has_calls([call(None, b"1234"),
> ```

Formatting.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955)*

shitz:

---

*[infrastructure/scion_elem.py, line 735 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2G4hcVwYdvnj047RU:-Ka39H2YA5ey5CSkiYVp:bs7wumb) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/infrastructure/scion_elem.py#L735)):*

I think this is too late. The revocation should be dropped immediately on receiving if the ifid is 0.
</blockquote></details>

Done.

---

*[infrastructure/path_server/base.py, line 166 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka1rAHdw2-WolEqmH8O:-Ka3BwmrGAtGbKfPMWQ_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/infrastructure/path_server/base.py#L166)):*

Can be replaced with a single line: `set.htroot_if2set.setdefault(key, set()).add(key)`.
</blockquote></details>

Done.

---

*[lib/packet/path.py, line 682 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2Gh20M8_CgKBBm2FA:-Ka3JJTi1Wx89fNofFDE:br0bztg) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/lib/packet/path.py#L682)):*

Formatting.
</blockquote></details>

Better now?

---

*[lib/packet/pcb.py, line 190 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2HtJd6cpRWzl1EqEf:-Ka3LwHBk1RkefOeeQRi:b52tu1h) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/lib/packet/pcb.py#L190)):*

This isn't legal, `p.exts.revInfos` hasn't been `init`'d.
</blockquote></details>

How do I init this? Does `p.exts.init("revInfos", len(rev_infos))` work?

---

*[proto/pcb.capnp, line 36 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2H7YMdxzFhTBihCdY:-Ka3MB2VR285CSIIiUby:b-rqgru6) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/proto/pcb.capnp#L36)):*

Why `Data`, and not just `RevInfo`?
</blockquote></details>

The reasoning is to parse it only on demand I guess. Should I changed to `RevInfo`?

---

*[test/lib/packet/path_test.py, line 949 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2JXV_VU6wqNGJ2QCc:-Ka3MWwn4_qrc-VqiRmF:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/test/lib/packet/path_test.py#L949)):*

Formatting.
</blockquote></details>

Done.

---

*[test/lib/packet/path_test.py, line 968 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2K6bN5VkLpnCpUAbL:-Ka3MU1n0z5vyDv2GlKq:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/test/lib/packet/path_test.py#L968)):*

Formatting.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955)*

shitz:
Review status: 5 of 13 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[infrastructure/path_server/core.py, line 258 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-K_jK4DT4edFfxWXvoed:-KaBd-oEJAbIY-jsdUGZ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9848c8e8d22682c23dc97b7a796007561afc6c9f/infrastructure/path_server/core.py#L258)):*

The downside to doing it JIT, as it were, is that you have to do this multiple times for every path request you get, which can get expensive. Hmm. One thing that would solve this is having a cache of PCBs with current revocations applied, which is expired every epoch. When a new revocation comes in, apply it to the cache. If a request comes in for one of the cached PCBs, just reply with it. If a request doesn't have a cached reply,  grab the PCB from the path store, apply current revs, cache it, and send the response. This should be very efficient in all cases.
</blockquote></details>

Done.

---

*[proto/pcb.capnp, line 36 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2H7YMdxzFhTBihCdY:-KaBd3ZBUMLyGnyG6WOa:bu2j8oa) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/proto/pcb.capnp#L36)):*

The reasoning is to parse it only on demand I guess. Should I changed to `RevInfo`?
</blockquote></details>

Changed it to `RevInfo` now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955)*

kormat:
Reviewed 8 of 8 files at r3, 1 of 1 files at r4.

---

*[infrastructure/beacon_server/base.py, line 669 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-KaITUrOxuUUCHuOqL1g:-KaITUrOxuUUCHuOqL1h:b99wyxh) ([raw file](https://github.com/netsec-ethz/scion/blob/7a99213dbdc748faacd529b136f5c83ff2eadc8a/infrastructure/beacon_server/base.py#L669)):*
> ```Python
> 
>     def _handle_revocation(self, rev_info, meta):
>         logging.info("Received revocation via UDP:\n%s", rev_info.short_desc())
> ```

Could also be TCP.

---

*[infrastructure/path_server/base.py, line 227 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-KaIUc9ddPwCJH-CPJUh:-KaIUc9ddPwCJH-CPJUi:b-4q34wc) ([raw file](https://github.com/netsec-ethz/scion/blob/7a99213dbdc748faacd529b136f5c83ff2eadc8a/infrastructure/path_server/base.py#L227)):*
> ```Python
>         self._remove_revoked_segments(rev_info)
>         # Update revocations for PCBs in the the PCB cache.
>         for segment in self.pcb_cache.values():
> ```

It looks like the cache'd PCBs are being accessed by multiple threads at once, while being modified. E.g. the worker thread and the request-handling thread.

---

*[infrastructure/path_server/base.py, line 403 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-KaNDNuL_JO_B6xg37c4:-KaNDNuL_JO_B6xg37c5:b-7rt1pf) ([raw file](https://github.com/netsec-ethz/scion/blob/c151e860d81a7db1541f7595f8b1f72dfe1a3e14/infrastructure/path_server/base.py#L403)):*
> ```Python
>             for rev_info in list(self.revocations):
>                 if rev_info.p.epoch != current_epoch:
>                     self.revocations.pop(rev_info)
> ```

This will delete any revocations that have been grace'd in from the previous epoch - is this desired?

---

*[lib/packet/pcb.py, line 190 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2HtJd6cpRWzl1EqEf:-KaMr7mERndJcCI0n-ZX:bhejbv0) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/lib/packet/pcb.py#L190)):*

How do I init this? Does `p.exts.init("revInfos", len(rev_infos))` work?
</blockquote></details>

I think it might be `p.init("exts.revInfos"...)` but i'm not 100% sure. If it's wrong, you'll get a method-not-found error.

---

*[lib/packet/pcb.py, line 191 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-KaMrXgzqUhQScm8l0yf:-KaMrXgzqUhQScm8l0yg:byk1u7b) ([raw file](https://github.com/netsec-ethz/scion/blob/7a99213dbdc748faacd529b136f5c83ff2eadc8a/lib/packet/pcb.py#L191)):*
> ```Python
>         if rev_infos:
>             p.exts.init("revInfos", len(rev_infos))
>             p.exts.revInfos = rev_infos
> ```

This also isn't legal, i believe. You need to copy each entry over individually.

---

*[lib/packet/pcb.py, line 252 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-KaNJMjWpS8TgaDBz3Yn:-KaNJMjWpS8TgaDBz3Yo:b-htvkv3) ([raw file](https://github.com/netsec-ethz/scion/blob/c151e860d81a7db1541f7595f8b1f72dfe1a3e14/lib/packet/pcb.py#L252)):*
> ```Python
>         existing = []
>         idxs_to_ignore = set()
>         for i in range(len(self.p.exts.revInfos)):
> ```

This seems more complex than necessary. Just drop revocations from previous epochs?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955)*

shitz:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---

*[infrastructure/beacon_server/base.py, line 669 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-KaITUrOxuUUCHuOqL1g:-KaNjtvKdQiHApvcOCvM:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7a99213dbdc748faacd529b136f5c83ff2eadc8a/infrastructure/beacon_server/base.py#L669)):*

Could also be TCP.
</blockquote></details>

Done.

---

*[infrastructure/path_server/base.py, line 227 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-KaIUc9ddPwCJH-CPJUh:-KaNl2EM9n0k9vwpekpM:b5xmra6) ([raw file](https://github.com/netsec-ethz/scion/blob/7a99213dbdc748faacd529b136f5c83ff2eadc8a/infrastructure/path_server/base.py#L227)):*

It looks like the cache'd PCBs are being accessed by multiple threads at once, while being modified. E.g. the worker thread and the request-handling thread.
</blockquote></details>

good point. fixed.

---

*[infrastructure/path_server/base.py, line 403 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-KaNDNuL_JO_B6xg37c4:-KaNm9vPj5G9p3nmcvic:b-54ak3u) ([raw file](https://github.com/netsec-ethz/scion/blob/c151e860d81a7db1541f7595f8b1f72dfe1a3e14/infrastructure/path_server/base.py#L403)):*

This will delete any revocations that have been grace'd in from the previous epoch - is this desired?
</blockquote></details>

I think the ones that are >= `current_epoch` should be included and others discarded, since they are only valid because of the grace period. Changed it that way.

---

*[lib/packet/pcb.py, line 190 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Ka2HtJd6cpRWzl1EqEf:-KaNmR6qYpMbwNZD_-8f:bepzpkx) ([raw file](https://github.com/netsec-ethz/scion/blob/fb4d0eb96a6332c92645269be9359fe866309983/lib/packet/pcb.py#L190)):*

I think it might be `p.init("exts.revInfos"...)` but i'm not 100% sure. If it's wrong, you'll get a method-not-found error.
</blockquote></details>

My version seems to work. I'm doing it in `PCB.add_rev_infos` and I didn't get any exceptions from that.

---

*[lib/packet/pcb.py, line 191 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-KaMrXgzqUhQScm8l0yf:-KaNmvUVkRb3nF_D7OzX:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7a99213dbdc748faacd529b136f5c83ff2eadc8a/lib/packet/pcb.py#L191)):*

This also isn't legal, i believe. You need to copy each entry over individually.
</blockquote></details>

Done.

---

*[lib/packet/pcb.py, line 252 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-KaNJMjWpS8TgaDBz3Yn:-KaRhAx43t2T3RAyJi76:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/c151e860d81a7db1541f7595f8b1f72dfe1a3e14/lib/packet/pcb.py#L252)):*

This seems more complex than necessary. Just drop revocations from previous epochs?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955)*

kormat:
Reviewed 3 of 3 files at r5.

---

*[lib/packet/pcb.py, line 261 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Kag7T6ZokR9vc89PYF5:-Kag7T6ZokR9vc89PYF6:b62nvdd) ([raw file](https://github.com/netsec-ethz/scion/blob/19144df9e55e741140a219bfd1014584f38c158d/lib/packet/pcb.py#L261)):*
> ```Python
>             if info_p.epoch >= current_epoch:
>                 existing[(info_p.isdas, info_p.ifID)] = orphan
>         # Remove revocations for which we already have a newer one.
> ```

This is a no-op, unless i'm missing something. The previous step will already have removed any older revocations.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955)*

shitz:

---

*[lib/packet/pcb.py, line 261 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Kag7T6ZokR9vc89PYF5:-Kag8bHRz9ABlo0Yo-5l:b-z8sbf) ([raw file](https://github.com/netsec-ethz/scion/blob/19144df9e55e741140a219bfd1014584f38c158d/lib/packet/pcb.py#L261)):*

This is a no-op, unless i'm missing something. The previous step will already have removed any older revocations.
</blockquote></details>

The previous step only removes revocations that were already in the PCB that are older than the current epoch. This step removes all revocations from the NEW revs that are either too old or already contained in the PCB.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955)*

kormat:

---

*[lib/packet/pcb.py, line 261 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Kag7T6ZokR9vc89PYF5:-KahImG65VU-rA8IpSgN:blfb6a4) ([raw file](https://github.com/netsec-ethz/scion/blob/19144df9e55e741140a219bfd1014584f38c158d/lib/packet/pcb.py#L261)):*

The previous step only removes revocations that were already in the PCB that are older than the current epoch. This step removes all revocations from the NEW revs that are either too old or already contained in the PCB.
</blockquote></details>

Ahh ok. I would prefer an explicit for loop. Otherwise LGTM.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955)*

shitz:
Review status: 12 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[lib/packet/pcb.py, line 261 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/955#-Kag7T6ZokR9vc89PYF5:-KahL9fGdMrGe4GPoO4e:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/19144df9e55e741140a219bfd1014584f38c158d/lib/packet/pcb.py#L261)):*

Ahh ok. I would prefer an explicit for loop. Otherwise LGTM.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/955)*

PR #968, opened 2017-02-13
Path Service

jpcsmith:
@klausman I created a different directory for the changes. The ScionSocket and SspSocket files are currently nothing substantial or final. Let me know if i should exclude them from the pull request.

The major parts of this commit include:
-  Basic PathService for interfacing with the SCION daemon and storing the retrieved paths.
- Document the PathPolicy (mostly the same as in the SSP).
- Utils and C code for parsing scion daemon responses.

Otherwise
- Some comments on the code in the SSP directory (no changes to the code were intended).
- Added missing includes in the base scion lib.
- The Mutex and MutexScion files are copied mostly verbatim

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/968)


klausman:
I like what you've been doing so far, despite the many picky comments I just added :)

One general thing:

I'm unsure if your intention was to always include doxygen comments (e.g. `set_timeout` doesn't have it at all). I don't mind only having human-readable comments, but we should be consistent. Alternatively, we could use doxygen comments for only the public parts of the API.

---

Reviewed 28 of 28 files at r1.
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[endhost/protocols/include/PathService.h, line 59 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcvsI8nFK_5jJ3figug:-KcvsI8nFK_5jJ3figuh:b-ch4enx) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L59)):*
> ```objc
>  *
>  * TODO(jsmith): Paths when returned are returned by copy to avoid fine-grained
>  * locking.
> ```

Can you clarify this?  Are they currently copy-returned and shouldn't be? Why is copy-return not wanted in this case?

---

*[endhost/protocols/include/PathService.h, line 73 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcvxICaph9N6OPWPYMs:-KcvxICaph9N6OPWPYMt:b-tlfft2) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L73)):*
> ```objc
>    *                   error.
>    */
>   static std::unique_ptr<PathService> create(uint32_t dest_isd_as,
> ```

The comment above does not mention `dest_isd_as`

---

*[endhost/protocols/include/PathService.h, line 97 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcvy4WMV-dWYop2Fclc:-Kcvy4WMV-dWYop2Fcld:b-gsi34d) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L97)):*
> ```objc
> 
>   // Verify that the record complies to the policy and inserts it
>   // int add_record();
> ```

Mark this as a `TODO`

---

*[endhost/protocols/include/PathService.h, line 112 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw0xBGxxXbY-hakm7r:-Kcw0xBGxxXbY-hakm7s:b-gjjr67) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L112)):*
> ```objc
>    *
>    * Requires that the buffer be able to accommodate at least the dispatcher
>    * header.
> ```

Can you add how long the header is? (at least the minimum, if variable)

---

*[endhost/protocols/include/PathService.h, line 114 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw2CL9nIHJhtHRSjVk:-Kcw2CL9nIHJhtHRSjVl:b-ke0io4) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L114)):*
> ```objc
>    * header.
>    */
>   int lookup_paths(uint32_t isd_as, uint8_t* buffer, int buffer_len)
> ```

Is there a reason the `buffer` is a a `uint8_t*` instead of a pointer type?

---

*[endhost/protocols/include/PathService.h, line 121 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw2OU3Wj9dG-auKrAl:-Kcw2OU3Wj9dG-auKrAm:b-b7hj3c) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L121)):*
> ```objc
>    *
>    * An existing record with the same interfaces is replaced by the new record,
>    * while maintinging the record's key.
> ```

maintaining*

---

*[endhost/protocols/include/PathService.h, line 124 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw2ZVGjOa5ziwPe0-C:-Kcw2ZVGjOa5ziwPe0-D:buaq1ar) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L124)):*
> ```objc
>    *
>    * Returns the identifier of the record where the insertion or replacement
>    * took place, or -1 if the insertion would excede the number of allowed
> ```

exceed*

Since `m_records` is a `map` and we're not using its order explicitly (I think), the term "addition" rather than an "insertion" would be more apropos, IMO.

---

*[endhost/protocols/include/PathService.h, line 135 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw2xeKbdOsKRYEfPo9:-Kcw2xeKbdOsKRYEfPoA:ba7sjv0) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L135)):*
> ```objc
> 
> 
>   // Removes non-conformant records
> ```

Can you expand what "non-conformant" means?

Also, the name of this method is a bit ambiguous: it sounds a bit like it clears the whole list. Maybe something along the lines of `cleanup_records` or somesuch? 

---

*[endhost/protocols/include/sciondlib.h, line 76 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw4kaXlmFdK_J0v_DI:-Kcw4kaY50gstVHIUDXa:brbn0eq) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/sciondlib.h#L76)):*
> ```objc
>  * On success, returns the number of bytes used for the host address. If the
>  * remaining input in the buffer, as indicated by data_len, is less than the
>  * needed data, zero is returned.
> ```

Add a comment regarding initialization of the buffer (I presume the caller needs to zero it before calling this, but it's better to be explicit).

---

*[endhost/protocols/include/UnixSocket.h, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw6b2qDN_xbRkywz5Z:-Kcw6b2qDN_xbRkywz5_:b-m5aa14) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/UnixSocket.h#L35)):*
> ```objc
>   virtual int send_all(uint8_t *buf, int len);
> 	virtual int setsockopt(int level, int optname, const void *optval,
> 												 socklen_t optlen);
> ```

Replace tabs with spaces.

---

*[endhost/protocols/include/util.h, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw743C6EzOGtEEk8gA:-Kcw743C6EzOGtEEk8gB:b-kkw3lf) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/util.h#L34)):*
> ```objc
> 
> 
> /* Blocks `timeout` milliseconds until data is available to be read then recvs
> ```

Should this be "Blocks at most `timeout` milliseconds..."?

---

*[endhost/protocols/include/util.h, line 43 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw7IgF8LvvEBvyFRHJ:-Kcw7IgF8LvvEBvyFRHK:bkfrv4o) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/util.h#L43)):*
> ```objc
>  *
>  * On success, a sockfd file descriptor is returned. On error, a negative
>  * system error code is returned, namely those relating to socket and connect.
> ```

In general it is good to write syscall and function call names with empty parens, i.e., "socket() and connect()", it makes it easier to get what it referred to.

---

*[endhost/protocols/src/PathService.cpp, line 132 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwCCd5rAeZBGWa76v_:-KcwCCd5rAeZBGWa76va:b-ihj0cx) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/PathService.cpp#L132)):*
> ```cpp
>   int bytes_used = 0, offset = 0;
>   m_records_mutex.Lock();
>   do {
> ```

Add a comment at the top of this loop explaining the two exit criteria on line 155

---

*[endhost/protocols/src/PathService.cpp, line 178 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwEnDdvB09hM65FbXk:-KcwEnDdvB09hM65FbXl:b-csi56) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/PathService.cpp#L178)):*
> ```cpp
>     updated = true;
>   } else {
>     // Flag that it would be an insertion
> ```

If my comment on `PathService.h:124` is right (`map` implying addition rather than insertion), change these comments from "insert" to "addition"

---

*[endhost/protocols/src/sciondlib.c, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwK_XYhiypc3P4oLYH:-KcwK_XZSgBpVJ5H0NXk:b645rtj) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/sciondlib.c#L87)):*
> ```C
>   offset += addr_len;
> 
>   // Copy the 2-byte port and change its edianness
> ```

Shouldn't this use `ntoh*()`? If not, add comment on why does the endianess need to change. There are more instances below.

---

*[endhost/protocols/src/sciondlib.c, line 116 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwM2C2xmGLvual2t4C:-KcwM2C2xmGLvual2t4D:b-my0j36) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/sciondlib.c#L116)):*
> ```C
>   // Now allocate the buffer space and set the path length
>   path_ptr->len = path_byte_len / LINE_LEN;
>   path_ptr->raw_path = malloc(path_byte_len);
> ```

Check if `malloc()` returned NULL? Not sure how to handle that situation.

---

*[endhost/protocols/src/sciondlib.c, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwLB48MfemQMaTryVc:-KcwLB48MfemQMaTryVd:b-q3msry) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/sciondlib.c#L128)):*
> ```C
> // Returns number of bytes parsed and assumes that the data in the buffer is
> // sufficient.
> int parse_interface(uint8_t* buffer, sinterface_t* interface)
> ```

Why can the caller not be expected to know the size of `buffer` (and tell us)? Then we could a t least make a quick check if it is long enough.
Or is the assumption that the caller makes these checks?

---

*[endhost/protocols/src/sciondlib.c, line 159 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwLor2Eyg51mp5BDFl:-KcwLor2Eyg51mp5BDFm:b-my0j36) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/sciondlib.c#L159)):*
> ```C
>   *interface_count = buffer[offset-1];
>   sinterface_t* interface_ptr =
>     malloc(sizeof(sinterface_t) * (*interface_count));
> ```

Check if `malloc()` returned NULL? Not sure how to handle that situation.

---

*[endhost/protocols/src/SspSocket.cpp, line 41 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw_-WbMqoeMbixVXBW:-Kcw_-Wcp-Z4lAxmaPLE:b-qli3xz) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/SspSocket.cpp#L41)):*
> ```cpp
> //
> // Src
> //
> ```

What is the purpose of this comment?

---

*[endhost/protocols/src/SspSocket.cpp, line 44 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw_RuK6ImWsNVxFtUC:-Kcw_RuK6ImWsNVxFtUD:b-3u3afb) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/SspSocket.cpp#L44)):*
> ```cpp
> 
> // TODO(jsmith): Check address validity
> int SspSocket::bind(const SCIONAddr& sockaddr)
> ```

The naming of this method is a bit odd, since it does none of the things the POSIX `bind()` does.

---

*[endhost/protocols/test/PathService-test.cpp, line 43 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcwbe0B7WX69w5jx-RQ:-Kcwbe0B7WX69w5jx-RR:b-m3rwxi) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/test/PathService-test.cpp#L43)):*
> ```cpp
> 
> 
> // Add a record byte vector the the out vector with the provided A path_lines
> ```

This sentence makes no sense

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-:-KcvwrkgIlD8hIwyONs2:bcomytm)*

jpcsmith:
I ended up falling back to doxygen style comments to be explicit with out or in,out parameters. If you dont think it's necessary i'll just writing that distinction in the description for the variable.

---

Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[endhost/protocols/include/PathService.h, line 59 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcvsI8nFK_5jJ3figug:-Kcwg-ElM2vmg_U83OTu:b-h2cnoi) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L59)):*

Can you clarify this?  Are they currently copy-returned and shouldn't be? Why is copy-return not wanted in this case?
</blockquote></details>

I haven't implemented the API for retrieving a path by it's associated integer id yet. But when I do it will be by copy or the caller will need to lock on the record to prevent it from being deleted while in use. Alternatively a shared pointer could be used, but I was concerned this may create complexities if/when converted to C.

---

*[endhost/protocols/include/PathService.h, line 73 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcvxICaph9N6OPWPYMs:-KcwsvYXV9kEEbFBov5B:b-rx4sv9) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L73)):*

The comment above does not mention `dest_isd_as`
</blockquote></details>

I was only commenting on ambiguous parameters. Should I do for all?

---

*[endhost/protocols/include/PathService.h, line 135 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw2xeKbdOsKRYEfPo9:-KcxDOHEo7vO2_BScvOM:b27yhml) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L135)):*

Can you expand what "non-conformant" means?

Also, the name of this method is a bit ambiguous: it sounds a bit like it clears the whole list. Maybe something along the lines of `cleanup_records` or somesuch? 
</blockquote></details>

"Reduce the extent of (something) by removing superfluous or unwanted parts". Cleanup to me sounds as if it deallocates the memory for the records. Since it currently only verifies the policy. How about something such as "ensure_policy"?

---

*[endhost/protocols/include/sciondlib.h, line 76 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw4kaXlmFdK_J0v_DI:-KcxEWeoXZ85X2uO_Y2G:bwaz66u) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/sciondlib.h#L76)):*

Add a comment regarding initialization of the buffer (I presume the caller needs to zero it before calling this, but it's better to be explicit).
</blockquote></details>

I don't follow, the buffer is the input argument.

---

*[endhost/protocols/include/UnixSocket.h, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw6b2qDN_xbRkywz5Z:-KcxFCjBIe-sIQ8Dm09H:btu26qg) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/UnixSocket.h#L35)):*

Replace tabs with spaces.
</blockquote></details>

I'll need to figure out how those got in there...

---

*[endhost/protocols/src/sciondlib.c, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwK_XYhiypc3P4oLYH:-KcxITKXdm83ZOQyR8uG:b-tytj13) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/sciondlib.c#L87)):*

Shouldn't this use `ntoh*()`? If not, add comment on why does the endianess need to change. There are more instances below.
</blockquote></details>

I believe ntoh*() is not necessary, as the "result of the computation" is stored in host byte order i.e. the computation and it's result ignores endianess.

---

*[endhost/protocols/src/sciondlib.c, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwLB48MfemQMaTryVc:-KcxLShmuk4BGGIOHMLb:b-on41f7) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/sciondlib.c#L128)):*

Why can the caller not be expected to know the size of `buffer` (and tell us)? Then we could a t least make a quick check if it is long enough.
Or is the assumption that the caller makes these checks?
</blockquote></details>

I'd defined it like that as it wasn't visible in the header file and the caller checks, but I'll change it to be more consistent with the rest.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-:-KcweYRzFRtkKG__mQjI:b18bchp)*

jpcsmith:
Review status: 21 of 28 files reviewed at latest revision, 21 unresolved discussions.

---

*[endhost/protocols/include/PathService.h, line 114 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw2CL9nIHJhtHRSjVk:-KcxPL4PO071g71e4spt:b-57j6h5) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L114)):*

Is there a reason the `buffer` is a a `uint8_t*` instead of a pointer type?
</blockquote></details>

Changed to vector.

---

*[endhost/protocols/include/PathService.h, line 121 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw2OU3Wj9dG-auKrAl:-KcxPNDCMyAv_na8UeA5:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L121)):*

maintaining*
</blockquote></details>

Done.

---

*[endhost/protocols/include/PathService.h, line 124 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw2ZVGjOa5ziwPe0-C:-KcxPObqdKF7652KaEJ-:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L124)):*

exceed*

Since `m_records` is a `map` and we're not using its order explicitly (I think), the term "addition" rather than an "insertion" would be more apropos, IMO.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/968)*

jpcsmith:
Review status: 20 of 28 files reviewed at latest revision, 21 unresolved discussions.

---

*[endhost/protocols/include/PathService.h, line 97 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcvy4WMV-dWYop2Fclc:-KcxRElkG04le2WDKoav:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L97)):*

Mark this as a `TODO`
</blockquote></details>

Done.

---

*[endhost/protocols/include/PathService.h, line 112 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw0xBGxxXbY-hakm7r:-KcxRH0K0YN39g_LcpFz:b-57j6h5) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L112)):*

Can you add how long the header is? (at least the minimum, if variable)
</blockquote></details>

Changed to vector.

---

*[endhost/protocols/include/util.h, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw743C6EzOGtEEk8gA:-KcxPurHcbom0hZCC_2s:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/util.h#L34)):*

Should this be "Blocks at most `timeout` milliseconds..."?
</blockquote></details>

Done.

---

*[endhost/protocols/include/util.h, line 43 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw7IgF8LvvEBvyFRHJ:-KcxQ3HAS4v5LvCQpm7-:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/util.h#L43)):*

In general it is good to write syscall and function call names with empty parens, i.e., "socket() and connect()", it makes it easier to get what it referred to.
</blockquote></details>

Done.

---

*[endhost/protocols/src/PathService.cpp, line 132 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwCCd5rAeZBGWa76v_:-KcxQ5ZYywgpSiBzIIA9:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/PathService.cpp#L132)):*

Add a comment at the top of this loop explaining the two exit criteria on line 155
</blockquote></details>

Done.

---

*[endhost/protocols/src/PathService.cpp, line 178 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwEnDdvB09hM65FbXk:-KcxQ7NGO6dgma5H_2C_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/PathService.cpp#L178)):*

If my comment on `PathService.h:124` is right (`map` implying addition rather than insertion), change these comments from "insert" to "addition"
</blockquote></details>

Done.

---

*[endhost/protocols/src/SspSocket.cpp, line 41 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw_-WbMqoeMbixVXBW:-KcxScptLoLcEhxP2pqM:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/SspSocket.cpp#L41)):*

What is the purpose of this comment?
</blockquote></details>

Done.

---

*[endhost/protocols/src/SspSocket.cpp, line 44 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw_RuK6ImWsNVxFtUC:-KcxST0pNn-IB8rkQvHT:b-x6f07g) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/SspSocket.cpp#L44)):*

The naming of this method is a bit odd, since it does none of the things the POSIX `bind()` does.
</blockquote></details>

Im uncertain what this should do beyond setting the local address?

---

*[endhost/protocols/test/PathService-test.cpp, line 43 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcwbe0B7WX69w5jx-RQ:-KcxR9aI1T9ZLlaiGRoc:b-bjpj7a) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/test/PathService-test.cpp#L43)):*

This sentence makes no sense
</blockquote></details>

Better?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/968)*

jpcsmith:
Review status: 20 of 28 files reviewed at latest revision, 21 unresolved discussions.

---

*[endhost/protocols/src/sciondlib.c, line 116 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwM2C2xmGLvual2t4C:-Kd-l1phg7la_0B4dg2D:b-nk5e2d) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/sciondlib.c#L116)):*

Check if `malloc()` returned NULL? Not sure how to handle that situation.
</blockquote></details>

Results <= 0 are now failure to account for the no-memory error.

---

*[endhost/protocols/src/sciondlib.c, line 159 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcwLor2Eyg51mp5BDFl:-Kd-ltQEJIRnR4byLF3r:bribihm) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/sciondlib.c#L159)):*

Check if `malloc()` returned NULL? Not sure how to handle that situation.
</blockquote></details>

As above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/968)*

klausman:
:lgtm:

---

Reviewed 8 of 8 files at r2.

---

*[endhost/protocols/include/PathService.h, line 73 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-KcvxICaph9N6OPWPYMs:-Kd1UedZsto7SB6lxEEe:b-kx7klw) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L73)):*
<details><summary><i>Previously, jpcsmith (Jean-Pierre Smith) wrote…</i></summary><blockquote>

I was only commenting on ambiguous parameters. Should I do for all?
</blockquote></details>

The problem with "ambiguous" is that it is ambiguous --- what is clear as day for you and me right now may be bewildering for someone else --- including both of us in six months. So, yes, either all or none.

---

*[endhost/protocols/include/PathService.h, line 135 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw2xeKbdOsKRYEfPo9:-Kd1V8JrRm_eRY3hjUjB:b-wgrmuf) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/PathService.h#L135)):*
<details><summary><i>Previously, jpcsmith (Jean-Pierre Smith) wrote…</i></summary><blockquote>

"Reduce the extent of (something) by removing superfluous or unwanted parts". Cleanup to me sounds as if it deallocates the memory for the records. Since it currently only verifies the policy. How about something such as "ensure_policy"?
</blockquote></details>

After mulling this over some more, I think you can leave it as prune for now.

---

*[endhost/protocols/include/sciondlib.h, line 76 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw4kaXlmFdK_J0v_DI:-Kd1VFnm_Zn-LLxzVS-7:b-sfmtxs) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/include/sciondlib.h#L76)):*
<details><summary><i>Previously, jpcsmith (Jean-Pierre Smith) wrote…</i></summary><blockquote>

I don't follow, the buffer is the input argument.
</blockquote></details>

Oh, I had a thinko there. 

---

*[endhost/protocols/src/SspSocket.cpp, line 44 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-Kcw_RuK6ImWsNVxFtUC:-Kd1VlEGr3rrZzJy4Kp8:by9wray) ([raw file](https://github.com/netsec-ethz/scion/blob/abb1a2d57a1800d2e7e45ac5a53356f6f1db7c23/endhost/protocols/src/SspSocket.cpp#L44)):*
<details><summary><i>Previously, jpcsmith (Jean-Pierre Smith) wrote…</i></summary><blockquote>

Im uncertain what this should do beyond setting the local address?
</blockquote></details>

As discussed, keep it as is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/968#-:-Kd1W3PMGeg699eSrnqW:bnfp4nl)*

PR #972, opened 2017-02-21
Add libcapnp-dev to packages list

kormat:
This greatly reduces the amount of work that the pycapnp installation
has to do:

With 500M of ram, it takes 3.5mins for pycapnp to install without the
headers, and 35s with the headers. With less than 500M of ram, pycapnp
install will fail without the headers, but with the headers it can
succeed with 275M of ram (although it takes 5.5mins).

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/972)

### DIFF
--- a/pkgs_debian.txt
+++ b/pkgs_debian.txt
@@ -9,6 +9,7 @@ file
 flake8
 g++
 gcc
+libcapnp-dev
 libcurl4-openssl-dev
 libffi-dev
 libjs-jquery-hotkeys


klausman:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/972#-:-KdUrn-6jhDfI4fgoqpg:bnfp4nl)*

PR #975, opened 2017-02-21
Fixed PCB cache handling

shitz:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/975)

### DIFF
--- a/infrastructure/path_server/base.py
+++ b/infrastructure/path_server/base.py
@@ -48,6 +48,48 @@ from lib.zk.id import ZkID
 from lib.zk.zk import Zookeeper
 
 
+class PCBCache:
+    """
+    Short lived path segment cache. Caches segments that contain revocations
+    for peer interfaces. This class is thread-safe.
+    """
+
+    def __init__(self, capacity=100):
+        self._cache = ExpiringDict(capacity, HASHTREE_EPOCH_TIME)
+        self._lock = Lock()
+
+    def get(self, key, default=None):
+        with self._lock:
+            return self._cache.get(key, default)
+
+    def add(self, key, pcb):
+        with self._lock:
+            self._cache[key] = pcb
+
+    def invalidate_entry(self, key):
+        """Removes an entry if it exists."""
+        with self._lock:
+            if key in self._cache:
+                del self._cache[key]
+                return True
+            return False
+
+    def update_with_rev(self, rev_info):
+        """
+        Updates the segments in the pcb_cache with a new revocation, if the
+        revocation is for a peering interface in that segment.
+        """
+        with self._lock:
+            for segment in self._cache.values():
+                for asm in segment.iter_asms():
+                    if asm.isd_as() != rev_info.isd_as():
+                        continue
+                    for pcbm in asm.iter_pcbms(start=1):
+                        hof = pcbm.hof()
+                        if rev_info.p.ifID in [hof.ingress_if, hof.egress_if]:
+                            segment.add_rev_infos([rev_info.copy()])
+
+
 class PathServer(SCIONElement, metaclass=ABCMeta):
     """
     The SCION Path Server.
@@ -78,8 +120,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
         self.waiting_targets = defaultdict(list)
         self.revocations = ExpiringDict(1000, HASHTREE_EPOCH_TIME)
         # Contains PCBs that include revocations.
-        self.pcb_cache = ExpiringDict(100, HASHTREE_EPOCH_TIME)
-        self.pcb_cache_lock = Lock()
+        self.pcb_cache = PCBCache()
         # A mapping from (hash tree root of AS, IFID) to segments
         self.htroot_if2seg = ExpiringDict(1000, HASHTREE_TTL)
         self.htroot_if2seglock = Lock()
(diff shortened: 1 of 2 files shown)


kormat:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[infrastructure/path_server/base.py, line 51 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/975#-Kdeo_ZF0ryVt2MRSUzp:-Kdeo_ZF0ryVt2MRSUzq:b8yj2ft) ([raw file](https://github.com/netsec-ethz/scion/blob/6e583fd9cb801052a62588348d16a62b45c55dea/infrastructure/path_server/base.py#L51)):*
> ```Python
> 
> 
> class PCBCache:
> ```

I suggest moving this to a separate file (e.g. `infrastructure/path_server/pcb_cache.py`), as there's already too much code in the main python files (base/main/local).

---

*[infrastructure/path_server/base.py, line 90 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/975#-Kdeo1WPCLbQLdKQ7EUL:-Kdeo1WPCLbQLdKQ7EUM:b-18efcq) ([raw file](https://github.com/netsec-ethz/scion/blob/6e583fd9cb801052a62588348d16a62b45c55dea/infrastructure/path_server/base.py#L90)):*
> ```Python
>                         hof = pcbm.hof()
>                         if rev_info.p.ifID in [hof.ingress_if, hof.egress_if]:
>                             segment.add_rev_infos([rev_info.copy()])
> ```

Once this happens you can skip to the next segment. Simplest way would be to move the `for asm in segment...` loop to a separate private method, and just `return` once a revocation has been added.

---

*[infrastructure/path_server/base.py, line 239 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/975#-KdepsMiw11wWjzlaZ6u:-KdepsMiw11wWjzlaZ6v:b-306i6v) ([raw file](https://github.com/netsec-ethz/scion/blob/6e583fd9cb801052a62588348d16a62b45c55dea/infrastructure/path_server/base.py#L239)):*
> ```Python
>             self._add_rev_mappings(pcb)
>             logging.debug("%s-Segment updated: %s", name, pcb.short_desc())
>         self.pcb_cache.invalidate_entry(pcb.get_hops_hash())
> ```

I'm confused. This looks like if an updated PCB arrives, the revocation is "forgotten"? The caching needs a full description of what it does and how it is used in the PR description.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/975)*

shitz:
The `pcb_cache` temporarily caches PCBs to which peer revocations have been added to save the effort of adding revocations to PCBs in case of multiple requests for the same segment within the lifetime of a revocation. It works as follows:

1) Whenever a PCB that contains revoked peer interfaces is handed out, the corresponding revocations are appended to the PCB and the PCB is then added to the `pcb_cache` (for at most 10s) to reuse if for further path requests.
2) An incoming peer revocation is applied to all PCBs in the `pcb_cache` that contain a matching interface.
3) If the path server receives a newer version of a PCB in the `pcb_cache`, the copy in the `pcb_cache` is removed from the `pcb_cache`. 

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[infrastructure/path_server/base.py, line 51 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/975#-Kdeo_ZF0ryVt2MRSUzp:-Kdf-WR5Ny3Dy4EsVxxl:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/6e583fd9cb801052a62588348d16a62b45c55dea/infrastructure/path_server/base.py#L51)):*

I suggest moving this to a separate file (e.g. `infrastructure/path_server/pcb_cache.py`), as there's already too much code in the main python files (base/main/local).
</blockquote></details>

Done.

---

*[infrastructure/path_server/base.py, line 90 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/975#-Kdeo1WPCLbQLdKQ7EUL:-Kdf03plxUVBkG0hB4X1:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/6e583fd9cb801052a62588348d16a62b45c55dea/infrastructure/path_server/base.py#L90)):*

Once this happens you can skip to the next segment. Simplest way would be to move the `for asm in segment...` loop to a separate private method, and just `return` once a revocation has been added.
</blockquote></details>

Done.

---

*[infrastructure/path_server/base.py, line 239 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/975#-KdepsMiw11wWjzlaZ6u:-Kdf0DFH4OKVLiHoiSfH:b-u9l3wr) ([raw file](https://github.com/netsec-ethz/scion/blob/6e583fd9cb801052a62588348d16a62b45c55dea/infrastructure/path_server/base.py#L239)):*

I'm confused. This looks like if an updated PCB arrives, the revocation is "forgotten"? The caching needs a full description of what it does and how it is used in the PR description.
</blockquote></details>

The revocation is still around in `revocations`, it's just that the outdated path segment is removed from the `pcb_cache`. Should the revocation still be valid when the updated segment is handed out to a requester, it will be added again to this segment and the segment will then be put into the `pcb_cache` again. Alternatively, the segment in the cache could be replaced with the updated segment and the revocation(s) could be readded to the updated segment in the cache. Imho, this unnecessarily complicates the `pcb_cache`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/975#-:-Kdf1C9FBJC44Gdxh9Ab:b-h3jo4c)*

shitz:
Closing this one. Will change peer revocation handling as discussed offline.

PR #996, opened 2017-03-01
Fix the `context` dependency error after allowing go 1.7.

kormat:
With govendor no longer blacklisting the go17 build tag, it pulls in
vendor/golang.org/x/net/context/go17.go, which depends on that "context"
package that is supplied by go 1.7. This file is not built under go 1.6
due to the build tag, but govendor is not aware of this, and hence under
1.6 it lists "context" as a missing dependency. This change specifically
excludes just this case.

("golang.org/x/net/context" was moved into the standard library as
"context" in go 1.7, hence this problem).

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/996)

### DIFF
--- a/go/Makefile
+++ b/go/Makefile
@@ -38,7 +38,7 @@ deps: vendor/.deps.stamp
 vendor/.deps.stamp: vendor/vendor.json
 	@echo "$$(date -Iseconds) Syncing deps"; govendor sync -v
 	@echo "$$(date -Iseconds) Installing deps"; go install ./vendor/...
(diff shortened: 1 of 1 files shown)


klausman:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/996#-:-Ke9X8CfZe-69RDOxV7t:bnfp4nl)*

Issue #986, opened 2017-02-28
sciond API


kormat:
The current sciond API is hand-coded, and very restricted. We need a new API based on capnp to allow greater freedom of development.

Things to support:
- path requests
- revocation notification
- querying the local ISD-AS id

shitz:
Gist to track the current proposal of the SCIOND API:
https://gist.github.com/shitz/89e613d16039206ba802dd2b32b6cb52

pszalach:
we also may consider adding `exp_time`/`ttl` to `SCIONPath`

kormat:
@shitz - a redesign was discussed ages ago in #408, which has some things that aren't (yet) covered by this design.

shitz:
Ok, so from #408 I think it makes sense to add the following:
1) Including the original request in the response (this could also be a random ID per request)
2) Adding a parameter k to the request
3) Adding error codes?

I think for the current iteration, the name lookup part is not needed, right?

kormat:
#1006 is also merged now. I think that's all the development, now we just need to start removing the deprecated use of a sciond object from within python.

shitz:
Not yet. The `GetBRs` and `GetSVCs` calls need to be implemented, such that we can remove all the `sd.get_first_hop` calls.

kormat:
Missing:
- blacklisting and whitelisting for both ISDs and ASes.

shitz:
I think for the purpose of the pre-production milestone, the SCIOND API is done and this can be closed. We can always open up new issues for feature requests.

PR #1002, opened 2017-03-03
First steps towards the new SCIOND API

shitz:
This PR defines the new SCIOND API (issue #986) and implements the plumbing for path requests and replies to and from SCIOND.
To that end, I introduced a new entity `FwdPathMeta` that contains a forwarding path (currently only a `SCIONPath` object is supported), a list of interfaces of the path and the MTU. The interfaces and MTU have been moved out of `SCIONPath`. `FwdPathMeta` is exclusively used by SCIOND and applications, while `SCIONPath` is the forwarding path contained in SCION packets. This change necessitated the move of `PathCombinator` into a separate file (which was overdue anyway).

Still missing features:
* Extend the API with a `GetFirstHop`-type call
* Implement the remaining SCIOND API calls
* Rename `SCIONPath` to `FwdPath`
* Make the revocation integration test use the new `flush` bit to fix an issue when to first hop border router is down and cannot respond with a revocation

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1002)


kormat:
Reviewed 6 of 18 files at r1.
Review status: 6 of 19 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[lib/path_combinator.py, line 29 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJ5uq9TOaMXjWFEddo:-KeJ5uq9TOaMXjWFEddp:bkevj1e) ([raw file](https://github.com/netsec-ethz/scion/blob/e72b88a5d49e6ccaefc96bfe3edcd1b16e4db4f8/lib/path_combinator.py#L29)):*
> ```Python
> 
> 
> class PathCombinator(object):
> ```

Now that it's in its own module, there's no need to have an artificial class for this.

---

*[lib/path_combinator.py, line 424 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJ6CohUqteYsnhKKY5:-KeJ6CohUqteYsnhKKY6:b4llk9e) ([raw file](https://github.com/netsec-ethz/scion/blob/e72b88a5d49e6ccaefc96bfe3edcd1b16e4db4f8/lib/path_combinator.py#L424)):*
> ```Python
> 
> 
> def valid_mtu(mtu):  # pragma: no cover
> ```

These two aren't used outside this module, so prepend them with `_`.

---

*[lib/packet/path_mgmt/seg_req.py, line 39 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJ9uxVHLAmo0j8PCSr:-KeJ9uxVHLAmo0j8PCSs:bocb35j) ([raw file](https://github.com/netsec-ethz/scion/blob/e72b88a5d49e6ccaefc96bfe3edcd1b16e4db4f8/lib/packet/path_mgmt/seg_req.py#L39)):*
> ```Python
>     def from_values(cls, src_ia, dst_ia, flags=None):
>         if not flags:
>             flags = set()
> ```

Good catch. :+1: 

---

*[lib/packet/sciond/__init__.py, line 0 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJA0clh9qV3D9yI4K-:-KeJA0clh9qV3D9yI4K0:bu04vs9) ([raw file](https://github.com/netsec-ethz/scion/blob/98f336cdc4e5db2aed6fe21f6ec70be53a76a0fa/lib/packet/sciond/__init__.py#L0)):*
This isn't packet-specific, really. And `sciond` is a bit ambigious. How about `lib/sciond_api/`?

---

*[test/lib/path_combinator_test.py, line 1 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJAs0jJWpkD_JDsoCl:-KeJAs0jJWpkD_JDsoCm:b-sdovbr) ([raw file](https://github.com/netsec-ethz/scion/blob/d07d2263ed965182803453a1e9fd0f7be176981a/test/lib/path_combinator_test.py#L1)):*
> ```Python
> # Stdlib
> ```

Copyright header etc missing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1002)*

shitz:
Review status: 3 of 19 files reviewed at latest revision, 5 unresolved discussions.

---

*[lib/path_combinator.py, line 29 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJ5uq9TOaMXjWFEddo:-KeJHP5ZB6OLO9VSVCoj:biw8y1v) ([raw file](https://github.com/netsec-ethz/scion/blob/e72b88a5d49e6ccaefc96bfe3edcd1b16e4db4f8/lib/path_combinator.py#L29)):*

Now that it's in its own module, there's no need to have an artificial class for this.
</blockquote></details>

Right. Done.

---

*[lib/path_combinator.py, line 424 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJ6CohUqteYsnhKKY5:-KeJKDBKfsQdnfw7aq82:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e72b88a5d49e6ccaefc96bfe3edcd1b16e4db4f8/lib/path_combinator.py#L424)):*

These two aren't used outside this module, so prepend them with `_`.
</blockquote></details>

Done.

---

*[test/lib/path_combinator_test.py, line 1 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJAs0jJWpkD_JDsoCl:-KeJNC7_qJmqjt_pG2Cn:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d07d2263ed965182803453a1e9fd0f7be176981a/test/lib/path_combinator_test.py#L1)):*

Copyright header etc missing.
</blockquote></details>

Done.

---

*[lib/packet/sciond/__init__.py, line  at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJA0clh9qV3D9yI4K-:-KeJNDVRvqoTMy_6-yIC:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/98f336cdc4e5db2aed6fe21f6ec70be53a76a0fa/lib/packet/sciond/__init__.py#L)):*

This isn't packet-specific, really. And `sciond` is a bit ambigious. How about `lib/sciond_api/`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1002)*

shitz:
Review status: 8 of 19 files reviewed at latest revision, 15 unresolved discussions.

---

*[lib/packet/path_mgmt/seg_req.py, line 39 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJ9uxVHLAmo0j8PCSr:-KeJnhTkVHg3vqfrPtOQ:bw12u8o) ([raw file](https://github.com/netsec-ethz/scion/blob/e72b88a5d49e6ccaefc96bfe3edcd1b16e4db4f8/lib/packet/path_mgmt/seg_req.py#L39)):*

Good catch. :+1: 
</blockquote></details>

Acknowledging, since the 'Ack' button doesn't seem to do it's job...

---

*[lib/sciond_api/base.py, line 26 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJTadwc_y22tRGKEh1:-KeJhbxtuaRP9xZa299G:b-d4hqli) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/base.py#L26)):*

The approach that `SCIONPayloadBaseProto` takes is to not define it in the base class at all, so that attempts to use a base class without defining it will break in an obvious way.
</blockquote></details>

Removed and added docstring explaining what subclasses need to do.

---

*[lib/sciond_api/parse.py, line 1 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJTw1HUWIbijEC-vxC:-KeJhweBURsDgEvKCiXK:b-oehs7t) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/parse.py#L1)):*

Given how small both this and `base.py` are, i'd merge them.
</blockquote></details>

That's not possible, since `parse.py` needs to import the various subclasses of `SCIONDMsgBase` which would lead to circular imports. (I assume that's the reason you did the same for the path mgmt classes.)

---

*[lib/sciond_api/path_meta.py, line 56 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJYa3ccCY9lyjTRvmd:-KeJj-nqDsI_Lu5GF-eh:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_meta.py#L56)):*

`short_desc` shouldn't contain `self.NAME`, but the parent class has `__str__` which does this for you instead.
</blockquote></details>

Done.

---

*[lib/sciond_api/path_meta.py, line 74 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJYn1xG_npZtTGOumm:-KeJiL-ekpuhLBwrw68K:b-fndbjn) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_meta.py#L74)):*

Wat? What is this even used for? Are the indexes usually floats?
</blockquote></details>

This is used in `cli_srv_ext_test.py`. I thought it would make sense to add it here, since `PathInterface` is essential a tuple. Indices are not floats but ints can be < 0 and > 1 which shouldn't be allowed in this case.

---

*[lib/sciond_api/path_req.py, line 44 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJZ245XiS91vVHLwqI:-KeJjeik5Qmh5nZhW7_6:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_req.py#L44)):*

Mmm. I'm not sure about setting `src_ia` if it isn't specified. `0` could also be a special value. There's a difference between "all paths from /any/ AS to the remote AS" vs "all paths from the /default/ AS(es) to the remote AS".
</blockquote></details>

Done.

---

*[lib/sciond_api/path_req.py, line 58 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJZW6_zk5wAygwpVHC:-KeJk5_nOCmoAULnJZaU:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_req.py#L58)):*

We have a more standard approach of using a list of strings, and `.join(" ")`.
</blockquote></details>

Done.

---

*[lib/sciond_api/path_req.py, line 69 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJ_7cLI65cXjBz2_-F:-KeJkU6JPQAvJcFcz1ad:b-tdjjc9) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_req.py#L69)):*

If this is used outside of this module, it needs a better name.
</blockquote></details>

renamed to 'SCIONDPathReplyError'

---

*[lib/sciond_api/path_req.py, line 112 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJ_spZeC38JDuC42PW:-KeJmEnuI1VoeRpJXYdN:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_req.py#L112)):*

Why not just let an `IndexError` happen to make it obvious?
</blockquote></details>

Done.

---

*[lib/sciond_api/path_req.py, line 144 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJas8LP2MGn0ahtnqq:-KeJkzeP9XM62AP27N89:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_req.py#L144)):*

`AddrType.NONE` is illegal.
</blockquote></details>

Done.

---

*[lib/sciond_api/path_req.py, line 160 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJbI4f-_9knbmDxVVT:-KeJl-weKDet5frTdw_S:bfxh10o) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_req.py#L160)):*

Why this?
</blockquote></details>

I thought it made processing a bit easier in the application. Turns out this isn't true at all. Changed it to return `None`.

---

*[lib/sciond_api/path_req.py, line 172 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJf71lkRtcuyTRI0nY:-KeJlfSaWMYDF-3Ifep6:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_req.py#L172)):*

Use list + join, please.
</blockquote></details>

Done.

---

*[proto/sciond.capnp, line 53 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJaZnGOtRm1nSo4a5A:-KeJlmXdB5GRIbtbaIrP:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/proto/sciond.capnp#L53)):*

`UInt64`.
</blockquote></details>

Done.

---

*[proto/sciond.capnp, line 60 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJagMH0_9A8PER33gz:-KeJluW-idRCQNfCFkot:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/proto/sciond.capnp#L60)):*

Maybe specify that the "default" one is first.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1002)*

kormat:
Friendly pass finished. ;)

---

Reviewed 1 of 22 files at r1, 1 of 2 files at r2, 6 of 17 files at r3, 5 of 5 files at r5.

---

*[endhost/sciond.py, line 207 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZFdx3s9-BchA8OcOx:-KeZFdx3s9-BchA8OcOy:b-tsy0x7) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/endhost/sciond.py#L207)):*
> ```Python
>         """
>         if msg.MSG_TYPE == SMT.PATH_REQUEST:
>             threading.Thread(
> ```

I'm not a fan of the one-thread-per-query approach, but that's something to fix later.

---

*[endhost/sciond.py, line 234 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZG6q9wW4270Zer2Q-:-KeZG6q9wW4270Zer2Q0:bge8vdm) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/endhost/sciond.py#L234)):*
> ```Python
>         if request.p.flags.flush:
>             flags = (_FLUSH_FLAG)
>         paths, error = self.get_paths(dst_ia, flags)[:request.p.maxPaths]
> ```

I.. don't see how this works. `get_paths` returns two values, the first of which is a list. You need to save the list first and then slice that.

---

*[endhost/sciond.py, line 242 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZGqbfk-HIN0txs4pb:-KeZGqbfk-HIN0txs4pc:b-v4s03k) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/endhost/sciond.py#L242)):*
> ```Python
>             # Set dummy host addr if path is empty.
>             if fwd_if == 0:
>                 haddr, port = None, SCION_UDP_EH_DATA_PORT
> ```

I don't think this is necessary now that we can capnp. Just don't set the values?

---

*[lib/sciond_api/path_meta.py, line 56 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJYa3ccCY9lyjTRvmd:-KeZDcZDP0SiZ_BnZ439:b-yt1nho) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_meta.py#L56)):*

Done.
</blockquote></details>

Change not visible.

---

*[lib/sciond_api/path_meta.py, line 74 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJYn1xG_npZtTGOumm:-KeZDndlBJoqeNMXLzM8:bnf07hu) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_meta.py#L74)):*

This is used in `cli_srv_ext_test.py`. I thought it would make sense to add it here, since `PathInterface` is essential a tuple. Indices are not floats but ints can be < 0 and > 1 which shouldn't be allowed in this case.
</blockquote></details>

Ah. We have an example of this in the code base, in ISD_AS: https://github.com/netsec-ethz/scion/blob/master/lib/packet/scion_addr.py#L111
I find that form to be easier to read.

---

*[lib/sciond_api/path_req.py, line 45 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZE3ayQxiRRdIso47F:-KeZE3azKBIg5H9PwkgN:b-3r2xa0) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/lib/sciond_api/path_req.py#L45)):*
> ```Python
>         p = cls.P_CLS.new_message(
>             id=id_, dst=int(dst_ia), maxPaths=max_paths)
>         if src_ia:
> ```

To allow for `0` to be a special value, this should be `if src_ia is not None`.

---

*[test/integration/base_cli_srv.py, line 137 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZJpwLPqx9G4cYcIqA:-KeZJpwMK9M6Mm5-u93P:b-sre6nl) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/test/integration/base_cli_srv.py#L137)):*
> ```Python
>         self.first_hop = (fh_addr, path_entry.p.port)
> 
>     def _try_sciond_api(self):
> ```

We talked about adding a "what's my AS?" call before the path lookup.

---

*[test/integration/base_cli_srv.py, line 150 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZM6bohexqm2LH-D81:-KeZM6bpr6Awg4pB16et:b5e1nsz) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/test/integration/base_cli_srv.py#L150)):*
> ```Python
>         while time.time() - start < API_TOUT:
>             logging.debug("Sending path request to local API at %s: %s",
>                           self.sd.api_addr, request)
> ```

`self.api_addr`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-:-KeZNs_ScJKqLuLHI0hr:bbqyssx)*

shitz:

---

*[endhost/sciond.py, line 207 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZFdx3s9-BchA8OcOx:-KeZRRX4kUzFHcqkf8D6:b2nhb2c) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/endhost/sciond.py#L207)):*

I'm not a fan of the one-thread-per-query approach, but that's something to fix later.
</blockquote></details>

Ack

---

*[endhost/sciond.py, line 234 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZG6q9wW4270Zer2Q-:-KeZPGgRvHj8nDzbVtsO:b8ycs33) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/endhost/sciond.py#L234)):*

I.. don't see how this works. `get_paths` returns two values, the first of which is a list. You need to save the list first and then slice that.
</blockquote></details>

I cannot explain why this actually didn't crash. Good catch!

---

*[endhost/sciond.py, line 242 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZGqbfk-HIN0txs4pb:-KeZQBZhyrdnqsIaoQlI:b-mvqsto) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/endhost/sciond.py#L242)):*

I don't think this is necessary now that we can capnp. Just don't set the values?
</blockquote></details>

Setting `haddr` to `None` will lead to no address being set in the proto. Removed the default udp port.

---

*[lib/sciond_api/path_meta.py, line 56 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJYa3ccCY9lyjTRvmd:-KeZRPy7X3daJik1jeDs:b-bi6kpv) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_meta.py#L56)):*

Change not visible.
</blockquote></details>

Now done.

---

*[lib/sciond_api/path_meta.py, line 74 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeJYn1xG_npZtTGOumm:-KeZQtK7165mwnA98XKg:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/eb9529909b20d766c49c97e9edd9daeb99f6e284/lib/sciond_api/path_meta.py#L74)):*

Ah. We have an example of this in the code base, in ISD_AS: https://github.com/netsec-ethz/scion/blob/master/lib/packet/scion_addr.py#L111
I find that form to be easier to read.
</blockquote></details>

Done.

---

*[lib/sciond_api/path_req.py, line 45 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZE3ayQxiRRdIso47F:-KeZR63kfEZRsVo8zzIU:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/lib/sciond_api/path_req.py#L45)):*

To allow for `0` to be a special value, this should be `if src_ia is not None`.
</blockquote></details>

Done.

---

*[test/integration/base_cli_srv.py, line 137 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZJpwLPqx9G4cYcIqA:-KeZR7TdjN8heBoD-wvX:bghhtu3) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/test/integration/base_cli_srv.py#L137)):*

We talked about adding a "what's my AS?" call before the path lookup.
</blockquote></details>

We did and the `ASInfoRequest` is part of the next PR.

---

*[test/integration/base_cli_srv.py, line 150 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KeZM6bohexqm2LH-D81:-KeZRI3pX3nP19vTouyC:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/2119488c40cde1ff37b82845c85364f77d93014e/test/integration/base_cli_srv.py#L150)):*

`self.api_addr`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1002)*

kormat:
One minor comment, LGTM otherwise.

---

Reviewed 2 of 17 files at r3, 4 of 4 files at r6.

---

*[test/integration/base_cli_srv.py, line 136 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KehCwMD6LCFscHOE5vO:-KehCwMD6LCFscHOE5vP:bulczry) ([raw file](https://github.com/netsec-ethz/scion/blob/0d05ee743cf66ac7b24a28585ef842df263e602f/test/integration/base_cli_srv.py#L136)):*
> ```Python
>             fh_addr = self.dst.host
>         port = (path_entry.p.port if path_entry.p.port
>                 else SCION_UDP_EH_DATA_PORT)
> ```

I think this can simply be: 
```
port = path_entry.p.port or SCION_UDP_EH_DATA_PORT
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-:-KehD8H7g2MIWkJjbfHh:b-sfg7c3)*

shitz:
Thanks for the review!

---

Review status: 18 of 19 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[test/integration/base_cli_srv.py, line 136 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-KehCwMD6LCFscHOE5vO:-KehHzwria_TS0_8g7MN:b-ejpnr0) ([raw file](https://github.com/netsec-ethz/scion/blob/0d05ee743cf66ac7b24a28585ef842df263e602f/test/integration/base_cli_srv.py#L136)):*

I think this can simply be: 
```
port = path_entry.p.port or SCION_UDP_EH_DATA_PORT
```
</blockquote></details>

Nice. Didn't know about this until now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-:-KehI5cwQTEyJuBEQcjU:b-9qdoz3)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1002#-:-KehRKfGEEFkUwsvHW93:bnfp4nl)*

PR #1008, opened 2017-03-09
Implemented missing calls of SCIOND API

shitz:
This PR implements all the remaining SCIOND API messages (`BRInfo` and `ServiceInfo`) (issue #986). Note, they are currently not tested/used anywhere. This will be done in a follow up PR where also an application layer wrapper around the low level SCIOND API will be provided.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1008)

PR #1011, opened 2017-03-10
lib.app.sciond

shitz:
This PR implements a wrapper around the low level SCIOND API (issue #986) for use by applications. With the introduction of this wrapper, all references to a SCIOND object were removed from the integration tests. Depends on #1008.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1011)


klausman:
Reviewed 10 of 10 files at r1.

---

*[lib/app/sciond.py, line 141 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfAvO3s9gRdfDb9grQh:-KfAvO3s9gRdfDb9grQi:bi5k9pk) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L141)):*
> ```Python
>             if len(br_infos) == len(if_list):
>                 return br_infos
>         br_req = SCIONDBRInfoRequest.from_values(if_list)
> ```

If I understand the logic here correctly, we only use cached values if all of the requested IFs are covered. I.e. a partial cache hit is discarded. Is that on purpose? (same applies to `get-service_info()` below)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

shitz:

---

*[lib/app/sciond.py, line 141 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfAvO3s9gRdfDb9grQh:-KfAy2HIn52RsmSK324b:bbvb857) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L141)):*

If I understand the logic here correctly, we only use cached values if all of the requested IFs are covered. I.e. a partial cache hit is discarded. Is that on purpose? (same applies to `get-service_info()` below)
</blockquote></details>

Yes, this is on purpose. My reasons to do it this way were twofold:
1) It simplifies the logic/implementation
2) If you have to request the info from SCIOND you might as well get a fresh copy for all of it, since the cache is very short-lived anyway (10s). This assumes that the work on the SCIOND side to get this information is cheap which is true for now, but might change in the future.

I can definitely see merits in using partial cache hits, but decided to go for a simpler variant for now. However, I can change this implementation to use partial hits if you insist.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

klausman:
:lgtm: from me, on to @kormat for his pass (i.e. wait for his lgtm before merging)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-:-KfAzuKVaW0nfNSZUxb2:b-e4ucc6)*

klausman:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

klausman:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

kormat:
Reviewed 2 of 10 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[lib/app/sciond.py, line 60 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB-i-cUbKWLG_unsBa:-KfB-i-cUbKWLG_unsBb:b-t3k73b) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L60)):*
> ```Python
>         self._socket = None
>         self._req_id = 0
>         self._br_infos = ExpiringDict(100, 10)
> ```

BR info will change a Lot less frequently than SVC info. This could be cached for a lot longer, say 1h.

If an app gets a path segment that uses an interface it doesn't have info for, it knows it needs to refresh its BR info.

A BR being removed (as opposed to an interface being revoked) is going to be a very rare occurrence, and will effectively be handled the same way as a BR being unreachable (i.e. a packet will timeout, and the client will get updated path segments, which then won't include the removed BR). Operationally, removing a BR would also include revoking its interfaces for the full period of segment validity before all mentions of it are removed.

---

*[lib/app/sciond.py, line 61 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB0qj0ExJEPsdKtjtT:-KfB0qj1yGDXIloRyC8B:b-mna0uw) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L61)):*
> ```Python
>         self._req_id = 0
>         self._br_infos = ExpiringDict(100, 10)
>         self._svc_infos = ExpiringDict(100, 10)
> ```

This is a bit tricky, as the discovery's exported topology view will contain parameters to saw how long to cache the contents. Ideally sciond would add such information to the replies it sends to the client application, and those entries would be cached accordingly. For now an expiring dict with a fixed expiry time will do, we just need to fix this in the future. I suggest adding a `TODO`.

---

*[lib/app/sciond.py, line 71 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB1TX7mTtR5YUYHwmY:-KfB1TX7mTtR5YUYHwmZ:bsd2fsp) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L71)):*
> ```Python
>             raise SCIONDConnectionError()
> 
>     def _get_response(self, expected_type):
> ```

I would move this and `_try_cache` below the public methods.

---

*[lib/app/sciond.py, line 72 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB22diX3c_PqFAJhg0:-KfB22diX3c_PqFAJhg1:be7qp78) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L72)):*
> ```Python
> 
>     def _get_response(self, expected_type):
>         data = self._socket.recv()[0]
> ```

This can raise exceptions -`SCIONIOError` at a minimum.

---

*[lib/app/sciond.py, line 79 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB3Ng3pDcvyPtfb-ae:-KfB3Ng3pDcvyPtfb-af:b1g7vcv) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L79)):*
> ```Python
>         except SCIONParseError as e:
>             raise SCIONDResponseError(str(e))
>         if response.MSG_TYPE != expected_type:
> ```

Hmm. This seems to mean that you can only do queries serially, which is problematic. I.e. if you do a path query which takes a few seconds, you can't do any other sort of query in the meantime. A more general approach would be to keep a list of query IDs (and maybe their expected response types) and match replies to the queries.

---

*[lib/app/sciond.py, line 113 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB48DvpxAxTTrRKufm:-KfB48DvpxAxTTrRKufn:b-3maxmw) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L113)):*
> ```Python
>             raise SCIONDResponseError(
>                 SCIONDPathReplyError.describe(response.p.errorCode))
>         self._req_id += 1
> ```

This is too late - errors in a given request would mean the request ID is reused. Also this isn't thread-safe.

---

*[lib/app/sciond.py, line 127 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB6f8LVH1wSfMZPxwf:-KfB6f8LVH1wSfMZPxwg:b-yi7l0n) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L127)):*
> ```Python
>         return list(response.iter_entries())
> 
>     def get_br_info(self, if_list=None):
> ```

This isn't quite correct - with the new topology format a BR can have multiple internal addresses, only some of which may be associated with a given interface. What the client really wants to know is what are the addresses and ports associated with the specified interface ID, not all addresses and ports the particular BR is using.

---

*[lib/app/sciond.py, line 141 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB6FZzPWpiB2yVOIAW:-KfB6F_-ohTnbaTuYFnI:b-vvighg) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L141)):*
> ```Python
>             if len(br_infos) == len(if_list):
>                 return br_infos
>         br_req = SCIONDBRInfoRequest.from_values(if_list)
> ```

The logic here is a bit confusing. If you can satisfy the complete request from the cache, you do so, otherwise you completely ignore the cache and fetch all the info again?
It would simplify things to change the return value to a dict of IFID -> IF Info. That way you can easily see which IF Infos still need to be fetched.

---

*[lib/app/sciond.py, line 162 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB7DoASFKpSGCc1MuI:-KfB7DoBSL25UYPeLrUv:b-bymlnl) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L162)):*
> ```Python
>             svc_infos = self._try_cache(self._svc_infos, service_types)
>             if len(svc_infos) == len(service_types):
>                 return svc_infos
> ```

Same comment applies here re: cache/fetch all.

---

*[lib/app/sciond.py, line 171 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB9wZK10s7nCzI_faJ:-KfB9wZK10s7nCzI_faK:bzgk06v) ([raw file](https://github.com/netsec-ethz/scion/blob/89561002be4b12cdb9530de24291bd623ee3c712/lib/app/sciond.py#L171)):*
> ```Python
>         return entries
> 
>     def get_first_hop(self, spkt):
> ```

Hum. I notice that `scion_elem` still has `get_first_hop` etc. I assume the plan is to remove that version.

I'm not actually convinced that this belongs in `lib.app.sciond`, then things which don't actually use sciond will be including a sciond library.

I'm also not convinced that `get_first_hop` is the right level of abstraction. I would instead have `get_first_ifid`, which will return `0` if it isn't going via a BR. Converting an IFID to an IF Info is already handled, or an SVC address to a host info, is already handled by the other functionality in this library. There can, of course, be a higher level function which calls the lower-level ones for the simple/common case (though i think `get_first_hop` is still a bad name - `get_overlay_dest` would be a lot more descriptive). E.g. the BS must already maintain its own IFID -> IF Info mapping, so making it use the mapping from `lib.app.sciond` leaves room for weird behaviour. It would instead want to use `get_first_ifid`.

---

*[lib/app/sciond.py, line 189 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfBF7dHwgPi0cWAU85h:-KfBF7dHwgPi0cWAU85i:br6e9uj) ([raw file](https://github.com/netsec-ethz/scion/blob/89561002be4b12cdb9530de24291bd623ee3c712/lib/app/sciond.py#L189)):*
> ```Python
>                 return self._empty_first_hop(dst)
>             if_id = path.get_fwd_if()
>         if if_id in self._br_infos:
> ```

This is a race-condition. The entry can expire between the `if` and the `return`.

---

*[lib/app/sciond.py, line 202 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfBFW7WlFOVZ_yPtK0u:-KfBFW7WlFOVZ_yPtK0v:bso78y4) ([raw file](https://github.com/netsec-ethz/scion/blob/89561002be4b12cdb9530de24291bd623ee3c712/lib/app/sciond.py#L202)):*
> ```Python
>                 return if_id
> 
>     def _empty_first_hop(self, dst):
> ```

This doesn't have the old check to make sure the destination is local - why remove it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

shitz:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[lib/app/sciond.py, line 60 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB-i-cUbKWLG_unsBa:-KfBwhtlgDJtg6ZiOYRk:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L60)):*

BR info will change a Lot less frequently than SVC info. This could be cached for a lot longer, say 1h.

If an app gets a path segment that uses an interface it doesn't have info for, it knows it needs to refresh its BR info.

A BR being removed (as opposed to an interface being revoked) is going to be a very rare occurrence, and will effectively be handled the same way as a BR being unreachable (i.e. a packet will timeout, and the client will get updated path segments, which then won't include the removed BR). Operationally, removing a BR would also include revoking its interfaces for the full period of segment validity before all mentions of it are removed.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 61 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB0qj0ExJEPsdKtjtT:-KfBwifoWZzsJ7rLInpv:bkikbrs) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L61)):*

This is a bit tricky, as the discovery's exported topology view will contain parameters to saw how long to cache the contents. Ideally sciond would add such information to the replies it sends to the client application, and those entries would be cached accordingly. For now an expiring dict with a fixed expiry time will do, we just need to fix this in the future. I suggest adding a `TODO`.
</blockquote></details>

Added TODO and already added a `ttl` field to the proto, such that we don't have to change the message format to accommodate for this extension in the future.

---

*[lib/app/sciond.py, line 71 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB1TX7mTtR5YUYHwmY:-KfBx-Rp9vJ4Jgi-3fmb:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L71)):*

I would move this and `_try_cache` below the public methods.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 72 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB22diX3c_PqFAJhg0:-KfBxSzqoReaLaty3SHY:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L72)):*

This can raise exceptions -`SCIONIOError` at a minimum.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 79 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB3Ng3pDcvyPtfb-ae:-KfBxUklWKWJCPLVXVYl:bg9r2vq) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L79)):*

Hmm. This seems to mean that you can only do queries serially, which is problematic. I.e. if you do a path query which takes a few seconds, you can't do any other sort of query in the meantime. A more general approach would be to keep a list of query IDs (and maybe their expected response types) and match replies to the queries.
</blockquote></details>

There was a big disclaimer that the class is not thread-safe :) Changed the implementation to use a new socket per request, which I think is acceptable for unix or udp sockets.

---

*[lib/app/sciond.py, line 113 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB48DvpxAxTTrRKufm:-KfHNpcrbBA37bn_Pj1y:bdv89gc) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L113)):*

This is too late - errors in a given request would mean the request ID is reused. Also this isn't thread-safe.
</blockquote></details>

Not applicable anymore.

---

*[lib/app/sciond.py, line 127 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB6f8LVH1wSfMZPxwf:-KfI-oG3jDFznQtYG7gm:b-41vloo) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L127)):*

This isn't quite correct - with the new topology format a BR can have multiple internal addresses, only some of which may be associated with a given interface. What the client really wants to know is what are the addresses and ports associated with the specified interface ID, not all addresses and ports the particular BR is using.
</blockquote></details>

Hmm then we either need a way to mark this in the BRInfoReply proto or SCIOND only returns addresses that are associated with an interface (if that's all an app will ever care about).

---

*[lib/app/sciond.py, line 141 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB6FZzPWpiB2yVOIAW:-KfHNkO5K48KsW7HN83_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L141)):*

The logic here is a bit confusing. If you can satisfy the complete request from the cache, you do so, otherwise you completely ignore the cache and fetch all the info again?
It would simplify things to change the return value to a dict of IFID -> IF Info. That way you can easily see which IF Infos still need to be fetched.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 162 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB7DoASFKpSGCc1MuI:-KfHNkp6_T04XfU4_DA4:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L162)):*

Same comment applies here re: cache/fetch all.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 171 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB9wZK10s7nCzI_faJ:-KfHOAOvbT5pFDwkGYfd:ba5ei5t) ([raw file](https://github.com/netsec-ethz/scion/blob/89561002be4b12cdb9530de24291bd623ee3c712/lib/app/sciond.py#L171)):*

Hum. I notice that `scion_elem` still has `get_first_hop` etc. I assume the plan is to remove that version.

I'm not actually convinced that this belongs in `lib.app.sciond`, then things which don't actually use sciond will be including a sciond library.

I'm also not convinced that `get_first_hop` is the right level of abstraction. I would instead have `get_first_ifid`, which will return `0` if it isn't going via a BR. Converting an IFID to an IF Info is already handled, or an SVC address to a host info, is already handled by the other functionality in this library. There can, of course, be a higher level function which calls the lower-level ones for the simple/common case (though i think `get_first_hop` is still a bad name - `get_overlay_dest` would be a lot more descriptive). E.g. the BS must already maintain its own IFID -> IF Info mapping, so making it use the mapping from `lib.app.sciond` leaves room for weird behaviour. It would instead want to use `get_first_ifid`.
</blockquote></details>

Refactored the code according to our discussions offline.

---

*[lib/app/sciond.py, line 189 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfBF7dHwgPi0cWAU85h:-KfHPa3edK744u78nWtv:boj1qgt) ([raw file](https://github.com/netsec-ethz/scion/blob/89561002be4b12cdb9530de24291bd623ee3c712/lib/app/sciond.py#L189)):*

This is a race-condition. The entry can expire between the `if` and the `return`.
</blockquote></details>

Good catch. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

shitz:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[lib/app/sciond.py, line 202 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfBFW7WlFOVZ_yPtK0u:-KfI0KA6oYUhr_7rIoDu:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/89561002be4b12cdb9530de24291bd623ee3c712/lib/app/sciond.py#L202)):*

This doesn't have the old check to make sure the destination is local - why remove it?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

kormat:
Reviewed 11 of 15 files at r4.
Review status: 15 of 19 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[lib/app/sciond.py, line 127 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB6f8LVH1wSfMZPxwf:-KfMPZyN6MpwmXJRoF2s:b6eau40) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L127)):*

Hmm then we either need a way to mark this in the BRInfoReply proto or SCIOND only returns addresses that are associated with an interface (if that's all an app will ever care about).
</blockquote></details>

My point is that it should be `get_if_info`, not `get_br_info`. Things a client will want to know are the internal addr/port(s) for that interface, and some of the interface attributes (BW, MTU, remote IA, link type), and maybe the br Id.

---

*[lib/app/sciond.py, line 23 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMFcyKY1SKEWRRRY4Y:-KfMFcyKY1SKEWRRRY4Z:bn8e2sp) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L23)):*
> ```Python
> import threading
> import time
> from contextlib import closing
> ```

TIL about `contextlib.closing`. Nice find :)

---

*[lib/app/sciond.py, line 53 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMFjE52heaCvuX4Zb9:-KfMFjE52heaCvuX4ZbA:b-y95f7h) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L53)):*
> ```Python
> _SVC_INFO_TTL = 10
> # Time after which a request gets retired.
> _SCIOND_TO = 5
> ```

`_SCIOND_TOUT` is probably more readable.

---

*[lib/app/sciond.py, line 103 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMF0iuFHjtk2-jHY5d:-KfMF0iuFHjtk2-jHY5e:b-31biuw) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L103)):*
> ```Python
>         self._as_info_lock = threading.Lock()
> 
>     def get_paths(self, dst_ia, src_ia, max_paths, flags):
> ```

`flags=None`?

---

*[lib/app/sciond.py, line 136 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMJRdkV-PkBU_gLbmZ:-KfMJRdlSTa-JlOA-GvM:b-drxhe0) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L136)):*
This breaks in hilarious ways:

```
>>> brs=lib.app.sciond.get_br_info(if_list=[12, 12])
>>> brs[0].p
<proto.sciond_capnp:BRInfoReplyEntry builder (ifID = 12, hostInfo = (port = 30087, addrs = (ipv4 = "\x00\x00\t")))>
>>> brs[1].p
<proto.sciond_capnp:BRInfoReplyEntry builder (ifID = 55, hostInfo = (port = 30067, addrs = (ipv4 = "\x00\x00\n")))>
```

Changing this to `if set(br_infos) == set(if_list)` fixes this. I would just convert `if_list` to a set at the start, and use it like that.

---

*[lib/app/sciond.py, line 137 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMLWUUE5KK0inEQ-Pm:-KfMLWUUE5KK0inEQ-Pn:blca56m) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L137)):*
> ```Python
>                 br_infos = self._try_cache(self._br_infos, if_list)
>                 if len(br_infos) == len(if_list):
>                     return br_infos
> ```

This returns a dict, the main `return` below returns a list. Returning a dict is _probably_ the most useful.

---

*[lib/app/sciond.py, line 138 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMMXA8chPLkz95j7EX:-KfMMXA8chPLkz95j7EY:b2yit8e) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L138)):*
The partially cached info is thrown away:
```
>>> lib.app.sciond.get_br_info(if_list=[12])
[<lib.sciond_api.br_req.SCIONDBRInfoReplyEntry object at 0x7f5a918a7240>]
>>> lib.app.sciond.get_br_info(if_list=[12,55])
[<lib.sciond_api.br_req.SCIONDBRInfoReplyEntry object at 0x7f5a91459518>]
```
I suggest instead that you have `br_infos = {}` at the top, and when you iterate over the results below that you add the entries to it, and then return that dict.

---

*[lib/app/sciond.py, line 180 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMNqoaffKs4hkCyHNw:-KfMNqoaffKs4hkCyHNx:b-i97iki) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L180)):*
> ```Python
>         if br_info:
>             return br_info.host_info()
>         br_infos = self.get_br_info([if_id])
> ```

Given that `get_br_info` looks in the cache first, i don't think the first 3 lines are needed.

---

*[lib/app/sciond.py, line 186 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMOQumn9fkwDvSnInw:-KfMOQumn9fkwDvSnInx:ba0btmd) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L186)):*
> ```Python
> 
>     def _resolve_dst_addr(self, dst):
>         as_info = self.get_as_info()
> ```

It's better to look at the source IA of the packet, as you can't determine which as_info to look at.

---

*[lib/app/sciond.py, line 195 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMOlMlHpG_3oa732sH:-KfMOlMlHpG_3oa732sI:b5jtnci) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L195)):*
> ```Python
>         if host.TYPE == AddrType.SVC:
>             svc_type = SVC_TO_SERVICE[host.addr]
>             svc_info = self._svc_infos.get(svc_type)
> ```

It seems simpler to just call `get_service_info`, as it will consult the cache anyway.

---

*[lib/sciond_api/service_req.py, line 113 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMRfJS3edODyAeyh3B:-KfMRfJS3edODyAeyh3C:bs1zv3o) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/sciond_api/service_req.py#L113)):*
> ```Python
>         ttl_str = "unset"
>         if self.p.ttl is not None:
>             ttl_str = "%s s" % self.p.ttl
> ```

I'd remove the space.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

shitz:
Review status: 15 of 19 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[lib/app/sciond.py, line 127 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB6f8LVH1wSfMZPxwf:-KfM_j2iAQ_tMGZTrhp9:b-vh7u78) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L127)):*

My point is that it should be `get_if_info`, not `get_br_info`. Things a client will want to know are the internal addr/port(s) for that interface, and some of the interface attributes (BW, MTU, remote IA, link type), and maybe the br Id.
</blockquote></details>

Uhh for that we need to expand the API. Currently, it only returns the addresses for a specific interface.

---

*[lib/app/sciond.py, line 53 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMFjE52heaCvuX4Zb9:-KfMUDBzNST_yweu_-_l:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L53)):*

`_SCIOND_TOUT` is probably more readable.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 103 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMF0iuFHjtk2-jHY5d:-KfMUGKUSpCAPoKzvVPn:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L103)):*

`flags=None`?
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 136 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMJRdkV-PkBU_gLbmZ:-KfMVKpuXDqjjGkQr4SE:b-we799c) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L136)):*

This breaks in hilarious ways:

```
>>> brs=lib.app.sciond.get_br_info(if_list=[12, 12])
>>> brs[0].p
<proto.sciond_capnp:BRInfoReplyEntry builder (ifID = 12, hostInfo = (port = 30087, addrs = (ipv4 = "\x00\x00\t")))>
>>> brs[1].p
<proto.sciond_capnp:BRInfoReplyEntry builder (ifID = 55, hostInfo = (port = 30067, addrs = (ipv4 = "\x00\x00\n")))>
```

Changing this to `if set(br_infos) == set(if_list)` fixes this. I would just convert `if_list` to a set at the start, and use it like that.
</blockquote></details>

Dayum, that's a good catch!

---

*[lib/app/sciond.py, line 137 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMLWUUE5KK0inEQ-Pm:-KfMWhRdwaE5dIyWTJim:b452poa) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L137)):*

This returns a dict, the main `return` below returns a list. Returning a dict is _probably_ the most useful.
</blockquote></details>

Done. Here and in `get_service_info`

---

*[lib/app/sciond.py, line 138 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMMXA8chPLkz95j7EX:-KfMZt8R2VK2YKN40NiZ:blv56qr) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L138)):*

The partially cached info is thrown away:
```
>>> lib.app.sciond.get_br_info(if_list=[12])
[<lib.sciond_api.br_req.SCIONDBRInfoReplyEntry object at 0x7f5a918a7240>]
>>> lib.app.sciond.get_br_info(if_list=[12,55])
[<lib.sciond_api.br_req.SCIONDBRInfoReplyEntry object at 0x7f5a91459518>]
```
I suggest instead that you have `br_infos = {}` at the top, and when you iterate over the results below that you add the entries to it, and then return that dict.
</blockquote></details>

Good catch! Done here and in `get_service_info`

---

*[lib/app/sciond.py, line 180 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMNqoaffKs4hkCyHNw:-KfM_3z8LDzOm5sUOTky:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L180)):*

Given that `get_br_info` looks in the cache first, i don't think the first 3 lines are needed.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 186 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMOQumn9fkwDvSnInw:-KfM_Kb06JazNxVfMBQE:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L186)):*

It's better to look at the source IA of the packet, as you can't determine which as_info to look at.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 195 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMOlMlHpG_3oa732sH:-KfM_QrtLPPlOxMfx9wm:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/app/sciond.py#L195)):*

It seems simpler to just call `get_service_info`, as it will consult the cache anyway.
</blockquote></details>

Done.

---

*[lib/sciond_api/service_req.py, line 113 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMRfJS3edODyAeyh3B:-KfM_zHeXUsonLHndtJp:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/531b7df15ce3b424630207505f6677783c2e331b/lib/sciond_api/service_req.py#L113)):*

I'd remove the space.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

kormat:
Reviewed 3 of 10 files at r1, 3 of 15 files at r4, 3 of 3 files at r5.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[lib/app/sciond.py, line 127 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB6f8LVH1wSfMZPxwf:-KfMrj5GlDQQbG8KF-Ra:b-6mar95) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L127)):*

Uhh for that we need to expand the API. Currently, it only returns the addresses for a specific interface.
</blockquote></details>

That's probably ok for now, but it's still all named about BR, which is misleading.

---

*[lib/app/sciond.py, line 90 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMrMFXhvPMIFhRU7Dg:-KfMrMFXhvPMIFhRU7Dh:b-1lbnjo) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L90)):*
> ```Python
>         with self._lock:
>             self._cntr += 1
>             return self._cntr
> ```

Oh. This should wrap, otherwise who knows what happens when the int value gets too big for the capnp field.

---

*[lib/app/sciond.py, line 141 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMqiIzWJTrMtP2sxh-:-KfMqiIzWJTrMtP2sxh0:b-42uz1l) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L141)):*
> ```Python
>                 if_list = set(if_list)
>                 cached_infos = self._try_cache(self._br_infos, if_list)
>                 if len(cached_infos) == len(if_list):
> ```

You can combine this with the next operation:
```
if_list = if_list - set(cached_infos)
if not if_list:
  return cached_infos
```

---

*[lib/app/sciond.py, line 143 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMq5VU1f5tqKyb0eic:-KfMq5VU1f5tqKyb0eid:b-y7hj0k) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L143)):*
> ```Python
>                 if len(cached_infos) == len(if_list):
>                     return cached_infos
>                 if_list = list(if_list - set(cached_infos.keys()))
> ```

You don't actually need `.keys()` here, as `set()` on a dict does that automatically. I'm also not sure if you need to convert it back to a list.

---

*[lib/app/sciond.py, line 152 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMqKRTIZd0rgdw0P07:-KfMqKRTIZd0rgdw0P08:b-42gnz1) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L152)):*
> ```Python
>             for if_id, entry in entries.items():
>                 self._br_infos[if_id] = entry
>             return {**cached_infos, **entries}
> ```

I think this is simpler:
```
for e in response.iter_entries():
   self._br_infos[e.p.ifID] = e
   cached_infos[e.p.ifID] = e
return cached_infos
```
(though i'd rename `cached_infos`).

---

*[lib/app/sciond.py, line 178 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMtrCnT0dK-m-Cvqle:-KfMtrCnT0dK-m-Cvqlf:b8mt1ws) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L178)):*
> ```Python
>             return {**cached_infos, **entries}
> 
>     def get_next_hop_overlay_dest(self, spkt):
> ```

`get_overlay_dest` is more succinct.

---

*[lib/app/sciond.py, line 263 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMrXpe2KgdziauzAiR:-KfMrXpe2KgdziauzAiS:bav53x1) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L263)):*
> ```Python
>     global _counter
>     if not _counter:
>         _counter = _Counter(random.randint(0, 10**9))
> ```

A power of 2 would be better than power of 10 (c.f. comment about wrapping in Counter).

---

*[test/integration/base_cli_srv.py, line 180 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMsOPv--9_RZ1UpHF4:-KfMsOPv--9_RZ1UpHF5:boki867) ([raw file](https://github.com/netsec-ethz/scion/blob/67a94321593f6f148d8919962793ff6bef91b1e1/test/integration/base_cli_srv.py#L180)):*
> ```Python
>         kill_self()
> 
>     def _get_path_direct(self, flags=0):
> ```

Diiiiiie! :grinning: 

---

*[test/integration/cert_req_test.py, line 43 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMs_cPAZRFQRVBYpqe:-KfMs_cPAZRFQRVBYpqf:bj3rbqp) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/test/integration/cert_req_test.py#L43)):*
> ```Python
>     def __init__(self, api_addr, finished, addr):
>         # We need the lib sciond here already.
>         connector = lib_sciond.init(api_addr)
> ```

We don't care about having a custom connector here, as there's only a client, no server.

---

*[test/integration/scmp_error_test.py, line 85 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMtysWCg_zBP-SHZfc:-KfMtysWCg_zBP-SHZfd:b-5dl3dw) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/test/integration/scmp_error_test.py#L85)):*
> ```Python
>         return False
> 
>     def _get_next_hop(self, spkt):
> ```

I think this functionality is used by many of the integration tests, so maybe move it into `TestClientBase` directly?

---

*[test/integration/scmp_error_test.py, line 276 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMuD9wE9jF5JMb-ifF:-KfMuD9wE9jF5JMb-ifG:bcwr6p4) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/test/integration/scmp_error_test.py#L276)):*
> ```Python
>             return True
>         # Only run if the remote AS is a core AS.
>         as_info = lib_sciond.get_as_info(self._connector)[0]
> ```

This.. is not doing the same thing. The important part is if the _destination_ is a core AS.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

kormat:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[lib/app/sciond.py, line 251 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQ_0KtOMUsaq9f7T8F:-KfQ_0KtOMUsaq9f7T8G:b-wu0dov) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L251)):*
> ```Python
> 
> 
> def init(api_addr):
> ```

Just occurred to me - please make `api_addr` optional. If it is not present, try to get the address from a `SCIOND_PATH` env var, and if that doesn't work, try `SCIOND_API_SOCKDIR`/`default.sock`. I don't think it's worth `stat`'ing the path at `init()` time, as sciond might not have started yet etc.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

shitz:
Review status: 10 of 21 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[lib/app/sciond.py, line 127 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfB6f8LVH1wSfMZPxwf:-KfMzI0qOL4XUkCP-sXO:b-4uc2y5) ([raw file](https://github.com/netsec-ethz/scion/blob/fa91bf0001fabcdd08f457fcad2f3c0358e14107/lib/app/sciond.py#L127)):*

That's probably ok for now, but it's still all named about BR, which is misleading.
</blockquote></details>

Globally rename `BRInfo` to `IFInfo`

---

*[lib/app/sciond.py, line 90 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMrMFXhvPMIFhRU7Dg:-KfMvlNqcFLKz2AzzHu1:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L90)):*

Oh. This should wrap, otherwise who knows what happens when the int value gets too big for the capnp field.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 141 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMqiIzWJTrMtP2sxh-:-KfMx98yI3IDhny2vhp_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L141)):*

You can combine this with the next operation:
```
if_list = if_list - set(cached_infos)
if not if_list:
  return cached_infos
```
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 143 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMq5VU1f5tqKyb0eic:-KfMvx8EIa6260djxdTF:b-y4lipx) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L143)):*

You don't actually need `.keys()` here, as `set()` on a dict does that automatically. I'm also not sure if you need to convert it back to a list.
</blockquote></details>

Yep no need to convert it back, since `from_values` expects any iterable.

---

*[lib/app/sciond.py, line 152 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMqKRTIZd0rgdw0P07:-KfMyQf6Sc7MnHt5Qk67:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L152)):*

I think this is simpler:
```
for e in response.iter_entries():
   self._br_infos[e.p.ifID] = e
   cached_infos[e.p.ifID] = e
return cached_infos
```
(though i'd rename `cached_infos`).
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 178 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMtrCnT0dK-m-Cvqle:-KfMyzE1N9FIzzGUllrg:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L178)):*

`get_overlay_dest` is more succinct.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 251 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQ_0KtOMUsaq9f7T8F:-KfQhWgI3MlK57Qd6KJ4:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L251)):*

Just occurred to me - please make `api_addr` optional. If it is not present, try to get the address from a `SCIOND_PATH` env var, and if that doesn't work, try `SCIOND_API_SOCKDIR`/`default.sock`. I don't think it's worth `stat`'ing the path at `init()` time, as sciond might not have started yet etc.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 263 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMrXpe2KgdziauzAiR:-KfMz7IuVq-nYgM9Mb5n:b-kgn1wy) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/lib/app/sciond.py#L263)):*

A power of 2 would be better than power of 10 (c.f. comment about wrapping in Counter).
</blockquote></details>

Don't think it makes a difference, but I changed it to 2**32 - 1.

---

*[test/integration/base_cli_srv.py, line 180 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMsOPv--9_RZ1UpHF4:-KfN0GTUcMQCRa-Tpnfl:bb6sh2i) ([raw file](https://github.com/netsec-ethz/scion/blob/67a94321593f6f148d8919962793ff6bef91b1e1/test/integration/base_cli_srv.py#L180)):*

Diiiiiie! :grinning: 
</blockquote></details>

Ack.

---

*[test/integration/cert_req_test.py, line 43 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMs_cPAZRFQRVBYpqe:-KfN0PfnwcX8WkFqQkUL:b-92jfj) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/test/integration/cert_req_test.py#L43)):*

We don't care about having a custom connector here, as there's only a client, no server.
</blockquote></details>

We do, otherwise all lookups return the cert server address of the AS that was tested first.

---

*[test/integration/scmp_error_test.py, line 85 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMtysWCg_zBP-SHZfc:-KfN0R2kB8pIDFLINMRq:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/test/integration/scmp_error_test.py#L85)):*

I think this functionality is used by many of the integration tests, so maybe move it into `TestClientBase` directly?
</blockquote></details>

Done.

---

*[test/integration/scmp_error_test.py, line 276 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfMuD9wE9jF5JMb-ifF:-KfQeAyLVMx9OQUwxc1g:btmnazv) ([raw file](https://github.com/netsec-ethz/scion/blob/85addda13bd86f44283535edcf1c4b2f3f624266/test/integration/scmp_error_test.py#L276)):*

This.. is not doing the same thing. The important part is if the _destination_ is a core AS.
</blockquote></details>

Fixed by adding an optional `ISD_AS` parameter to `get_as_info`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

kormat:
Reviewed 12 of 12 files at r6.

---

*[endhost/sciond.py, line 288 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfR04aqIHH1LTTCGIiP:-KfR04arT1o9YEzqKozE:ba3o3qj) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/endhost/sciond.py#L288)):*
> ```Python
>         if not all_brs:
>             if_list = list(request.iter_ids())
>         br_entries = []
> ```

Rename.

---

*[lib/app/sciond.py, line 147 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQnyF4a3djYokpfx-L:-KfQnyF4a3djYokpfx-M:b-rkkbsv) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/lib/app/sciond.py#L147)):*
> ```Python
>                 if_list = set(if_list)
>                 if_infos = self._try_cache(self._if_infos, if_list)
>                 if_list = list(if_list - set(if_infos))
> ```

Why `list()`?

---

*[lib/app/sciond.py, line 169 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQoDNTVzBCGD6wtHxW:-KfQoDNTVzBCGD6wtHxX:b-rkkbsv) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/lib/app/sciond.py#L169)):*
> ```Python
>                 service_types = set(service_types)
>                 svc_infos = self._try_cache(self._svc_infos, service_types)
>                 service_types = list(service_types - set(svc_infos))
> ```

Why `list()`?

---

*[lib/app/sciond.py, line 277 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQobnmwojWw5fKJ1No:-KfQobnmwojWw5fKJ1Np:bdibw4v) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/lib/app/sciond.py#L277)):*
> ```Python
> 
> def _get_api_addr():
>     return os.environ.get(
> ```

`os.getenv(...)` is slightly cleaner.

---

*[lib/sciond_api/as_req.py, line 38 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQosnAM9nVijMTt8K6:-KfQosnAM9nVijMTt8K7:bruqldl) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/lib/sciond_api/as_req.py#L38)):*
> ```Python
>         if isd_as:
>             p.isdas = int(isd_as)
> ```

`isd_as.int()` is slightly clearer (it's a little more obvious how the conversion happens).

---

*[proto/sciond.capnp, line 82 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQlmc_z6zPQRLZjrsi:-KfQlmc_z6zPQRLZjrsj:b-bq5mgg) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/proto/sciond.capnp#L82)):*
> ```
> 
> struct IFInfoRequest {
>     ifIDs @0 :List(UInt64);  # The if IDs for which a client requests the host infos. Empty list means all interfaces of all BRs.
> ```

"Empty list means all interfaces of the local AS."

---

*[proto/sciond.capnp, line 90 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQlqX1jXH2QXd6crpP:-KfQlqX2ODcPrPqU_rcv:bhl6x7y) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/proto/sciond.capnp#L90)):*
> ```
> 
> struct IFInfoReplyEntry {
>     ifID @0 :UInt64;  # The if ID of the BR.
> ```

"The ID of the interface."

---

*[proto/sciond.capnp, line 91 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQm6rMvTRB2hzUrVk4:-KfQm6rMvTRB2hzUrVk5:b-sfx432) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/proto/sciond.capnp#L91)):*
> ```
> struct IFInfoReplyEntry {
>     ifID @0 :UInt64;  # The if ID of the BR.
>     hostInfo @1 :HostInfo;  # The host info of the BR.
> ```

"The host info of the internal address of the interface."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

shitz:
Review status: 16 of 20 files reviewed at latest revision, 9 unresolved discussions.

---

*[endhost/sciond.py, line 288 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfR04aqIHH1LTTCGIiP:-KfR8zibKZTpPX-t-Y4R:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/endhost/sciond.py#L288)):*

Rename.
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 147 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQnyF4a3djYokpfx-L:-KfR960hYpLGKiHYek2O:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/lib/app/sciond.py#L147)):*

Why `list()`?
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 169 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQoDNTVzBCGD6wtHxW:-KfR98XRBFn2atFwB17W:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/lib/app/sciond.py#L169)):*

Why `list()`?
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 277 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQobnmwojWw5fKJ1No:-KfR9PmN0t5uJwTcwfZ3:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/lib/app/sciond.py#L277)):*

`os.getenv(...)` is slightly cleaner.
</blockquote></details>

Done.

---

*[lib/sciond_api/as_req.py, line 38 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQosnAM9nVijMTt8K6:-KfR9Wh6D6wjiXUpp9Vz:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/lib/sciond_api/as_req.py#L38)):*

`isd_as.int()` is slightly clearer (it's a little more obvious how the conversion happens).
</blockquote></details>

Done.

---

*[proto/sciond.capnp, line 82 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQlmc_z6zPQRLZjrsi:-KfR9erryxZUQ7PY8HfP:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/proto/sciond.capnp#L82)):*

"Empty list means all interfaces of the local AS."
</blockquote></details>

Done.

---

*[proto/sciond.capnp, line 90 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQlqX1jXH2QXd6crpP:-KfR9inSF2Y1nV-QsKZy:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/proto/sciond.capnp#L90)):*

"The ID of the interface."
</blockquote></details>

Done.

---

*[proto/sciond.capnp, line 91 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfQm6rMvTRB2hzUrVk4:-KfR9lvzEtLFCGEiFTBP:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e6ffc031bae2e94440d3de7f10df49e2b07211fa/proto/sciond.capnp#L91)):*

"The host info of the internal address of the interface."
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

kormat:
Reviewed 4 of 5 files at r7.

---

*[endhost/sciond.py, line 294 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfRPI7SGhTG1bijCKZM:-KfRPI7SGhTG1bijCKZN:b95yjck) ([raw file](https://github.com/netsec-ethz/scion/blob/51de5d0b35d1ab4984f2151583daa4d81af28f5f/endhost/sciond.py#L294)):*
> ```Python
>                 reply_entry = SCIONDIFInfoReplyEntry.from_values(if_id, info)
>                 if_entries.append(reply_entry)
>         br_reply = SCIONDIFInfoReply.from_values(request.id, if_entries)
> ```

Rename `br_reply` :P

---

*[lib/app/sciond.py, line 278 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfRPRfbE2HtzB0osSU5:-KfRPRfbE2HtzB0osSU6:bpixj3l) ([raw file](https://github.com/netsec-ethz/scion/blob/51de5d0b35d1ab4984f2151583daa4d81af28f5f/lib/app/sciond.py#L278)):*
> ```Python
> def _get_api_addr():
>     return os.getenv(
>         "SCIOND_PATH", os.path.join(SCIOND_API_SOCKDIR, "default.sock"))
> ```

Aand it just occurs to me that i should have added a constant for the _name_ of the env var. Can you please add one to `lib.defines`? And probably also for `default.sock`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

shitz:
Review status: 18 of 22 files reviewed at latest revision, 2 unresolved discussions.

---

*[endhost/sciond.py, line 294 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfRPI7SGhTG1bijCKZM:-KfRRMiQNO9ND7BxuWyk:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/51de5d0b35d1ab4984f2151583daa4d81af28f5f/endhost/sciond.py#L294)):*

Rename `br_reply` :P
</blockquote></details>

Done.

---

*[lib/app/sciond.py, line 278 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-KfRPRfbE2HtzB0osSU5:-KfRRDuoLU70UavuNDF9:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/51de5d0b35d1ab4984f2151583daa4d81af28f5f/lib/app/sciond.py#L278)):*

Aand it just occurs to me that i should have added a constant for the _name_ of the env var. Can you please add one to `lib.defines`? And probably also for `default.sock`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011)*

kormat:
:lgtm:

---

Reviewed 3 of 4 files at r8, 1 of 2 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1011#-:-KfRUORg5gW-enjObFXW:bnfp4nl)*

PR #1012, opened 2017-03-10
Check that signatures on capnp objects are versioned correctly.

kormat:
Signatures for capnp objects used in PCBs are defined based on the
version of the capnp definition used. I.e. which fields are covered by
the signature (and in which order) is specific to a given capnp
definition version. This means that when new fields are added, a new
signature definition (which fields, in what order) should be specified.
While this doesn't matter for a homogenous test-bed, as soon as there
are changes made in a production protocol, backwards compatibility is
critical.

The version of a capnp object is defined to be the highest field number,
for simplicity. This change simply calculates the current version
number, and checks to make sure that the correct signature method is
being called. This will have to change once we start implementing actual
backwards compatibility, but will at least in the meantime catch most
cases of fields being added (or removed) without the signature
definitions being updated.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1012)

### DIFF
--- a/infrastructure/beacon_server/base.py
+++ b/infrastructure/beacon_server/base.py
@@ -567,7 +567,7 @@ class BeaconServer(SCIONElement, metaclass=ABCMeta):
         cert_ia = asm.isd_as()
         trc = self.trust_store.get_trc(cert_ia[0], asm.p.trcVer)
         return verify_sig_chain_trc(
-            pcb.sig_pack(), asm.p.sig, str(cert_ia), asm.chain(), trc,
+            pcb.sig_pack3(), asm.p.sig, str(cert_ia), asm.chain(), trc,
             asm.p.trcVer)
 
     @abstractmethod
--- a/lib/errors.py
+++ b/lib/errors.py
@@ -75,3 +75,7 @@ class SCIONTCPError(SCIONBaseError):
 
 class SCIONTCPTimeout(SCIONBaseError):
     """SCION TCP timeout"""
+
+
+class SCIONSigVerError(SCIONBaseError):
+    """SCION Signature Version error"""
--- a/lib/packet/pcb.py
+++ b/lib/packet/pcb.py
@@ -27,6 +27,7 @@ import proto.pcb_capnp as P
 from lib.crypto.asymcrypto import sign
 from lib.crypto.certificate_chain import CertificateChain
 from lib.defines import EXP_TIME_UNIT
+from lib.errors import SCIONSigVerError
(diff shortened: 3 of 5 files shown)


shitz:
Reviewed 4 of 5 files at r1.

---

*[lib/packet/pcb.py, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1012#-KetECIqD85hPf_fQWjU:-KetECIqD85hPf_fQWjV:b69x6mr) ([raw file](https://github.com/netsec-ethz/scion/blob/59e0771e0b00cae8091ebb12ccd3f5e17c3218b1/lib/packet/pcb.py#L128)):*
> ```Python
>         self.p.from_dict(d)
> 
>     def sig_pack8(self):
> ```

What's the reason to hardcode the version in the function name over having a `version` parameter? Each time you the version changes, the interface of the class changes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1012)*

kormat:
Review status: 4 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[lib/packet/pcb.py, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1012#-KetECIqD85hPf_fQWjU:-Kf6f2tc3IGevtB4PVjN:b-34r0x2) ([raw file](https://github.com/netsec-ethz/scion/blob/59e0771e0b00cae8091ebb12ccd3f5e17c3218b1/lib/packet/pcb.py#L128)):*

What's the reason to hardcode the version in the function name over having a `version` parameter? Each time you the version changes, the interface of the class changes.
</blockquote></details>

That's on purpose. The intention is to force the coder who is adding fields to handle this appropriately, or at the very least make it obvious if they're not doing so. With the generic `sig_pack` method it's much less obvious when this happens. C.f. pr #1000 which prompted this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1012)*

shitz:
:lgtm:

---

Review status: 4 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1012#-:-Kf6sTDAO2s5EzEMFm_n:bnfp4nl)*

shitz:
Reviewed 1 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1012)*

PR #1021, opened 2017-03-20
Adjust socket timeout in lib.app.sciond

shitz:
Adjust socket timeout in `lib.app.sciond` to 3 seconds and the path request timeout it `endhost.sciond` to 2s. Before, it was the case that the socket timeout in `lib.app.sciond` matched the path request timeout in `endhost.sciond`. It could happen that the app closed the socket, but the `SCIOND` try to send a response over the already closed socket.

Also raise an exception in case send fails.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1021)

### DIFF
--- a/endhost/sciond.py
+++ b/endhost/sciond.py
@@ -77,7 +77,7 @@ class SCIONDaemon(SCIONElement):
     The SCION Daemon used for retrieving and combining paths.
     """
     # Max time for a path lookup to succeed/fail.
-    TIMEOUT = 2
+    PATH_REQ_TOUT = 2
     # Time a path segment is cached at a host (in seconds).
     SEGMENT_TTL = 300
 
@@ -95,7 +95,8 @@ class SCIONDaemon(SCIONElement):
         req_name = "SCIONDaemon Requests %s" % self.addr.isd_as
         self.requests = RequestHandler.start(
             req_name, self._check_segments, self._fetch_segments,
-            self._reply_segments, ttl=self.TIMEOUT, key_map=self._req_key_map,
+            self._reply_segments, ttl=self.PATH_REQ_TOUT,
+            key_map=self._req_key_map,
         )
         self._api_sock = None
         self.daemon_thread = None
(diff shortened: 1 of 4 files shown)


kormat:
Reviewed 2 of 2 files at r1.

---

*[lib/app/sciond.py, line 60 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1021#-KflTDBRzvP7NyVtlTxQ:-KflTDBRzvP7NyVtlTxR:b-934kcx) ([raw file](https://github.com/netsec-ethz/scion/blob/a6302001015bb9895eacd1de5c48891c5b465ee8/lib/app/sciond.py#L60)):*
> ```Python
> _SVC_INFO_TTL = 10
> # Time after which a request gets retired.
> _SCIOND_TOUT = 10
> ```

I don't think we need to increase the timeout in lib.app.sciond. Sciond itself needs to handle the case that the requesting socket is closed by the app, instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1021)*

shitz:
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[lib/app/sciond.py, line 60 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1021#-KflTDBRzvP7NyVtlTxQ:-KfvAaPXGmTfbuQ7IWH3:bc0jxae) ([raw file](https://github.com/netsec-ethz/scion/blob/a6302001015bb9895eacd1de5c48891c5b465ee8/lib/app/sciond.py#L60)):*

I don't think we need to increase the timeout in lib.app.sciond. Sciond itself needs to handle the case that the requesting socket is closed by the app, instead.
</blockquote></details>

I changed the timeout in SCION daemon to 2 seconds and in lib.app.sciond to 3 seconds. This will definitely lead to more timeouts during the test, but we will see how flaky it will be. Note, that  an application gets notified with a timeout error code from SCIOND after 2 seconds. `_SCIOND_TOUT` only comes into play when the connection to SCIOND broke for some reason. `_SCIOND_TOUT` should be >= `PATH_REQ_OUT` to not make SCIOND work more than is needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1021)*

kormat:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1021#-:-Kfvh3Ozq0ktoHW86snL:bnfp4nl)*

PR #1034, opened 2017-04-03
Integration tests use local SCIOND

shitz:
Make integration tests use the local SCIOND instance that gets spawned as part of the infrastructure. Also _try_sciond_api now stops trying to get a path when the connection to SCIOND fails.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1034)

### DIFF
--- a/lib/app/sciond.py
+++ b/lib/app/sciond.py
@@ -229,9 +229,9 @@ class SCIONDConnector:
         socket.settimeout(_SCIOND_TOUT)
         try:
             socket.connect(self._api_addr)
-        except OSError:
+        except OSError as e:
             socket.close()
-            raise SCIONDConnectionError()
+            raise SCIONDConnectionError(str(e))
         return socket
 
     def _get_response(self, socket, expected_id, expected_type):  # pragma: no cover
--- a/test/integration/base_cli_srv.py
+++ b/test/integration/base_cli_srv.py
@@ -31,13 +31,15 @@ from itertools import product
 
 # SCION
 import lib.app.sciond as lib_sciond
-from endhost.sciond import SCIOND_API_SOCKDIR
-from lib.defines import AS_LIST_FILE, GEN_PATH, SCION_UDP_EH_DATA_PORT
+from lib.defines import (
+    AS_LIST_FILE,
+    GEN_PATH,
+    SCIOND_API_SOCKDIR,
+    SCION_UDP_EH_DATA_PORT
+)
 from lib.log import init_logging
 from lib.main import main_wrapper
-from lib.packet.host_addr import (
-    haddr_parse_interface,
-)
+from lib.packet.host_addr import haddr_parse_interface
 from lib.packet.packet_base import PayloadRaw
 from lib.packet.scion import SCIONL4Packet, build_base_hdrs
 from lib.packet.scion_addr import ISD_AS, SCIONAddr
@@ -59,8 +61,8 @@ class ResponseRV:
 
 
 class TestBase(object, metaclass=ABCMeta):
-    def __init__(self, data, finished, addr, timeout=1.0):
-        self.api_addr = SCIOND_API_SOCKDIR + "sd%s.sock" % (addr.isd_as)
+    def __init__(self, data, finished, addr, timeout=1.0, api_addr=None):
+        self.api_addr = api_addr or (SCIOND_API_SOCKDIR + "sd%s.sock" % (addr.isd_as))
         self.data = data
         self.finished = finished
         self.addr = addr
@@ -111,7 +113,7 @@ class TestClientBase(TestBase):
     Base client app
     """
     def __init__(self, data, finished, addr, dst, dport, api=True,
-                 timeout=3.0, retries=0):
+                 timeout=3.0, retries=0, api_addr=None):
         self.dst = dst
         self.dport = dport
         self.api = api
@@ -119,7 +121,7 @@ class TestClientBase(TestBase):
         self.first_hop = None
         self.retries = retries
         self._req_id = 0
-        super().__init__(data, finished, addr, timeout)
+        super().__init__(data, finished, addr, timeout, api_addr)
         self._get_path(api)
 
     def _get_path(self, api, flush=False):
--- a/test/integration/base_cli_srv.py
+++ b/test/integration/base_cli_srv.py
@@ -31,7 +31,7 @@ from itertools import product
 
 # SCION
 import lib.app.sciond as lib_sciond
-from endhost.sciond import SCIOND_API_SOCKDIR, SCIONDaemon
+from endhost.sciond import SCIOND_API_SOCKDIR
 from lib.defines import AS_LIST_FILE, GEN_PATH, SCION_UDP_EH_DATA_PORT
 from lib.log import init_logging
 from lib.main import main_wrapper
@@ -59,8 +59,8 @@ class ResponseRV:
 
 
 class TestBase(object, metaclass=ABCMeta):
-    def __init__(self, api_addr, data, finished, addr, timeout=1.0):
-        self.api_addr = api_addr
+    def __init__(self, data, finished, addr, timeout=1.0):
+        self.api_addr = SCIOND_API_SOCKDIR + "sd%s.sock" % (addr.isd_as)
         self.data = data
         self.finished = finished
         self.addr = addr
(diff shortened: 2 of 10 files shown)


kormat:
Reviewed 2 of 10 files at r1.
Review status: 2 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[test/integration/base_cli_srv.py, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1034#-Kgse8DDkOgWNgA_Ynkm:-Kgse8DDkOgWNgA_Ynkn:bsgmkq4) ([raw file](https://github.com/netsec-ethz/scion/blob/086315a414432ec8ba4d32ade7d6920f998cdcce/test/integration/base_cli_srv.py#L34)):*
> ```Python
> # SCION
> import lib.app.sciond as lib_sciond
> from endhost.sciond import SCIOND_API_SOCKDIR
> ```

This is in `lib.defines` now i believe.

---

*[test/integration/base_cli_srv.py, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1034#-KgsdN_ooM56HLS6kABd:-KgsdN_ooM56HLS6kABe:bn316rd) ([raw file](https://github.com/netsec-ethz/scion/blob/086315a414432ec8ba4d32ade7d6920f998cdcce/test/integration/base_cli_srv.py#L63)):*
> ```Python
> class TestBase(object, metaclass=ABCMeta):
>     def __init__(self, data, finished, addr, timeout=1.0):
>         self.api_addr = SCIOND_API_SOCKDIR + "sd%s.sock" % (addr.isd_as)
> ```

This is a bit too hard-coded for my taste. I would still pass in `api_addr`, and do the calculation in the caller.

(Also it should be `os.path.join(SCIOND_API_SOCKDIR, "sd%s.sock" % addr.isd_as)`)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1034)*

shitz:
Review status: 2 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[test/integration/base_cli_srv.py, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1034#-Kgse8DDkOgWNgA_Ynkm:-KgtD2P1sRYPWyEG26fV:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/086315a414432ec8ba4d32ade7d6920f998cdcce/test/integration/base_cli_srv.py#L34)):*

This is in `lib.defines` now i believe.
</blockquote></details>

Done.

---

*[test/integration/base_cli_srv.py, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1034#-KgsdN_ooM56HLS6kABd:-KgtDMijf1qk3-5OnxGE:bmcndcl) ([raw file](https://github.com/netsec-ethz/scion/blob/086315a414432ec8ba4d32ade7d6920f998cdcce/test/integration/base_cli_srv.py#L63)):*

This is a bit too hard-coded for my taste. I would still pass in `api_addr`, and do the calculation in the caller.

(Also it should be `os.path.join(SCIOND_API_SOCKDIR, "sd%s.sock" % addr.isd_as)`)
</blockquote></details>

Made it an optional parameter to the constructor.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1034)*

kormat:
Reviewed 1 of 1 files at r2.
Review status: 2 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[test/integration/base_cli_srv.py, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1034#-Kgse8DDkOgWNgA_Ynkm:-KgtFPadQVSHlvNHMnJz:b2492sx) ([raw file](https://github.com/netsec-ethz/scion/blob/086315a414432ec8ba4d32ade7d6920f998cdcce/test/integration/base_cli_srv.py#L34)):*

Done.
</blockquote></details>

Same applies elsewhere :P

---

*[test/integration/base_cli_srv.py, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1034#-KgsdN_ooM56HLS6kABd:-KgtFeb99zQ4JyZmvQ_x:by29jw0) ([raw file](https://github.com/netsec-ethz/scion/blob/086315a414432ec8ba4d32ade7d6920f998cdcce/test/integration/base_cli_srv.py#L63)):*

Made it an optional parameter to the constructor.
</blockquote></details>

I read that and thought "Ok, he means sometimes a test element won't need access to sciond", but no :P Augh, and this is still hacked in `TestCertClient.__init__`. Please, just calculate the default in a single place, and pass the real api_addr everywhere that needs it.

When i did a quick version of this PR before, i just made a small method for this. Something like:
```
def get_def_api_addr(addr):
    return os.path.join(SCIOND_API_SOCKDIR, "sd%s.sock" % addr.isd_as)
```
That could be imported from `base_cli_srv` by anything else that needs to use it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1034)*

shitz:
Review status: 1 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[test/integration/base_cli_srv.py, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1034#-Kgse8DDkOgWNgA_Ynkm:-Kgtb-WvLE0h_mfMYHZn:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/086315a414432ec8ba4d32ade7d6920f998cdcce/test/integration/base_cli_srv.py#L34)):*

Same applies elsewhere :P
</blockquote></details>

Done.

---

*[test/integration/base_cli_srv.py, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1034#-KgsdN_ooM56HLS6kABd:-KgtbyYnedmWDfmxuPKx:b-9xr9nr) ([raw file](https://github.com/netsec-ethz/scion/blob/086315a414432ec8ba4d32ade7d6920f998cdcce/test/integration/base_cli_srv.py#L63)):*

I read that and thought "Ok, he means sometimes a test element won't need access to sciond", but no :P Augh, and this is still hacked in `TestCertClient.__init__`. Please, just calculate the default in a single place, and pass the real api_addr everywhere that needs it.

When i did a quick version of this PR before, i just made a small method for this. Something like:
```
def get_def_api_addr(addr):
    return os.path.join(SCIOND_API_SOCKDIR, "sd%s.sock" % addr.isd_as)
```
That could be imported from `base_cli_srv` by anything else that needs to use it.
</blockquote></details>

I really don't see why we shouldn't make this optional. All tests so far need the default API address and all tests inherit from `TestBase`. Why not set it there if nothing special is provided? In case some test in the future needs a different api_addr for whatever reason, it can just pass in a different one. This is exactly what default values are for. I did add the `get_sciond_api_addr` method, so it can be used in `cert_req_test` as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1034)*

kormat:
:lgtm:

---

Reviewed 7 of 10 files at r1, 2 of 2 files at r3.

---

*[test/integration/base_cli_srv.py, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1034#-KgsdN_ooM56HLS6kABd:-KgyajhopXyEPEShy-7s:baab1qs) ([raw file](https://github.com/netsec-ethz/scion/blob/086315a414432ec8ba4d32ade7d6920f998cdcce/test/integration/base_cli_srv.py#L63)):*

I really don't see why we shouldn't make this optional. All tests so far need the default API address and all tests inherit from `TestBase`. Why not set it there if nothing special is provided? In case some test in the future needs a different api_addr for whatever reason, it can just pass in a different one. This is exactly what default values are for. I did add the `get_sciond_api_addr` method, so it can be used in `cert_req_test` as well.
</blockquote></details>

You're right; having a utility function to do this is sufficient, and making it optional also makes sense. Thanks :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1034#-:-Kgyaw53otob1QFNn-6-:bnfp4nl)*

kormat:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1034)*

PR #1037, opened 2017-04-04
Locking for router config

shitz:
This PR adds a lock around the router config to prepare for updating topology information without restarting the router.

There now is a package level lock in `conf` and the global configuration object can only be acquired by calling `conf.GetConfig()` which protects access to the configuration object by a RWMutex. When loading a new config, the new config first gets completely loaded before the global reference is updated (protected by a write lock).

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1037)

### DIFF
--- a/go/border/conf/conf.go
+++ b/go/border/conf/conf.go
@@ -36,10 +36,6 @@ import (
 
 // Conf is the main config structure.
 type Conf struct {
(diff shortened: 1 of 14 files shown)


shitz:
Closing, since this won't be checked in.

PR #1039, opened 2017-04-10
Added CN field to CA certificates

pszalach:
Currently, certificate generation does not add the common name field, that PR fixes it.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1039)

### DIFF
--- a/topology/Default.topo
+++ b/topology/Default.topo
@@ -70,46 +70,22 @@ ASes:
 CAs:
   CA1-1:
     ISD: 1
-    countryName: US
-    stateOrProvinceName: Minnesota
-    localityName: Minnetonka
-    organizationName: my company
-    organizationalUnitName: my organization
+    commonName: CA1-1
   CA1-2:
     ISD: 1
-    countryName: US
-    stateOrProvinceName: Minnesota
-    localityName: Minnetonka
-    organizationName: my company
-    organizationalUnitName: my organization
+    commonName: CA1-2
   CA1-3:
     ISD: 1
-    countryName: US
-    stateOrProvinceName: Minnesota
-    localityName: Minnetonka
-    organizationName: my company
-    organizationalUnitName: my organization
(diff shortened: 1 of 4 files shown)


pszalach:
@sezergueler @kormat

sezergueler:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1039#-:-KhMtAswOzXaaOmbsV0-:bnfp4nl)*

pszalach:
The latest commit removes all CA description expect `CN` (as the description is identical everywhere, it is rather confusing than helpful). 
@sezergueler do you use this description anywhere?

sezergueler:
@pszalach I am not using them right now, it is the same everywhere for now but I think in real deployment they would all be different. I think the CAs should have a globally unique name. This could be done with all those fields. We can still drop them, but then we need to make sure that the names are unique.

One could argue that the name contains the ISD identifier and makes the name unique. But we do not have any restrictions on CA names yet, they could have an arbitrary name. I would suggest that if we want to drop the fields, we set CN field to something like "ISD x, ca_name".

pszalach:
@sezergueler but these names are unique. Names must be unique within ISD (i.e., a CA can be included in two ISDs' TRCs under the same name, why not?).
Sure, I agree that in reality they have descriptions etc... but I just want to drop it as it makes debugging more difficult.

kormat:
LGTM, but please update the commit description to cover why the other fields are being removed.

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1039#-:-KinlEs9w3JZy7-6wunL:b-tcc7ma)*

PR #1042, opened 2017-04-10
Timeout IFs if they are down after a reset

shitz:
A BS resets all negative information about interfaces after becoming master, i.e., set their state to `INACTIVE`. If the interface really was down, however, it never timed it out and thus would never issue revocations for that interface. This PR fixes that. 

Also reduced the interface timeout to 3s (down from 10s).

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1042)


shitz:
Superseded by #1050

PR #999, opened 2017-03-02
CMAC instead of CBC-MAC

pszalach:
fixes #985
~~CMAC.py is taken from the current version of pycrypto (unfortunately, currently neither `apt` nor `pip` provide it)~~
fixes #1003 

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/999)

PR #1047, opened 2017-04-20
Cryptography package added

pszalach:
required by #999

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1047)


pszalach:
@kormat

PR #1050, opened 2017-04-24
Fix handling of inactive interfaces in BS.

kormat:
If a BS starts up (or becomes master) when a given interface is down, it
currently stays permenantly marked as INACTIVE, and is never revoked.
This change makes inactive interfaces timeout the same as active ones,
in both the startup and master-election cases.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1050)

### DIFF
--- a/infrastructure/beacon_server/base.py
+++ b/infrastructure/beacon_server/base.py
@@ -761,14 +761,16 @@ class BeaconServer(SCIONElement, metaclass=ABCMeta):
             with self.ifid_state_lock:
                 for (if_id, if_state) in self.ifid_state.items():
                     cur_epoch = ConnectedHashTree.get_current_epoch()
-                    # Check if interface has timed-out.
-                    if ((if_state.is_expired() or if_state.is_revoked()) and
-                       (if_id_last_revoked[if_id] != cur_epoch)):
-                            if_id_last_revoked[if_id] = cur_epoch
-                            if not if_state.is_revoked():
-                                logging.info("IF %d appears to be down.", if_id)
-                            self._issue_revocation(if_id)
-                            if_state.revoke_if_expired()
+                    if not if_state.is_expired() or (
+                            if_state.is_revoked() and if_id_last_revoked[if_id] == cur_epoch):
+                        # Either the interface hasn't timed out, or it's already revoked for this
+                        # epoch
+                        continue
+                    if_id_last_revoked[if_id] = cur_epoch
+                    if not if_state.is_revoked():
+                        logging.info("IF %d appears to be down.", if_id)
+                    self._issue_revocation(if_id)
+                    if_state.revoke_if_expired()
             sleep_interval(start_time, self.IF_TIMEOUT_INTERVAL,
                            "Handle IF timeouts")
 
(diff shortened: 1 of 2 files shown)


shitz:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[infrastructure/beacon_server/if_state.py, line 86 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1050#-KiUFE38kZW6fl6HgnNa:-KiUFE38kZW6fl6HgnNb:bl8yphw) ([raw file](https://github.com/netsec-ethz/scion/blob/e4dba093daf437943636a49066d08e278e3a45e5/infrastructure/beacon_server/if_state.py#L86)):*
> ```Python
>     def is_expired(self):
>         with self._lock:
>             if self._state in [self.TIMED_OUT, self.REVOKED]:
> ```

What's the reason for adding `REVOKED` here as well? Is it intentional that an interface can now be `EXPIRED` and `REVOKED`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1050)*

kormat:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[infrastructure/beacon_server/if_state.py, line 86 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1050#-KiUFE38kZW6fl6HgnNa:-KiUIBldYr3QcvD7AYp2:btef8kv) ([raw file](https://github.com/netsec-ethz/scion/blob/e4dba093daf437943636a49066d08e278e3a45e5/infrastructure/beacon_server/if_state.py#L86)):*

What's the reason for adding `REVOKED` here as well? Is it intentional that an interface can now be `EXPIRED` and `REVOKED`?
</blockquote></details>

Well, part of the issue here is that the semantics of the states aren't defined, `is_expired` doesn't actually map to a single state (there is no `EXPIRED` state), and the semantics of "expired" isn't defined either. I'm considering `REVOKED` to be an extension of `TIMED_OUT`. This also simplifies the logic in the BS.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1050)*

shitz:
:lgtm:

---

Reviewed 1 of 2 files at r1.

---

*[infrastructure/beacon_server/if_state.py, line 86 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1050#-KiUFE38kZW6fl6HgnNa:-KiUK-fAHO-olXv5uasX:bg2m6kk) ([raw file](https://github.com/netsec-ethz/scion/blob/e4dba093daf437943636a49066d08e278e3a45e5/infrastructure/beacon_server/if_state.py#L86)):*

Well, part of the issue here is that the semantics of the states aren't defined, `is_expired` doesn't actually map to a single state (there is no `EXPIRED` state), and the semantics of "expired" isn't defined either. I'm considering `REVOKED` to be an extension of `TIMED_OUT`. This also simplifies the logic in the BS.
</blockquote></details>

`is_expired` is just a bad naming of `is_timed_out`. I'm fine with considering `REVOKED` as being an extensions of `TIMED_OUT`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1050#-:-KiUKHjRXQFBcZXaXz9a:bnfp4nl)*

Issue #983, opened 2017-02-28
Dynamic reloading of topology & config formats


kormat:
Currently configs are read only on startup, and processes need to restart to pick up config changes. This needs to be fixed.

kormat:
The BR supports this, but we're not going to implement this for any of the existing python code. All new Go services need to support this.

scrye:
Go services now support reloading for most basic fields. We'll add support for more as use cases arise.

PR #1053, opened 2017-04-25
Dynamically reloading config in the Router

shitz:
This PR adds the capability of dynamically reloading the router configuration, i.e., without restarting the router (#983) . 
There exists now a router context `rctx.RtrCtx` which contains the current configuration as well as all the I/O routines for each socket. Instead of access the global configuration, rpkts now get passed in a snapshot of the current router context that is then used to process the entire packet.
When the router configuration changes, a new context object is first set up completely, before it replaces the current context. If any of the interfaces (internal/external) or local addresses change new input goroutines are started and old ones removed.

Things that are tested:
* Normal operation with static configuration
* Reloading the same configuration
* Changing the router's local address/port
* Changing a router's own end of an interface (address/port)
* Changing a router's far end of an interface (address/port)
* HSR

Things that are untested:
* Adding a new interface
* Removing an interface

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1053)


klausman:
Only one small clarification, otherwise :lgtm: (please also await review by @kormat )

---

Reviewed 25 of 25 files at r1.

---

*[go/border/io.go, line 116 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KidGcjMFt5cmVX3kYAY:-KidGcjMFt5cmVX3kYAZ:bw2fe9w) ([raw file](https://github.com/netsec-ethz/scion/blob/37380d5a199eb1e5263b4077cc1a11cd95f330cc/go/border/io.go#L116)):*
> ```Go
> 			metrics.PktsRecv.With(args.Labels).Inc()
> 			metrics.BytesRecv.With(args.Labels).Add(float64(length))
> 			// TODO(kormat): experiment with performance by calling processPacket directly instead.
> ```

Is this TODO still valid?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-:-KidKaLCmkH8Aki4mx_A:b-ly3o89)*

kormat:
Reviewed 8 of 25 files at r1.
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[go/border/main.go, line 61 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KipAPlKvFUqIcpNBByR:-KipAPlKvFUqIcpNBByS:b-mblb06) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/main.go#L61)):*
> ```Go
> 	if err := r.Run(); err != nil {
> 		log.Crit("Run failed", err.Ctx...)
> 		liblog.Flush()
> ```

Hmm. In that case, add this to the `"Startup Failed"` case as well.

---

*[go/border/router.go, line 61 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KipGssuOTrAadq1oNLO:-KipGssuOTrAadq1oNLP:bwzljtt) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/router.go#L61)):*
> ```Go
> 	// TODO(shitz): Here should be some code to periodically check the discovery
> 	// service for updated info.
> 	for {
> ```

This can just `return nil` instead of looping.

---

*[go/border/metrics/metrics.go, line 149 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip6UPjHMSs6AVQg8YJ:-Kip6UPjHMSs6AVQg8YK:bxwsn6c) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/metrics/metrics.go#L149)):*
> ```Go
> 	for _, addr := range addresses {
> 		if _, ok := servers[addr]; !ok {
> 			log.Debug("Exporting metrics", "addr", addr)
> ```

I'd make this `log.Info`, same for the stopping below.

---

*[go/border/metrics/metrics.go, line 163 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip6eF-zuVMNXIqeRAv:-Kip6eF-zuVMNXIqeRAw:b-1v5mmr) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/metrics/metrics.go#L163)):*
> ```Go
> 		if !contains(addresses, addr) {
> 			log.Debug("Stop exporting metrics", "addr", addr)
> 			closer.Close()
> ```

The entry isn't removed from the map. Suggested refactoring of this function:
```
new_servers := make(map[string]io.Closer)
for _, addr := range addresses {
   if _, ok := servers[addr]; ok {
       new_servers[addr] = servers[addr]
       continue // Exporter already running on this address
   }
   ... // run new exporter
   new_servers[addr] = closer
}
for addr, closer := range servers {
    ... // close unconditionally
}
servers = new_servers
```

---

*[go/border/metrics/metrics.go, line 189 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip5s4xTl9Snzth_jSW:-Kip5s4xTl9Snzth_jSX:b-yea9kb) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/metrics/metrics.go#L189)):*
> ```Go
> 	go func() {
> 		err := srv.Serve(listener)
> 		if err != nil {
> ```

If the listener is closed, `err` will be `http.ErrServerClosed`, which wouldn't be an error in this case. I suggest adding an info log entry for that case, saying the http server was shut down on the given address.

---

*[go/border/rctx/rctx.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kiomx8WkbBfCME2K-gQ:-Kiomx8WkbBfCME2K-gR:bq1udkl) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L35)):*
> ```Go
> 	Bytes() common.RawBytes
> 	// LogError can be used log errors during output.
> 	LogError(msg string, ctx ...interface{})
> ```

Why not just `Error`? That way rpkt would already satisfy it.

---

*[go/border/rctx/rctx.go, line 41 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kiony4KHEFcjPBtdhEK:-Kiony4KHEFcjPBtdhEL:b-i0k81y) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L41)):*
> ```Go
> type OutputFunc func(OutputObj, *net.UDPAddr)
> 
> // InputFunc defines an interface for starting and stopping input goroutines.
> ```

Naming suggestion: `IOCtrl`. Output funcs may well be goroutines at some point too.

---

*[go/border/rctx/rctx.go, line 48 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KionEFJJCpSw6AEpXRa:-KionEFJJCpSw6AEpXRb:bcen8su) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L48)):*
> ```Go
> 
> // RtrCtx is the main context structure.
> type RtrCtx struct {
> ```

`Ctx` would be sufficient (you then get `rctx.Ctx` vs `rctx.RtrCtx`, which is shorter, and sufficiently informative).

---

*[go/border/rctx/rctx.go, line 59 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KioncxK9XEP1o5td6cI:-KioncxK9XEP1o5td6cJ:b-solksz) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L59)):*
> ```Go
> 	IntfOutFs map[spath.IntfID]OutputFunc
> 	// InputFuncs is a slice of InputFunc objects to stop the corresponding input goroutines.
> 	InputFuncs map[string]InputFunc
> ```

I'm not sure this should be `string`. Or, rather, i'm not sure it's ok to have a single map for both internal and external input goroutines. It looks like you're using the string representation of the bind address as the key, which could be the same for multiple input goroutines. (E.g. the BR has the same internal and external address).

---

*[go/border/rctx/rctx.go, line 62 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KiooPKhGlJ__o3tKApB:-KiooPKis0pWBnF7dOc7:b-h3slm4) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L62)):*
> ```Go
> }
> 
> func NewContext(conf *conf.Conf) *RtrCtx {
> ```

As `Ctx` is the main class of this package, i would name these funcs `New`, `Get`, `Set`.

---

*[go/border/rctx/rctx.go, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KiooZRefUtg7ItCOdQG:-KiooZRefUtg7ItCOdQH:b65dxu5) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L63)):*
> ```Go
> 
> func NewContext(conf *conf.Conf) *RtrCtx {
> 	ctx := &RtrCtx{
> ```

`return &Ctrx{`

---

*[go/border/rctx/rctx.go, line 89 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip-QsvtwIfI-3l9OAr:-Kip-Qswnvrd_iwKjz3j:boscrqi) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L89)):*
> ```Go
> 	ctxLock.Lock()
> 	ctx = newCtx
> 	ctxLock.Unlock()
> ```

I would use `defer` here as well for consistency with `GetContext`

---

*[go/border/rctx/rctx.go, line 95 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip-eW07aW4pIK0va3P:-Kip-eW07aW4pIK0va3Q:b-3cor6e) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L95)):*
> ```Go
> // locks, since the ISD-AS cannot change dynamically.
> func IA() *addr.ISD_AS {
> 	return ctx.Conf.IA
> ```

This is not safe. https://golang.org/ref/mem

---

*[go/lib/overlay/udp.go, line 68 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip9_My_j_mvW6D-xFA:-Kip9_My_j_mvW6D-xFB:b-6i4mrn) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/lib/overlay/udp.go#L68)):*
> ```Go
> 
> func (u *UDP) String() string {
> 	return u.BindAddr().String()
> ```

`String` seems like it should print the full information of the struct. It's generally intended to produce human-useful output.

---

*[go/lib/overlay/udp.go, line 73 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip9ozvwoPo2Dk4U6V2:-Kip9ozvwoPo2Dk4U6V3:btyaioq) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/lib/overlay/udp.go#L73)):*
> ```Go
> func (u *UDP) Equal(other *UDP) bool {
> 	return (u.PublicAddr().String() == other.PublicAddr().String() &&
> 		u.BindAddr().String() == other.BindAddr().String())
> ```

Mmph. Not a big fan of string-conversion-then-comparison.
```
return (u.pubIP.Equal(other.pubIP) && u.pubPort == other.pubPort && u.bindIP.Equal(other.bindIP) && u.bindPort == other.bindPort)
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

shitz:
Review status: 10 of 25 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[go/border/io.go, line 116 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KidGcjMFt5cmVX3kYAY:-KjCH-OBIBbZ9mtPcazi:bs1sy0u) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/io.go#L116)):*

Is this TODO still valid?
</blockquote></details>

Still valid, since we haven't done any benchmarking of this approach.

---

*[go/border/main.go, line 61 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KipAPlKvFUqIcpNBByR:-KjCHUfTHFLcOsD4yGnt:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/main.go#L61)):*

Hmm. In that case, add this to the `"Startup Failed"` case as well.
</blockquote></details>

Done.

---

*[go/border/router.go, line 61 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KipGssuOTrAadq1oNLO:-KjCHrnA1KNwGqSRvSCu:b-imizrg) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/router.go#L61)):*

This can just `return nil` instead of looping.
</blockquote></details>

 Exiting the main function quits all threads and goroutines (see http://stackoverflow.com/questions/37142494/make-goroutines-stay-running-after-returning-from-function#comment61824560_37143130). Replaced loop with reading from a null channel.

---

*[go/border/metrics/metrics.go, line 149 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip6UPjHMSs6AVQg8YJ:-KjCI2U4lvrQ12YAarHs:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/metrics/metrics.go#L149)):*

I'd make this `log.Info`, same for the stopping below.
</blockquote></details>

Done.

---

*[go/border/metrics/metrics.go, line 163 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip6eF-zuVMNXIqeRAv:-KjCNT_PdteoG3qlzF-e:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/metrics/metrics.go#L163)):*

The entry isn't removed from the map. Suggested refactoring of this function:
```
new_servers := make(map[string]io.Closer)
for _, addr := range addresses {
   if _, ok := servers[addr]; ok {
       new_servers[addr] = servers[addr]
       continue // Exporter already running on this address
   }
   ... // run new exporter
   new_servers[addr] = closer
}
for addr, closer := range servers {
    ... // close unconditionally
}
servers = new_servers
```
</blockquote></details>

Done.

---

*[go/border/metrics/metrics.go, line 189 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip5s4xTl9Snzth_jSW:-KjCNz_KzwzvL0kuS3u4:b-tw27ru) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/metrics/metrics.go#L189)):*

If the listener is closed, `err` will be `http.ErrServerClosed`, which wouldn't be an error in this case. I suggest adding an info log entry for that case, saying the http server was shut down on the given address.
</blockquote></details>

I think this only happens in Go1.8, where `http.Server` has explicit `Close()` and `Shutdown()` functions. All this code is only needed because we are not using Go1.8 yet. Afaics, `Serve()` returns the error the listener returns when closing, which is some generic `net.OpError`.

---

*[go/border/rctx/rctx.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kiomx8WkbBfCME2K-gQ:-KjCPWwXx5kRSjd7Uu7S:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L35)):*

Why not just `Error`? That way rpkt would already satisfy it.
</blockquote></details>

Done.

---

*[go/border/rctx/rctx.go, line 41 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kiony4KHEFcjPBtdhEK:-KjCPxwso7FnTq9dbGX6:b-ol9l4l) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L41)):*

Naming suggestion: `IOCtrl`. Output funcs may well be goroutines at some point too.
</blockquote></details>

Nice! Done.

---

*[go/border/rctx/rctx.go, line 48 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KionEFJJCpSw6AEpXRa:-KjCQTs1NxURhxJ413eT:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L48)):*

`Ctx` would be sufficient (you then get `rctx.Ctx` vs `rctx.RtrCtx`, which is shorter, and sufficiently informative).
</blockquote></details>

Done.

---

*[go/border/rctx/rctx.go, line 59 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KioncxK9XEP1o5td6cI:-KjCe3k5RAIvHzjmh9zE:bfr29e5) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L59)):*

I'm not sure this should be `string`. Or, rather, i'm not sure it's ok to have a single map for both internal and external input goroutines. It looks like you're using the string representation of the bind address as the key, which could be the same for multiple input goroutines. (E.g. the BR has the same internal and external address).
</blockquote></details>

Split them up in two different maps.

---

*[go/border/rctx/rctx.go, line 62 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KiooPKhGlJ__o3tKApB:-KjCTmwQaXXLzlbIxLMP:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L62)):*

As `Ctx` is the main class of this package, i would name these funcs `New`, `Get`, `Set`.
</blockquote></details>

Done.

---

*[go/border/rctx/rctx.go, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KiooZRefUtg7ItCOdQG:-KjCTr4kW5XQlyRJqC9D:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L63)):*

`return &Ctrx{`
</blockquote></details>

Done.

---

*[go/border/rctx/rctx.go, line 89 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip-QsvtwIfI-3l9OAr:-KjCTxefI6sJEiTDylP2:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L89)):*

I would use `defer` here as well for consistency with `GetContext`
</blockquote></details>

Done.

---

*[go/border/rctx/rctx.go, line 95 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip-eW07aW4pIK0va3P:-KjCVE6455GDbwbwo0b1:ba5t3d3) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L95)):*

This is not safe. https://golang.org/ref/mem
</blockquote></details>

Afaict, this is only unsafe if it gets called before a `ctx` has been initialized (but then a lot of the code would be unsafe, since nothing checks for a `nil` context). Since we assume that `ctx.Conf.IA` can never change it doesn't matter if the version that gets returned is from an old config (in case the write to `ctx` happens to interleave the call to `IA()`). We can discuss if this shortcut is really necessary for performance reasons though.

---

*[go/lib/overlay/udp.go, line 68 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip9_My_j_mvW6D-xFA:-KjCe7aE9GZVR65CUkIq:bnhm957) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/lib/overlay/udp.go#L68)):*

`String` seems like it should print the full information of the struct. It's generally intended to produce human-useful output.
</blockquote></details>

Removed entirely.

---

*[go/lib/overlay/udp.go, line 73 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip9ozvwoPo2Dk4U6V2:-KjCX42q5G2OaCl2btQw:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/lib/overlay/udp.go#L73)):*

Mmph. Not a big fan of string-conversion-then-comparison.
```
return (u.pubIP.Equal(other.pubIP) && u.pubPort == other.pubPort && u.bindIP.Equal(other.bindIP) && u.bindPort == other.bindPort)
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

kormat:
Reviewed 4 of 15 files at r2.
Review status: 9 of 26 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[go/border/router.go, line 59 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kjr2KMynUfeBeCCcKKC:-Kjr2KMynUfeBeCCcKKD:b-r3u0wm) ([raw file](https://github.com/netsec-ethz/scion/blob/38fd3eee5f913487f61e06119452c633098f3160/go/border/router.go#L59)):*
> ```Go
> 	// TODO(shitz): Here should be some code to periodically check the discovery
> 	// service for updated info.
> 	done := make(chan struct{})
> ```

`var wait chan struct{}`, no need to use `make`, we want a nil channel.

---

*[go/border/metrics/metrics.go, line 150 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQrzoDk4f_FlgjFNbu:-KkQrzoEHj1OVbR2P5WJ:bkq15dg) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/metrics/metrics.go#L150)):*
> ```Go
> 	for _, addr := range addresses {
> 		if _, ok := servers[addr]; ok {
> 			newServers[addr] = servers[addr]
> ```

Oops, i think there's a `del servers[addr]` missing after this line, otherwise i think all existing servers get stopped regardless.

---

*[go/border/rctx/rctx.go, line 95 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip-eW07aW4pIK0va3P:-KkQsqPS6cjV2i47Pddc:b-wcaq8g) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L95)):*
The memory model is quite explicit: 
> "Programs that modify data being simultaneously accessed by multiple goroutines must serialize such access."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

kormat:
Reviewed 1 of 15 files at r3.
Review status: 10 of 26 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[go/border/setup.go, line 82 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQxC8sMd9yKG21ex22:-KkQxC8sMd9yKG21ex23:b-gyb36l) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup.go#L82)):*
> ```Go
> }
> 
> // loadCapabilities drops and immediately reloads the capabilities on startup.
> ```

`// clearCapabilities drops unnecessary capabilities after startup`
(name change implied)

---

*[go/border/setup.go, line 118 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQxuW4032_TQGyD_NW:-KkQxuW4032_TQGyD_NX:b-nis41r) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup.go#L118)):*
> ```Go
> 	// Load capabilities
> 	if loadCaps {
> 		if err := r.loadCapabilities(); err != nil {
> ```

As clearing the capabilities will only happen once, at startup, i'd move this to the end of `setup()`.

---

*[go/border/setup.go, line 134 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQyIdthQTYYwPorRSE:-KkQyIdthQTYYwPorRSF:bhx8mt1) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup.go#L134)):*
> ```Go
> 
> // setupNet configures networking for the router, using any setup hooks that
> // have been registered.
> ```

Maybe add a comment about how if an oldCtx is supplied, then it reconfigures networking 

---

*[go/border/setup.go, line 198 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQyiiZPTR8KHXI2TbD:-KkQyiiZPTR8KHXI2TbE:b-lveztr) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup.go#L198)):*
> ```Go
> 	// Stop input functions that are no longer needed.
> 	if oldCtx != nil {
> 		for k, pif := range oldCtx.LocInputFs {
> ```

Why `pif`? My best guess is that it means "posixInputFunction"? If so that wouldn't be accurate, as there will be non-posix functions in here.`f` would be fine.

---

*[go/border/rctx/rctx.go, line 71 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQzrEsEck7_9BtVR1b:-KkQzrEtI3mSuN0AugDv:bkz1u9v) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/rctx/rctx.go#L71)):*
> ```Go
> 		LocOutFs:   make(map[int]OutputFunc),
> 		IntfOutFs:  make(map[spath.IntfID]OutputFunc),
> 		LocInputFs: make(map[string]IOCtrl),
> ```

Why is this a string and not an index?

---

*[lib/libhsr/hsr_dpdk.c, line 374 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQtxJ4LqEHrHylFp29:-KkQtxJ4LqEHrHylFp2A:b-qdi1sv) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/lib/libhsr/hsr_dpdk.c#L374)):*
> ```C
>         return 1;
>     }
>     for(int i=0; i<argc; ++i) {
> ```

Is this still needed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

kormat:
Reviewed 1 of 15 files at r2, 1 of 15 files at r3.
Review status: 12 of 26 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[go/border/io.go, line 37 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRAvtquJLAZS1Lpmfb:-KkRAvtquJLAZS1Lpmfc:bgnbu0z) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/io.go#L37)):*
> ```Go
> type PosixInputFuncArgs struct {
> 	// Router is a reference to the main router object.
> 	Router *Router
> ```

As far as i can tell, this is only used to call `processPacket` on the router instance? In that case, i'd prefer to just have a function pointer here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

kormat:
Reviewed 3 of 15 files at r2, 6 of 15 files at r3.
Review status: 21 of 26 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[go/border/error.go, line 126 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkREhDTZD_y9IeEMWrh:-KkREhDTZD_y9IeEMWri:b-l421wc) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/error.go#L126)):*
> ```Go
> 	}
> 	// Only (potentially) call IncPath if the dest is not in the local AS.
> 	if !dstIA.Eq(rctx.IA()) {
> ```

This can just access `rp.Ctx.Conf.IA`.

---

*[go/border/error.go, line 176 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkREyhlhMI0_43Zjigc:-KkREyhlhMI0_43Zjigd:bl8e0ai) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/error.go#L176)):*
> ```Go
> 	}
> 	// Use the ingress address as the source host
> 	sp.SrcIA = rctx.IA()
> ```

`rp.Ctx.Conf.IA`

---

*[go/border/error.go, line 184 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRF0wHzXp9Hl9-VEFd:-KkRF0wHzXp9Hl9-VEFe:biy1ras) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/error.go#L184)):*
> ```Go
> // address to use when replying to a packet.
> func (r *Router) replyEgress(rp *rpkt.RtrPkt) (rpkt.EgressPair, *common.Error) {
> 	ctx := rctx.Get()
> ```

`rp.Ctx.Conf.IA`.

---

*[go/border/gen.go, line 50 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRF5bAgoPZjxU979w3:-KkRF5bAgoPZjxU979w4:b9exa9a) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/gen.go#L50)):*
> ```Go
> 	ctx := rctx.Get()
> 	dirTo := rpkt.DirExternal
> 	if dstIA.Eq(rctx.IA()) {
> ```

`ctx.Conf.IA`.

---

*[go/border/gen.go, line 55 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRF8r49-Wzy0dH0sgL:-KkRF8r49-Wzy0dH0sgM:bgj96ly) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/gen.go#L55)):*
> ```Go
> 	// Create base packet
> 	rp, err := rpkt.RtrPktFromScnPkt(&spkt.ScnPkt{
> 		DstIA: dstIA, SrcIA: rctx.IA(), DstHost: dstHost, SrcHost: addr.HostFromIP(srcAddr.IP),
> ```

`ctx.Conf.IA`.
etc.

---

*[go/border/setup-hsr.go, line 56 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRDfFGmD4lhECP6i3s:-KkRDfFGmD4lhECP6i3t:b-jgvgq9) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup-hsr.go#L56)):*
> ```Go
> }
> 
> func setupHSRNetStart(r *Router, ctx *rctx.Ctx, _ *rctx.Ctx) (rpkt.HookResult, *common.Error) {
> ```

As `ctx` isn't used here either, it can also just be `_`.

---

*[go/border/setup-hsr.go, line 128 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkREIAF59xcOEpPPNXx:-KkREIAF59xcOEpPPNXy:bfmdwlp) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup-hsr.go#L128)):*
> ```Go
> 		return rpkt.HookError, err
> 	}
> 	hi := &HSRInput{
> ```

`ctx.LocInputFs["hsr"] = &HSRInput{`

---

*[go/border/ifstate/ifstate.go, line 79 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRFPAd_M9hOejiTPs-:-KkRFPAd_M9hOejiTPs0:b3519sy) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/ifstate/ifstate.go#L79)):*
> ```Go
> 	ctx := rctx.Get()
> 	ctx.Conf.IFStates.Lock()
> 	defer ctx.Conf.IFStates.Unlock()
> ```

Hmmm. If `ctx` changes during this time, the update will be lost.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

kormat:
Reviewed 2 of 25 files at r1, 5 of 15 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

shitz:

---

*[go/border/error.go, line 126 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkREhDTZD_y9IeEMWrh:-KkUgo1jXpvNFQFNeEvT:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/error.go#L126)):*

This can just access `rp.Ctx.Conf.IA`.
</blockquote></details>

Done.

---

*[go/border/error.go, line 176 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkREyhlhMI0_43Zjigc:-KkUgqZW90pJFaGbcBfu:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/error.go#L176)):*

`rp.Ctx.Conf.IA`
</blockquote></details>

Done.

---

*[go/border/error.go, line 184 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRF0wHzXp9Hl9-VEFd:-KkUgr6X0jjwWD5YPwqw:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/error.go#L184)):*

`rp.Ctx.Conf.IA`.
</blockquote></details>

Done.

---

*[go/border/gen.go, line 50 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRF5bAgoPZjxU979w3:-KkUh--dpCDWZGvbSAmS:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/gen.go#L50)):*

`ctx.Conf.IA`.
</blockquote></details>

Done.

---

*[go/border/gen.go, line 55 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRF8r49-Wzy0dH0sgL:-KkUh-SpPn5yk9HbjjqB:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/gen.go#L55)):*

`ctx.Conf.IA`.
etc.
</blockquote></details>

Done.

---

*[go/border/io.go, line 37 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRAvtquJLAZS1Lpmfb:-KkUi68ZJouU3ZS6g3lC:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/io.go#L37)):*

As far as i can tell, this is only used to call `processPacket` on the router instance? In that case, i'd prefer to just have a function pointer here.
</blockquote></details>

Done.

---

*[go/border/router.go, line 59 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kjr2KMynUfeBeCCcKKC:-KkUiYc4JRI0k_eNroCs:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/38fd3eee5f913487f61e06119452c633098f3160/go/border/router.go#L59)):*

`var wait chan struct{}`, no need to use `make`, we want a nil channel.
</blockquote></details>

Done.

---

*[go/border/setup-hsr.go, line 56 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRDfFGmD4lhECP6i3s:-KkUicYP65K2ozEtH9oA:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup-hsr.go#L56)):*

As `ctx` isn't used here either, it can also just be `_`.
</blockquote></details>

Done.

---

*[go/border/setup-hsr.go, line 128 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkREIAF59xcOEpPPNXx:-KkUinsuC31lEq45NwaV:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup-hsr.go#L128)):*

`ctx.LocInputFs["hsr"] = &HSRInput{`
</blockquote></details>

Done.

---

*[go/border/setup.go, line 82 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQxC8sMd9yKG21ex22:-KkUrNDRIQ_gNtydwTpP:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup.go#L82)):*

`// clearCapabilities drops unnecessary capabilities after startup`
(name change implied)
</blockquote></details>

Done.

---

*[go/border/setup.go, line 118 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQxuW4032_TQGyD_NW:-KkUrM5A-z1muPiYz3_U:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup.go#L118)):*

As clearing the capabilities will only happen once, at startup, i'd move this to the end of `setup()`.
</blockquote></details>

Done.

---

*[go/border/setup.go, line 134 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQyIdthQTYYwPorRSE:-KkUkphsr__P0gl0NrN2:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup.go#L134)):*

Maybe add a comment about how if an oldCtx is supplied, then it reconfigures networking 
</blockquote></details>

Done.

---

*[go/border/setup.go, line 198 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQyiiZPTR8KHXI2TbD:-KkUkzPPkSqr-u8yXIvP:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/setup.go#L198)):*

Why `pif`? My best guess is that it means "posixInputFunction"? If so that wouldn't be accurate, as there will be non-posix functions in here.`f` would be fine.
</blockquote></details>

Done.

---

*[go/border/ifstate/ifstate.go, line 79 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkRFPAd_M9hOejiTPs-:-KkUlDsx3UOs9jTw-c9e:biac3n) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/ifstate/ifstate.go#L79)):*

Hmmm. If `ctx` changes during this time, the update will be lost.
</blockquote></details>

That's a very good catch. The problem is actually deeper: the `IFStates` weren't copied to the new context from the old one. I moved `IFStates` out of the context and made it a globally accessible struct.

---

*[go/border/metrics/metrics.go, line 150 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQrzoDk4f_FlgjFNbu:-KkUzYT6dmAQwAnkGmrX:b-5w9wtm) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/metrics/metrics.go#L150)):*

Oops, i think there's a `del servers[addr]` missing after this line, otherwise i think all existing servers get stopped regardless.
</blockquote></details>

I don't think so. `if !contains(addresses, addr)` won't evaluate to `true` for these addresses, since here we iterate over all addresses and copy the exporter out of `servers` if an entry already exists. 

---

*[go/border/rctx/rctx.go, line 95 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip-eW07aW4pIK0va3P:-KkV0WqOMU6Yu5NrW26L:bzh4a3b) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L95)):*

The memory model is quite explicit: 
> "Programs that modify data being simultaneously accessed by multiple goroutines must serialize such access."
</blockquote></details>

I removed it entirely, since all the places where `IA()` is called also need the context.

---

*[go/border/rctx/rctx.go, line 71 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQzrEsEck7_9BtVR1b:-KkV-G6eGjMBECxxumur:b-o0ycix) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/rctx/rctx.go#L71)):*

Why is this a string and not an index?
</blockquote></details>

I use the string of the `BindAddr` of the corresponding socket as a key, since there needs to be one input function per (IP, port) pair. I don't immediately see another possibility for an index, at least not one that would need an additional (IP, port) -> index mapping.

---

*[lib/libhsr/hsr_dpdk.c, line 374 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQtxJ4LqEHrHylFp29:-KkV176ovMn1zHL5OkiP:bwtjl24) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/lib/libhsr/hsr_dpdk.c#L374)):*

Is this still needed?
</blockquote></details>

Wait what? This part was already removed on my branch. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

kormat:
:lgtm:

---

Reviewed 13 of 13 files at r4.

---

*[go/border/rctx/rctx.go, line 95 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kip-eW07aW4pIK0va3P:-KkVJEYZO8F0l-MpuSAN:bmi8ppe) ([raw file](https://github.com/netsec-ethz/scion/blob/14ac7b4ed4bb0556b426f028ae0afdbffc9f5c14/go/border/rctx/rctx.go#L95)):*

I removed it entirely, since all the places where `IA()` is called also need the context.
</blockquote></details>

:100: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-:-KkVZR1h0c-MgpNAPc3H:bnfp4nl)*

shitz:

---

*[go/border/metrics/metrics.go, line 150 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQrzoDk4f_FlgjFNbu:-KkWY_hEeu9y01WmbECM:b-kcgaxy) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/metrics/metrics.go#L150)):*

I don't think so. `if !contains(addresses, addr)` won't evaluate to `true` for these addresses, since here we iterate over all addresses and copy the exporter out of `servers` if an entry already exists. 
</blockquote></details>

Changed it like you suggested. More elegant that way.

---

*[go/border/rctx/rctx.go, line 71 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-KkQzrEsEck7_9BtVR1b:-KkWa_tfNuPNhtxS_29w:bwh25yh) ([raw file](https://github.com/netsec-ethz/scion/blob/62dc3cd14c5b2d25eb023392253e90995063ae6d/go/border/rctx/rctx.go#L71)):*

I use the string of the `BindAddr` of the corresponding socket as a key, since there needs to be one input function per (IP, port) pair. I don't immediately see another possibility for an index, at least not one that would need an additional (IP, port) -> index mapping.
</blockquote></details>

Changed to to use `localAddrIdx`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

kormat:
Reviewed 4 of 4 files at r5.

---

*[go/border/setup-hsr.go, line 118 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kkj7KF6h1JstRZ32qCB:-Kkj7KF6h1JstRZ32qCC:btj3ucw) ([raw file](https://github.com/netsec-ethz/scion/blob/0c04f56edd19248781c38bd634921cf5319bf5ca/go/border/setup-hsr.go#L118)):*
> ```Go
> 	}
> 	if oldCtx != nil {
> 		if f, ok := oldCtx.LocInputFs[0]; ok {
> ```

What is the purpose of this (and why is `0` special)? This looks like it should be handled by `setupHSRAddLocal`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

shitz:
Review status: 23 of 26 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[go/border/setup-hsr.go, line 118 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-Kkj7KF6h1JstRZ32qCB:-KkjLPTmJDB39oRTlOBK:bcr2p1t) ([raw file](https://github.com/netsec-ethz/scion/blob/0c04f56edd19248781c38bd634921cf5319bf5ca/go/border/setup-hsr.go#L118)):*

What is the purpose of this (and why is `0` special)? This looks like it should be handled by `setupHSRAddLocal`.
</blockquote></details>

Good catch. Defined a constant for the HSR input index, which should never be overwritten by a Posix input function. Having it in `setupHSRNetFinish` seems ok to be, since there can only be one HSR input function ever and that one is not dependent on any of the parameters passed to `setupHSRAddLocal` (at least as long as `lib/hsr` doesn't change).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

kormat:
Reviewed 3 of 3 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053)*

kormat:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1053#-:-KkjjrWfnoFidlP1VUZb:bnfp4nl)*

PR #1064, opened 2017-05-08
Warnings in PSes removed

sezergueler:
These warnings are not valid anymore. 
Before, the path_resolution was called only if the path could be resolved. Now, the path_resolution is called for every request. Therefore, the warnings are printed very often.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1064)


kormat:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1064#-:-KkAwj2Va81UNSCCa_Wi:bnfp4nl)*

shitz:
The warnings in question have been removed by #1076.

PR #952, opened 2016-12-14
Nonblocking send_msg() with TCP and other fixes

pszalach:
- nonblocking send_msg() with TCP (see #944)
- fix in PS with prematurely closed connections
- less annoying logging
- sub/lwip-contrib updated

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/952)

PR #1067, opened 2017-05-09
Few prerequisites for #952

pszalach:
This PR:
- increases limit for open files in dispatcher. Dispatcher opens many files and with TCP (during a heavy load) it goes beyond the default limit (1024)
- introduces `_build_meta()` for creating UDP/TCP meta objects
- introduces the `reuse` flag, that allows to indicate that a given msg should be sent via a shared (TCP) socket

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1067)

### DIFF
--- a/endhost/dispatcher.c
+++ b/endhost/dispatcher.c
@@ -13,6 +13,7 @@
 #include <sys/resource.h>
 #include <sys/select.h>
 #include <sys/socket.h>
+#include <sys/time.h>
 #include <sys/types.h>
 #include <sys/uio.h>
 #include <sys/un.h>
@@ -145,15 +146,7 @@ int main(int argc, char **argv)
     signal(SIGINT, handle_signal);
     signal(SIGPIPE, handle_signal);
 
-    struct rlimit rl;
     int res;
-
-    rl.rlim_cur = MAX_SOCKETS;
-    rl.rlim_max = MAX_SOCKETS;
-    if (setrlimit(RLIMIT_NOFILE, &rl)< 0) {
-        fprintf(stderr, "failed to set fileno limit\n");
-        return -1;
-    }
 
     setenv("TZ", "UTC", 1);
 
@@ -168,6 +161,19 @@ int main(int argc, char **argv)
     if (!zc) {
         fprintf(stderr, "failed to get dispatcher zlog category\n");
         zlog_fini();
+        return -1;
+    }
+
+    /* TCPMW uses many open files, set limit as high as possible */
+    struct rlimit rlim;
+    if (getrlimit(RLIMIT_NOFILE, &rlim) < 0){
+        zlog_fatal(zc, "getrlimit(): %s", strerror(errno));
+        return -1;
+    }
+    zlog_debug(zc, "Changing RLIMIT_NOFILE %lu -> %lu", rlim.rlim_cur, rlim.rlim_max);
+    rlim.rlim_cur = rlim.rlim_max;
+    if (setrlimit(RLIMIT_NOFILE, &rlim) < 0){
+        zlog_fatal(zc, "setrlimit(): %s", strerror(errno));
         return -1;
     }
 
--- a/endhost/sciond.py
+++ b/endhost/sciond.py
@@ -577,7 +577,7 @@ class SCIONDaemon(SCIONElement):
             return
         req = PathSegmentReq.from_values(self.addr.isd_as, dst_ia, flags=flags)
         logging.debug("Sending path request (%s) to [%s]:%s", req.short_desc(), addr, port)
-        meta = self.DefaultMeta.from_values(host=addr, port=port)
+        meta = self._build_meta(host=addr, port=port)
         self.send_meta(req, meta)
 
     def _calc_core_segs(self, dst_isd, up_segs, down_segs):
--- a/infrastructure/beacon_server/base.py
+++ b/infrastructure/beacon_server/base.py
@@ -47,10 +47,8 @@ from lib.errors import (
     SCIONParseError,
     SCIONServiceLookupError,
 )
-from lib.flagtypes import TCPFlags
-from lib.msg_meta import TCPMetadata, UDPMetadata
+from lib.msg_meta import UDPMetadata
 from lib.path_seg_meta import PathSegMeta
-from lib.packet.ext.one_hop_path import OneHopPathExt
 from lib.packet.opaque_field import HopOpaqueField, InfoOpaqueField
 from lib.packet.path import SCIONPath
 from lib.packet.path_mgmt.ifstate import (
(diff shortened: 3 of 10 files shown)


kormat:
Reviewed 10 of 10 files at r1.

---

*[lib/msg_meta.py, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1067#-KjqgM0paECbJpfR-X4f:-KjqgM0paECbJpfR-X4g:b-kb1jpc) ([raw file](https://github.com/netsec-ethz/scion/blob/fa32f15b4c67e9417ec2dcaa5d1eafd806e10217/lib/msg_meta.py#L34)):*
> ```Python
>     @classmethod
>     def from_values(cls, ia=None, host=None, path=None, ext_hdrs=(),
>                     reuse=False, port=0):
> ```

Why is `reuse=False` the default?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1067)*

pszalach:

---

*[lib/msg_meta.py, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1067#-KjqgM0paECbJpfR-X4f:-KjqgsIiTefTdE5JGYeu:bmrx4qc) ([raw file](https://github.com/netsec-ethz/scion/blob/fa32f15b4c67e9417ec2dcaa5d1eafd806e10217/lib/msg_meta.py#L34)):*

Why is `reuse=False` the default?
</blockquote></details>

Because reusing is only good for (path) requests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1067)*

kormat:

---

*[lib/msg_meta.py, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1067#-KjqgM0paECbJpfR-X4f:-KjqnGAojdll7SVwd_Uj:bx9ud2l) ([raw file](https://github.com/netsec-ethz/scion/blob/fa32f15b4c67e9417ec2dcaa5d1eafd806e10217/lib/msg_meta.py#L34)):*
<details><summary><i>Previously, pszalach (Paweł Szałachowski) wrote…</i></summary><blockquote>

Because reusing is only good for (path) requests.
</blockquote></details>

What about cert requests, or local path registrations, or beacon propagation?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1067)*

pszalach:

---

*[lib/msg_meta.py, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1067#-KjqgM0paECbJpfR-X4f:-KjvYHRkIv6bz2HV770u:b-nuituj) ([raw file](https://github.com/netsec-ethz/scion/blob/fa32f15b4c67e9417ec2dcaa5d1eafd806e10217/lib/msg_meta.py#L34)):*

What about cert requests, or local path registrations, or beacon propagation?
</blockquote></details>

Cert/trc distribution is tied to path requests/responses. So, cert/trc is asked using the socket on which the corresponding path was received (if you mean these requests). (`reuse` flag makes sense only if meta is not associated with a socket yet.)

I thought about local path registrations, but it is pretty infrequent (compared with requests) and local. If you want it I'm fine with adding that.

Beacon propagation should not use `reuse`, as `reuse` will work per destination and beacon propagation is per interface.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1067)*

kormat:

---

*[lib/msg_meta.py, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1067#-KjqgM0paECbJpfR-X4f:-KkBDQgdGUBmHgL2-1SX:b-wftowc) ([raw file](https://github.com/netsec-ethz/scion/blob/fa32f15b4c67e9417ec2dcaa5d1eafd806e10217/lib/msg_meta.py#L34)):*
> Cert/trc distribution is tied to path requests/responses. So, cert/trc is asked using the socket on which the corresponding path was received (if you mean these requests). (reuse flag makes sense only if meta is not associated with a socket yet.)

I don't mean those, no. E.g. a non-master LPS needs a cert for a PCB it received over ZK, so it requests it from the local CS. A BS receives a new PCB but is missing certs, when it gets the certs it registers them with the cert server. Etc.

> I thought about local path registrations, but it is pretty infrequent (compared with requests) and local. If you want it I'm fine with adding that.

While it doesn't happen frequently, it's very bursty. Every 5s the BS will register a bunch of paths, and without `reuse` it will open a separate connection for each, which doesn't make any sense. For registering down-paths in the core, doing a connection per segment still makes sense, as it acts as a sanity check the the segment is valid (though maybe we'd need to reconsider that if it turns out to be too much overhead for the CPSes in large ISDs).

> Beacon propagation should not use reuse, as reuse will work per destination and beacon propagation is per interface.

Uh? At the very least this has the same bursty nature as local path registrations. Every 5s, a BS will propagate a bunch of beacons to each downstream AS. Why wouldn't reuse be appropriate? Just set `reuse` without an associated socket, and the network stack should do the right thing with it, no?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1067)*

pszalach:

---

*[lib/msg_meta.py, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1067#-KjqgM0paECbJpfR-X4f:-KkBTQWNSyaxo4tVqRJB:b-onard3) ([raw file](https://github.com/netsec-ethz/scion/blob/fa32f15b4c67e9417ec2dcaa5d1eafd806e10217/lib/msg_meta.py#L34)):*

> Cert/trc distribution is tied to path requests/responses. So, cert/trc is asked using the socket on which the corresponding path was received (if you mean these requests). (reuse flag makes sense only if meta is not associated with a socket yet.)

I don't mean those, no. E.g. a non-master LPS needs a cert for a PCB it received over ZK, so it requests it from the local CS. A BS receives a new PCB but is missing certs, when it gets the certs it registers them with the cert server. Etc.

> I thought about local path registrations, but it is pretty infrequent (compared with requests) and local. If you want it I'm fine with adding that.

While it doesn't happen frequently, it's very bursty. Every 5s the BS will register a bunch of paths, and without `reuse` it will open a separate connection for each, which doesn't make any sense. For registering down-paths in the core, doing a connection per segment still makes sense, as it acts as a sanity check the the segment is valid (though maybe we'd need to reconsider that if it turns out to be too much overhead for the CPSes in large ISDs).

> Beacon propagation should not use reuse, as reuse will work per destination and beacon propagation is per interface.

Uh? At the very least this has the same bursty nature as local path registrations. Every 5s, a BS will propagate a bunch of beacons to each downstream AS. Why wouldn't reuse be appropriate? Just set `reuse` without an associated socket, and the network stack should do the right thing with it, no?
</blockquote></details>

1)  this communication is still UDP-only, but I'll remember to set `reuse=True` while fixing CS.
2) added.
3) `reuse` will break propagation in the case when a sender AS is connected with a receiver AS by multiple links. The first beacon will go through well, but the second will reuse the socket, thus will go through incorrect link and will be dropped by the receiving BS.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1067)*

kormat:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1067#-:-KkGawIS4vWO7ut95Yba:bnfp4nl)*

kormat:
Reviewed 8 of 8 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1067)*

PR #1069, opened 2017-05-10
Bandwidth Enforcement in the Go router

forstern:
This pull request adds the bandwidth enforcement to the go border router by using a moving average.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1069)

### DIFF
--- a/go/border/enforcement/bw-enforcement-loader.go
+++ b/go/border/enforcement/bw-enforcement-loader.go
@@ -1,0 +1,144 @@
+// Copyright 2016 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file handles the loading, parsing and set up of the bandwidth
+// enforcement mechanism.
+
+package enforcement
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+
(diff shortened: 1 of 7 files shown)


forstern:
@ercanucan @kormat 
This PR is not finished yet. I will notify you as soon as it is ready for review.

forstern:
@ercanucan @kormat
the pr should now be ready for review. It adds most of my master thesis work to the code.

forstern:
@ercanucan @kormat
This PR should now be finally ready. 

Summary:
This PR add the basic BW mechanism implemented as part of my master thesis. 

In order to get the BW enforcement working, one has to add a config file bw.yml to brX-YY-Z. If no config file has been added, the code works as before. Where should I put such an example config file?

The enforcement mechanism is initialized at start-up of the border router. (see class bw-enforcement-loader.go)  if there is any problem while initializing the bandwidth enforcement mechanism, I write a warning to the log and then the router starts up without BW enforcement!

How does it work:
In the config file one can add the interfaces one wants to have BW enforcement enabled. Each interface can have a list of ASes that have reserved/restricted BW (which term to use here depends on the use case). Egress and Ingress BW per AS can be restricted individually.
Additionally there exists a field 'unknwon' in order to restrict egress(ingress traffic from unknown ASes (ASes not listed)

Under the hood, the BW enforcement mechanism uses a bucketed ring buffer in order to keep track of the current moving average. The number of buckets and the duration one bucket represents can be configured in the code. At the moment it is set to 5 buckets, 1 second each bucket. 

The maximum bandwidth per AS defined in the config file is a hard limit. As soon as a AS overshoots the maximum BW its packets get dropped (monitored with prometheus).

If the AS uses more than 95% of its maximum BW we start to monitor it with prometheus as well. This way a network/AS admin is able to see ASes with high loads.

At 50 Mbps the average processing time for a packet is at 450ns. If the packet gets dropped the the processing time rises to 750ns. Prometheus adds an considerable amount of processing time to the packet. 

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1069#-:-KpFsFYzH8NvS9XKPXiF:b3aossy)*

scrye:
Closing due to age.

PR #1073, opened 2017-05-15
Add cache-only flag to path requests.

kormat:
This allows a path server to be quereies without causing the PS to send
any further queries itself.

By setting this flag when a CPS queries the local master CPS, this
prevents query loops (which we've seen when two non-master CPSes thought
the other was the master).

Also:
- Un-set the master ID when a CPS becomes master. This helps prevent
  stale master information, but doesn't completely solve it. It's still
  possible for a non-master CPS to be disconnected from ZK and have
  cached a stale master ID, but the cache-only flag should stop that
  from causing harm at least.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1073)


shitz:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1073#-:-KkAvtlAPoqXdpE3Ko1p:bnfp4nl)*

PR #1086, opened 2017-05-18
Fix 2 issues with dependencies.

kormat:
- `pip freeze` does not include pip, setuptools, distribute, wheel in
  the output, which breaks `env/pip3/check` if any of those are
  specified in the requirements.txt.
- Run ./env/deps with APTARGS=-y in `tools/ci/build_prep`, otherwise it
  will immediately fail because it can't read an answer from stdin.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1086)

### DIFF
--- a/env/common.sh
+++ b/env/common.sh
@@ -8,7 +8,7 @@ lower() {
 }
 
 pip_installed() {
-    "${1:?}" --disable-pip-version-check freeze | lower | LC_ALL=C sort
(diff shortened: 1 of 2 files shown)


pszalach:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1086#-:-KkQjWH-snlSneRxzmvN:bnfp4nl)*

PR #1096, opened 2017-05-22
Add BR cmdline flag to specify metric export address.

kormat:
This makes it possible to export metrics on an interface that the BR
isn't using itself.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1096)

### DIFF
--- a/go/border/metrics/metrics.go
+++ b/go/border/metrics/metrics.go
@@ -17,17 +17,19 @@
 package metrics
 
 import (
+	"flag"
 	"io"
 	"net"
 	"net/http"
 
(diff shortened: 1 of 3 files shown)


shitz:
I'm a bit sad to see the re-exporting functionality go, but you are probably right that if you change the local address of the router you probably restart it anyway at that point. :lgtm: 

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1096#-:-KkkyvqDvYBwig_D8-Nm:b-8c0wca)*

PR #1101, opened 2017-05-23
IssuerVersion in Certificates

Oncilla:
Add issuer version to Certificates

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1101)

### DIFF
--- a/infrastructure/scion_elem.py
+++ b/infrastructure/scion_elem.py
@@ -512,8 +512,7 @@ class SCIONElement(object):
             trc = self.trust_store.get_trc(cert_ia[0], asm.p.trcVer)
             chain = self.trust_store.get_cert(asm.isd_as(), asm.p.certVer)
             ver_seg.add_asm(asm)
-            verify_sig_chain_trc(ver_seg.sig_pack3(), asm.p.sig,
-                                 str(cert_ia), chain, trc, asm.p.trcVer)
+            verify_sig_chain_trc(ver_seg.sig_pack3(), asm.p.sig, cert_ia, chain, trc)
 
     def _get_handler(self, pkt):
         # FIXME(PSz): needed only by python router.
--- a/lib/crypto/certificate.py
+++ b/lib/crypto/certificate.py
@@ -26,6 +26,7 @@ from lib.errors import SCIONVerificationError
 
 SUBJECT_STRING = 'Subject'
 ISSUER_STRING = 'Issuer'
+ISSUER_VERSION_STRING = "IssuerVersion"
 VERSION_STRING = 'Version'
 COMMENT_STRING = 'Comment'
 CAN_ISSUE_STRING = 'CanIssue'
@@ -45,6 +46,7 @@ class Certificate(object):
 
     :ivar str subject: the certificate subject.
     :ivar str issuer: the certificate issuer. It can only be an AS.
+    :ivar int issuer_version: the version of the issuing certificate/trc.
     :ivar int version: the certificate version.
     :ivar str comment: is an arbitrary and optional string used by the subject
         to describe the certificate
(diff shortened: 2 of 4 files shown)


pszalach:
I'm not sure about that change. Every signed object (certificate, beacon, path, ...) must be verified with the current TRC, or exceptionally with the previous TRC (if this TRC was updated and the grace period is active). Specifically, an object should not be verified against the TRC it was created with. Just imagine the following sequence:
1) TRCVer=n and a misbehaving AS creates many certificates (that have TRCVer=n)
2) the core notices misbehaving and updates TRC (to n+1) removing the malicious AS
3) the update is propagated through the network
4) the malicious certificates are still valid as relying parties use TRCVer=n to verify it, ignoring the fact that the bad AS was removed.

TRC version in certificates may be helpful for synchronization or audit (we have timestamps already that may help as well), but shouldn't determine TRCs used for verification.
What makes sense to me is to have an API to ask for the current TRC (or maybe the two latest, to handle the grace period corner case). AFAIR we have such an API between app<->sciond.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/crypto/certificate.py, line 29 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1101#-KktlaiEZMEEqTkxfa8w:-KktlaiEZMEEqTkxfa8x:b-zd1oju) ([raw file](https://github.com/netsec-ethz/scion/blob/f1ee3f66766d2bf1a6424a80989418885b5acd20/lib/crypto/certificate.py#L29)):*
> ```Python
> SUBJECT_STRING = 'Subject'
> ISSUER_STRING = 'Issuer'
> ISSUER_VERSION_STRING = "IssuerVersion"
> ```

`TRCVersion` or `IssuerTRCVersion` would be more meaningful

---

*[lib/crypto/certificate.py, line 120 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1101#-KktlyRPZIal3XFjOTmR:-KktlyRPZIal3XFjOTmS:b-kwnlw0) ([raw file](https://github.com/netsec-ethz/scion/blob/f1ee3f66766d2bf1a6424a80989418885b5acd20/lib/crypto/certificate.py#L120)):*
> ```Python
>         if int(time.time()) >= issuer_cert.expiration_time:
>             raise SCIONVerificationError("The issuer certificate expired:\n%s" % issuer_cert)
>         if self.issuer_version != issuer_cert.version:
> ```

it doesn't have to match. We always should verify with the current TRC (exceptionally, with the previous one). See my general comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1101#-:-KktmMnSr4FsKrZtDFqv:b-o058eo)*

Oncilla:
Okay. So if I want to verify a CertChain, I should always use the (second) newest TRC. If that fails, the certificate is invalid.
If this holds true, then we can abandon this PR.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1101#-:-KktwWg8X844lB8JovRS:b-j8f6v7)*

pszalach:
Just to clarify:
1) you always try to verify with the newest TRC
2) if that fails you get the second newest and if the grace period is active you can try to verify against that TRC.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1101#-:-KktyGTB8L_pj2dlBUd3:b2wxzbc)*

Oncilla:
Okay. But if that fails, the CertChain is invalid?

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1101#-:-Kktyntdf0sYAeNgAR9Q:b-7we5el)*

pszalach:
Correct.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1101#-:-KktzC3JO2QCwSWiI3cM:btqqu5l)*

PR #1107, opened 2017-05-29
Add tarball hash for cffi.

kormat:
On amd64 machines, pip3 downloads the wheel file for cffi, which matches
the original hash in `env/pip3/requirements.txt`. On non-amd64 machines
(e.g. the odroid arm devices), there may not be a wheel available, so it
will download the bare tarball instead, which then fails to match the
hash. Pip supports providing multiple hashes, for exactly this use-case
(https://pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode).

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1107)


klausman:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1107#-:-KlJODmTaT7dFbngngXH:bnfp4nl)*

PR #1111, opened 2017-05-30
Trc update

sezergueler:
First version of TRC update. A new TRC is generated on topology generation by a selected AS defined in the topology file. The new TRC is read from fs by the CS 10 seconds after start up and disseminated to other ASes. Receiving ASes first verify the new TRC and then disseminate it. 

Cross signatures are not checked yet. This will be implemented once the xsigs PR is merged.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1111)

### DIFF
--- a/python/cert_server/main.py
+++ b/python/cert_server/main.py
@@ -21,10 +21,10 @@ import datetime
 import logging
 import os
 import threading
+import time
 
 # External packages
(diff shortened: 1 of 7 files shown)


shitz:
This is outdated and will be superseded by our current work on the Control-Plane PKI.

PR #1115, opened 2017-06-01
Go API for SCIOND

scrye:
The API exposes structures that mirror the Cap'n Proto messages for communication with SCIOND. The entire SCIOND reply is passed back to the application. To facilitate ease of use, some reply fields (such as ISD-AS pairs) are converted directly by the API to the appropriate structs.

**Note**: This API is tested using the end-to-end connectivity tests between random sources and destinations. The test is over 500 lines long and not included in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1115)


shitz:
Reviewed 1 of 4 files at r1.
Review status: 1 of 4 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[go/lib/sciond/sciond.go, line 26 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KldClG_l_FSzFUsVj3l:-KldClGaY9g-r1zpJvMq:b-5kdayf) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L26)):*
> ```Go
> // These are rarely useful, and the same fields without the prefix (e.g., ErrorCode) should be used
> // instead.
> package sciond
> ```

Add a TODO somewhere that revocation notifications need to be implemented.

---

*[go/lib/sciond/sciond.go, line 111 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KlcyhlojH-qHgb8gq5O:-KlcyhlpUfPYglt4cMY5:b-9zcz) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L111)):*
> ```Go
> // Paths requests from SCIOND a set of end to end paths between src and dst. max specifices the
> // maximum number of paths returned.
> func (c *Connector) Paths(src, dst *addr.ISD_AS, max uint16, flush bool,
> ```

This is not very extendable. If we add flags for path requests the API will change. If you put the flags in a `PathReqFlags` struct then the API of `Paths` wouldn't change when adding/removing flags.

---

*[go/lib/sciond/sciond.go, line 130 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld-jq8eZp6E9Kll0qC:-Kld-jq8eZp6E9Kll0qD:bfi1hgn) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L130)):*
> ```Go
> 
> 	// Expose "pretty" ISD-ASes
> 	reply.PathReply.postprocess()
> ```

It's more common in our codebase to have getters. So for the ISD-AS you'd have something like
```
type PathInterface struct {
	RawIsdas uint32 `capnp:"isdas"`
	IfID     uint64
	ia       *addr.ISD_AS `capnp:"-"`
}

func (pi *PathInterface) IA() *addr.ISD_AS {
    if pi.ia == null {
        pi.ia = addr.IAFromInt(int(pi.RawIsdas))
    }
    return pi.ia
}
```

I think you also don't have to expose the Raw* members at all. Just provide an appropriate getter.

---

*[go/lib/sciond/sciond.go, line 138 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld2X4ZjMXS14wSV-An:-Kld2X4_O8Ih_VX8eS3k:b3dsi3i) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L138)):*
> ```Go
> 	// Check if information for this ISD-AS is cached
> 	key := ia.String()
> 	value, found := c.asInfos.Get(key)
> ```

This can be simplified to 
```
	if value, found := c.asInfos.Get(key); found {
		return value.(*ASInfoReply), nil
	}
```
This applies to various other places as well.

---

*[go/lib/sciond/sciond.go, line 158 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld2v-AHKJCUo-up1hS:-Kld2v-AHKJCUo-up1hT:b-stg8uv) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L158)):*
> ```Go
> 	reply.AsInfoReply.postprocess()
> 	// Cache result
> 	c.asInfos.Set(key, &reply.AsInfoReply, cache.DefaultExpiration)
> ```

Using `SetDefault` you don't have to pass `cache.DefaultExpiration`

---

*[go/lib/sciond/sciond.go, line 162 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld3XTG-0MzRpF5G5z2:-Kld3XTG-0MzRpF5G5z3:briqwz8) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L162)):*
> ```Go
> }
> 
> // IFInfo requests from SCIOND information about addresses and ports of border routers (BRs).
> ```

"... addresses and ports of interfaces" (feel free to change it in the python version as well)

---

*[go/lib/sciond/sciond.go, line 163 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld39jRGAotCvkJUkID:-Kld39jRGAotCvkJUkIE:bm6m2l7) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L163)):*
> ```Go
> 
> // IFInfo requests from SCIOND information about addresses and ports of border routers (BRs).
> // Slice ifs contains interface IDs of BRs. If unset, all BRs are
> ```

"If empty, all interfaces are returned."

---

*[go/lib/sciond/sciond.go, line 211 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld8EM0TGdFmmJiX7ob:-Kld8EM0TGdFmmJiX7oc:b-itgb58) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L211)):*
> ```Go
> // Slice svcTypes contains a list of desired service types. If unset, all service types are
> // returned.
> func (c *Connector) SVCInfo(svcTypes []addr.HostSVC) (*ServiceInfoReply, error) {
> ```

This won't actually work. The service type is an enum and capnp generated the following:
```
// Values of ServiceInfoRequest_ServiceType.
const (
	ServiceInfoRequest_ServiceType_bs ServiceInfoRequest_ServiceType = 0
	ServiceInfoRequest_ServiceType_ps ServiceInfoRequest_ServiceType = 1
	ServiceInfoRequest_ServiceType_cs ServiceInfoRequest_ServiceType = 2
	ServiceInfoRequest_ServiceType_br ServiceInfoRequest_ServiceType = 3
	ServiceInfoRequest_ServiceType_sb ServiceInfoRequest_ServiceType = 4
)
```
I suggest you define something like
```
type ServiceType uint16 
const (
	SvcBS ServiceType = iota
	SvcPS
	SvcCS
	SvcBR
	SvcSB 
)
```
so we don't to work with these unwieldy capnp enums.

---

*[go/lib/sciond/types.go, line 26 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KldDYvlvKLkjYJmQm4n:-KldDYvlvKLkjYJmQm4o:b-tqw4ov) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/types.go#L26)):*
> ```Go
> type PathErrorCode uint16
> 
> const (
> ```

You can define enums like this
```
type PathErrorCode uint16
const (
    ErrorOk PathErrorCode = iota
    ErrorNoPaths
    ErrorPSTimeout
    ErrorInternal
)
```

---

*[go/lib/sciond/types.go, line 108 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KldDxeBQB_hYd2Qfv0C:-KldDxeBQB_hYd2Qfv0D:bolp5yk) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/types.go#L108)):*
> ```Go
> 
> type PathInterface struct {
> 	RawIsdas uint32 `capnp:"isdas"`
> ```

See my comment above about the RawIsdas struct members.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115)*

shitz:
[Port SCIOND API to Go](https://trello.com/c/VhuxkN1n/18-port-sciond-api-to-go)

scrye:
Review status: 1 of 5 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[go/lib/sciond/sciond.go, line 26 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KldClG_l_FSzFUsVj3l:-KlwR5MHyAh7pyCyUZ4b:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L26)):*

Add a TODO somewhere that revocation notifications need to be implemented.
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 111 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KlcyhlojH-qHgb8gq5O:-KlwRp9I663SyMqH2OJg:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L111)):*

This is not very extendable. If we add flags for path requests the API will change. If you put the flags in a `PathReqFlags` struct then the API of `Paths` wouldn't change when adding/removing flags.
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 130 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld-jq8eZp6E9Kll0qC:-KlwVCdbgvjQOC3ntnUC:bx7m0lg) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L130)):*

It's more common in our codebase to have getters. So for the ISD-AS you'd have something like
```
type PathInterface struct {
	RawIsdas uint32 `capnp:"isdas"`
	IfID     uint64
	ia       *addr.ISD_AS `capnp:"-"`
}

func (pi *PathInterface) IA() *addr.ISD_AS {
    if pi.ia == null {
        pi.ia = addr.IAFromInt(int(pi.RawIsdas))
    }
    return pi.ia
}
```

I think you also don't have to expose the Raw* members at all. Just provide an appropriate getter.
</blockquote></details>

Sadly I need to set them to public for reflection to work. I added getters for all `Raw` fields, if apps want to use the prettier structures instead of packed data.

---

*[go/lib/sciond/sciond.go, line 138 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld2X4ZjMXS14wSV-An:-Klw_AjE9ylI2NNv5Y0Y:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L138)):*

This can be simplified to 
```
	if value, found := c.asInfos.Get(key); found {
		return value.(*ASInfoReply), nil
	}
```
This applies to various other places as well.
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 158 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld2v-AHKJCUo-up1hS:-Klw_RIpknLZVZ4zlqZV:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L158)):*

Using `SetDefault` you don't have to pass `cache.DefaultExpiration`
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 162 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld3XTG-0MzRpF5G5z2:-Klwc0qd82H4JOqmLvRM:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L162)):*

"... addresses and ports of interfaces" (feel free to change it in the python version as well)
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 163 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld39jRGAotCvkJUkID:-Klwbz6Jsxj60Dv5DaPl:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L163)):*

"If empty, all interfaces are returned."
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 211 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld8EM0TGdFmmJiX7ob:-KlwdUGdrB3NQbQZSx95:b31izlp) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L211)):*

This won't actually work. The service type is an enum and capnp generated the following:
```
// Values of ServiceInfoRequest_ServiceType.
const (
	ServiceInfoRequest_ServiceType_bs ServiceInfoRequest_ServiceType = 0
	ServiceInfoRequest_ServiceType_ps ServiceInfoRequest_ServiceType = 1
	ServiceInfoRequest_ServiceType_cs ServiceInfoRequest_ServiceType = 2
	ServiceInfoRequest_ServiceType_br ServiceInfoRequest_ServiceType = 3
	ServiceInfoRequest_ServiceType_sb ServiceInfoRequest_ServiceType = 4
)
```
I suggest you define something like
```
type ServiceType uint16 
const (
	SvcBS ServiceType = iota
	SvcPS
	SvcCS
	SvcBR
	SvcSB 
)
```
so we don't to work with these unwieldy capnp enums.
</blockquote></details>

From a type point of view, I don't think there are any issues; addr.HostSVC is a uint16, same as ServiceInfoRequest_ServiceType.

However, I now noticed there's a mismatch between capnp and the defines in `addr/host.go`. For clarity, I think it's best if we have the same value in all places (e.g., hardcode capnp with our `addr.host.go` values). Otherwise, API users will see two incompatible definitions for service types.

---

*[go/lib/sciond/types.go, line 26 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KldDYvlvKLkjYJmQm4n:-Klxd1n0cJaqv6Gtli5_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/types.go#L26)):*

You can define enums like this
```
type PathErrorCode uint16
const (
    ErrorOk PathErrorCode = iota
    ErrorNoPaths
    ErrorPSTimeout
    ErrorInternal
)
```
</blockquote></details>

Done.

---

*[go/lib/sciond/types.go, line 108 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KldDxeBQB_hYd2Qfv0C:-KlxY3D4bKd3o4XWlQ0k:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/types.go#L108)):*

See my comment above about the RawIsdas struct members.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115)*

shitz:
Reviewed 1 of 4 files at r1, 1 of 3 files at r2.
Review status: 3 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[go/lib/sciond/sciond.go, line 211 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld8EM0TGdFmmJiX7ob:-KlyCFNb_SPNJFzN3pfN:b85s77f) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L211)):*
I don't agree with this. First of all, if the multicast bit is set then the value of `HostSVC` wouldn't correspond to `0, 1, 2, ...` anymore. It would probably still work, since a client should not ask for a multicast SVC address anyway, but we shouldn't rely on that. The enum in capnp is completely independent of the values of `HostSVC*` and I think it should stay that way. One is a network address the other just an arbitrary enumeration. This is from the capnp definition

>Like fields, enumerants must be numbered sequentially starting from zero. In languages where enums have numeric values, these numbers will be used, but in general Cap’n Proto enums should not be considered numeric.

In hindsight, I'm not a big fan of capnp enums anymore and we might just want the service type field a `uint8` or `uint16`.

---

*[go/lib/sciond/sciond.go, line 24 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KlyB4wj3XiMGVQXPa67:-KlyB4wj3XiMGVQXPa68:b8ct8rm) ([raw file](https://github.com/netsec-ethz/scion/blob/6ef11ff613525c8e8b529fdd1f71c2330b72ec28/go/lib/sciond/sciond.go#L24)):*
> ```Go
> //
> // Fields prefixed with Raw (e.g., RawErrorCode) contain data in the format received from SCIOND.
> // These are rarely useful, and the same fields without the prefix (e.g., ErrorCode) should be used
> ```

Update the comment to reflect that the API now offers accessors like `ErrorCode()`.

---

*[go/lib/sciond/types.go, line 26 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KldDYvlvKLkjYJmQm4n:-KlyD8_rTXZUVpvuOysE:bfhqmln) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/types.go#L26)):*

Done.
</blockquote></details>

Not done?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115)*

scrye:
Review status: 3 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[go/lib/sciond/sciond.go, line 211 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Kld8EM0TGdFmmJiX7ob:-Km25vlh4nHjEkhtbGuh:bkrhlla) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/sciond.go#L211)):*

I don't agree with this. First of all, if the multicast bit is set then the value of `HostSVC` wouldn't correspond to `0, 1, 2, ...` anymore. It would probably still work, since a client should not ask for a multicast SVC address anyway, but we shouldn't rely on that. The enum in capnp is completely independent of the values of `HostSVC*` and I think it should stay that way. One is a network address the other just an arbitrary enumeration. This is from the capnp definition

>Like fields, enumerants must be numbered sequentially starting from zero. In languages where enums have numeric values, these numbers will be used, but in general Cap’n Proto enums should not be considered numeric.

In hindsight, I'm not a big fan of capnp enums anymore and we might just want the service type field a `uint8` or `uint16`.
</blockquote></details>

You're right, I initially thought they referred to the same thing but now I see that they're different. I fixed it using your suggestion.

---

*[go/lib/sciond/sciond.go, line 24 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KlyB4wj3XiMGVQXPa67:-Km20N4XhrOAbPtLGZi6:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/6ef11ff613525c8e8b529fdd1f71c2330b72ec28/go/lib/sciond/sciond.go#L24)):*

Update the comment to reflect that the API now offers accessors like `ErrorCode()`.
</blockquote></details>

Done.

---

*[go/lib/sciond/types.go, line 26 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KldDYvlvKLkjYJmQm4n:-Km26fUwhppjbkbdBs5Z:b-24rp) ([raw file](https://github.com/netsec-ethz/scion/blob/0bb6f4a92764b0be3ee08a428bed2d9562ebddc8/go/lib/sciond/types.go#L26)):*

Not done?
</blockquote></details>

No idea how I reverted this one because I'm sure I rewrote it once. Done now (hopefully).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115)*

shitz:
The only thing that bothers me a bit is that the Go version of the API returns the reply from SCIOND directly, whereas the Python version returns a list of reply entries. I think they should be consistent and we can discuss which way it makes more sense.

---

Review status: 3 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/sciond/sciond.go, line 192 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Km2RrG8oJuYMsPKIMKp:-Km2RrG8oJuYMsPKIMKq:bwjykyc) ([raw file](https://github.com/netsec-ethz/scion/blob/001660ed144c08101efa51fbc98d148308a21fac/go/lib/sciond/sciond.go#L192)):*
> ```Go
> 	}
> 
> 	if len(uncachedIfs) == 0 {
> ```

Wait a minute. Does this work? If `ifs` is empty then `uncachedIfs` is also empty but you actually want to request SCIOND about all interfaces. Would also apply to `SVCInfo`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-:-Km2T2iq_uaZllJAX4iL:b1vkirc)*

scrye:
*[go/lib/sciond/sciond.go, line 192 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Km2RrG8oJuYMsPKIMKp:-Km6I1B1ss9NVR5OMo1C:bh76kex) ([raw file](https://github.com/netsec-ethz/scion/blob/001660ed144c08101efa51fbc98d148308a21fac/go/lib/sciond/sciond.go#L192)):*

Wait a minute. Does this work? If `ifs` is empty then `uncachedIfs` is also empty but you actually want to request SCIOND about all interfaces. Would also apply to `SVCInfo`.
</blockquote></details>

You're right, it doesn't work; this kind of bug should have been caught in testing, but the end to end tests do not provide enough coverage.

I'm going to write an integration test for SCIOND and include it in the next update to this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115)*

scrye:
Review status: 3 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/sciond/sciond.go, line 192 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Km2RrG8oJuYMsPKIMKp:-Km6o14ljxL75PHqYrZU:brkujap) ([raw file](https://github.com/netsec-ethz/scion/blob/001660ed144c08101efa51fbc98d148308a21fac/go/lib/sciond/sciond.go#L192)):*

You're right, it doesn't work; this kind of bug should have been caught in testing, but the end to end tests do not provide enough coverage.

I'm going to write an integration test for SCIOND and include it in the next update to this PR.
</blockquote></details>

Included some very basic integration tests for SCIOND.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115)*

shitz:
> The only thing that bothers me a bit is that the Go version of the API returns the reply from SCIOND directly, whereas the Python version returns a list of reply entries. I think they should be consistent and we can discuss which way it makes more sense.

As discussed offline, we are leaving the inconsistency between the different languages.

:lgtm: 

---

Reviewed 1 of 5 files at r4.
Review status: 3 of 5 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-:-KmB3cj0L7G2p_DMaMcv:bd0k3v3)*

kormat:
Reviewed 2 of 4 files at r1, 3 of 5 files at r4.

---

*[go/lib/sciond/sciond.go, line 50 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBBg7-3IHtimvSsVkT:-KmBBg7-3IHtimvSsVkU:bplf4es) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L50)):*
> ```Go
> 	ASInfoTTL  = time.Hour
> 	IFInfoTTL  = time.Hour
> 	SVCInfoTTL = 10 * time.Second
> ```

Minor nitpick. The order here is AS/IF/SVC. In `Connector`'s attributes, it's IF/SVC/AS. In the connector's methods, it's AS/IF/SVC again. The simplest fix would be to change the order of `Connector`'s attributes.

---

*[go/lib/sciond/sciond.go, line 103 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBCnELNAd0Bo30fRLB:-KmBCnELNAd0Bo30fRLC:bt2t4he) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L103)):*
> ```Go
> 		return nil, err
> 	}
> 	message := new(SCIONDMsg)
> ```

`message := &SCIONDMsg{}` is equivalent, and more consistent with the rest of the code.

---

*[go/lib/sciond/sciond.go, line 115 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBAVSHN1lrtIBzLeFp:-KmBAVSHN1lrtIBzLeFq:br27dlz) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L115)):*
> ```Go
> // guaranteed to be fresh, as the returned connector caches ASInfo replies for ASInfoTTL time,
> // IFInfo replies for IFInfoTTL time and SVCInfo for SVCInfoTTL time.
> func Connect(socketName string) (*Connector, error) {
> ```

The constructor should be directly after the type declaration.

---

*[go/lib/sciond/sciond.go, line 132 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBDMITm7JexOMOJugp:-KmBDMITm7JexOMOJugq:blqzujf) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L132)):*
> ```Go
> // Paths requests from SCIOND a set of end to end paths between src and dst. max specifices the
> // maximum number of paths returned.
> func (c *Connector) Paths(src, dst *addr.ISD_AS, max uint16, f PathReqFlags) (*PathReply, error) {
> ```

The python code has a different AS order (`dst, src`), which is probably worth keeping in Go.

---

*[go/lib/sciond/sciond.go, line 178 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBRx_xTuOmCeYi8VKA:-KmBRx_yeWmHwPzGSdEL:b-9f4y1t) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L178)):*
> ```Go
> // IFInfo requests from SCIOND addresses and ports of interfaces.
> // Slice ifs contains interface IDs of BRs. If empty, all interfaces are
> // returned.
> ```

It's probably worth mentioning that the cache isn't used in the latter case.

---

*[go/lib/sciond/sciond.go, line 179 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBRg0lD5VZ6GRN_fjD:-KmBRg0lD5VZ6GRN_fjE:bujl953) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L179)):*
> ```Go
> // Slice ifs contains interface IDs of BRs. If empty, all interfaces are
> // returned.
> func (c *Connector) IFInfo(ifs []uint64) (*IFInfoReply, error) {
> ```

This should probably be returning a map instead of a list.

---

*[go/lib/sciond/types.go, line 75 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmB7o0CsOunoJpHru1I:-KmB7o0CsOunoJpHru1J:b-rmw2nw) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/types.go#L75)):*
> ```Go
> 
> type PathReply struct {
> 	RawErrorCode uint16 `capnp:"errorCode"`
> ```

What happens if you just use `PathErrorCode` directly here?

---

*[go/lib/sciond/types.go, line 111 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmB95-ryHwc6DsPUWPd:-KmB95-ryHwc6DsPUWPe:bj9uvd3) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/types.go#L111)):*
> ```Go
> }
> 
> func (entry PathInterface) String() string {
> ```

s/entry/iface/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115)*

scrye:
Review status: 3 of 6 files reviewed at latest revision, 9 unresolved discussions.

---

*[go/lib/sciond/sciond.go, line 192 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-Km2RrG8oJuYMsPKIMKp:-KmBeB6vBd8ZavPrdyAm:b-77zy9c) ([raw file](https://github.com/netsec-ethz/scion/blob/001660ed144c08101efa51fbc98d148308a21fac/go/lib/sciond/sciond.go#L192)):*

Included some very basic integration tests for SCIOND.
</blockquote></details>

Following discussions, we decided to move the integration tests to a different PR.

---

*[go/lib/sciond/sciond.go, line 50 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBBg7-3IHtimvSsVkT:-KmBTuyuO8siWjrbtOid:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L50)):*

Minor nitpick. The order here is AS/IF/SVC. In `Connector`'s attributes, it's IF/SVC/AS. In the connector's methods, it's AS/IF/SVC again. The simplest fix would be to change the order of `Connector`'s attributes.
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 103 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBCnELNAd0Bo30fRLB:-KmBU1-u3svEoxvOrR-2:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L103)):*

`message := &SCIONDMsg{}` is equivalent, and more consistent with the rest of the code.
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 115 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBAVSHN1lrtIBzLeFp:-KmBUGsQLt0qshxDVWWF:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L115)):*

The constructor should be directly after the type declaration.
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 132 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBDMITm7JexOMOJugp:-KmBUUgH5OQ4I4ecKdBU:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L132)):*

The python code has a different AS order (`dst, src`), which is probably worth keeping in Go.
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 178 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBRx_xTuOmCeYi8VKA:-KmBXIXw1ovCf5rzbTFo:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L178)):*

It's probably worth mentioning that the cache isn't used in the latter case.
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 179 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmBRg0lD5VZ6GRN_fjD:-KmBaBKIew9WiZah6FaN:b-mn302j) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/sciond.go#L179)):*

This should probably be returning a map instead of a list.
</blockquote></details>

For the sake of API consistency, I kept the returned type but added an accessor `Entries()` that returns the map and *hid* the list behind `RawEntries`. Repeatedly calling the accessor is inefficient, but it prevents weird inconsistent states (say, if for some arcane reason some app developer ever overwrites the contents of `RawEntries` between calls to `Entries()`).

---

*[go/lib/sciond/types.go, line 75 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmB7o0CsOunoJpHru1I:-KmBYNby8sHerjyVsyTt:bii0qyv) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/types.go#L75)):*

What happens if you just use `PathErrorCode` directly here?
</blockquote></details>

It works out of the box, which means we can drop the accessor. `ServiceInfoReplyEntry` used `ServiceType` in a similar way, but I somehow missed this one.

---

*[go/lib/sciond/types.go, line 111 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-KmB95-ryHwc6DsPUWPd:-KmBYlexC--Xqs3U1qlB:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4e270a1b06d38924f7091e20d69b1d4de7c75dd7/go/lib/sciond/types.go#L111)):*

s/entry/iface/
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115)*

kormat:
:lgtm:

---

Reviewed 3 of 3 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1115#-:-KmBfQnNTQfhnswuebZA:bnfp4nl)*

PR #1117, opened 2017-06-02
Cert sync with book

sezergueler:
In the current implementation every AS gets a certificate issued by its parent AS, thus creating long certificate chains.
This PR contains the following changes:

- Clear separation between core and non-core AS certificates

- Core AS certificates are signed and verified based on online root keys

- Only core ASes issue certificates

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1117)

### DIFF
--- a/python/lib/crypto/certificate.py
+++ b/python/lib/crypto/certificate.py
@@ -45,7 +45,7 @@ class Certificate(object):
     information for further use.
 
     :ivar str subject: the certificate subject.
-    :ivar str issuer: the certificate issuer. It can only be an AS.
+    :ivar str issuer: the certificate issuer. It can only be a core AS.
     :ivar int trc_version: the version of the issuing trc.
     :ivar int version: the certificate version.
     :ivar str comment: is an arbitrary and optional string used by the subject
@@ -60,13 +60,15 @@ class Certificate(object):
     :ivar bytes subject_sig_key: the public key used for signing.
     :ivar bytes signature: the certificate signature. It is computed over the
         rest of the certificate.
-    :cvar int validity_period:
-        default validity period (in real seconds) of a new certificate.
+    :cvar int as_validity_period:
+        default validity period (in real seconds) of a new regular AS certificate.
+    :cvar int core_as_validity_period:
+        default validity period (in real seconds) of a new core AS certificate.
     :cvar str sign_algortihm: default algorithm used to sign a certificate.
     :cvar str enc_alorithm: default algorithm used to encrypt messages.
     """
-
-    VALIDITY_PERIOD = 365 * 24 * 60 * 60
+    AS_VALIDITY_PERIOD = 3 * 24 * 60 * 60
+    CORE_AS_VALIDITY_PERIOD = 7 * 24 * 60 * 60
     SIGN_ALGORTIHM = 'ed25519'
     ENC_ALGORITHM = 'curve25519xsalsa20poly1305'
     FIELDS_MAP = {
@@ -99,14 +101,14 @@ class Certificate(object):
         self.subject_sig_key_raw = base64.b64decode(self.subject_sig_key)
         self.signature_raw = base64.b64decode(self.signature)
 
-    def verify(self, subject, issuer_cert):
+    def verify(self, subject, verifying_key):
         """
         Perform one step verification.
 
         :param str subject:
             the certificate subject. It can either be an AS, an email address or
             a domain address.
-        :param str issuer_cert: the certificate issuer. It can only be an AS.
+        :param bytes verifying_key: the key to be used for signature verification.
         :raises: SCIONVerificationError if the verification fails.
         """
         if subject != self.subject:
@@ -115,23 +117,10 @@ class Certificate(object):
                 (subject, self.subject, self))
         if int(time.time()) >= self.expiration_time:
             raise SCIONVerificationError("This certificate expired:\n%s" % self)
-        if int(time.time()) >= issuer_cert.expiration_time:
-            raise SCIONVerificationError("The issuer certificate expired:\n%s" % issuer_cert)
         try:
-            self._verify_signature(self.signature_raw, issuer_cert.subject_sig_key_raw)
+            self._verify_signature(self.signature_raw, verifying_key)
         except SCIONVerificationError:
             raise SCIONVerificationError("Signature verification failed:\n%s" % self)
-
-    def verify_core(self, pub_online_root_key):
-        """
-        Verify core signature with given online root key.
-
-        :param bytes pub_online_root_key:
-            The online root key of the core AS  which signed this
-            root certificate
-        :raises: SCIONVerificationError if the verification fails.
-        """
-        self._verify_signature(self.signature_raw, pub_online_root_key)
 
     def _verify_signature(self, signature, public_key):
         """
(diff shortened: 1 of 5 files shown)


shitz:
:lgtm:

---

Reviewed 5 of 5 files at r1.

---

*[python/lib/crypto/certificate_chain.py, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-KmgXAjr3FCPq1Hsfs6M:-KmgXAjr3FCPq1Hsfs6N:b-8iz1wf) ([raw file](https://github.com/netsec-ethz/scion/blob/045fee61017e2ce0c845624734ce7d1d73592d29/python/lib/crypto/certificate_chain.py#L115)):*
> ```Python
>             self.as_cert.verify(subject, self.core_as_cert.subject_sig_key_raw)
>         except SCIONVerificationError as e:
>             raise SCIONVerificationError("AS certificate verification failed: \n%s" % e)
> ```

Don't use a new line after the colon.

---

*[python/lib/crypto/certificate_chain.py, line 121 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-KmgXE79l4o6u0Igl-OU:-KmgXE79l4o6u0Igl-OV:b-ehpfwb) ([raw file](https://github.com/netsec-ethz/scion/blob/045fee61017e2ce0c845624734ce7d1d73592d29/python/lib/crypto/certificate_chain.py#L121)):*
> ```Python
>                                      trc.core_ases[self.core_as_cert.issuer][ONLINE_KEY_STRING])
>         except SCIONVerificationError:
>             raise SCIONVerificationError("Core AS certificate verification failed.")
> ```

Include the reason why the signature verification failed here as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-:-KmgXlo0YJ2yotUaYkAl:bnfp4nl)*

pszalach:
I'm only not sure about limiting certificate chains to two certificates. @oncilla don't you need longer chains (i.e., certificates for endhosts)?

---

Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[python/lib/crypto/certificate_chain.py, line 85 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-Kmjd6PEIUFJSQnoT1p5:-Kmjd6PEIUFJSQnoT1p6:b-ijl133) ([raw file](https://github.com/netsec-ethz/scion/blob/045fee61017e2ce0c845624734ce7d1d73592d29/python/lib/crypto/certificate_chain.py#L85)):*
> ```Python
>         :param str cert_list: certificate chain as list.
>         """
>         assert len(cert_list) == 2
> ```

`raise SCIONParseError` ?

---

*[python/topology/generator.py, line 351 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-KmjdHtLManEwZ6nLBP1:-KmjdHtLManEwZ6nLBP2:b-rqxvnq) ([raw file](https://github.com/netsec-ethz/scion/blob/045fee61017e2ce0c845624734ce7d1d73592d29/python/topology/generator.py#L351)):*
> ```Python
>         issuer = TopoID(as_conf.get('cert_issuer', str(topo_id)))
>         # Make sure that issuer is a core AS
>         assert issuer in self.pub_online_root_keys
> ```

ditto?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-:-KmjdLFkKUnxBThvW-Rx:b-lquwl1)*

sezergueler:
Review status: 3 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[python/lib/crypto/certificate_chain.py, line 85 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-Kmjd6PEIUFJSQnoT1p5:-KmjtZfc3oYWZZeZbQAE:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/045fee61017e2ce0c845624734ce7d1d73592d29/python/lib/crypto/certificate_chain.py#L85)):*
<details><summary><i>Previously, pszalach (Paweł Szałachowski) wrote…</i></summary><blockquote>

`raise SCIONParseError` ?
</blockquote></details>

Done.

---

*[python/lib/crypto/certificate_chain.py, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-KmgXAjr3FCPq1Hsfs6M:-Kmjt_DQAbmot1rFChX1:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/045fee61017e2ce0c845624734ce7d1d73592d29/python/lib/crypto/certificate_chain.py#L115)):*

Don't use a new line after the colon.
</blockquote></details>

Done.

---

*[python/lib/crypto/certificate_chain.py, line 121 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-KmgXE79l4o6u0Igl-OU:-Kmjt_fcaXf2wmiSeXPv:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/045fee61017e2ce0c845624734ce7d1d73592d29/python/lib/crypto/certificate_chain.py#L121)):*

Include the reason why the signature verification failed here as well.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 351 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-KmjdHtLManEwZ6nLBP1:-KmjtaHIuwhMcgosdG0T:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/045fee61017e2ce0c845624734ce7d1d73592d29/python/topology/generator.py#L351)):*
<details><summary><i>Previously, pszalach (Paweł Szałachowski) wrote…</i></summary><blockquote>

ditto?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1117)*

Oncilla:
Currently, we do not use this functionality (since authenticating etc is left to the AS). However, I do not see why the CertChain should be restricted to two either.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-:-Kmk_Xvm0HOlwsW8sNQN:bo1kp9k)*

sezergueler:
We need to have a separation between core AS certificates and regular AS certificates, because they are different and are signed/verified basing on different type of keys etc. So, for sure we would have the core AS certificate as an attribute that keeps the certificate. 

What one could do is to make a list out of the regular AS attribute. But I do not think that this is better for two reasons:

- We are not using chains longer than two and we do not want to allow chains longer than two. So, why should one parse longer chains and then always treat them as not valid. I think in something security critical as certificates it makes sense to have the implementation to be as precise as possible. I am not sure if generalization makes sense here.

- If one wants to add another level later on, like certificates for endhosts, it would anyway be cleaner to extend this class in a subclass instead of keeping certificates with different types in the same list.

---

Review status: 3 of 5 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-:-Kmkg6l8sXhr7I1zXq1M:b-uqvwhk)*

pszalach:
Reviewed 3 of 5 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1117)*

pszalach:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1117#-:-Kml6shHUF_C_vXIh1Np:bnfp4nl)*

PR #1127, opened 2017-06-13
Replacing Deque for segment propagation with self expiring caches

### LINKED ISSUE #1126: Change CPS propagation queues to expiring dicts
Currently when a non-master CPS gets a down-segment registration (https://github.com/netsec-ethz/scion/blob/8773e206f4529168145edb273f56c90498ad3007/python/path_server/core.py#L110) it adds it to a deque for propagation to other core ASes. But as it is not master, that queue is never serviced, and grows continuously. If we replace this deque with an expiring dict instead, using the pcb ID as the key, this would both remove the unbounded queuing behaviour, as well as remove duplicate propagations.

The deque for propagating core PCBs to the master CPS probably also needs the same treatment, as if the connection to zookeeper breaks, that queue doesn't get cleared either.

shitz:
Refactored `rev_cache.py` in something more general, such that it can also be used for segments. ExpiringDicts are not the datastructure of choice, since path segments can have different expiration times.

Fixes #1126.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1127)


shitz:
[[1] Change CPS propagation queues to expiring dicts](https://trello.com/c/7jGuTIah/24-1-change-cps-propagation-queues-to-expiring-dicts)

klausman:
:lgtm:

---

Reviewed 13 of 13 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1127#-:-KmWQFPRAnDyPknyv619:bnfp4nl)*

klausman:
Reviewed 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1127)*

kormat:
Reviewed 2 of 13 files at r1, 1 of 1 files at r3.

---

*[python/lib/cache/base.py, line 30 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1127#-Kmk8UEtdArP73lU2MKk:-Kmk8UEtdArP73lU2MKl:b-gqlb0d) ([raw file](https://github.com/netsec-ethz/scion/blob/5029e53996415b9939747186f3210e3f1f5ae3c7/python/lib/cache/base.py#L30)):*
> ```Python
>     """
>     Thread-safe cache with auto expiration of entries. Subclasses have to implement
>     their own way how to validate an entry.
> ```

"have to implement their own entry validation" maybe?

---

*[python/lib/cache/base.py, line 38 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1127#-Kmk9w0-WQdeiLHyi5GZ:-Kmk9w0-WQdeiLHyi5G_:bcnjh4k) ([raw file](https://github.com/netsec-ethz/scion/blob/5029e53996415b9939747186f3210e3f1f5ae3c7/python/lib/cache/base.py#L38)):*
> ```Python
>         self._capacity = capacity
> 
>     def __contains__(self, key):  # pragma: no cover
> ```

It's probably simpler to do: `return self.get(key) is not None`

---

*[python/lib/cache/base.py, line 47 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1127#-Kmk9S96eCaltnPpveOe:-Kmk9S96eCaltnPpveOf:b-j5sbeo) ([raw file](https://github.com/netsec-ethz/scion/blob/5029e53996415b9939747186f3210e3f1f5ae3c7/python/lib/cache/base.py#L47)):*
> ```Python
> 
>     def __len__(self):  # pragma: no cover
>         return len(self._cache)
> ```

Is this thread-safe?

---

*[python/lib/cache/base.py, line 83 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1127#-KmkAo3F8JuzCziOssQm:-KmkAo3F8JuzCziOssQn:b-l4gpz) ([raw file](https://github.com/netsec-ethz/scion/blob/5029e53996415b9939747186f3210e3f1f5ae3c7/python/lib/cache/base.py#L83)):*
> ```Python
>         calculated. In case the cache is at its capacity, the oldest entry is replaced.
> 
>         :returns: True if the entry was added, False if a newer entry was present.
> ```

It also returns false if the entry failed to validate.

---

*[python/lib/cache/base.py, line 98 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1127#-KmkCBAo4RG664wmmQ_0:-KmkCBAo4RG664wmmQ_1:b-8ug2tg) ([raw file](https://github.com/netsec-ethz/scion/blob/5029e53996415b9939747186f3210e3f1f5ae3c7/python/lib/cache/base.py#L98)):*
> ```Python
>                 # Couldn't free up enough space, evict the oldest entry.
>                 if len(self._cache) >= self._capacity:
>                     self._cache.popitem(last=False)
> ```

For paranoia, you could add `assert len(self._cache) < self._capacity` here.

---

*[python/lib/cache/seg.py, line 28 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1127#-KmkCjJND3LdwhWDEmU8:-KmkCjJND3LdwhWDEmU9:b-trlnd6) ([raw file](https://github.com/netsec-ethz/scion/blob/5029e53996415b9939747186f3210e3f1f5ae3c7/python/lib/cache/seg.py#L28)):*
> ```Python
> 
>     def _is_newer(self, seg1, seg2):
>         return seg1.get_timestamp() > seg2.get_timestamp()
> ```

If an intermediate AS sends multiple segments based on the same PCB from a core, then the latest one should "win". This code will drop any updates an AS makes if it doesn't also use a newer core PCB as the base, which is not desirable.

---

*[python/lib/cache/seg.py, line 31 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1127#-KmkD8QQ_e1FtFc2ee9I:-KmkD8QQ_e1FtFc2ee9J:bhggui) ([raw file](https://github.com/netsec-ethz/scion/blob/5029e53996415b9939747186f3210e3f1f5ae3c7/python/lib/cache/seg.py#L31)):*
> ```Python
> 
>     def _validate_entry(self, seg):
>         return seg.get_expiration_time() > time.time()
> ```

As discussed, it doesn't make sense cache segments for _their_ expiration time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1127)*

shitz:
Superseded by #1133.

PR #1133, opened 2017-06-16
Replace deques in path servers with expiring dicts

### LINKED ISSUE #1126: Change CPS propagation queues to expiring dicts

shitz:
Replace deques in path servers with expiring dicts to prevent them from growing forever. Fixes #1126.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1133)

### DIFF
--- a/python/path_server/base.py
+++ b/python/path_server/base.py
@@ -423,7 +423,10 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
         pcbs = defaultdict(list)
         while container:
             count += 1
-            _, (type_, pcb) = container.popitem(last=False)
+            try:
+                _, (type_, pcb) = container.popitem(last=False)
+            except KeyError:
+                continue
             pcbs[type_].append(pcb.copy())
             if count >= limit:
                 yield(pcbs)
@@ -476,7 +479,11 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
             return
         logging.info("Sharing %d revocation(s) via ZK", len(self._revs_to_zk))
         while self._revs_to_zk:
-            self._zk_write_rev(self._revs_to_zk.popitem(last=False)[1])
+            try:
+                data = self._revs_to_zk.popitem(last=False)[1]
+            except KeyError:
+                continue
+            self._zk_write_rev(data)
 
     def _zk_write(self, data):
         hash_ = crypto_hash(data).hex()
--- a/python/path_server/base.py
+++ b/python/path_server/base.py
@@ -19,7 +19,7 @@
 import logging
 import random
 import threading
-from collections import defaultdict, deque
+from collections import defaultdict
 from abc import ABCMeta, abstractmethod
 from threading import Lock
 
@@ -73,6 +73,8 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
     ZK_SHARE_LIMIT = 10
     # Time to store revocations in zookeeper
     ZK_REV_OBJ_MAX_AGE = HASHTREE_EPOCH_TIME
+    # TTL of segments in the queue for ZK (in seconds)
+    SEGS_TO_ZK_TTL = 10 * 60
 
     def __init__(self, server_id, conf_dir):
         """
@@ -111,8 +113,8 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
                 SCMPPathClass.REVOKED_IF: self._handle_scmp_revocation,
             },
         }
-        self._segs_to_zk = deque()
-        self._revs_to_zk = deque()
+        self._segs_to_zk = ExpiringDict(1000, self.SEGS_TO_ZK_TTL)
+        self._revs_to_zk = ExpiringDict(1000, HASHTREE_EPOCH_TIME)
         self._zkid = ZkID.from_values(self.addr.isd_as, self.id,
                                       [(self.addr.host, self._port)])
         self.zk = Zookeeper(self.topology.isd_as, PATH_SERVICE,
@@ -219,7 +221,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
             return False
         self.revocations.add(rev_info)
         logging.debug("Received revocation from %s: %s", meta, rev_info.short_desc())
-        self._revs_to_zk.append(rev_info.copy().pack())  # have to pack copy
+        self._revs_to_zk[rev_info] = rev_info.copy().pack()  # have to pack copy
         # Remove segments that contain the revoked interface.
         self._remove_revoked_segments(rev_info)
         # Forward revocation to other path servers.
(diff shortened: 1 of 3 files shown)


kormat:
Reviewed 3 of 3 files at r1.

---

*[python/path_server/base.py, line 426 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1133#-KmzJsKsG_NhygdYQ-sC:-KmzJsKtLAmR1KvmkpmD:bq67b52) ([raw file](https://github.com/netsec-ethz/scion/blob/19d66db66a8b5d893a22754fac2e3b7cb51c79d0/python/path_server/base.py#L426)):*
> ```Python
>         while container:
>             count += 1
>             _, (type_, pcb) = container.popitem(last=False)
> ```

This isn't thread-safe, add a `try`/`except` to catch `KeyError`.

---

*[python/path_server/base.py, line 479 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1133#-KmzKSkb9l5KtzAI9rDi:-KmzKSkb9l5KtzAI9rDj:b474d6a) ([raw file](https://github.com/netsec-ethz/scion/blob/19d66db66a8b5d893a22754fac2e3b7cb51c79d0/python/path_server/base.py#L479)):*
> ```Python
>         logging.info("Sharing %d revocation(s) via ZK", len(self._revs_to_zk))
>         while self._revs_to_zk:
>             self._zk_write_rev(self._revs_to_zk.popitem(last=False)[1])
> ```

Same here re: thread-safe.

---

*[python/path_server/core.py, line 49 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1133#-KmzKrtqG-zF64GPs8MP:-KmzKrtqG-zF64GPs8MQ:b-h20kfj) ([raw file](https://github.com/netsec-ethz/scion/blob/19d66db66a8b5d893a22754fac2e3b7cb51c79d0/python/path_server/core.py#L49)):*
> ```Python
>         assert self.topology.is_core_as, "This shouldn't be a local PS!"
>         self._master_id = None  # Address of master core Path Server.
>         self._segs_to_master = ExpiringDict(1000, 10)
> ```

What's the reasoning for this ttl value?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1133)*

shitz:

---

*[python/path_server/base.py, line 426 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1133#-KmzJsKsG_NhygdYQ-sC:-KmzVVi0xGSBHXlTc-H6:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/19d66db66a8b5d893a22754fac2e3b7cb51c79d0/python/path_server/base.py#L426)):*

This isn't thread-safe, add a `try`/`except` to catch `KeyError`.
</blockquote></details>

Done.

---

*[python/path_server/base.py, line 479 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1133#-KmzKSkb9l5KtzAI9rDi:-KmzVu6ZuUXt8qH3ZHYc:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/19d66db66a8b5d893a22754fac2e3b7cb51c79d0/python/path_server/base.py#L479)):*

Same here re: thread-safe.
</blockquote></details>

Done.

---

*[python/path_server/core.py, line 49 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1133#-KmzKrtqG-zF64GPs8MP:-KmzW070w6J_iBDgU4Ys:b-tyzizm) ([raw file](https://github.com/netsec-ethz/scion/blob/19d66db66a8b5d893a22754fac2e3b7cb51c79d0/python/path_server/core.py#L49)):*

What's the reasoning for this ttl value?
</blockquote></details>

`_segs_to_master` only has to keep around segments as long as it takes to detect a disconnect to ZK and reelect a new master. We discussed offline that this should be in the order of 10s.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1133)*

shitz:
[[1] Change CPS propagation queues to expiring dicts](https://trello.com/c/7jGuTIah/24-1-change-cps-propagation-queues-to-expiring-dicts)

kormat:
Reviewed 1 of 1 files at r2.

---

*[python/path_server/base.py, line 425 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1133#-Kmz_Ir7UdG0SRVu1Jxq:-Kmz_Ir7UdG0SRVu1Jxr:b-2wxsft) ([raw file](https://github.com/netsec-ethz/scion/blob/be39269ebe3bca640f9dca17104cbba4f644f3e8/python/path_server/base.py#L425)):*
> ```Python
>         pcbs = defaultdict(list)
>         while container:
>             count += 1
> ```

`count` should be moved after the try/except block, just so that it doesn't get out of sync. (It doesn't matter right now, but if this got refactored it could cause problems).

---

*[python/path_server/core.py, line 49 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1133#-KmzKrtqG-zF64GPs8MP:-KmzaG-gTgqucpeCwmRN:b-836gv8) ([raw file](https://github.com/netsec-ethz/scion/blob/19d66db66a8b5d893a22754fac2e3b7cb51c79d0/python/path_server/core.py#L49)):*

`_segs_to_master` only has to keep around segments as long as it takes to detect a disconnect to ZK and reelect a new master. We discussed offline that this should be in the order of 10s.
</blockquote></details>

Ahh, because `_sync_master` will cover longer outages. Ok :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1133)*

shitz:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[python/path_server/base.py, line 425 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1133#-Kmz_Ir7UdG0SRVu1Jxq:-KmzdNHE3Qrl8E1eAY4k:boj1qgt) ([raw file](https://github.com/netsec-ethz/scion/blob/be39269ebe3bca640f9dca17104cbba4f644f3e8/python/path_server/base.py#L425)):*

`count` should be moved after the try/except block, just so that it doesn't get out of sync. (It doesn't matter right now, but if this got refactored it could cause problems).
</blockquote></details>

Good catch. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1133)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1133#-:-Kmziiv_kT0o_dC951xE:bnfp4nl)*

PR #1139, opened 2017-06-23
Set -y flag to prevent user interaction

ErebosM:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1139)


kormat:
This is specifically not desired. If a user wants to achieve this, they would do `APTARGS=-y ./env/debian/deps`, for example.

ErebosM:
Ah I see. Thanks!

PR #1159, opened 2017-07-07
Fix memory leak in PS and BS.

shitz:
A flaw in how segment verification was handled when receiving path segments through Zookeeper lead to a memory leak.

On startup of a server it can happen that the first path segments come from Zookeeper. If not all certificates are available the server requests it from the local CS with the cache only flag set. However, the CS doesn't have this certificate yet, which lead to a rerequesting loop. This loop also prevented requesting the certs from the appropriate beacon servers when new path segments were arriving.

Fixed the issue by having newer path segments always overwrite older ones in the unverified segments map.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1159)

### DIFF
--- a/python/lib/msg_meta.py
+++ b/python/lib/msg_meta.py
@@ -50,10 +50,6 @@ class MetadataBase(object):
     def __str__(self):
         return "%s:%d" % (self.get_addr(), self.port)
 
-    def __eq__(self, other):
-        return self.host == other.host and self.port == other.port and (
-            (not self.ia and not other.ia) or (self.ia and other.ia and self.ia == other.ia))
-
 
 class UDPMetadata(MetadataBase):
     """
--- a/python/lib/msg_meta.py
+++ b/python/lib/msg_meta.py
@@ -51,7 +51,8 @@ class MetadataBase(object):
         return "%s:%d" % (self.get_addr(), self.port)
 
     def __eq__(self, other):
-        return self.ia == other.ia and self.host == other.host and self.port == other.port
+        return self.host == other.host and self.port == other.port and (
+            (not self.ia and not other.ia) or (self.ia and other.ia and self.ia == other.ia))
 
 
 class UDPMetadata(MetadataBase):
--- a/python/lib/msg_meta.py
+++ b/python/lib/msg_meta.py
@@ -50,6 +50,9 @@ class MetadataBase(object):
     def __str__(self):
         return "%s:%d" % (self.get_addr(), self.port)
 
+    def __eq__(self, other):
+        return self.ia == other.ia and self.host == other.host and self.port == other.port
+
 
 class UDPMetadata(MetadataBase):
     """
--- a/python/lib/path_seg_meta.py
+++ b/python/lib/path_seg_meta.py
@@ -19,7 +19,7 @@
 import threading
 
 
-class PathSegMeta(object):
+class PathSegMeta(object):  # pragma: no cover
     """
     The PathSegMeta class holds missing trcs and certificates and
     necessary metadata for a path segment.
@@ -36,6 +36,7 @@ class PathSegMeta(object):
         self.meta = meta
         self.type = type_
         self.params = params
+        self.id = seg.get_hops_hash()
 
     def verifiable(self):
         with self.miss_cert_lock and self.miss_trc_lock:
--- a/python/scion_elem/scion_elem.py
+++ b/python/scion_elem/scion_elem.py
@@ -195,7 +195,7 @@ class SCIONElement(object):
             self._DefaultMeta = TCPMetadata
         else:
             self._DefaultMeta = UDPMetadata
-        self.unverified_segs = {}
+        self.unverified_segs = ExpiringDict(500, 60 * 60)
         self.unv_segs_lock = threading.RLock()
         self.requested_trcs = {}
         self.req_trcs_lock = threading.Lock()
--- a/python/scion_elem/scion_elem.py
+++ b/python/scion_elem/scion_elem.py
@@ -25,6 +25,7 @@ import time
 from collections import defaultdict
 
 # External packages
+from external.expiring_dict import ExpiringDict
 from prometheus_client import Counter, Gauge, start_http_server
 
 # SCION
@@ -427,8 +428,8 @@ class SCIONElement(object):
         for isd, ver in missing_trcs:
             with self.req_trcs_lock:
                 _, meta = self.requested_trcs.get((isd, ver), (None, None))
-                # Prefer requesting from remote BS.
-                if meta and (not seg_meta.meta or seg_meta.meta == self._get_cs()):
+                # There is already an outstanding request for the missing TRC.
+                if meta:
                     continue
             trc_req = TRCRequest.from_values(ISD_AS.from_values(isd, 0), ver, cache_only=True)
             meta = seg_meta.meta or self._get_cs()
@@ -439,7 +440,7 @@ class SCIONElement(object):
             logging.info("Requesting %sv%s TRC from %s, for PCB %s",
                          isd, ver, meta, seg_meta.seg.short_id())
             with self.req_trcs_lock:
-                self.requested_trcs[(isd, ver)] = (time.time(), meta)
+                self.requested_trcs[(isd, ver)] = (time.time(), seg_meta.meta)
                 if self._labels:
                     PENDING_TRC_REQS_TOTAL.labels(**self._labels).set(len(self.requested_trcs))
             self.send_meta(trc_req, meta)
@@ -459,8 +460,8 @@ class SCIONElement(object):
         for isd_as, ver in missing_certs:
             with self.req_certs_lock:
                 _, meta = self.requested_certs.get((isd_as, ver), (None, None))
-                # Prefer requesting from remote BS.
-                if meta and (not seg_meta.meta or seg_meta.meta == self._get_cs()):
+                # There is already an outstanding request for the missing TRC.
+                if meta:
                     continue
             cert_req = CertChainRequest.from_values(isd_as, ver, cache_only=True)
             meta = seg_meta.meta or self._get_cs()
@@ -471,7 +472,7 @@ class SCIONElement(object):
             logging.info("Requesting %sv%s CERTCHAIN from %s for PCB %s",
                          isd_as, ver, meta, seg_meta.seg.short_id())
             with self.req_certs_lock:
-                self.requested_certs[(isd_as, ver)] = (time.time(), meta)
+                self.requested_certs[(isd_as, ver)] = (time.time(), seg_meta.meta)
                 if self._labels:
                     PENDING_CERT_REQS_TOTAL.labels(**self._labels).set(len(self.requested_certs))
             self.send_meta(cert_req, meta)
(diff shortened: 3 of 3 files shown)


ercanucan:
@shitz what's the status on this PR? is it going to be merged soon so that we can deploy?

kormat:
Reviewed 1 of 3 files at r1, 1 of 1 files at r2.
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[python/scion_elem/scion_elem.py, line 193 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL2Xk1hvuTHmZoIsiR:-KpL2Xk1hvuTHmZoIsiS:b-ykp28a) ([raw file](https://github.com/netsec-ethz/scion/blob/c6b4df5461c0a8c5745e22305f7716b54f4c7e86/python/scion_elem/scion_elem.py#L193)):*
> ```Python
>         else:
>             self._DefaultMeta = UDPMetadata
>         self.unverified_segs = {}
> ```

It probably makes sense to make this an expiring dict, say with the TTL of 1h. Any sgement that you get that can't be verified within an hour is highly unlikely to be useful. This will also prevent any segments that are never verifiable from staying around permanently.

---

*[python/scion_elem/scion_elem.py, line 422 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL3ljcvXveOVGEgUXF:-KpL3ljcvXveOVGEgUXG:b-dxgo2g) ([raw file](https://github.com/netsec-ethz/scion/blob/c6b4df5461c0a8c5745e22305f7716b54f4c7e86/python/scion_elem/scion_elem.py#L422)):*
> ```Python
>             with self.req_trcs_lock:
>                 _, meta = self.requested_trcs.get((isd, ver), (None, None))
>                 # Prefer requesting from remote BS.
> ```

"remote" and "BS" assumptions are incorrect. I'm also not sure what you're trying to do, here.

---

*[python/scion_elem/scion_elem.py, line 431 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL4MRbQhM61P76Jyua:-KpL4MRbQhM61P76Jyub:b-nbf72f) ([raw file](https://github.com/netsec-ethz/scion/blob/70dfac84c0ef5ab1cddbdf5c027f7dc77390fa51/python/scion_elem/scion_elem.py#L431)):*
> seg_meta.meta == self._get_cs()

That's.. a bizarre check. What's the intention?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1159)*

shitz:
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[python/scion_elem/scion_elem.py, line 193 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL2Xk1hvuTHmZoIsiR:-KpLQ8k4wpofyx1SADG8:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/c6b4df5461c0a8c5745e22305f7716b54f4c7e86/python/scion_elem/scion_elem.py#L193)):*

It probably makes sense to make this an expiring dict, say with the TTL of 1h. Any sgement that you get that can't be verified within an hour is highly unlikely to be useful. This will also prevent any segments that are never verifiable from staying around permanently.
</blockquote></details>

Done.

---

*[python/scion_elem/scion_elem.py, line 422 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL3ljcvXveOVGEgUXF:-KpLT-vvSUOpFT5CSPW5:b-8jw2g5) ([raw file](https://github.com/netsec-ethz/scion/blob/c6b4df5461c0a8c5745e22305f7716b54f4c7e86/python/scion_elem/scion_elem.py#L422)):*

"remote" and "BS" assumptions are incorrect. I'm also not sure what you're trying to do, here.
</blockquote></details>

Where is it requested from then?

---

*[python/scion_elem/scion_elem.py, line 431 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL4MRbQhM61P76Jyua:-KpLSaRy07VSw_QRF4LP:b87pmr9) ([raw file](https://github.com/netsec-ethz/scion/blob/70dfac84c0ef5ab1cddbdf5c027f7dc77390fa51/python/scion_elem/scion_elem.py#L431)):*

> seg_meta.meta == self._get_cs()

That's.. a bizarre check. What's the intention?
</blockquote></details>

`requested_trcs` contains the outstanding trc requests. The problem is that if the path segment that triggered the TRC request was from zookeeper then `meta` will point to the local CS. Thus you want to issue a new request `seg_meta.meta` does not point to the local CS.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1159)*

kormat:
Reviewed 1 of 1 files at r3.
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[python/scion_elem/scion_elem.py, line 422 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL3ljcvXveOVGEgUXF:-KpOyOa2T3jIoaLP6i9h:b-ixmnsw) ([raw file](https://github.com/netsec-ethz/scion/blob/c6b4df5461c0a8c5745e22305f7716b54f4c7e86/python/scion_elem/scion_elem.py#L422)):*

Where is it requested from then?
</blockquote></details>

E.g.:
- path server receives up-segment from a local BS.
- path server receives any segment from remote PS during lookup.

---

*[python/scion_elem/scion_elem.py, line 431 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL4MRbQhM61P76Jyua:-KpOysWcjH5T4r5MPhve:b-h3c1st) ([raw file](https://github.com/netsec-ethz/scion/blob/70dfac84c0ef5ab1cddbdf5c027f7dc77390fa51/python/scion_elem/scion_elem.py#L431)):*
> The problem is that if the path segment that triggered the TRC request was from zookeeper then meta will point to the local CS.

That's a bug. If there is no original meta in the seg_meta, then one should not be added for the local CS. Also you're assuming there's only a single local CS instance, which is bad.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1159)*

shitz:
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[python/scion_elem/scion_elem.py, line 422 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL3ljcvXveOVGEgUXF:-KpPfdzM7VIT-onqgCIe:b-fi6yci) ([raw file](https://github.com/netsec-ethz/scion/blob/c6b4df5461c0a8c5745e22305f7716b54f4c7e86/python/scion_elem/scion_elem.py#L422)):*

E.g.:
- path server receives up-segment from a local BS.
- path server receives any segment from remote PS during lookup.
</blockquote></details>

Changed comment entirely.

---

*[python/scion_elem/scion_elem.py, line 431 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL4MRbQhM61P76Jyua:-KpPfgxCp_TZmGCOQcuy:b-u2e77x) ([raw file](https://github.com/netsec-ethz/scion/blob/70dfac84c0ef5ab1cddbdf5c027f7dc77390fa51/python/scion_elem/scion_elem.py#L431)):*

> The problem is that if the path segment that triggered the TRC request was from zookeeper then meta will point to the local CS.

That's a bug. If there is no original meta in the seg_meta, then one should not be added for the local CS. Also you're assuming there's only a single local CS instance, which is bad.
</blockquote></details>

Changed according to offline discussion.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1159)*

kormat:
Reviewed 1 of 2 files at r4.

---

*[python/scion_elem/scion_elem.py, line 431 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL4MRbQhM61P76Jyua:-KpPtyPvlWqf6HH1pNoM:bxj7r35) ([raw file](https://github.com/netsec-ethz/scion/blob/70dfac84c0ef5ab1cddbdf5c027f7dc77390fa51/python/scion_elem/scion_elem.py#L431)):*

Changed according to offline discussion.
</blockquote></details>

This breaks in `_check_trc_reqs`, as it assumes that `requested_trcs` always contains a useful `meta`.

---

*[python/scion_elem/scion_elem.py, line 431 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpPu6TRaImQNSgxi7KZ:-KpPu6TRaImQNSgxi7K_:b-jqrji3) ([raw file](https://github.com/netsec-ethz/scion/blob/5339f76ee374694759c30b05fa135c59936eb3a0/python/scion_elem/scion_elem.py#L431)):*
> ```Python
>             with self.req_trcs_lock:
>                 _, meta = self.requested_trcs.get((isd, ver), (None, None))
>                 # There is already an outstanding request for the missing TRC.
> ```

This comment should be inside the `if` clause, so it's scope is unambiguous.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1159)*

shitz:
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[python/scion_elem/scion_elem.py, line 431 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpL4MRbQhM61P76Jyua:-KpPuywEtkP6teFXhfkP:boj1qgt) ([raw file](https://github.com/netsec-ethz/scion/blob/70dfac84c0ef5ab1cddbdf5c027f7dc77390fa51/python/scion_elem/scion_elem.py#L431)):*

This breaks in `_check_trc_reqs`, as it assumes that `requested_trcs` always contains a useful `meta`.
</blockquote></details>

Good catch. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1159)*

kormat:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1159)*

shitz:

---

*[python/scion_elem/scion_elem.py, line 431 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-KpPu6TRaImQNSgxi7KZ:-KpPvTU_DARFPPUVEpsk:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5339f76ee374694759c30b05fa135c59936eb3a0/python/scion_elem/scion_elem.py#L431)):*

This comment should be inside the `if` clause, so it's scope is unambiguous.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1159)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-:-KpPverj5T3EACIicE6x:bnfp4nl)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1159#-:-KpQFxVtC_6a814jmAgR:bnfp4nl)*

PR #1161, opened 2017-07-10
Separate retrieved TRCs and certs in separate directory

klausman:
To avoid deleting cached data from `gen/` and thus triggering restarts
via Ansible, move those files to a separate subdirectory, `gen-cache`.
Also fix `scion.sh` to nuke and (re)make the directory if the topology
is (re)generated.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1161)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ Thumbs.db
 /logs/*
 !/logs/.keepme
 /gen/
+/gen-cache/
 /traces/
 
 # Python generated files #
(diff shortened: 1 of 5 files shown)


klausman:
@kormat

kormat:
Elegant solution :)

:lgtm: 

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1161#-:-KogwQqJVdfwVgXwLka9:bbswrvp)*

PR #1168, opened 2017-07-23
Make SIG use an alternate routing table.

kormat:
This allows the main forwarding traffic to be easily switched to/away
from the SIG (by changing an ip rule). This also allows SIG healthchecks
to always use the SIG, so it can be checked even when it's not
forwarding the main traffic.

Also:
- Add bin/sig to .gitignore

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1168)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,7 @@ bin/border
 bin/discovery
 bin/dispatcher
 bin/logdog
+bin/sig
 c/dispatcher/dispatcher
 c/ssp/test/*client
 c/ssp/test/*server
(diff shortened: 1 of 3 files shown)


scrye:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1168#-:-KpibuhKxnYmXOlt-MdD:bnfp4nl)*

PR #1172, opened 2017-07-23
Reimplemented ingress worker

shitz:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1172)


kormat:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[go/sig/base/ingress.go, line 30 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpjZv0i71BD43B9aBXS:-KpjZv0i71BD43B9aBXT:b5badv0) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L30)):*
> reassemble

---

*[go/sig/base/ingress.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpj_NhX_Eceg0j8uzoj:-Kpj_NhX_Eceg0j8uzok:bm9v38s) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L35)):*
> ```Go
> 	seqNr          int
> 	index          int
> 	len            int
> ```

`frameLen` might be less ambigious.

---

*[go/sig/base/ingress.go, line 36 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpj_AhTeEKp6pf_oq7X:-Kpj_AhUOHyvX47gci77:b-lgytpr) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L36)):*
> ```Go
> 	index          int
> 	len            int
> 	endProcessed   bool
> ```

I'd move this after `startProcessed`, so they're grouped.

Ah. Hmm. The names are somewhat counter-intuitive. I think you're using `end` to mean there's the _end_ of a packet fragment at the _start_ of the frame. This gets even murkier then a frame contains only a middle fragment of a packet. `nonStartProcessed` might be clearer. Or, maybe, `fragNonStart bool; fragStartOffset int;`. The former is always true if `index` is 0, and otherwise false. When the fragment is processed, set the bool to false. `fragStartOffset` can be set to 0 when that fragment is processed.

---

*[go/sig/base/ingress.go, line 38 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpj_1eRP9vHvNoLVTSu:-Kpj_1eRP9vHvNoLVTSv:bdexas9) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L38)):*
> ```Go
> 	endProcessed   bool
> 	hasStart       bool
> 	pktStart       int
> ```

I think `hasStart` is probably redundant, if you can just do `pktStart != 0` instead?

---

*[go/sig/base/ingress.go, line 41 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpj_EVGITpgftvRz8UB:-Kpj_EVHHuXqR_Fl6gJJ:b-ybljjg) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L41)):*
> ```Go
> 	startProcessed bool
> 	pktLen         int
> 	raw            []byte
> ```

Use `common.RawBytes` everywhere instead of `[]byte`, for easier logging.

---

*[go/sig/base/ingress.go, line 314 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpje17y9wdZQoSAqZ7p:-Kpje17y9wdZQoSAqZ7q:b-zbf9a3) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L314)):*
> ```Go
> // IngressWorker handles decapsulation of SIG frames. There is one IngressWorker per
> // remote SIG.
> func IngressWorker(scionNet *scion.SCIONNet, listenAddr addr.HostAddr, listenPort uint16) {
> ```

The order of stuff in this file is hard to follow. Suggested order: 
- `FrameBuf` (could also go after `IngressWorker`)
-`IngressWorker`
- `IngressState`
- `processFrame`
- `processPkts`
- `ReassemblyList`

---

*[go/sig/base/ingress.go, line 393 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpjfJdN27yNgMMjNVkk:-KpjfJdN27yNgMMjNVkl:b-r2i6j1) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L393)):*
> ```Go
> 		offset += 2
> 		rawPkt := frame.raw[offset:frame.len]
> 		if len(rawPkt) <= pktLen {
> ```

Uh. Why would having less than the packet length be "we got everything"? That should probably be `>=`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1172)*

kormat:
Review status: 1 of 2 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/sig/base/ingress.go, line 402 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpjgLJaYwTKzZTmKLSS:-KpjgLJaYwTKzZTmKLST:b-16uo13) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L402)):*
> ```Go
> 			continue
> 		}
> 		// There is an incomplete packet at the end of the frame.
> ```

It would be cleaner to do `if len(rawPkt) < pktLen { break}; // We got everything...` and move the block below outside the for loop.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1172)*

kormat:
Review status: 1 of 2 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[go/sig/base/ingress.go, line 332 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpjoPQytTaOitxEbBvl:-KpjoPQytTaOitxEbBvm:b3cdpfy) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L332)):*
> ```Go
> 			go state.cleanUp()
> 		default:
> 			frame := state.getFrameBuf()
> ```

You get a framebuf here, but you don't always release it. E.g. if you hit the read error below, it doesn't get released. I think `processFrame` also doesn't release it. The cleanest is just to always release it in this method. 

```
if err != nil {
  log.Error...
  goto cleanup
}
...
cleanup:
   bufPool <- frame
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1172)*

shitz:
Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[go/sig/base/ingress.go, line 30 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpjZv0i71BD43B9aBXS:-KpjjQzb46hSBDSuMprQ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L30)):*

> reassemble
</blockquote></details>

Done.

---

*[go/sig/base/ingress.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpj_NhX_Eceg0j8uzoj:-Kpjkd2XvU7NiMvzEz6R:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L35)):*

`frameLen` might be less ambigious.
</blockquote></details>

Done.

---

*[go/sig/base/ingress.go, line 36 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpj_AhTeEKp6pf_oq7X:-Kpjuv1pVkO1_jA0QMcK:b90bgrn) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L36)):*

I'd move this after `startProcessed`, so they're grouped.

Ah. Hmm. The names are somewhat counter-intuitive. I think you're using `end` to mean there's the _end_ of a packet fragment at the _start_ of the frame. This gets even murkier then a frame contains only a middle fragment of a packet. `nonStartProcessed` might be clearer. Or, maybe, `fragNonStart bool; fragStartOffset int;`. The former is always true if `index` is 0, and otherwise false. When the fragment is processed, set the bool to false. `fragStartOffset` can be set to 0 when that fragment is processed.
</blockquote></details>

Not sure if this would be more intuitive. I replaced every `end` with `trailingFrag` and every `start` with `leadingFrag`

---

*[go/sig/base/ingress.go, line 38 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpj_1eRP9vHvNoLVTSu:-KpjknzqGz7UPLq605yQ:boxdwf) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L38)):*

I think `hasStart` is probably redundant, if you can just do `pktStart != 0` instead?
</blockquote></details>

Was only there to make it consistent, ie., a flag for both. Removed both anyway.

---

*[go/sig/base/ingress.go, line 41 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpj_EVGITpgftvRz8UB:-KpjlRQgIXXtJrQ-KVor:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L41)):*

Use `common.RawBytes` everywhere instead of `[]byte`, for easier logging.
</blockquote></details>

Done.

---

*[go/sig/base/ingress.go, line 314 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpje17y9wdZQoSAqZ7p:-Kpjm1G6l9QRg1zuBQZs:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L314)):*

The order of stuff in this file is hard to follow. Suggested order: 
- `FrameBuf` (could also go after `IngressWorker`)
-`IngressWorker`
- `IngressState`
- `processFrame`
- `processPkts`
- `ReassemblyList`
</blockquote></details>

Done.

---

*[go/sig/base/ingress.go, line 332 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpjoPQytTaOitxEbBvl:-KpjxrnMYTmc4JJOI2kp:blc1ng6) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L332)):*

You get a framebuf here, but you don't always release it. E.g. if you hit the read error below, it doesn't get released. I think `processFrame` also doesn't release it. The cleanest is just to always release it in this method. 

```
if err != nil {
  log.Error...
  goto cleanup
}
...
cleanup:
   bufPool <- frame
```
</blockquote></details>

Good catch. I release the frame now in `processFrame` if it doesn't have to be added to the reassembly list and in the `IngressWorker` after a read error.

---

*[go/sig/base/ingress.go, line 393 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpjfJdN27yNgMMjNVkk:-Kpjm2VX3f099bGwYjfT:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L393)):*

Uh. Why would having less than the packet length be "we got everything"? That should probably be `>=`.
</blockquote></details>

Done.

---

*[go/sig/base/ingress.go, line 402 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpjgLJaYwTKzZTmKLSS:-KpjonrHyVIutJ6QIP84:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/30e89d1afddd0647e80deb90fcfdb318a1929a2d/go/sig/base/ingress.go#L402)):*

It would be cleaner to do `if len(rawPkt) < pktLen { break}; // We got everything...` and move the block below outside the for loop.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1172)*

kormat:
Reviewed 1 of 2 files at r2.
Review status: 1 of 2 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[go/sig/base/forward.go, line 150 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpk2aSzutyzXGQxgo7n:-Kpk2aSzutyzXGQxgo7o:b7se8y4) ([raw file](https://github.com/netsec-ethz/scion/blob/d2e8013e83014ded8567e00fec8925546271c680/go/sig/base/forward.go#L150)):*
> ```Go
> 	}
> 	// New packets always starts at a 8 byte boundary.
> 	e.frameOff = pad(e.frameOff)
> ```

Noo. This has to happen _before_ the index is set.

---

*[go/sig/base/ingress.go, line 133 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpk3q-DO1rnER-fkSjq:-Kpk3q-DO1rnER-fkSjr:b-j90omz) ([raw file](https://github.com/netsec-ethz/scion/blob/d2e8013e83014ded8567e00fec8925546271c680/go/sig/base/ingress.go#L133)):*
> ```Go
> 
> func (s *IngressState) getFrameBuf() *FrameBuf {
> 	buf := <-s.bufPool
> ```

This can be simplified to `return <-s.bufPool`. Note that it _will_ block if there's no buffers available.

---

*[go/sig/base/ingress.go, line 154 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpk49BgX2lqfj9cXg00:-Kpk49BgX2lqfj9cXg01:bjdwdv5) ([raw file](https://github.com/netsec-ethz/scion/blob/d2e8013e83014ded8567e00fec8925546271c680/go/sig/base/ingress.go#L154)):*
> ```Go
> 			log.Debug("Removing reassembly list", "epoch", epoch)
> 			rlist.removeAll()
> 			delete(s.reassemblyLists, epoch)
> ```

This is racy as there is no locking. What would be safe is:
- don't call `cleanUp` as a go routine
- change the order here:
```
delete(s.reassemblyLists, epoch)
go rlist.removeAll()
```
The expensive part is then done in the go routine, after it no longer touches shared state.

---

*[go/sig/base/ingress.go, line 338 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpkBu4Ipvprsh3aXHi9:-KpkBu4Ipvprsh3aXHiA:b-a6qdnb) ([raw file](https://github.com/netsec-ethz/scion/blob/d2e8013e83014ded8567e00fec8925546271c680/go/sig/base/ingress.go#L338)):*
> ```Go
> 		if currFrame.seqNr == prevSeqNr+1 {
> 			prevSeqNr = currFrame.seqNr
> 			// Add number of bytes contained in this frame.
> ```

I think this can be simplified to:
```
bytes += currFrame.len - SIGHdrSize
if bytes >= fromFrame.pktLen {
  to = e
  break
}
if currFrame.index != 0 {
  // a framing error has occurred. we've consumed all the bytes in this frame, there aren't enough for our current packet, and another packet starts in this frame.
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1172)*

kormat:
Reviewed 1 of 2 files at r2.

---

*[go/sig/base/ingress.go, line 376 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpkD0h8IWJkm08sx_Ck:-KpkD0h8IWJkm08sx_Cl:bjl2xeb) ([raw file](https://github.com/netsec-ethz/scion/blob/d2e8013e83014ded8567e00fec8925546271c680/go/sig/base/ingress.go#L376)):*
> ```Go
> 	// Collect rest.
> 	var next *list.Element
> 	for e := from.Next(); l.buf.Len() < fromFrame.pktLen && e != to.Next(); e = next {
> ```

`< pktLen`? Seeing as you already have a variable for it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1172)*

shitz:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/sig/base/forward.go, line 150 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpk2aSzutyzXGQxgo7n:-KpkQ-7Vpla4E2gzojJd:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d2e8013e83014ded8567e00fec8925546271c680/go/sig/base/forward.go#L150)):*

Noo. This has to happen _before_ the index is set.
</blockquote></details>

Done.

---

*[go/sig/base/ingress.go, line 133 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpk3q-DO1rnER-fkSjq:-KpkQ8kPwSb9zqp2GJUP:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d2e8013e83014ded8567e00fec8925546271c680/go/sig/base/ingress.go#L133)):*

This can be simplified to `return <-s.bufPool`. Note that it _will_ block if there's no buffers available.
</blockquote></details>

Done.

---

*[go/sig/base/ingress.go, line 154 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-Kpk49BgX2lqfj9cXg00:-KpkQGV5LgPxvspq62RJ:bmswhn2) ([raw file](https://github.com/netsec-ethz/scion/blob/d2e8013e83014ded8567e00fec8925546271c680/go/sig/base/ingress.go#L154)):*

This is racy as there is no locking. What would be safe is:
- don't call `cleanUp` as a go routine
- change the order here:
```
delete(s.reassemblyLists, epoch)
go rlist.removeAll()
```
The expensive part is then done in the go routine, after it no longer touches shared state.
</blockquote></details>

Nice solution! Thought I would need to use locks for this.

---

*[go/sig/base/ingress.go, line 338 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpkBu4Ipvprsh3aXHi9:-KpkUmuC22QnSPatI1IF:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d2e8013e83014ded8567e00fec8925546271c680/go/sig/base/ingress.go#L338)):*

I think this can be simplified to:
```
bytes += currFrame.len - SIGHdrSize
if bytes >= fromFrame.pktLen {
  to = e
  break
}
if currFrame.index != 0 {
  // a framing error has occurred. we've consumed all the bytes in this frame, there aren't enough for our current packet, and another packet starts in this frame.
}
```
</blockquote></details>

Done.

---

*[go/sig/base/ingress.go, line 376 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-KpkD0h8IWJkm08sx_Ck:-KpkUs69x0nOfsKcnSL0:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d2e8013e83014ded8567e00fec8925546271c680/go/sig/base/ingress.go#L376)):*

`< pktLen`? Seeing as you already have a variable for it.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1172)*

kormat:
Reviewed 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1172)*

kormat:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-:-KpkbOa0XHvHSW25Eu0a:bnfp4nl)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1172#-:-Kpkbf_bl4au2ZLBLVBE:bnfp4nl)*

PR #1185, opened 2017-07-26
Fix stale dependency handling for govendor.

kormat:
If you're on a branch that requires go packages installed that aren't
required by master, when you switch back to master those packages aren't
removed, and can then cause build failures. This change will check for
this issue and work around it, but it's slightly brute-force: any
unneeded packages are also removed from govendor's cache.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1185)

### DIFF
--- a/go/Makefile
+++ b/go/Makefile
@@ -36,6 +36,12 @@ deps_proto: deps proto
 deps: vendor/.deps.stamp ../proto/go.capnp
 
 vendor/.deps.stamp: vendor/vendor.json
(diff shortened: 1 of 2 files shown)


klausman:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1185#-:-Kpz59aTeXvupELVwjYb:bnfp4nl)*

PR #1192, opened 2017-08-02
Multiple message Reads/Writes for ReliableSock

scrye:
This PR optimizes ReliableSock reads and writes by allowing multiple slices to be read/written in a single syscall via the new functions `WriteN` and `ReadN`.


shitz:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[go/lib/sock/reliable/appaddr.go, line 38 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqY1SME4xfNB-Hid-cX:-KqY1SME4xfNB-Hid-cY:bct7ad5) ([raw file](https://github.com/netsec-ethz/scion/blob/e76397383c861b8e7a801d7fb82463e79aa70c58/go/lib/sock/reliable/appaddr.go#L38)):*
> ```Go
> }
> 
> func ParseAppAddr(buf common.RawBytes, addrType addr.HostAddrType) (*AppAddr, error) {
> ```

Call this `AppAddrFromRaw`

---

*[go/lib/sock/reliable/appaddr.go, line 68 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kqb6h2vYrdU-4VD8b-H:-KqY1_34dLhcIFx7RWpI:bbj1w30) ([raw file](https://github.com/netsec-ethz/scion/blob/e76397383c861b8e7a801d7fb82463e79aa70c58/go/lib/sock/reliable/appaddr.go#L68)):*
> ```Go
> }
> 
> func init() {
> ```

I prefer having the `init` method at the beginning of a file, since it's run first and it usually initializes variables declared in a toplevel `var` block.

---

*[go/lib/sock/reliable/reliable.go, line 80 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kqb6ssli2GLelCE0uba:-Kqb6ssli2GLelCE0ubb:b-179u36) ([raw file](https://github.com/netsec-ethz/scion/blob/e76397383c861b8e7a801d7fb82463e79aa70c58/go/lib/sock/reliable/reliable.go#L80)):*
> ```Go
> const (
> 	regCommandField = 0x03 // Register command (0x01) with SCMP enabled (0x02)
> 	dfltBufSize     = 1 << 18
> ```

`dflt` looks weird. 

---

*[go/lib/sock/reliable/reliable.go, line 94 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbIjQSsjicCvGi4MsH:-KqbIjQSsjicCvGi4MsI:b-1461f4) ([raw file](https://github.com/netsec-ethz/scion/blob/e76397383c861b8e7a801d7fb82463e79aa70c58/go/lib/sock/reliable/reliable.go#L94)):*
> ```Go
> 	sendBuf   []byte
> 	recvBuf   []byte
> 	header    []byte
> ```

Why is `header` separate? It doesn't really make sense, since this is an object for a connection, whereas a header is packet specific. From what I can tell by looking at the code, you could remove `header` and write directly to `sendBuf` instead. (Would also avoid an extra copy of the header)

---

*[go/lib/sock/reliable/reliable.go, line 102 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kqb7OO0zOyubyE1MZth:-Kqb7OO0zOyubyE1MZti:b8igm4u) ([raw file](https://github.com/netsec-ethz/scion/blob/e76397383c861b8e7a801d7fb82463e79aa70c58/go/lib/sock/reliable/reliable.go#L102)):*
> ```Go
> 	c, err := net.DialTimeout("unix", address, timeout)
> 	if err != nil {
> 		return nil, common.NewError("Unable to connect", "address", address)
> ```

Include the original error in the context.

---

*[go/lib/sock/reliable/reliable.go, line 322 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbEpxXWGShPNhkXP70:-KqbEpxXWGShPNhkXP71:bre2r5f) ([raw file](https://github.com/netsec-ethz/scion/blob/e76397383c861b8e7a801d7fb82463e79aa70c58/go/lib/sock/reliable/reliable.go#L322)):*
> ```Go
> 			return 0, err
> 		}
> 		if n != 0 {
> ```

`n > 0` is more appropriate

---

*[go/lib/sock/reliable/reliable.go, line 338 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbG8ylr4ySAeZC0yb8:-KqbG8ylr4ySAeZC0yb9:b-v4x7v2) ([raw file](https://github.com/netsec-ethz/scion/blob/e76397383c861b8e7a801d7fb82463e79aa70c58/go/lib/sock/reliable/reliable.go#L338)):*
> ```Go
> // WriteN copies data sequentially from each buf slice contained in msgs.
> // Destination address information can be specified in the addr field of each
> // message. If the address is nil, then a HostTypeNone address is assumed. The
> ```

Is there a case for which the host addr is `None`? I think I remember that we wanted to get rid of `HostAddrNone`

---

*[go/lib/sock/reliable/reliable.go, line 374 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbGaqrx6MT0ZHi-V-H:-KqbGaqrx6MT0ZHi-V-I:b6o2l0a) ([raw file](https://github.com/netsec-ethz/scion/blob/e76397383c861b8e7a801d7fb82463e79aa70c58/go/lib/sock/reliable/reliable.go#L374)):*
> ```Go
> }
> 
> func (conn *Conn) computeHeader(buf []byte, dst *AppAddr) error {
> ```

`copyHeader` or `fillHeader` would be better names

---

*[go/lib/sock/reliable/reliable.go, line 374 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbHhXl2l6pOic-OCAn:-KqbHhXl2l6pOic-OCAo:bewdjk2) ([raw file](https://github.com/netsec-ethz/scion/blob/e76397383c861b8e7a801d7fb82463e79aa70c58/go/lib/sock/reliable/reliable.go#L374)):*
> ```Go
> }
> 
> func (conn *Conn) computeHeader(buf []byte, dst *AppAddr) error {
> ```

It seems that only `len(buf)` is needed here which you can pass in directly as `pldLen` or something

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 17 unresolved discussions.

---

*[go/lib/sock/reliable/appaddr.go, line 32 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr19OzdWszV4zNAgrNV:-Kr19OzdWszV4zNAgrNW:biedj0e) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/appaddr.go#L32)):*
> ```Go
> 
> func (a *AppAddr) Pack() []byte {
> 	buf := make([]byte, 0, a.Len())
> ```

Mmph. For something that we're doing a Lot, these small allocations are bad. It would be better to have `Write()` methods for AppAddr and HostAddr, and supply a buffer for them to write to.

---

*[go/lib/sock/reliable/reliable.go, line 80 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kqb6ssli2GLelCE0uba:-Kr0toqTZNEtYToc3aHs:b3f085g) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L80)):*

`dflt` looks weird. 
</blockquote></details>

`defBufSize` would be more normal.  

---

*[go/lib/sock/reliable/reliable.go, line 254 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr126pocW97sxl805rl:-Kr126ppOlWkjipadwOg:b-iex546) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L254)):*
> ```Go
> 
> 	// Peek to see if packet complete
> 	peekData := conn.recvBuf[conn.readHead:conn.writeHead]
> ```

`readHead` and `writeHead` are pretty misleading names when it turns out that they are only relevant to the `recvBuf`, and not the `sendBuf`.

---

*[go/lib/sock/reliable/reliable.go, line 264 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr14vNOcXhNmLgTQt-0:-Kr14vNOcXhNmLgTQt-1:b-f6lbub) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L264)):*
> ```Go
> 		return false, common.NewError("ReliableSock protocol desynchronized", "conn", conn)
> 	}
> 	rcvdAddrType := addr.HostAddrType(header[len(cookie)])
> ```

It would read a bit easier to do:
```
peekOffset += len(cookie)
recvAddrType := addr.HostAddrType(header[peekOffset])
peekOffset +=1
...
```

---

*[go/lib/sock/reliable/reliable.go, line 306 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr0wWcXnzywll2VT_zc:-Kr0wWcXnzywll2VT_zd:bf0sk9n) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L306)):*
> ```Go
> // of bytes is always len(buf).
> func (conn *Conn) Write(buf []byte) (int, error) {
> 	a, _ := addr.HostFromRaw(nil, addr.HostTypeNone)
> ```

`NilAppAddr`?

---

*[go/lib/sock/reliable/reliable.go, line 338 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbG8ylr4ySAeZC0yb8:-Kr1-dM_5FJJXhRsDGTo:b-e7qjhg) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L338)):*

Is there a case for which the host addr is `None`? I think I remember that we wanted to get rid of `HostAddrNone`
</blockquote></details>

@shitz - AIUI things like sciond's API uses it, as the address isn't relevant.

---

*[go/lib/sock/reliable/reliable.go, line 339 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1-TLsyOFCSfLDTWpx:-Kr1-TLsyOFCSfLDTWpy:bws3gzv) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L339)):*
> returns immediately if writing the next packet would block
That doesn't appear to be true, afaict. It's also not knowable.

---

*[go/lib/sock/reliable/reliable.go, line 342 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr11fArp-oY-0Es04sj:-Kr11fArp-oY-0Es04sk:b-tup70w) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L342)):*
> ```Go
> // does not guarantee that at least one packet is written.  Each buffer is
> // copied in its entirety (similarly to datagram oriented protocols); the
> // copied field of struct Msg is reset to 0 for written packets.  WriteN
> ```

This field isn't accessible to users of this library, as it is not exported.

---

*[go/lib/sock/reliable/reliable.go, line 362 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr0znpbVWFTEu8QSav0:-Kr0znpbVWFTEu8QSav1:b-8lq1l1) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L362)):*
> we'll rarely block on writes anyway
[citation needed]?

---

*[go/lib/sock/reliable/reliable.go, line 396 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr0ycVYWsd29hu_u2Dj:-Kr0ycVYWsd29hu_u2Dk:bn7ht39) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L396)):*
> ```Go
> 	offset += 4
> 	copy(conn.header[offset:], dst.Pack())
> 	offset += dst.Len()
> ```

This isn't used.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
Review status: 2 of 3 files reviewed at latest revision, 18 unresolved discussions.

---

*[go/lib/sock/reliable/reliable_test.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1IAvuGP0zARf2-_IH:-Kr1IAvuGP0zARf2-_II:byfcxer) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L207)):*
> ```Go
> }
> 
> func getRandFile() string {
> ```

https://golang.org/pkg/io/ioutil/#TempFile

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

scrye:
Review status: 0 of 3 files reviewed at latest revision, 18 unresolved discussions, some commit checks failed.

---

*[go/lib/sock/reliable/appaddr.go, line 32 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr19OzdWszV4zNAgrNV:-Kr5Zrs87hCDDo3vMFZz:b-b4jvx7) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/appaddr.go#L32)):*

Mmph. For something that we're doing a Lot, these small allocations are bad. It would be better to have `Write()` methods for AppAddr and HostAddr, and supply a buffer for them to write to.
</blockquote></details>

Agreed. `HostAddr.Pack()`for IPv4 and IPv6 does not allocate new buffers, so we can keep using it for now.

Using `Write()` as the name for these functions is confusing to me, because the Go standard library uses `structure.Write(buffer)` to write _from_ the buffer _to_ the structure. I propose we use `PutRaw()` (or something else that's different from `Write`).

---

*[go/lib/sock/reliable/appaddr.go, line 38 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqY1SME4xfNB-Hid-cX:-Kr5b4St5PXFsA91IKTb:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/appaddr.go#L38)):*

Call this `AppAddrFromRaw`
</blockquote></details>

Done.

---

*[go/lib/sock/reliable/appaddr.go, line 68 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kqb6h2vYrdU-4VD8b-H:-Kr5b62XLM3nvBsNe5fV:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/appaddr.go#L68)):*

I prefer having the `init` method at the beginning of a file, since it's run first and it usually initializes variables declared in a toplevel `var` block.
</blockquote></details>

Done.

---

*[go/lib/sock/reliable/reliable.go, line 80 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kqb6ssli2GLelCE0uba:-Kr5bCZ3ZgAeMHZT9YZs:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L80)):*

`defBufSize` would be more normal.  
</blockquote></details>

Done.

---

*[go/lib/sock/reliable/reliable.go, line 94 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbIjQSsjicCvGi4MsH:-Kr5ci31cayIwtXK9mUK:bxw0dt3) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L94)):*

Why is `header` separate? It doesn't really make sense, since this is an object for a connection, whereas a header is packet specific. From what I can tell by looking at the code, you could remove `header` and write directly to `sendBuf` instead. (Would also avoid an extra copy of the header)
</blockquote></details>

Good point, it's an artifact from a previous version. The headers are now built directly on top of `sendBuf`.

---

*[go/lib/sock/reliable/reliable.go, line 102 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kqb7OO0zOyubyE1MZth:-Kr5mRDvyG7TLR3DYiF5:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L102)):*

Include the original error in the context.
</blockquote></details>

Done.

---

*[go/lib/sock/reliable/reliable.go, line 254 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr126pocW97sxl805rl:-Kr5miauXgDEXTgZikUl:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L254)):*

`readHead` and `writeHead` are pretty misleading names when it turns out that they are only relevant to the `recvBuf`, and not the `sendBuf`.
</blockquote></details>

Done.

---

*[go/lib/sock/reliable/reliable.go, line 264 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr14vNOcXhNmLgTQt-0:-Kr5okS9ngjC_NhWiBnh:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L264)):*

It would read a bit easier to do:
```
peekOffset += len(cookie)
recvAddrType := addr.HostAddrType(header[peekOffset])
peekOffset +=1
...
```
</blockquote></details>

Done.

---

*[go/lib/sock/reliable/reliable.go, line 306 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr0wWcXnzywll2VT_zc:-Kr5p7sfAyfcbtZf69Sv:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L306)):*

`NilAppAddr`?
</blockquote></details>

Done.

---

*[go/lib/sock/reliable/reliable.go, line 322 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbEpxXWGShPNhkXP70:-Kr5pG_zJEW9rKdJCReC:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L322)):*

`n > 0` is more appropriate
</blockquote></details>

Done.

---

*[go/lib/sock/reliable/reliable.go, line 339 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1-TLsyOFCSfLDTWpx:-Kr5pS-diGTc4PpdvSQa:b73nowb) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L339)):*

> returns immediately if writing the next packet would block
That doesn't appear to be true, afaict. It's also not knowable.
</blockquote></details>

The text here is referring to blocking when copying to the internal buffer. Since `WriteN` always flushes that buffer at the end, it's kinda misleading. I reworded the docstring to give a better overview of how the function works internally.

(as a side note, you can tell when you'd block via `EAGAIN/EWOULDBLOCK` errors and configuring sockets with `O_NONBLOCK`)

---

*[go/lib/sock/reliable/reliable.go, line 342 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr11fArp-oY-0Es04sj:-Kr5rWgm8ns2VQ8jRnkh:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L342)):*

This field isn't accessible to users of this library, as it is not exported.
</blockquote></details>

Done.

---

*[go/lib/sock/reliable/reliable.go, line 362 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr0znpbVWFTEu8QSav0:-Kr5rrBfM6YrNO3aL8bs:b-1uzype) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L362)):*

> we'll rarely block on writes anyway
[citation needed]?
</blockquote></details>

Blocking on writes for datagram oriented protocols only happens when the reader (which can be another app, like the dispatcher in our case, or the kernel) is unable to process the data in the buffer fast enough. This only happens during reader failures or when the reader is slower than the writer during high traffic.

This is much rarer than for `reads`, which often block due to unavailable data (it's more common to be under low load than high load) or write blocks for connection oriented protocols which need acknowledgments from across the network to continue emptying the buffer.

---

*[go/lib/sock/reliable/reliable.go, line 374 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbGaqrx6MT0ZHi-V-H:-Kr5sfZgO8DDI-Xauon1:bs04p2w) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L374)):*

`copyHeader` or `fillHeader` would be better names
</blockquote></details>

Function deleted.

---

*[go/lib/sock/reliable/reliable.go, line 374 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbHhXl2l6pOic-OCAn:-Kr5siJOgt5xBu41amxQ:bs04p2w) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L374)):*

It seems that only `len(buf)` is needed here which you can pass in directly as `pldLen` or something
</blockquote></details>

Function deleted.

---

*[go/lib/sock/reliable/reliable.go, line 396 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr0ycVYWsd29hu_u2Dj:-Kr5srWZw6fre5UyF2up:bs04p2w) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L396)):*

This isn't used.
</blockquote></details>

Function deleted.

---

*[go/lib/sock/reliable/reliable_test.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1IAvuGP0zARf2-_IH:-Kr5tUOdjkxQX9wnUqNZ:b-tak5u4) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L207)):*

https://golang.org/pkg/io/ioutil/#TempFile
</blockquote></details>

Sadly, this does not work for UNIX Domain Sockets.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
Reviewed 1 of 3 files at r2.
Review status: 1 of 3 files reviewed at latest revision, 20 unresolved discussions.

---

*[go/lib/sock/reliable/appaddr.go, line 32 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr19OzdWszV4zNAgrNV:-Kr5yzgmK4YnxybkrtzX:b7m8uz) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/appaddr.go#L32)):*

Agreed. `HostAddr.Pack()`for IPv4 and IPv6 does not allocate new buffers, so we can keep using it for now.

Using `Write()` as the name for these functions is confusing to me, because the Go standard library uses `structure.Write(buffer)` to write _from_ the buffer _to_ the structure. I propose we use `PutRaw()` (or something else that's different from `Write`).
</blockquote></details>

That's a fair point re: naming, it looks like `WriteTo` is a more standard name for this in the std lib. (net/http seems to be an exception, {Header,Response,Request}.Write all take an `io.Writer`).

That said, we currently have many examples of `.Write()` for this in the current codebase (`lib.addr.ISD_AS`, 'lib.spkt.CmnHdr`, `lib.common.Extension`, `lib.l4.L4Header`, `lib.common.Payload`, etc), so it would be good to stay consistent with them, and maybe rename all of them in the future.

---

*[go/lib/sock/reliable/appaddr.go, line 23 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr61WmGCCtUS1lETkCo:-Kr61WmGCCtUS1lETkCp:b-6ml8af) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/appaddr.go#L23)):*
> ```Go
> }
> 
> func (a *AppAddr) PutRaw(buf common.RawBytes) (int, error) {
> ```

The Go convention (or at least the one we're using) is for the order to be:
- `type ...`
- constructors (e.g. `AppAddrFromRaw`)
- methods

---

*[go/lib/sock/reliable/appaddr.go, line 26 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr61sO29dJN0LM75uEj:-Kr61sO29dJN0LM75uEk:bzgd16f) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/appaddr.go#L26)):*
> ```Go
> 	if len(buf) < a.Len() {
> 		return 0, common.NewError("Unable to write AppAddr, insufficient size",
> 			"have", len(buf), "want", a.Len())
> ```

The terms we use elsewhere are "expected" and "actual".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
Reviewed 1 of 3 files at r2.
Review status: 2 of 3 files reviewed at latest revision, 18 unresolved discussions.

---

*[go/lib/sock/reliable/reliable.go, line 339 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1-TLsyOFCSfLDTWpx:-Kr6GA7EwQvlE6BKYzRd:b8pblgr) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L339)):*

The text here is referring to blocking when copying to the internal buffer. Since `WriteN` always flushes that buffer at the end, it's kinda misleading. I reworded the docstring to give a better overview of how the function works internally.

(as a side note, you can tell when you'd block via `EAGAIN/EWOULDBLOCK` errors and configuring sockets with `O_NONBLOCK`)
</blockquote></details>

Wording looks better, thanks.

(That's not the same as knowing you can write a full message without blocking. That tells you can you write 1B without blocking, it doesn't tell you if you can write 1500B without blocking, for example.)

---

*[go/lib/sock/reliable/reliable.go, line 362 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr0znpbVWFTEu8QSav0:-Kr6KoI3p_uogGPDGFv_:b-pl88ol) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L362)):*
> Blocking on writes for datagram oriented protocols only happens when the reader (which can be another app, like the dispatcher in our case, or the kernel) is unable to process the data in the buffer fast enough.

I don't think that's data-gram specific.

> This only happens ... when the reader is slower than the writer during high traffic.

This is _precisely_ what we're concerned about, which is why the comment struck me as odd.

---

*[go/lib/sock/reliable/reliable.go, line 343 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr6KEe2H_9XV6BVUUH3:-Kr6KEe2H_9XV6BVUUH4:b-1zf4ql) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/reliable.go#L343)):*
>  WriteN returns the number of packets written.

This kinda sucks for callers who want to send a lot of packets. Each call site would need to loop over calls to `WriteN` to ensure everything is sent. With a bit of refactoring `WriteN` could just write everything its been given, and only return when everything is sent, or if an error happens.

---

*[go/lib/sock/reliable/reliable.go, line 347 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr6IEE9FeL2NAnaNWzG:-Kr6IEE9FeL2NAnaNWzH:b-s3f22p) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/reliable.go#L347)):*
> ```Go
> 	copiedMsgs := 0
> 	index := 0
> 	for _, msg := range bufs {
> ```

The contents of this for loop could be moved to a separate method, for improved readability, and just loop over `bufs` calling the method on each member. This would also allow WriteN to write more than 1 buffer's worth per invocation.

---

*[go/lib/sock/reliable/reliable.go, line 362 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr6HlRP1SvCTtzhbXBt:-Kr6HlRP1SvCTtzhbXBu:b2chga4) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/reliable.go#L362)):*
> ```Go
> 		default:
> 			return 0, common.NewError("Unknown address type", "type", dst.Addr.Type())
> 		}
> ```

It looks like this can just be reduced to `destLen := dst.Len()`? I'm not too concerned about other host addr types being invented but not being valid here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

scrye:
Review status: 0 of 3 files reviewed at latest revision, 17 unresolved discussions.

---

*[go/lib/sock/reliable/appaddr.go, line 32 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr19OzdWszV4zNAgrNV:-Kr6lcfNWke-49NeSDUw:b-5hdta) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/appaddr.go#L32)):*

That's a fair point re: naming, it looks like `WriteTo` is a more standard name for this in the std lib. (net/http seems to be an exception, {Header,Response,Request}.Write all take an `io.Writer`).

That said, we currently have many examples of `.Write()` for this in the current codebase (`lib.addr.ISD_AS`, 'lib.spkt.CmnHdr`, `lib.common.Extension`, `lib.l4.L4Header`, `lib.common.Payload`, etc), so it would be good to stay consistent with them, and maybe rename all of them in the future.
</blockquote></details>

`WriteTo` is also somewhat iffy, because in `net.Conn` it means write from buffer to connection using a specified address as the destination.

If we have a better name, renaming everything can be done pretty easily; the vim Go plugin I'm using, for example, does renames across all files at once.

---

*[go/lib/sock/reliable/appaddr.go, line 23 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr61WmGCCtUS1lETkCo:-Kr6kqk3HDVkkMxYYdCq:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/appaddr.go#L23)):*

The Go convention (or at least the one we're using) is for the order to be:
- `type ...`
- constructors (e.g. `AppAddrFromRaw`)
- methods
</blockquote></details>

Done.

---

*[go/lib/sock/reliable/appaddr.go, line 26 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr61sO29dJN0LM75uEj:-Kr6ky8ZBpxvHY10x_wx:b4bvjzv) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/appaddr.go#L26)):*

The terms we use elsewhere are "expected" and "actual".
</blockquote></details>

Yes, but they feel a bit... off, here. It's not really the expected value, it's more like a maximum. I tried to rephrase the error to make it sound better.

---

*[go/lib/sock/reliable/reliable.go, line 362 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr0znpbVWFTEu8QSav0:-Kr6x9nEUyWV7K4s7MnP:bsmq7pg) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L362)):*

> Blocking on writes for datagram oriented protocols only happens when the reader (which can be another app, like the dispatcher in our case, or the kernel) is unable to process the data in the buffer fast enough.

I don't think that's data-gram specific.

> This only happens ... when the reader is slower than the writer during high traffic.

This is _precisely_ what we're concerned about, which is why the comment struck me as odd.


</blockquote></details>

Well, connection oriented protocols can block on writes if the ACKs are not arriving. Datagrams will always be flushed right away, so it's a matter of overpowering the reader.

If we overpower the reader, there's not much we can do. To preserve freshness, we'll probably want to drop leftovers anyway (I remember you suggested this a while ago and we agreed that discarding what cannot bet written will help us from forwarding information that is _too_ stale).

---

*[go/lib/sock/reliable/reliable.go, line 343 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr6KEe2H_9XV6BVUUH3:-Kr6mRN9iAKaVwIPLJb1:b-wtjcil) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/reliable.go#L343)):*

>  WriteN returns the number of packets written.

This kinda sucks for callers who want to send a lot of packets. Each call site would need to loop over calls to `WriteN` to ensure everything is sent. With a bit of refactoring `WriteN` could just write everything its been given, and only return when everything is sent, or if an error happens.
</blockquote></details>

Yes, but this approach is a bit more flexible. Some writers might prefer to try to write 1000 msgs and if only 400 go through without blocking, do some more work and then try to write the remaining 600 and 300 newly processed ones. This also allows the writer to discard packets that are stale (e.g., was I only able to write 200 packets out of my 800? then drop the extra 600 and send only fresh ones next time). TCP traffic might prefer the first one and VoIP traffic the second one. In any case, it's better to have the option than hardcode the library internally.

Granted, this is not the case now because the function _does_ block internally. But this is a necessary evil for now and will no longer be the case when we have `sendmmsg` and `recvmmsg` there.

---

*[go/lib/sock/reliable/reliable.go, line 347 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr6IEE9FeL2NAnaNWzG:-Kr6va-revxIWVzQELRJ:b5zox7p) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/reliable.go#L347)):*

The contents of this for loop could be moved to a separate method, for improved readability, and just loop over `bufs` calling the method on each member. This would also allow WriteN to write more than 1 buffer's worth per invocation.
</blockquote></details>

Done.

Also, it's not a problem if `WriteN` writes only one buffer per invocation, because it's the syscall that's the actual cost and writing multiple buffers inevitably means multiple syscalls. The caller will just have to call the function again, and that's cheap. Also, if the buffer is smaller than the Unix socket buffer we decrease the probability we would block. It is a hacky heuristic which is not always correct, but atm it should work ok and it is also better than what we have on `master`.

Optimally, this function will be refactored in the future to have the following behavior:
  - get N messages
  - do non-blocking write on the unix socket for all of them in a single syscall
  - see how many were sent
  - drop packets that it could not send (or, better yet, leave this to the caller)
  - return how many packets were sent

This will be very clean and feels close to the best we can achieve in userspace.

To get there we'll need `sendmmsg`, `fcntl` to enable non-blocking operation and `unix-gram` support to simplify parsing.

---

*[go/lib/sock/reliable/reliable.go, line 362 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr6HlRP1SvCTtzhbXBt:-Kr6ofV57qxYTlKqVSqj:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/reliable.go#L362)):*

It looks like this can just be reduced to `destLen := dst.Len()`? I'm not too concerned about other host addr types being invented but not being valid here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
Reviewed 2 of 2 files at r3.
Review status: 2 of 3 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[go/lib/sock/reliable/reliable.go, line 362 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr0znpbVWFTEu8QSav0:-KrAkV_Hzulb4fp-cKwi:bxe6t85) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L362)):*
> To preserve freshness, we'll probably want to drop leftovers anyway (I remember you suggested this a while ago and we agreed that discarding what cannot bet written will help us from forwarding information that is too stale).

Where this gets tricky is that it's very hard to know that this is the right thing to do. Example: the socket buffer can hold 100 msgs, and is currently full. WriteN gets called with a single buffer, which it can't write currently, and so simple dropping logic would just throw it away, despite that often being the wrong thing to do. We don't have enough information to make a good decision.

---

*[go/lib/sock/reliable/reliable.go, line 343 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr6KEe2H_9XV6BVUUH3:-KrAFw1BH1t02BlqqM0W:b-24gtud) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/reliable.go#L343)):*

Yes, but this approach is a bit more flexible. Some writers might prefer to try to write 1000 msgs and if only 400 go through without blocking, do some more work and then try to write the remaining 600 and 300 newly processed ones. This also allows the writer to discard packets that are stale (e.g., was I only able to write 200 packets out of my 800? then drop the extra 600 and send only fresh ones next time). TCP traffic might prefer the first one and VoIP traffic the second one. In any case, it's better to have the option than hardcode the library internally.

Granted, this is not the case now because the function _does_ block internally. But this is a necessary evil for now and will no longer be the case when we have `sendmmsg` and `recvmmsg` there.
</blockquote></details>

A simple fix would be to pass in a `bool` that says whether `WriteN` should write everything or not. Any non-performance-oriented caller would set it to `true` and gets simple semantics, and anything which wants more control sets it to `false` and can make more informed decisions.

---

*[go/lib/sock/reliable/reliable.go, line 347 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr6IEE9FeL2NAnaNWzG:-KrAJ_jhEBc-iTdF3a4f:b-9fe2jd) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/reliable.go#L347)):*

Done.

Also, it's not a problem if `WriteN` writes only one buffer per invocation, because it's the syscall that's the actual cost and writing multiple buffers inevitably means multiple syscalls. The caller will just have to call the function again, and that's cheap. Also, if the buffer is smaller than the Unix socket buffer we decrease the probability we would block. It is a hacky heuristic which is not always correct, but atm it should work ok and it is also better than what we have on `master`.

Optimally, this function will be refactored in the future to have the following behavior:
  - get N messages
  - do non-blocking write on the unix socket for all of them in a single syscall
  - see how many were sent
  - drop packets that it could not send (or, better yet, leave this to the caller)
  - return how many packets were sent

This will be very clean and feels close to the best we can achieve in userspace.

To get there we'll need `sendmmsg`, `fcntl` to enable non-blocking operation and `unix-gram` support to simplify parsing.
</blockquote></details>

I'm not talking about the efficiency, i'm talking about the complexity cost for most callers which won't care about these issues. Alternatively to the `bool` suggestion i made above, we could have a `WriteNAll` method that just calls `WriteN` repeatedly until everything it has been given is written.

---

*[go/lib/sock/reliable/reliable.go, line 405 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KrAl7lL_doeXObVcLQz:-KrAl7lL_doeXObVcLR-:b-qob90p) ([raw file](https://github.com/netsec-ethz/scion/blob/bd12a60195dc6a5225ed167cc0da9ad1dcc87e30/go/lib/sock/reliable/reliable.go#L405)):*
> ```Go
> }
> 
> func (conn *Conn) copyMsg(msg *Msg, index, copiedMsgs int) (indexNew, copiedMsgsNew int, err error) {
> ```

This is a `Conn` method, it should be with the rest of the Conn methods.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
Review status: 2 of 3 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/lib/sock/reliable/reliable_test.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1IAvuGP0zARf2-_IH:-KrAmWqn6NLB3diS2-Ub:bv371ns) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L207)):*

Sadly, this does not work for UNIX Domain Sockets.
</blockquote></details>

Doh, right. It would be better then to use a temp dir (with a descriptive name :P), and have the domain sockets inside it.

---

*[go/lib/sock/reliable/reliable_test.go, line 278 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KrAm1YjAKS7FMYWmToD:-KrAm1YkbXsSx7bQDLRI:b-8fhlj2) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L278)):*
> ```Go
> 
> // Run 1000 writes individually
> func BenchmarkReadWrite(b *testing.B) {
> ```

Is this really a benchmark? All of these take >1s/op on my machine:
```
BenchmarkReadWrite-12               1000           1772373 ns/op
BenchmarkWriteRead-12               1000           1745416 ns/op
BenchmarkReadNWriteN-12             2000           1168009 ns/op
BenchmarkWriteNReadN-12             1000           1203258 ns/op
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
Reviewed 1 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
Review status: 2 of 3 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/lib/sock/reliable/reliable_test.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1IAvuGP0zARf2-_IH:-KrAn_MEz0M0dO5elKd6:b-vn5ax8) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L207)):*

Doh, right. It would be better then to use a temp dir (with a descriptive name :P), and have the domain sockets inside it.
</blockquote></details>

Hmm, why is this needed actually?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

scrye:
Review status: 2 of 3 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/lib/sock/reliable/reliable.go, line 362 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr0znpbVWFTEu8QSav0:-KrAnDmKU6DzatVQx-Lj:b8tc7ap) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L362)):*

> To preserve freshness, we'll probably want to drop leftovers anyway (I remember you suggested this a while ago and we agreed that discarding what cannot bet written will help us from forwarding information that is too stale).

Where this gets tricky is that it's very hard to know that this is the right thing to do. Example: the socket buffer can hold 100 msgs, and is currently full. WriteN gets called with a single buffer, which it can't write currently, and so simple dropping logic would just throw it away, despite that often being the wrong thing to do. We don't have enough information to make a good decision.
</blockquote></details>

I'd argue that for a lossy protocol you want to drop that one packet. What benefit does keeping it give you in a best effort case during congestion?

---

*[go/lib/sock/reliable/reliable.go, line 347 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr6IEE9FeL2NAnaNWzG:-KrAnSNUzBlaoPQ7E0CM:b-y51f6u) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/reliable.go#L347)):*

I'm not talking about the efficiency, i'm talking about the complexity cost for most callers which won't care about these issues. Alternatively to the `bool` suggestion i made above, we could have a `WriteNAll` method that just calls `WriteN` repeatedly until everything it has been given is written.
</blockquote></details>

`WriteNAll` is a good solution I think.

The SIG might want to care about that complexity though, to reduce jitter and latency (packets arrive much later if you block on writes and decide).

Imagine an extreme case where the SIG has a 10MB internal buffer. That buffer gets full, and the reader to which we're sending is slow. Packets at the end of the 10MB buffer will be very late (possibly too late), while we're dropping fresh packets due to our own receive queue getting full.

This won't come up in tests though, because the writer will be slower than the reader.

---

*[go/lib/sock/reliable/reliable_test.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1IAvuGP0zARf2-_IH:-KrAoRSL2S3EaNJieGGF:b-zemwh9) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L207)):*

Hmm, why is this needed actually?
</blockquote></details>

To let the OS clean up all the files eventually (or fail on open).

---

*[go/lib/sock/reliable/reliable_test.go, line 278 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KrAm1YjAKS7FMYWmToD:-KrAmyIL5EPCAixWa6lN:b-pdmm3p) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L278)):*

Is this really a benchmark? All of these take >1s/op on my machine:
```
BenchmarkReadWrite-12               1000           1772373 ns/op
BenchmarkWriteRead-12               1000           1745416 ns/op
BenchmarkReadNWriteN-12             2000           1168009 ns/op
BenchmarkWriteNReadN-12             1000           1203258 ns/op
```
</blockquote></details>

One `op` there is writing 1000 messages. Also, that's 1ms. That means about 1Mpps.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

scrye:
Review status: 1 of 3 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/sock/reliable/reliable.go, line 343 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr6KEe2H_9XV6BVUUH3:-KrAvGWUk-NvmqFu6xUA:blru2o9) ([raw file](https://github.com/netsec-ethz/scion/blob/4c937735e61d4ca161c6591d365abfd82b943bd7/go/lib/sock/reliable/reliable.go#L343)):*

A simple fix would be to pass in a `bool` that says whether `WriteN` should write everything or not. Any non-performance-oriented caller would set it to `true` and gets simple semantics, and anything which wants more control sets it to `false` and can make more informed decisions.
</blockquote></details>

I went with the `WriteNAll` helper function suggestion.

---

*[go/lib/sock/reliable/reliable.go, line 405 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KrAl7lL_doeXObVcLQz:-KrAtUaEMGaR8g6q3jkp:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/bd12a60195dc6a5225ed167cc0da9ad1dcc87e30/go/lib/sock/reliable/reliable.go#L405)):*

This is a `Conn` method, it should be with the rest of the Conn methods.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
Reviewed 1 of 3 files at r2, 1 of 1 files at r4.

---

*[go/lib/sock/reliable/reliable_test.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1IAvuGP0zARf2-_IH:-KrAzBZ4n4p9FtBn_sgU:b-sdnbpq) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L207)):*

To let the OS clean up all the files eventually (or fail on open).
</blockquote></details>

The OS doesn't clean up unix domain sockets, they persist unless specifically removed.

---

*[go/lib/sock/reliable/reliable_test.go, line 278 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KrAm1YjAKS7FMYWmToD:-KrAn6O-Om3w8IKUiB9P:bimlegz) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L278)):*

One `op` there is writing 1000 messages. Also, that's 1ms. That means about 1Mpps.
</blockquote></details>

Ohh, damnit. Off by a factor of 1000, oops :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

scrye:

---

*[go/lib/sock/reliable/reliable_test.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1IAvuGP0zARf2-_IH:-KrBZpmK6RJTYn7KCwX-:b-rnsjk3) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L207)):*

The OS doesn't clean up unix domain sockets, they persist unless specifically removed.
</blockquote></details>

The test socket files are under `/tmp` so I guess they get cleaned up daily? Not 100% sure.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

scrye:

---

*[go/lib/sock/reliable/reliable.go, line 338 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KqbG8ylr4ySAeZC0yb8:-KrBaajq9vrbMQTnPXmW:b-50toif) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable.go#L338)):*

@shitz - AIUI things like sciond's API uses it, as the address isn't relevant.
</blockquote></details>

@kormat - Exactly, `None` type addresses are used when talking with SCIOND over `ReliableSock`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:

---

*[go/lib/sock/reliable/reliable_test.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1IAvuGP0zARf2-_IH:-KrBkZzu8hPOD-ieO12K:b-ag3418) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L207)):*

The test socket files are under `/tmp` so I guess they get cleaned up daily? Not 100% sure.
</blockquote></details>

That's entirely machine-specific. Many will only wipe on reboot, many not even then. Putting them under a single dir makes it easier to clean everything up afterwards.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

scrye:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/sock/reliable/reliable_test.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1IAvuGP0zARf2-_IH:-KrBmhjKHzJNkKQmIihk:b-mljqd7) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L207)):*

That's entirely machine-specific. Many will only wipe on reboot, many not even then. Putting them under a single dir makes it easier to clean everything up afterwards.
</blockquote></details>

I agree, that way we don't pollute the root of `/tmp`.

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

scrye:
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/sock/reliable/reliable_test.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-Kr1IAvuGP0zARf2-_IH:-KrFvFT_3ERAfMGqn2GP:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/34fee0b721d4077bb0bc341b752c72265ff24fc6/go/lib/sock/reliable/reliable_test.go#L207)):*

`getRandFile` hasn't been changed. It looks like `getRandFile` could take the changes you made in `TestWriteTo`, and return a random file within the tmp dir. It could then be called from `TestWriteTo` as well.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
Reviewed 1 of 1 files at r6.

---

*[go/lib/sock/reliable/reliable_test.go, line 105 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KrFwY9ITME3cEQtAmJN:-KrFwY9ITME3cEQtAmJO:b-x6ryf3) ([raw file](https://github.com/netsec-ethz/scion/blob/554d036e41470716372f7c1cdbbf06edf0328c27/go/lib/sock/reliable/reliable_test.go#L105)):*
> ```Go
> 
> func TestWriteTo(t *testing.T) {
> 	// Create dir for test sockets if it does not exist
> ```

Abandoned comment :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

scrye:
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/lib/sock/reliable/reliable_test.go, line 105 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-KrFwY9ITME3cEQtAmJN:-KrFx9Xs0y2t3gtPOnRe:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/554d036e41470716372f7c1cdbbf06edf0328c27/go/lib/sock/reliable/reliable_test.go#L105)):*

Abandoned comment :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1192#-:-KrG15JUtClJiPDnn5fN:bnfp4nl)*

PR #1198, opened 2017-08-07
Improve checksum performance for dispatcher.

kormat:
- Instead of allocating a new chk_input struct for each packet inside
  `scion_udp_checksum()`, have the caller allocate one and pass it in
  repeatedly. This saves a Large number of mallocs/frees.
- Simplify `checksum()`:
  - The checksum algorithm is already endian-agnostic, so remove all
    byte-swapping.
  - With a 32-bit accumulator, overflow is avoided, removing the need
    for checks.

Some simple benchmarks of `checksum()` put this at >2x faster than
the existing implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1198)

### DIFF
--- a/c/dispatcher/Makefile
+++ b/c/dispatcher/Makefile
@@ -1,7 +1,7 @@
 .PHONY: all clean install uninstall
 
 CC = clang
-CFLAGS ?= -Wall -Werror -g
+CFLAGS ?= -Wall -Werror -g -O2
 LDFLAGS ?= -lscion -lfilter -lpthread -lzlog -ltcpmw -llwip
 
 LIB_DIR = ../lib
--- a/c/dispatcher/dispatcher.c
+++ b/c/dispatcher/dispatcher.c
@@ -117,6 +117,8 @@ static int data_v4_socket;
 static int data_v6_socket;
 static int app_socket;
 
+static chk_input *chk_udp_input;
+
 static zlog_category_t *zc;
 
 void handle_signal(int signal);
@@ -199,6 +201,8 @@ int main(int argc, char **argv)
         zlog_fatal(zc, "setrlimit(): %s", strerror(errno));
         return -1;
     }
+    /* Allocate for later use */
+    chk_udp_input = mk_chk_input(UDP_CHK_INPUT_SIZE);
 
     zlog_info(zc, "dispatcher with zlog starting up");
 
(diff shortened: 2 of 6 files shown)


klausman:
:lgtm:

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1198#-:-KqxVxOpVLrseiw1Aziv:bnfp4nl)*

scrye:
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[c/lib/scion/checksum.c, line 40 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1198#-Kr-zHky_-Q7Uaw-MoiT:-Kr-zHky_-Q7Uaw-MoiU:bt2ro91) ([raw file](https://github.com/netsec-ethz/scion/blob/574ec54011425bddbd4a741f79a25014e06d8778/c/lib/scion/checksum.c#L40)):*
> ```C
> 
>     // Return ones-complement.
>     return ~sum;
> ```

The semantics for LE machines are a bit muddled here; this is a leftover from how the dispatcher works internally with the checksum values (and also how the old checksum worked). The returned uint16 has the checksum encoded in network byte order, and not host byte order. Since the dispatcher is coded with this in mind, this is not an issue.

A nasty side effect is that if you extract the High Byte and Low Byte with bit shifts afterwards you get wrong values. If you only `memcmp` and `memcpy` it's ok.

Scapy, for example, does things in reverse.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1198)*

scrye:
:lgtm:

---

Reviewed 5 of 6 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1198#-:-Kr00qyvPHxM3KGK9XuE:bnfp4nl)*

kormat:
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[c/lib/scion/checksum.c, line 40 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1198#-Kr-zHky_-Q7Uaw-MoiT:-Kr03Q0ujHkeAk5AWpdR:b-g5w9r4) ([raw file](https://github.com/netsec-ethz/scion/blob/574ec54011425bddbd4a741f79a25014e06d8778/c/lib/scion/checksum.c#L40)):*

The semantics for LE machines are a bit muddled here; this is a leftover from how the dispatcher works internally with the checksum values (and also how the old checksum worked). The returned uint16 has the checksum encoded in network byte order, and not host byte order. Since the dispatcher is coded with this in mind, this is not an issue.

A nasty side effect is that if you extract the High Byte and Low Byte with bit shifts afterwards you get wrong values. If you only `memcmp` and `memcpy` it's ok.

Scapy, for example, does things in reverse.

</blockquote></details>

We _could_ do a byteswap before returning, but it then means that callers need to do another byteswap to do a comparison, so i don't see a real gain. Added a comment to say that this is in network byte order.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1198)*

scrye:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1198)*

PR #1201, opened 2017-08-08
New build flag to disable debug logging

scrye:
The command line logging level disables writing out to files but does not disable log processing. This severely impacts performance.

This PR adds a new build flag `debug`. If the flag is not specified, all calls to `log.Debug` are deleted by the compiler.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1201)

### DIFF
--- a/go/lib/log/const_off.go
+++ b/go/lib/log/const_off.go
@@ -1,0 +1,19 @@
+// Copyright 2017 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !debug
+
+package liblog
+
+const DebugOn = false
--- a/go/lib/log/const_on.go
+++ b/go/lib/log/const_on.go
@@ -1,4 +1,4 @@
-// Copyright 2016 ETH Zurich
+// Copyright 2017 ETH Zurich
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/go/lib/log/const_on.go
+++ b/go/lib/log/const_on.go
@@ -1,0 +1,19 @@
+// Copyright 2016 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build debug
+
+package liblog
+
+const DebugOn = true
--- a/go/lib/log/log.go
+++ b/go/lib/log/log.go
@@ -88,3 +88,25 @@ func PanicLog() {
 func Flush() {
 	logBuf.Flush()
 }
+
+func Debug(msg string, ctx ...interface{}) {
+	if DebugOn {
+		log.Debug(msg, ctx...)
+	}
+}
+
+func Info(msg string, ctx ...interface{}) {
+	log.Info(msg, ctx...)
+}
+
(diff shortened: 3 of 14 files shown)


kormat:
I would prefer to instead remove debug log statements that aren't useful.  `git grep 'log\.Debug' -- go/sig` shows no useful debug statements in the fast path, they should just be removed. The only time they would be needed is if we were trying to debug a problem with the encap protocol, which seems to be pretty solid now, and adding specific debug statements for that case wouldn't be hard.

Additionally, most logging statements in go currently are in the border router, and almost all of them use a specific logger for their context, so this change doesn't touch them at all.

---

Review status: 0 of 14 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1201#-:-Kr5Vp8FTW46XZbVtG5e:b-nd4fhj)*

scrye:
I agree that some of the logging statements should be pruned, but having some per packet logging that can be switched on/off using a build flag is a really powerful tool (for at least packet lengths, destination, and some other critical fields). At some point we'll add QoS code, and it will also help there to debug path selection.

I feel there is 0 downside to adding this feature. It can be switched to having debug statements on by default, so nothing changes for anyone. If someone is interested in deploying that exact same code in production they can just build it with debugging off. We can run with debug on in our VMs and debug off in production without manually changing sources, which will help us be more efficient.

The BR not being impacted by this change is a plus.

---

Review status: 0 of 14 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1201#-:-Kr5v6wlti3DYthPrCMn:b-k6ktky)*

kormat:
The biggest problem is that it's indiscriminate. Have a look at the debug logging statements in the border router:
```
$ git grep '\.Debug(' -- go/border
go/border/revinfo.go:           log.Debug("Dropping rev token")
go/border/revinfo.go:           log.Debug("Forwarding revocation", "revInfo", revInfo.Pretty(), "targets", args.Addrs)
go/border/router.go:    //rp.Debug("processPacket", "raw", rp.Raw)
go/border/rpkt/l4.go:           rp.Debug("Skipping checksum verification of L4 header", "type", rp.L4Type)
go/border/setup.go:     log.Debug("Startup capabilities", "caps", caps)
go/border/setup.go:     log.Debug("Cleared capabilities", "caps", caps)
go/border/setup.go:     log.Debug("Topology and AS config loaded", "IA", config.IA, "IfIDs", config.BR, "dir", r.confDir)
go/border/setup.go:             log.Debug("No change detected for local socket.", "bindaddr", ba)
go/border/setup.go:     log.Debug("Set up new local socket.", "conn", over.LocalAddr())
go/border/setup.go:             log.Debug("Existing interface changed.", "old", oldIntf, "new", intf)
go/border/setup.go:             log.Debug("No change detected for external socket.", "conn",
go/border/setup.go:     log.Debug("Set up new external socket.", "intf", intf)
```
There's only one there that's in the per-packet path, and it's the one that's commented out. Your change would remove all of the other debug statements, despite them being not harmful, and actually useful (unless you want `log` to be `liblog` in some files/modules, and `log15` in others, which is going to cause confusion, and doesn't allow you to mix the two in a single file).

It's also generally a good idea to use specifically scoped loggers in general, as they remove a lot of context boilerplate. The SIG should be moving towards that as it matures.

---

Review status: 0 of 14 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1201#-:-Kr6NWi4B8RRAX_ngHr7:b-l056l8)*

kormat:
My suggestion is that, if you want, put a single `Debug` log statement for both ingress and ingress, and have it commented out by default. Anyone who needs to see that debug statement can uncomment it.

---

Review status: 0 of 14 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1201#-:-Kr6OjXjTPA8GYHfMQoX:b-xk1uma)*

scrye:
Debug statements that constantly supply useful information should probably be bumped up to `Info`. This leaves `Debug` for detailed information that can be switched off at a whim.

The consistency argument is good though. Initially I `sed`ed all go files to use the new library and it broke stuff in the Go router.

I think having a `dev` working copy with extra features like fast on/off debugging for performance stuff is very useful, but it's fine if we don't merge it in master. I can set it up locally in my own environment anyway. I personally lose a lot of time commenting and uncommenting (because it's rarely only one statement), but I can script a local switch to do that for me.

As a side note, the code pattern in `log` can also be used to simplify `assert` code in the Go router (no more need for `if assert.On` blocks). Empty functions also get compiled out so performance stays the same.

---

Review status: 0 of 14 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1201#-:-KrAFgQDGa55ZFpMtMpc:b33lqhx)*

PR #1214, opened 2017-08-14
Re-factor sring to a single ringbuffer

kormat:
This simplifies its use in places where you are passing data through
multiple ring-buffers without copying. This also means it can just be
used as a free buffer pool regardless of how filled buffers are passed
around.

Also:
- Add a `Close()` method, to allow shutting down of go routines who are
  blocked reading/writing to the ring buffer.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1214)

### DIFF
--- a/go/lib/ringbuf/metrics.go
+++ b/go/lib/ringbuf/metrics.go
@@ -1,0 +1,70 @@
+// Copyright 2017 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ringbuf
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type metrics struct {
+	writeCalls   prometheus.Counter
+	readCalls    prometheus.Counter
+	writeEntries prometheus.Counter
+	readEntries  prometheus.Counter
+}
+
+func newMetrics(desc string, labels prometheus.Labels) *metrics {
+	l := copyLabels(labels)
+	l["desc"] = desc
+	return &metrics{
+		writeCalls:   WriteCalls.With(l),
+		readCalls:    ReadCalls.With(l),
+		writeEntries: WriteEntries.With(l),
+		readEntries:  ReadEntries.With(l),
+	}
+}
+
+func copyLabels(labels prometheus.Labels) prometheus.Labels {
+	l := make(prometheus.Labels)
+	for k, v := range labels {
+		l[k] = v
+	}
+	return l
+}
+
+var WriteCalls *prometheus.CounterVec
+var ReadCalls *prometheus.CounterVec
+var WriteEntries *prometheus.CounterVec
+var ReadEntries *prometheus.CounterVec
+
+func InitMetrics(namespace string) {
+	WriteCalls = newCounterVec(namespace, "write_calls_total", "Number of calls to Write.")
+	ReadCalls = newCounterVec(namespace, "read_calls_total", "Number of calls to Read.")
+	WriteEntries = newCounterVec(namespace, "write_entries_total",
+		"Number of written entries.")
+	ReadEntries = newCounterVec(namespace, "read_entries_total",
+		"Number of read entries.")
+}
+
+func newCounterVec(namespace, name, help string) *prometheus.CounterVec {
+	return prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      name,
+			Help:      help,
+		},
+		[]string{"id", "desc"},
+	)
+}
--- a/go/lib/ringbuf/ringbuf.go
+++ b/go/lib/ringbuf/ringbuf.go
@@ -1,0 +1,146 @@
+// Copyright 2017 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ringbuf
+
+import (
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type Entry interface{}
+type EntryList []Entry
+type NewEntryF func() interface{}
+
+// Ring is a classic generic ring buffer on top of a fixed-sized slice.
+type Ring struct {
+	mutex      sync.Mutex
+	writableC  *sync.Cond
+	readableC  *sync.Cond
+	entries    EntryList
+	writeIndex int
+	readIndex  int
+	writable   int
+	readable   int
+	closed     bool
+	metrics    *metrics
+}
+
+func New(count int, newf NewEntryF, desc string, labels prometheus.Labels) *Ring {
+	r := &Ring{}
+	r.readableC = sync.NewCond(&r.mutex)
+	r.entries = make(EntryList, count)
+	// Only allocate memory if caller requested it
+	if newf != nil {
+		for i := 0; i < count; i++ {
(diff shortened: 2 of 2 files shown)


scrye:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/lib/ringbuf/ringbuf.go, line 41 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWOLkHYGFxwAy9Ggus:-KrWOLkIm0doAUE31EyK:b6kcsyp) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L41)):*
> ```Go
> }
> 
> func New(count int, newf NewEntryF, desc string, labels prometheus.Labels) *Ring {
> ```

Add comment describing arguments, especially `newf` which has special semantics depending on whether it's nil or not.

---

*[go/lib/ringbuf/ringbuf.go, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWObLYRcjKo1tict50:-KrWObLYRcjKo1tict51:b-glm0c6) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L63)):*
> ```Go
> // Write copies entries to the internal ring buffer.
> // Returns the number of entries written, or -1 if the RingBuf is closed.
> func (r *Ring) Write(entries EntryList, block bool) int {
> ```

Add a short description of what `block`does (maybe at the top of the file); it's also important to make it clear that `block` is only relevant for reading/writing at least one entry, and that the ringbuffer does not repeatedly block until everything is written.

---

*[go/lib/ringbuf/ringbuf.go, line 66 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWMRnvuKDHgyDDY0ex:-KrWMRnwx4iq0XbW4WNZ:b-ytn7rr) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L66)):*
> ```Go
> 	r.mutex.Lock()
> 	defer r.mutex.Unlock()
> 	r.metrics.writeCalls.Inc()
> ```

I would move this external call outside the locked region, just to be 100% sure it doesn't do any funky stuff while locking the entire `Ring`.

---

*[go/lib/ringbuf/ringbuf.go, line 90 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWMd-gqS3flB2obY-v:-KrWMd-gqS3flB2obY-w:b2nreh6) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L90)):*
> ```Go
> 	r.mutex.Lock()
> 	defer r.mutex.Unlock()
> 	r.metrics.readCalls.Inc()
> ```

Same as above.

---

*[go/lib/ringbuf/ringbuf.go, line 116 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWNZsEk_o3L45YaL3h:-KrWNZsEk_o3L45YaL3i:b-6bilsl) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L116)):*
> ```Go
> 	r.closed = true
> 	r.writableC.Signal()
> 	r.readableC.Signal()
> ```

These two need to be `Broadcast()`, otherwise if two readers/writers are blocked in the same spot they won't both wake up.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1214)*

scrye:
Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions.

---

*[go/lib/ringbuf/ringbuf.go, line 81 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWPg2LZYn86tCm8GCB:-KrWPg2LZYn86tCm8GCC:b-4wg9ll) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L81)):*
> ```Go
> 	r.readable += n
> 	r.readableC.Signal()
> 	r.metrics.writeEntries.Add(float64(n))
> ```

Also moving this would be great, although that would mean `defer` can't be used.

---

*[go/lib/ringbuf/ringbuf.go, line 105 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWPlfZR9q_gG03mbsa:-KrWPlfZR9q_gG03mbsb:b2nreh6) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L105)):*
> ```Go
> 	r.writable += n
> 	r.writableC.Signal()
> 	r.metrics.readEntries.Add(float64(n))
> ```

Same as above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1214)*

kormat:
Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions.

---

*[go/lib/ringbuf/ringbuf.go, line 41 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWOLkHYGFxwAy9Ggus:-KrWRYmVCwHt5WBoKi_p:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L41)):*

Add comment describing arguments, especially `newf` which has special semantics depending on whether it's nil or not.
</blockquote></details>

Done.

---

*[go/lib/ringbuf/ringbuf.go, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWObLYRcjKo1tict50:-KrWS1Bi5BICFPp1x0sB:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L63)):*

Add a short description of what `block`does (maybe at the top of the file); it's also important to make it clear that `block` is only relevant for reading/writing at least one entry, and that the ringbuffer does not repeatedly block until everything is written.
</blockquote></details>

Done.

---

*[go/lib/ringbuf/ringbuf.go, line 66 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWMRnvuKDHgyDDY0ex:-KrWS98fcgx7dHQMVT9t:bv9imq1) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L66)):*

I would move this external call outside the locked region, just to be 100% sure it doesn't do any funky stuff while locking the entire `Ring`.
</blockquote></details>

While the write/readCalls could be moved outside, moving the write/readEntries outside the lock complicates things more than i'd like. Looking at the prometheus client library, Inc/Add both end up doing `atomic.LoadUint64` followed by `atomic.CompareAndSwapUint64`. So there's no locking or blocking.

---

*[go/lib/ringbuf/ringbuf.go, line 116 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWNZsEk_o3L45YaL3h:-KrWSnApm5BkJ7deraqJ:bjl0zbf) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L116)):*

These two need to be `Broadcast()`, otherwise if two readers/writers are blocked in the same spot they won't both wake up.
</blockquote></details>

Ah, good catch. I should have looked at the docs for sync.Cond. Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1214)*

scrye:
:lgtm:

---

Reviewed 2 of 2 files at r1, 1 of 1 files at r2.

---

*[go/lib/ringbuf/ringbuf.go, line 66 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-KrWMRnvuKDHgyDDY0ex:-KrWT_Z54qy9Sv04qz9O:b-jimqd9) ([raw file](https://github.com/netsec-ethz/scion/blob/17f275619da5757afa96694db3c10d28e8b209d6/go/lib/ringbuf/ringbuf.go#L66)):*

While the write/readCalls could be moved outside, moving the write/readEntries outside the lock complicates things more than i'd like. Looking at the prometheus client library, Inc/Add both end up doing `atomic.LoadUint64` followed by `atomic.CompareAndSwapUint64`. So there's no locking or blocking.
</blockquote></details>

If they're only using atomic functions then we should be safe. We can leave them as is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1214#-:-KrWTqK2bY4bLOKS_Nv0:bnfp4nl)*

PR #1221, opened 2017-08-16
Refactor prometheus metrics configuration.

kormat:
By making metrics be configured on start rather than at compile time, we
can easily include things like the element ID in all labels a process
exports. It also removes a lot of boiler-plate.

Note: this change adds a mandatory -id flag to the sig.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1221)

### DIFF
--- a/go/border/metrics/metrics.go
+++ b/go/border/metrics/metrics.go
@@ -18,7 +18,6 @@ package metrics
 
 import (
 	"flag"
-	"io"
 	"net"
 	"net/http"
 
@@ -28,6 +27,7 @@ import (
 
 	"github.com/netsec-ethz/scion/go/lib/common"
 	"github.com/netsec-ethz/scion/go/lib/overlay/conn"
+	"github.com/netsec-ethz/scion/go/lib/prom"
 	"github.com/netsec-ethz/scion/go/lib/ringbuf"
 )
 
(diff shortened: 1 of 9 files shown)


shitz:
:lgtm:

---

Reviewed 9 of 9 files at r1.

---

*[go/border/metrics/metrics.go, line 60 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1221#-Krg-ZSioxxAFWXgX3k8:-Krg-ZSioxxAFWXgX3k9:bmq9o55) ([raw file](https://github.com/netsec-ethz/scion/blob/1230d6c34cefffaeee6d4fedabe7f515c77fdd74/go/border/metrics/metrics.go#L60)):*
> ```Go
> 	newC := func(name, help string) prometheus.Counter {
> 		v := prom.NewCounter(namespace, "", name, help, constLabels)
> 		prometheus.MustRegister(v)
> ```

Nit - If you only register one metric at a time then you can call `Register` instead of `MustRegister`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1221#-:-Krg2Pu11CZy0ptHNXYD:bnfp4nl)*

kormat:

---

*[go/border/metrics/metrics.go, line 60 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1221#-Krg-ZSioxxAFWXgX3k8:-Krj68c0fJcMHL0jliUN:b-9cj2cw) ([raw file](https://github.com/netsec-ethz/scion/blob/1230d6c34cefffaeee6d4fedabe7f515c77fdd74/go/border/metrics/metrics.go#L60)):*

Nit - If you only register one metric at a time then you can call `Register` instead of `MustRegister`
</blockquote></details>

There's a bigger difference, `MustRegister` will panic if there's a problem with the registration, whereas `Register` will just return an error, which in any case we should be treating as fatal.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1221)*

klausman:
:lgtm:

---

Reviewed 9 of 9 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1221#-:-Krj9ekWm5iHE7AHKNDT:bnfp4nl)*

PR #1222, opened 2017-08-16
Added asynchronous path resolver to main lib

scrye:
Features:
  - fast path only needs to do one atomic load to grab fresh paths
  - paths are continuously updated in the background
  - transparently tolerates SCIOND restarts, connection desyncs, and other network related events

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1222)


scrye:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/pr/pr.go, line 154 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krjs9_wQrg8QCMeQO6B:-Krjs9_wQrg8QCMeQO6C:brfr17a) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L154)):*
> ```Go
> 			"src", src, "dst", dst)
> 	}
> ```

Reminder: Add the ISD-AS to the map.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1222)*

kormat:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[go/lib/pr/pr.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KrjAZZ2OJmO-R3NnqXg:-KrjAZZ2OJmO-R3NnqXh:b-v4r7g3) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L15)):*
> ```Go
> 
> // package pr implement an asynchronous Path Resolver for SCION Paths.
> ```

This package name is a bit too short, even by _my_ standards :) How about `pathmgr`?

---

*[go/lib/pr/pr.go, line 39 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KrjAub7ihnL9EGy70up:-KrjAub81YrXnsmud5Cy:b9xppnt) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L39)):*
> ```Go
> 
> 	// For now, log errors in library because there's no asynchronous
> 	// mechanism to gather them
> ```

A better way would be to have a base logger be passed in to `New`, and use a sub-logger created from that. `pr.Logger = logger.New("pathmgr")`.

---

*[go/lib/pr/pr.go, line 97 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk0ATrdqBcDPAWqmsS:-Krk0ATrdqBcDPAWqmsT:b-smlcrw) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L97)):*
> ```Go
> }
> 
> type PR struct {
> ```

This type is big enough to deserve its own file. I would move the other stuff to another file.

---

*[go/lib/pr/pr.go, line 99 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk0Wn3IVZOiruM2C1t:-Krk0Wn4Xh4N-NpO1XBF:b-d7dm81) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L99)):*
> ```Go
> type PR struct {
> 	sciondPath string
> 	// sciond is _not_ currently concurrency safe and must be protected by locks
> ```

Hum. It would be better to move this lock into sciond.Connector itself.

---

*[go/lib/pr/pr.go, line 126 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk2MiYwP5E0yUAvkg2:-Krk2MiYwP5E0yUAvkg3:brtrk64) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L126)):*
> ```Go
> 	pr := &PR{sciondPath: sciondPath,
> 		sciond:         sciondSock,
> 		regMap:         make(map[string]bool),
> ```

If this is just being used as a set, it's better to use `map[string]struct{}`, as that makes it clear that the only significant information is presence/absence.

---

*[go/lib/pr/pr.go, line 148 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk2I7BAmAxMRrNfVq_:-Krk2I7BAmAxMRrNfVqa:bvexoor) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L148)):*
> ```Go
> 	// Sanity check for double registrations
> 	r.regMapMutex.Lock()
> 	_, duplicate := r.regMap[src.String()+"."+dst.String()]
> ```

`regMap` is never populated.

---

*[go/lib/pr/pr.go, line 155 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk3NG5BztqogLUOL3u:-Krk3NG5BztqogLUOL3v:b-4r07kv) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L155)):*
> ```Go
> 	}
> 
> 	if atomic.CompareAndSwapUint64(&r.regCount, queryChanCap, queryChanCap) {
> ```

I don't think we need low-level atomic operations for this kind of thing. Having `SyncPaths` use `atomic.Value` covers exactly what i was looking for. A mutex to cover the mutable state inside `PR` is plenty (and would simplify the code a bit), especially as none of the functionality is on the fast-path.

---

*[go/lib/pr/pr.go, line 164 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk58Tmq4T7frrqPX-Y:-Krk58Tmq4T7frrqPX-Z:b-55jv3g) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L164)):*
> ```Go
> 	// Reserve memory location for pointer to path
> 	sp := &SyncPaths{}
> 	sp.Store([]*sciond.FwdPathMeta{})
> ```

I believe this is not needed. Any call to `Load` before anything is stored will return `nil`, which is a valid slice.

---

*[go/lib/pr/pr.go, line 183 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KrkBrGkTJk5SjKKz0wB:-KrkBrGkTJk5SjKKz0wC:b-ikgh3j) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L183)):*
> ```Go
> 	for query := range r.queries {
> 		// Spawn blocking sub-goroutine
> 		go r.lookup(query)
> ```

Why a goroutine for this? All queries will be immediately serialised when talking to sciond anyway, as there's only a single Connector.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1222)*

scrye:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[go/lib/pr/pr.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KrjAZZ2OJmO-R3NnqXg:-KrkQ7EjOwwGGzVuu1-T:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L15)):*

This package name is a bit too short, even by _my_ standards :) How about `pathmgr`?
</blockquote></details>

Done.

---

*[go/lib/pr/pr.go, line 39 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KrjAub7ihnL9EGy70up:-KrkRbIyRo_hYnSU_cV-:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L39)):*

A better way would be to have a base logger be passed in to `New`, and use a sub-logger created from that. `pr.Logger = logger.New("pathmgr")`.
</blockquote></details>

Done.

---

*[go/lib/pr/pr.go, line 97 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk0ATrdqBcDPAWqmsS:-KrkcRjC35Ew6Re9Qr5Y:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L97)):*

This type is big enough to deserve its own file. I would move the other stuff to another file.
</blockquote></details>

Done.

---

*[go/lib/pr/pr.go, line 99 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk0Wn3IVZOiruM2C1t:-KrkRoi7uMm10CkKo4wF:be3aaap) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L99)):*

Hum. It would be better to move this lock into sciond.Connector itself.
</blockquote></details>

Now that the SCIOND locks are internal, the lock is no longer needed.

---

*[go/lib/pr/pr.go, line 126 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk2MiYwP5E0yUAvkg2:-KrkTS2XxCl8y85_4aR-:b-le5ffm) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L126)):*

If this is just being used as a set, it's better to use `map[string]struct{}`, as that makes it clear that the only significant information is presence/absence.
</blockquote></details>

To support reconnects (which happen between the same source and destination) the map values are now `*SyncPaths`. Duplicate registrations now return a pointer to the thread-safe slice of paths.

---

*[go/lib/pr/pr.go, line 148 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk2I7BAmAxMRrNfVq_:-KrkTkisC9T43jvcVQ2X:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L148)):*

`regMap` is never populated.
</blockquote></details>

Done.

---

*[go/lib/pr/pr.go, line 155 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk3NG5BztqogLUOL3u:-KrkcmEFF1vpCZQu0e4o:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L155)):*

I don't think we need low-level atomic operations for this kind of thing. Having `SyncPaths` use `atomic.Value` covers exactly what i was looking for. A mutex to cover the mutable state inside `PR` is plenty (and would simplify the code a bit), especially as none of the functionality is on the fast-path.
</blockquote></details>

Done.

---

*[go/lib/pr/pr.go, line 164 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk58Tmq4T7frrqPX-Y:-KrkTucNYRKsXo59cH8N:bxwgszd) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L164)):*

I believe this is not needed. Any call to `Load` before anything is stored will return `nil`, which is a valid slice.
</blockquote></details>

It is not _needed_, but I like the semantics behind returning a slice with 0 elements instead of nil. Also, it's untyped nil (as the type is pinned inside `sync.Value` on first `Store`), which is not a valid slice. This means that `Load` callers need to check for untyped `nil` and return a slice `nil`, which is a bit more involved.

---

*[go/lib/pr/pr.go, line 183 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KrkBrGkTJk5SjKKz0wB:-KrkVh0XrrAhZBj0bsdq:b-hncbn9) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pr/pr.go#L183)):*

Why a goroutine for this? All queries will be immediately serialised when talking to sciond anyway, as there's only a single Connector.
</blockquote></details>

True, as long as `sciond` does nothing in parallel underneath the goroutine is pointless.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1222)*

scrye:
Review status: 0 of 3 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[go/lib/pathmgr/pathmgr_test.go, line 51 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KrlHRD9Q_noW8AGpdCM:-KrlHRDAz6bsTkOzcC4s:b-ck5uhr) ([raw file](https://github.com/netsec-ethz/scion/blob/26ff049d49eb8a95ca87d4db1d5ce6e3ed94cf9f/go/lib/pathmgr/pathmgr_test.go#L51)):*
> ```Go
> 		time.Sleep(2 * time.Second)
> 	}
> 	// Output: foo
> ```

Reminder to myself to remove this to stop it from being executed during unit tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1222)*

kormat:
Reviewed 2 of 5 files at r2, 2 of 3 files at r3.
Review status: 2 of 3 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/pathmgr/pathmgr.go, line 155 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk3NG5BztqogLUOL3u:-KsJCvqTEUxwkHXpVJ7D:bihdv4g) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pathmgr/pathmgr.go#L155)):*

Done.
</blockquote></details>

This code didn't change.

---

*[go/lib/pathmgr/pathmgr.go, line 164 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk58Tmq4T7frrqPX-Y:-KsJANEA-0c7kIq_D9Vr:b-tzg14n) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pathmgr/pathmgr.go#L164)):*

It is not _needed_, but I like the semantics behind returning a slice with 0 elements instead of nil. Also, it's untyped nil (as the type is pinned inside `sync.Value` on first `Store`), which is not a valid slice. This means that `Load` callers need to check for untyped `nil` and return a slice `nil`, which is a bit more involved.
</blockquote></details>

They don't, though. If i comment out this line, the Example code works fine. `SyncPaths.Load` already takes care of it.

---

*[go/lib/pathmgr/pathmgr.go, line 69 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJBv1UKNop243YLQN-:-KsJBv1UKNop243YLQN0:b2cbwln) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr.go#L69)):*
> ```Go
> 	}
> 
> 	pr := &PR{sciondPath: sciondPath,
> ```

It's a bit easier to read multi-line literals when the first member assigment starts on the next line. I missed this part when reading the code.

(i.e. ```
&PR{
   sciondPath: sciondPath,
   ...
```)
```


---

*[go/lib/pathmgr/pathmgr.go, line 73 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJBekyOMWn4WuN3hsf:-KsJBekyOMWn4WuN3hsg:b-manrhm) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr.go#L73)):*
> ```Go
> 		regMap:         make(map[string]*SyncPaths),
> 		queries:        make(chan query, queryChanCap),
> 		state:          sciondUp,
> ```

This probably belongs with the `sciond` vars above.

---

*[go/lib/pathmgr/pathmgr.go, line 77 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJCKYHkHlwQIhUx3ll:-KsJCKYHkHlwQIhUx3lm:b-wsu8s3) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr.go#L77)):*
> ```Go
> 		Logger:         logger.New("pathmgr")}
> 
> 	// Start resolver, which periodically spawns SCIOND query goroutines
> ```

Outdated

---

*[go/lib/pathmgr/pathmgr.go, line 206 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJFbo7UkpgCs26CJfP:-KsJFbo7UkpgCs26CJfQ:b-2b9hn7) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr.go#L206)):*
> ```Go
> // changes are made
> func (r *PR) stateTransition(old, new sciondState) bool {
> 	if r.state == old {
> ```

This logic is a bit odd. E.g., if `old` doesn't equal `r.state`, that means we have inconsistent state, but it's silently ignored. How about instead only passing in a the new state, and comparing _that_ to `r.state`?

---

*[go/lib/pathmgr/pathmgr_test.go, line 13 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJ9hiAgzroFcIfYxF8:-KsJ9hiAgzroFcIfYxF9:bf88d9n) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr_test.go#L13)):*
> ```Go
> // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
> // See the License for the specific language governing permissions and
> ```

As discussed: `infrarunning` build flag, with predictable Output.

---

*[go/lib/pathmgr/pathmgr_test.go, line 29 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJ7EoIHQ5JJO_6vqcq:-KsJ7EoIHQ5JJO_6vqcr:b-m3g5g3) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr_test.go#L29)):*
> ```Go
> func ExamplePR() {
> 	src, _ := addr.IAFromString("1-14")
> 	dst, _ := addr.IAFromString("2-21")
> ```

It would be nice if these were configurable. E.g.:
```
var srcStr = flag.String("srcIA", "1-14", "Source ISD-AS")
var dstStr = flag.String("dstIA", "2-21", "Dest ISD-AS")
...
src, err := addr.IAFromString(*srcStr)
...
```
It can then be run like this: `go test -v ./go/lib/pathmgr/ -args -srcIA 1-11 -dstIA 1-13`. That way it can still be used when not running the default topo.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1222)*

scrye:
Review status: 0 of 3 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/pathmgr/pathmgr.go, line 155 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk3NG5BztqogLUOL3u:-KsJIRmbbkqImOHP_J2y:b-x4rjhb) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pathmgr/pathmgr.go#L155)):*

This code didn't change.
</blockquote></details>

Whoops. Done.

---

*[go/lib/pathmgr/pathmgr.go, line 164 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk58Tmq4T7frrqPX-Y:-KsJQFQ8s9h88XJboG0Q:by49ezq) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pathmgr/pathmgr.go#L164)):*

They don't, though. If i comment out this line, the Example code works fine. `SyncPaths.Load` already takes care of it.
</blockquote></details>

This is because the first call to `lookup` succeeds and populates the slice. If the first lookup fails (e.g., timeout), `Load` attempts to cast untyped nil to an empty slice, which panics.

---

*[go/lib/pathmgr/pathmgr.go, line 69 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJBv1UKNop243YLQN-:-KsJI2mmrg25JxxF8FXG:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr.go#L69)):*

It's a bit easier to read multi-line literals when the first member assigment starts on the next line. I missed this part when reading the code.

(i.e. ```
&PR{
   sciondPath: sciondPath,
   ...
```)
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 73 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJBekyOMWn4WuN3hsf:-KsJIAJMxnLz602VBu9w:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr.go#L73)):*

This probably belongs with the `sciond` vars above.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 77 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJCKYHkHlwQIhUx3ll:-KsJIKle7J-n2d1M9YO2:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr.go#L77)):*

Outdated
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 206 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJFbo7UkpgCs26CJfP:-KsJQ_EPHtbd6qBHQmCS:bfutorf) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr.go#L206)):*

This logic is a bit odd. E.g., if `old` doesn't equal `r.state`, that means we have inconsistent state, but it's silently ignored. How about instead only passing in a the new state, and comparing _that_ to `r.state`?
</blockquote></details>

Since there are only two states (and we are no longer using multiple goroutines competing for atomic state transitions), the two approaches are equivalent. I went with the proposed variant since it's a bit shorter.

---

*[go/lib/pathmgr/pathmgr_test.go, line 13 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJ9hiAgzroFcIfYxF8:-KsJQAKMb8ZRkyK80ZAH:bsz7qak) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr_test.go#L13)):*

As discussed: `infrarunning` build flag, with predictable Output.
</blockquote></details>

Done.

However, this has the unexpected downside of completely removing the example from generated documentation.

---

*[go/lib/pathmgr/pathmgr_test.go, line 29 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJ7EoIHQ5JJO_6vqcq:-KsJTIdorWMC8yxqdxs_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr_test.go#L29)):*

It would be nice if these were configurable. E.g.:
```
var srcStr = flag.String("srcIA", "1-14", "Source ISD-AS")
var dstStr = flag.String("dstIA", "2-21", "Dest ISD-AS")
...
src, err := addr.IAFromString(*srcStr)
...
```
It can then be run like this: `go test -v ./go/lib/pathmgr/ -args -srcIA 1-11 -dstIA 1-13`. That way it can still be used when not running the default topo.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1222)*
<!-- Sent from Reviewable.io -->

kormat:
Reviewed 2 of 5 files at r2, 1 of 3 files at r3, 2 of 2 files at r4.

---

*[go/lib/pathmgr/pathmgr.go, line 154 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krjs9_wQrg8QCMeQO6B:-KsJjs2u7mTJpB3Gi5Tg:b2vnq24) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pathmgr/pathmgr.go#L154)):*

Reminder: Add the ISD-AS to the map.
</blockquote></details>

Is this still valid?

---

*[go/lib/pathmgr/pathmgr.go, line 155 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk3NG5BztqogLUOL3u:-KsJixF8L_4zXPhzBQ1f:b-uej1m4) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pathmgr/pathmgr.go#L155)):*

Whoops. Done.
</blockquote></details>

There's still an `atomic.AddUint64` below ;P

---

*[go/lib/pathmgr/pathmgr.go, line 164 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk58Tmq4T7frrqPX-Y:-KsJjRR8XyXjSQEfyvI1:bbtx0ep) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pathmgr/pathmgr.go#L164)):*

This is because the first call to `lookup` succeeds and populates the slice. If the first lookup fails (e.g., timeout), `Load` attempts to cast untyped nil to an empty slice, which panics.
</blockquote></details>

Ahh, ok. So it _is_ needed. :)

---

*[go/lib/pathmgr/pathmgr_test.go, line 13 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJ9hiAgzroFcIfYxF8:-KsJj_uqFG28CpYucuwZ:b-cs4m6x) ([raw file](https://github.com/netsec-ethz/scion/blob/bc77df9247931e3a84a70810f7c349bd88c7f5d6/go/lib/pathmgr/pathmgr_test.go#L13)):*

Done.

However, this has the unexpected downside of completely removing the example from generated documentation.
</blockquote></details>

Damnit :/

---

*[go/lib/pathmgr/pathmgr_test.go, line 50 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJjlKPjZEdqr-ofrDc:-KsJjlKPjZEdqr-ofrDd:b-yoi6dd) ([raw file](https://github.com/netsec-ethz/scion/blob/6dc51f4a7831198bbd4b8c613de25c3afcb6cc97/go/lib/pathmgr/pathmgr_test.go#L50)):*
> ```Go
> 
> 	// Initialize path resolver
> 	pr, err := New("/run/shm/sciond/sd1-14.sock", time.Second, log.Root())
> ```

Ah, this path is hard-coded. Using the srcIA in there would help :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1222)*

kormat:
:lgtm:

---

Reviewed 2 of 5 files at r2, 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-:-KsJxCz2VaYOUUhVm4DH:bnfp4nl)*

scrye:

---

*[go/lib/pathmgr/pathmgr.go, line 154 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krjs9_wQrg8QCMeQO6B:-KsJv5BGkDEEJx0tYj_T:byutpig) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pathmgr/pathmgr.go#L154)):*

Is this still valid?
</blockquote></details>

No, it was fixed a while ago but I forgot to remove it.

---

*[go/lib/pathmgr/pathmgr.go, line 155 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-Krk3NG5BztqogLUOL3u:-KsJvApyDOZyLnfS-xLU:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/649f78560c1984d0fae3162047e2acbbdd8d2733/go/lib/pathmgr/pathmgr.go#L155)):*

There's still an `atomic.AddUint64` below ;P
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 50 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1222#-KsJjlKPjZEdqr-ofrDc:-KsJvVX-mkZj6rgtZ5A3:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/6dc51f4a7831198bbd4b8c613de25c3afcb6cc97/go/lib/pathmgr/pathmgr_test.go#L50)):*

Ah, this path is hard-coded. Using the srcIA in there would help :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1222)*

PR #1224, opened 2017-08-18
Fix a memory leak in br/ringbuf.

kormat:
When a packet was created by the BR (e.g. via `genPkt()`), references to
this packet would be retained in the egress ringbuf's until overwritten,
preventing it from being GC'd. Over time, this would lead to 1024 64KiB
packet references been kept, causing the RSS to balloon up to a maximum
size of ~200MiB.

This change:
- Clears references from a ringbuf on `Read()`
- Clears references from read/write slices passed to a ringbuf.

Also:
- Reduce the packet buffer size to handle jumbo frames (a 7x reduction),
  as there is currently no reason to support packet sizes beyond that.
- Make RtrPkt.Reset() slightly more comprehensive.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1224)

### DIFF
--- a/go/border/io.go
+++ b/go/border/io.go
@@ -86,6 +86,8 @@ Top:
 			bytesRecv.Add(float64(length))
 			pktRecvSizes.Observe(float64(length))
 			s.Ring.Write(ringbuf.EntryList{pkts[i]}, true)
+			// Clear RtrPkt reference
+			pkts[i] = nil
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


shitz:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1224#-:-KroWyejXn07TLbi_qFD:bnfp4nl)*

PR #1228, opened 2017-08-23
Add SCION Control messages to Go lib

shitz:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1228)


kormat:
Reviewed 23 of 23 files at r1.

---

*[go/lib/ctrl/ctrl.go, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1228#-KsNMZ86zE4dWsBuHm_d:-KsNMZ86zE4dWsBuHm_e:b39k064) ([raw file](https://github.com/netsec-ethz/scion/blob/cf5f677ac6d9dbf6684e8213451090604fdaabf7/go/lib/ctrl/ctrl.go#L34)):*
> ```Go
> var _ common.Payload = (*CtrlPld)(nil)
> 
> type CtrlPld struct {
> ```

This can just be `Pld`, it's already scoped under `ctrl`.

---

*[go/lib/ctrl/ctrl.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1228#-KsNNOlZQ7dQ1Edxkzkb:-KsNNOlZQ7dQ1Edxkzkc:bgr7ttd) ([raw file](https://github.com/netsec-ethz/scion/blob/cf5f677ac6d9dbf6684e8213451090604fdaabf7/go/lib/ctrl/ctrl.go#L35)):*
> ```Go
> 
> type CtrlPld struct {
> 	Which       proto.SCION_Which
> ```

Wait, why is this duplicating the structure of the SCION proto?

---

*[go/lib/util/time.go, line 25 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1228#-KsNKmA0BYN_QghR7EtQ:-KsNKmA0BYN_QghR7EtR:bxxlycb) ([raw file](https://github.com/netsec-ethz/scion/blob/cf5f677ac6d9dbf6684e8213451090604fdaabf7/go/lib/util/time.go#L25)):*
> ```Go
> // ISOTimestamp returns a ISO 8601 formatted UTC timestamp.
> func ISOTimestamp(t time.Time) string {
> 	return t.UTC().Format("2017-01-01 15:04:05+00:00")
> ```

This is an invalid timestamp. The go time package requires that the format be given for the _reference_ time. Also, this doesn't match the timestamp format we're using elsewhere. Have a look at `go/lib/log/log.go`, where `fmt15.TimeFmt` is initialised. I suggest you move that string from there into `go/lib/common/defs.go`, and reference it from `go/lib/log`, at which point i'm also not sure that `ISOTimestamp` is needed separately.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1228)*

shitz:
Refactored the code based on feedback. PTAL

---

Review status: 10 of 27 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1228#-:-KsPM5O70j5jXdXmHS49:b-53pieo)*

kormat:
Replaced by #1228

PR #1233, opened 2017-08-29
Add demo/connetivity tool for SCION in Go

scrye:
This PR adds a simple tool, written in Go, for echo request/reply connectivity tests.

This tool:
- showcases a simple use of the new SCION connection library;
- serves as a starting point for SCION end to end integration tests;
- can be used to test connectivity between any pair of ASes.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1233)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ bin/dispatcher
 bin/logdog
 bin/sig
 bin/topopruner
+bin/echo
 c/dispatcher/dispatcher
 c/lib/scion/checksum_bench
 c/ssp/test/*client
--- a/go/lib/snet/addr.go
+++ b/go/lib/snet/addr.go
@@ -77,9 +77,6 @@ func AddrFromString(s string) (*Addr, error) {
 	if err != nil {
 		return nil, common.NewError("Invalid port string", "port", parts["port"], "err", err)
 	}
-	if port == 0 {
-		return nil, common.NewError("Invalid port number", "port", parts["port"])
-	}
 	return &Addr{IA: ia, Host: addr.HostFromIP(ip), L4Port: uint16(port)}, nil
 }
 
--- a/go/lib/snet/addr_test.go
+++ b/go/lib/snet/addr_test.go
@@ -64,7 +64,6 @@ func Test_AddrFromString(t *testing.T) {
 		{address: "1-20,[abc]:12", isError: true},
 		{address: "1-2]:14,[1.2.3.4]", isError: true},
 		{address: "1-1,[1.2.3.4]:70000", isError: true},
-		{address: "1-1,[1.2.3.4]:0", isError: true},
 		{address: "", isError: true},
 		{address: "1-1,[1.2.3.4]:80",
 			ia:   "1-1",
--- a/go/tools/echo/Readme.md
+++ b/go/tools/echo/Readme.md
@@ -1,0 +1,15 @@
+To run an echo connectivity test on the default topology from AS1-19 to AS2-25,
+first make sure the infrastructure is running.
+
+Then, start the server using:
+```
+go run echo.go -mode server -local 2-25,[127.0.0.1]:40002 -sciond /run/shm/sciond/sd2-25.sock \
+	-dispatcher /run/shm/dispatcher/default.sock -count 10
+```
+
+Finally, start the client using:
+```
+go run echo.go -mode client -remote 2-25,[127.0.0.1]:40002 -local 1-19,[127.0.0.1]:40001 \
+	-sciond /run/shm/sciond/sd1-19.sock -dispatcher /run/shm/dispatcher/default.sock \
+	-count 10
+```
--- a/go/tools/echo/Readme.md
+++ b/go/tools/echo/Readme.md
@@ -9,7 +9,7 @@ go run echo.go -mode server -local 2-25,[127.0.0.1]:40002 -sciond /run/shm/scion
 
 Finally, start the client using:
 ```
-go run echo.go -mode client -remote 2-25,[127.0.0.1]:40002 -local 1-19,[127.0.0.1]:40001 \
+go run echo.go -mode client -remote 2-25,[127.0.0.1]:40002 -local 1-19,[127.0.0.1]:0 \
 	-sciond /run/shm/sciond/sd1-19.sock -dispatcher /run/shm/dispatcher/default.sock \
 	-count 10
 ```
--- a/go/tools/echo/Readme.md
+++ b/go/tools/echo/Readme.md
@@ -3,13 +3,10 @@ first make sure the infrastructure is running.
 
 Then, start the server using:
 ```
-go run echo.go -mode server -local 2-25,[127.0.0.1]:40002 -sciond /run/shm/sciond/sd2-25.sock \
-	-dispatcher /run/shm/dispatcher/default.sock -count 10
+go run echo.go -mode server -local 2-25,[127.0.0.1]:40002 -count 10
 ```
 
 Finally, start the client using:
 ```
-go run echo.go -mode client -remote 2-25,[127.0.0.1]:40002 -local 1-19,[127.0.0.1]:0 \
-	-sciond /run/shm/sciond/sd1-19.sock -dispatcher /run/shm/dispatcher/default.sock \
-	-count 10
+go run echo.go -mode client -remote 2-25,[127.0.0.1]:40002 -local 1-19,[127.0.0.1]:0 -count 10
 ```
--- a/go/tools/echo/echo.go
+++ b/go/tools/echo/echo.go
@@ -61,7 +61,7 @@ func init() {
 func main() {
 	validateFlags()
 	liblog.Setup(*id)
-	defer liblog.PanicLog()
+	defer liblog.LogPanicAndExit()
 	switch *mode {
 	case "client":
 		Client()
--- a/go/tools/echo/echo.go
+++ b/go/tools/echo/echo.go
@@ -37,7 +37,7 @@ const (
 )
 
 func GetDefaultSCIONDPath(ia *addr.ISD_AS) string {
-	return fmt.Sprintf("/run/shm/sd%v.sock", ia)
+	return fmt.Sprintf("/run/shm/sciond/sd%v.sock", ia)
 }
 
 var (
@@ -112,11 +112,13 @@ func Client() {
 		before := time.Now()
 		written, err := conn.Write([]byte(ReqMsg))
 		if err != nil {
-			LogFatal("Unable to write", "err", err)
+			log.Error("Unable to write", "err", err)
+			continue
 		}
 		if written != len(ReqMsg) {
-			LogFatal("Wrote incomplete message", "expected", len(ReqMsg),
+			log.Error("Wrote incomplete message", "expected", len(ReqMsg),
 				"actual", written)
+			continue
 		}
 
 		// Receive echo reply with timeout
@@ -125,6 +127,7 @@ func Client() {
 		conn.SetDeadline(time.Time{})
 		if err != nil {
 			log.Error("Unable to read", "err", err)
+			continue
 		}
 		if string(b[:read]) != ReplyMsg {
 			fmt.Println("Received bad message", "expected", ReplyMsg,
@@ -133,7 +136,7 @@ func Client() {
 		after := time.Now()
 		elapsed := after.Sub(before)
 		fmt.Printf("%d bytes from %v: seq=%d time=%s\n", read, &remote, i, elapsed)
-		time.Sleep((*interval) - elapsed)
+		time.Sleep(*interval)
 	}
 }
 
--- a/go/tools/echo/echo.go
+++ b/go/tools/echo/echo.go
@@ -29,12 +29,11 @@ import (
 )
 
 const (
-	DefaultDispatcherPath = "/run/shm/dispatcher/default.sock"
-	DefaultInterval       = 2 * time.Second
-	DefaultTimeout        = 2 * time.Second
-	MaxEchoes             = 1 << 16
-	ReqMsg                = "ping!"
-	ReplyMsg              = "pong!"
+	DefaultInterval = 2 * time.Second
+	DefaultTimeout  = 2 * time.Second
+	MaxEchoes       = 1 << 16
+	ReqMsg          = "ping!"
+	ReplyMsg        = "pong!"
 )
 
 func GetDefaultSCIONDPath(ia *addr.ISD_AS) string {
@@ -47,10 +46,10 @@ var (
 	id         = flag.String("id", "echo", "Element ID")
 	mode       = flag.String("mode", "client", "Run in client or server mode")
 	sciond     = flag.String("sciond", "", "Path to sciond socket")
-	dispatcher = flag.String("dispatcher", "", "Path to dispatcher socket")
-	count      = flag.Int("count", 0,
-		fmt.Sprintf("Number of echoes, between 0 and %d; "+
-			"a count of 0 means infinity", MaxEchoes))
+	dispatcher = flag.String("dispatcher", "/run/shm/dispatcher/default.sock",
+		"Path to dispatcher socket")
+	count = flag.Int("count", 0,
+		fmt.Sprintf("Number of echoes, between 0 and %d; a count of 0 means infinity", MaxEchoes))
 	interval = flag.Duration("interval", DefaultInterval, "time between echoes")
 )
 
--- a/go/tools/echo/echo.go
+++ b/go/tools/echo/echo.go
@@ -18,29 +18,36 @@ package main
 import (
 	"flag"
 	"fmt"
-	"io/ioutil"
-	"log"
+	"os"
 	"time"
 
-	log15 "github.com/inconshreveable/log15"
+	log "github.com/inconshreveable/log15"
 
+	"github.com/netsec-ethz/scion/go/lib/addr"
+	liblog "github.com/netsec-ethz/scion/go/lib/log"
 	"github.com/netsec-ethz/scion/go/lib/snet"
 )
 
 const (
-	DefaultInterval = 2 * time.Second
-	DefaultTimeout  = 2 * time.Second
-	MaxEchoes       = 1 << 16
-	ReqMsg          = "ping!"
-	ReplyMsg        = "pong!"
+	DefaultDispatcherPath = "/run/shm/dispatcher/default.sock"
+	DefaultInterval       = 2 * time.Second
+	DefaultTimeout        = 2 * time.Second
+	MaxEchoes             = 1 << 16
+	ReqMsg                = "ping!"
+	ReplyMsg              = "pong!"
 )
+
+func GetDefaultSCIONDPath(ia *addr.ISD_AS) string {
+	return fmt.Sprintf("/run/shm/sd%v.sock", ia)
+}
 
 var (
 	local      snet.Addr
 	remote     snet.Addr
+	id         = flag.String("id", "echo", "Element ID")
 	mode       = flag.String("mode", "client", "Run in client or server mode")
-	sciond     = flag.String("sciond", "", "(Mandatory) path to sciond socket")
-	dispatcher = flag.String("dispatcher", "", "(Mandatory) path to dispatcher socket")
+	sciond     = flag.String("sciond", "", "Path to sciond socket")
+	dispatcher = flag.String("dispatcher", "", "Path to dispatcher socket")
 	count      = flag.Int("count", 0,
 		fmt.Sprintf("Number of echoes, between 0 and %d; "+
 			"a count of 0 means infinity", MaxEchoes))
@@ -54,10 +61,8 @@ func init() {
 
 func main() {
 	validateFlags()
-
-	// Disable logging
-	log15.Root().SetHandler(log15.StreamHandler(ioutil.Discard, log15.LogfmtFormat()))
-
+	liblog.Setup(*id)
+	defer liblog.PanicLog()
 	switch *mode {
 	case "client":
 		Client()
@@ -68,29 +73,20 @@ func main() {
 
 func validateFlags() {
 	flag.Parse()
-
 	if *mode != "client" && *mode != "server" {
-		log.Fatal("Unknown mode, must be either 'client' or 'server'")
+		LogFatal("Unknown mode, must be either 'client' or 'server'")
 	}
-
 	if *mode == "client" && remote.Host == nil {
-		log.Fatal("Error: missing remote address")
+		LogFatal("Missing remote address")
 	}
-
-	if *mode == "server" && local.Host == nil {
-		log.Fatal("Error: missing local address")
+	if local.Host == nil {
+		LogFatal("Missing local address")
 	}
-
 	if *sciond == "" {
-		log.Fatal("Error: missing path to sciond")
+		*sciond = GetDefaultSCIONDPath(local.IA)
 	}
-
-	if *dispatcher == "" {
-		log.Fatal("Error: missing path to dispatcher")
-	}
-
 	if *count < 0 || *count > MaxEchoes {
-		log.Fatal(fmt.Sprintf("Error: invalid count, must be between 0 and %d", MaxEchoes))
+		LogFatal("Invalid count", "min", 0, "max", MaxEchoes, "actual", *count)
 	}
 }
 
@@ -99,33 +95,29 @@ func validateFlags() {
 // message with the round trip time is printed. On errors (including timeouts),
 // the Client exits.
 func Client() {
-	// Initialize default SCION networking context
-	err := snet.Init(local.IA, *sciond, *dispatcher)
-	if err != nil {
-		log.Fatal("Error: unable to initialize SCION network")
-	}
-	fmt.Println("SCION network successfully initialized.")
+	initNetwork()
 
 	// Connect to remote address. Note that currently the SCION library
-	// does not support automatic binding to local addresses, so the
-	// local IP address and port need to be supplied explicitly
+	// does not support automatic binding to local addresses, so the local
+	// IP address needs to be supplied explicitly. When supplied a local
+	// port of 0, DialSCION will assign a random free local port.
 	conn, err := snet.DialSCION("udp4", &local, &remote)
 	if err != nil {
-		log.Fatal("Error: unable to dial", err)
+		LogFatal("Unable to dial", "err", err)
 	}
-	fmt.Printf("Connected to %v.\n", &remote)
+	log.Debug("Connected", "local", &local, "remote", &remote)
 
 	b := make([]byte, 1<<12)
 	for i := 0; i < *count || *count == 0; i++ {
-		before := time.Now()
-
 		// Send echo request to destination
+		before := time.Now()
 		written, err := conn.Write([]byte(ReqMsg))
 		if err != nil {
-			log.Fatal("Error: unable to write", err)
+			LogFatal("Unable to write", "err", err)
 		}
 		if written != len(ReqMsg) {
-			log.Fatal("Error: wrote incomplete message")
+			LogFatal("Wrote incomplete message", "expected", len(ReqMsg),
+				"actual", written)
 		}
 
 		// Receive echo reply with timeout
@@ -133,17 +125,15 @@ func Client() {
 		read, err := conn.Read(b)
 		conn.SetDeadline(time.Time{})
 		if err != nil {
-			log.Fatal("Error: unable to read", err)
+			log.Error("Unable to read", "err", err)
 		}
 		if string(b[:read]) != ReplyMsg {
 			fmt.Println("Received bad message", "expected", ReplyMsg,
 				"actual", string(b[:read]))
 		}
 		after := time.Now()
-
 		elapsed := after.Sub(before)
 		fmt.Printf("%d bytes from %v: seq=%d time=%s\n", read, &remote, i, elapsed)
-
 		time.Sleep((*interval) - elapsed)
 	}
 }
@@ -151,26 +141,21 @@ func Client() {
 // Server listens on a SCION address and replies to any echo request messages.
 // On any error, the server exits.
 func Server() {
-	// Initialize default SCION networking context
-	err := snet.Init(local.IA, *sciond, *dispatcher)
-	if err != nil {
-		log.Fatal("Error: unable to initialize SCION network")
-	}
-	fmt.Println("SCION network successfully initialized.")
+	initNetwork()
 
 	// Listen on SCION address
 	conn, err := snet.ListenSCION("udp4", &local)
 	if err != nil {
-		log.Fatal("Error: unable to listen", err)
+		LogFatal("Unable to listen", "err", err)
 	}
-	fmt.Printf("Listening to %v.\n", &local)
+	log.Debug("Listening", "local", &local)
 
 	b := make([]byte, 1<<12)
 	for i := 0; i < *count || *count == 0; i++ {
 		// Receive echo request
 		read, senderAddr, err := conn.ReadFrom(b)
 		if err != nil {
-			log.Fatal("Error: unable to read", err)
+			LogFatal("Unable to read", "err", err)
 		}
 		if string(b[:read]) != ReqMsg {
 			fmt.Println("Received bad message", "expected", ReqMsg,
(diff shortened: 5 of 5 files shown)


kormat:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[go/tools/echo/echo.go, line 1 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmSDH-3phvAEeKFtbE:-KsmSDH-3phvAEeKFtbF:bhfv5dx) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L1)):*
> ```Go
> ```

Add `bin/echo` to the top-level `.gitignore`.

---

*[go/tools/echo/echo.go, line 43 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmSMWYSyftsaJ5o-PE:-KsmSMWYSyftsaJ5o-PF:bmk4zt4) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L43)):*
> ```Go
> 	mode       = flag.String("mode", "client", "Run in client or server mode")
> 	sciond     = flag.String("sciond", "", "(Mandatory) path to sciond socket")
> 	dispatcher = flag.String("dispatcher", "", "(Mandatory) path to dispatcher socket")
> ```

These two do not need to be mandatory. The first you can calculate from the local IA supplied, the second has a static default. Keep the flags, of course, just don't make them mandatory.

---

*[go/tools/echo/echo.go, line 59 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmSbHfiMqvrmx2OjOO:-KsmSbHfiMqvrmx2OjOP:b7t0ohf) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L59)):*
> ```Go
> 
> 	// Disable logging
> 	log15.Root().SetHandler(log15.StreamHandler(ioutil.Discard, log15.LogfmtFormat()))
> ```

Why this? If any of the scion libraries are logging, we should know.

Additionally, not using our standard logging library makes this program distinctly less useful for us. The log output is in a different format, the date/timestamps aren't usefully comparable, and it doesn't log to a file, either, which makes it not a good fit for CI testing.

---

*[go/tools/echo/echo.go, line 80 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmT0qe6WFHChFG76kF:-KsmT0qe6WFHChFG76kG:b-txz1l9) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L80)):*
> ```Go
> 	}
> 
> 	if *mode == "server" && local.Host == nil {
> ```

As the local address is mandatory, this shouldn't be dependant on the mode.

---

*[go/tools/echo/echo.go, line 93 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmZeHtwaK-tmVXiUcw:-KsmZeHtwaK-tmVXiUcx:bienl0i) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L93)):*
> ```Go
> 
> 	if *count < 0 || *count > MaxEchoes {
> 		log.Fatal(fmt.Sprintf("Error: invalid count, must be between 0 and %d", MaxEchoes))
> ```

If using the stdlib logger, there's `log.Fatalf` for exactly this.

---

*[go/tools/echo/echo.go, line 103 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmY0-UYDK8sHPJ90eK:-KsmY0-UYDK8sHPJ90eL:b-wocjlx) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L103)):*
> ```Go
> func Client() {
> 	// Initialize default SCION networking context
> 	err := snet.Init(local.IA, *sciond, *dispatcher)
> ```

This code is ~duplicated between client and server. I suggest splitting it out to a common function.

---

*[go/tools/echo/echo.go, line 105 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmXr3MiblwTLfG7-PM:-KsmXr3MiblwTLfG7-PN:b-24e03) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L105)):*
> ```Go
> 	err := snet.Init(local.IA, *sciond, *dispatcher)
> 	if err != nil {
> 		log.Fatal("Error: unable to initialize SCION network")
> ```

Include the error, otherwise this is impossible to debug. E.g. i was getting `2017/08/30 11:41:45 Error: unable to initialize SCION network`, when the reason was that it was failing to connect to sciond.

---

*[go/tools/echo/echo.go, line 107 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmY6FJtKngGUQ2MCtz:-KsmY6FJtKngGUQ2MCu-:b-j07buf) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L107)):*
> ```Go
> 		log.Fatal("Error: unable to initialize SCION network")
> 	}
> 	fmt.Println("SCION network successfully initialized.")
> ```

Mixing logging and fmt.Print* is bad; the output is inconsistently formatted, and fmt calls bypass whatever logging configuration might be in place (e.g. logging also adds to log file, but stuff written to stdout won't appear there).

---

*[go/tools/echo/echo.go, line 111 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmYQI3I7rtK12jL1Y3:-KsmYQI3I7rtK12jL1Y4:b-jpoc3u) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L111)):*
> ```Go
> 	// Connect to remote address. Note that currently the SCION library
> 	// does not support automatic binding to local addresses, so the
> 	// local IP address and port need to be supplied explicitly
> ```

The _port_ isn't required, i.e. it can be set to 0. 

---

*[go/tools/echo/echo.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmYophwfirlbzUJTYS:-KsmYophwfirlbzUJTYT:bwr0u6g) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L123)):*
> ```Go
> 
> 		// Send echo request to destination
> 		written, err := conn.Write([]byte(ReqMsg))
> ```

It would be good to use packets that are the full MTU size (or have it configurable, like with `pktgen.py`). Ah. `snet` doesn't expose this, right? That's... a problem.

---

*[go/tools/echo/echo.go, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmZD4sOxV67xgRXukm:-KsmZD4sOxV67xgRXukn:b-rtbb1o) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L128)):*
> ```Go
> 		}
> 		if written != len(ReqMsg) {
> 			log.Fatal("Error: wrote incomplete message")
> ```

Include how much was written, and how much should have been written.

---

*[go/tools/echo/Readme.md, line 12 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-Ksm_2r5GywEdiMCfq1f:-Ksm_2r5GywEdiMCfq1g:bkl3ik0) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/Readme.md#L12)):*
> ```Markdown
> Finally, start the client using:
> ```
> go run echo.go -mode client -remote 2-25,[127.0.0.1]:40002 -local 1-19,[127.0.0.1]:40001 \
> ```

I'd use port `0` for the client, to demonstrate that's supported.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1233)*

scrye:
Review status: 0 of 3 files reviewed at latest revision, 12 unresolved discussions.

---

*[go/tools/echo/echo.go, line 1 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmSDH-3phvAEeKFtbE:-KsrGRcohok5enpQ5ZLv:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L1)):*

Add `bin/echo` to the top-level `.gitignore`.
</blockquote></details>

Done.

---

*[go/tools/echo/echo.go, line 43 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmSMWYSyftsaJ5o-PE:-KsrHK3RWHa_6qWhZIoj:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L43)):*

These two do not need to be mandatory. The first you can calculate from the local IA supplied, the second has a static default. Keep the flags, of course, just don't make them mandatory.
</blockquote></details>

Done.

---

*[go/tools/echo/echo.go, line 59 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmSbHfiMqvrmx2OjOO:-KsrLHaLLAw3DNou_cit:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L59)):*

Why this? If any of the scion libraries are logging, we should know.

Additionally, not using our standard logging library makes this program distinctly less useful for us. The log output is in a different format, the date/timestamps aren't usefully comparable, and it doesn't log to a file, either, which makes it not a good fit for CI testing.
</blockquote></details>

Done.

---

*[go/tools/echo/echo.go, line 80 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmT0qe6WFHChFG76kF:-KsrM-fnaqjqld6sUE-o:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L80)):*

As the local address is mandatory, this shouldn't be dependant on the mode.
</blockquote></details>

Done.

---

*[go/tools/echo/echo.go, line 93 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmZeHtwaK-tmVXiUcw:-KsrM1D34U1jEzbNYy9l:b-owl8zw) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L93)):*

If using the stdlib logger, there's `log.Fatalf` for exactly this.
</blockquote></details>

Changed to `log15` now.

---

*[go/tools/echo/echo.go, line 103 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmY0-UYDK8sHPJ90eK:-KsrMVe9nLzsA4_mLGry:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L103)):*

This code is ~duplicated between client and server. I suggest splitting it out to a common function.
</blockquote></details>

Done.

---

*[go/tools/echo/echo.go, line 105 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmXr3MiblwTLfG7-PM:-KsrMjUJOtnDt-VGRY0A:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L105)):*

Include the error, otherwise this is impossible to debug. E.g. i was getting `2017/08/30 11:41:45 Error: unable to initialize SCION network`, when the reason was that it was failing to connect to sciond.
</blockquote></details>

Done.

---

*[go/tools/echo/echo.go, line 107 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmY6FJtKngGUQ2MCtz:-KsrMl_r34pQwwHKpEso:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L107)):*

Mixing logging and fmt.Print* is bad; the output is inconsistently formatted, and fmt calls bypass whatever logging configuration might be in place (e.g. logging also adds to log file, but stuff written to stdout won't appear there).
</blockquote></details>

Done.

---

*[go/tools/echo/echo.go, line 111 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmYQI3I7rtK12jL1Y3:-KsrNYrqoWJe8WMHN0G3:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L111)):*

The _port_ isn't required, i.e. it can be set to 0. 
</blockquote></details>

Done.

---

*[go/tools/echo/echo.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmYophwfirlbzUJTYS:-KsrNccsx_L74q1E27mq:b-srmhys) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L123)):*

It would be good to use packets that are the full MTU size (or have it configurable, like with `pktgen.py`). Ah. `snet` doesn't expose this, right? That's... a problem.
</blockquote></details>

The MTU is only relevant if you know the path your traffic will take, but for apps it is completely decided by the networking layer.

When we implement path policy support we can extend the networking API to expose the MTU, but that is only relevant if you commit to a certain path.

---

*[go/tools/echo/echo.go, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmZD4sOxV67xgRXukm:-KsrPDSKMESl0puEjHPz:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L128)):*

Include how much was written, and how much should have been written.
</blockquote></details>

Done.

---

*[go/tools/echo/Readme.md, line 12 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-Ksm_2r5GywEdiMCfq1f:-KsrPPNNBRYmLtHEmSPX:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/Readme.md#L12)):*

I'd use port `0` for the client, to demonstrate that's supported.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1233)*

kormat:
Reviewed 3 of 3 files at r2.

---

*[go/tools/echo/echo.go, line 107 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmY6FJtKngGUQ2MCtz:-KssMWmQrO_h1LwYzq_K:bldg95) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L107)):*

Done.
</blockquote></details>

There are still 3 fmt.Prinln/fmt.Printf calls left.

---

*[go/tools/echo/echo.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmYophwfirlbzUJTYS:-KssMhQGw1g38wqT9_Oo:b-h452t3) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L123)):*

The MTU is only relevant if you know the path your traffic will take, but for apps it is completely decided by the networking layer.

When we implement path policy support we can extend the networking API to expose the MTU, but that is only relevant if you commit to a certain path.
</blockquote></details>

That's not strictly true. E.g. if there were multiple paths with the same MTU, the client wouldn't have to care which of them were being used. Or, if there are multiple paths, and the client is using the minimum MTU, the same applies. In any case, for a datagram protocol it's vital that we expose the MTU to the client, otherwise they have absolutely no clue how large the packets they can send are, nor will they know about it changing (in either direction). This is not something the posix socket API envisioned, so we'll have to figure out a good way to extend it.

---

*[go/tools/echo/echo.go, line 32 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssLUUqwE7jnP_uWqnI:-KssLUUqwE7jnP_uWqnJ:b8gnbf6) ([raw file](https://github.com/netsec-ethz/scion/blob/56b27da192e72ed779c747b0e1ef568e0dd38d72/go/tools/echo/echo.go#L32)):*
> ```Go
> 
> const (
> 	DefaultDispatcherPath = "/run/shm/dispatcher/default.sock"
> ```

This isn't used. I would instead have below:
```
dispatcher = flag.String("dispatcher", "/run/shm/dispatcher/default.sock", "Path to dispatcher socket")
```

---

*[go/tools/echo/echo.go, line 52 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssM8NA4GvRUsLpsscY:-KssM8NA4GvRUsLpsscZ:btb8mjm) ([raw file](https://github.com/netsec-ethz/scion/blob/56b27da192e72ed779c747b0e1ef568e0dd38d72/go/tools/echo/echo.go#L52)):*
> ```Go
> 	dispatcher = flag.String("dispatcher", "", "Path to dispatcher socket")
> 	count      = flag.Int("count", 0,
> 		fmt.Sprintf("Number of echoes, between 0 and %d; "+
> ```

There's no need to split the string.
```
count = flag.Int("count, 0, 
    fmt.Sprintf("Number of echoes, between 0 and %d; a count of 0 means infinity", MaxEchoes))
```
reads better, and still fits into our 100-char max line length.

---

*[go/tools/echo/echo.go, line 128 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssOLUGjEFhmmI-5UUM:-KssOLUGjEFhmmI-5UUN:boicyru) ([raw file](https://github.com/netsec-ethz/scion/blob/56b27da192e72ed779c747b0e1ef568e0dd38d72/go/tools/echo/echo.go#L128)):*
> ```Go
> 		conn.SetDeadline(time.Time{})
> 		if err != nil {
> 			log.Error("Unable to read", "err", err)
> ```

Why continue the loop if read failed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1233)*

scrye:

---

*[go/tools/echo/echo.go, line 107 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmY6FJtKngGUQ2MCtz:-KssP-uNluxX9ne3umWf:b6e2ny) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L107)):*

There are still 3 fmt.Prinln/fmt.Printf calls left.
</blockquote></details>

I was trying to emulate `ping` as closely as possible. I'm possibly missing some use case here, do you want to run the tool with purely logging output?

---

*[go/tools/echo/echo.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmYophwfirlbzUJTYS:-KssQt9PZ0iusW-oj7jC:b11jm88) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L123)):*

That's not strictly true. E.g. if there were multiple paths with the same MTU, the client wouldn't have to care which of them were being used. Or, if there are multiple paths, and the client is using the minimum MTU, the same applies. In any case, for a datagram protocol it's vital that we expose the MTU to the client, otherwise they have absolutely no clue how large the packets they can send are, nor will they know about it changing (in either direction). This is not something the posix socket API envisioned, so we'll have to figure out a good way to extend it.
</blockquote></details>

Agreed. In my comment I was referring to _legacy_ apps, but failed to make this clear.

Yes, we clearly need to expose MTUs in the future. The real problem here is that we do not have have TCP right now, which would fix all of this.

Side note: we also need a way to standardize errors (right now, it's difficult to tell apart timeouts, from missing paths, mtu exceeded, etc.).

---

*[go/tools/echo/echo.go, line 32 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssLUUqwE7jnP_uWqnI:-KssPMHN1XQNJK-eWGCF:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/56b27da192e72ed779c747b0e1ef568e0dd38d72/go/tools/echo/echo.go#L32)):*

This isn't used. I would instead have below:
```
dispatcher = flag.String("dispatcher", "/run/shm/dispatcher/default.sock", "Path to dispatcher socket")
```
</blockquote></details>

Done.

---

*[go/tools/echo/echo.go, line 52 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssM8NA4GvRUsLpsscY:-KssPe_HVxy7_wZNxFGf:btim5rp) ([raw file](https://github.com/netsec-ethz/scion/blob/56b27da192e72ed779c747b0e1ef568e0dd38d72/go/tools/echo/echo.go#L52)):*

There's no need to split the string.
```
count = flag.Int("count, 0, 
    fmt.Sprintf("Number of echoes, between 0 and %d; a count of 0 means infinity", MaxEchoes))
```
reads better, and still fits into our 100-char max line length.
</blockquote></details>

Ah, it was due to tab width. I've updated my `vimrc`.

---

*[go/tools/echo/echo.go, line 128 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssOLUGjEFhmmI-5UUM:-KssSOYOhPcILpsNkiaO:b6nqxrp) ([raw file](https://github.com/netsec-ethz/scion/blob/56b27da192e72ed779c747b0e1ef568e0dd38d72/go/tools/echo/echo.go#L128)):*

Why continue the loop if read failed?
</blockquote></details>

This allows the tool to be used when the network experiences issues (e.g., loss of connectivity due to revocations), and also monitor how the network recovers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1233)*

kormat:
Reviewed 1 of 1 files at r3.

---

*[go/tools/echo/echo.go, line 107 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmY6FJtKngGUQ2MCtz:-Kssa594f5MZxDRh7Ge2:b6di8dz) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L107)):*

I was trying to emulate `ping` as closely as possible. I'm possibly missing some use case here, do you want to run the tool with purely logging output?
</blockquote></details>

Ah, i see. Well, i had been thinking we might run it as part of integration tests, in which case non-logging output isn't great. But if it's not intended to be used for that, then std output is fine.

---

*[go/tools/echo/echo.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmYophwfirlbzUJTYS:-KssYnHJOikCB7PMQz8I:bflif7j) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L123)):*

Agreed. In my comment I was referring to _legacy_ apps, but failed to make this clear.

Yes, we clearly need to expose MTUs in the future. The real problem here is that we do not have have TCP right now, which would fix all of this.

Side note: we also need a way to standardize errors (right now, it's difficult to tell apart timeouts, from missing paths, mtu exceeded, etc.).
</blockquote></details>

TCP has nothing to do with _datagram_ protocols, as specified above :package: 

The SIG currently has a hard-coded assumption about MTU. This is going to bite us in the real world sooner rather than later, and can't be fixed until snet exposes MTU (and updates) in a useful way.

Re: errors, there is a way to do this. `*common.Error` contains a `Data` field that can be used for exactly this. Have a look at `go/border/errors.go` for an example of examining the data to determine the error type.

---

*[go/tools/echo/echo.go, line 128 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssOLUGjEFhmmI-5UUM:-KssZo8xGY0-wDsrj4oT:bj9w6xu) ([raw file](https://github.com/netsec-ethz/scion/blob/56b27da192e72ed779c747b0e1ef568e0dd38d72/go/tools/echo/echo.go#L128)):*

This allows the tool to be used when the network experiences issues (e.g., loss of connectivity due to revocations), and also monitor how the network recovers.
</blockquote></details>

I think you're answering a different question (i.e. why not exit the program). If the read failed, why not `continue`? The remaining code below doesn't make much sense if there's no data to operate on.

---

*[go/tools/echo/echo.go, line 40 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssXojealv3K8xalPyL:-KssXojfF6szQxc49S9P:b-nlxubg) ([raw file](https://github.com/netsec-ethz/scion/blob/455e50458b1a7697024f118e93340efed717415b/go/tools/echo/echo.go#L40)):*
> ```Go
> 
> func GetDefaultSCIONDPath(ia *addr.ISD_AS) string {
> 	return fmt.Sprintf("/run/shm/sd%v.sock", ia)
> ```

This is missing `sciond/`.

---

*[go/tools/echo/Readme.md, line 10 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssXxhdmF5ilCpNDw1m:-KssXxhdmF5ilCpNDw1n:brhpe1s) ([raw file](https://github.com/netsec-ethz/scion/blob/455e50458b1a7697024f118e93340efed717415b/go/tools/echo/Readme.md#L10)):*
> ```Markdown
> ```
> 
> Finally, start the client using:
> ```

The examples can be simplified a lot now, by removing the `-sciond` and `-dispatcher` flags.

---

*[go/tools/echo/Readme.md, line 12 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssY0c5RqN04lY4XcFe:-KssY0c5RqN04lY4XcFf:b-wpv0z4) ([raw file](https://github.com/netsec-ethz/scion/blob/455e50458b1a7697024f118e93340efed717415b/go/tools/echo/Readme.md#L12)):*
> ```Markdown
> Finally, start the client using:
> ```
> go run echo.go -mode client -remote 2-25,[127.0.0.1]:40002 -local 1-19,[127.0.0.1]:0 \
> ```

Hurm. `:0` doesn't actually work:
```
invalid value "1-19,[127.0.0.1]:0" for flag -local: [desc Invalid port number port 0]
```
Ah. `snet.AddrFromString` is overly picky.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1233)*

scrye:
Review status: 1 of 5 files reviewed at latest revision, 6 unresolved discussions.

---

*[go/tools/echo/echo.go, line 107 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmY6FJtKngGUQ2MCtz:-Kssd7PDsQ9V1Od_RcJa:b4wmhrj) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L107)):*

Ah, i see. Well, i had been thinking we might run it as part of integration tests, in which case non-logging output isn't great. But if it's not intended to be used for that, then std output is fine.
</blockquote></details>

It will probably get modified a bit in the future to be useful for integration tests. For the moment its main function is just to debug SCION communication from the CLI.

---

*[go/tools/echo/echo.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmYophwfirlbzUJTYS:-KssdknRUq8J-qd90DSJ:bhzut3) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L123)):*

TCP has nothing to do with _datagram_ protocols, as specified above :package: 

The SIG currently has a hard-coded assumption about MTU. This is going to bite us in the real world sooner rather than later, and can't be fixed until snet exposes MTU (and updates) in a useful way.

Re: errors, there is a way to do this. `*common.Error` contains a `Data` field that can be used for exactly this. Have a look at `go/border/errors.go` for an example of examining the data to determine the error type.
</blockquote></details>

Well, the SIG is an atypical SCION app because it needs a lot of control over the network (e.g., policies, ToS, paths), which is typically implemented at OS level. Other SCION apps won't need these advanced features.

Regarding the SIG itself, yes, we need to expose the MTU at some point. This is quite tricky in practice, because the choice of path might depend on policy. So you want something like (build_header/metadata -> choose path from policy -> get path -> get mtu -> build_packet).

Thanks for the errors advice. As that needs a lot of changes in most of the network stack, I'm going to include it as a separate PR in the future.

---

*[go/tools/echo/echo.go, line 128 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssOLUGjEFhmmI-5UUM:-KsshlkVUEP4P5o0f3O6:b-hpacbf) ([raw file](https://github.com/netsec-ethz/scion/blob/56b27da192e72ed779c747b0e1ef568e0dd38d72/go/tools/echo/echo.go#L128)):*

I think you're answering a different question (i.e. why not exit the program). If the read failed, why not `continue`? The remaining code below doesn't make much sense if there's no data to operate on.
</blockquote></details>

Ah, yes, my bad. This used to be a fatal error in a previous version so `continue` didn't matter; forgot to add it when it got changed.

---

*[go/tools/echo/echo.go, line 40 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssXojealv3K8xalPyL:-Kssd45tWl8EX-8c0baY:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/455e50458b1a7697024f118e93340efed717415b/go/tools/echo/echo.go#L40)):*

This is missing `sciond/`.
</blockquote></details>

Done.

---

*[go/tools/echo/Readme.md, line 10 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssXxhdmF5ilCpNDw1m:-Kssi2RLdarmDFSz4hZT:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/455e50458b1a7697024f118e93340efed717415b/go/tools/echo/Readme.md#L10)):*

The examples can be simplified a lot now, by removing the `-sciond` and `-dispatcher` flags.
</blockquote></details>

Done.

---

*[go/tools/echo/Readme.md, line 12 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KssY0c5RqN04lY4XcFe:-KssiDZuBe30M2K2SeMc:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/455e50458b1a7697024f118e93340efed717415b/go/tools/echo/Readme.md#L12)):*

Hurm. `:0` doesn't actually work:
```
invalid value "1-19,[127.0.0.1]:0" for flag -local: [desc Invalid port number port 0]
```
Ah. `snet.AddrFromString` is overly picky.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1233)*

kormat:
Reviewed 4 of 4 files at r4.

---

*[go/tools/echo/echo.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-KsmYophwfirlbzUJTYS:-Ksspfl1ENxgs6jJfNYM:bo162ra) ([raw file](https://github.com/netsec-ethz/scion/blob/a7a72be340a95d24239d8d3c5ed4f70d795295cb/go/tools/echo/echo.go#L123)):*

Well, the SIG is an atypical SCION app because it needs a lot of control over the network (e.g., policies, ToS, paths), which is typically implemented at OS level. Other SCION apps won't need these advanced features.

Regarding the SIG itself, yes, we need to expose the MTU at some point. This is quite tricky in practice, because the choice of path might depend on policy. So you want something like (build_header/metadata -> choose path from policy -> get path -> get mtu -> build_packet).

Thanks for the errors advice. As that needs a lot of changes in most of the network stack, I'm going to include it as a separate PR in the future.
</blockquote></details>

(Discussed off line, we have a plan we're happy with for the (near)future)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1233)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1233#-:-KsxVPCXP0DA7KJSRhLM:bnfp4nl)*

PR #1234, opened 2017-08-29
Initial work for the hidden path communication

jonghoonkwon:
Major changes:
 - New class for hidden path configuration
 - Hidden path server stores hidden down-segments
 - Local beacon server registers the hidden down-segments to HPS
 - Local path server dispatches the hidden down segments from HPS
 - PCBRecordReq, PCBRecordReg contain PCB-set info
 - Topology generator generates the hidden path infrastructure using a param (e.g., -H True)
 - Integration test supports end2end communication test through hidden path

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1234)


jonghoonkwon:
@shitz @kormat @ercanucan  Hi folks, I just made a PR for the hidden pass communication.
It would be nice if you can have a look at the PR when you get a chance. :)

shitz:
Reviewed 8 of 33 files at r1.
Review status: 8 of 33 files reviewed at latest revision, 7 unresolved discussions.

---

*[c/lib/scion/checksum_bench, line 0 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1234#-KsnEDQtfM-NRHILdYQk:-KsnEDQtfM-NRHILdYQl:b-kep22t) ([raw file](https://github.com/netsec-ethz/scion/blob/03699e0b6a8407533c42c457b5dd130fdfb84d63/c/lib/scion/checksum_bench#L0)):*
Binary file. Shouldn't be added to the index

---

*[proto/path_mgmt.capnp, line 14 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1234#-KsnFCwH6Mz4VdN7T_bF:-KsnFCwH6Mz4VdN7T_bG:b9fnm84) ([raw file](https://github.com/netsec-ethz/scion/blob/03699e0b6a8407533c42c457b5dd130fdfb84d63/proto/path_mgmt.capnp#L14)):*
> ```
>     srcIA @0 :UInt32;
>     dstIA @1 :UInt32;
>     setID @2 :Data;
> ```

Just to clarify: This should be 8 bytes for the moment, the upper 32bit are the `ISD_AS` of the AS that manages the hidden path set, and the lower 32bit are randomly chosen. However, keep using `Data` instead of `UInt64` if we ever decide to increase the `setID` length. What should be changed is that this can be a list of `setID`s

---

*[proto/path_mgmt.capnp, line 24 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1234#-KsnHi-lzuZKGrjqF3bT:-KsnHi-lzuZKGrjqF3bU:bjsg1ay) ([raw file](https://github.com/netsec-ethz/scion/blob/03699e0b6a8407533c42c457b5dd130fdfb84d63/proto/path_mgmt.capnp#L24)):*
> ```
>     recs @0 :List(PCB.PathSegMeta);
>     revInfos @1 :List(RevInfo.RevInfo);
>     setInfos @2 :List(SetInfo.SetInfo);
> ```

Path registrations should not contain the entire ACL. Hidden path set configurations should be distributed out of band. What you want to do is add a `setIDs :List(Data);` field to `PCB.PathSegMeta` that can be set to include all the `setID`'s for which you want to register this hidden path segment.

---

*[python/beacon_server/local.py, line 146 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1234#-KsnUS5_kDF5tn28sNmf:-KsnUS5_kDF5tn28sNmg:bggvchy) ([raw file](https://github.com/netsec-ethz/scion/blob/03699e0b6a8407533c42c457b5dd130fdfb84d63/python/beacon_server/local.py#L146)):*
> ```Python
>         """
>         with self._rev_seg_lock:
>             self.up_segments.add_segment(pcb)
> ```

Hmm if you add the hidden path segment here, it will get registered as an `up-segment` but also loses all information that it's hidden. This could be a problem, if for example an organization only wants to use the hidden link when needed and otherwise the public. I'm aware that this would need some more engineering and maybe for a first implementation it's fine, however, it's something we should keep in mind.

---

*[python/beacon_server/local.py, line 225 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1234#-KsnWBAfL5w7elGahGtp:-KsnWBAfL5w7elGahGtq:bb8xdgd) ([raw file](https://github.com/netsec-ethz/scion/blob/03699e0b6a8407533c42c457b5dd130fdfb84d63/python/beacon_server/local.py#L225)):*
> ```Python
>         self._log_registrations(registered_paths, "hidden")
> 
>     def _get_path_via_api(self, isd_as, flush=False):
> ```

This should go into `SCIONElement` now that it's shared between the BS, CS, and PS.

---

*[python/lib/hps_config.py, line 50 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1234#-Ksnbjn1QNmWDQ92CUqb:-Ksnbjn1QNmWDQ92CUqc:b-5dazb4) ([raw file](https://github.com/netsec-ethz/scion/blob/03699e0b6a8407533c42c457b5dd130fdfb84d63/python/lib/hps_config.py#L50)):*
> ```Python
> 
> 
> class HPSProvider(object):
> ```

Could you explain to me the config format:
```
{
  "AuthorizedGroups": {
    "ETHZ": {
      "Providers": [
        "1-18"
      ],
      "Clients": [
        "1-14",
        "1-17",
        "2-26",
        "1-15",
        "1-16",
        "2-25",
        "1-19",
        "1-10",
        "2-23",
        "2-24"
      ],
      "HPS": "2-26"
    }
  },
  "HiddenASes": {
    "1-18": {
      "26": [
        {
          "SetID": "ETHZ",
          "HPS": "2-26"
        }
      ]
    }
  }
}
```

---

*[python/path_server/hidden.py, line 33 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1234#-KsnYgt-gwntps10-aaC:-KsnYgt-gwntps10-aaD:b-mm3u88) ([raw file](https://github.com/netsec-ethz/scion/blob/03699e0b6a8407533c42c457b5dd130fdfb84d63/python/path_server/hidden.py#L33)):*
> ```Python
> 
> 
> class HiddenPathServer(PathServer):
> ```

The hidden path server shouldn't be a `PathServer`, since that one contains many functionalities not needed by the `HiddenPathServer`. It should be a standalone service that only inherits from `SCIONElement`. All it does is handling path requests/registrations and authorization thereof. It doesn't need to do any path resolution. Also, it shouldn't forward revocations anywhere, but only process them when some arrive.

I suggest you remove the hidden path service from this PR and all the related endtoend test stuff. In a subsequent PR you can add an implementation of the hidden path service and the corresponding tests. Also there is no authentication/authorization of any kind. A simple scheme authentication scheme should be implemented.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1234)*

shitz:
This is heavily outdated and any hidden path work will not be done as part of this PR.

PR #1236, opened 2017-08-30
Add SCMP support to end-host packet parsing lib

scrye:
This PR adds SCMP parsing to lib hpkt. Also, the parsing code is now more flexible to allow complex extensions in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1236)


kormat:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[go/lib/hpkt/hpkt.go, line 30 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssdLqF9avWnK9xg52m:-KssdLqF9avWnK9xg52n:b-x4mpfj) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L30)):*
> ```Go
> )
> 
> // Processing/parsing callback type
> ```

This is enough new code that it could be moved to a separate file in the same module. Maybe `write.go` and `read.go`?

---

*[go/lib/hpkt/hpkt.go, line 31 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssdRmeFypsvNEaxhet:-KssdRmeFypsvNEaxheu:b-3vag6o) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L31)):*
> ```Go
> 
> // Processing/parsing callback type
> type PktParser func() error
> ```

Do this and `Offsets` need to be publicly visible?

---

*[go/lib/hpkt/hpkt.go, line 76 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssfSpJn3qyTXD0QNXO:-KssfSpJn3qyTXD0QNXP:bx6hab4) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L76)):*
> ```Go
> 		b:      b,
> 		cmnHdr: &spkt.CmnHdr{}}
> 	pCtx.CmnHdrParser = pCtx.DefaultCmnHdrParser
> ```

This can't be overridden, so it doesn't need to be here.

---

*[go/lib/hpkt/hpkt.go, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KsseL1fyCT3CP1ehK9U:-KsseL1fyCT3CP1ehK9V:bd1nmr4) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L87)):*
> ```Go
> }
> 
> // ParseScnPkt populates the SCION fields in s with information from b
> ```

This is a function between a constructor and a method for the `parseCtx` type.

---

*[go/lib/hpkt/hpkt.go, line 98 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssekmYQUGPsRKco77k:-KssekmYQUGPsRKco77l:b-1reoch) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L98)):*
> ```Go
> 	//  1. Common header
> 	//  2. Extension headers, in the order they are placed in the packet.
> 	//  Note that extension headers can also overwrite default behavior for
> ```

Specifically, hbh extensions can override 2-6, and e2e extensions can override 5-6 (an e2e ext can't touch anything that affects routing, as by definition the ext isn't inspected in transit).

---

*[go/lib/hpkt/hpkt.go, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssgKgxWZJF7sYqTsjH:-KssgKgxWZJF7sYqTsjI:b3ti97p) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L115)):*
> ```Go
> 	p.offset = p.extHdrOffsets.start
> ProtoLoop:
> 	for {
> ```

Hmm. Doing this in a loop leads to invalid ext header ordering being silently accepted. The rules for extension headers go like this:
1. HBH exts appear directly after the path header, and are limited to 3 by default.
1. If there is an scmp HBH extension present, it _must_ be the first extension, and raises the allowed HBH limit to 4 (so that a packet with 3 HBH exts that triggers an error can retain the original HBH exts for the reply).
1. E2E exts appear directly after all HBH exts, if any (and otherwise directly after the path header).

---

*[go/lib/hpkt/hpkt.go, line 119 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssgFIO7FHCyZhN7lQd:-KssgFIO7FHCyZhN7lQe:b8u8j52) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L119)):*
> ```Go
> 		case common.HopByHopClass:
> 			if err := p.HBHExtParser(); err != nil {
> 				return common.NewError("Unable to parse extension", "err", err)
> ```

HBH extension.

---

*[go/lib/hpkt/hpkt.go, line 150 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KsshdUXClt8DpMz3Fww:-KsshdUXClt8DpMz3Fwx:b-lkzjlt) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L150)):*
> ```Go
> 		return common.NewError("Unable to parse payload header", "err", err)
> 	}
> 	if err := p.ChecksumF(); err != nil {
> ```

This should probably just be done as part of parsing the l4 header. That parser will know  what if any checksum to use, and how to run it.

---

*[go/lib/hpkt/hpkt.go, line 173 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssiBw-giQ4jkHQTmqB:-KssiBw-giQ4jkHQTmqC:b-bxwqmq) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L173)):*
> ```Go
> 	// Parse 3-byte extension header first
> 	// We know the type of the next header, so we save it for the protocol loop
> 	p.cmnHdr.NextHdr = common.L4ProtocolType(p.b[p.offset])
> ```

It would be cleaner to have a `nextHdr` field in `parseCtx`.

---

*[go/lib/hpkt/hpkt.go, line 180 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssifZfxTHVKvLpMFjn:-KssifZfxTHVKvLpMFjo:b-jslft2) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L180)):*
> ```Go
> 	switch extnType {
> 	case common.ExtnSCMPType.Type:
> 		extn, cerr := scmp.ExtnFromRaw(p.b[p.offset+3:])
> ```

This needs to pass in only the bytes from the extension header. E.g. the traceroute extension uses the length of data to know much many hops it can store.

---

*[go/lib/hpkt/hpkt.go, line 191 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-Kssixxop8AkrWQiFw0b:-Kssixxop8AkrWQiFw0c:b-c22w4h) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L191)):*
> ```Go
> 	// Finished parsing another extension, advance the end offset
> 	p.extHdrOffsets.end += int(hdrLen * common.LineLen)
> 	p.offset += int(hdrLen * common.LineLen)
> ```

As these are the same value (right?), you could just do
```
p.offset += int(hdrLen * common.LineLen)
p.extHdrOffsets.end = p.offset
```

---

*[go/lib/hpkt/hpkt.go, line 274 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-Ksst9hBX0rGztrnOcsY:-Ksst9hBX0rGztrnOcsZ:br3pbbj) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L274)):*
> ```Go
> 	p.pldOffsets.start = p.offset
> 	pldLen := int(p.cmnHdr.TotalLen) - p.cmnHdr.HdrLenBytes() - p.s.L4.L4Len() -
> 		(p.extHdrOffsets.end - p.extHdrOffsets.start)
> ```

The common header parser should already have asserted that `p.cmnHdr.TotalLen == len(p.b)`. `l4.L4Header` also defines a `Validate` method, that takes the bytes remaining, and asserts that it matches the l4 header length field, which can be done in `DefaultL4HdrParser`. After that, all of the relevant lengths have been checked.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

scrye:
Review status: 0 of 6 files reviewed at latest revision, 12 unresolved discussions.

---

*[go/lib/hpkt/hpkt.go, line 30 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssdLqF9avWnK9xg52m:-KtBCcabi9rPUfpoE63V:b-x6i2d) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L30)):*

This is enough new code that it could be moved to a separate file in the same module. Maybe `write.go` and `read.go`?
</blockquote></details>

Done, moved to `read.go`.

---

*[go/lib/hpkt/hpkt.go, line 31 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssdRmeFypsvNEaxhet:-KtG5D1hqnr_ilz9Ufrw:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L31)):*

Do this and `Offsets` need to be publicly visible?
</blockquote></details>

Done.

---

*[go/lib/hpkt/hpkt.go, line 76 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssfSpJn3qyTXD0QNXO:-KtG5iUk6LHRzaSCRwTl:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L76)):*

This can't be overridden, so it doesn't need to be here.
</blockquote></details>

Done.

---

*[go/lib/hpkt/hpkt.go, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KsseL1fyCT3CP1ehK9U:-KtG66kAU5l29_8FDURh:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L87)):*

This is a function between a constructor and a method for the `parseCtx` type.
</blockquote></details>

Done.

---

*[go/lib/hpkt/hpkt.go, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssgKgxWZJF7sYqTsjH:-KtGBMPAfx34NHriZ8RB:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L115)):*

Hmm. Doing this in a loop leads to invalid ext header ordering being silently accepted. The rules for extension headers go like this:
1. HBH exts appear directly after the path header, and are limited to 3 by default.
1. If there is an scmp HBH extension present, it _must_ be the first extension, and raises the allowed HBH limit to 4 (so that a packet with 3 HBH exts that triggers an error can retain the original HBH exts for the reply).
1. E2E exts appear directly after all HBH exts, if any (and otherwise directly after the path header).
</blockquote></details>

Done.

---

*[go/lib/hpkt/hpkt.go, line 119 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssgFIO7FHCyZhN7lQd:-KtGCbIIAxIF1VBIKoia:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L119)):*

HBH extension.
</blockquote></details>

Done.

---

*[go/lib/hpkt/hpkt.go, line 150 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KsshdUXClt8DpMz3Fww:-KtGET0X2hocGz4SzSTK:bs2ky6d) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L150)):*

This should probably just be done as part of parsing the l4 header. That parser will know  what if any checksum to use, and how to run it.
</blockquote></details>

Done, this also meant moving payload parsing to l4 header parser. This makes sense anyway, since it's all dictated by the l4 header.

---

*[go/lib/hpkt/hpkt.go, line 173 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssiBw-giQ4jkHQTmqB:-KtGF6M4-lh1QM03XXI4:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L173)):*

It would be cleaner to have a `nextHdr` field in `parseCtx`.
</blockquote></details>

Done.

---

*[go/lib/hpkt/hpkt.go, line 180 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KssifZfxTHVKvLpMFjn:-KtGFZB2VULbzbg3D1kE:bl0vbkk) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L180)):*

This needs to pass in only the bytes from the extension header. E.g. the traceroute extension uses the length of data to know much many hops it can store.
</blockquote></details>

Hmm, this is quite dangerous because currently the function looks very innocent. I'll add a docstring to the function specifying this.

Done.

---

*[go/lib/hpkt/hpkt.go, line 191 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-Kssixxop8AkrWQiFw0b:-KtGObXhSeVzBmAN-QJ3:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L191)):*

As these are the same value (right?), you could just do
```
p.offset += int(hdrLen * common.LineLen)
p.extHdrOffsets.end = p.offset
```
</blockquote></details>

Done.

---

*[go/lib/hpkt/hpkt.go, line 274 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-Ksst9hBX0rGztrnOcsY:-KtGdbwdCncISPoUMopA:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L274)):*

The common header parser should already have asserted that `p.cmnHdr.TotalLen == len(p.b)`. `l4.L4Header` also defines a `Validate` method, that takes the bytes remaining, and asserts that it matches the l4 header length field, which can be done in `DefaultL4HdrParser`. After that, all of the relevant lengths have been checked.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

kormat:
Reviewed 5 of 6 files at r2.
Review status: 5 of 6 files reviewed at latest revision, 13 unresolved discussions.

---

*[go/lib/common/pld.go, line 26 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGhOXoL77RWy5k5mUx:-KtGhOXoL77RWy5k5mUy:b-lyem4d) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/common/pld.go#L26)):*
> ```Go
> 	Copy() (Payload, *Error)
> 	// Copies data from payload to b
> 	WritePld(b RawBytes) (int, *Error)
> ```

This looks like a rebase issue?

---

*[go/lib/hpkt/hpkt.go, line 274 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-Ksst9hBX0rGztrnOcsY:-KtGj8z6GkTFWYLBa5Sz:bihdv4g) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L274)):*

Done.
</blockquote></details>

This code didn't change.

---

*[go/lib/hpkt/read.go, line 83 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGmbPJ-epD-TvMDYCY:-KtGmbPJ-epD-TvMDYCZ:bqc4ozx) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L83)):*
> ```Go
> 		b:        b,
> 		cmnHdr:   &spkt.CmnHdr{},
> 		hbhLimit: common.ExtnMaxHBH}
> ```

In general, if a literal starts with a `{\n`, it should end with `\n}`, for readability.

---

*[go/lib/hpkt/read.go, line 122 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGlz8zPW5dJ5xj3dnU:-KtGlz8zPW5dJ5xj3dnV:bobks3f) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L122)):*
> ```Go
> 	// header.
> ProtoLoop:
> 	for {
> ```

Doing this in a single big loop still makes it somewhat hard to follow.  E.g. it wasn't until i read the details of `DefaultHBHExtParser` could i see that SCMP was handled correctly. I would instead have 2 different methods. `parseHBHExts`, `parseE2EExts`. Both would return as soon as `p.nextHdr` isn't the appropriate class.

---

*[go/lib/hpkt/read.go, line 127 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGkEv7pDko91cWWLk4:-KtGkEv7pDko91cWWLk5:b40og8q) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L127)):*
> ```Go
> 			p.hbhCounter += 1
> 			if p.hbhCounter > p.hbhLimit {
> 				return common.NewError("HBH extension limit exceeded", "limit", p.hbhLimit)
> ```

Include the type and position of the current extension.

---

*[go/lib/hpkt/read.go, line 131 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGkHinaPcvhAzThgbV:-KtGkHinaPcvhAzThgbW:b40og8q) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L131)):*
> ```Go
> 			if p.parsedFirstE2E {
> 				return common.NewError("Invalid placement of HBH extension, must appear before " +
> 					"E2E extensions but found after")
> ```

Include the type and position of the current extension.

---

*[go/lib/hpkt/read.go, line 208 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGkyULJPP8_oQIsHCm:-KtGkyULJPP8_oQIsHCn:b-5fm70u) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L208)):*
> ```Go
> 		extn, cerr := scmp.ExtnFromRaw(p.b[p.offset+common.ExtnSubHdrLen : p.extHdrOffsets.end])
> 		if cerr != nil {
> 			return common.NewError("Unable to parse extension header", "err", cerr)
> ```

Include ext type and position.

---

*[go/lib/hpkt/read.go, line 212 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGl5VIfDDr_XFmSwjN:-KtGl5VIfDDr_XFmSwjO:bfgo7lm) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L212)):*
> ```Go
> 		p.s.HBHExt = append(p.s.HBHExt, extn)
> 	default:
> 		return common.NewError("Unsupported HBH extension type", "type", extnType)
> ```

Include the ext position.

---

*[go/lib/hpkt/write.go, line 34 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGnq8v5hnjxO27hvij:-KtGnq8v5hnjxO27hvik:b1afq7o) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/write.go#L34)):*
> ```Go
> 
> 	if s.L4.L4Type() != common.L4UDP {
> 		return 0, common.NewError("Unsupported protocol", "expected",
> ```

Is this actually needed? If another header type was used, it would "Just Work™", as this function only interacts with the l4 header via the interface, right?

---

*[go/lib/scmp/extn.go, line 38 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGmnhyOHuK8_cotgEY:-KtGmnhyOHuK8_cotgEZ:b-rj5hgu) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/scmp/extn.go#L38)):*
> ```Go
> // include the sub extension header. Because some extensions rely on the length
> // of b during parsing, b should contain the entire extension header and no
> // additional bytes.
> ```

Well, while this is correct, this comment would belong much better on the places this is _called_, as anywhere that calls ext parsing is the only correct/possible place to handle this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

scrye:
Review status: 3 of 6 files reviewed at latest revision, 13 unresolved discussions.

---

*[go/lib/common/pld.go, line 26 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGhOXoL77RWy5k5mUx:-KtGwcPMV0Lpc6DKnCWK:b-x39vi9) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/common/pld.go#L26)):*

This looks like a rebase issue?
</blockquote></details>

You mean the `b`? I added it manually for documentation purposes, as the comment above didn't make sense.

---

*[go/lib/hpkt/hpkt.go, line 274 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-Ksst9hBX0rGztrnOcsY:-KtGwrwZIS-1PFS1W-e0:bke732s) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L274)):*

This code didn't change.
</blockquote></details>

Moving to a new file broke Reviewable, I should avoid collapsing commits which include moves before pushing.

The  `Validate` is in `DefaultL4Parser` (line 290) and the total packet length check is in `CmnHdrParser`(line 149)  in `go/lib/hpkt/read.go`.

---

*[go/lib/hpkt/read.go, line 83 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGmbPJ-epD-TvMDYCY:-KtGyU632MbFa12HQyK3:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L83)):*

In general, if a literal starts with a `{\n`, it should end with `\n}`, for readability.
</blockquote></details>

Done.

---

*[go/lib/hpkt/read.go, line 122 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGlz8zPW5dJ5xj3dnU:-KtH-EqjMUUdKbbNBaOi:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L122)):*

Doing this in a single big loop still makes it somewhat hard to follow.  E.g. it wasn't until i read the details of `DefaultHBHExtParser` could i see that SCMP was handled correctly. I would instead have 2 different methods. `parseHBHExts`, `parseE2EExts`. Both would return as soon as `p.nextHdr` isn't the appropriate class.
</blockquote></details>

Done.

---

*[go/lib/hpkt/read.go, line 127 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGkEv7pDko91cWWLk4:-KtH-zLYahA6ZAmFiUpQ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L127)):*

Include the type and position of the current extension.
</blockquote></details>

Done.

---

*[go/lib/hpkt/read.go, line 131 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGkHinaPcvhAzThgbV:-KtH-aE6Q-DE4j3VSorD:bduyt6b) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L131)):*

Include the type and position of the current extension.
</blockquote></details>

This type of error is now caught by the L4 parser.

---

*[go/lib/hpkt/read.go, line 208 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGkyULJPP8_oQIsHCm:-KtH0DDM_2i5xNBLTL7i:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L208)):*

Include ext type and position.
</blockquote></details>

Done.

---

*[go/lib/hpkt/read.go, line 212 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGl5VIfDDr_XFmSwjN:-KtH0JvwfyMQSYnYVm5E:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L212)):*

Include the ext position.
</blockquote></details>

Done.

---

*[go/lib/hpkt/write.go, line 34 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGnq8v5hnjxO27hvij:-KtH0aVnFRJHRK7bE5za:b-h7weom) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/write.go#L34)):*

Is this actually needed? If another header type was used, it would "Just Work™", as this function only interacts with the l4 header via the interface, right?
</blockquote></details>

Don't know, looks like it should work but we are never using it to write other protocols. Should be ok though.

---

*[go/lib/scmp/extn.go, line 38 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGmnhyOHuK8_cotgEY:-KtH0orAwYgEnTL-c4wP:b-dm2id6) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/scmp/extn.go#L38)):*

Well, while this is correct, this comment would belong much better on the places this is _called_, as anywhere that calls ext parsing is the only correct/possible place to handle this.
</blockquote></details>

When I godoc the call this shows up and its uses don't, so I think it's better to have such caveats close to definitions. I like just pressing shift+K and getting all the info instantly without having to go digging.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

kormat:
Reviewed 1 of 6 files at r2, 2 of 2 files at r3.

---

*[go/lib/hpkt/hpkt.go, line 274 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-Ksst9hBX0rGztrnOcsY:-KtH2CfVylRhlQOxRFOl:bctmj8e) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L274)):*

Moving to a new file broke Reviewable, I should avoid collapsing commits which include moves before pushing.

The  `Validate` is in `DefaultL4Parser` (line 290) and the total packet length check is in `CmnHdrParser`(line 149)  in `go/lib/hpkt/read.go`.
</blockquote></details>

The code i'm referring to is read.go:288
```
pldLen := int(p.cmnHdr.TotalLen) - p.cmnHdr.HdrLenBytes() - p.s.L4.L4Len() -
    (p.extHdrOffsets.end - p.extHdrOffsets.start)
```

---

*[go/lib/hpkt/read.go, line 127 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGkEv7pDko91cWWLk4:-KtH3Dq99DmhpJtpZ0I1:bbwmu1t) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L127)):*

Done.
</blockquote></details>

Mm, the position should probably be reported 0-indexed, otherwise it gets confusing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

scrye:
Review status: 5 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/lib/hpkt/hpkt.go, line 274 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-Ksst9hBX0rGztrnOcsY:-KtH6QrwUoajA2uT7H9N:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/00b512c54ae0db8ef210dbd8b81de5e4c6837517/go/lib/hpkt/hpkt.go#L274)):*

The code i'm referring to is read.go:288
```
pldLen := int(p.cmnHdr.TotalLen) - p.cmnHdr.HdrLenBytes() - p.s.L4.L4Len() -
    (p.extHdrOffsets.end - p.extHdrOffsets.start)
```
</blockquote></details>

Done.

---

*[go/lib/hpkt/read.go, line 127 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGkEv7pDko91cWWLk4:-KtH6sJMlsZZsZwBvqJ6:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/hpkt/read.go#L127)):*

Mm, the position should probably be reported 0-indexed, otherwise it gets confusing.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

kormat:
Reviewed 1 of 1 files at r4.

---

*[go/lib/scmp/extn.go, line 38 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGmnhyOHuK8_cotgEY:-KtH7Fyj17-BJtQZbB-x:bdbjkhu) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/scmp/extn.go#L38)):*

When I godoc the call this shows up and its uses don't, so I think it's better to have such caveats close to definitions. I like just pressing shift+K and getting all the info instantly without having to go digging.
</blockquote></details>

What about all the other extensions, then? You've added the comment to one of the extensions that _don't_ care about this. That's a pretty undiscoverable location.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

kormat:

---

*[go/lib/scmp/extn.go, line 38 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGmnhyOHuK8_cotgEY:-KtH7zigZ9HL8Fjd2jpR:bsu1gou) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/scmp/extn.go#L38)):*

What about all the other extensions, then? You've added the comment to one of the extensions that _don't_ care about this. That's a pretty undiscoverable location.
</blockquote></details>

Ideally the description would be in a (fairly) obvious central location, and then all places that is relevant can refer to that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

scrye:

---

*[go/lib/scmp/extn.go, line 38 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGmnhyOHuK8_cotgEY:-KtHIf-dRm0uT7rM9ur8:b-pvuq7d) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/scmp/extn.go#L38)):*

Ideally the description would be in a (fairly) obvious central location, and then all places that is relevant can refer to that.
</blockquote></details>

Hmm, since this is a constructor I guess its usage is on a case by case basis. So yeah, other constructors that require callers to do _special stuff_ should include the comment.

Since this one doesn't do anything, I just deleted the comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

kormat:
Reviewed 1 of 1 files at r5.

---

*[go/lib/common/pld.go, line 26 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGhOXoL77RWy5k5mUx:-KtHJLPTvijFrbrjXlIl:bwwlae1) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/common/pld.go#L26)):*

You mean the `b`? I added it manually for documentation purposes, as the comment above didn't make sense.
</blockquote></details>

Ah. I would just rephrase the comment, `// Writes payload to buffer`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

scrye:

---

*[go/lib/common/pld.go, line 26 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-KtGhOXoL77RWy5k5mUx:-KtHKUxwyC6wBxT8lxmw:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/53057b81fbe7dda23d21786761be8976afd797cb/go/lib/common/pld.go#L26)):*

Ah. I would just rephrase the comment, `// Writes payload to buffer`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1236#-:-KtHKpMCp1dHKVWlN3uW:bnfp4nl)*

PR #1239, opened 2017-09-01
Add reply message to SCIOND revocation notifications

scrye:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1239)


scrye:
Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[python/sciond/sciond.py, line 341 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KsxfqdoReoGEM34NwSl:-KsxfqdoReoGEM34NwSm:bxvr32x) ([raw file](https://github.com/netsec-ethz/scion/blob/8d09f56a42cd2b36a01ed0803be01a4e39482378/python/sciond/sciond.py#L341)):*
> ```Python
>         self.handle_revocation(rev_info, meta)
> 
>     def handle_revocation(self, rev_info, meta, id_=-1):
> ```

For revocations coming via the `ReliableSocket`, the `id_` is needed to build the reply. However, `handle_revocation` is also called when SCIOND gets an SCMP message directly; in this case, `id_` is undefined.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

shitz:
Review status: 0 of 9 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/lib/sciond/sciond.go, line 252 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-Kt2Z-piVAX3Uaq_RJQF:-Kt2Z-piVAX3Uaq_RJQG:bctbds9) ([raw file](https://github.com/netsec-ethz/scion/blob/6bd81a89e804d521fff28520bc1eccc2da9ee74a/go/lib/sciond/sciond.go#L252)):*
> ```Go
> 
> // RevNotification sends a RevocationInfo message to SCIOND.
> func (c *Connector) RevNotification(info []byte) (*RevReply, error) {
> ```

Hmm is this the right level of abstraction? Doesn't `snet` or the `pathmgr` have to parse the revocation anyway to then later remove paths that contain the revoked interface and thus `RevNotification` could take `RevInfo`as a parameter?

---

*[go/lib/sciond/types.go, line 197 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-Kt2aBWx8lo5L634oZ19:-Kt2aBWx8lo5L634oZ1A:b-n3mfwo) ([raw file](https://github.com/netsec-ethz/scion/blob/6bd81a89e804d521fff28520bc1eccc2da9ee74a/go/lib/sciond/types.go#L197)):*
> ```Go
> const (
> 	RemovedSegments StatusCode = iota
> 	EpochOk
> ```

Hmm actually just the epoch the app could easily check itself: `crypto.VerifyHashTreeEpoch(revInfo.Epoch)`. The status codes could then just be 
```
RevocationValid,
RevocationInvalid,
Unknown
```
or something. As soon as we switch to self-contained revocation authentication, `Unknown` can be dropped. The only use-case for the `SCIOND` response I can see in the future is if you are a power/memory/CPU-limited device that cannot do signature verification and you want to outsource this to SCIOND.
Until then, apps would just do the checks that we discussed offline (including the epoch verification) and if those succeed remove paths containing revoked interfaces.

---

*[python/lib/sciond_api/revocation.py, line 67 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-Kt2daaFuI-bRNHsnXUd:-Kt2daaFuI-bRNHsnXUe:btrjjnp) ([raw file](https://github.com/netsec-ethz/scion/blob/6bd81a89e804d521fff28520bc1eccc2da9ee74a/python/lib/sciond_api/revocation.py#L67)):*
> ```Python
> 
> 
> class SCIONDRevReplyStatus:  # pragma: no cover
> ```

If you make this a `TypeBase` subclass you get a `to_str()` method for free.

---

*[python/sciond/sciond.py, line 341 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KsxfqdoReoGEM34NwSl:-Kt2e6aVVRauXaZAhBvu:bfzegzm) ([raw file](https://github.com/netsec-ethz/scion/blob/8d09f56a42cd2b36a01ed0803be01a4e39482378/python/sciond/sciond.py#L341)):*

For revocations coming via the `ReliableSocket`, the `id_` is needed to build the reply. However, `handle_revocation` is also called when SCIOND gets an SCMP message directly; in this case, `id_` is undefined.
</blockquote></details>

When SCIOND receives a `RevNotification` it shouldn't directly call `handle_revocation`. Create a function `_api_handle_rev_notification` that then calls `handle_revocation`. `handle_revocation` should be extended to return a status code that then ca be used by `_api_handle_revocation` to put in the reply.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

kormat:
Reviewed 5 of 8 files at r1, 4 of 4 files at r2.

---

*[go/lib/sciond/sciond.go, line 252 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-Kt2Z-piVAX3Uaq_RJQF:-KtAzLYeyQTpkCdTm7Rf:b-n5jle6) ([raw file](https://github.com/netsec-ethz/scion/blob/6bd81a89e804d521fff28520bc1eccc2da9ee74a/go/lib/sciond/sciond.go#L252)):*

Hmm is this the right level of abstraction? Doesn't `snet` or the `pathmgr` have to parse the revocation anyway to then later remove paths that contain the revoked interface and thus `RevNotification` could take `RevInfo`as a parameter?
</blockquote></details>

Not necessarily. We shouldn't design the sciond lib api to assume that snet/pathmgr etc are being used, in any case. We could provide two api calls, `RevNotificationFromRaw` and `RevNotification`, with the former calling the latter.

---

*[go/lib/sciond/types.go, line 197 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-Kt2aBWx8lo5L634oZ19:-KtAwLuAzUs14gpqRmfa:b82kg3) ([raw file](https://github.com/netsec-ethz/scion/blob/6bd81a89e804d521fff28520bc1eccc2da9ee74a/go/lib/sciond/types.go#L197)):*

Hmm actually just the epoch the app could easily check itself: `crypto.VerifyHashTreeEpoch(revInfo.Epoch)`. The status codes could then just be 
```
RevocationValid,
RevocationInvalid,
Unknown
```
or something. As soon as we switch to self-contained revocation authentication, `Unknown` can be dropped. The only use-case for the `SCIOND` response I can see in the future is if you are a power/memory/CPU-limited device that cannot do signature verification and you want to outsource this to SCIOND.
Until then, apps would just do the checks that we discussed offline (including the epoch verification) and if those succeed remove paths containing revoked interfaces.
</blockquote></details>

Signature verification requires handling/requesting TRCs/Certs, which apps will not do (outside of very specialised security-sensitive apps, perhaps). We will still want to use sciond for this. You also don't want to depend on every app correctly implementing signature verification, because that's not going to end well. Checking the epoch in two places is also racey (if the app says the epoch is valid, and sciond checks slightly later and says that it isn't...). In that context, it's not that that the revocation is _invalid_ (which implies that it should be treated as an error, or malicious), just that it's too old (and therefore not fitting into the 3 categories you've listed). Something like this might be closer to what we want:

- RevValid
- RevTooOld
- RevInvalid
- RevUnknown

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

scrye:
*[go/lib/sciond/types.go, line 197 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-Kt2aBWx8lo5L634oZ19:-KtAzLBdHVTwNDmdTmOZ:bik7h2w) ([raw file](https://github.com/netsec-ethz/scion/blob/6bd81a89e804d521fff28520bc1eccc2da9ee74a/go/lib/sciond/types.go#L197)):*

Hmm actually just the epoch the app could easily check itself: `crypto.VerifyHashTreeEpoch(revInfo.Epoch)`. The status codes could then just be 
```
RevocationValid,
RevocationInvalid,
Unknown
```
or something. As soon as we switch to self-contained revocation authentication, `Unknown` can be dropped. The only use-case for the `SCIOND` response I can see in the future is if you are a power/memory/CPU-limited device that cannot do signature verification and you want to outsource this to SCIOND.
Until then, apps would just do the checks that we discussed offline (including the epoch verification) and if those succeed remove paths containing revoked interfaces.
</blockquote></details>

I agree that in the future we want to do most of it in Go. However, verifying the epoch in Go right now splits the semantics behind RevInfo validation in two places (and the python code checks the epoch anyway).

The purpose behind the larger number of status codes is that `Unknown` does not have any clear semantics behind it. If we have detailed status codes, the Go code can just say, ok, these status codes map to valid, these to invalid and these to unknown as per our discussion. I think it's useful to know _exactly_ what caused the failure, so the Go code can log it / make more informed decisions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

scrye:
Review status: 5 of 9 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/lib/sciond/sciond.go, line 252 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-Kt2Z-piVAX3Uaq_RJQF:-KtBAVjQSOjAMfuLG-gQ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/6bd81a89e804d521fff28520bc1eccc2da9ee74a/go/lib/sciond/sciond.go#L252)):*

Not necessarily. We shouldn't design the sciond lib api to assume that snet/pathmgr etc are being used, in any case. We could provide two api calls, `RevNotificationFromRaw` and `RevNotification`, with the former calling the latter.
</blockquote></details>

Done.

---

*[python/lib/sciond_api/revocation.py, line 67 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-Kt2daaFuI-bRNHsnXUd:-KtB0xNmUXTlyKUKDL4t:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/6bd81a89e804d521fff28520bc1eccc2da9ee74a/python/lib/sciond_api/revocation.py#L67)):*

If you make this a `TypeBase` subclass you get a `to_str()` method for free.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 341 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KsxfqdoReoGEM34NwSl:-KtB4qdqMJUPOqpZuQsN:bfbq99i) ([raw file](https://github.com/netsec-ethz/scion/blob/8d09f56a42cd2b36a01ed0803be01a4e39482378/python/sciond/sciond.py#L341)):*

When SCIOND receives a `RevNotification` it shouldn't directly call `handle_revocation`. Create a function `_api_handle_rev_notification` that then calls `handle_revocation`. `handle_revocation` should be extended to return a status code that then ca be used by `_api_handle_revocation` to put in the reply.
</blockquote></details>

Good point, done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

shitz:
Review status: 5 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/lib/sciond/types.go, line 197 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-Kt2aBWx8lo5L634oZ19:-KtBS85g8m09AIUiZVb4:bn7z26x) ([raw file](https://github.com/netsec-ethz/scion/blob/6bd81a89e804d521fff28520bc1eccc2da9ee74a/go/lib/sciond/types.go#L197)):*

I agree that in the future we want to do most of it in Go. However, verifying the epoch in Go right now splits the semantics behind RevInfo validation in two places (and the python code checks the epoch anyway).

The purpose behind the larger number of status codes is that `Unknown` does not have any clear semantics behind it. If we have detailed status codes, the Go code can just say, ok, these status codes map to valid, these to invalid and these to unknown as per our discussion. I think it's useful to know _exactly_ what caused the failure, so the Go code can log it / make more informed decisions.
</blockquote></details>

Lets go with @kormat's proposal

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

scrye:
Review status: 4 of 15 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/lib/sciond/types.go, line 197 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-Kt2aBWx8lo5L634oZ19:-KtC06KyX3Fkt7rOvrrc:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/6bd81a89e804d521fff28520bc1eccc2da9ee74a/go/lib/sciond/types.go#L197)):*

Lets go with @kormat's proposal
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

shitz:
:lgtm:

---

Reviewed 3 of 8 files at r1, 1 of 4 files at r2, 1 of 4 files at r3, 8 of 10 files at r4.
Review status: 13 of 15 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[python/sciond/sciond.py, line 361 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC15_uXehAxxui5lWz:-KtC15_uXehAxxui5lX-:bhukl5s) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L361)):*
> ```Python
> 
>         if epoch_status == ConnectedHashTree.EPOCH_FUTURE:
>             logging.debug(
> ```

This should probably we a warning. Also the text should reflect that SCIOND received a revocation from the future.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-:-KtC1JwwsgRg9lNVNHOf:bnfp4nl)*

kormat:
Reviewed 2 of 4 files at r3, 10 of 10 files at r4.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/sciond/types.go, line 193 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC363Q2j12aH2anVw-:-KtC363Q2j12aH2anVw0:b-fe6app) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/go/lib/sciond/types.go#L193)):*
> ```Go
> }
> 
> type StatusCode uint16
> ```

This name is very generic.

---

*[go/lib/sciond/types.go, line 195 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC23eDWljL8Dly3-hu:-KtC23eDWljL8Dly3-hv:bemhlzv) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/go/lib/sciond/types.go#L195)):*
> ```Go
> type StatusCode uint16
> 
> func (c StatusCode) String() string {
> ```

For this kind of constant type declaration, our convention is
```
type ...
const (
   val type
)
...
func ... String()
```
See go/lib/common/l4.go for an example.

---

*[python/lib/crypto/hash_tree.py, line 263 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC3esvnFMktnBbSD6m:-KtC3esvnFMktnBbSD6n:b-40gkw0) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/lib/crypto/hash_tree.py#L263)):*
> ```Python
>         gap_time = cls.get_time_since_epoch()
>         if (epoch == cur_epoch or
>                 cur_epoch == epoch + 1 and gap_time < HASHTREE_EPOCH_TOLERANCE):
> ```

Hmm. Looking at this, have we confirmed the operator precedence does what we want it to do?

---

*[python/lib/crypto/hash_tree.py, line 266 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC4OhnAW_HTevL71NW:-KtC4OhnAW_HTevL71NX:brk7d4l) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/lib/crypto/hash_tree.py#L266)):*
> ```Python
>             return cls.EPOCH_OK
>         if epoch < cur_epoch:
>             return cls.EPOCH_PAST
> ```

This fails to distinguish between an epoch that was very recent (i.e. <= 2 epochs different from cur_epoch), and one that wasn't. The former is probably legitimate, the latter is definitely not. This also effects how the failure should be logged.

---

*[python/sciond/sciond.py, line 376 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC5DuuYwzR6QXIA7x7:-KtC5DuuYwzR6QXIA7x8:bruyow) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L376)):*
> ```Python
>             return SCIONDRevReplyStatus.VALID
>         else:
>             return SCIONDRevReplyStatus.UNKNOWN
> ```

This logic isn't sound. Removing 0 segments does not imply that the revocation could not be verified. E.g. if the revocation is for a peer interface, normally no segments should be removed, but the revocation is still verified.

---

*[python/sciond/sciond.py, line 392 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC5WPaP7cnPw1yF0I_:-KtC5WPaP7cnPw1yF0Ia:bc36kll) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L392)):*
> ```Python
>         """
> 
>         if ConnectedHashTree.verify_epoch(rev_info.p.epoch) != ConnectedHashTree.EPOCH_OK:
> ```

Hum. I'm not convinced by this check, actually. It ignores the grace period. It's also duplicating the checks in `handle_revocation`, which is the only place this is called from. I think it can just be dropped.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

scrye:
Review status: 9 of 16 files reviewed at latest revision, 7 unresolved discussions.

---

*[go/lib/sciond/types.go, line 193 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC363Q2j12aH2anVw-:-KtCaxbD2mRAD4HM5Ly3:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/go/lib/sciond/types.go#L193)):*

This name is very generic.
</blockquote></details>

Done.

---

*[go/lib/sciond/types.go, line 195 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC23eDWljL8Dly3-hu:-KtCbHRFJ1Kpu5dA1gzR:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/go/lib/sciond/types.go#L195)):*

For this kind of constant type declaration, our convention is
```
type ...
const (
   val type
)
...
func ... String()
```
See go/lib/common/l4.go for an example.
</blockquote></details>

Done.

---

*[python/lib/crypto/hash_tree.py, line 263 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC3esvnFMktnBbSD6m:-KtCe-NEO3gsdVAg6iOS:blmb3h2) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/lib/crypto/hash_tree.py#L263)):*

Hmm. Looking at this, have we confirmed the operator precedence does what we want it to do?
</blockquote></details>

`and` has precedence over `or`, so yes, it seems like it's ok.

---

*[python/lib/crypto/hash_tree.py, line 266 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC4OhnAW_HTevL71NW:-KtCjLozbfx8Owpzk6r_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/lib/crypto/hash_tree.py#L266)):*

This fails to distinguish between an epoch that was very recent (i.e. <= 2 epochs different from cur_epoch), and one that wasn't. The former is probably legitimate, the latter is definitely not. This also effects how the failure should be logged.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 361 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC15_uXehAxxui5lWz:-KtClybk5IaLV8z25HFT:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L361)):*

This should probably we a warning. Also the text should reflect that SCIOND received a revocation from the future.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 376 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC5DuuYwzR6QXIA7x7:-KtCmGVHNrMNQXFClJlE:bjj26nn) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L376)):*

This logic isn't sound. Removing 0 segments does not imply that the revocation could not be verified. E.g. if the revocation is for a peer interface, normally no segments should be removed, but the revocation is still verified.
</blockquote></details>

I'm massively struggling to map error conditions to external results, I'm possibly missing some scenarios. For a complete picture, this covers the whole possibility space:
  a. epoch in the distant (more than 2 epochs) past, nothing is revoked
  b. epoch in the near past, nothing is revoked
  c. epoch ok (current or past 1 but within threshold), nothing is revoked
   c1. nothing was revoked because we did not find any matching segments
   c2. nothing was revoked because we matched a peering link
   c3. nothing was revoked because of crypto errors
  d. epoch ok, some segments are revoked
  e. epoch in the future, nothing is revoked

As per suggested results, these need to be mapped to:
  1. revocation valid
  2. revocation too old
  3. revocation invalid
  4. revocation unknown

Distinguishing between c1, c2 and c3 is not possible without changing `ConnectedHashTree.verify`.

The closest mapping I can think of:
  1. d
  2. a
  3. e
  4.. b, c1, c2, c3

External applications have a binary choice (remove/not remove) to make according to 1-4:
  - revoke paths which match AS+IFID: 1, 4
  - do not revoke: 2, 3

---

*[python/sciond/sciond.py, line 392 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC5WPaP7cnPw1yF0I_:-KtCqXnOYRMHDOb__t9k:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L392)):*

Hum. I'm not convinced by this check, actually. It ignores the grace period. It's also duplicating the checks in `handle_revocation`, which is the only place this is called from. I think it can just be dropped.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

shitz:
Reviewed 1 of 10 files at r4.
Review status: 9 of 16 files reviewed at latest revision, 6 unresolved discussions.

---

*[python/sciond/sciond.py, line 376 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC5DuuYwzR6QXIA7x7:-KtCyuLE4fpqcVFTYxy7:b-4o54m4) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L376)):*

I'm massively struggling to map error conditions to external results, I'm possibly missing some scenarios. For a complete picture, this covers the whole possibility space:
  a. epoch in the distant (more than 2 epochs) past, nothing is revoked
  b. epoch in the near past, nothing is revoked
  c. epoch ok (current or past 1 but within threshold), nothing is revoked
   c1. nothing was revoked because we did not find any matching segments
   c2. nothing was revoked because we matched a peering link
   c3. nothing was revoked because of crypto errors
  d. epoch ok, some segments are revoked
  e. epoch in the future, nothing is revoked

As per suggested results, these need to be mapped to:
  1. revocation valid
  2. revocation too old
  3. revocation invalid
  4. revocation unknown

Distinguishing between c1, c2 and c3 is not possible without changing `ConnectedHashTree.verify`.

The closest mapping I can think of:
  1. d
  2. a
  3. e
  4.. b, c1, c2, c3

External applications have a binary choice (remove/not remove) to make according to 1-4:
  - revoke paths which match AS+IFID: 1, 4
  - do not revoke: 2, 3
</blockquote></details>

Why would b map to 4? A revocation is too old if it's not for the current epoch or the previous epoch + grace period. What @kormat meant with current epoch > 2 epochs in the past is that that case might be considered malicious/misconfigured and should be logged as such (same as an epoch from the future). Whether we want to expose this to the app or just let SCIOND log a warning remains to be seen.
A revocation that matches a peering interface should definitely be valid, since SCIOND was able to verify the epoch and the revocation token. The fact that SCIOND doesn't remove a path segment because a peering interface got revoked is a policy decision. Here is the mapping as I understand it:

1. c2, d
2. b, (a)
3. (a), c3, e
4. c1

a is either 2 or 3.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

scrye:
Review status: 9 of 16 files reviewed at latest revision, 6 unresolved discussions.

---

*[python/sciond/sciond.py, line 376 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC5DuuYwzR6QXIA7x7:-KtD00xl_-wKRHP_812T:b-e4mdpw) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L376)):*

Why would b map to 4? A revocation is too old if it's not for the current epoch or the previous epoch + grace period. What @kormat meant with current epoch > 2 epochs in the past is that that case might be considered malicious/misconfigured and should be logged as such (same as an epoch from the future). Whether we want to expose this to the app or just let SCIOND log a warning remains to be seen.
A revocation that matches a peering interface should definitely be valid, since SCIOND was able to verify the epoch and the revocation token. The fact that SCIOND doesn't remove a path segment because a peering interface got revoked is a policy decision. Here is the mapping as I understand it:

1. c2, d
2. b, (a)
3. (a), c3, e
4. c1

a is either 2 or 3.
</blockquote></details>

We cannot tell c1, c2 and c3 apart with the current code. To make different decisions we need to change `ConnectedHashTree.verify` to expose the various cases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

scrye:
Review status: 9 of 16 files reviewed at latest revision, 6 unresolved discussions.

---

*[python/sciond/sciond.py, line 376 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC5DuuYwzR6QXIA7x7:-KtD3c0ZjK8rqMa1U2jV:bdax8o6) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L376)):*

We cannot tell c1, c2 and c3 apart with the current code. To make different decisions we need to change `ConnectedHashTree.verify` to expose the various cases.
</blockquote></details>

Follow-up from offline chat: telling c1, c2, and c3 apart is quite difficult with the way peering links are treated right now. Those revocations come directly from the path server, and are checked on path construction.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

kormat:
Reviewed 7 of 7 files at r5.

---

*[python/lib/crypto/hash_tree.py, line 266 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC4OhnAW_HTevL71NW:-KtGIjC1CeXB10eN-Ugc:b-jlqtbq) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/lib/crypto/hash_tree.py#L266)):*

Done.
</blockquote></details>

Upon further consideration, `-2` is too generous, `-1` is sufficient.

---

*[python/sciond/sciond.py, line 376 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC5DuuYwzR6QXIA7x7:-KtGIKrDCw6lrX82ERO3:br95ifv) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L376)):*

Follow-up from offline chat: telling c1, c2, and c3 apart is quite difficult with the way peering links are treated right now. Those revocations come directly from the path server, and are checked on path construction.
</blockquote></details>

Conclusion: current logic will do for now, just need to add a large FIXME to the `UNKNOWN` return value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

scrye:
Review status: 14 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[python/lib/crypto/hash_tree.py, line 266 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC4OhnAW_HTevL71NW:-KtGhBc0pRR4LGo_p8qi:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/lib/crypto/hash_tree.py#L266)):*

Upon further consideration, `-2` is too generous, `-1` is sufficient.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 376 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtC5DuuYwzR6QXIA7x7:-KtGka1_ljYlnUlu4ZVC:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/51b2d0662600bc090f823b66ecd860d77044454c/python/sciond/sciond.py#L376)):*

Conclusion: current logic will do for now, just need to add a large FIXME to the `UNKNOWN` return value.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

kormat:
Reviewed 2 of 2 files at r6.

---

*[python/sciond/sciond.py, line 358 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtGwrKhvsPyNEE1NCw_:-KtGwrKhvsPyNEE1NCwa:bcnr0jo) ([raw file](https://github.com/netsec-ethz/scion/blob/e69033aafaafe3b9d7ee34d862882dfe712c1c30/python/sciond/sciond.py#L358)):*
> ```Python
>                 "Failed to verify epoch: epoch in the past %d,current epoch %d."
>                 % (rev_info.p.epoch, ConnectedHashTree.get_current_epoch()))
>             return SCIONDRevReplyStatus.TOO_OLD
> ```

This should be an error, and `INVALID`.

---

*[python/sciond/sciond.py, line 371 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtGxdz2YN8Pi4sjtcfo:-KtGxdz361UPWQDmmbsI:b-56h4wc) ([raw file](https://github.com/netsec-ethz/scion/blob/e69033aafaafe3b9d7ee34d862882dfe712c1c30/python/sciond/sciond.py#L371)):*
> ```Python
>                 % (rev_info.p.epoch, ConnectedHashTree.get_current_epoch()))
>             # For reasoning behind UNKNOWN return value, see FIXME below
>             return SCIONDRevReplyStatus.UNKNOWN
> ```

This should be `STALE`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

scrye:
Review status: 12 of 16 files reviewed at latest revision, 3 unresolved discussions.

---

*[python/sciond/sciond.py, line 358 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtGwrKhvsPyNEE1NCw_:-KtH32DxlvzIx-cInQyj:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e69033aafaafe3b9d7ee34d862882dfe712c1c30/python/sciond/sciond.py#L358)):*

This should be an error, and `INVALID`.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 371 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtGxdz2YN8Pi4sjtcfo:-KtH34QgnKZIo3CcRijL:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/e69033aafaafe3b9d7ee34d862882dfe712c1c30/python/sciond/sciond.py#L371)):*

This should be `STALE`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

kormat:
Reviewed 4 of 4 files at r7.

---

*[python/sciond/sciond.py, line 370 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtH5UZtqa41FRSyY2Qq:-KtH5UZtqa41FRSyY2Qr:b-b3jbv4) ([raw file](https://github.com/netsec-ethz/scion/blob/1f2ff6b05481b4e0ef888e5edb7b38150e1bad07/python/sciond/sciond.py#L370)):*
> ```Python
>                 "Failed to verify epoch: epoch in the near past %d, current epoch %d."
>                 % (rev_info.p.epoch, ConnectedHashTree.get_current_epoch()))
>             # For reasoning behind UNKNOWN return value, see FIXME below
> ```

Obsolete.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

scrye:
Review status: 15 of 17 files reviewed at latest revision, 2 unresolved discussions.

---

*[python/sciond/sciond.py, line 370 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtH5UZtqa41FRSyY2Qq:-KtH7Qz3tw4Jj9JJ1SGF:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/1f2ff6b05481b4e0ef888e5edb7b38150e1bad07/python/sciond/sciond.py#L370)):*

Obsolete.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

shitz:
Reviewed 2 of 2 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

kormat:
Reviewed 1 of 2 files at r8, 1 of 1 files at r9.

---

*[python/lib/util.py, line 270 at r9](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtHJbDsPbIlFWR0Aa__:-KtHJbDsPbIlFWR0Aa_a:bor4eet) ([raw file](https://github.com/netsec-ethz/scion/blob/9c4a30662f1f3181b8382cd8c04add754976bd45/python/lib/util.py#L270)):*
> ```Python
>         except ConnectionResetError:
>             logging.error("ConnectionResetError: sock = %s, total_len = %s"
>                           " read_len = %s, data = %s" % (sock, total_len, len(barr), barr))
> ```

Use `hex_str(barr)` (from `lib.util`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

scrye:

---

*[python/lib/util.py, line 270 at r9](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-KtHJbDsPbIlFWR0Aa__:-KtHKI_zr_9lFz3qeAsl:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9c4a30662f1f3181b8382cd8c04add754976bd45/python/lib/util.py#L270)):*

Use `hex_str(barr)` (from `lib.util`).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1239#-:-KtHKjvYXEA14XnduLgE:bnfp4nl)*

PR #1240, opened 2017-09-01
PathDB for Go

shitz:
This PR adds a `PathDB` akin to the Python version to the Go codebase. It currently uses SQLite as backend storage. The major changes to the Python versions are:

* Being able to query for all path segments that contain a specific interface
* Being able to store UP, DOWN, and CORE segments in one database
* Being able to store hidden path configuration identifiers along side the path segments
* Being able to read and store a database from a file.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1240)


kormat:
Reviewed 3 of 9 files at r1.
Review status: 3 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/seg/seg.go, line 69 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCLuBMrmrBD3i7yG2f:-KtCLuBMrmrBD3i7yG2g:bxkdamr) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L69)):*
> ```Go
> }
> 
> func (ps *PathSegment) Pack() (common.RawBytes, *common.Error) {
> ```

So, my question here is if we need this to be a method on the class itself? (The same goes for `Write()`, too).

---

*[go/lib/ctrl/seg/seg.go, line 107 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCM89gdvKjOD-say7X:-KtCM89gdvKjOD-say7Y:b-mqzu6w) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L107)):*
> ```Go
> func (t Type) String() string {
> 	switch t {
> 	case 0:
> ```

These constants should be defined.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

kormat:
Reviewed 4 of 9 files at r1.
Review status: 7 of 9 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/seg/seg.go, line 51 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtC_22mT96Kc_QPykie:-KtC_22nbIZaPZVy8xDy:b-lq44m4) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L51)):*
> ```Go
> 		common.Order.PutUint32(data, as.RawIA)
> 		common.Order.PutUint64(data, as.HopEntries[0].InIF)
> 		common.Order.PutUint64(data, as.HopEntries[0].OutIF)
> ```

Uff. I've just realised that this is wrong (as is the python version). This is relying on the interface IDs reported by _neighbours_, not the relevant ASes themselves. We should fix this.

---

*[go/lib/pathdb/pathdb.go, line 42 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCPnWM1ihy6g-Hguol:-KtCPnWM1ihy6g-Hguom:b-juuc9u) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/pathdb.go#L42)):*
> ```Go
> 		db.conn, cerr = sqlite.New(path)
> 	default:
> 		return nil, common.NewError(fmt.Sprintf("Unknown backend: '%s'", backend))
> ```

`common.NewError("Unsupported backend", "backend", backend)`

---

*[go/lib/pathdb/pathdb.go, line 58 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCQ201iVzwa9FZZAIc:-KtCQ201iVzwa9FZZAId:bxvt0hc) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/pathdb.go#L58)):*
> ```Go
> // InsertWithCfgIDs inserts or updates a path segment with a set of HPCfgIDs. It
> // returns the number of path segments that have been inserted/updated.
> func (db *DB) InsertWithCfgIDs(pseg *seg.PathSegment,
> ```

`Cfg` is too generic.

---

*[go/lib/pathdb/sqlite/schema.go, line 36 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCWlsuZBh95BSD3qZg:-KtCWlsuZBh95BSD3qZh:b2n9jio) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/schema.go#L36)):*
> ```Go
> 		IntfID INTEGER NOT NULL,
> 		SegID INTEGER NOT NULL,
> 		PRIMARY KEY (IsdID, AsID, IntfID, SegID) ON CONFLICT IGNORE,
> ```

Why ignore?

---

*[go/lib/pathdb/sqlite/schema.go, line 56 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCXasQKYRbZdLQTiP-:-KtCXasQKYRbZdLQTiP0:bsgvnnt) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/schema.go#L56)):*
> ```Go
> 		FOREIGN KEY (SegID) REFERENCES Segments(ID) ON DELETE CASCADE
> 	);
> 	CREATE TABLE HPCfgIds(
> ```

Where is the actual config stored?

---

*[go/lib/pathdb/sqlite/sqlite.go, line 26 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCXiEbAqJ5byutNTsC:-KtCXiEbAqJ5byutNTsD:bt4c5px) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L26)):*
> ```Go
> 	"time"
> 
> 	"github.com/netsec-ethz/scion/go/lib/addr"
> ```

Import grouping.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 52 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCXyRjhGYDM9JjlR_J:-KtCXyRjhGYDM9JjlR_K:b-7vo595) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L52)):*
> if the schema version of the existing database is smaller than the schema version in schema.go, a new database is be created. 

That.. instant data-loss?

---

*[go/lib/pathdb/sqlite/sqlite.go, line 67 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCY9Wvce207pm9nJYS:-KtCY9Wvce207pm9nJYT:b-wjp2hw) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L67)):*
> ```Go
> 	}
> 	if version > SchemaVersion {
> 		return nil, common.NewError("The database schema version is newer than supported")
> ```

Include the relevant versions.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 84 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCYUVUm4Y7OhcOk8j_:-KtCYUVUm4Y7OhcOk8ja:b-7em8c1) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L84)):*
> ```Go
> 	var err error
> 	b.db, err = sql.Open("sqlite3", uri)
> 	if err != nil {
> ```

The advantage of pre-declaring `err` is that you can do the call and error check on a single line. `if b.db, err = sql.Open(...); err != nil {`

---

*[go/lib/pathdb/sqlite/sqlite.go, line 91 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCYoix5Z_x35pIvUxr:-KtCYoix5Z_x35pIvUxs:b-15xcam) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L91)):*
> ```Go
> 	err = b.db.QueryRow("PRAGMA foreign_keys;").Scan(&enabled)
> 	if err != nil {
> 		return common.NewError("Foreign keys not supported", "err", err)
> ```

Is that the only possible reason that QueryRow might return an error?

---

*[go/lib/pathdb/sqlite/sqlite.go, line 201 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCaR7oUCfDSyLepmvO:-KtCaR7oUCfDSyLepmvP:b1f5h5q) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L201)):*
> ```Go
> 	}
> 	// Update segment.
> 	if cerr := updateSeg(tx, meta); err != nil {
> ```

`cerr`!

---

*[go/lib/pathdb/sqlite/sqlite.go, line 202 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCalzh0-JUPPOzM7Pd:-KtCalzh0-JUPPOzM7Pe:b-sdv7f0) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L202)):*
> ```Go
> 	// Update segment.
> 	if cerr := updateSeg(tx, meta); err != nil {
> 		tx.Rollback()
> ```

Note to self: reviewed to here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

kormat:
Reviewed 1 of 9 files at r1.
Review status: 8 of 9 files reviewed at latest revision, 29 unresolved discussions, some commit checks failed.

---

*[go/lib/pathdb/sqlite/schema.go, line 27 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGc-b35ABR-ts6Fced:-KtGc-b35ABR-ts6Fcee:bh4xkyr) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/schema.go#L27)):*
> ```Go
> 	Schema = `CREATE TABLE Segments(
> 		ID INTEGER PRIMARY KEY AUTOINCREMENT,
> 		SegID DATA UNIQUE NOT NULL,
> ```

These two field names are.. problematic. E.g. a lot of the code refers to just `rowID`, which maps to `Segments.ID`, but the sql schema refers to that as `SegID`. These really need to be consistent, and much easier to distinguish.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGO5tkcMgRZf63yx7E:-KtGO5tkcMgRZf63yx7F:b-9yqs1b) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L207)):*
> ```Go
> 	// Check if the existing segment is registered as the given type(s).
> 	for _, segType := range segTypes {
> 		ok, cerr := b.checkType(meta.RowID, segType)
> ```

Would it maybe be cheaper to just do an insert/update?

---

*[go/lib/pathdb/sqlite/sqlite.go, line 241 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGOrtxyStmHtAl16b_:-KtGOrtxyStmHtAl16ba:b-lvzex9) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L241)):*
> ```Go
> }
> 
> func updateSeg(tx *sql.Tx, meta *segMeta) *common.Error {
> ```

Functions should come after methods. That said, there's probably no reason they couldn't become methods, either.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 306 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGPMtcfsHig3jVpRge:-KtGPMtcfsHig3jVpRgf:b-8mgc40) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L306)):*
> ```Go
> }
> 
> func insertType(tx *sql.Tx, rowID int64, segType seg.Type) *common.Error {
> ```

`rowID` is a bit too generic. `segRowID`?

---

*[go/lib/pathdb/sqlite/sqlite.go, line 309 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGcd9VBzfIxZ901P4M:-KtGcd9VBzfIxZ901P4N:b-gfy4cc) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L309)):*
> Failed

---

*[go/lib/pathdb/sqlite/sqlite.go, line 354 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGdZU8T6k86KMe9PlM:-KtGdZU8T6k86KMe9PlN:b-7yef51) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L354)):*
> ```Go
> 	}
> 	// Insert ISD-AS to StartsAt.
> 	cerr = insertStartOrEnd(tx, pseg.ASEntries[0], rowID, "StartsAt")
> ```

These table names should be constants.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 395 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGcfhke_CXtbvT7NyC:-KtGcfhke_CXtbvT7NyD:bwokgkc) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L395)):*
> ```Go
> 		stmt, err := tx.Prepare(stmtStr)
> 		if err != nil {
> 			return common.NewError("Failed to prepare statement", "err", err)
> ```

This is too generic.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 399 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGcpYp8h9SpYG9kVld:-KtGcpYp8h9SpYG9kVle:b51dfeh) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L399)):*
> ```Go
> 		defer stmt.Close()
> 		for _, hop := range as.HopEntries {
> 			if hop.InIF != 0 {
> ```

This is the wrong interface. Look at the descriptions in `proto/pcb.capnp`.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 408 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGd3ZTnKdamKNmQAex:-KtGd3ZTnKdamKNmQAey:byypiq5) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L408)):*
> ```Go
> 				_, err = stmt.Exec(ia.I, ia.A, hop.OutIF, rowID)
> 				if err != nil {
> 					return common.NewError("Failed to insert into IntfToSeg", "err", err)
> ```

This message is a duplicate of the previous one, and doesn't contain useful information for debugging.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 438 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGeFoMwzrr5-0bsNv3:-KtGeFoMwzrr5-0bsNv4:b1f5h5q) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L438)):*
> ```Go
> 	}
> 	res, cerr := prepareAndExec(tx, "DELETE FROM Segments WHERE SegID=?", segID)
> 	if err != nil {
> ```

`cerr`!

---

*[go/lib/pathdb/sqlite/sqlite.go, line 464 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGe_8vPsmmokBNr-tg:-KtGe_8vPsmmokBNr-th:b-tqoqzh) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L464)):*
> ```Go
> 		SELECT * FROM IntfToSeg WHERE IsdID=? AND AsID=? AND IntfID=?)`
> 	res, cerr := prepareAndExec(tx, delStmt, intf.IA.I, intf.IA.A, intf.IfID)
> 	if err != nil {
> ```

`cerr`

---

*[go/lib/pathdb/sqlite/sqlite.go, line 536 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGfgGGSaJ2apGBKWv3:-KtGfgGHZD6dqpDa9rJZ:b-7a4t50) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L536)):*
> ```Go
> 	where := []string{}
> 	if len(params.SegID) > 0 {
> 		where = append(where, fmt.Sprintf("s.SegID=x'%v'", params.SegID))
> ```

`%s`, as you're deliberately calling `.String()` on `RawBytes`.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 549 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGgYUU8j8FVBYyc8Lz:-KtGgYUU8j8FVBYyc8M-:b-12ug1b) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L549)):*
> ```Go
> 		subQ := []string{}
> 		for _, cfgID := range params.CfgIDs {
> 			subQ = append(subQ, fmt.Sprintf("(h.IsdID='%v' AND h.AsID='%v' AND h.CfgID='%v')",
> ```

These should be %d's.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 558 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGgULUPtZ0Pnog6vO-:-KtGgULUPtZ0Pnog6vO0:b-12ug1b) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L558)):*
> ```Go
> 		subQ := []string{}
> 		for _, spec := range params.Intfs {
> 			subQ = append(subQ, fmt.Sprintf("(i.IsdID='%v' AND i.AsID='%v' AND i.IntfID='%v')",
> ```

These should be %d's.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 568 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGgnplAxXsSdWUdTTs:-KtGgnplAxXsSdWUdTTt:b-12ug1b) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L568)):*
> ```Go
> 		for _, as := range params.StartsAt {
> 			subQ = append(subQ, fmt.Sprintf("(st.IsdID='%v' AND st.AsID='%v')", as.I, as.A))
> 		}
> ```

These should be %d's.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 584 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGh0VDh84hN36-_NvP:-KtGh0VDh84hN36-_NvQ:bqzuthm) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L584)):*
> ```Go
> 	}
> 	if len(where) > 0 {
> 		query = append(query, fmt.Sprintf("WHERE %s", strings.Join(where, " AND\n")))
> ```

Hmm. There's no grouping of the various where clauses.

Also the `\n` seems unnecessary, doesn't the line below already handle that?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

shitz:
Review status: 3 of 10 files reviewed at latest revision, 29 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/seg/seg.go, line 51 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtC_22mT96Kc_QPykie:-KtH44znJ6w_6SM5cKIB:b75p4b1) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L51)):*

Uff. I've just realised that this is wrong (as is the python version). This is relying on the interface IDs reported by _neighbours_, not the relevant ASes themselves. We should fix this.
</blockquote></details>

Ack. Added TODO, since this PR is already large enough.

---

*[go/lib/ctrl/seg/seg.go, line 69 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCLuBMrmrBD3i7yG2f:-KtH4PjcM1C-Vzq4Nhb_:b6vg0zz) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L69)):*

So, my question here is if we need this to be a method on the class itself? (The same goes for `Write()`, too).
</blockquote></details>

Hmm there are here for convenience, but not necessarily needed. I think they do add something to the readability of code, but ultimately I leave it up to you whether to keep them or not (they do add some boilerplate of course)

---

*[go/lib/ctrl/seg/seg.go, line 107 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCM89gdvKjOD-say7X:-KtH7-kn0xcy4ec6KMx7:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L107)):*

These constants should be defined.
</blockquote></details>

Done.

---

*[go/lib/pathdb/pathdb.go, line 42 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCPnWM1ihy6g-Hguol:-KtH78C20G3c_uMbcanN:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/pathdb.go#L42)):*

`common.NewError("Unsupported backend", "backend", backend)`
</blockquote></details>

Done.

---

*[go/lib/pathdb/pathdb.go, line 58 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCQ201iVzwa9FZZAIc:-KtH7MBjXEO3izhyacS3:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/pathdb.go#L58)):*

`Cfg` is too generic.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/schema.go, line 27 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGc-b35ABR-ts6Fced:-KtHA2KjuCrkwVgKwfhL:b4tu9ub) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/schema.go#L27)):*

These two field names are.. problematic. E.g. a lot of the code refers to just `rowID`, which maps to `Segments.ID`, but the sql schema refers to that as `SegID`. These really need to be consistent, and much easier to distinguish.
</blockquote></details>

Changed `ID` to `RowID` and `SegID` in meta tables to `SegRowID`

---

*[go/lib/pathdb/sqlite/schema.go, line 36 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCWlsuZBh95BSD3qZg:-KtHA7rQcJbgDnpJEihz:b-okxm8h) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/schema.go#L36)):*

Why ignore?
</blockquote></details>

`<Isd, As, Intf>` triples can appear multiple times in case an `ASEntry` contains peering interfaces. This made sure that only one of those occurrences was in the DB. I now handle this in the code and removed the `ON CONFLICT IGNORE` statement.

---

*[go/lib/pathdb/sqlite/schema.go, line 56 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCXasQKYRbZdLQTiP-:-KtHJMJvg8Ny8nnjp6jJ:b-8odzdc) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/schema.go#L56)):*

Where is the actual config stored?
</blockquote></details>

Do we really want to store it in the pathdb? That one I would store as a normal file I guess. It is also not applicable for a wide range of `PathDB` clients.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 26 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCXiEbAqJ5byutNTsC:-KtHJaxg5G9jbDpqetHh:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L26)):*

Import grouping.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 52 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCXyRjhGYDM9JjlR_J:-KtHKSZCMxvWwq-RHh7d:b-pasv4k) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L52)):*

> if the schema version of the existing database is smaller than the schema version in schema.go, a new database is be created. 

That.. instant data-loss?
</blockquote></details>

Changed to throw an error in both cases.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 67 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCY9Wvce207pm9nJYS:-KtHKY56GP5yCgBFnUgw:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L67)):*

Include the relevant versions.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 84 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCYUVUm4Y7OhcOk8j_:-KtHKjQ43zmbJhJg06QK:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L84)):*

The advantage of pre-declaring `err` is that you can do the call and error check on a single line. `if b.db, err = sql.Open(...); err != nil {`
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 91 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCYoix5Z_x35pIvUxr:-KtHLTS8_qetVH6umJt2:bhuu6lo) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L91)):*

Is that the only possible reason that QueryRow might return an error?
</blockquote></details>

Clarified

---

*[go/lib/pathdb/sqlite/sqlite.go, line 201 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCaR7oUCfDSyLepmvO:-KtHLdWDKvBiBdXfIBhs:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L201)):*

`cerr`!
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 207 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGO5tkcMgRZf63yx7E:-KtHO7t0xvKtNaLKjYP5:b-e15kvr) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L207)):*

Would it maybe be cheaper to just do an insert/update?
</blockquote></details>

Hmm instead of an extra query I defined `(SegRowID, SegType)` as `UNIQUE` and then add `ON CONFLICT IGNORE` (same for the HPCfgIDs). I'm not sure if that's faster, but it reduces complexity of the code.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 241 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGOrtxyStmHtAl16b_:-KtHLnpemfr23fBg-Qob:b7upeac) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L241)):*

Functions should come after methods. That said, there's probably no reason they couldn't become methods, either.
</blockquote></details>

Hmm there isn't really a reason for them to be methods, since they do not operate on the `Backend` object. The grouping in this file is that you have the main method first and then all calls in order of appearance (in the code).

---

*[go/lib/pathdb/sqlite/sqlite.go, line 306 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGPMtcfsHig3jVpRge:-KtHPTkuEYQx94NqpBAE:bdwh20r) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L306)):*

`rowID` is a bit too generic. `segRowID`?
</blockquote></details>

Replaced all `rowID` with `segRowID`

---

*[go/lib/pathdb/sqlite/sqlite.go, line 309 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGcd9VBzfIxZ901P4M:-KtHPZ4x4dvUgXqmr9iq:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L309)):*

> Failed
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 354 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGdZU8T6k86KMe9PlM:-KtHQxozgf-uH6dsE18F:b-v0ac6f) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L354)):*

These table names should be constants.
</blockquote></details>

That would mean (for consistency) that I need to create string constants for all tables and lots of `fmt.Sprintf` which I doubt would be very useful.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 395 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGcfhke_CXtbvT7NyC:-KtHPsfDy3GdH7ZUOVM7:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L395)):*

This is too generic.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 399 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGcpYp8h9SpYG9kVld:-KtHPu9AKHBh39wysDA3:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L399)):*

This is the wrong interface. Look at the descriptions in `proto/pcb.capnp`.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 408 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGd3ZTnKdamKNmQAex:-KtHQ0U1Q9QqLy_oRRap:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L408)):*

This message is a duplicate of the previous one, and doesn't contain useful information for debugging.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 438 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGeFoMwzrr5-0bsNv3:-KtHQ5Bf5kWvrtvKUehJ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L438)):*

`cerr`!
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 464 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGe_8vPsmmokBNr-tg:-KtHQ8bW8i4drInYV_Dp:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L464)):*

`cerr`
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 536 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGfgGGSaJ2apGBKWv3:-KtHQJ04tqXdKdni-YU5:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L536)):*

`%s`, as you're deliberately calling `.String()` on `RawBytes`.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 549 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGgYUU8j8FVBYyc8Lz:-KtHQSA_3v63Nlkjm_84:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L549)):*

These should be %d's.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 558 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGgULUPtZ0Pnog6vO-:-KtHQSa8b9yyFyjXxEtH:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L558)):*

These should be %d's.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 568 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGgnplAxXsSdWUdTTs:-KtHQT0elVd7c1WWiDSa:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L568)):*

These should be %d's.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 584 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGh0VDh84hN36-_NvP:-KtHQZzWWiwMw19C6V-B:b-1s18u1) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L584)):*

Hmm. There's no grouping of the various where clauses.

Also the `\n` seems unnecessary, doesn't the line below already handle that?
</blockquote></details>

What do you mean by grouping? Each entry in `where` is a condition in `()`. Here we just join all the where clauses with an `AND`. The newline is only there to have a more readable string, but all the where clauses could be on one line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

kormat:
Reviewed 6 of 7 files at r2.
Review status: 9 of 10 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/seg/seg.go, line 51 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtC_22mT96Kc_QPykie:-KtHTnN4dvus9zNKMBeB:b-du4vqy) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L51)):*

Ack. Added TODO, since this PR is already large enough.
</blockquote></details>

I've filed https://github.com/netsec-ethz/scion/issues/1244 for this, maybe include that in the FIXME comment.

---

*[go/lib/ctrl/seg/seg.go, line 69 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCLuBMrmrBD3i7yG2f:-KtHTsTxxAyRZE0YSd0w:bdf2npa) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L69)):*

Hmm there are here for convenience, but not necessarily needed. I think they do add something to the readability of code, but ultimately I leave it up to you whether to keep them or not (they do add some boilerplate of course)
</blockquote></details>

Yeah, ok. Readability is nice, as is insulating the caller from the actual underlying mechanism. We should add a `Pack()` method to `Cerealizable` (#NotInThisPR)

---

*[go/lib/ctrl/seg/seg.go, line 107 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCM89gdvKjOD-say7X:-KtHVxR6a8RYgxESG6rP:b-fsvg6q) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L107)):*

Done.
</blockquote></details>

Except you're not using them :P

---

*[go/lib/ctrl/seg/seg.go, line 121 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtHW0T0WGkV3FJm69NN:-KtHW0T0WGkV3FJm69NO:b-1ss6vn) ([raw file](https://github.com/netsec-ethz/scion/blob/764b17c48692159a1eb10685497a6fd6dc0ebfa8/go/lib/ctrl/seg/seg.go#L121)):*
> ```Go
> 		return "CORE"
> 	default:
> 		return "UNKNOWN"
> ```

This isn't a great string. Check out one of the many examples we have for this, e.g. `go/lib/common/l4.go`

---

*[go/lib/pathdb/pathdb.go, line 58 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCQ201iVzwa9FZZAIc:-KtHWa9gV873BxQoFsRl:b-f58tmc) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/pathdb.go#L58)):*

Done.
</blockquote></details>

Not even slightly. :P

---

*[go/lib/pathdb/sqlite/schema.go, line 56 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCXasQKYRbZdLQTiP-:-KtHXZjo7IjhhqieAO1F:bqigiq5) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/schema.go#L56)):*

Do we really want to store it in the pathdb? That one I would store as a normal file I guess. It is also not applicable for a wide range of `PathDB` clients.
</blockquote></details>

Well, in that case you'd have an unused table or two.

From discussion: lets store it externally for now, and revisit this later if it becomes an issue.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 241 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGOrtxyStmHtAl16b_:-KtH_2hcEVK34oTpl6yl:b-a0dpuf) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L241)):*

Hmm there isn't really a reason for them to be methods, since they do not operate on the `Backend` object. The grouping in this file is that you have the main method first and then all calls in order of appearance (in the code).
</blockquote></details>

Pls methodize.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 309 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGcd9VBzfIxZ901P4M:-KtHbHqVUfbpCcGRlOUD:bwqr66m) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L309)):*

Done.
</blockquote></details>

You Faild.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 354 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGdZU8T6k86KMe9PlM:-KtH_eNSpyepkT_0XbK4:blxuh8z) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L354)):*

That would mean (for consistency) that I need to create string constants for all tables and lots of `fmt.Sprintf` which I doubt would be very useful.
</blockquote></details>

It would be good to have it for bare table names, if nothing else.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 584 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGh0VDh84hN36-_NvP:-KtHcWLZgE00l6gKD2iL:bocmkb3) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L584)):*

What do you mean by grouping? Each entry in `where` is a condition in `()`. Here we just join all the where clauses with an `AND`. The newline is only there to have a more readable string, but all the where clauses could be on one line.
</blockquote></details>

Here's an example result:
```
  WHERE s.SegID=x'ae561bfbaa9777242bf22a81f2db3bd1e8b9ae325bb80ef5b55710161a913598' AND
  t.Type='0' AND
  (st.IsdID='1' AND st.AsID='13') OR (st.IsdID='1' AND st.AsID='19') 
```
That will match all paths that start in 1-19, regardless of anything else on that line.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 355 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtHa8fQK6xTpcQX81kV:-KtHa8fQK6xTpcQX81kW:b-dwp7zs) ([raw file](https://github.com/netsec-ethz/scion/blob/764b17c48692159a1eb10685497a6fd6dc0ebfa8/go/lib/pathdb/sqlite/sqlite.go#L355)):*
> ```Go
> 				}
> 			}
> 			// Only insert the Ingress interface for the first hop entry in an AS entry.
> ```

Why? That would mean that peer interfaces aren't findable.

---

*[go/lib/spath/hop.go, line 111 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtHb-qXcmEZjo8t5ppR:-KtHb-qXcmEZjo8t5ppS:b-bhtx9m) ([raw file](https://github.com/netsec-ethz/scion/blob/764b17c48692159a1eb10685497a6fd6dc0ebfa8/go/lib/spath/hop.go#L111)):*
> ```Go
> }
> 
> func (h *HopField) Raw() common.RawBytes {
> ```

I'm not convinced about this. If an app has modified any of the HopField fields without calling `Write()`, then `Raw()` will return stale/wrong data.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

shitz:
Review status: 9 of 10 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/seg/seg.go, line 51 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtC_22mT96Kc_QPykie:-KtL4QA-psCWiBHSMHsq:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L51)):*

I've filed https://github.com/netsec-ethz/scion/issues/1244 for this, maybe include that in the FIXME comment.
</blockquote></details>

Done.

---

*[go/lib/ctrl/seg/seg.go, line 107 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCM89gdvKjOD-say7X:-KtL4ZGQm7ZfxQcEk9Vp:b-5gp5uj) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/ctrl/seg/seg.go#L107)):*

Except you're not using them :P
</blockquote></details>

Yeah well, I tried ;)

---

*[go/lib/ctrl/seg/seg.go, line 121 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtHW0T0WGkV3FJm69NN:-KtL4l7fxHD38q7WT1JS:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/764b17c48692159a1eb10685497a6fd6dc0ebfa8/go/lib/ctrl/seg/seg.go#L121)):*

This isn't a great string. Check out one of the many examples we have for this, e.g. `go/lib/common/l4.go`
</blockquote></details>

Done.

---

*[go/lib/pathdb/pathdb.go, line 58 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtCQ201iVzwa9FZZAIc:-KtL6qf8H9cb0E-DNPES:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/pathdb.go#L58)):*

Not even slightly. :P
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 241 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGOrtxyStmHtAl16b_:-KtLBEbXF5sAickGF1dB:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L241)):*

Pls methodize.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 309 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGcd9VBzfIxZ901P4M:-KtLBHW1gNMQcc8vrPPv:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L309)):*

You Faild.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 354 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGdZU8T6k86KMe9PlM:-KtLC0jrQGGHtD_BUvrc:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L354)):*

It would be good to have it for bare table names, if nothing else.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 584 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtGh0VDh84hN36-_NvP:-KtL0LY_RgI1Nwsa6c74:bhcmzz7) ([raw file](https://github.com/netsec-ethz/scion/blob/fee632e205d851072d113521a4c946bf6836dd0a/go/lib/pathdb/sqlite/sqlite.go#L584)):*

Here's an example result:
```
  WHERE s.SegID=x'ae561bfbaa9777242bf22a81f2db3bd1e8b9ae325bb80ef5b55710161a913598' AND
  t.Type='0' AND
  (st.IsdID='1' AND st.AsID='13') OR (st.IsdID='1' AND st.AsID='19') 
```
That will match all paths that start in 1-19, regardless of anything else on that line.
</blockquote></details>

Ah good catch. Added grouping around all the `OR` clauses

---

*[go/lib/pathdb/sqlite/sqlite.go, line 355 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtHa8fQK6xTpcQX81kV:-KtL-7gecdAnQ1uTNYK8:b-7kmxjf) ([raw file](https://github.com/netsec-ethz/scion/blob/764b17c48692159a1eb10685497a6fd6dc0ebfa8/go/lib/pathdb/sqlite/sqlite.go#L355)):*

Why? That would mean that peer interfaces aren't findable.
</blockquote></details>

Oh for some reason I thought that the `Egress` would contain the peering interface. Looking at the beacon server it seems that the `Ingress` contains the peering interface.

---

*[go/lib/spath/hop.go, line 111 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtHb-qXcmEZjo8t5ppR:-KtL3EXfJmCJXXnnkFRu:b29tf8d) ([raw file](https://github.com/netsec-ethz/scion/blob/764b17c48692159a1eb10685497a6fd6dc0ebfa8/go/lib/spath/hop.go#L111)):*

I'm not convinced about this. If an app has modified any of the HopField fields without calling `Write()`, then `Raw()` will return stale/wrong data.
</blockquote></details>

Prepended a `Write`and renamed `Raw` to `Pack` to make it more consistent with other types.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

kormat:
Reviewed 6 of 7 files at r3.
Review status: 9 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 158 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtLH8sPeBnDcNKtz4on:-KtLH8sPeBnDcNKtz4oo:b-qhsoqj) ([raw file](https://github.com/netsec-ethz/scion/blob/c476e771e827c1624e534b4e04e621cb82fe1352/go/lib/pathdb/sqlite/sqlite.go#L158)):*
> ```Go
> 	}
> 	if err := b.tx.Commit(); err != nil {
> 		return common.NewError("Failed to commit transaction", "err", err)
> ```

Hmm. Should this still set `b.tx` to `nil`?

---

*[go/lib/pathdb/sqlite/sqlite.go, line 465 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtLIFpxvB_k7A8itS9-:-KtLIFpxvB_k7A8itS90:b-9ohm7v) ([raw file](https://github.com/netsec-ethz/scion/blob/c476e771e827c1624e534b4e04e621cb82fe1352/go/lib/pathdb/sqlite/sqlite.go#L465)):*
> ```Go
> 		var cfgIsd int
> 		var cfgAs int
> 		var cfgID uint64
> ```

*cough*

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

shitz:
Review status: 9 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 158 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtLH8sPeBnDcNKtz4on:-KtLKOelT3KyWgbim7Xp:bb77ufg) ([raw file](https://github.com/netsec-ethz/scion/blob/c476e771e827c1624e534b4e04e621cb82fe1352/go/lib/pathdb/sqlite/sqlite.go#L158)):*

Hmm. Should this still set `b.tx` to `nil`?
</blockquote></details>

Yes.

---

*[go/lib/pathdb/sqlite/sqlite.go, line 465 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtLIFpxvB_k7A8itS9-:-KtLKWdI9h1mazKZDkN0:bsubj2k) ([raw file](https://github.com/netsec-ethz/scion/blob/c476e771e827c1624e534b4e04e621cb82fe1352/go/lib/pathdb/sqlite/sqlite.go#L465)):*

*cough*
</blockquote></details>

Fixed as discussed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

kormat:
Reviewed 2 of 3 files at r4.
Review status: 8 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

kormat:
Reviewed 1 of 4 files at r5.
Review status: 6 of 9 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[go/lib/pathdb/sqlite/sqlite_test.go, line 70 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtLQsbJ8TBfYYHS2Cj8:-KtLQsbJ8TBfYYHS2Cj9:b-u7gw5g) ([raw file](https://github.com/netsec-ethz/scion/blob/a8d0c829671f96457e7eb386f2ab548646daed8e/go/lib/pathdb/sqlite/sqlite_test.go#L70)):*
> ```Go
> 	for i := 0; i < len(ifs)/2; i++ {
> 		rawHops[i] = make([]byte, 8)
> 		_ = spath.NewHopField(rawHops[i], common.IFIDType(ifs[2*i]),
> ```

Assignment isn't necessary.

---

*[go/lib/pathdb/sqlite/sqlite_test.go, line 76 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtLkoFzlji9cLCCuMdG:-KtLkoFzlji9cLCCuMdH:b-b0etyy) ([raw file](https://github.com/netsec-ethz/scion/blob/a8d0c829671f96457e7eb386f2ab548646daed8e/go/lib/pathdb/sqlite/sqlite_test.go#L76)):*
> ```Go
> 		{
> 			RawIA: ia13.Uint32(),
> 			HopEntries: []*seg.HopEntry{
> ```

It looks like it's worth creating a helper function to create a single `*seg.HopEntry`.

---

*[go/lib/pathdb/sqlite/sqlite_test.go, line 269 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtMCdLrAvqlmD_uYRbi:-KtMCdLrAvqlmD_uYRbj:b-7m8wcz) ([raw file](https://github.com/netsec-ethz/scion/blob/f7d650fe8a2ffe28a419d34b60aa944004975e3e/go/lib/pathdb/sqlite/sqlite_test.go#L269)):*
> ```Go
> 		// Check Insert.
> 		checkInsert(t, b, &ExpectedInsert{1, segID, TS, ifspecs, ia13, ia19, types, hpCfgIDs})
> 		os.Remove(tmpF)
> ```

Use a `defer` statement at the top to handle this, otherwise failures don't clean up.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

shitz:
Review status: 6 of 9 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[go/lib/pathdb/sqlite/sqlite_test.go, line 70 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtLQsbJ8TBfYYHS2Cj8:-KtMYDAY9QvgCzBeP3T4:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a8d0c829671f96457e7eb386f2ab548646daed8e/go/lib/pathdb/sqlite/sqlite_test.go#L70)):*

Assignment isn't necessary.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite_test.go, line 76 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtLkoFzlji9cLCCuMdG:-KtMZ7VIuVf-Wp6arF4k:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a8d0c829671f96457e7eb386f2ab548646daed8e/go/lib/pathdb/sqlite/sqlite_test.go#L76)):*

It looks like it's worth creating a helper function to create a single `*seg.HopEntry`.
</blockquote></details>

Done.

---

*[go/lib/pathdb/sqlite/sqlite_test.go, line 269 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-KtMCdLrAvqlmD_uYRbi:-KtM_a-L1AwpRT22eI7P:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/f7d650fe8a2ffe28a419d34b60aa944004975e3e/go/lib/pathdb/sqlite/sqlite_test.go#L269)):*

Use a `defer` statement at the top to handle this, otherwise failures don't clean up.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

kormat:
Reviewed 3 of 4 files at r5, 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

kormat:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240#-:-KtMadPP39zEmZh9Knqw:bnfp4nl)*

shitz:
Reviewed 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1240)*

PR #1253, opened 2017-09-20
Enable concurrent read/writes in the end host sock

jpcsmith:
This is an attempt to support concurrent reads & writes through the snet.Conn interface by splitting the associated mutex.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1253)

### DIFF
--- a/go/lib/snet/conn.go
+++ b/go/lib/snet/conn.go
@@ -16,7 +16,6 @@ package snet
 
 import (
 	"net"
-	"sync"
 	"time"
 
 	"github.com/netsec-ethz/scion/go/lib/addr"
@@ -40,8 +39,7 @@ var _ net.Conn = (*Conn)(nil)
 var _ net.PacketConn = (*Conn)(nil)
 
 type Conn struct {
-	dispMutex sync.Mutex
-	conn      *reliable.Conn
+	conn *reliable.Conn
 	// Local and remote SCION addresses (IA, L3, L4)
 	laddr *Addr
 	raddr *Addr
@@ -101,9 +99,7 @@ func (c *Conn) read(b []byte) (int, *Addr, error) {
 	var err error
 	var remote *Addr
 
-	c.dispMutex.Lock()
 	n, err := c.conn.Read(c.recvBuffer)
-	c.dispMutex.Unlock()
 	if err != nil {
 		return 0, nil, common.NewCError("Dispatcher read error", "err", err)
 	}
@@ -225,9 +221,7 @@ func (c *Conn) write(b []byte, raddr *Addr) (int, error) {
 	}
 
 	// Send message
-	c.dispMutex.Lock()
 	n, err = c.conn.WriteTo(c.sendBuffer[:n], appAddr)
-	c.dispMutex.Unlock()
 	if err != nil {
 		return 0, common.NewCError("Dispatcher write error", "err", err)
 	}
--- a/go/lib/sock/reliable/reliable.go
+++ b/go/lib/sock/reliable/reliable.go
@@ -53,7 +53,7 @@
 // To send messages to remote SCION hosts, hosts fill in the common header
 // with the address type, the address and the layer 4 port of the remote host.
 //
-// The connections are not thread safe.
+// Reads and writes to the connection are thread safe.
 //
 package reliable
 
@@ -61,6 +61,7 @@ import (
 	"bytes"
 	"fmt"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/netsec-ethz/scion/go/lib/addr"
@@ -93,6 +94,9 @@ type Conn struct {
 	recvBuf       []byte
 	recvReadHead  int
 	recvWriteHead int
+
+	readMutex  sync.Mutex
+	writeMutex sync.Mutex
 }
 
 func DialTimeout(address string, timeout time.Duration) (*Conn, error) {
@@ -203,6 +207,9 @@ func (conn *Conn) ReadFrom(buf []byte) (int, *AppAddr, error) {
 // copied field of the Msg struct contains the number of bytes in the read
 // packet.
 func (conn *Conn) ReadN(msgs []Msg) (int, error) {
+	conn.readMutex.Lock()
+	defer conn.readMutex.Unlock()
+
 	fillIndex := 0
 	// If we do not have enough data for a full packet, try a blocking read
 	for fillIndex < len(msgs) {
@@ -334,6 +341,9 @@ func (conn *Conn) WriteTo(buf []byte, dst AppAddr) (int, error) {
 // oriented protocols); the copied field of struct Msg is reset to 0 for
 // written packets.  WriteN returns the number of packets written.
 func (conn *Conn) WriteN(bufs []Msg) (int, error) {
+	conn.writeMutex.Lock()
+	defer conn.writeMutex.Unlock()
+
 	copiedMsgs := 0
 	index := 0
 	for i := range bufs {


kormat:
Reviewed 2 of 2 files at r1.

---

*[go/lib/snet/conn.go, line 106 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1253#-Kusx7EvzjlylFiQwBHu:-Kusx7EvzjlylFiQwBHv:b-1rmqn8) ([raw file](https://github.com/netsec-ethz/scion/blob/80c388c7100f42e9879c464c9744eca10501dc4c/go/lib/snet/conn.go#L106)):*
> ```Go
> 	c.dispMutex.Lock()
> 	n, err := c.conn.Read(c.recvBuffer)
> 	c.dispMutex.Unlock()
> ```

Oh, actually, this was _already_ racey. The call above reads into a static buffer belonging to `c`. If two `read()` calls happen at the ~same time, then the second call could easily overwrite the data of the first one. The same problem exists with `c.sendBuffer` below. 

So, at a _minimum_, `Conn` needs to have mutexes to protect the send/recv buffers. A single mutex for both would be ok, but it would be counterproductive for your aim :)

---

*[go/lib/sock/reliable/reliable.go, line 312 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1253#-KusyO-NZrtfuLGyR_ap:-KusyO-NZrtfuLGyR_aq:b-lsjlri) ([raw file](https://github.com/netsec-ethz/scion/blob/1da5047e56d8204a108cf38ae436f179114b1608/go/lib/sock/reliable/reliable.go#L312)):*
> ```Go
> // WriteTo works similarly to Write. In addition to Write, the ReliableSocket message header
> // will contain the address and port information in dst.
> func (conn *Conn) WriteTo(buf []byte, dst AppAddr) (int, error) {
> ```

This also needs to be protected, so that two calls to WriteTo don't cause interleaved writes. This means that the current `WriteN` can't grab the mutex. You'd then end up with:
```
func ... WriteTo(...) {
   conn.writeMutex.Lock()
   defer conn.writeMutex.Unlock()
   ...
       n, err := conn.writeN(msgs)
```

```
func ... WriteN(...) {
   conn.writeMutex.Lock()
   defer conn.writeMutex.Unlock()
   return conn.writeN(...)
}
```

and the current `WriteN` would be renamed to `writeN`. `WriteNAll` would need to similarly change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1253)*

jpcsmith:
Reviewed 1 of 2 files at r1.

---

*[go/lib/snet/conn.go, line 106 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1253#-Kusx7EvzjlylFiQwBHu:-Kutsnms7e2ArxCZ4hux:bqczrkc) ([raw file](https://github.com/netsec-ethz/scion/blob/80c388c7100f42e9879c464c9744eca10501dc4c/go/lib/snet/conn.go#L106)):*

Oh, actually, this was _already_ racey. The call above reads into a static buffer belonging to `c`. If two `read()` calls happen at the ~same time, then the second call could easily overwrite the data of the first one. The same problem exists with `c.sendBuffer` below. 

So, at a _minimum_, `Conn` needs to have mutexes to protect the send/recv buffers. A single mutex for both would be ok, but it would be counterproductive for your aim :)
</blockquote></details>

Okay. I don't see any reason for the buffer to be shared - they're only ever used within the respective functions. Can I just use local buffers instead of locking?

---

*[go/lib/sock/reliable/reliable.go, line 312 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1253#-KusyO-NZrtfuLGyR_ap:-KuttbDrWczaWby_obbh:bx9olo) ([raw file](https://github.com/netsec-ethz/scion/blob/1da5047e56d8204a108cf38ae436f179114b1608/go/lib/sock/reliable/reliable.go#L312)):*

This also needs to be protected, so that two calls to WriteTo don't cause interleaved writes. This means that the current `WriteN` can't grab the mutex. You'd then end up with:
```
func ... WriteTo(...) {
   conn.writeMutex.Lock()
   defer conn.writeMutex.Unlock()
   ...
       n, err := conn.writeN(msgs)
```

```
func ... WriteN(...) {
   conn.writeMutex.Lock()
   defer conn.writeMutex.Unlock()
   return conn.writeN(...)
}
```

and the current `WriteN` would be renamed to `writeN`. `WriteNAll` would need to similarly change.
</blockquote></details>

Okay, I'd missed the "for" loop in the WriteTo function.

Wouldn't WriteNAll be okay without the change? Or should interleaving of messages be disallowed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1253)*

kormat:

---

*[go/lib/snet/conn.go, line 106 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1253#-Kusx7EvzjlylFiQwBHu:-KuyTRzG-fduCFT6IHcq:br14dcc) ([raw file](https://github.com/netsec-ethz/scion/blob/80c388c7100f42e9879c464c9744eca10501dc4c/go/lib/snet/conn.go#L106)):*
<details><summary><i>Previously, jpcsmith (Jean-Pierre Smith) wrote…</i></summary><blockquote>

Okay. I don't see any reason for the buffer to be shared - they're only ever used within the respective functions. Can I just use local buffers instead of locking?
</blockquote></details>

The buffers are 65KB each. Allocating them for each read/write call is far too expensive.

---

*[go/lib/sock/reliable/reliable.go, line 312 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1253#-KusyO-NZrtfuLGyR_ap:-KuyTgisNWj5WYmgzWVG:b-ucr9qb) ([raw file](https://github.com/netsec-ethz/scion/blob/1da5047e56d8204a108cf38ae436f179114b1608/go/lib/sock/reliable/reliable.go#L312)):*
<details><summary><i>Previously, jpcsmith (Jean-Pierre Smith) wrote…</i></summary><blockquote>

Okay, I'd missed the "for" loop in the WriteTo function.

Wouldn't WriteNAll be okay without the change? Or should interleaving of messages be disallowed?
</blockquote></details>

Interleaving should be disallowed, there's just too much room for subtle bugs. Once a single Write/WriteTo/WriteN/WriteNAll call is made, all other such calls must block until the first one finishes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1253)*

jpcsmith:

---

*[go/lib/snet/conn.go, line 106 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1253#-Kusx7EvzjlylFiQwBHu:-KuyUpFjEmRcPJFaVD4Z:b43t11v) ([raw file](https://github.com/netsec-ethz/scion/blob/80c388c7100f42e9879c464c9744eca10501dc4c/go/lib/snet/conn.go#L106)):*

The buffers are 65KB each. Allocating them for each read/write call is far too expensive.
</blockquote></details>

Okay, final question :D... Why is it 65KB?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1253)*

kormat:

---

*[go/lib/snet/conn.go, line 106 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1253#-Kusx7EvzjlylFiQwBHu:-Kv1WEqglyn5_Ekw5puR:b-6438w4) ([raw file](https://github.com/netsec-ethz/scion/blob/80c388c7100f42e9879c464c9744eca10501dc4c/go/lib/snet/conn.go#L106)):*
<details><summary><i>Previously, jpcsmith (Jean-Pierre Smith) wrote…</i></summary><blockquote>

Okay, final question :D... Why is it 65KB?
</blockquote></details>

Because that's the maximum supported size for a packet (common header has a 2-byte total length field, same for udp header, etc).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1253)*

kormat:
Closing, replaced by #1276 (merged).

PR #1267, opened 2017-10-04
Major refactoring of the non-forwarding logic of the SIG.

kormat:
NB: this is still WIP. There's some polishing and cleanup that still needs to be done,

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1267)


shitz:
Reviewed 1 of 20 files at r1.
Review status: 1 of 21 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/sig/base/as.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc0ik4KJpzM5ARUqIR:-Kvc0ik4KJpzM5ARUqIS:b-kn2d3r) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/base/as.go#L15)):*
> ```Go
> 
> package base
> ```

How about having `package as` that contains `Entry` and `Map`. Then `base` would become less a collection of unrelated files. It's also nicer to refer to `as.Map` instead of `base.Map`. The former makes it very clear what kind of map this is.

---

*[go/sig/base/as.go, line 192 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc-0h8OYvftxRGW30R:-Kvc-0h8OYvftxRGW30S:bikxmri) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/base/as.go#L192)):*
> ```Go
> 	defer ae.RUnlock()
> 
> 	output := fmt.Sprintf("ISDAS %v:\n", ae.IA)
> ```

Make `output` a slice of `string` to which you append all the lines and then call `strings.Join(output, "\n")` at the end. `+=` on string is very inefficient.

---

*[go/sig/base/sig.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc4-hU5dbhG3tXCq3K:-Kvc4-hU5dbhG3tXCq3L:b-kskqa0) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/base/sig.go#L15)):*
> ```Go
> 
> package base
> ```

This would also become part of `package as`

---

*[go/sig/sigcmn/common.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc6LT7hzP19zzsZrgQ:-Kvc6LT7hzP19zzsZrgR:beaox55) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/sigcmn/common.go#L15)):*
> ```Go
> 
> package sigcmn
> ```

Not happy with the name. What about just `sig`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

shitz:
Review status: 1 of 21 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/sig/base/pathmgmt.go, line 28 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc6fzYViabetQ_G_pp:-Kvc2ivHyG92AntG7767:bqsyb41) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/base/pathmgmt.go#L28)):*
> ```Go
> type PathPolicy struct {
> 	//Policy  pathmgr.Policy // Describes what interfaces to route through
> 	Pool    *pathmgr.AppPathSet // Pool of paths that meet the policy requirement, managed by pathmgr
> ```

This should be a `SyncPaths`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

kormat:
Review status: 1 of 21 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/sig/base/as.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc0ik4KJpzM5ARUqIR:-Kvc6wOD1CTjQnaeZI39:b1o5u5j) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/base/as.go#L15)):*

How about having `package as` that contains `Entry` and `Map`. Then `base` would become less a collection of unrelated files. It's also nicer to refer to `as.Map` instead of `base.Map`. The former makes it very clear what kind of map this is.
</blockquote></details>

Currently the only unrelated file is `forward.go`, which will get removed when the Egress path gets refactored. Everything else works closely together. The keepalive code will be going into this dir, though, so i don't think `as` is a good name for it.

---

*[go/sig/base/as.go, line 192 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc-0h8OYvftxRGW30R:-Kvc7M7VVARvavSd-Iq5:b-ydd4so) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/base/as.go#L192)):*

Make `output` a slice of `string` to which you append all the lines and then call `strings.Join(output, "\n")` at the end. `+=` on string is very inefficient.
</blockquote></details>

This is old code, and i'm not sure it's needed. I'll just remove it.

---

*[go/sig/base/pathmgmt.go, line 28 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc6fzYViabetQ_G_pp:-Kvc86xvxplIpayNQdtq:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/base/pathmgmt.go#L28)):*

This should be a `SyncPaths`
</blockquote></details>

Done.

---

*[go/sig/sigcmn/common.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc6LT7hzP19zzsZrgQ:-Kvc88vPtuG3tda8E2RT:b-mxurd3) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/sigcmn/common.go#L15)):*

Not happy with the name. What about just `sig`?
</blockquote></details>

Do you want to also rename the dir as well, to `go/sig/sig/common.go`? That's going to be confusing. I based the name on `go/border/rtrcmn`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

shitz:
Review status: 1 of 21 files reviewed at latest revision, 3 unresolved discussions.

---

*[go/sig/base/as.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc0ik4KJpzM5ARUqIR:-Kvc9D86PfioIkFAl9S2:b-3yhfbc) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/base/as.go#L15)):*

Currently the only unrelated file is `forward.go`, which will get removed when the Egress path gets refactored. Everything else works closely together. The keepalive code will be going into this dir, though, so i don't think `as` is a good name for it.
</blockquote></details>

But the keepalive module would work on all those `as.X` datastructures, so I think this should be fine. Besides I'd rather have a `keepalive` package and an `as` package than everything in `base`. `base` is just a very non-verbose name. `sigcmn` could also be `base` for example.

---

*[go/sig/sigcmn/common.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc6LT7hzP19zzsZrgQ:-Kvc8f769m-uQNlTR7n5:b36fvi9) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/sigcmn/common.go#L15)):*

Do you want to also rename the dir as well, to `go/sig/sig/common.go`? That's going to be confusing. I based the name on `go/border/rtrcmn`.
</blockquote></details>

Ah damn, that is indeed confusing. Actually, having lib/common is the problem since I feel like these should all be called `common`. Anyway, changing that wouldn't be worth it, so I guess we could start this convention of `<basepkgname>cmn`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

kormat:
Review status: 1 of 21 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/sig/base/as.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc0ik4KJpzM5ARUqIR:-KvcAq0BSf_QF66dL-Tp:beqauvm) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/base/as.go#L15)):*

But the keepalive module would work on all those `as.X` datastructures, so I think this should be fine. Besides I'd rather have a `keepalive` package and an `as` package than everything in `base`. `base` is just a very non-verbose name. `sigcmn` could also be `base` for example.
</blockquote></details>

I agree `base` is not a good name, but right now i don't have a better suggestion. Let's postpone a decision until we at least have the basic keepalive code in. It's going to integrate into the guts of the current base/*.go structures, and i'd rather not split things out only to have to un-split them later.

---

*[go/sig/sigcmn/common.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc6LT7hzP19zzsZrgQ:-KvcAkfDfUPEllUkwkxx:bw13l1n) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/sigcmn/common.go#L15)):*

Ah damn, that is indeed confusing. Actually, having lib/common is the problem since I feel like these should all be called `common`. Anyway, changing that wouldn't be worth it, so I guess we could start this convention of `<basepkgname>cmn`.
</blockquote></details>

ITYM _continue_ the convention :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

shitz:
:lgtm:

---

Review status: 1 of 21 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/sig/base/as.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc0ik4KJpzM5ARUqIR:-KvcBZLmg2kFmZ7qXOT-:b-dafekm) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/base/as.go#L15)):*

I agree `base` is not a good name, but right now i don't have a better suggestion. Let's postpone a decision until we at least have the basic keepalive code in. It's going to integrate into the guts of the current base/*.go structures, and i'd rather not split things out only to have to un-split them later.
</blockquote></details>

Alright, lets revisit after that code is in, or maybe you see a better option while you are implementing it.

---

*[go/sig/sigcmn/common.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvc6LT7hzP19zzsZrgQ:-KvcBOGK6kock7wSTSaU:b-9axx0p) ([raw file](https://github.com/netsec-ethz/scion/blob/403256e3f26d3fd6d279a3ac3ff84bb7297897ad/go/sig/sigcmn/common.go#L15)):*

ITYM _continue_ the convention :P
</blockquote></details>

Well you cannot really call it a convention if there is only one example so far and no document that specifies this convention :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-:-KvcBmxV8ZQ1HsghWmM6:bnfp4nl)*

shitz:
Reviewed 13 of 20 files at r1, 7 of 7 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

scrye:
Reviewed 13 of 20 files at r1, 7 of 7 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[go/lib/addr/isdas.go, line 79 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvfbo1Pb5pDuffD4RUQ:-Kvfbo1QGP_9_-_OYiLK:b-50n8ui) ([raw file](https://github.com/netsec-ethz/scion/blob/b26763547a759bd5ad84e1ae287d370900378d9b/go/lib/addr/isdas.go#L79)):*
> ```Go
> }
> 
> func (ia *ISD_AS) IAInt() IAInt {
> ```

This should include a FIXME to replace method `Uint32` with `IAInt`in the future. Having two methods doing the same thing is bloat.

---

*[go/sig/sig.go, line 108 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvfcg4xfFzL6ATB6czP:-Kvfcg4xfFzL6ATB6czQ:b-mec69l) ([raw file](https://github.com/netsec-ethz/scion/blob/b26763547a759bd5ad84e1ae287d370900378d9b/go/sig/sig.go#L108)):*
> ```Go
> }
> 
> func loadConfig(path string) bool {
> ```

We could probably replace this with walking the structure output by the `config` package (https://github.com/scrye/scion/tree/sig-framer/go/sig/config) , which already handles full config file parsing (including classes and path filters) and is unit tested with high coverage. This also separates JSON error handling from main program logic.

---

*[go/sig/base/as.go, line 99 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-KvfePbc78TNeeUQ6ns_:-KvfePbc78TNeeUQ6nsa:bpw5x9s) ([raw file](https://github.com/netsec-ethz/scion/blob/b26763547a759bd5ad84e1ae287d370900378d9b/go/sig/base/as.go#L99)):*
> ```Go
> 	// and the egress dispatcher would be spawned here (and if we move away from tun-per-IA,
> 	// then it would be spawned by main())
> 	go NewEgressWorker(ae, ae.tunIO).Run()
> ```

`ASEntry` already includes `tunIO`. Passing `ae` should be enough.

(On a sidenote, `EgressDispatcher` will probably take `ae` as well).

---

*[go/sig/base/pathmgmt.go, line 26 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvffg22R5-VPDMzyLSI:-Kvffg22R5-VPDMzyLSJ:bo5csqb) ([raw file](https://github.com/netsec-ethz/scion/blob/b26763547a759bd5ad84e1ae287d370900378d9b/go/sig/base/pathmgmt.go#L26)):*
> ```Go
> // a pool of paths that match the specified policy, metrics about those paths,
> // as well as maintaining the currently favoured path and remote SIG to use.
> type PathPolicy struct {
> ```

This also needs to somehow include the RingBuffer connecting `EgressDispatcher` to `EgressWorker`. The parent goroutine needs to create this, as the child goroutines have no knowledge of each other.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

kormat:

---

*[go/lib/addr/isdas.go, line 79 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvfbo1Pb5pDuffD4RUQ:-Kvfjyo00WlHmTUKonPw:b-y3ns1i) ([raw file](https://github.com/netsec-ethz/scion/blob/b26763547a759bd5ad84e1ae287d370900378d9b/go/lib/addr/isdas.go#L79)):*

This should include a FIXME to replace method `Uint32` with `IAInt`in the future. Having two methods doing the same thing is bloat.
</blockquote></details>

Well, we could do that, but it means that the current calls to `Uint32` would become `uint32(....IAInt())`, which would silently break if we ever change the size of ISD_AS. They are all calls which specifically care about the type specifically being 32bits, whereas calls to IAInt() don't care.

---

*[go/sig/sig.go, line 108 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvfcg4xfFzL6ATB6czP:-Kvfl5T7NUePQJKdPTgY:b3usgzs) ([raw file](https://github.com/netsec-ethz/scion/blob/b26763547a759bd5ad84e1ae287d370900378d9b/go/sig/sig.go#L108)):*

We could probably replace this with walking the structure output by the `config` package (https://github.com/scrye/scion/tree/sig-framer/go/sig/config) , which already handles full config file parsing (including classes and path filters) and is unit tested with high coverage. This also separates JSON error handling from main program logic.
</blockquote></details>

I did initially start with that code, but it's not actually that useful. We want to use the same mechanism for updating the data structures regardless of whether it's from JSON on startup, JSON on reload, or dynamically from discovery. That's why the API for ASMap/ASEntry/SIGEntry etc are all designed to add/remove single items at a time, so it can be done iteratively, and not in a single large replace.

Thinking about it, the problem is the error handling is in the wrong place - it should be inside `ASInfo.AddSIG` etc. That is, _parse_ errors should be caught by `config` (in the JSON decoder where possible), and _logic_ errors should be caught by the `base` API.

I'll rework things a bit along those lines, and see how it looks.

---

*[go/sig/base/as.go, line 99 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-KvfePbc78TNeeUQ6ns_:-Kvfpiq1su2SwNm5nPrV:b-8ykd6v) ([raw file](https://github.com/netsec-ethz/scion/blob/b26763547a759bd5ad84e1ae287d370900378d9b/go/sig/base/as.go#L99)):*

`ASEntry` already includes `tunIO`. Passing `ae` should be enough.

(On a sidenote, `EgressDispatcher` will probably take `ae` as well).
</blockquote></details>

I _could_ change it to simply just pass `ae`, and have the worker call `ae.TunIO()`, but given that `setupNet` is possibly being _called_ by `TunIO`, i'm not sure it's worth it. The `tunIO` attribute is deliberately not exported, and won't be accessible to the egress worker/dispatcher once they move to their own module (like `ingress` has).

---

*[go/sig/base/pathmgmt.go, line 26 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvffg22R5-VPDMzyLSI:-KvfqbPavIHM23nI3RFh:bz1l50s) ([raw file](https://github.com/netsec-ethz/scion/blob/b26763547a759bd5ad84e1ae287d370900378d9b/go/sig/base/pathmgmt.go#L26)):*

This also needs to somehow include the RingBuffer connecting `EgressDispatcher` to `EgressWorker`. The parent goroutine needs to create this, as the child goroutines have no knowledge of each other.
</blockquote></details>

Yes, agreed. This is currently not used, yet, this is just a rough sketch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

kormat:
Review status: 16 of 31 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/lib/addr/isdas.go, line 79 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-Kvfbo1Pb5pDuffD4RUQ:-KvgPvhn5TRgjks5G0gg:bljcqgm) ([raw file](https://github.com/netsec-ethz/scion/blob/b26763547a759bd5ad84e1ae287d370900378d9b/go/lib/addr/isdas.go#L79)):*

Well, we could do that, but it means that the current calls to `Uint32` would become `uint32(....IAInt())`, which would silently break if we ever change the size of ISD_AS. They are all calls which specifically care about the type specifically being 32bits, whereas calls to IAInt() don't care.
</blockquote></details>

Decided to just go with it :) scion/1269 switches everything fully to `IAInt`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

scrye:
Went in detail over the config stuff, looks good.

The pattern seems to be quite clean now, `cfg` parses, `loadConfig` walks the result and fills in the blanks and validation happens when calling functions that affect the state of the SIG's internal tables.

As an aside, I feel like code should always (1) catch errors as soon as possible if it has full knowledge that what it's looking at constitutes an error (so parsing errors in json should be caught in json, and broken references within the same file, if we consider them to be an error condition, should also be caught there, (2) a function should always expect the utmost garbage when receiving input. I feel like the current version of the code respects these mantras quite well. It's clear where the parsing ends and loading begins, and the SIG's core API validates everything that goes in.

---

Review status: 16 of 21 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/sig/base/as.go, line 99 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-KvfePbc78TNeeUQ6ns_:-KvhAfq_JS1TXgyPxq1X:b-qhngjv) ([raw file](https://github.com/netsec-ethz/scion/blob/b26763547a759bd5ad84e1ae287d370900378d9b/go/sig/base/as.go#L99)):*

I _could_ change it to simply just pass `ae`, and have the worker call `ae.TunIO()`, but given that `setupNet` is possibly being _called_ by `TunIO`, i'm not sure it's worth it. The `tunIO` attribute is deliberately not exported, and won't be accessible to the egress worker/dispatcher once they move to their own module (like `ingress` has).
</blockquote></details>

Well, if it's going to get scoped out there's no way around having it as a separate parameter, so I guess it's fine.

---

*[go/sig/base/as.go, line 150 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-KvhBUN13z2UKf4SukEJ:-KvhBUN13z2UKf4SukEK:bbcpl8d) ([raw file](https://github.com/netsec-ethz/scion/blob/325331af4d1ac30fb8c6957ada30fca89b7117ec/go/sig/base/as.go#L150)):*
> ```Go
> 	}
> 	if ip == nil {
> 		return common.NewCError("AddSig: zero-length SIG address", "ia", ae.IA)
> ```

"zero-length SIG address" might look a little misleading when seen in logs.  Maybe explicitly say "nil SIG address"?

---

*[go/sig/base/as.go, line 153 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-KvhBvtHTBWcCvo3FbSq:-KvhBvtHTBWcCvo3FbSr:b-902xom) ([raw file](https://github.com/netsec-ethz/scion/blob/325331af4d1ac30fb8c6957ada30fca89b7117ec/go/sig/base/as.go#L153)):*
> ```Go
> 	}
> 	if ctrlPort == 0 {
> 		return common.NewCError("AddSig: Ctrl port must be non-zero", "ia", ae.IA, "id", id)
> ```

Since this function does most of the validation, maybe also checking if the port numbers are over max? I know the json parser verifies that it's uint16, but this function should not make any assumptions that data it receives is not garbage.

---

*[go/sig/base/as.go, line 156 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-KvhC7Ddvxcq9eF-KVHd:-KvhC7Ddvxcq9eF-KVHe:b2nreh6) ([raw file](https://github.com/netsec-ethz/scion/blob/325331af4d1ac30fb8c6957ada30fca89b7117ec/go/sig/base/as.go#L156)):*
> ```Go
> 	}
> 	if encapPort == 0 {
> 		return common.NewCError("AddSig: Encap port must be non-zero", "ia", ae.IA, "id", id)
> ```

Same as above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-:-KvhJn3d5-kJjZFMsTOD:b-zbhs0f)*

kormat:
Review status: 15 of 22 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/sig/base/as.go, line 150 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-KvhBUN13z2UKf4SukEJ:-KvkfYOT-THiUar2DZ2B:b6o0qdj) ([raw file](https://github.com/netsec-ethz/scion/blob/325331af4d1ac30fb8c6957ada30fca89b7117ec/go/sig/base/as.go#L150)):*

"zero-length SIG address" might look a little misleading when seen in logs.  Maybe explicitly say "nil SIG address"?
</blockquote></details>

Yeah, i wasn't very happy with the wording either. I've reworded by the id and address errors.

---

*[go/sig/base/as.go, line 153 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-KvhBvtHTBWcCvo3FbSq:-KvkfckCqyK3skrf6Ufv:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/325331af4d1ac30fb8c6957ada30fca89b7117ec/go/sig/base/as.go#L153)):*

Since this function does most of the validation, maybe also checking if the port numbers are over max? I know the json parser verifies that it's uint16, but this function should not make any assumptions that data it receives is not garbage.
</blockquote></details>

Done.

---

*[go/sig/base/as.go, line 156 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-KvhC7Ddvxcq9eF-KVHd:-KvkfcytdPC3QPHU4pZi:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/325331af4d1ac30fb8c6957ada30fca89b7117ec/go/sig/base/as.go#L156)):*

Same as above.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

scrye:
Reviewed 1 of 20 files at r1, 3 of 5 files at r3, 4 of 4 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267)*

scrye:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1267#-:-Kvkzm1gsYx1Zu3U93nz:bnfp4nl)*

PR #1271, opened 2017-10-05
Certificates in Go

Oncilla:
Add go implementation for:
- Certificate
- CertificateChain
- TrustStore (TRC part missing)
- verify and sign based on ed25519

Additionally, two libraries are added:
golang.org/x/crypto/ed25519
github.com/pierrec/lz4

TRC implementation will follow

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1271)


shitz:
Review status: 0 of 7 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[go/lib/crypto/asymcrypto.go, line 1 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwiczy8n6LZvVKNQai5:-Kwiczy8n6LZvVKNQai6:b3rpf4n) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/asymcrypto.go#L1)):*
> ```Go
> ```

Filename can be `asym.go`

---

*[go/lib/crypto/asymcrypto.go, line 29 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwicmLxjJRkfUh_44K4:-KwicmLxjJRkfUh_44K5:b-wqrq7x) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/asymcrypto.go#L29)):*
> ```Go
> func Sign(sigInput, signKey common.RawBytes, signAlgo string) (common.RawBytes, error) {
> 	switch strings.ToLower(signAlgo) {
> 	case "ed25519":
> ```

Define constants for the algorithm names and use those.

---

*[go/lib/crypto/cert/cert.go, line 1 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwifewM49ktz8pv64Mw:-KwifewM49ktz8pv64Mx:b3r134v) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L1)):*
> ```Go
> ```

This file needs unit tests.

---

*[go/lib/crypto/cert/cert.go, line 31 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwiduGmoe4aLJ8_v3wy:-KwiduGmoe4aLJ8_v3wz:b-99h13) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L31)):*
> ```Go
> 
> type Certificate struct {
> 	// The following fields need to be sorted
> ```

which fields and sorted for what?

---

*[go/lib/crypto/cert/cert.go, line 71 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwiem-YBHxfzaKlOSRb:-Kwiem-YBHxfzaKlOSRc:b9ptbh9) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L71)):*
> ```Go
> // Verify checks the signature of the certificate based on a trusted verifying key,
> // and further verifies, that the certificate belongs to the given subject.
> func (c *Certificate) Verify(subject *addr.ISD_AS, verifyKey common.RawBytes, signAlgo string) error {
> ```

Why does this take `signAlgo` and doesn't just use `c.SignAlgorithm`?

---

*[go/lib/crypto/cert/cert.go, line 104 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwifWkSF0P8e-AJmcFy:-KwifWkSF0P8e-AJmcFz:bb7kry6) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L104)):*
> ```Go
> 	c.Signature = sig
> 	return nil
> ```

Remove empty line

---

*[go/lib/crypto/cert/cert.go, line 108 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwiezJ25IgckQQfxwp1:-KwiezJ25IgckQQfxwp2:b-wglmtf) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L108)):*
> ```Go
> 
> // signatureInput creates a sorted json object of all fields, except for the signature field.
> func (c *Certificate) signatureInput() (common.RawBytes, error) {
> ```

We call this `sigPack` in other parts of the code.

---

*[go/lib/crypto/cert/chain.go, line 1 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwijqKgDgQXpijPxPuu:-KwijqKgDgQXpijPxPuv:b-6r3ua9) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L1)):*
> ```Go
> ```

This file needs unit tests

---

*[go/lib/crypto/cert/chain.go, line 42 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwij4cNRsh4RwEzUM0a:-Kwij4cNRsh4RwEzUM0b:b85r6w6) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L42)):*
> ```Go
> }
> 
> type Chain struct {
> ```

docstring

---

*[go/lib/crypto/cert/chain.go, line 49 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwiiVggWuds-Xt_jq5f:-KwiiVggWuds-Xt_jq5g:b-gpxbso) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L49)):*
> ```Go
> func ChainFromRaw(raw common.RawBytes, lz4_ bool) (*Chain, error) {
> 	if lz4_ {
> 		byteLen := binary.LittleEndian.Uint32(raw[:4])
> ```

this should be `common.Order.Uint32(raw[:4])`

---

*[go/lib/crypto/cert/chain.go, line 68 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwijL_2_6hUp01NnyKy:-KwijL_2_6hUp01NnyKz:bjy0crh) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L68)):*
> ```Go
> }
> 
> func (c *Chain) Verify(subject *addr.ISD_AS, trc int) error {
> ```

Use `interface{}` as a type for `trc` for now.

---

*[go/lib/crypto/cert/chain.go, line 84 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwijWRcnG8VaJGYcPfa:-KwijWRcnG8VaJGYcPfb:btkkoca) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L84)):*
> ```Go
> 	}
> 	comp := make([]byte, lz4.CompressBlockBound(len(raw))+4)
> 	binary.LittleEndian.PutUint32(comp[:4], uint32(len(raw)))
> ```

`common.Order`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

shitz:
Review status: 0 of 7 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[go/lib/trust/store.go, line 45 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwjfu4vXIe72uL082an:-Kwik6pCpcfkGCVaD4PI:b-ec7y9l) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L45)):*
> ```Go
> }
> 
> func NewTrustStore(confDir, cacheDir, eName string) (*Store, error) {
> ```

rename to `NewStore`

---

*[go/lib/trust/store.go, line 46 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwjfv34d94t3jaCJD8x:-Kwjfv34d94t3jaCJD8y:b-ccqfjm) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L46)):*
> ```Go
> 
> func NewTrustStore(confDir, cacheDir, eName string) (*Store, error) {
> 	dir := filepath.Join(confDir, CertDir)
> ```

`NewStore` should take the certs directory directly as parameter, otherwise you are hardcoding the filesystem layout into a library.

---

*[go/lib/trust/store.go, line 62 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwjgMbIW23E61qt19_O:-KwjgMbIW23E61qt19_P:b-eigpo8) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L62)):*
> ```Go
> 	}
> 	cachedFiles, err := filepath.Glob(fmt.Sprintf("%s/%s*.crt", s.cacheDir, s.eName))
> 	if err != nil {
> ```

This will return an error if there are no cached certificates, but that will be the case on a fresh installation.

---

*[go/lib/trust/store.go, line 90 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwjh2OBblM4UQa-Erht:-Kwjh2OBblM4UQa-Erhu:b4v1em4) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L90)):*
> ```Go
> 	s.chainMap[*chain.Key()] = chain
> 	v, ok := s.maxChainMap[*ia]
> 	if ver > v || !ok {
> ```

Change the order of evaluation `if !ok || ver > v`

---

*[go/lib/trust/store.go, line 96 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwjgsafn1ND5Nx3ps98:-Kwjgsafn1ND5Nx3ps99:b-xm1zgp) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L96)):*
> ```Go
> 	s.chainLock.Unlock()
> 	if write && !ok {
> 		j, err := json.MarshalIndent(chain, "", "    ")
> ```

use `strings.Repeat`

---

*[go/lib/trust/store.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwjhT9581HM9wRnvsdQ:-KwjhT9581HM9wRnvsdR:b-xunapr) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L110)):*
> ```Go
> // GetChain returns the certificate chain for the specified values or nil, if it is not present.
> func (s *Store) GetChain(ia *addr.ISD_AS, ver int) *cert.Chain {
> 	s.chainLock.RLock()
> ```

Use the 
```
s.chainLock.RLock()
defer s.chainLock.RUnlock()
```
pattern. Same below

---

*[go/lib/trust/store.go, line 117 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwjiDHS850CQkh_LLr3:-KwjiDHS850CQkh_LLr4:b-ae57uv) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L117)):*
> ```Go
> 
> // GetMaxChain the certificate chain with the highest version for the specified ISD-AS.
> func (s *Store) GetMaxChain(ia *addr.ISD_AS) *cert.Chain {
> ```

Would `GetNewestChain` be a better name or is higher version number not necessarily = newer?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

Oncilla:
Review status: 0 of 10 files reviewed at latest revision, 19 unresolved discussions.

---

*[go/lib/crypto/cert/cert.go, line 1 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwifewM49ktz8pv64Mw:-KwkKozD9swBLq6exz-6:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L1)):*

This file needs unit tests.
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/cert.go, line 31 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwiduGmoe4aLJ8_v3wy:-Kwivo6jWS_Os_NQkmUM:b-7dtemf) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L31)):*

which fields and sorted for what?
</blockquote></details>

Golang marshals a JSON with fields in order of struct declaration. I could either overwrite the MarshalJSON function, or just sort the fields.
(I prefer latter)

---

*[go/lib/crypto/cert/cert.go, line 71 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwiem-YBHxfzaKlOSRb:-KwiwqRWHaTun9wnPAG3:b99cr5b) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L71)):*

Why does this take `signAlgo` and doesn't just use `c.SignAlgorithm`?
</blockquote></details>

Because c.SignAlgorithm is the algorithm, which is used to create signatures verifiable with c. But signAlgo is the algorithm used to sign c.

---

*[go/lib/crypto/cert/cert.go, line 104 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwifWkSF0P8e-AJmcFy:-KwixU5r0KDFbIbrtckj:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L104)):*

Remove empty line
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/cert.go, line 108 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwiezJ25IgckQQfxwp1:-KwixbPEPyRCxcVsxwES:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L108)):*

We call this `sigPack` in other parts of the code.
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/chain.go, line 1 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwijqKgDgQXpijPxPuu:-Kwou0bXX05cx3oMuUxN:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L1)):*

This file needs unit tests
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/chain.go, line 42 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwij4cNRsh4RwEzUM0a:-Kwj-LC38TGVSBFz0DHu:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L42)):*

docstring
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/chain.go, line 49 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwiiVggWuds-Xt_jq5f:-Kwj-Y8M0c0KzkqzuVSZ:bvus1wc) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L49)):*

this should be `common.Order.Uint32(raw[:4])`
</blockquote></details>

Our python library uses little endian for this integer.

---

*[go/lib/crypto/cert/chain.go, line 68 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwijL_2_6hUp01NnyKy:-Kwj-PyxAtYL9CNBW-AD:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L68)):*

Use `interface{}` as a type for `trc` for now.
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/chain.go, line 84 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwijWRcnG8VaJGYcPfa:-Kwj-YmZgeGgfe5-RhQ8:b-ya81ng) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L84)):*

`common.Order`
</blockquote></details>

Same as above

---

*[go/lib/trust/store.go, line 45 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwjfu4vXIe72uL082an:-Kwk6rlk1gurNgpcoURn:bs7wumb) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L45)):*

rename to `NewStore`
</blockquote></details>

Done.

---

*[go/lib/trust/store.go, line 46 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwjfv34d94t3jaCJD8x:-Kwk8-aIbROmEUQlhjBJ:buegu00) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L46)):*

`NewStore` should take the certs directory directly as parameter, otherwise you are hardcoding the filesystem layout into a library.
</blockquote></details>

(remnant from python code)
Done.

---

*[go/lib/trust/store.go, line 62 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwjgMbIW23E61qt19_O:-Kwk9ZszH-TXF-g8V-HE:b-80dpc4) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L62)):*

This will return an error if there are no cached certificates, but that will be the case on a fresh installation.
</blockquote></details>

How so?
https://golang.org/pkg/path/filepath/#Glob

---

*[go/lib/trust/store.go, line 90 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwjh2OBblM4UQa-Erht:-Kwk9iPb8JgnEWhJDJYj:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L90)):*

Change the order of evaluation `if !ok || ver > v`
</blockquote></details>

Done.

---

*[go/lib/trust/store.go, line 96 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwjgsafn1ND5Nx3ps98:-KwkA2Ave8SuHOsB1fTe:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L96)):*

use `strings.Repeat`
</blockquote></details>

Done.

---

*[go/lib/trust/store.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwjhT9581HM9wRnvsdQ:-KwkAnBcx7lenf5KARVM:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L110)):*

Use the 
```
s.chainLock.RLock()
defer s.chainLock.RUnlock()
```
pattern. Same below
</blockquote></details>

Done.

---

*[go/lib/trust/store.go, line 117 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwjiDHS850CQkh_LLr3:-KwkCMRMjcTaTjLY1HkM:b-hr11y6) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L117)):*

Would `GetNewestChain` be a better name or is higher version number not necessarily = newer?
</blockquote></details>

Hmm. According to the book:

"Version: version of the certificate. It has to be a unique (per subject) nonnegative integer"

However, in the current implementation it is always assumed that the newest certificate also has the max version.
I think in a later stage, we should improve TrustStore to return the newest active Certificate.

---

*[go/lib/crypto/asymcrypto.go, line 1 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwiczy8n6LZvVKNQai5:-KwivHs-Yzd-ORA5gy3i:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/asymcrypto.go#L1)):*

Filename can be `asym.go`
</blockquote></details>

Done.

---

*[go/lib/crypto/asymcrypto.go, line 29 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwicmLxjJRkfUh_44K4:-KwivIo00uHTrz58DTly:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/asymcrypto.go#L29)):*

Define constants for the algorithm names and use those.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

shitz:
Review status: 0 of 10 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[go/lib/crypto/cert/cert.go, line 31 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwiduGmoe4aLJ8_v3wy:-KwsqPVX1yaOE6HtcVm6:b-uvufg7) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L31)):*

Golang marshals a JSON with fields in order of struct declaration. I could either overwrite the MarshalJSON function, or just sort the fields.
(I prefer latter)
</blockquote></details>

Add that the field names need to be sorted alphabetically because MarshalJSON marshals the fields in order of declaration.

---

*[go/lib/crypto/cert/cert.go, line 71 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwiem-YBHxfzaKlOSRb:-Kwsr4hBkOjNDhJ2Xkas:b-qn8nf) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L71)):*

Because c.SignAlgorithm is the algorithm, which is used to create signatures verifiable with c. But signAlgo is the algorithm used to sign c.
</blockquote></details>

Then the docstring for `c.SignAlgorithm` is wrong and needs to be adjusted.

---

*[go/lib/crypto/cert/cert_test.go, line 71 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwsursPJMRG-Y03kAfw:-KwsursPJMRG-Y03kAfx:bpxrlrc) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/crypto/cert/cert_test.go#L71)):*
> ```Go
> 		SoMsg("CanIssue", cert.CanIssue, ShouldEqual, false)
> 		SoMsg("Comment", cert.Comment, ShouldEqual, "AS Certificate☂☂☂☂")
> 		SoMsg("EncAlgo", cert.EncAlgorithm, ShouldEqual, "curve25519xsalsa20poly1305")
> ```

This probably should also use a constant

---

*[go/lib/crypto/cert/cert_test.go, line 74 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwsunjE2rY2rOPx7H08:-KwsunjFYYSahLIA813f:b-ik2oxc) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/crypto/cert/cert_test.go#L74)):*
> ```Go
> 		SoMsg("ExpTime", cert.ExpirationTime, ShouldEqual, 1539868933)
> 		SoMsg("IssueTime", cert.IssuingTime, ShouldEqual, 1508332933)
> 		SoMsg("SignAlgo", cert.SignAlgorithm, ShouldEqual, "ed25519")
> ```

use `crypto.Ed25519`

---

*[go/lib/crypto/cert/cert_test.go, line 101 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwswWgYrdeQHsG0sONz:-KwswWgYrdeQHsG0sOO-:bqys1i6) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/crypto/cert/cert_test.go#L101)):*
> ```Go
> 
> 		valid := time.Date(2017, 10, 19, 0, 0, 0, 0, time.UTC)
> 		monkey.Patch(time.Now, func() time.Time { return valid })
> ```

We don't want to use this module. Without patching `time.Now` you probably have to parametrize the creation of certificates, such that you can pass in timestamps.

---

*[go/lib/crypto/cert/cert_test.go, line 165 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt3SRMyqXSRu0eXCnT:-Kwt3SRMyqXSRu0eXCnU:b-hx6lok) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/crypto/cert/cert_test.go#L165)):*
> ```Go
> 	Convey("Certificate is returned as String correctly", t, func() {
> 		cert, err := CertificateFromRaw(rawCert)
> 		s := `{"CanIssue":false,"Comment":"AS Certificate☂☂☂☂","EncAlgorithm":"curve25519xsalsa20poly1305","ExpirationTime":1539868933,"Issuer":"1-13","IssuingTime":1508332933,"SignAlgorithm":"ed25519","Signature":"/hoJBGTQ0F2+4OqpfCTrPgZjAEX7/3XuqTLbPhmZpsVhX4E+gLHKVG0/+/ASyq6PZjF97WtzApPjVw5jOIEtAg==","Subject":"1-10","SubjectEncKey":"nP1HkZwkW8ujqeEO82Rb9cN6AVqFPO1UIiypdZU+dHI=","SubjectSigKey":"5YYo/Djor8KoUPbcG89m0sOXbhaxU/wserVf7X4w0W4=","TRCVersion":2,"Version":1}`
> ```

Hmm this makes me think. Shouldn't `String()` return a more readable representation?

---

*[go/lib/crypto/cert/chain.go, line 42 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwij4cNRsh4RwEzUM0a:-Kwt4zc9ftL4y9v7PRO0:b-gmgjbe) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L42)):*

Done.
</blockquote></details>

The struct still needs a docstring.

---

*[go/lib/crypto/cert/chain.go, line 49 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwiiVggWuds-Xt_jq5f:-Kwt4tENsGE7wm8IIDep:b-z96uae) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L49)):*

Our python library uses little endian for this integer.
</blockquote></details>

Nevermind then.

---

*[go/lib/crypto/cert/chain_test.go, line 163 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt6A1NeV0GX5IkTe4h:-Kwt6A1NeV0GX5IkTe4i:bnj7hl0) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/crypto/cert/chain_test.go#L163)):*
> ```Go
> 		// FIXME(roosd): Update with TRC implementation
> 		valid := time.Date(2017, 10, 19, 0, 0, 0, 0, time.UTC)
> 		monkey.Patch(time.Now, func() time.Time { return valid })
> ```

same as above

---

*[go/lib/trust/store.go, line 62 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwjgMbIW23E61qt19_O:-Kwt6SBDZrnvu5fDz738:b-rwa1tu) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/trust/store.go#L62)):*

How so?
https://golang.org/pkg/path/filepath/#Glob
</blockquote></details>

Nevermind, sorry.

---

*[go/lib/trust/store.go, line 29 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt6HjbaPqOqTzO0MOZ:-Kwt6HjbaPqOqTzO0MO_:bkypl70) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/trust/store.go#L29)):*
> ```Go
> )
> 
> type Store struct {
> ```

docstring for `Store`

---

*[go/lib/trust/store.go, line 32 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt6N0afS3KEirCVTjh:-Kwt6N0afS3KEirCVTji:bxf0472) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/trust/store.go#L32)):*
> ```Go
> 	// certDir is the certificate directory.
> 	certDir string
> 	// certDir is the directory to cache TRCs and certs in.
> ```

certDir -> cacheDir

---

*[python/lib/crypto/certificate.py, line 200 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt7fl-dEtb3P0QqWNI:-Kwt7fl-dEtb3P0QqWNJ:bj6drqf) ([raw file](https://github.com/netsec-ethz/scion/blob/66e9e607bfc8081d7b3148b1283ac6c5fce74983/python/lib/crypto/certificate.py#L200)):*
> ```Python
>     def _sig_input(self):
>         d = self.dict(False)
>         for k in d:
> ```

Why is this change needed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

Oncilla:
Review status: 0 of 10 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[go/lib/crypto/cert/cert.go, line 31 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwiduGmoe4aLJ8_v3wy:-Kwt_otl1-4qJO2Ef9te:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L31)):*

Add that the field names need to be sorted alphabetically because MarshalJSON marshals the fields in order of declaration.
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/cert.go, line 71 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwiem-YBHxfzaKlOSRb:-Kwt_pGEOjRHAYxD4Va1:b8h99dl) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/cert.go#L71)):*

Then the docstring for `c.SignAlgorithm` is wrong and needs to be adjusted.
</blockquote></details>

Right.
Done.

---

*[go/lib/crypto/cert/cert_test.go, line 71 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwsursPJMRG-Y03kAfw:-KwteEuAKCWGBGWSP96i:bdx921w) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/crypto/cert/cert_test.go#L71)):*

This probably should also use a constant
</blockquote></details>

I did not find a better name.
Maybe C25519XS20P1305, but that is really hard to parse. 

---

*[go/lib/crypto/cert/cert_test.go, line 74 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwsunjE2rY2rOPx7H08:-KwteSsXXG_h3DtzahWZ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/crypto/cert/cert_test.go#L74)):*

use `crypto.Ed25519`
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/cert_test.go, line 101 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KwswWgYrdeQHsG0sONz:-KwtA2YxgRUBzs50-r3g:b-xf73o) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/crypto/cert/cert_test.go#L101)):*

We don't want to use this module. Without patching `time.Now` you probably have to parametrize the creation of certificates, such that you can pass in timestamps.
</blockquote></details>

Hope you like my solution ;)
Done.

---

*[go/lib/crypto/cert/cert_test.go, line 165 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt3SRMyqXSRu0eXCnT:-KwttZcYLT5Kv8ZvUIkt:b-m9xnsc) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/crypto/cert/cert_test.go#L165)):*

Hmm this makes me think. Shouldn't `String()` return a more readable representation?
</blockquote></details>

I refactored it to String() and JSON(), where String() prints basic identifier and JSON() the complete certificate.
Same for Chain.

---

*[go/lib/crypto/cert/chain.go, line 42 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwij4cNRsh4RwEzUM0a:-KwtgtdmB3y4XW_XRF0r:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/64ab68c474fe3ca70a4c6591306cf7270ceb0529/go/lib/crypto/cert/chain.go#L42)):*

The struct still needs a docstring.
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/chain_test.go, line 163 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt6A1NeV0GX5IkTe4h:-Kwtmk1J5grO3da0jkHt:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/crypto/cert/chain_test.go#L163)):*

same as above
</blockquote></details>

Done.

---

*[go/lib/trust/store.go, line 29 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt6HjbaPqOqTzO0MOZ:-KwttQ28GdnFwKIzuesg:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/trust/store.go#L29)):*

docstring for `Store`
</blockquote></details>

Done.

---

*[go/lib/trust/store.go, line 32 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt6N0afS3KEirCVTjh:-KwttUon4O3-4kH40bC_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/1fd43d43b4aa966f23544470a7ad755f0b4ceb34/go/lib/trust/store.go#L32)):*

certDir -> cacheDir
</blockquote></details>

Done.

---

*[python/lib/crypto/certificate.py, line 200 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt7fl-dEtb3P0QqWNI:-Kwttoe3Echp0vWqK9vf:bapnaag) ([raw file](https://github.com/netsec-ethz/scion/blob/66e9e607bfc8081d7b3148b1283ac6c5fce74983/python/lib/crypto/certificate.py#L200)):*

Why is this change needed?
</blockquote></details>

It is not necessary. However, i do not see any value in encoding utf-8 strings in base64.
Base64 takes up approx 30% more space. Also, we cannot expresse more information by doing so.
So, what's the point?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

kormat:
Reviewed 1 of 8 files at r1, 2 of 10 files at r2, 6 of 8 files at r3.
Review status: 8 of 10 files reviewed at latest revision, 21 unresolved discussions.

---

*[go/lib/addr/isdas.go, line 85 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDSaYtlXOjwemDMDi2:-KxDSaYtlXOjwemDMDi3:b-ff29pl) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/addr/isdas.go#L85)):*
> ```Go
> }
> 
> func (ia *ISD_AS) MarshalText() ([]byte, error) {
> ```

Move before `UnmarshalText`.

---

*[go/lib/crypto/asym.go, line 26 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDVumdd9kKprc5LPkG:-KxDVumdd9kKprc5LPkH:b-srpvpc) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/asym.go#L26)):*
> ```Go
> 
> const (
> 	Ed25519                    string = "ed25519"
> ```

The type isn't needed, it's derived from the value.

---

*[go/lib/crypto/cert/chain.go, line 43 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDYX6356G_lgN0N9TN:-KxDYX6356G_lgN0N9TO:b-9bza3v) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain.go#L43)):*
> ```Go
> }
> 
> // Chain contains two certificates, one fore the leave and one for the core. The leave certificate
> ```

"one for the leaf".. "The leaf certificate"

---

*[go/lib/crypto/cert/chain.go, line 46 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDYdu6s8xLPKDyIxNQ:-KxDYdu6s8xLPKDyIxNR:b1fhdk0) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain.go#L46)):*
> ```Go
> // is signed by the core certificate, which is signed by the TRC of the corresponding ISD.
> type Chain struct {
> 	// Leave is the leave certificate of the chain. It is signed by the Core certificate.
> ```

Leaf, leaf, leaf, leaf, leaf. (leaf)

---

*[go/lib/crypto/cert/chain.go, line 82 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDZJqz5WLSXFI4P6H8:-KxDZJqz5WLSXFI4P6H9:b-xzjwel) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain.go#L82)):*
> ```Go
> 
> // Compress compresses the JSON generated from the certificate chain using lz4 and
> // prepends the original length. (4 bytes, little endian, unsigned)
> ```

Wait, why? Why does it need a length? Why little-endian? The python code doesn't do this, afaik.

Compression can simply be (rough sketch):
```
buf := bytes.NewBuffer()
lzW := lz4.NewWriter(buf)
lzW.Write(raw)
return buf.Bytes()
```
Decompression would be:
```
var in bytes.Reader(raw)
var buf &bytes.Buffer
lzR := lz4.NewReader(in)
lzR.WriteTo(buf)
// buf.Bytes() is now the uncompressed raw chain
```

This also means that `MaxChainByteLength` can be removed.

---

*[go/lib/trust/store.go, line 29 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDVDDl4vahxo1CIEMy:-KxDVDDl4vahxo1CIEMz:bjwfov0) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/trust/store.go#L29)):*
> ```Go
> )
> 
> // Store handles storage and management of trust objects (certificate chains and TRCs)
> ```

This mentions TRCs, but the code below doesn't appear to deal with them.

---

*[go/lib/trust/store.go, line 94 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDVNEnYIrwrWCMheuG:-KxDVNEnYIrwrWCMheuH:b-tyswrk) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/trust/store.go#L94)):*
> ```Go
> 	}
> 	s.chainLock.Unlock()
> 	if write && !ok {
> ```

So this will only store a cert if it's a new maximum? What's the rationale?

---

*[go/lib/trust/store.go, line 95 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDaID3g8roJ7svmbtF:-KxDaID3g8roJ7svmbtG:b-li73y) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/trust/store.go#L95)):*
> ```Go
> 	s.chainLock.Unlock()
> 	if write && !ok {
> 		j, err := json.MarshalIndent(chain, "", strings.Repeat(" ", 4))
> ```

`chain.JSON(true)`?

---

*[go/lib/trust/store.go, line 111 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDVk_GFUoav7Y2PLZ5:-KxDVk_GFUoav7Y2PLZ6:bltz335) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/trust/store.go#L111)):*
> ```Go
> 	s.chainLock.RLock()
> 	defer s.chainLock.RUnlock()
> 	chain := s.chainMap[cert.Key{IA: *ia, Ver: ver}]
> ```

You can just return this directly, without `chain`.

---

*[python/lib/crypto/certificate.py, line 200 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-Kwt7fl-dEtb3P0QqWNI:-KxDTt5TuQ2rlkjjk8a5:b-32w0iv) ([raw file](https://github.com/netsec-ethz/scion/blob/66e9e607bfc8081d7b3148b1283ac6c5fce74983/python/lib/crypto/certificate.py#L200)):*

It is not necessary. However, i do not see any value in encoding utf-8 strings in base64.
Base64 takes up approx 30% more space. Also, we cannot expresse more information by doing so.
So, what's the point?
</blockquote></details>

Sounds reasonable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

kormat:
Reviewed 1 of 8 files at r3.
Review status: 9 of 10 files reviewed at latest revision, 24 unresolved discussions.

---

*[go/lib/crypto/cert/cert_test.go, line 79 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDbRKNP6LqruTi8cm3:-KxDbRKNP6LqruTi8cm4:bmwiiy0) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/cert_test.go#L79)):*
> ```Go
> 		SoMsg("Issuer", cert.Issuer.String(), ShouldEqual, "1-13")
> 		SoMsg("Subject", cert.Subject.String(), ShouldEqual, "1-10")
> 		SoMsg("Signature", bytes.Compare(cert.Signature, rawSignature), ShouldEqual, 0)
> ```

You can use `ShouldResemble` for this. It will (attempt to) give readable output of the differences.

---

*[go/lib/crypto/cert/cert_test.go, line 85 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDc7_3XcemfQE0IE81:-KxDc7_3XcemfQE0IE82:bf0rbh2) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/cert_test.go#L85)):*
> ```Go
> 
> 	Convey("CertificateFromRaw should throw error for invalid bytes", t, func() {
> 		rawCert[0] ^= 0xFF
> ```

This is not safe. If a panic happens, this won't be reverted, and all other tests will fail, too. You need to make a copy and modify that.
```
tmpCert = append([]byte(nil), rawCert...)
tmpCert[0] ^= 0xFF
```

---

*[go/lib/crypto/cert/cert_test.go, line 103 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDcsFGgRxrY0rlkUwu:-KxDcsFGgRxrY0rlkUwv:b-9c9ace) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/cert_test.go#L103)):*
> ```Go
> 		cert.Sign(privRaw, crypto.Ed25519)
> 
> 		Convey("Certificate is verifyable", func() {
> ```

"verifiable"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

kormat:
Reviewed 1 of 8 files at r3.
Review status: all files reviewed at latest revision, 26 unresolved discussions.

---

*[go/lib/crypto/cert/chain_test.go, line 228 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDhcxpDjEVYZdx9i9J:-KxDhcxpDjEVYZdx9i9K:bb6u05g) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain_test.go#L228)):*
> ```Go
> 		})
> 		Convey("Chains are unequal (Core)", func() {
> 			c1.Leave.CanIssue = true
> ```

Core

---

*[go/lib/crypto/cert/chain_test.go, line 239 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDhYQ5RHGKnm0p5fHT:-KxDhYQ5RHGKnm0p5fHU:b718zb3) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain_test.go#L239)):*
> ```Go
> 		SoMsg("err", err, ShouldBeNil)
> 		key := *chain.Key()
> 		SoMsg("Key", reflect.DeepEqual(key, Key{IA: addr.ISD_AS{I: 1, A: 10}, Ver: 1}), ShouldBeTrue)
> ```

`ShouldResemble`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

Oncilla:
Review status: 4 of 10 files reviewed at latest revision, 26 unresolved discussions.

---

*[go/lib/addr/isdas.go, line 85 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDSaYtlXOjwemDMDi2:-KxDhs0dZb8isoOtntNu:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/addr/isdas.go#L85)):*

Move before `UnmarshalText`.
</blockquote></details>

Done.

---

*[go/lib/crypto/asym.go, line 26 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDVumdd9kKprc5LPkG:-KxDi1bLuKsshvmX7euq:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/asym.go#L26)):*

The type isn't needed, it's derived from the value.
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/cert_test.go, line 79 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDbRKNP6LqruTi8cm3:-KxDjBrUYyXvh9uGHNFA:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/cert_test.go#L79)):*

You can use `ShouldResemble` for this. It will (attempt to) give readable output of the differences.
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/cert_test.go, line 85 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDc7_3XcemfQE0IE81:-KxDjV66_NkFjcY9Z9tI:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/cert_test.go#L85)):*

This is not safe. If a panic happens, this won't be reverted, and all other tests will fail, too. You need to make a copy and modify that.
```
tmpCert = append([]byte(nil), rawCert...)
tmpCert[0] ^= 0xFF
```
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/cert_test.go, line 103 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDcsFGgRxrY0rlkUwu:-KxDjc3yPgjB80vUAxoU:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/cert_test.go#L103)):*

"verifiable"
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/chain.go, line 43 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDYX6356G_lgN0N9TN:-KxDjzx87JJb93cUPLk0:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain.go#L43)):*

"one for the leaf".. "The leaf certificate"
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/chain.go, line 46 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDYdu6s8xLPKDyIxNQ:-KxDkJLetpZKhlDPUtBi:brhcttj) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain.go#L46)):*

Leaf, leaf, leaf, leaf, leaf. (leaf)
</blockquote></details>

Whoops.
Done.

---

*[go/lib/crypto/cert/chain.go, line 82 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDZJqz5WLSXFI4P6H8:-KxDlVEMhU3-RL7r2qgE:bzbefsv) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain.go#L82)):*

Wait, why? Why does it need a length? Why little-endian? The python code doesn't do this, afaik.

Compression can simply be (rough sketch):
```
buf := bytes.NewBuffer()
lzW := lz4.NewWriter(buf)
lzW.Write(raw)
return buf.Bytes()
```
Decompression would be:
```
var in bytes.Reader(raw)
var buf &bytes.Buffer
lzR := lz4.NewReader(in)
lzR.WriteTo(buf)
// buf.Bytes() is now the uncompressed raw chain
```

This also means that `MaxChainByteLength` can be removed.
</blockquote></details>

This is how the python library does it.

Decompression in python code
        if lz4_:
            chain_raw = lz4.loads(chain_raw).decode("utf-8")
Compression in python code
        if lz4_:
            return lz4.dumps(ret)
manual (load/dumps are deprecated names of compress/uncompress)
http://python-lz4.readthedocs.io/en/latest/lz4.block.html#lz4.block.compress

  Parameters:	
    - store_size (bool) – If True (the default) then the size of the uncompressed data is stored at the start of the compressed block.

---

*[go/lib/crypto/cert/chain_test.go, line 228 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDhcxpDjEVYZdx9i9J:-KxDmgzU2YwmNmyov7sH:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain_test.go#L228)):*

Core
</blockquote></details>

Done.

---

*[go/lib/crypto/cert/chain_test.go, line 239 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDhYQ5RHGKnm0p5fHT:-KxDmoJRNuMbiA9Qp0KX:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain_test.go#L239)):*

`ShouldResemble`
</blockquote></details>

Done.

---

*[go/lib/trust/store.go, line 29 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDVDDl4vahxo1CIEMy:-KxDnCpdHaHx8RzyhyTW:bfygb96) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/trust/store.go#L29)):*

This mentions TRCs, but the code below doesn't appear to deal with them.
</blockquote></details>

It eventually will.
Removed for now.

---

*[go/lib/trust/store.go, line 94 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDVNEnYIrwrWCMheuG:-KxDnYLRDUKG1LHTd-kV:b8pw6ho) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/trust/store.go#L94)):*

So this will only store a cert if it's a new maximum? What's the rationale?
</blockquote></details>

Good catch.
Some flawed logic
Done.

---

*[go/lib/trust/store.go, line 95 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDaID3g8roJ7svmbtF:-KxDpi69Tod3Rwupq47g:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/trust/store.go#L95)):*

`chain.JSON(true)`?
</blockquote></details>

Done.

---

*[go/lib/trust/store.go, line 111 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDVk_GFUoav7Y2PLZ5:-KxDpnRo0yqwm7QW0zJZ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/trust/store.go#L111)):*

You can just return this directly, without `chain`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

kormat:
Reviewed 1 of 10 files at r2, 6 of 6 files at r4.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[go/lib/crypto/cert/chain.go, line 82 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDZJqz5WLSXFI4P6H8:-KxDzFb-ON9okGqIesXH:b-3zw3jo) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain.go#L82)):*

This is how the python library does it.

Decompression in python code
        if lz4_:
            chain_raw = lz4.loads(chain_raw).decode("utf-8")
Compression in python code
        if lz4_:
            return lz4.dumps(ret)
manual (load/dumps are deprecated names of compress/uncompress)
http://python-lz4.readthedocs.io/en/latest/lz4.block.html#lz4.block.compress

  Parameters:	
    - store_size (bool) – If True (the default) then the size of the uncompressed data is stored at the start of the compressed block.
</blockquote></details>

:man_facepalming: Welp. And the python library requires knowing the size for decompression. Guh. And they don't actually _document_ the byte-ordering. Guh. Well, good job spotting this. Please document _why_ we have to do this in Go.

---

*[go/lib/trust/store.go, line 94 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDVNEnYIrwrWCMheuG:-KxE0EUs4KlYslK2n9EB:bi77wz8) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/trust/store.go#L94)):*

Good catch.
Some flawed logic
Done.

</blockquote></details>

Hmm. If  an app does `AddChain(c, false)`, and then later `AddChain(c, true)`, the write won't actually happen.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

Oncilla:
Review status: 8 of 10 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[go/lib/crypto/cert/chain.go, line 82 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDZJqz5WLSXFI4P6H8:-KxEW_fJ_9L4v7ZXWhY0:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/crypto/cert/chain.go#L82)):*

:man_facepalming: Welp. And the python library requires knowing the size for decompression. Guh. And they don't actually _document_ the byte-ordering. Guh. Well, good job spotting this. Please document _why_ we have to do this in Go.
</blockquote></details>

Done.

---

*[go/lib/trust/store.go, line 94 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-KxDVNEnYIrwrWCMheuG:-KxEWbo-KCBF0tPwHXX0:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/cf49da7fd31514ee06ca50f2b129f89dd3585071/go/lib/trust/store.go#L94)):*

Hmm. If  an app does `AddChain(c, false)`, and then later `AddChain(c, true)`, the write won't actually happen.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

kormat:
:lgtm:

---

Reviewed 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271#-:-KxIi2iRDZtyorEp-TC_:bnfp4nl)*

shitz:
Reviewed 2 of 10 files at r2, 3 of 8 files at r3, 4 of 6 files at r4, 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

Oncilla:
Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1271)*

PR #1272, opened 2017-10-06
Full revocation support in snet

scrye:
This PR adds revocation support to the Go application networking stack.

Due to concurrency issues, it depends on the changes suggested by @jpcsmith in https://github.com/netsec-ethz/scion/pull/1253.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1272)

### DIFF
--- a/go/lib/hpkt/read.go
+++ b/go/lib/hpkt/read.go
@@ -299,7 +299,10 @@ func (p *parseCtx) DefaultL4Parser() error {
 	case common.L4UDP:
 		p.s.Pld = common.RawBytes(p.b[p.offset : p.offset+pldLen])
 	case common.L4SCMP:
-		hdr, _ := p.s.L4.(*scmp.Hdr)
+		hdr, ok := p.s.L4.(*scmp.Hdr)
+		if !ok {
+			return common.NewCError("Unable to extract SCMP payload, type assertion failed.")
+		}
 		p.s.Pld, err = scmp.PldFromRaw(p.b[p.offset:p.offset+pldLen],
 			scmp.ClassType{Class: hdr.Class, Type: hdr.Type})
 		if err != nil {
--- a/go/lib/hpkt/read.go
+++ b/go/lib/hpkt/read.go
@@ -295,7 +295,17 @@ func (p *parseCtx) DefaultL4Parser() error {
 	if err = p.s.L4.Validate(pldLen); err != nil {
 		return common.NewCError("L4 validation failed", "err", err)
 	}
-	p.s.Pld = common.RawBytes(p.b[p.offset : p.offset+pldLen])
+	switch p.nextHdr {
+	case common.L4UDP:
+		p.s.Pld = common.RawBytes(p.b[p.offset : p.offset+pldLen])
+	case common.L4SCMP:
+		hdr, _ := p.s.L4.(*scmp.Hdr)
+		p.s.Pld, err = scmp.PldFromRaw(p.b[p.offset:p.offset+pldLen],
+			scmp.ClassType{Class: hdr.Class, Type: hdr.Type})
+		if err != nil {
+			return common.NewCError("Unable to parse SCMP payload", "err", err)
+		}
+	}
 	p.offset += pldLen
 	p.pldOffsets.end = p.offset
 
--- a/go/lib/pathmgr/apppath.go
+++ b/go/lib/pathmgr/apppath.go
@@ -17,6 +17,8 @@ package pathmgr
 import (
 	"crypto/sha256"
 	"encoding/binary"
+	"fmt"
+	"strings"
 
 	"github.com/netsec-ethz/scion/go/lib/common"
 	"github.com/netsec-ethz/scion/go/lib/sciond"
@@ -55,12 +57,26 @@ func (aps AppPathSet) Copy() AppPathSet {
 	return newAPS
 }
 
-// GetAppPath returns an arbitrary path from aps.
-func (aps AppPathSet) GetAppPath() *AppPath {
+// GetAppPath returns an AppPath from the set. It first tries to find
+// a path with key pref; if one cannot be found, an arbitrary one
+// is returned.
+func (aps AppPathSet) GetAppPath(pref PathKey) *AppPath {
+	ap, ok := aps[pref]
+	if ok {
+		return ap
+	}
 	for _, v := range aps {
 		return v
 	}
 	return nil
+}
+
+func (aps AppPathSet) String() string {
+	var desc []string
+	for _, path := range aps {
+		desc = append(desc, fmt.Sprintf("%v", path.Entry.Path.Interfaces))
+	}
+	return "{" + strings.Join(desc, ";") + "}"
 }
 
 // AppPath contains a SCIOND path entry.
--- a/go/lib/pathmgr/apppath.go
+++ b/go/lib/pathmgr/apppath.go
@@ -17,7 +17,6 @@ package pathmgr
 import (
 	"crypto/sha256"
 	"encoding/binary"
-	"fmt"
 	"strings"
 
 	"github.com/netsec-ethz/scion/go/lib/common"
@@ -74,7 +73,7 @@ func (aps AppPathSet) GetAppPath(pref PathKey) *AppPath {
 func (aps AppPathSet) String() string {
 	var desc []string
 	for _, path := range aps {
-		desc = append(desc, fmt.Sprintf("%v", path.Entry.Path.Interfaces))
+		desc = append(desc, path.Entry.Path.String())
 	}
 	return "{" + strings.Join(desc, ";") + "}"
 }
--- a/go/lib/pathmgr/apppath.go
+++ b/go/lib/pathmgr/apppath.go
@@ -60,9 +60,11 @@ func (aps AppPathSet) Copy() AppPathSet {
 // a path with key pref; if one cannot be found, an arbitrary one
 // is returned.
 func (aps AppPathSet) GetAppPath(pref PathKey) *AppPath {
-	ap, ok := aps[pref]
-	if ok {
-		return ap
+	if len(pref) > 0 {
+		ap, ok := aps[pref]
+		if ok {
+			return ap
+		}
 	}
 	for _, v := range aps {
 		return v
@@ -75,7 +77,7 @@ func (aps AppPathSet) String() string {
 	for _, path := range aps {
 		desc = append(desc, path.Entry.Path.String())
 	}
-	return "{" + strings.Join(desc, ";") + "}"
+	return "{" + strings.Join(desc, "; ") + "}"
 }
 
 // AppPath contains a SCIOND path entry.
--- a/go/lib/scmp/info.go
+++ b/go/lib/scmp/info.go
@@ -182,9 +182,10 @@ func NewInfoRevocation(infoF, hopF, ifID uint16, ingress bool,
 
 func InfoRevocationFromRaw(b common.RawBytes) (*InfoRevocation, error) {
 	p := &InfoRevocation{InfoPathOffsets: &InfoPathOffsets{}}
-	if err := restruct.Unpack(b, common.Order, &p.InfoPathOffsets); err != nil {
-		return nil, common.NewCError("Failed to unpack SCMP Revocation info", "err", err)
-	}
+	p.InfoPathOffsets.InfoF = common.Order.Uint16(b[0:])
+	p.InfoPathOffsets.HopF = common.Order.Uint16(b[2:])
+	p.InfoPathOffsets.IfID = common.Order.Uint16(b[4:])
+	p.InfoPathOffsets.Ingress = (b[6] & 0x01) == 0x01
 	p.RevToken = b[8:]
 	return p, nil
 }
--- a/go/lib/scmp/info.go
+++ b/go/lib/scmp/info.go
@@ -134,9 +134,10 @@ type InfoPathOffsets struct {
 
 func InfoPathOffsetsFromRaw(b common.RawBytes) (*InfoPathOffsets, error) {
 	p := &InfoPathOffsets{}
-	if err := restruct.Unpack(b, common.Order, p); err != nil {
-		return nil, common.NewCError("Failed to unpack SCMP Path Offsets info", "err", err)
-	}
+	p.InfoF = common.Order.Uint16(b[0:])
+	p.HopF = common.Order.Uint16(b[2:])
+	p.IfID = common.Order.Uint16(b[4:])
+	p.Ingress = (b[6] & 0x01) == 0x01
 	return p, nil
 }
 
@@ -181,11 +182,12 @@ func NewInfoRevocation(infoF, hopF, ifID uint16, ingress bool,
 }
 
 func InfoRevocationFromRaw(b common.RawBytes) (*InfoRevocation, error) {
+	var err error
 	p := &InfoRevocation{InfoPathOffsets: &InfoPathOffsets{}}
-	p.InfoPathOffsets.InfoF = common.Order.Uint16(b[0:])
-	p.InfoPathOffsets.HopF = common.Order.Uint16(b[2:])
-	p.InfoPathOffsets.IfID = common.Order.Uint16(b[4:])
-	p.InfoPathOffsets.Ingress = (b[6] & 0x01) == 0x01
+	p.InfoPathOffsets, err = InfoPathOffsetsFromRaw(b)
+	if err != nil {
+		return nil, common.NewCError("Unable to parse path offsets", "err", err)
+	}
 	p.RevToken = b[8:]
 	return p, nil
 }
--- a/go/lib/scmp/info.go
+++ b/go/lib/scmp/info.go
@@ -183,7 +183,7 @@ func NewInfoRevocation(infoF, hopF, ifID uint16, ingress bool,
 
 func InfoRevocationFromRaw(b common.RawBytes) (*InfoRevocation, error) {
 	var err error
-	p := &InfoRevocation{InfoPathOffsets: &InfoPathOffsets{}}
+	p := &InfoRevocation{}
 	p.InfoPathOffsets, err = InfoPathOffsetsFromRaw(b)
 	if err != nil {
 		return nil, common.NewCError("Unable to parse path offsets", "err", err)
--- a/go/lib/snet/conn.go
+++ b/go/lib/snet/conn.go
@@ -39,6 +39,25 @@ const (
 	BufSize = 1<<16 - 1
 )
 
+type Error interface {
+	error
+	SCMP() *scmp.Hdr
+}
+
+var _ Error = (*OpError)(nil)
+
+type OpError struct {
+	scmp *scmp.Hdr
+}
+
+func (e *OpError) SCMP() *scmp.Hdr {
+	return e.scmp
+}
+
+func (e *OpError) Error() string {
+	return e.scmp.String()
+}
+
 var _ net.Conn = (*Conn)(nil)
 var _ net.PacketConn = (*Conn)(nil)
 
@@ -85,17 +104,16 @@ func (c *Conn) ReadFromSCION(b []byte) (int, *Addr, error) {
 	if c.scionNet == nil {
 		return 0, nil, common.NewCError("SCION network not initialized")
 	}
-	// Loop until we have a data packet
-	for {
-		c.readMutex.Lock()
-		n, a, appData, err := c.read(b, true)
-		c.readMutex.Unlock()
-		if err != nil {
-			return 0, nil, common.NewCError("Dispatcher error", "err", err)
-		}
-		if appData {
-			return n, a, err
-		}
+	c.readMutex.Lock()
+	n, a, scmp, err := c.read(b, true)
+	c.readMutex.Unlock()
+	if err != nil {
+		return 0, nil, common.NewCError("Unable to read", "err", err)
+	}
+	if scmp == nil {
+		return n, a, err
+	} else {
+		return 0, nil, &OpError{scmp: scmp}
 	}
 }
 
@@ -110,15 +128,15 @@ func (c *Conn) Read(b []byte) (int, error) {
 	return n, err
 }
 
-// read returns the number of bytes read, the address that sent the bytes,
-// whether the message should be delivered to an application or if an error
-// occurred.
-func (c *Conn) read(b []byte, from bool) (int, *Addr, bool, error) {
+// read returns the number of bytes read, the address that sent the bytes, the
+// SCMP header (if an SCMP message was received, or nil otherwise) and if an
+// error occurred.
+func (c *Conn) read(b []byte, from bool) (int, *Addr, *scmp.Hdr, error) {
 	var err error
 	var remote *Addr
 	n, lastHop, err := c.conn.ReadFrom(c.recvBuffer)
 	if err != nil {
-		return 0, nil, false, common.NewCError("Dispatcher read error", "err", err)
+		return 0, nil, nil, common.NewCError("Dispatcher read error", "err", err)
 	}
 	if !from {
 		lastHop = nil
@@ -130,12 +148,12 @@ func (c *Conn) read(b []byte, from bool) (int, *Addr, bool, error) {
 	}
 	err = hpkt.ParseScnPkt(pkt, c.recvBuffer[:n])
 	if err != nil {
-		return 0, nil, false, common.NewCError("SCION packet parse error", "err", err)
+		return 0, nil, nil, common.NewCError("SCION packet parse error", "err", err)
 	}
 	// Copy data, extract address
 	n, err = pkt.Pld.WritePld(b)
 	if err != nil {
-		return 0, nil, false, common.NewCError("Unable to copy payload", "err", err)
+		return 0, nil, nil, common.NewCError("Unable to copy payload", "err", err)
 	}
 	// On UDP4 network we can get either UDP traffic or SCMP messages
 	if c.net == "udp4" {
@@ -151,23 +169,23 @@ func (c *Conn) read(b []byte, from bool) (int, *Addr, bool, error) {
 			if lastHop != nil {
 				path := pkt.Path
 				if err = path.Reverse(); err != nil {
-					return 0, nil, true,
+					return 0, nil, nil,
 						common.NewCError("Unable to reverse path on received packet", "err", err)
 				}
 				remote.Path = path
 				remote.NextHopHost = lastHop.Addr
 				remote.NextHopPort = lastHop.Port
 			}
-			return n, remote, true, nil
+			return n, remote, nil, nil
 		case *scmp.Hdr:
 			c.handleSCMP(hdr, pkt)
-			return n, remote, false, nil
+			return n, remote, hdr, nil
 		default:
-			return n, remote, true, common.NewCError("Unexpected SCION L4 protocol", "expected",
+			return n, remote, nil, common.NewCError("Unexpected SCION L4 protocol", "expected",
 				"UDP or SCMP", "actual", pkt.L4.L4Type())
 		}
 	}
-	return 0, nil, true, common.NewCError("Unknown network", "net", c.net)
+	return 0, nil, nil, common.NewCError("Unknown network", "net", c.net)
 }
 
 func (c *Conn) handleSCMP(hdr *scmp.Hdr, pkt *spkt.ScnPkt) {
--- a/go/lib/snet/conn.go
+++ b/go/lib/snet/conn.go
@@ -319,10 +319,7 @@ func (c *Conn) selectPathEntry(raddr *Addr) (*sciond.PathReplyEntry, error) {
 	var pathSet pathmgr.AppPathSet
 	// If the remote address is fixed, register source and destination for
 	// continous path updates
-	// FIXME(scrye): Temporarily, while the SIG uses Listen for the data plane
-	// have all traffic use Register instead of one shot Query
-	// if c.raddr == nil {
-	if false {
+	if c.raddr == nil {
 		pathSet = c.scionNet.pathResolver.Query(c.laddr.IA, raddr.IA)
 	} else {
 		// Sanity check, as Dial already initializes this
--- a/go/lib/snet/conn.go
+++ b/go/lib/snet/conn.go
@@ -19,7 +19,7 @@ import (
 	"sync"
 	"time"
 
-	//log "github.com/inconshreveable/log15"
+	log "github.com/inconshreveable/log15"
 
 	"github.com/netsec-ethz/scion/go/lib/addr"
 	"github.com/netsec-ethz/scion/go/lib/common"
@@ -28,6 +28,7 @@ import (
 	"github.com/netsec-ethz/scion/go/lib/overlay"
 	"github.com/netsec-ethz/scion/go/lib/pathmgr"
 	"github.com/netsec-ethz/scion/go/lib/sciond"
+	"github.com/netsec-ethz/scion/go/lib/scmp"
 	"github.com/netsec-ethz/scion/go/lib/sock/reliable"
 	"github.com/netsec-ethz/scion/go/lib/spath"
 	"github.com/netsec-ethz/scion/go/lib/spkt"
@@ -57,6 +58,8 @@ type Conn struct {
 	sp *pathmgr.SyncPaths
 	// Reference to SCION networking context
 	scionNet *Network
+	// Key of last used path, used to select the same path for the next packet
+	prefPathKey pathmgr.PathKey
 }
 
 // DialSCION calls DialSCION on the default networking context.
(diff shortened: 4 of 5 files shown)


scrye:
NB: Regarding concurrency, @kormat's points still stand. In `snet.Conn` readers race against readers and writers against writers due to the use of internal buffers. This will need to be fixed in the future.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-:-Kvl95O2JQFH1A8zjDXD:b-ke5g4y)*

shitz:
Reviewed 1 of 7 files at r2, 3 of 4 files at r3.
Review status: 4 of 8 files reviewed at latest revision, 9 unresolved discussions.

---

*[go/lib/pathmgr/pathmgr.go, line 214 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwA6jmJ9wmNO0k14d6:-KvwA6jmJ9wmNO0k14d7:b6m9d2a) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/pathmgr/pathmgr.go#L214)):*
> ```Go
> 	case sciond.RevUnknown, sciond.RevValid:
> 		uifid := UIFIDFromValues(parsedRev.IA(), common.IFIDType(parsedRev.IfID))
> 		log.Info("Revocation is valid", "uifid", uifid)
> ```

This should log `parsedRev`

---

*[go/lib/pathmgr/pathmgr.go, line 261 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwAHG4AKcKZN91a2TR:-KvwAHG4AKcKZN91a2TS:bz3enji) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/pathmgr/pathmgr.go#L261)):*
> ```Go
> 
> 	reply, err := r.sciond.Paths(q.dst, q.src, numReqPaths, sciond.PathReqFlags{})
> 	log.Info("SCIOND response received", "count", len(reply.Entries), "data", reply)
> ```

This looks like a debug statement to me.

---

*[go/lib/scmp/info.go, line 188 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-Kvw60oqQonYUMV5wV7f:-Kvw60oqQonYUMV5wV7g:bbj52go) ([raw file](https://github.com/netsec-ethz/scion/blob/c3d890fdd9968f1d60b7e2f6765b680cd2a3be7f/go/lib/scmp/info.go#L188)):*
> ```Go
> 	p.InfoPathOffsets.HopF = common.Order.Uint16(b[2:])
> 	p.InfoPathOffsets.IfID = common.Order.Uint16(b[4:])
> 	if (b[6] & 0x01) == 0 {
> ```

this can just be `b.InfoPathOffsets.Ingress = (b[6] & 0x01) == 1`

---

*[go/lib/snet/conn.go, line 139 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwApr9AHITUbgA8MH_:-KvwApr9AHITUbgA8MHa:b-eee4ta) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/conn.go#L139)):*
> ```Go
> 				L4Port: hdr.SrcPort,
> 			}
> 		case *scmp.Hdr:
> ```

Might be worth to split `read` into smaller chunks, e.g., revocation handling could be done in a `handleSCMPRev` method.

---

*[go/lib/snet/conn.go, line 140 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwAlO1mYd1WpGx9JId:-KvwAlO1mYd1WpGx9JIe:b-l6mwxj) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/conn.go#L140)):*
> ```Go
> 			}
> 		case *scmp.Hdr:
> 			log.Info("Received SCMP message")
> ```

remove

---

*[go/lib/snet/conn.go, line 143 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwAgg2PtAJWkEyhvsS:-KvwAgg2PtAJWkEyhvsT:bgk5ij0) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/conn.go#L143)):*
> ```Go
> 			// Only handle revocations for now
> 			if hdr.Class == scmp.C_Path && hdr.Type == scmp.T_P_RevokedIF {
> 				log.Info("SCMP message is revocation request")
> ```

Log the revocation once and nothing else.

---

*[go/lib/snet/conn.go, line 281 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwBGUp1emp1lK6BwKO:-KvwBGUp1emp1lK6BwKP:b87p2in) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/conn.go#L281)):*
> ```Go
> 	// If the remote address is fixed, register source and destination for
> 	// continous path updates
> 	// FIXME(scrye): Temporarily, while the SIG uses Listen for the data plane
> ```

Why is that?

---

*[go/lib/snet/conn.go, line 296 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwBEZvE5-t72tCUmQc:-KvwBEZvE5-t72tCUmQd:b-l6mwxj) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/conn.go#L296)):*
> ```Go
> 		}
> 		pathSet = c.sp.Load()
> 		//log.Info("Current path set is", "set", fmt.Sprintf("%v", pathSet))
> ```

remove

---

*[go/lib/snet/snet.go, line 100 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwBK7YgxUnB9wd1NMk:-KvwBK7YgxUnB9wd1NMl:b-60d6fr) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/snet.go#L100)):*
> ```Go
> 		dispatcherPath: dPath,
> 		localIA:        ia}
> 	pathResolver, err := pathmgr.New(sPath, 10*time.Second, log.Root())
> ```

Will this stay at 10 seconds or should we revert again to one minute?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272)*

scrye:
Review status: 4 of 8 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[go/lib/pathmgr/pathmgr.go, line 214 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwA6jmJ9wmNO0k14d6:-KvwUCJMvBLPzsv8dZ4J:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/pathmgr/pathmgr.go#L214)):*

This should log `parsedRev`
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 261 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwAHG4AKcKZN91a2TR:-KvwUHfhn7DVdgTddEE7:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/pathmgr/pathmgr.go#L261)):*

This looks like a debug statement to me.
</blockquote></details>

Done.

---

*[go/lib/scmp/info.go, line 188 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-Kvw60oqQonYUMV5wV7f:-KvwUeFySm_PEOlgugRb:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/c3d890fdd9968f1d60b7e2f6765b680cd2a3be7f/go/lib/scmp/info.go#L188)):*

this can just be `b.InfoPathOffsets.Ingress = (b[6] & 0x01) == 1`
</blockquote></details>

Done.

---

*[go/lib/snet/conn.go, line 139 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwApr9AHITUbgA8MH_:-KvwVlNvWbOtBP8QFkZo:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/conn.go#L139)):*

Might be worth to split `read` into smaller chunks, e.g., revocation handling could be done in a `handleSCMPRev` method.
</blockquote></details>

Done.

---

*[go/lib/snet/conn.go, line 140 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwAlO1mYd1WpGx9JId:-KvwVlu8qDUgxWq6GJjO:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/conn.go#L140)):*

remove
</blockquote></details>

Done.

---

*[go/lib/snet/conn.go, line 143 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwAgg2PtAJWkEyhvsS:-KvwWMF-jBqZxY2FAaxJ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/conn.go#L143)):*

Log the revocation once and nothing else.
</blockquote></details>

Done.

---

*[go/lib/snet/conn.go, line 281 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwBGUp1emp1lK6BwKO:-KvwWN4gnM9yDwOXJdQ3:b1slwju) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/conn.go#L281)):*

Why is that?
</blockquote></details>

`Dial` is currently hardcoded to `Register` source IA and destination IA for periodic lookups. 'Listen` is not.

Both should be changed to just use the expiring cache, and I will include this as part of the `pathmgr` refactoring. This is because the SIG registers them anyway via policies. For now, while the SIG code is in a constant state of flux revocations are easier to test if we make sure each source and destination are registered.

---

*[go/lib/snet/conn.go, line 296 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwBEZvE5-t72tCUmQc:-Kvwbw0azwxnhv3ObmNj:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/conn.go#L296)):*

remove
</blockquote></details>

Done.

---

*[go/lib/snet/snet.go, line 100 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwBK7YgxUnB9wd1NMk:-Kvwbzu2ui328jRqi0Pr:b-xvna45) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/snet.go#L100)):*

Will this stay at 10 seconds or should we revert again to one minute?
</blockquote></details>

It depends on how fast you want a path to go back up. This value describes the worst-case on how long it takes the app to recognize that the path has gone back up after SCIOND has it.

For the demo I'd have it be even more aggressive than 10 seconds. For normal operation, one minute is more natural.

(As an aside, having push notifications for such events would be great.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272)*

shitz:
Review status: 3 of 9 files reviewed at latest revision, 3 unresolved discussions.

---

*[go/lib/snet/conn.go, line 142 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwniXqa6_W9o10nrFc:-KvwniXqa6_W9o10nrFd:b-w4iy3w) ([raw file](https://github.com/netsec-ethz/scion/blob/59bb3083170ac6b154e0afbeb42af1127cefe3ee/go/lib/snet/conn.go#L142)):*
> ```Go
> 			c.handleSCMP(hdr, pkt)
> 			// Recursively try to read again to give the application UDP data
> 			return c.read(b)
> ```

This is dangerous if you continuously receive SCMP messages. I suggest you do this iteratively.

---

*[go/lib/snet/conn.go, line 166 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwnS4gj2gLUhRdNjBM:-KvwnS4gj2gLUhRdNjBN:b-jxebsu) ([raw file](https://github.com/netsec-ethz/scion/blob/59bb3083170ac6b154e0afbeb42af1127cefe3ee/go/lib/snet/conn.go#L166)):*
> ```Go
> 		log.Error("Unable to type assert payload to SCMP payload")
> 	}
> 	log.Info("SCMP payload dump", "payload", scmpPayload.String())
> ```

Remove this since you log the revocation below.

---

*[go/lib/snet/snet.go, line 100 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwBK7YgxUnB9wd1NMk:-Kvwoskq2Yc4GxkCPL5B:bu8nbg2) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/snet.go#L100)):*

It depends on how fast you want a path to go back up. This value describes the worst-case on how long it takes the app to recognize that the path has gone back up after SCIOND has it.

For the demo I'd have it be even more aggressive than 10 seconds. For normal operation, one minute is more natural.

(As an aside, having push notifications for such events would be great.)
</blockquote></details>

Yeah I agree push would be great for this and yes for the demo we want this to be much lower. Probably in the range of 1-2 seconds.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272)*

scrye:
Review status: 3 of 9 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/snet/conn.go, line 142 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwniXqa6_W9o10nrFc:-Kw0LxghvKNWG7cUg1UR:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/59bb3083170ac6b154e0afbeb42af1127cefe3ee/go/lib/snet/conn.go#L142)):*

This is dangerous if you continuously receive SCMP messages. I suggest you do this iteratively.
</blockquote></details>

Done.

---

*[go/lib/snet/conn.go, line 166 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwnS4gj2gLUhRdNjBM:-Kw0MMQYTIvX7G1wFg9p:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/59bb3083170ac6b154e0afbeb42af1127cefe3ee/go/lib/snet/conn.go#L166)):*

Remove this since you log the revocation below.
</blockquote></details>

Done.

---

*[go/lib/snet/snet.go, line 100 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KvwBK7YgxUnB9wd1NMk:-Kw0MNs2In_75524JMC3:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/3f4874b1fa0fb44b63a2e615f273d5ea023b006d/go/lib/snet/snet.go#L100)):*

Yeah I agree push would be great for this and yes for the demo we want this to be much lower. Probably in the range of 1-2 seconds.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272)*

scrye:
:warning:  IMPORTANT: For testing purposes, this PR has multiple hacks in the SIG. All SIG code should be discarded prior to merging.

---

Review status: 3 of 9 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-:-Kw0a62r5hl9x1RH1wsm:b-xda9sw)*

scrye:
As @shitz mentioned in an offline talk, the `SCMPReceiver` goroutine should still be included during merging.

---

Review status: 3 of 9 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-:-Kw0spgmDO95obH2kEqx:b66gncq)*

kormat:
Reviewed 1 of 7 files at r2, 2 of 4 files at r3, 1 of 6 files at r5, 1 of 3 files at r7.
Review status: 5 of 9 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[go/lib/hpkt/read.go, line 296 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAQb2Ykuo8ACSrf37U:-KwAQb2Z9QHIDByI7Qfx:b9s2ab3) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/hpkt/read.go#L296)):*
> ```Go
> 		p.s.Pld = common.RawBytes(p.b[p.offset : p.offset+pldLen])
> 	case common.L4SCMP:
> 		hdr, _ := p.s.L4.(*scmp.Hdr)
> ```

Why `_`? If the `L4` isn't an scmp.Hdr, that'll make this line pass, but the next line will panic.

---

*[go/lib/pathmgr/apppath.go, line 67 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAR-z9m6Zgv0z_qu6k:-KwAR-zANo2eF9L7LvKF:b8mxsz3) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/pathmgr/apppath.go#L67)):*
> ```Go
> 	var desc []string
> 	for _, path := range aps {
> 		desc = append(desc, fmt.Sprintf("%v", path.Entry.Path.Interfaces))
> ```

`path.Entry.Path.String()`

---

*[go/lib/pathmgr/pathmgr.go, line 260 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAQ7ZUYUXAScf1gxcr:-KwAQ7ZUYUXAScf1gxcs:b-vpvte1) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/pathmgr/pathmgr.go#L260)):*
> ```Go
> 	}
> 
> 	reply, err := r.sciond.Paths(q.dst, q.src, numReqPaths, sciond.PathReqFlags{Flush: true})
> ```

Is `Flush: true` required?

---

*[go/lib/pathmgr/pathmgr.go, line 261 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAQ4AbBpaRqU-kJRtC:-KwAQ4AbBpaRqU-kJRtD:bhra9x1) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/pathmgr/pathmgr.go#L261)):*
> ```Go
> 
> 	reply, err := r.sciond.Paths(q.dst, q.src, numReqPaths, sciond.PathReqFlags{Flush: true})
> 	log.Debug("SCIOND response received", "count", len(reply.Entries), "data", reply)
> ```

This can _probably_ be removed.

---

*[go/lib/scmp/info.go, line 188 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAQGoB7oaYy-Eq_-mx:-KwAQGoB7oaYy-Eq_-my:bj5b1gd) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/scmp/info.go#L188)):*
> ```Go
> 	p.InfoPathOffsets.HopF = common.Order.Uint16(b[2:])
> 	p.InfoPathOffsets.IfID = common.Order.Uint16(b[4:])
> 	p.InfoPathOffsets.Ingress = (b[6] & 0x01) == 0x01
> ```

Nooo. This code should be in `InfoPathOffsetsFromRaw`.

---

*[go/lib/snet/conn.go, line 88 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwATT_N1heBPoxhFeDK:-KwATT_N1heBPoxhFeDL:bmayebw) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/snet/conn.go#L88)):*
> ```Go
> 		return 0, nil, common.NewCError("SCION network not initialized")
> 	}
> 	// Loop until we have a data packet
> ```

I don't think this is the right choice. There are many types of scmp messages, and this logic means that the app will never be notified about them. Instead i suggest that in the case of an scmp packet arriving, return it to the application wrapped in an snet.Error, similar to https://golang.org/pkg/net/#Error. That can have an `SCMP()` method, that will return the scmp header/payload to the application, if it desires, and otherwise it can just call `ReadX` again.

---

*[go/sig/base/egress_worker.go, line 113 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAUHX8jgslzwgTj02W:-KwAUHX8jgslzwgTj02X:bcnuqxn) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/sig/base/egress_worker.go#L113)):*
> ```Go
> }
> 
> func (e *EgressWorker) SCMPReceiver() {
> ```

This can be replaced by `snet/PktDispatcher` when #1282 1282 is merged.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272)*

scrye:
Review status: 0 of 4 files reviewed at latest revision, 7 unresolved discussions.

---

*[go/lib/hpkt/read.go, line 296 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAQb2Ykuo8ACSrf37U:-KxxLkt4ZCnLMr3J40_U:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/hpkt/read.go#L296)):*

Why `_`? If the `L4` isn't an scmp.Hdr, that'll make this line pass, but the next line will panic.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/apppath.go, line 67 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAR-z9m6Zgv0z_qu6k:-KxxM0VHRFte8QD7amDM:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/pathmgr/apppath.go#L67)):*

`path.Entry.Path.String()`
</blockquote></details>

Done.

---

*[go/lib/scmp/info.go, line 188 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAQGoB7oaYy-Eq_-mx:-KyL6lrpVX5PsJPhl5wF:bme7zgx) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/scmp/info.go#L188)):*

Nooo. This code should be in `InfoPathOffsetsFromRaw`.
</blockquote></details>

Whoops, you're right. Moved.

---

*[go/lib/snet/conn.go, line 88 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwATT_N1heBPoxhFeDK:-KyL57YVJs7SwKafyQD8:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/snet/conn.go#L88)):*

I don't think this is the right choice. There are many types of scmp messages, and this logic means that the app will never be notified about them. Instead i suggest that in the case of an scmp packet arriving, return it to the application wrapped in an snet.Error, similar to https://golang.org/pkg/net/#Error. That can have an `SCMP()` method, that will return the scmp header/payload to the application, if it desires, and otherwise it can just call `ReadX` again.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 260 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAQ7ZUYUXAScf1gxcr:-KyL5PpBJtgxV4L2CkDx:bj487tf) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/pathmgr/pathmgr.go#L260)):*

Is `Flush: true` required?
</blockquote></details>

It wasn't needed. Refactoring deleted the code anyway.

---

*[go/lib/pathmgr/pathmgr.go, line 261 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAQ4AbBpaRqU-kJRtC:-KyL6VSWENpfnj09pS_1:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/lib/pathmgr/pathmgr.go#L261)):*

This can _probably_ be removed.
</blockquote></details>

Done.

---

*[go/sig/base/egress_worker.go, line 113 at r7](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KwAUHX8jgslzwgTj02W:-KyL6UqYfFRzQiwqC9Ho:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/20508c2e44eb02851f45c10724fc21dd0332486b/go/sig/base/egress_worker.go#L113)):*

This can be replaced by `snet/PktDispatcher` when #1282 1282 is merged.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272)*

scrye:
Review status: 1 of 5 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[go/lib/pathmgr/apppath.go, line 63 at r8](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyMKsTEOEGrw-t28u70:-KyPxC3c_pzYQOXKN4kA:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fbb80b8565d7adeddb396e86298b60961acaae71/go/lib/pathmgr/apppath.go#L63)):*

I'd feel happier if this was wrapped in `if len(pref) > 0 {}`, as `""` is a valid map key.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/apppath.go, line 78 at r8](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyME_OKLlJH7hXTwIPv:-KyPy-Ofdnsn_12rTC0w:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fbb80b8565d7adeddb396e86298b60961acaae71/go/lib/pathmgr/apppath.go#L78)):*

This looks like it will be very dense to read. Suggestion, use `"; "` as the separator (i.e. add a space after the `;`).
</blockquote></details>

Done.

---

*[go/lib/scmp/info.go, line 186 at r8](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyMGDeH0aD9WCV6z0dN:-KyPy6URJl9aEUk9Bg3T:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fbb80b8565d7adeddb396e86298b60961acaae71/go/lib/scmp/info.go#L186)):*

There's no need to initialise `InfoPathOffsets` here, as it's getting overwritten by the next line.
</blockquote></details>

Done.

---

*[go/lib/snet/conn.go, line 102 at r8](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyMK7md7f8Wn8tLkeIB:-KyPyC9ByEHP8em8oF8r:behxl6d) ([raw file](https://github.com/netsec-ethz/scion/blob/b8dcdd6b3f9087af3ff71f3708de4c0f1cba8c9a/go/lib/snet/conn.go#L102)):*

Changing this defeats the point of the `from` parameter to `read`.
</blockquote></details>

Yes, I wanted all calls to go through the same base logic. Refactored a bit s.t. only `ReadFrom`s reverse paths.

---

*[go/lib/snet/conn.go, line 134 at r8](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyMJoPxx_NEBAxWT02q:-KyQ1hu7perwbPmaLOi-:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fbb80b8565d7adeddb396e86298b60961acaae71/go/lib/snet/conn.go#L134)):*

Why add another return value, and not just return an OpError with the scmp header inside if an scmp error happens?
</blockquote></details>

Done.

---

*[go/lib/snet/conn.go, line 197 at r8](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyMKVZaWwTdp5HB9Ran:-KyQ1ork_qUCp9i1ZvBN:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/fbb80b8565d7adeddb396e86298b60961acaae71/go/lib/snet/conn.go#L197)):*

No need to line-wrap this.
</blockquote></details>

Done.

---

*[go/lib/snet/conn.go, line 222 at r8](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyMKej5vLrH-3MNQEno:-KyQ1peZBUzqrTz0_dUa:bj04da0) ([raw file](https://github.com/netsec-ethz/scion/blob/fbb80b8565d7adeddb396e86298b60961acaae71/go/lib/snet/conn.go#L222)):*

Why move the mutex lock/unlock?
</blockquote></details>

In the previous version, this was inside a loop; of course, it doesn't make sense to have it like this any more.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272)*

kormat:
Reviewed 3 of 3 files at r9.

---

*[go/lib/snet/conn.go, line 222 at r8](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyMKej5vLrH-3MNQEno:-KyR507H9MjqtEWyRzQL:br6rrro) ([raw file](https://github.com/netsec-ethz/scion/blob/fbb80b8565d7adeddb396e86298b60961acaae71/go/lib/snet/conn.go#L222)):*

In the previous version, this was inside a loop; of course, it doesn't make sense to have it like this any more.
</blockquote></details>

Nope :)

---

*[go/lib/snet/conn.go, line 120 at r9](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyR3KN1sRBDD3R2x5wQ:-KyR3KN1sRBDD3R2x5wR:b-4nh4gw) ([raw file](https://github.com/netsec-ethz/scion/blob/44e6ccfc639ed3dfd2e7ddf3b0cb2fb6ef922bc2/go/lib/snet/conn.go#L120)):*
> ```Go
> // read returns the number of bytes read, the address that sent the bytes, the
> // SCMP header (if an SCMP message was received, or nil otherwise) and if an
> // error occurred.
> ```

Needs updating.

---

*[go/lib/snet/conn.go, line 174 at r9](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyR1uMi6CPnZzra10OZ:-KyR1uMjEPzWTlmxD_fy:b-hxu0cv) ([raw file](https://github.com/netsec-ethz/scion/blob/44e6ccfc639ed3dfd2e7ddf3b0cb2fb6ef922bc2/go/lib/snet/conn.go#L174)):*
> ```Go
> 		case *scmp.Hdr:
> 			c.handleSCMP(hdr, pkt)
> 			return n, remote, &OpError{scmp: hdr}
> ```

`remote` is not populated for SCMP packets.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272)*

scrye:
Review status: 3 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[go/lib/snet/conn.go, line 222 at r8](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyMKej5vLrH-3MNQEno:-KyR5fEf1nlp-xhbtOf7:b-93pfs6) ([raw file](https://github.com/netsec-ethz/scion/blob/fbb80b8565d7adeddb396e86298b60961acaae71/go/lib/snet/conn.go#L222)):*

Nope :)
</blockquote></details>

Ugh, right (write, get it?). I now understood what happened here. One version of the code had one set of mutexes, and the version I rebased on implemented the same functionality in a different place. During rebasing one replaced the other by mistake. Fixed now.

---

*[go/lib/snet/conn.go, line 120 at r9](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyR3KN1sRBDD3R2x5wQ:-KyR8pTcp_LJjUYt4Xc2:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/44e6ccfc639ed3dfd2e7ddf3b0cb2fb6ef922bc2/go/lib/snet/conn.go#L120)):*

Needs updating.
</blockquote></details>

Done.

---

*[go/lib/snet/conn.go, line 174 at r9](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-KyR1uMi6CPnZzra10OZ:-KyR8dVXOzW1QPpjmWvf:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/44e6ccfc639ed3dfd2e7ddf3b0cb2fb6ef922bc2/go/lib/snet/conn.go#L174)):*

`remote` is not populated for SCMP packets.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1272#-:-KyRATXjx6U9OvVl1RPI:bnfp4nl)*

PR #1242, opened 2017-09-01
Initial work for hidden path control packets

jonghoonkwon:
New Payload class "HPATH" that includes the following six payload types:
    - Hidden path register
    - Hidden path request
    - Hidden path reply
    - Hidden path config register
    - Hidden path config request
    - Hidden path config reply

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1242)

PR #1273, opened 2017-10-07
 Added hidden path mgmt packet formats for Python

shitz:
Also:
* Added a helper function to calculate the length of a proto schema
* Renamed all `sig_packX` to `sig_pack`

Supersedes #1242

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1273)

### DIFF
--- a/proto/ctrl_pld.capnp
+++ b/proto/ctrl_pld.capnp
@@ -10,6 +10,7 @@ using PathMgmt = import "path_mgmt.capnp";
 using SIBRA = import "sibra.capnp";
 using DRKeyMgmt = import "drkey_mgmt.capnp";
 using SIG = import "sig.capnp";
+using HPMgmt = import "hp_mgmt.capnp";
 
 struct CtrlPld {
     union {
@@ -21,5 +22,6 @@ struct CtrlPld {
         sibra @5 :SIBRA.SibraPayload;
         drkeyMgmt @6 :DRKeyMgmt.DRKeyMgmt;
         sig @7 :SIG.SIGCtrl;
+        hpMgmt @8 :HPMgmt.HPMgmt;
     }
 }
--- a/proto/hp_cfg.capnp
+++ b/proto/hp_cfg.capnp
@@ -1,0 +1,18 @@
+@0xfb570d9dd09c7fb7;
+
+using Go = import "go.capnp";
+$Go.package("proto");
+$Go.import("github.com/netsec-ethz/scion/go/proto");
+
+struct HPCfgId {
+    masterIA @0 :Data;
+    cfgId @1 :UInt64;
+}
+
+struct HPCfg {
+    id @0 :HPCfgId;
+    version @1: UInt64;
+    hpsIAs @2 :List(Data);  # ISD_AS of hidden path servers that stores hidden paths
+    writerIAs @3 :List(Data);  # ISD_AS of ASes that are authorized to register hidden paths
+    readerIAs @4 :List(Data);  # ISD_AS of ASes authorized to use the hidden paths
+}
--- a/proto/hp_mgmt.capnp
+++ b/proto/hp_mgmt.capnp
@@ -1,0 +1,29 @@
+@0xe8550b6088706947;
+
+using Go = import "go.capnp";
+$Go.package("proto");
+$Go.import("github.com/netsec-ethz/scion/go/proto");
+
+using PathMgmt = import "path_mgmt.capnp";
+using HPCfg = import "hp_cfg.capnp";
+
+struct HPCfgReq {
+    hpCfgIds @0 :List(HPCfg.HPCfgId);
+}
+
+struct HPCfgRecs {
+    hpCfgs @0 :List(HPCfg.HPCfg);
+}
+
+struct HPMgmt {
+    timestamp @0 :UInt64;
+    signature @1 :Data;
+    union {
+        hpCfgReq @2 :HPCfgReq;
+        hpCfgReply @3 :HPCfgRecs;
+        hpCfgReg @4 :HPCfgRecs;
+        hpSegReq @5 :PathMgmt.SegReq;
+        hpSegReply @6 :PathMgmt.SegRecs;
+        hpSegReg @7 :PathMgmt.SegRecs;
+    }
+}
--- a/proto/path_mgmt.capnp
+++ b/proto/path_mgmt.capnp
@@ -6,6 +6,7 @@ $Go.import("github.com/netsec-ethz/scion/go/proto");
 using PCB = import "pcb.capnp";
 using IFState = import "if_state.capnp";
 using RevInfo = import "rev_info.capnp";
+using HPCfg = import "hp_cfg.capnp";
 
 struct SegReq {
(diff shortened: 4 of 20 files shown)


kormat:
Reviewed 6 of 20 files at r1.
Review status: 6 of 20 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[proto/hp_cfg.capnp, line 2 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1273#-KxItGnaz05NPluPSM6G:-KxItGnaz05NPluPSM6H:b-8q7bbn) ([raw file](https://github.com/netsec-ethz/scion/blob/17c4ad15d3bdad8be79f9e29f0983179a4ec2127/proto/hp_cfg.capnp#L2)):*
> ```
> @0xfb570d9dd09c7fb7;
> ```

Remove blank line, same for other new files.

---

*[proto/hp_cfg.capnp, line 15 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1273#-KxIuj02us_iXktxdMeg:-KxIuj02us_iXktxdMeh:b-wwmot2) ([raw file](https://github.com/netsec-ethz/scion/blob/17c4ad15d3bdad8be79f9e29f0983179a4ec2127/proto/hp_cfg.capnp#L15)):*
> ISD_AS of ASes which run the hidden path services for this HPCfg.

---

*[proto/hp_mgmt.capnp, line 23 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1273#-KxItzx-YMP7UAIInnsA:-KxItzx-YMP7UAIInnsB:b-i75515) ([raw file](https://github.com/netsec-ethz/scion/blob/17c4ad15d3bdad8be79f9e29f0983179a4ec2127/proto/hp_mgmt.capnp#L23)):*
> ```
>     union {
>         hpCfgReq @2 :HPCfgReq;
>         hpCfgReply @3 :HPCfgRecs;
> ```

Don't include "Reply" in the names. This ends up with nonsensical things like cert servers logging that they received a CertChainReply when they have never sent a CertChainReq. `hpCfgs`.

---

*[proto/path_mgmt.capnp, line 27 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1273#-KxItfzI669l1Fr6ea5V:-KxItfzI669l1Fr6ea5W:bz8nh0j) ([raw file](https://github.com/netsec-ethz/scion/blob/17c4ad15d3bdad8be79f9e29f0983179a4ec2127/proto/path_mgmt.capnp#L27)):*
> ```
>     meta :group {
>         revInfos @1 :List(RevInfo.RevInfo);
>         hpCfgIds @2 :List(HPCfg.HPCfgId);
> ```

This doesn't look right. How are the hpCfgIds mapped to the segments (and vice-versa)? Or is the implication that all recs in this message have the same set of hpCfgIds? My original proposal was to put a list of hpCfgIDs into `PathSegMeta`.

https://github.com/kormat/scion/commit/71843d2c01e77c508f50598d346473e88144057a

---

*[python/lib/util.py, line 399 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1273#-KxJ-hPTizeo9WebHUGW:-KxJ-hPTizeo9WebHUGX:b-p60ws9) ([raw file](https://github.com/netsec-ethz/scion/blob/17c4ad15d3bdad8be79f9e29f0983179a4ec2127/python/lib/util.py#L399)):*
> ```Python
>     """Calculates the number of fields in a capnp schema."""
>     union_fields = schema.union_fields
>     # Add up all the fields in a union.
> ```

This is misleading. `Add up all fields in the unnamed union, if present`.

---

*[python/lib/util.py, line 402 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1273#-KxJ-rEIc_zqWjdfasxs:-KxJ-rEIc_zqWjdfasxt:b6owhof) ([raw file](https://github.com/netsec-ethz/scion/blob/17c4ad15d3bdad8be79f9e29f0983179a4ec2127/python/lib/util.py#L402)):*
> ```Python
>     cnt = len(union_fields)
>     for name, field in schema.fields.items():
>         # Fields within unions are already accounted for. Skip.
> ```

`Fields within the unnamed union (if any) are already accounted for, skip.`

But, if this for-loop is going to loop over those fields _anyway_, there's no need to special-case them.

---

*[python/lib/util.py, line 411 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1273#-KxJ3v5gBqumiwn6UjAN:-KxJ3v5gBqumiwn6UjAO:b-fq6i9b) ([raw file](https://github.com/netsec-ethz/scion/blob/17c4ad15d3bdad8be79f9e29f0983179a4ec2127/python/lib/util.py#L411)):*
> ```Python
>             else:
>                 cnt += 1
>         except capnp.lib.capnp.KjException:
> ```

This is rather crude - it's not possible to tell what's causing the exception.

Here's my proposal:
```
cnt = 0
for name, field in schema.fields.items():
  try:
    s = field.schema
   except capnp.lib.capnp.KjException:
    cnt += 1
    continue
  if isinstance(s, capnp.lib.capnp._StructSchema) and s.node.struct.isGroup:
    cnt += len(s.fields)
  else:
    cnt += 1   
```

---

*[python/lib/util.py, line 421 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1273#-KxJ2HMlFCp4tAFiBhOx:-KxJ2HMlFCp4tAFiBhOy:b-1g0pnr) ([raw file](https://github.com/netsec-ethz/scion/blob/17c4ad15d3bdad8be79f9e29f0983179a4ec2127/python/lib/util.py#L421)):*
> ```Python
>     # Simple heuristic to distinguish toplevel groups from toplevel structs:
>     # If schema name contains a '.' it's a group, otherwise a struct.
>     field_name = schema.node.displayName.split(":")[1]
> ```

I think you can just do `return schema.node.struct.isGroup`

---

*[python/lib/packet/pcb.py, line 133 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1273#-KxJ66BXetdzzfPy6ao1:-KxJ66BXetdzzfPy6ao2:b-h8gq75) ([raw file](https://github.com/netsec-ethz/scion/blob/17c4ad15d3bdad8be79f9e29f0983179a4ec2127/python/lib/packet/pcb.py#L133)):*
> ```Python
>         self.p.from_dict(d)
> 
>     def sig_pack(self):
> ```

Changing this is a mistake. Version 8 of the PCB _must_ call pcbm.sig_pack5, as an example. Otherwise host A which has ver 8 PCB and ver 5 PCBM proto definitions will get a different result than someone with a ver 8 PCB and a ver 6 PCBM proto definition, and signatures become unusable.

Well, crap. Actually, the current scheme _can't_ cope with an internal proto change without there being a change in the top-level proto. That's a big problem. Ok, we can't use the number of fields to version things. We probably need to add an explicit version field. Ugh, this is going to get messy.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1273)*

scrye:
Python is no longer in development and will soon be deprecated, so closing this.

PR #1275, opened 2017-10-07
Refactor sig egress to be able to use path policies.

kormat:
This change also splits up egress into a dispatchers (reads from a TUN
device, and writes to policies ringbuf's) and workers (read from a
policiy's ringbuf, encapsulates the packets, and send them to the remote
SIG).

As packet classifiers and path selection policies aren't merged just
yet, currently there is a single default "policy" that the dispatcher
writes to.

Note:
- Currently the path an egress worker gets is never changed, that
  functionality will be in a following PR.
- The egress code should _probably_ be moved to its own module,
  but that will also require moving PathPolicies and SIGInfo out to
  their own modules, too, at a minimum, so i've decided to not do it
  just yet.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1275)


shitz:
Review status: 0 of 10 files reviewed at latest revision, 9 unresolved discussions.

---

*[go/sig/base/as.go, line 183 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrLm6RgH9qYx5Yn8AY:-KvrLm6SO9hsJMe0hkkW:b-9dyuds) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/as.go#L183)):*
> ```Go
> // Internal method to return an arbitrary SIG
> func (ae *ASEntry) getSig() *SIGEntry {
> 	for _, se := range ae.Sigs {
> ```

Hmm this could be heavily biased towards one specific SIG. If this should really return an arbitrary then the choice should be random.

---

*[go/sig/base/as.go, line 200 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrMJeqF2AEVRWZJlP8:-KvrMJeru_wPmYb7Sl3p:buipuhp) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/as.go#L200)):*
> ```Go
> 	ae.PathPolicies.Store(pps)
> 	go NewEgressWorker(ae, pp, ae.conn).Run()
> 	if len(pps) == 1 {
> ```

We have to be careful about this when we add support for removing policies. Also it seems that you decided to have one dispatcher per AS, which is different from the design doc.

---

*[go/sig/base/egress_dispatcher.go, line 40 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrC-kw6CzezLOs_SM0:-KvrC-kw6CzezLOs_SM1:bx9baqq) ([raw file](https://github.com/netsec-ethz/scion/blob/77e806fb3776fee15d88e505d211ba34bd3d085b/go/sig/base/egress_dispatcher.go#L40)):*
> ```Go
> 	egressFreePkts = ringbuf.New(egressFreePktsCap, func() interface{} {
> 		return make(common.RawBytes, 1<<16)
> 	}, "egress", prometheus.Labels{"ringId": "freeFrames"})
> ```

`freePkts` prevents confusing this with the ingress `freeFrames` ringbuffer.

---

*[go/sig/base/egress_worker.go, line 173 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrJ6N1WEN19tgBTyCA:-KvrJ6N1WEN19tgBTyCB:berzo8c) ([raw file](https://github.com/netsec-ethz/scion/blob/77e806fb3776fee15d88e505d211ba34bd3d085b/go/sig/base/egress_worker.go#L173)):*
> ```Go
> 
> func (e *EgressWorker) resetFrame(f *frame) {
> 	e.currPath = e.pol.CurrPath()
> ```

It is a bit weird that you also refresh the path when you reset a frame and maybe also not necessary after every write to a frame. I suggest you move refreshing the path elsewhere, e.g., at the start of Write (and in case you haven't received something for a long time). You could also consider doing it only every N frames should this become a bottleneck.

---

*[go/sig/base/egress_worker.go, line 145 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrKRD0Piq8i8N8Fwdp:-KvrKRD0Piq8i8N8Fwdq:b-5wfn22) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/egress_worker.go#L145)):*
> ```Go
> 	if e.currPath == nil {
> 		// FIXME(kormat): add some metrics to track this.
> 		return nil
> ```

Hmm I feel like this should return some error to let the worker know that it has to do something about the frame. At least add a TODO if you plan to do this in your upcoming PR.

---

*[go/sig/base/egress_worker.go, line 150 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrKbtrwdI1QXQRz926:-KvrKbtrwdI1QXQRz927:brhumzt) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/egress_worker.go#L150)):*
> ```Go
> 	if sig == nil {
> 		// FIXME(kormat): add some metrics to track this.
> 		return nil
> ```

Same here.

---

*[go/sig/base/pathmgmt.go, line 46 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrNIEBhIbN37RmGCIv:-KvrNIEBhIbN37RmGCIw:b-dx39uk) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/pathmgmt.go#L46)):*
> ```Go
> // a pool of paths that match the specified policy, metrics about those paths,
> // as well as maintaining the currently favoured path and remote SIG to use.
> type PathPolicy struct {
> ```

I feel like this should be called differently, since this struct contains much more than just a policy. 

---

*[go/sig/base/pathmgmt.go, line 48 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrNrAXLtKBSNInwQJ3:-KvrNrAXLtKBSNInwQJ4:b-h8wn51) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/pathmgmt.go#L48)):*
> ```Go
> type PathPolicy struct {
> 	IA     *addr.ISD_AS
> 	policy interface{}
> ```

Add FIXME, since this seems to be temporary.

---

*[go/sig/sigcmn/common.go, line 71 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrOFhnpeAmEFZoe9fg:-KvrOFhnpeAmEFZoe9fh:b-3vkq6f) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/sigcmn/common.go#L71)):*
> ```Go
> 	}
> 	if PathMgr, err = pathmgr.New(*sciondPath, time.Minute, log.Root()); err != nil {
> ```

Do something here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1275)*

kormat:
Review status: 0 of 10 files reviewed at latest revision, 9 unresolved discussions.

---

*[go/sig/base/as.go, line 183 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrLm6RgH9qYx5Yn8AY:-KvrOXK1eZX3mmCLliVM:b-tj3vvj) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/as.go#L183)):*

Hmm this could be heavily biased towards one specific SIG. If this should really return an arbitrary then the choice should be random.
</blockquote></details>

This is currently only used to just pick _some_ SIG to start with. When we have more structure, this will skip non-active SIGs, and maybe favour SIGs with which we already have exchanged keys. This depends on the keepalive mechanism. Besides, the remote SIG can decide that it doesn't want so many sessions, and refer you to another SIG.

---

*[go/sig/base/as.go, line 200 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrMJeqF2AEVRWZJlP8:-KvrOoL6LbKQ1AjeJj8t:b-mga5ln) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/as.go#L200)):*

We have to be careful about this when we add support for removing policies. Also it seems that you decided to have one dispatcher per AS, which is different from the design doc.
</blockquote></details>

The issue is that right now the SIG relies on having a `tun` device per remote AS. Until we change that, and do the ip-range matching ourselves, having a single dispatcher doesn't really work.

The counter-part to this is when `DelPolicy` exists, when the number of policies hit zero, we should close the `tun` device (which will stop the egress dispatcher). I've added a comment for this.

---

*[go/sig/base/egress_dispatcher.go, line 40 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrC-kw6CzezLOs_SM0:-KvrPqB-CTYJ65fVcYv8:bfv3b4q) ([raw file](https://github.com/netsec-ethz/scion/blob/77e806fb3776fee15d88e505d211ba34bd3d085b/go/sig/base/egress_dispatcher.go#L40)):*

`freePkts` prevents confusing this with the ingress `freeFrames` ringbuffer.
</blockquote></details>

Oops, good catch.

---

*[go/sig/base/egress_worker.go, line 173 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrJ6N1WEN19tgBTyCA:-KvrQTo-Ld8uRh_ys2Q-:b-e17pdh) ([raw file](https://github.com/netsec-ethz/scion/blob/77e806fb3776fee15d88e505d211ba34bd3d085b/go/sig/base/egress_worker.go#L173)):*

It is a bit weird that you also refresh the path when you reset a frame and maybe also not necessary after every write to a frame. I suggest you move refreshing the path elsewhere, e.g., at the start of Write (and in case you haven't received something for a long time). You could also consider doing it only every N frames should this become a bottleneck.
</blockquote></details>

The path is tied to the frame resetting, as the frame length depends on the current path. It's not possible to unconditionally replace the path at the start of Write - if the new path has a lower mtu, then the current frame would have to be dropped (hence the comments at the start of Write). To put it another way, the MTU has to be known before any data is written to the frame, unless we want to add complexity.

Also, this path update is _very_ cheap. It's a single pointer load from `atomic.Value`. We could also replace it with https://golang.org/pkg/sync/atomic/#LoadPointer to make it effectively a single instruction, but i really can't see that being necessary. The main egress copies and writes are far _far_ more expensive.

---

*[go/sig/base/egress_worker.go, line 145 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrKRD0Piq8i8N8Fwdp:-KvrQ5HUwV95DJ8gEeke:b-rt8y0a) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/egress_worker.go#L145)):*

Hmm I feel like this should return some error to let the worker know that it has to do something about the frame. At least add a TODO if you plan to do this in your upcoming PR.
</blockquote></details>

The worker doesn't have to do anything, the frame is automatically dropped. Adding logic to not drop frames for a certain period until a new path appears would greatly increase complexity, and not something i'm sure i want to add.

---

*[go/sig/base/pathmgmt.go, line 46 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrNIEBhIbN37RmGCIv:-KvrRuYgAiAo-YcJCy3r:b-bjr658) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/pathmgmt.go#L46)):*

I feel like this should be called differently, since this struct contains much more than just a policy. 
</blockquote></details>

I'm open to suggestions.

---

*[go/sig/base/pathmgmt.go, line 48 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrNrAXLtKBSNInwQJ3:-KvrSowzprNvgLGcGG4Q:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/pathmgmt.go#L48)):*

Add FIXME, since this seems to be temporary.
</blockquote></details>

Done.

---

*[go/sig/sigcmn/common.go, line 71 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrOFhnpeAmEFZoe9fg:-KvrSz3vuHnbU6v21aiS:b-63i335) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/sigcmn/common.go#L71)):*

Do something here?
</blockquote></details>

Oops, done. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1275)*

shitz:
Review status: 0 of 10 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/sig/base/as.go, line 183 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrLm6RgH9qYx5Yn8AY:-Kvrg78dvQ78ReZlRjV2:b-677hhm) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/as.go#L183)):*

This is currently only used to just pick _some_ SIG to start with. When we have more structure, this will skip non-active SIGs, and maybe favour SIGs with which we already have exchanged keys. This depends on the keepalive mechanism. Besides, the remote SIG can decide that it doesn't want so many sessions, and refer you to another SIG.
</blockquote></details>

Even if this is just used initially you'd want to avoid the possibility that the same SIG is chosen very often. It's not super crucial for the small deployments that we have so far, but then adding a random, unbiased choice here also isn't a big overhead.

---

*[go/sig/base/egress_worker.go, line 173 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrJ6N1WEN19tgBTyCA:-Kvrfpy6cM-ZvAfa-Xio:b-mx9h11) ([raw file](https://github.com/netsec-ethz/scion/blob/77e806fb3776fee15d88e505d211ba34bd3d085b/go/sig/base/egress_worker.go#L173)):*

The path is tied to the frame resetting, as the frame length depends on the current path. It's not possible to unconditionally replace the path at the start of Write - if the new path has a lower mtu, then the current frame would have to be dropped (hence the comments at the start of Write). To put it another way, the MTU has to be known before any data is written to the frame, unless we want to add complexity.

Also, this path update is _very_ cheap. It's a single pointer load from `atomic.Value`. We could also replace it with https://golang.org/pkg/sync/atomic/#LoadPointer to make it effectively a single instruction, but i really can't see that being necessary. The main egress copies and writes are far _far_ more expensive.
</blockquote></details>

Right that makes sense.

---

*[go/sig/base/egress_worker.go, line 145 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrKRD0Piq8i8N8Fwdp:-KvreqFZRrPD07bKRhAn:bix9lgg) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/egress_worker.go#L145)):*

The worker doesn't have to do anything, the frame is automatically dropped. Adding logic to not drop frames for a certain period until a new path appears would greatly increase complexity, and not something i'm sure i want to add.
</blockquote></details>

Hmm but what about the case where you have some high prio traffic that should use a certain (set of) paths. If all of those paths become unavailable, then we'd be dropping this traffic. Instead we should probably add the packet to some other path policy (e.g., the default one) and try sending it out there. I'm aware that this increases the complexity and that this can be done later, but not doing anything doesn't seem the right decision for some cases.

---

*[go/sig/base/pathmgmt.go, line 46 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrNIEBhIbN37RmGCIv:-Kvrl9vVI0URM3wGkciW:b-1e40cm) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/pathmgmt.go#L46)):*

I'm open to suggestions.
</blockquote></details>

I was hoping this would inspire you to come up with something better. I have to admit I'm struggling finding a better term.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1275)*

kormat:
Review status: 0 of 10 files reviewed at latest revision, 3 unresolved discussions.

---

*[go/sig/base/as.go, line 183 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrLm6RgH9qYx5Yn8AY:-Kvv1kMR3eUezEpqtOnW:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/as.go#L183)):*

Even if this is just used initially you'd want to avoid the possibility that the same SIG is chosen very often. It's not super crucial for the small deployments that we have so far, but then adding a random, unbiased choice here also isn't a big overhead.
</blockquote></details>

Done.

---

*[go/sig/base/egress_worker.go, line 145 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrKRD0Piq8i8N8Fwdp:-Kvv1nAPHMSzDBpEz1SC:bbrnaoc) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/egress_worker.go#L145)):*

Hmm but what about the case where you have some high prio traffic that should use a certain (set of) paths. If all of those paths become unavailable, then we'd be dropping this traffic. Instead we should probably add the packet to some other path policy (e.g., the default one) and try sending it out there. I'm aware that this increases the complexity and that this can be done later, but not doing anything doesn't seem the right decision for some cases.
</blockquote></details>

The answer is that this is handled at a higher layer. When the packet classification is implemented, the dispatcher will walk the list of classifiers looking for a match. When a match is found, it will then walk the associated list of path policies, looking for a healthy policy (i.e. that the PathPolicy currently has a path + sig), and will then use that. A path policy without a path will be skipped.

```
type ASEntry struct {
    ...
    // Contains rules for classifying a packet, and a list of path policies to try in sequence:
    Classes []Classifier
    PathPolicies []PathPolicy
    ...
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1275)*

shitz:
:lgtm:

---

Reviewed 6 of 10 files at r1, 4 of 4 files at r3.

---

*[go/sig/base/egress_worker.go, line 145 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrKRD0Piq8i8N8Fwdp:-Kvv7NDJV8lsSpkUJsxb:brxp9sq) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/egress_worker.go#L145)):*

The answer is that this is handled at a higher layer. When the packet classification is implemented, the dispatcher will walk the list of classifiers looking for a match. When a match is found, it will then walk the associated list of path policies, looking for a healthy policy (i.e. that the PathPolicy currently has a path + sig), and will then use that. A path policy without a path will be skipped.

```
type ASEntry struct {
    ...
    // Contains rules for classifying a packet, and a list of path policies to try in sequence:
    Classes []Classifier
    PathPolicies []PathPolicy
    ...
}
```
</blockquote></details>

Sounds good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-:-Kvv7tS96DCDxmn3yGgi:bnfp4nl)*

scrye:
Review status: all files reviewed at latest revision, 13 unresolved discussions, all commit checks successful.

---

*[go/sig/base/egress_dispatcher.go, line 34 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvwE0PD2wNc0cBfQFT:-KvvwE0PD2wNc0cBfQFU:b5s5tp4) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_dispatcher.go#L34)):*
> ```Go
> 
> var (
> 	egressFreePkts *ringbuf.Ring
> ```

One thing I noticed throughout the code is that there quite some boilerplate relating to casting and initialization relating to `ringbuf.Ring` (`EntryList` initialization, casts to `common.RawBytes`, etc.). Embedding the type into a new struct that operates directly on `common.RawBytes` and just provides methods for raw `ringbuf` operations would make code a lot cleaner. We could have a `FreePkts` type and a `PayloadPkts` type. The `FreePkts` type would handle capacity resets internally. The `PayloadPkts` can also provide a `Write` for only one element so we avoid clunkier syntax like `ringbuf.EntryList{buf}`.

---

*[go/sig/base/egress_dispatcher.go, line 39 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvovLb1kiD-hBPyeuO:-KvvovLb1kiD-hBPyeuP:b55urwo) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_dispatcher.go#L39)):*
> ```Go
> func Init() {
> 	egressFreePkts = ringbuf.New(egressFreePktsCap, func() interface{} {
> 		return make(common.RawBytes, 1<<16)
> ```

Use `common.MaxMTU` instead of `1<<16` here.

---

*[go/sig/base/egress_dispatcher.go, line 56 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvvp4xSk7wmmnIhp7Hz:-Kvvp4xSk7wmmnIhp7I-:b-ob83rl) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_dispatcher.go#L56)):*
> ```Go
> func (ed *egressDispatcher) Run() {
> 	defer liblog.LogPanicAndExit()
> 	bufs := make(ringbuf.EntryList, 32)
> ```

Change 32 to a const definition that's near `egressFreePktsCap`.

Also, there is a corner case here if the number of frames an EgressDispatcher reserves is over `egressFreePktsCap`/`number of dispatchers`, as reservations made by some EgressDispatchers that are not receiving data can starve a different one. For example, if we have 4 frames, 3 dispatchers and each one reserves 2 without actually getting traffic, the third one will block while trying to get frames.

One solution would be to have `egressFreePkts.Read(bufs, true)` never grab more than `egreeFreePktsCap`/`number of known ASes`. Better solutions need peeking into the ringbuffer which gets too complex for now.

---

*[go/sig/base/egress_dispatcher.go, line 71 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvqPzLonZLVKWMp4oa:-KvvqPzLonZLVKWMp4ob:b-ttfb4w) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_dispatcher.go#L71)):*
> ```Go
> 		for i := 0; i < n; i++ {
> 			buf := bufs[i].(common.RawBytes)
> 			bufs[i] = nil
> ```

Does this assignment do anything? `bufs[i]` is never read before the next call to `egressFrePkts.Read` which overwrites it.

---

*[go/sig/base/egress_dispatcher.go, line 72 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvrTox0D85RAY6YWMo:-KvvrTox0D85RAY6YWMp:b-s5knzg) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_dispatcher.go#L72)):*
> ```Go
> 			buf := bufs[i].(common.RawBytes)
> 			bufs[i] = nil
> 			buf = buf[:cap(buf)]
> ```

Resetting here to `MaxMTU` instead of `cap` would cause a panic whenever a buffer is mishandled somewhere else in the code and replaced with a different one with lower capacity. I'm not sure if this would be helpful in any way, just throwing the suggestion out there.

---

*[go/sig/base/egress_worker.go, line 34 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvsCOnIaEYNHUyRh7B:-KvvsCOnIaEYNHUyRh7C:bbjwwez) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_worker.go#L34)):*
> ```Go
> )
> 
> //   SIG Frame Header, used to encapsulate SIG to SIG traffic. The sequence
> ```

This comment should also include how the frame is organized internally (i.e., a 2-byte network byte order length precedes each packet in the frame, 8 byte alignment). This will also help clarity when we move to the new version which will change header fields and alignment rules. I had to check the code to understand which version we're currently running.

---

*[go/sig/base/egress_worker.go, line 65 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvvT1i-c2a-zcvRllR:-KvvvT1i-c2a-zcvRllS:b-evavf) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_worker.go#L65)):*
> ```Go
> func NewEgressWorker(ae *ASEntry, pol *PathPolicy, conn *snet.Conn) *EgressWorker {
> 	return &EgressWorker{ae: ae, pol: pol, conn: conn, currPath: pol.CurrPath(),
> 		pkts: make(ringbuf.EntryList, 0, 32)}
> ```

Replace magic value with a const (could probably reuse the one to be added in `egress_dispatcher.go`).

---

*[go/sig/base/egress_worker.go, line 75 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvvymc1R5f9PWqpogbr:-Kvvymc1R5f9PWqpogbs:b6uom87) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_worker.go#L75)):*
> ```Go
> 	for {
> 		// If the frame is empty, block indefinitely for more packets.
> 		fEmpty := f.offset == sigcmn.SIGHdrSize
> ```

Wait, on the first loop doesn't `fEmpty` evaluate to `false`? Since `f.offset == 0` after `newFrame`. This means `e.Read()` below doesn't block and busy loops on `TopLoop` until there's data?

---

*[go/sig/base/egress_worker.go, line 80 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvw0RhLEIIBod9hZMa4:-Kvw0RhLEIIBod9hZMa5:bnzntrz) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_worker.go#L80)):*
> ```Go
> 		}
> 		if fEmpty {
> 			// Cover the case where no packets have arrived in a while, and the
> ```

I am not sure I understand this comment. The path is refreshed below in `resetFrame`, but I do not see the logical connection to "no packets have arrived in a while". We can still reach this code if we start from a new empty frame and the blocking read got something immediately.

---

*[go/sig/base/egress_worker.go, line 100 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvw1OVeLLxY2uzq5FgT:-Kvw1OVeLLxY2uzq5FgU:baug53m) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_worker.go#L100)):*
> ```Go
> 		egressFreePkts.Write(e.pkts, true)
> 		for i := range e.pkts {
> 			e.pkts[i] = nil
> ```

Does removing references here do anything? They are overwritten on the next `Read` anyway (and it's not like that memory is ever GC'd).

---

*[go/sig/base/pathmgmt.go, line 46 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrNIEBhIbN37RmGCIv:-Kvw4fBZxJqXSgvK1siu:bjaggik) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/pathmgmt.go#L46)):*

I was hoping this would inspire you to come up with something better. I have to admit I'm struggling finding a better term.
</blockquote></details>

Since we have one per `EgressWorker`, maybe `EgressEntry`, `EgressInfo`, `EgressState`?

---

*[go/sig/base/pathmgmt.go, line 50 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvw4s4_XRq8CCZKmssz:-Kvw4s4_XRq8CCZKmst-:b-eh4kv3) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/pathmgmt.go#L50)):*
> ```Go
> 	policy interface{} // FIXME(kormat): not implemented yet :P
> 	//Policy  pathmgr.Policy // Describes what interfaces to route through
> 	pool     *pathmgr.SyncPaths // pool of paths that meet the policy requirement, managed by pathmgr
> ```

Our convention is to put the comments above the line.

---

*[go/sig/base/pathmgmt.go, line 61 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvw5F2GjUwuuN0ILvJC:-Kvw5F2GjUwuuN0ILvJD:b-yl56mo) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/pathmgmt.go#L61)):*
> ```Go
> 	var err error
> 	pp := &PathPolicy{IA: dstIA, policy: policy}
> 	pp.pool, err = sigcmn.PathMgr.Register(sigcmn.IA, dstIA)
> ```

Add a small FIXME to change this to `RegisterFilter` after we merge the classifier, just so we don't forget since we won't see a build error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1275)*

kormat:
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[go/sig/base/egress_dispatcher.go, line 34 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvwE0PD2wNc0cBfQFT:-KvwCtFgCd1W1WTlnfGq:bjsft8u) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_dispatcher.go#L34)):*

One thing I noticed throughout the code is that there quite some boilerplate relating to casting and initialization relating to `ringbuf.Ring` (`EntryList` initialization, casts to `common.RawBytes`, etc.). Embedding the type into a new struct that operates directly on `common.RawBytes` and just provides methods for raw `ringbuf` operations would make code a lot cleaner. We could have a `FreePkts` type and a `PayloadPkts` type. The `FreePkts` type would handle capacity resets internally. The `PayloadPkts` can also provide a `Write` for only one element so we avoid clunkier syntax like `ringbuf.EntryList{buf}`.
</blockquote></details>

Yeah, i was also vaguely pondering something like that. It seems like a good idea. I'm going to leave it on the "future work" list though, for now.

---

*[go/sig/base/egress_dispatcher.go, line 39 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvovLb1kiD-hBPyeuO:-Kvw8bLB04jauNZvB7-i:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_dispatcher.go#L39)):*

Use `common.MaxMTU` instead of `1<<16` here.
</blockquote></details>

Done.

---

*[go/sig/base/egress_dispatcher.go, line 56 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvvp4xSk7wmmnIhp7Hz:-Kvw9OePa1p1fr0Zh92Y:bk7wxp6) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_dispatcher.go#L56)):*

Change 32 to a const definition that's near `egressFreePktsCap`.

Also, there is a corner case here if the number of frames an EgressDispatcher reserves is over `egressFreePktsCap`/`number of dispatchers`, as reservations made by some EgressDispatchers that are not receiving data can starve a different one. For example, if we have 4 frames, 3 dispatchers and each one reserves 2 without actually getting traffic, the third one will block while trying to get frames.

One solution would be to have `egressFreePkts.Read(bufs, true)` never grab more than `egreeFreePktsCap`/`number of known ASes`. Better solutions need peeking into the ringbuffer which gets too complex for now.
</blockquote></details>

Another would be to dynamically change how much each dispatcher requests. Anyway, keeping this number small relative to the size of the free pkts ringbuf is "good enough", for now. I've added a comment to the `const`'s, as a reminder.

---

*[go/sig/base/egress_dispatcher.go, line 71 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvqPzLonZLVKWMp4oa:-Kvw9cvh7lcFd__11SVi:biu7agb) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_dispatcher.go#L71)):*

Does this assignment do anything? `bufs[i]` is never read before the next call to `egressFrePkts.Read` which overwrites it.
</blockquote></details>

In this particular case, no. I'm doing it because we ran into very-hard-to-debug memory leak issues in the border router in a _similar_ situation, so by default i'm now making sure to remove references. If we ever changed the code such that new buffers are occasionally created/destroyed, this will prevent the same issue.

---

*[go/sig/base/egress_dispatcher.go, line 72 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvrTox0D85RAY6YWMo:-KvwDL6JpPrl1-5YDXRp:bz8z4nb) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_dispatcher.go#L72)):*

Resetting here to `MaxMTU` instead of `cap` would cause a panic whenever a buffer is mishandled somewhere else in the code and replaced with a different one with lower capacity. I'm not sure if this would be helpful in any way, just throwing the suggestion out there.
</blockquote></details>

We could, but i'm coming down on the side of not doing that (for now, at least), as that seems to be an unlikely error source, and the current approach avoids hard-coding the buffer size in an extra place.

---

*[go/sig/base/egress_worker.go, line 65 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvvT1i-c2a-zcvRllR:-KvwB2tOS5agOWHDnQ2e:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_worker.go#L65)):*

Replace magic value with a const (could probably reuse the one to be added in `egress_dispatcher.go`).
</blockquote></details>

Done.

---

*[go/sig/base/egress_worker.go, line 75 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvvymc1R5f9PWqpogbr:-KvwB7LaZ1Nh49XA3fpc:bwj81h9) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_worker.go#L75)):*

Wait, on the first loop doesn't `fEmpty` evaluate to `false`? Since `f.offset == 0` after `newFrame`. This means `e.Read()` below doesn't block and busy loops on `TopLoop` until there's data?
</blockquote></details>

Good catch, fixed in `newFrame`.

---

*[go/sig/base/egress_worker.go, line 80 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvw0RhLEIIBod9hZMa4:-KvwBBQqo0kl1ZtcN3ji:b-i67liw) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_worker.go#L80)):*

I am not sure I understand this comment. The path is refreshed below in `resetFrame`, but I do not see the logical connection to "no packets have arrived in a while". We can still reach this code if we start from a new empty frame and the blocking read got something immediately.
</blockquote></details>

If there had been traffic in the past, the cached path could be stale. This check ensures that we update the path when it's safe to do so. Yes, it's more frequent than strictly necessary, but the operation is cheap in any case.

---

*[go/sig/base/egress_worker.go, line 100 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvw1OVeLLxY2uzq5FgT:-KvwBdzAnFb1OuStFQkg:byvujft) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_worker.go#L100)):*

Does removing references here do anything? They are overwritten on the next `Read` anyway (and it's not like that memory is ever GC'd).
</blockquote></details>

See response in `egress_dispatcher.go`

---

*[go/sig/base/pathmgmt.go, line 46 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvrNIEBhIbN37RmGCIv:-KvwCbvmkQJru-4YNpCI:b-o5ib1q) ([raw file](https://github.com/netsec-ethz/scion/blob/6be992984d0f7ed7980c5128ce10b55aae4c1eae/go/sig/base/pathmgmt.go#L46)):*

Since we have one per `EgressWorker`, maybe `EgressEntry`, `EgressInfo`, `EgressState`?
</blockquote></details>

@scrye - ah hah. Yeah, something like that works. I'll see about doing the rename in the next PR.

---

*[go/sig/base/pathmgmt.go, line 50 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvw4s4_XRq8CCZKmssz:-KvwC1xq0IRGYDRn0n9i:b-mbrq2) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/pathmgmt.go#L50)):*

Our convention is to put the comments above the line.
</blockquote></details>

Fixed in the PR i'm currently working on (i'd rather not change it here and then have to deal with merge conflicts when rebasing).

---

*[go/sig/base/pathmgmt.go, line 61 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-Kvw5F2GjUwuuN0ILvJC:-KvwCaGpqI-7hoBcVIua:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/pathmgmt.go#L61)):*

Add a small FIXME to change this to `RegisterFilter` after we merge the classifier, just so we don't forget since we won't see a build error.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1275)*

kormat:
Review status: 7 of 10 files reviewed at latest revision, 3 unresolved discussions.

---

*[go/sig/base/egress_worker.go, line 34 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-KvvsCOnIaEYNHUyRh7B:-KvwFqtqD7ic2L2fJ3ij:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5ed45e87c3fb841366613ed4f7af36579f7320c8/go/sig/base/egress_worker.go#L34)):*

This comment should also include how the frame is organized internally (i.e., a 2-byte network byte order length precedes each packet in the frame, 8 byte alignment). This will also help clarity when we move to the new version which will change header fields and alignment rules. I had to check the code to understand which version we're currently running.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1275)*

scrye:
:lgtm:

---

Review status: 7 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1275#-:-KvwG2N-x9aYqjKOLbHN:bnfp4nl)*

scrye:
Reviewed 3 of 3 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1275)*

PR #383, opened 2015-09-23
New types of PathMgmtPacket

pszal:
- replacing `RECORDS` type by three new types (`REPLY, SYNC, REG`)
[...]

PR #1280, opened 2017-10-10
Make SCIOND return multiple paths and not just the first one

shitz:
Also:
PathReplies now include the PathRequest they are responding to.
PathRecordsSync is now used to propagate path segments to other core PSes and to the master PS.
Fix a race condition in path request handling

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1280)

### DIFF
--- a/go/lib/ctrl/path_mgmt/seg_recs.go
+++ b/go/lib/ctrl/path_mgmt/seg_recs.go
@@ -48,12 +48,6 @@ func (s *SegRecs) String() string {
 	return strings.Join(desc, "\n")
 }
 
-var _ proto.Cerealizable = (*SegReply)(nil)
-
-type SegReply struct {
-	*SegRecs
-}
-
 var _ proto.Cerealizable = (*SegReg)(nil)
 
 type SegReg struct {
--- a/go/lib/ctrl/path_mgmt/seg_reply.go
+++ b/go/lib/ctrl/path_mgmt/seg_reply.go
@@ -1,0 +1,48 @@
+// Copyright 2017 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file contains the Go representation of segment replies.
+
+package path_mgmt
+
+import (
+	"fmt"
+
+	"github.com/netsec-ethz/scion/go/lib/common"
+	"github.com/netsec-ethz/scion/go/proto"
+)
+
+var _ proto.Cerealizable = (*SegReply)(nil)
+
+type SegReply struct {
+	Req  *SegReq
+	Recs *SegRecs
+}
+
+func NewSegReplyFromRaw(b common.RawBytes) (*SegReply, error) {
+	s := &SegReply{}
+	return s, proto.ParseFromRaw(s, s.ProtoId(), b)
+}
+
+func (s *SegReply) ProtoId() proto.ProtoIdType {
+	return proto.SegReply_TypeID
+}
+
+func (s *SegReply) Write(b common.RawBytes) (int, error) {
+	return proto.WriteRoot(s, b)
+}
+
+func (s *SegReply) String() string {
+	return fmt.Sprintf("Req: %s Reply:\n%s", s.Req, s.Recs)
+}
--- a/go/lib/ctrl/path_mgmt/seg_req.go
+++ b/go/lib/ctrl/path_mgmt/seg_req.go
@@ -24,10 +24,16 @@ import (
 	"github.com/netsec-ethz/scion/go/proto"
 )
 
+type SegReqId uint64
+
+func (id SegReqId) String() string {
+	return fmt.Sprintf("%016x", uint64(id))
+}
+
 var _ proto.Cerealizable = (*SegReq)(nil)
 
 type SegReq struct {
-	Id       uint64
+	Id       SegReqId
 	RawSrcIA addr.IAInt `capnp:"srcIA"`
 	RawDstIA addr.IAInt `capnp:"dstIA"`
 	Flags    struct {
@@ -58,5 +64,5 @@ func (s *SegReq) Write(b common.RawBytes) (int, error) {
 }
 
 func (s *SegReq) String() string {
-	return fmt.Sprintf("Id: %16x %s -> %s, Flags: %v", s.Id, s.SrcIA(), s.DstIA(), s.Flags)
+	return fmt.Sprintf("Id: %s %s -> %s, Flags: %v", s.Id, s.SrcIA(), s.DstIA(), s.Flags)
 }
--- a/go/lib/ctrl/path_mgmt/seg_req.go
+++ b/go/lib/ctrl/path_mgmt/seg_req.go
@@ -58,27 +58,5 @@ func (s *SegReq) Write(b common.RawBytes) (int, error) {
 }
 
 func (s *SegReq) String() string {
-	return fmt.Sprintf("Id: %08x %v -> %v, Flags: %v", s.Id, s.SrcIA(), s.DstIA(), s.Flags)
-}
-
-type SegReply struct {
-	Id   uint64
-	Recs *SegRecs
-}
-
-func NewSegReplyFromRaw(b common.RawBytes) (*SegReply, error) {
-	s := &SegReply{}
-	return s, proto.ParseFromRaw(s, s.ProtoId(), b)
-}
-
-func (s *SegReply) ProtoId() proto.ProtoIdType {
-	return proto.SegReply_TypeID
-}
-
-func (s *SegReply) Write(b common.RawBytes) (int, error) {
-	return proto.WriteRoot(s, b)
-}
-
-func (s *SegReply) String() string {
-	return fmt.Sprintf("Id: %08x\n%s", s.Id, s.Recs)
+	return fmt.Sprintf("Id: %16x %s -> %s, Flags: %v", s.Id, s.SrcIA(), s.DstIA(), s.Flags)
 }
--- a/go/lib/ctrl/path_mgmt/seg_req.go
+++ b/go/lib/ctrl/path_mgmt/seg_req.go
@@ -27,6 +27,7 @@ import (
 var _ proto.Cerealizable = (*SegReq)(nil)
 
 type SegReq struct {
+	Id       uint64
 	RawSrcIA addr.IAInt `capnp:"srcIA"`
 	RawDstIA addr.IAInt `capnp:"dstIA"`
 	Flags    struct {
@@ -57,5 +58,27 @@ func (s *SegReq) Write(b common.RawBytes) (int, error) {
 }
 
 func (s *SegReq) String() string {
-	return fmt.Sprintf("SrcIA: %v, DstIA: %d, Flags: %v", s.SrcIA(), s.DstIA(), s.Flags)
+	return fmt.Sprintf("Id: %08x %v -> %v, Flags: %v", s.Id, s.SrcIA(), s.DstIA(), s.Flags)
+}
+
+type SegReply struct {
+	Id   uint64
+	Recs *SegRecs
+}
+
+func NewSegReplyFromRaw(b common.RawBytes) (*SegReply, error) {
+	s := &SegReply{}
+	return s, proto.ParseFromRaw(s, s.ProtoId(), b)
+}
+
+func (s *SegReply) ProtoId() proto.ProtoIdType {
+	return proto.SegReply_TypeID
+}
+
+func (s *SegReply) Write(b common.RawBytes) (int, error) {
+	return proto.WriteRoot(s, b)
+}
+
+func (s *SegReply) String() string {
+	return fmt.Sprintf("Id: %08x\n%s", s.Id, s.Recs)
 }
--- a/integration/revocation_test.sh
+++ b/integration/revocation_test.sh
@@ -43,7 +43,7 @@ if [ ${result} -ne 0 ]; then
     exit ${result}
 fi
 # Bring down routers.
-SLEEP=10
+SLEEP=3
 log "Stopping routers and waiting for ${SLEEP}s."
 ./supervisor/supervisor.sh stop "$@"
 if [ $? -ne 0 ]; then
--- a/integration/revocation_test.sh
+++ b/integration/revocation_test.sh
@@ -43,7 +43,7 @@ if [ ${result} -ne 0 ]; then
     exit ${result}
 fi
 # Bring down routers.
-SLEEP=3
+SLEEP=4
 log "Stopping routers and waiting for ${SLEEP}s."
 ./supervisor/supervisor.sh stop "$@"
 if [ $? -ne 0 ]; then
--- a/proto/path_mgmt.capnp
+++ b/proto/path_mgmt.capnp
@@ -24,7 +24,7 @@ struct SegRecs {
 }
 
 struct SegReply {
-    id @0 :UInt64;
+    req @0 :SegReq;
     recs @1 :SegRecs;
 }
 
--- a/proto/path_mgmt.capnp
+++ b/proto/path_mgmt.capnp
@@ -8,12 +8,14 @@ using IFState = import "if_state.capnp";
 using RevInfo = import "rev_info.capnp";
 
 struct SegReq {
-    srcIA @0 :UInt32;
-    dstIA @1 :UInt32;
+    id @0 :UInt64;  # Request ID
+    srcIA @1 :UInt32;
+    dstIA @2 :UInt32;
     flags :group {
-        sibra @2 :Bool;
-        cacheOnly @3 :Bool;
+        sibra @3 :Bool;
+        cacheOnly @4 :Bool;
     }
+
 }
 
 struct SegRecs {
@@ -21,11 +23,16 @@ struct SegRecs {
     revInfos @1 :List(RevInfo.RevInfo);
 }
 
+struct SegReply {
+    id @0 :UInt64;
+    recs @1 :SegRecs;
+}
+
 struct PathMgmt {
     union {
         unset @0 :Void;
         segReq @1 :SegReq;
-        segReply @2 :SegRecs;
+        segReply @2 :SegReply;
         segReg @3 :SegRecs;
         segSync @4 :SegRecs;
         revInfo @5 :RevInfo.RevInfo;
--- a/python/lib/packet/path_mgmt/base.py
+++ b/python/lib/packet/path_mgmt/base.py
@@ -24,12 +24,8 @@ from lib.packet.packet_base import CerealBox
 from lib.types import PathMgmtType
 from lib.packet.path_mgmt.ifstate import IFStatePayload, IFStateRequest
 from lib.packet.path_mgmt.rev_info import RevocationInfo
-from lib.packet.path_mgmt.seg_recs import (
-    PathRecordsReg,
-    PathRecordsReply,
-    PathRecordsSync,
-)
-from lib.packet.path_mgmt.seg_req import PathSegmentReq
+from lib.packet.path_mgmt.seg_recs import PathRecordsReg, PathRecordsSync
+from lib.packet.path_mgmt.seg_req import PathSegmentReq, PathSegmentReply
 
 
 class PathMgmt(CerealBox):  # pragma: no cover
@@ -37,7 +33,7 @@ class PathMgmt(CerealBox):  # pragma: no cover
     P_CLS = P.PathMgmt
     CLASS_FIELD_MAP = {
         PathSegmentReq: PathMgmtType.REQUEST,
-        PathRecordsReply: PathMgmtType.REPLY,
+        PathSegmentReply: PathMgmtType.REPLY,
         PathRecordsReg: PathMgmtType.REG,
         PathRecordsSync: PathMgmtType.SYNC,
         RevocationInfo: PathMgmtType.REVOCATION,
--- a/python/lib/packet/path_mgmt/seg_recs.py
+++ b/python/lib/packet/path_mgmt/seg_recs.py
@@ -86,10 +86,6 @@ class PathSegmentRecords(Cerealizable):  # pragma: no cover
         return "\n".join(s)
 
 
-class PathRecordsReply(PathSegmentRecords):
-    NAME = "PathRecordsReply"
-
-
 class PathRecordsReg(PathSegmentRecords):
     NAME = "PathRecordsReg"
 
--- a/python/lib/packet/path_mgmt/seg_recs.py
+++ b/python/lib/packet/path_mgmt/seg_recs.py
@@ -31,6 +31,7 @@ class PathSegmentRecords(Cerealizable):  # pragma: no cover
     Path Record class used for sending list of down/up-paths. Paths are
     represented as objects of the PathSegment class.
     """
+    NAME = "PathSegmentRecords"
     P_CLS = P.SegRecs
 
     @classmethod
--- a/python/lib/packet/path_mgmt/seg_req.py
+++ b/python/lib/packet/path_mgmt/seg_req.py
@@ -68,7 +68,8 @@ class PathSegmentReq(Cerealizable):  # pragma: no cover
         return self.p.id
 
     def short_desc(self):
-        return "Id: %16x %s -> %s  %s" % (self.req_id(), self.src_ia(), self.dst_ia(), self.flags())
+        return "Id: %016x %s -> %s  %s" % (
+            self.req_id(), self.src_ia(), self.dst_ia(), self.flags())
 
 
 class PathSegmentReply(Cerealizable):  # pragma: no cover
--- a/python/lib/packet/path_mgmt/seg_req.py
+++ b/python/lib/packet/path_mgmt/seg_req.py
@@ -22,6 +22,7 @@ import capnp  # noqa
 import proto.path_mgmt_capnp as P
 from lib.defines import PATH_FLAG_CACHEONLY, PATH_FLAG_SIBRA
 from lib.packet.packet_base import Cerealizable
+from lib.packet.path_mgmt.seg_recs import PathSegmentRecords
 from lib.packet.scion_addr import ISD_AS
 
 
@@ -32,15 +33,19 @@ class PathSegmentReq(Cerealizable):  # pragma: no cover
     P_CLS = P.SegReq
 
     @classmethod
-    def from_values(cls, src_ia, dst_ia, flags=None):
+    def from_values(cls, req_id, src_ia, dst_ia, flags=None):
         if not flags:
             flags = set()
-        p = cls.P_CLS.new_message(srcIA=int(src_ia), dstIA=int(dst_ia))
+        p = cls.P_CLS.new_message(
+            id=req_id, srcIA=int(src_ia), dstIA=int(dst_ia))
         if PATH_FLAG_SIBRA in flags:
             p.flags.sibra = True
         if PATH_FLAG_CACHEONLY in flags:
             p.flags.cacheOnly = True
         return cls(p)
+
+    def req_id(self):
+        return self.p.id
 
     def src_ia(self):
         return ISD_AS(self.p.srcIA)
@@ -57,4 +62,23 @@ class PathSegmentReq(Cerealizable):  # pragma: no cover
         return tuple(flags)
 
     def short_desc(self):
-        return "%s -> %s %s" % (self.src_ia(), self.dst_ia(), self.flags())
+        return "Id: %08x %s -> %s  %s" % (self.req_id(), self.src_ia(), self.dst_ia(), self.flags())
+
+
+class PathSegmentReply(Cerealizable):  # pragma: no cover
+    NAME = "PathSegmentReply"
+    P_CLS = P.SegReply
+
+    @classmethod
+    def from_values(cls, req_id, recs):
+        p = cls.P_CLS.new_message(id=req_id, recs=recs.p)
+        return cls(p)
+
+    def req_id(self):
+        return self.p.id
+
+    def recs(self):
+        return PathSegmentRecords(self.p.recs)
+
+    def short_desc(self):
+        return "Id: %08x\n%s" % (self.req_id(), self.recs())
--- a/python/lib/packet/path_mgmt/seg_req.py
+++ b/python/lib/packet/path_mgmt/seg_req.py
@@ -61,6 +61,12 @@ class PathSegmentReq(Cerealizable):  # pragma: no cover
             flags.add(PATH_FLAG_CACHEONLY)
         return tuple(flags)
 
+    def __eq__(self, other):
+        return self.p.id == other.p.id
+
+    def __hash__(self):
+        return self.p.id
+
     def short_desc(self):
         return "Id: %08x %s -> %s  %s" % (self.req_id(), self.src_ia(), self.dst_ia(), self.flags())
 
@@ -70,15 +76,15 @@ class PathSegmentReply(Cerealizable):  # pragma: no cover
     P_CLS = P.SegReply
 
     @classmethod
-    def from_values(cls, req_id, recs):
-        p = cls.P_CLS.new_message(id=req_id, recs=recs.p)
+    def from_values(cls, req, recs):
+        p = cls.P_CLS.new_message(req=req.p, recs=recs.p)
         return cls(p)
 
-    def req_id(self):
-        return self.p.id
+    def req(self):
+        return PathSegmentReq(self.p.req)
 
     def recs(self):
         return PathSegmentRecords(self.p.recs)
 
     def short_desc(self):
-        return "Id: %08x\n%s" % (self.req_id(), self.recs())
+        return "Req: %s\n%s" % (self.req(), self.recs())
--- a/python/lib/packet/path_mgmt/seg_req.py
+++ b/python/lib/packet/path_mgmt/seg_req.py
@@ -62,7 +62,7 @@ class PathSegmentReq(Cerealizable):  # pragma: no cover
 
     def __eq__(self, other):
         return (self.p.id == other.p.id and self.p.srcIA == other.p.srcIA and
-            self.p.dstIA == other.p.dstIA and self.flags() == other.flags())
+                self.p.dstIA == other.p.dstIA and self.flags() == other.flags())
 
     def __hash__(self):
         return self.p.id
--- a/python/lib/packet/path_mgmt/seg_req.py
+++ b/python/lib/packet/path_mgmt/seg_req.py
@@ -29,7 +29,6 @@ from lib.packet.scion_addr import ISD_AS
 class PathSegmentReq(Cerealizable):  # pragma: no cover
     """Describes a request for path segment(s)"""
     NAME = "PathSegmentReq"
-    LEN = 1 + 2 * ISD_AS.LEN
     P_CLS = P.SegReq
 
     @classmethod
@@ -62,13 +61,14 @@ class PathSegmentReq(Cerealizable):  # pragma: no cover
         return tuple(flags)
 
     def __eq__(self, other):
-        return self.p.id == other.p.id
+        return (self.p.id == other.p.id and self.p.srcIA == other.p.srcIA and
+            self.p.dstIA == other.p.dstIA and self.flags() == other.flags())
 
     def __hash__(self):
         return self.p.id
 
     def short_desc(self):
-        return "Id: %08x %s -> %s  %s" % (self.req_id(), self.src_ia(), self.dst_ia(), self.flags())
+        return "Id: %16x %s -> %s  %s" % (self.req_id(), self.src_ia(), self.dst_ia(), self.flags())
 
 
 class PathSegmentReply(Cerealizable):  # pragma: no cover
--- a/python/lib/util.py
+++ b/python/lib/util.py
@@ -21,6 +21,7 @@ Various utilities for SCION functionality.
 import json
 import logging
 import os
+import random
 import shutil
 import signal
 import sys
@@ -274,6 +275,11 @@ def recv_all(sock, total_len, flags):
             return None
         barr += buf
     return bytes(barr)
+
+
+def random_uint64():
+    """Returns a random number in the range [0, 2**64 - 1]"""
+    return random.randint(0, 2**64 - 1)
 
 
 class SCIONTime(object):
--- a/python/path_server/base.py
+++ b/python/path_server/base.py
@@ -64,8 +64,7 @@ from scion_elem.scion_elem import SCIONElement
 
 # Exported metrics.
 REQS_TOTAL = Counter("ps_reqs_total", "# of path requests", ["server_id", "isd_as"])
-REQS_PENDING = Gauge("ps_req_pending_total",
-                     "# of pending path requests", ["server_id", "isd_as"])
+REQS_PENDING = Gauge("ps_req_pending_total", "# of pending path requests", ["server_id", "isd_as"])
 SEGS_TO_ZK = Gauge("ps_segs_to_zk_total", "# of path segments to ZK", ["server_id", "isd_as"])
 REVS_TO_ZK = Gauge("ps_revs_to_zk_total", "# of revocations to ZK", ["server_id", "isd_as"])
 HT_ROOT_MAPPTINGS = Gauge("ps_ht_root_mappings_total", "# of hashtree root to segment mappings",
@@ -539,8 +538,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
     def _zk_write(self, data):
         hash_ = crypto_hash(data).hex()
         try:
-            self.path_cache.store("%s-%s" %
-                                  (hash_, SCIONTime.get_time()), data)
+            self.path_cache.store("%s-%s" % (hash_, SCIONTime.get_time()), data)
         except ZkNoConnection:
             logging.warning("Unable to store segment(s) in shared path: "
                             "no connection to ZK")
@@ -561,7 +559,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
         # Create new formatter to include the request in the log.
         formatter = formatter = Rfc3339Formatter(
             "%(asctime)s [%(levelname)s] (%(threadName)s) %(message)s "
-            "{req=%(req)s, from=%(from)s}")
+            "{id=%(id)s, from=%(from)s}")
         add_formatter('RequestLogger', formatter)
 
     def get_request_logger(self, req, meta):
@@ -571,7 +569,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
         # Create a logger for the request to log with context.
         return logging.LoggerAdapter(
             self._request_logger,
-            {"req": req.short_desc(), "from": str(meta)})
+            {"id": req.req_id(), "from": str(meta)})
 
     def _init_metrics(self):
         super()._init_metrics()
--- a/python/path_server/base.py
+++ b/python/path_server/base.py
@@ -336,7 +336,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
             {PST.UP: up, PST.CORE: core, PST.DOWN: down},
             revs_to_add
         )
-        pld = PathSegmentReply.from_values(req.req_id(), recs)
+        pld = PathSegmentReply.from_values(req.copy(), recs)
         self.send_meta(CtrlPayload(PathMgmt(pld)), meta)
         logger.info("Sending PATH_REPLY with %d segment(s).", len(all_segs))
 
--- a/python/path_server/base.py
+++ b/python/path_server/base.py
@@ -63,14 +63,11 @@ from scion_elem.scion_elem import SCIONElement
 
 
 # Exported metrics.
-REQS_TOTAL = Counter("ps_reqs_total", "# of path requests", [
-                     "server_id", "isd_as"])
+REQS_TOTAL = Counter("ps_reqs_total", "# of path requests", ["server_id", "isd_as"])
 REQS_PENDING = Gauge("ps_req_pending_total",
                      "# of pending path requests", ["server_id", "isd_as"])
-SEGS_TO_ZK = Gauge("ps_segs_to_zk_total", "# of path segments to ZK", [
-                   "server_id", "isd_as"])
-REVS_TO_ZK = Gauge("ps_revs_to_zk_total", "# of revocations to ZK", [
-                   "server_id", "isd_as"])
+SEGS_TO_ZK = Gauge("ps_segs_to_zk_total", "# of path segments to ZK", ["server_id", "isd_as"])
+REVS_TO_ZK = Gauge("ps_revs_to_zk_total", "# of revocations to ZK", ["server_id", "isd_as"])
 HT_ROOT_MAPPTINGS = Gauge("ps_ht_root_mappings_total", "# of hashtree root to segment mappings",
                           ["server_id", "isd_as"])
 IS_MASTER = Gauge("ps_is_master", "true if this process is the replication master",
@@ -103,25 +100,19 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
         :param str prom_export: prometheus export address.
         """
         super().__init__(server_id, conf_dir, prom_export=prom_export)
-        down_labels = {**self._labels,
-                       "type": "down"} if self._labels else None
-        core_labels = {**self._labels,
-                       "type": "core"} if self._labels else None
-        self.down_segments = PathSegmentDB(
-            max_res_no=self.MAX_SEG_NO, labels=down_labels)
-        self.core_segments = PathSegmentDB(
-            max_res_no=self.MAX_SEG_NO, labels=core_labels)
+        down_labels = {**self._labels, "type": "down"} if self._labels else None
+        core_labels = {**self._labels, "type": "core"} if self._labels else None
+        self.down_segments = PathSegmentDB(max_res_no=self.MAX_SEG_NO, labels=down_labels)
+        self.core_segments = PathSegmentDB(max_res_no=self.MAX_SEG_NO, labels=core_labels)
         # Dict of pending requests.
-        self.pending_req = defaultdict(
-            lambda: ExpiringDict(1000, PATH_REQ_TOUT))
+        self.pending_req = defaultdict(lambda: ExpiringDict(1000, PATH_REQ_TOUT))
         self.pen_req_lock = threading.Lock()
         self._request_logger = None
         # Used when l/cPS doesn't have up/dw-path.
         self.waiting_targets = defaultdict(list)
         self.revocations = RevCache(labels=self._labels)
         # A mapping from (hash tree root of AS, IFID) to segments
-        self.htroot_if2seg = ExpiringDict(
-            1000, self.config.revocation_tree_ttl)
+        self.htroot_if2seg = ExpiringDict(1000, self.config.revocation_tree_ttl)
         self.htroot_if2seglock = Lock()
         self.CTRL_PLD_CLASS_MAP = {
             PayloadClass.PATH: {
@@ -242,8 +233,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
         assert isinstance(infos, IFStatePayload), type(infos)
         for info in infos.iter_infos():
             if not info.p.active and info.p.revInfo:
-                self._handle_revocation(CtrlPayload(
-                    PathMgmt(info.rev_info())), meta)
+                self._handle_revocation(CtrlPayload(PathMgmt(info.rev_info())), meta)
 
     def _handle_scmp_revocation(self, pld, meta):
         rev_info = RevocationInfo.from_raw(pld.info.rev_info)
@@ -509,8 +499,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
         """
         dst_ia = pcb.first_ia()
         if not self.is_core_as(dst_ia):
-            logging.warning(
-                "Invalid waiting target, not a core AS: %s", dst_ia)
+            logging.warning("Invalid waiting target, not a core AS: %s", dst_ia)
             return
         self._send_waiting_queries(dst_ia[0], pcb)
 
@@ -522,8 +511,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
         src_ia = pcb.first_ia()
         while targets:
             (seg_req, logger) = targets.pop(0)
-            meta = self._build_meta(
-                ia=src_ia, path=path, host=SVCType.PS_A, reuse=True)
+            meta = self._build_meta(ia=src_ia, path=path, host=SVCType.PS_A, reuse=True)
             self.send_meta(CtrlPayload(PathMgmt(seg_req)), meta)
             logger.info("Waiting request (%s) sent to %s via %s",
                         seg_req.short_desc(), meta, pcb.short_desc())
@@ -570,8 +558,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
         Initializes the request logger.
         """
         self._request_logger = logging.getLogger("RequestLogger")
-        # Create new formatter to include the random request id and the request
-        # in the log.
+        # Create new formatter to include the request in the log.
         formatter = formatter = Rfc3339Formatter(
             "%(asctime)s [%(levelname)s] (%(threadName)s) %(message)s "
             "{req=%(req)s, from=%(from)s}")
--- a/python/path_server/base.py
+++ b/python/path_server/base.py
@@ -41,7 +41,8 @@ from lib.packet.ctrl_pld import CtrlPayload
 from lib.packet.path_mgmt.base import PathMgmt
 from lib.packet.path_mgmt.ifstate import IFStatePayload
 from lib.packet.path_mgmt.rev_info import RevocationInfo
-from lib.packet.path_mgmt.seg_recs import PathRecordsReply, PathSegmentRecords
+from lib.packet.path_mgmt.seg_req import PathSegmentReply
+from lib.packet.path_mgmt.seg_recs import PathSegmentRecords
 from lib.packet.scmp.types import SCMPClass, SCMPPathClass
 from lib.packet.svc import SVCType
 from lib.path_db import DBResult, PathSegmentDB
@@ -62,10 +63,14 @@ from scion_elem.scion_elem import SCIONElement
 
 
 # Exported metrics.
-REQS_TOTAL = Counter("ps_reqs_total", "# of path requests", ["server_id", "isd_as"])
-REQS_PENDING = Gauge("ps_req_pending_total", "# of pending path requests", ["server_id", "isd_as"])
-SEGS_TO_ZK = Gauge("ps_segs_to_zk_total", "# of path segments to ZK", ["server_id", "isd_as"])
-REVS_TO_ZK = Gauge("ps_revs_to_zk_total", "# of revocations to ZK", ["server_id", "isd_as"])
+REQS_TOTAL = Counter("ps_reqs_total", "# of path requests", [
+                     "server_id", "isd_as"])
+REQS_PENDING = Gauge("ps_req_pending_total",
+                     "# of pending path requests", ["server_id", "isd_as"])
+SEGS_TO_ZK = Gauge("ps_segs_to_zk_total", "# of path segments to ZK", [
+                   "server_id", "isd_as"])
+REVS_TO_ZK = Gauge("ps_revs_to_zk_total", "# of revocations to ZK", [
+                   "server_id", "isd_as"])
 HT_ROOT_MAPPTINGS = Gauge("ps_ht_root_mappings_total", "# of hashtree root to segment mappings",
                           ["server_id", "isd_as"])
 IS_MASTER = Gauge("ps_is_master", "true if this process is the replication master",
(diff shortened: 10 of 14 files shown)


scrye:
Reviewed 8 of 10 files at r1, 2 of 3 files at r2.
Review status: 9 of 12 files reviewed at latest revision, 6 unresolved discussions.

---

*[proto/path_mgmt.capnp, line 11 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw4svbL3aauZtIuRPek:-Kw4svbL3aauZtIuRPel:b-9rl54x) ([raw file](https://github.com/netsec-ethz/scion/blob/3d4354c7cd90a6275e58c6b7f543943f3541208e/proto/path_mgmt.capnp#L11)):*
> ```
> 
> struct SegReq {
>     id @0 :UInt64;  # Request ID
> ```

Renumbering IDs here breaks compatibility with applications still running the old version of the protocol.

The recommended way of updating a Cap'n Proto structure is by using the next available ID (i.e., `id @4`).

---

*[python/path_server/core.py, line 234 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw4wwqbe2sKnqj7rQnG:-Kw4wwqdYzb5hRXihwk-:byy4tpu) ([raw file](https://github.com/netsec-ethz/scion/blob/3d4354c7cd90a6275e58c6b7f543943f3541208e/python/path_server/core.py#L234)):*
> ```Python
>         sflags.add(PATH_FLAG_CACHEONLY)
>         flags = tuple(sflags)
>         req_id = random.randint(0, 2**64 - 1)
> ```

Refactor this to get rid of the "magic" max 64-bit uint (maybe define a helper function like random_uint64).

---

*[python/sciond/req.py, line 45 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw5cHtAfOG1rvVawBdO:-Kw5cHtByCW-bHnasgki:b-vrrrlk) ([raw file](https://github.com/netsec-ethz/scion/blob/a9f40be7baaa994b486ad01d3f994517c12ba8c5/python/sciond/req.py#L45)):*
> ```Python
>         self._segs_to_verify = self.reply.recs().num_segs()
>         self._ver_tout = threading.Timer(0.3, self._ver_tout_fired)
>         self._ver_tout.start()
> ```

I don't understand why the timer is started in `set_reply`. This function looks to be called when the reply comes back in, why is a timer necessary at that point?

Also, just for the sake of testing, can you set it something extremely small (including firing instantly), just so we can confirm that this code execution path does what it is intended to do?

---

*[python/sciond/req.py, line 77 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw5cDHB-yiE2lnm3y75:-Kw5cDHDl3gOZg70mFoe:br2v1ln) ([raw file](https://github.com/netsec-ethz/scion/blob/a9f40be7baaa994b486ad01d3f994517c12ba8c5/python/sciond/req.py#L77)):*
> ```Python
>         """
>         Gets called when the verification timer fired. From now on we try to immediately
>         fullfil requests as new path segments get verified.
> ```

fulfill

---

*[python/sciond/req.py, line 84 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw5dCGFprFsu5hf2QrO:-Kw5dCGGj053AOg46YDO:bp3hwv7) ([raw file](https://github.com/netsec-ethz/scion/blob/a9f40be7baaa994b486ad01d3f994517c12ba8c5/python/sciond/req.py#L84)):*
> ```Python
>         self._check()
> 
>     def _timeout_fired(self):
> ```

Naming is inconsistent (`_ver_tout_fired` vs `_timeout_fired`). `tout` isn't a lot shorter than `timeout`, so in my opinion for clarity the whole word can be used.

Also, since there are two timeouts it would help if the generic name one would also indicate what it is tracking -- for example, `path_req_timeout` vs `path_ver_timeout`.

---

*[python/sciond/sciond.py, line 521 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw4z7t346Wc0VzAh1SQ:-Kw4z7t346Wc0VzAh1SR:b55zt4l) ([raw file](https://github.com/netsec-ethz/scion/blob/3d4354c7cd90a6275e58c6b7f543943f3541208e/python/sciond/sciond.py#L521)):*
> ```Python
>                 if key not in self.requested_paths:
>                     # No previous outstanding request
>                     req_id = random.randint(0, 2**64 - 1)
> ```

The same random uint64 helper above could be used here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

shitz:
Review status: 8 of 13 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[proto/path_mgmt.capnp, line 11 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw4svbL3aauZtIuRPek:-Kw5xU5hzTgYW10fCcwC:b-xeulm) ([raw file](https://github.com/netsec-ethz/scion/blob/3d4354c7cd90a6275e58c6b7f543943f3541208e/proto/path_mgmt.capnp#L11)):*

Renumbering IDs here breaks compatibility with applications still running the old version of the protocol.

The recommended way of updating a Cap'n Proto structure is by using the next available ID (i.e., `id @4`).
</blockquote></details>

I'm aware of that, but since we are still at the stage where we control the entire deployment I'm willing to make these breaking changes, since an Id should definitely be at the beginning of the proto. As soon as we have deployments that are controlled by other people, we cannot do this anymore of course.

---

*[python/path_server/core.py, line 234 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw4wwqbe2sKnqj7rQnG:-Kw5ygZny-M_k8N1Imuo:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/3d4354c7cd90a6275e58c6b7f543943f3541208e/python/path_server/core.py#L234)):*

Refactor this to get rid of the "magic" max 64-bit uint (maybe define a helper function like random_uint64).
</blockquote></details>

Done.

---

*[python/sciond/req.py, line 77 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw5cDHB-yiE2lnm3y75:-Kw5ynCnO3Pee2v0Pg32:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/a9f40be7baaa994b486ad01d3f994517c12ba8c5/python/sciond/req.py#L77)):*

fulfill
</blockquote></details>

Done.

---

*[python/sciond/req.py, line 84 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw5dCGFprFsu5hf2QrO:-Kw6-urJphHFGhCYQIk3:bneelzz) ([raw file](https://github.com/netsec-ethz/scion/blob/a9f40be7baaa994b486ad01d3f994517c12ba8c5/python/sciond/req.py#L84)):*

Naming is inconsistent (`_ver_tout_fired` vs `_timeout_fired`). `tout` isn't a lot shorter than `timeout`, so in my opinion for clarity the whole word can be used.

Also, since there are two timeouts it would help if the generic name one would also indicate what it is tracking -- for example, `path_req_timeout` vs `path_ver_timeout`.
</blockquote></details>

As discussed there is no only on timer which is simply called `_timer`

---

*[python/sciond/sciond.py, line 521 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw4z7t346Wc0VzAh1SQ:-Kw6-zVxFg8hJ1z613FU:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/3d4354c7cd90a6275e58c6b7f543943f3541208e/python/sciond/sciond.py#L521)):*

The same random uint64 helper above could be used here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

scrye:
I was a bit scared by the lack of sync primitives on shared data, so I searched for races and possibly found a few. There might be others lurking at the edges.

I think locking all code except `wait()` would lead to safer code.

---

Review status: 8 of 13 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[python/sciond/req.py, line 44 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw6NmBV2YOpOPNxVJqV:-Kw6NmBV2YOpOPNxVJqW:buzbmes) ([raw file](https://github.com/netsec-ethz/scion/blob/ff6db7dab5f2bbd942bc690d05346f72276a2df3/python/sciond/req.py#L44)):*
> ```Python
>         self._segs_to_verify = self.reply.recs().num_segs()
> 
>     def verified_segment(self):
> ```

`verified_segment` callers race against each other; this can lead to `self._segs_to_verify < 0`. I think there's only one caller anyway, so it's ok, but it's something to keep in mind.

---

*[python/sciond/req.py, line 78 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw6QnlM9D2lO-WVnuvE:-Kw6QnlM9D2lO-WVnuvF:buund8c) ([raw file](https://github.com/netsec-ethz/scion/blob/ff6db7dab5f2bbd942bc690d05346f72276a2df3/python/sciond/req.py#L78)):*
> ```Python
>         if not self._check():
>             self._timer = threading.Timer(PATH_REQ_TOUT - _WAIT_TIME, self._second_timer_fired)
>             self._timer.start()
> ```

There's a race here, I think. Sequence of events:
  - first timer fires with nothing received yet, `self._check()` evaluates to `false`, a new timer is created but the thread is preempted before it is started
  - main thread enters `verified_segment`, finds out that it no longer has to wait so runs `_check`
  - `check` sees everything ready, calls `_done()` thus resetting `self._timer` to `None`
  - first timer thread comes back, attempts `self._timer.start()` which raises an exception

---

*[python/sciond/req.py, line 92 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw6MLJ6peV2wHAG1Ty4:-Kw6MLJ74Mq7hmblSjJY:b-empe95) ([raw file](https://github.com/netsec-ethz/scion/blob/ff6db7dab5f2bbd942bc690d05346f72276a2df3/python/sciond/req.py#L92)):*
> ```Python
>         """Wake up all waiters."""
>         # Cancel outstanding timer if there is one.
>         if self._timer:
> ```

There's a race here I think. Is the following sequence of events possible?
  - We need exactly one path
  - The first timer expires without a path, sets `_wait` to false, arms the second timer
  - In the main thread, `verified_segments` is called for the path just before the second timer expires; because we only needed one path, the `done` on line `58` gets called
  - In the main thread, `self._timer` above evaluates to true, main thread is preempted
  - Timer 2 expires, runs `_second_timer_fired`, sets `self._timer` to `None`
  - Main thread gets rescheduled and an `Exception` is raised as `cancel()` below is called on `None`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-:-Kw6RnXSudDf9MzZPrIh:b1cltb0)*

shitz:
You are right. I did as you proposed.

---

Review status: 7 of 13 files reviewed at latest revision, 3 unresolved discussions.

---

*[python/sciond/req.py, line 44 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw6NmBV2YOpOPNxVJqV:-Kw9A_7YPk8G7I-Axj8W:b-9c8uc4) ([raw file](https://github.com/netsec-ethz/scion/blob/ff6db7dab5f2bbd942bc690d05346f72276a2df3/python/sciond/req.py#L44)):*

`verified_segment` callers race against each other; this can lead to `self._segs_to_verify < 0`. I think there's only one caller anyway, so it's ok, but it's something to keep in mind.
</blockquote></details>

There is only one caller right now, but maybe this changes in the future (unlikely). Anyway, I made all code acquire a lock first.

---

*[python/sciond/req.py, line 78 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw6QnlM9D2lO-WVnuvE:-Kw9Aiepdgcs9JyYpQg0:b-o7vuhc) ([raw file](https://github.com/netsec-ethz/scion/blob/ff6db7dab5f2bbd942bc690d05346f72276a2df3/python/sciond/req.py#L78)):*

There's a race here, I think. Sequence of events:
  - first timer fires with nothing received yet, `self._check()` evaluates to `false`, a new timer is created but the thread is preempted before it is started
  - main thread enters `verified_segment`, finds out that it no longer has to wait so runs `_check`
  - `check` sees everything ready, calls `_done()` thus resetting `self._timer` to `None`
  - first timer thread comes back, attempts `self._timer.start()` which raises an exception
</blockquote></details>

Good catch! Done.

---

*[python/sciond/req.py, line 92 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kw6MLJ6peV2wHAG1Ty4:-Kw9AlseQA2YDXgjGprD:b-o7vuhc) ([raw file](https://github.com/netsec-ethz/scion/blob/ff6db7dab5f2bbd942bc690d05346f72276a2df3/python/sciond/req.py#L92)):*

There's a race here I think. Is the following sequence of events possible?
  - We need exactly one path
  - The first timer expires without a path, sets `_wait` to false, arms the second timer
  - In the main thread, `verified_segments` is called for the path just before the second timer expires; because we only needed one path, the `done` on line `58` gets called
  - In the main thread, `self._timer` above evaluates to true, main thread is preempted
  - Timer 2 expires, runs `_second_timer_fired`, sets `self._timer` to `None`
  - Main thread gets rescheduled and an `Exception` is raised as `cancel()` below is called on `None`
</blockquote></details>

Good catch! Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-:-Kw9ApUzMzHjCtrOYkvd:b-ac0bub)*

scrye:
:lgtm:

---

Reviewed 3 of 5 files at r3, 3 of 3 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-:-Kw9YAefQAaxiiuhrYZ6:bnfp4nl)*

kormat:
Reviewed 6 of 10 files at r1, 1 of 3 files at r2, 3 of 5 files at r3, 1 of 3 files at r4.

---

*[go/lib/ctrl/path_mgmt/seg_req.go, line 61 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSmjelHT1tYqQu8epN:-KxSmjelHT1tYqQu8epO:b-6pd15t) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/go/lib/ctrl/path_mgmt/seg_req.go#L61)):*
> ```Go
> 
> func (s *SegReq) String() string {
> 	return fmt.Sprintf("Id: %08x %v -> %v, Flags: %v", s.Id, s.SrcIA(), s.DstIA(), s.Flags)
> ```

`"Id: %016x %s -> %s Flags: %v"`
(the width is actually 16B, as hex is 2-chars-per-byte, use `%s` instead of `%v` for types that implement fmt.Stringer)

---

*[go/lib/ctrl/path_mgmt/seg_req.go, line 64 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSorX9xWhLpLhAbdNJ:-KxSorX9xWhLpLhAbdNK:b-dh2111) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/go/lib/ctrl/path_mgmt/seg_req.go#L64)):*
> ```Go
> }
> 
> type SegReply struct {
> ```

I think this should be in its own file, but also this doesn't seem to match the capnp definition.

---

*[go/lib/ctrl/path_mgmt/seg_req.go, line 83 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSpF0TBksFnYb42DM8:-KxSpF0TBksFnYb42DM9:b-qvr665) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/go/lib/ctrl/path_mgmt/seg_req.go#L83)):*
> ```Go
> 
> func (s *SegReply) String() string {
> 	return fmt.Sprintf("Id: %08x\n%s", s.Id, s.Recs)
> ```

Same issue with width here. Actually, just make a seg req Id its own type, and implement fmt.Stringer for it.

---

*[integration/revocation_test.sh, line 46 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSpfJ521YJpzC5FN-7:-KxSpfJ521YJpzC5FN-8:b-223fgf) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/integration/revocation_test.sh#L46)):*
> ```Shell
> fi
> # Bring down routers.
> SLEEP=3
> ```

This looks slightly short. The current ifid timeout in the BS is 3s. `SLEEP=4` would reduce the chances of race-conditions here.

---

*[python/lib/packet/path_mgmt/seg_req.py, line 32 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSrvsGt8zCP6NS8alo:-KxSrvsGt8zCP6NS8alp:b-66zg8r) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/lib/packet/path_mgmt/seg_req.py#L32)):*
> ```Python
>     """Describes a request for path segment(s)"""
>     NAME = "PathSegmentReq"
>     LEN = 1 + 2 * ISD_AS.LEN
> ```

This pre-dates capnp, can can be removed.

---

*[python/lib/packet/path_mgmt/seg_req.py, line 65 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSrEcguEWyIRfigpxq:-KxSrEcguEWyIRfigpxr:b-rv2lt8) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/lib/packet/path_mgmt/seg_req.py#L65)):*
> ```Python
> 
>     def __eq__(self, other):
>         return self.p.id == other.p.id
> ```

This should also compare the other attributes.

---

*[python/lib/packet/path_mgmt/seg_req.py, line 71 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSrKZz8V3hVGKadIT7:-KxSrKZz8V3hVGKadIT8:b-uosugs) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/lib/packet/path_mgmt/seg_req.py#L71)):*
> ```Python
> 
>     def short_desc(self):
>         return "Id: %08x %s -> %s  %s" % (self.req_id(), self.src_ia(), self.dst_ia(), self.flags())
> ```

`%016x` damn you.

---

*[python/path_server/base.py, line 66 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSsBMMdRPpspakpMXd:-KxSsBMMdRPpspakpMXe:bpqfql8) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/path_server/base.py#L66)):*
> ```Python
> 
> # Exported metrics.
> REQS_TOTAL = Counter("ps_reqs_total", "# of path requests", [
> ```

All of these lines were already <100 (and if they weren't, the lint step would catch it in CI). There's no need to change them, and it adds a lot of noise to the review/history.

---

*[python/path_server/core.py, line 184 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxT1tf5E9A0QdS60LRK:-KxT1tf5E9A0QdS60LRL:bgpbazk) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/path_server/core.py#L184)):*
> ```Python
>                       len(self._segs_to_prop))
>         for pcbs in self._gen_prop_recs(self._segs_to_prop):
>             recs = PathRecordsSync.from_values(pcbs)
> ```

The commit description should mention how PathRecordsSync is being repurposed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

kormat:
Reviewed 2 of 3 files at r4.
Review status: all files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[python/sciond/req.py, line 24 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTC8gBlf44l6ICG3au:-KxTC8gBlf44l6ICG3av:b-e0f1uu) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/sciond/req.py#L24)):*
> ```Python
> 
> 
> class RequestState:  # pragma: no cover
> ```

This needs some extensive documentation to explain what it does and why. So many attributes, that presumably interact in some complex fashion.

---

*[python/sciond/req.py, line 29 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTFpX8sOCPkXRbDow7:-KxTFpX8sOCPkXRbDow8:bshatg1) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/sciond/req.py#L29)):*
> ```Python
>         self.req = req
>         self.reply = None
>         self.checkf = checkf
> ```

`self._checkf` is probably better, as there's no need for it to be externally accessed.

---

*[python/sciond/req.py, line 36 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTGVY7zf2uauqFm2Ao:-KxTGVY7zf2uauqFm2Ap:b-p434m3) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/sciond/req.py#L36)):*
> ```Python
>         self._lock = threading.Lock()
>         self._timer = threading.Timer(_WAIT_TIME, self._first_timer_fired)
>         self._timer.start()
> ```

This changes the semantics from what we had discussed, is that intended? What i remember is that a timer would start as soon as `set_reply` was called. When that timer expires (or all possible segs are verified, whichever is sooner), a response would be sent with all available segments. I thiink that would also simplify the logic quite a bit.

---

*[python/sciond/req.py, line 43 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTFP9hWX6YFZfifef2:-KxTFP9hWX6YFZfifef3:bkv033i) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/sciond/req.py#L43)):*
> ```Python
>     def set_reply(self, reply):
>         with self._lock:
>             self.reply = reply
> ```

`.reply` is never used anywhere else, so it could be dropped, and the name of the method updated.

---

*[python/sciond/req.py, line 53 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTCRw95etETmQHtv2m:-KxTCRwArwmrvM0XBCcj:bhajejd) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/sciond/req.py#L53)):*
> ```Python
>         """
>         with self._lock:
>             if self._segs_to_verify == 0:
> ```

This seems like it should be an error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

kormat:
Review status: all files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[python/sciond/req.py, line 36 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTGVY7zf2uauqFm2Ao:-KxTPswvtGXyrdT_FDVa:b4qqha2) ([raw file](https://github.com/netsec-ethz/scion/blob/af8c5dc45199d8c711ed1e96b25ad98b7ef379ef/python/sciond/req.py#L36)):*

This changes the semantics from what we had discussed, is that intended? What i remember is that a timer would start as soon as `set_reply` was called. When that timer expires (or all possible segs are verified, whichever is sooner), a response would be sent with all available segments. I thiink that would also simplify the logic quite a bit.
</blockquote></details>

Just to make sure i wasn't completely wrong, i refactored it with what i have in mind, and i think it looks a lot better, and still provides the desired behaviour: https://github.com/kormat/scion/tree/pr1280

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

shitz:
Review status: 7 of 14 files reviewed at latest revision, 14 unresolved discussions.

---

*[go/lib/ctrl/path_mgmt/seg_req.go, line 61 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSmjelHT1tYqQu8epN:-KxheoqaahyILpv4A5tg:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/go/lib/ctrl/path_mgmt/seg_req.go#L61)):*

`"Id: %016x %s -> %s Flags: %v"`
(the width is actually 16B, as hex is 2-chars-per-byte, use `%s` instead of `%v` for types that implement fmt.Stringer)
</blockquote></details>

Done.

---

*[go/lib/ctrl/path_mgmt/seg_req.go, line 64 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSorX9xWhLpLhAbdNJ:-KxmdJ4PgjsXS9f9lh1g:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/go/lib/ctrl/path_mgmt/seg_req.go#L64)):*

I think this should be in its own file, but also this doesn't seem to match the capnp definition.
</blockquote></details>

Done.

---

*[go/lib/ctrl/path_mgmt/seg_req.go, line 83 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSpF0TBksFnYb42DM8:-KxmdL-HuGD_41Zxmsua:b-yyclk7) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/go/lib/ctrl/path_mgmt/seg_req.go#L83)):*

Same issue with width here. Actually, just make a seg req Id its own type, and implement fmt.Stringer for it.
</blockquote></details>

It's now only used in `SegReq`.

---

*[integration/revocation_test.sh, line 46 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSpfJ521YJpzC5FN-7:-KxmdUakPO1-JZ_g7qBv:bkw28wo) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/integration/revocation_test.sh#L46)):*

This looks slightly short. The current ifid timeout in the BS is 3s. `SLEEP=4` would reduce the chances of race-conditions here.
</blockquote></details>

The intention was to increase the chance of race-condition to (possibly) find new bugs. I agree 3s is probably a bit too aggressive for the test.

---

*[python/lib/packet/path_mgmt/seg_req.py, line 32 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSrvsGt8zCP6NS8alo:-KxmdvHGjxRjNz6fXRZn:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/lib/packet/path_mgmt/seg_req.py#L32)):*

This pre-dates capnp, can can be removed.
</blockquote></details>

Done.

---

*[python/lib/packet/path_mgmt/seg_req.py, line 65 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSrEcguEWyIRfigpxq:-KxmeLxRjxNM4zCw_Uep:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/lib/packet/path_mgmt/seg_req.py#L65)):*

This should also compare the other attributes.
</blockquote></details>

Done.

---

*[python/lib/packet/path_mgmt/seg_req.py, line 71 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSrKZz8V3hVGKadIT7:-KxmeOpIibn7ueUBaaFe:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/lib/packet/path_mgmt/seg_req.py#L71)):*

`%016x` damn you.
</blockquote></details>

Done.

---

*[python/path_server/base.py, line 66 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSsBMMdRPpspakpMXd:-KxmiP464llwpyxr144s:bj9yscp) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/path_server/base.py#L66)):*

All of these lines were already <100 (and if they weren't, the lint step would catch it in CI). There's no need to change them, and it adds a lot of noise to the review/history.
</blockquote></details>

My editor was set to autopep8 for whatever reason. Reverted those.

---

*[python/path_server/core.py, line 184 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxT1tf5E9A0QdS60LRK:-KxmjICBPm-52eXMs9rn:b-t395w0) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/path_server/core.py#L184)):*

The commit description should mention how PathRecordsSync is being repurposed.
</blockquote></details>

It should have been `PathRecordsSync` in the first place.

---

*[python/sciond/req.py, line 24 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTC8gBlf44l6ICG3au:-KxsQWA1sZkueWgO25IR:b-3bkalf) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/sciond/req.py#L24)):*

This needs some extensive documentation to explain what it does and why. So many attributes, that presumably interact in some complex fashion.
</blockquote></details>

Changed to your proposed simplified version.

---

*[python/sciond/req.py, line 29 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTFpX8sOCPkXRbDow7:-Kxmk54DhZaJ-SeL51JJ:bgreq8h) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/sciond/req.py#L29)):*

`self._checkf` is probably better, as there's no need for it to be externally accessed.
</blockquote></details>

not applicable

---

*[python/sciond/req.py, line 36 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTGVY7zf2uauqFm2Ao:-KxsOn3gK7pSdPKYH85I:b1xwmg4) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/sciond/req.py#L36)):*

Just to make sure i wasn't completely wrong, i refactored it with what i have in mind, and i think it looks a lot better, and still provides the desired behaviour: https://github.com/kormat/scion/tree/pr1280
</blockquote></details>

Yes the semantics are changed. In my version, SCIOND returns all available paths after at most 0.5s. This includes the lookup time. Your version will give an additional 0.2s for verification even if the lookup took > 1s. 

Besides that your version definitely streamlined the entire timeout handling, thus I opted to include your version.

---

*[python/sciond/req.py, line 43 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTFP9hWX6YFZfifef2:-KxsPT6zPZ0SIsyTwS23:bgreq8h) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/sciond/req.py#L43)):*

`.reply` is never used anywhere else, so it could be dropped, and the name of the method updated.
</blockquote></details>

not applicable

---

*[python/sciond/req.py, line 53 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxTCRw95etETmQHtv2m:-KxsQ9Umfes_Wocvlb1O:bgreq8h) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/sciond/req.py#L53)):*

This seems like it should be an error.
</blockquote></details>

not applicable

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

kormat:
Reviewed 7 of 7 files at r5.

---

*[go/lib/ctrl/path_mgmt/seg_req.go, line 61 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSmjelHT1tYqQu8epN:-KxwjtPjUa4LvsX0Hz0Q:b9thxhv) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/go/lib/ctrl/path_mgmt/seg_req.go#L61)):*

Done.
</blockquote></details>

You dropped the `0`.

---

*[go/lib/ctrl/path_mgmt/seg_req.go, line 83 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSpF0TBksFnYb42DM8:-KxwjKfFIOpc2ffnjYty:b-6dqnls) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/go/lib/ctrl/path_mgmt/seg_req.go#L83)):*

It's now only used in `SegReq`.
</blockquote></details>

The type approach is still useful, e.g. for a path server/sciond logging just the id.

---

*[python/lib/packet/path_mgmt/seg_req.py, line 71 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSrKZz8V3hVGKadIT7:-Kxwk-g52wKVOlz2--7Z:b9thxhv) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/lib/packet/path_mgmt/seg_req.py#L71)):*

Done.
</blockquote></details>

You dropped the `0`.

---

*[python/path_server/base.py, line 564 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwkB2DBfuhGM1jGJNI:-KxwkB2DBfuhGM1jGJNJ:b67mx6r) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/base.py#L564)):*
> ```Python
>         formatter = formatter = Rfc3339Formatter(
>             "%(asctime)s [%(levelname)s] (%(threadName)s) %(message)s "
>             "{req=%(req)s, from=%(from)s}")
> ```

Including the entire request seems a bit overkill. If the request (and its id) are logged on reception, then just logging the req id thereafter should be plenty, and reduce noise.

---

*[python/path_server/core.py, line 184 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxT1tf5E9A0QdS60LRK:-KxwkLudOnue66Y9pycB:b-rt31h4) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/path_server/core.py#L184)):*

It should have been `PathRecordsSync` in the first place.
</blockquote></details>

That's not what`PathRecordsSync` was added for. (See https://github.com/netsec-ethz/scion/pull/383 for some historical context). It was specifically for syncing path segments between cPSes in the _same_ AS.

---

*[python/sciond/sciond.py, line 502 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwqCXKywzmhBDtoAKg:-KxwqCXKywzmhBDtoAKh:blvar3v) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L502)):*
> ```Python
>                 logging.error("No paths found for %s", dst_ia)
>                 return [], SCIONDPathReplyError.NO_PATHS
>         return paths, error_code
> ```

`error_code` is only ever set once, so it can just be replaced with the `.OK` version.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

kormat:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[python/path_server/base.py, line 68 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwqguS-HvD6Mg_AW9U:-KxwqguS-HvD6Mg_AW9V:b7c8anj) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/base.py#L68)):*
> ```Python
> REQS_TOTAL = Counter("ps_reqs_total", "# of path requests", ["server_id", "isd_as"])
> REQS_PENDING = Gauge("ps_req_pending_total",
>                      "# of pending path requests", ["server_id", "isd_as"])
> ```

Unnecessary change.

---

*[python/path_server/base.py, line 543 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwqkhUVGg-PbK_GMAs:-KxwqkhUVGg-PbK_GMAt:b7c8anj) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/base.py#L543)):*
> ```Python
>         try:
>             self.path_cache.store("%s-%s" %
>                                   (hash_, SCIONTime.get_time()), data)
> ```

Unnecessary change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

kormat:
Review status: all files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[python/path_server/core.py, line 235 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwqtMuG_Je6kAVaJON:-KxwqtMuG_Je6kAVaJOO:b7c8anj) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/core.py#L235)):*
> ```Python
>         flags = tuple(sflags)
>         req = PathSegmentReq.from_values(random_uint64(), src_ia, dst_ia, flags=flags)
>         logger.debug("Asking master (%s) for segment: %s" %
> ```

Unnecessary change.

---

*[python/path_server/core.py, line 292 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kxwr0Q2xWBulNns8ylj:-Kxwr0Q2xWBulNns8ylk:b7c8anj) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/core.py#L292)):*
> ```Python
> 
>         self._send_path_segments(
>             req, meta, logger, core=core_segs, down=down_segs)
> ```

Unnecessary change.

---

*[python/path_server/core.py, line 307 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kxwr-OY4j4UVSbKrwIH:-Kxwr-OZjuyT9IaNBVUW:b7c8anj) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/core.py#L307)):*
> ```Python
>             # Segments not found and it is a new request.
>             self.pending_req[(dst_ia, sibra)][
>                 req.req_id()] = (req, meta, logger)
> ```

Unnecessary change.

---

*[python/path_server/core.py, line 339 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kxwr7KPnXNkNA5E1WRJ:-Kxwr7KPnXNkNA5E1WRK:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/core.py#L339)):*
> ```Python
>                 # Core segment not found and it is a new request.
>                 self.pending_req[(dseg_ia, sibra)][
>                     seg_req.req_id()] = (seg_req, meta, logger)
> ```

Unnecessary line-wrap.

---

*[python/path_server/core.py, line 407 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrCraYG78SiZiV4c1:-KxwrCraYG78SiZiV4c2:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/core.py#L407)):*
> ```Python
>         if self._labels:
>             SEGS_TO_MASTER.labels(
>                 **self._labels).set(len(self._segs_to_master))
> ```

Unnecessary line-wrap.

---

*[python/path_server/local.py, line 49 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrHMmVZQohRn3YPIX:-KxwrHMmVZQohRn3YPIY:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/local.py#L49)):*
> ```Python
>         up_labels = {**self._labels, "type": "up"} if self._labels else None
>         self.up_segments = PathSegmentDB(
>             max_res_no=self.MAX_SEG_NO, labels=up_labels)
> ```

Unnecessary line-wrap.

---

*[python/path_server/local.py, line 97 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrIzJohSi9stQ_GJ7:-KxwrIzJohSi9stQ_GJ8:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/local.py#L97)):*
> ```Python
>         if up_segs | core_segs | down_segs:
>             self._send_path_segments(
>                 req, meta, logger, up_segs, core_segs, down_segs)
> ```

Unnecessary line-wrap.

---

*[python/path_server/local.py, line 102 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrJRoXAlGvF5mHAZR:-KxwrJRoXAlGvF5mHAZS:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/local.py#L102)):*
> ```Python
>             self._request_paths_from_core(req, logger)
>             self.pending_req[(dst_ia, req.p.flags.sibra)][
>                 req.req_id()] = (req, meta, logger)
> ```

Unnecessary line-wrap.

---

*[python/sciond/sciond.py, line 112 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrP1HpU0nOj1dTMR1:-KxwrP1HpU0nOj1dTMR2:b5xuytr) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L112)):*
> ```Python
>             segment_ttl=self.SEGMENT_TTL, labels=down_labels)
>         self.core_segments = PathSegmentDB(
>             segment_ttl=self.SEGMENT_TTL, labels=core_labels)
> ```

Unnecessary line-wraps.

---

*[python/sciond/sciond.py, line 144 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrQFNSFiCs5aa38o9:-KxwrQFNSFiCs5aa38oA:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L144)):*
> ```Python
>         if run_local_api:
>             self._api_sock = ReliableSocket(
>                 bind_unix=(self.api_addr, "sciond"))
> ```

Unnecessary line-wrap.

---

*[python/sciond/sciond.py, line 308 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrUSSvbkGY4B1t0hx:-KxwrUSTg9q_OPgrQ_xC:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L308)):*
> ```Python
>     def _send_path_reply(self, req_id, reply_entries, error, meta):
>         path_reply = SCIONDMsg(SCIONDPathReply.from_values(
>             reply_entries, error), req_id)
> ```

Unnecessary line-wrap.

---

*[python/sciond/sciond.py, line 322 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrVJ1mslRdmfCjK7X:-KxwrVJ2r4wyAXkDXn3y:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L322)):*
> ```Python
>                 self.addr.isd_as, self.is_core_as(), self.topology.mtu)
>         as_reply = SCIONDMsg(
>             SCIONDASInfoReply.from_values([reply_entry]), pld.id)
> ```

Unnecessary line-wrap.

---

*[python/sciond/sciond.py, line 361 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrViDID-Ib_D9Fjou:-KxwrViDID-Ib_D9Fjov:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L361)):*
> ```Python
>                 svc_entries.append(reply_entry)
>         svc_reply = SCIONDMsg(
>             SCIONDServiceInfoReply.from_values(svc_entries), pld.id)
> ```

Unnecessary line-wrap.

---

*[python/sciond/sciond.py, line 368 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrWK2S63FzyR-z6PK:-KxwrWK2S63FzyR-z6PL:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L368)):*
> ```Python
>         assert isinstance(request, SCIONDRevNotification), type(request)
>         status = self.handle_revocation(
>             CtrlPayload(PathMgmt(request.rev_info())), meta)
> ```

Unnecessary line-wrap.

---

*[python/sciond/sciond.py, line 451 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrXT_tvxBalRoxI6B:-KxwrXT_tvxBalRoxI6C:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L451)):*
> ```Python
>                 if self._verify_revocation_for_asm(rev_info, asm):
>                     logging.debug("Removing segment: %s" %
>                                   segment.short_desc())
> ```

Unnecessary line-wrap.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

shitz:
Review status: all files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/path_mgmt/seg_req.go, line 61 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSmjelHT1tYqQu8epN:-KxxTz2M4w1rqkX3tK53:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/go/lib/ctrl/path_mgmt/seg_req.go#L61)):*

You dropped the `0`.
</blockquote></details>

Done.

---

*[go/lib/ctrl/path_mgmt/seg_req.go, line 83 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSpF0TBksFnYb42DM8:-KxxVzQPj0PSFZXrNL4z:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/go/lib/ctrl/path_mgmt/seg_req.go#L83)):*

The type approach is still useful, e.g. for a path server/sciond logging just the id.
</blockquote></details>

Done.

---

*[python/lib/packet/path_mgmt/seg_req.py, line 71 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxSrKZz8V3hVGKadIT7:-KxxW7m9eM_BRH8mR5ti:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/lib/packet/path_mgmt/seg_req.py#L71)):*

You dropped the `0`.
</blockquote></details>

Done.

---

*[python/path_server/base.py, line 68 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwqguS-HvD6Mg_AW9U:-KxxWHE0Ww12RxG6OjTv:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/base.py#L68)):*

Unnecessary change.
</blockquote></details>

Done.

---

*[python/path_server/base.py, line 543 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwqkhUVGg-PbK_GMAs:-KxxWN1OH9nH05GfZSm_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/base.py#L543)):*

Unnecessary change.
</blockquote></details>

Done.

---

*[python/path_server/base.py, line 564 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwkB2DBfuhGM1jGJNI:-KxxWic180PrKHriYXa8:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/base.py#L564)):*

Including the entire request seems a bit overkill. If the request (and its id) are logged on reception, then just logging the req id thereafter should be plenty, and reduce noise.
</blockquote></details>

Done.

---

*[python/path_server/core.py, line 184 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxT1tf5E9A0QdS60LRK:-KxxX5sP8hSqrDVjCAVo:b-39jyeg) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/path_server/core.py#L184)):*

That's not what`PathRecordsSync` was added for. (See https://github.com/netsec-ethz/scion/pull/383 for some historical context). It was specifically for syncing path segments between cPSes in the _same_ AS.
</blockquote></details>

That seems odd. Was this before we had ZK, because ZK is used to do that and I think there we store path segments directly. Will update description.

---

*[python/path_server/core.py, line 235 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwqtMuG_Je6kAVaJON:-KxxXcYn4UcHdUwyZ56m:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/core.py#L235)):*

Unnecessary change.
</blockquote></details>

Done.

---

*[python/path_server/core.py, line 292 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kxwr0Q2xWBulNns8ylj:-KxxYhiZu7WS-3fUb8SD:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/core.py#L292)):*

Unnecessary change.
</blockquote></details>

Done.

---

*[python/path_server/core.py, line 307 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kxwr-OY4j4UVSbKrwIH:-KxxXifTZPjVJLBSdQqE:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/core.py#L307)):*

Unnecessary change.
</blockquote></details>

Done.

---

*[python/path_server/core.py, line 339 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-Kxwr7KPnXNkNA5E1WRJ:-KxxXniL4J-xccKPhA8Z:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/core.py#L339)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/path_server/core.py, line 407 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrCraYG78SiZiV4c1:-KxxXryEfHbgWsXSy0T1:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/core.py#L407)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/path_server/local.py, line 49 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrHMmVZQohRn3YPIX:-KxxXyVvUURzXoO8A9gn:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/local.py#L49)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/path_server/local.py, line 97 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrIzJohSi9stQ_GJ7:-KxxY1ZSgsI4cuja2wBS:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/local.py#L97)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/path_server/local.py, line 102 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrJRoXAlGvF5mHAZR:-KxxY1qio9SYMxYzwfuF:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/path_server/local.py#L102)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 112 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrP1HpU0nOj1dTMR1:-KxxYBjLmE_ZYuX0ig3z:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L112)):*

Unnecessary line-wraps.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 144 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrQFNSFiCs5aa38o9:-KxxYHH57jT2TYFXsl-w:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L144)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 308 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrUSSvbkGY4B1t0hx:-KxxYLMbaqCJRftuqh6a:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L308)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 322 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrVJ1mslRdmfCjK7X:-KxxYNNxCVL6n8z-5hBh:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L322)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 361 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrViDID-Ib_D9Fjou:-KxxYPhFREir_w-aW0Ru:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L361)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 368 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrWK2S63FzyR-z6PK:-KxxYRRudbB7CfUWIJss:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L368)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 451 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwrXT_tvxBalRoxI6B:-KxxYVZC-wyYOp85LcET:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L451)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 502 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxwqCXKywzmhBDtoAKg:-KxxYcmJnA7Bfif_pRAF:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/330daff0fd7e8770b782168c1a178df46d512b18/python/sciond/sciond.py#L502)):*

`error_code` is only ever set once, so it can just be replaced with the `.OK` version.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

kormat:
Reviewed 5 of 6 files at r6.
Review status: 13 of 14 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[python/path_server/core.py, line 184 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KxT1tf5E9A0QdS60LRK:-KyGkYlTiuNQ1HE6ZjzS:bpomv4p) ([raw file](https://github.com/netsec-ethz/scion/blob/d1c3b0ac04c3e5d16135c9a6928e4296833cf175/python/path_server/core.py#L184)):*

That seems odd. Was this before we had ZK, because ZK is used to do that and I think there we store path segments directly. Will update description.
</blockquote></details>

ZK is used for all down-segments for the local AS. The master-sync approach is used for non-local core/down segments.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280)*

kormat:
Please remove "WIP" from the description :)

---

Reviewed 1 of 6 files at r6.

---

*[python/path_server/core.py, line 66 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGm5qyneFaXl8eqBJW:-KyGm5qyneFaXl8eqBJX:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/path_server/core.py#L66)):*
> ```Python
>         self._segs_to_master = ExpiringDict(1000, 10)
>         self._segs_to_prop = ExpiringDict(
>             1000, 2 * self.config.propagation_time)
> ```

Unnecessary line-wrap.

---

*[python/sciond/sciond.py, line 101 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGmODpT26l_x6iaCpL:-KyGmODpT26l_x6iaCpM:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/sciond/sciond.py#L101)):*
> ```Python
>         """
>         super().__init__("sciond", conf_dir,
>                          prom_export=prom_export, public=[(addr, port)])
> ```

Unnecessary line-wrap.

---

*[python/sciond/sciond.py, line 104 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGmSCFUlgPxpP7rBLo:-KyGmSCFUlgPxpP7rBLp:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/sciond/sciond.py#L104)):*
> ```Python
>         up_labels = {**self._labels, "type": "up"} if self._labels else None
>         down_labels = {**self._labels,
>                        "type": "down"} if self._labels else None
> ```

Unnecessary line-wrap.

---

*[python/sciond/sciond.py, line 106 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGmSkjv49im74VCqST:-KyGmSkjv49im74VCqSU:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/sciond/sciond.py#L106)):*
> ```Python
>                        "type": "down"} if self._labels else None
>         core_labels = {**self._labels,
>                        "type": "core"} if self._labels else None
> ```

Unnecessary line-wrap.

---

*[python/sciond/sciond.py, line 476 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGmc5juIGfdsBj1PcB:-KyGmc5juIGfdsBj1PcC:b-gl47h2) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/sciond/sciond.py#L476)):*
> ```Python
>                     self.requested_paths[key] = RequestState(req.copy())
>                     self._fetch_segments(req)
>                 r = self.requested_paths[key]
> ```

As discussed earlier, this is a race-condition. The simplest fix (that at least will stop crashes) is to do:
```
r = self.requested_paths.get(key)
if r is None:
  # No previous outstanding request
  r = PathSegmentReq...
```

---

*[python/sciond/sciond.py, line 660 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGmvUCoSyVOhoii7zw:-KyGmvUCoSyVOhoii7zx:b-vvxcli) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/sciond/sciond.py#L660)):*
> ```Python
>             return
>         logging.debug(
>             "Sending path request (%s) to [%s]:%s", req.short_desc(), addr, port)
> ```

Unnecessary line-wrap.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-:-KyGmx3yPIzoXb45MSKd:bqz63wz)*

shitz:
Done

---

Review status: 12 of 14 files reviewed at latest revision, 7 unresolved discussions.

---

*[python/path_server/core.py, line 66 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGm5qyneFaXl8eqBJW:-KyGz0zxm_XLMjraQ1nx:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/path_server/core.py#L66)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 101 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGmODpT26l_x6iaCpL:-KyGzDv4tFNKSJDlvliG:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/sciond/sciond.py#L101)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 104 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGmSCFUlgPxpP7rBLo:-KyGzEDxKeK8Uu_G4DOF:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/sciond/sciond.py#L104)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 106 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGmSkjv49im74VCqST:-KyGzEYH5blktA4nCFcs:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/sciond/sciond.py#L106)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 476 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGmc5juIGfdsBj1PcB:-KyGzldAQmB1fx6JXNEw:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/sciond/sciond.py#L476)):*

As discussed earlier, this is a race-condition. The simplest fix (that at least will stop crashes) is to do:
```
r = self.requested_paths.get(key)
if r is None:
  # No previous outstanding request
  r = PathSegmentReq...
```
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 660 at r6](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-KyGmvUCoSyVOhoii7zw:-KyH05Uoq-B2p0d1EDJV:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/4b86c7ac1df9c143f18ea730c3a160ae33b99bc2/python/sciond/sciond.py#L660)):*

Unnecessary line-wrap.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-:-KyH070ry2LiEnRltuU0:bb6un1j)*

kormat:
:lgtm:

---

Reviewed 2 of 2 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1280#-:-KyH0jq8g3gZPY-qOBS1:bnfp4nl)*

PR #1282, opened 2017-10-10
Implement session monitoring for the SIG

kormat:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1282)


shitz:
Reviewed 10 of 27 files at r1.
Review status: 10 of 25 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/ctrl.go, line 136 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9h9b5OTGOpTl7gwkL:-Kw9h9b5OTGOpTl7gwkM:b-c276ab) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/ctrl/ctrl.go#L136)):*
> ```Go
> }
> 
> func (p *Pld) PackPld() (common.RawBytes, error) {
> ```

Alternative:
```
func (p *Pld) PackPld() (common.RawBytes, error) {
	full := make(common.RawBytes, 1 << 16)
	n, err := p.WritePld(full)
	if err != nil {
		return err
	}
	return full[:n], nil
}
```
This allocates too much memory, but avoids the copy. Not sure which one is better.

---

*[go/lib/snet/disp.go, line 25 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9PFIIZlxRY76vf1p9:-Kw9PFIIZlxRY76vf1pA:blxt8gk) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L25)):*
> ```Go
> 
> type DispPkt struct {
> 	B    common.RawBytes
> ```

We usually call this field `Raw`

---

*[go/lib/snet/disp.go, line 32 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9PXK6IfjZ7DTDMrOi:-Kw9PXK6IfjZ7DTDMrOj:bz5q8z5) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L32)):*
> ```Go
> 
> // XXX(kormat): N.B. the DispPkt is reused, applications should make a copy if this is a problem.
> func PktDispatcher(c *Conn, f DispatchFunc) {
> ```

At some point `snet` really needs a `doc.go` explaining the common usage scenarios. A short docstring for `PktDispatcher` would still be nice though.

---

*[go/sig/base/as.go, line 197 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9UObsZex7k95dkLZ9:-Kw9UObsZex7k95dkLZA:b-ahldpf) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/base/as.go#L197)):*
> ```Go
> 			break
> 		case <-ticker.C:
> 			smap := ae.SigMap()
> ```

Hmm if you acquire `ae.Lock()` here you don't need to copy the entire sig map. Copying the sig map also acquires the lock and `ExpireFails()` is not heavy, so I don't see a reason to copy the entire map.

---

*[go/sig/base/as.go, line 209 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9SgzvlxH_bqYHN6TL:-Kw9SgzvlxH_bqYHN6TM:bdbxnd5) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/base/as.go#L209)):*
> ```Go
> 	defer ae.Unlock()
> 	// Clean up sigMgr goroutine.
> 	ae.sigMgrStop <- struct{}{}
> ```

The idiomatic way is to close the channel `close(ae.SigMgrStop)`

---

*[go/sig/disp/disp.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9feFNIYJ1ZGNtoaMB:-Kw9feFNIYJ1ZGNtoaMC:bd8y17b) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L15)):*
> ```Go
> 
> package disp
> ```

A lot of the functionality in this package will be universally useful for a application that need to handle control messages. We should think about how to generalize this and put it into a library (in the future).

---

*[go/sig/disp/disp.go, line 114 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9fLOgvIVRDjp7hscV:-Kw9fLOgvIVRDjp7hscW:b-g8d7a3) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L114)):*
> ```Go
> 	case *mgmt.PollRep:
> 		regPld := &RegPld{Id: msgId, P: pld, Addr: addr}
> 		if pld.Addr == nil || pld.Addr.Ctrl == nil {
> ```

Shouldn't this be handled by whoever registered to receive this message?

---

*[go/sig/egress/dispatcher.go, line 48 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9Q9rrtdI1kQvXiC5o:-Kw9Q9rrtdI1kQvXiC5p:b-qdgf5x) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/dispatcher.go#L48)):*
> ```Go
> 	devName string
> 	devIO   io.ReadWriteCloser
> 	spp     *SyncSession
> ```

Rename to `syncSess`

---

*[go/sig/egress/dispatcher.go, line 62 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9QLdn4xtnpt-sOaJv:-Kw9QLdn4xtnpt-sOaJw:b-g0zsvt) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/dispatcher.go#L62)):*
> ```Go
> 	pktBytesRecv := metrics.PktBytesRecv.WithLabelValues(ed.devName)
> 	pps := ed.spp.Load()
> 	var pp *Session
> ```

Rename to `sess`. Lots of places in this file that still has some form of `pp` in the variable name.

---

*[go/sig/egress/sessmon.go, line 36 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9_ryy_qIjLjET9cz-:-Kw9_ryy_qIjLjET9cz0:b-da3vms) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/sessmon.go#L36)):*
> ```Go
> )
> 
> type sessMonitor struct {
> ```

The understandability of the code in here would greatly benefit from having docstrings for the various fields of `sessMonitor`

---

*[go/sig/egress/sessmon.go, line 70 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9WaBdjV8CiJVPwJSw:-Kw9WaBdjV8CiJVPwJSx:b-61d1lu) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/sessmon.go#L70)):*
> ```Go
> 		select {
> 		case <-sm.sess.sessMonStop:
> 			sm.Info("sessMonitor: stopping")
> ```

Not sure what value this logging statement adds if you already have the one at the end of the function.

---

*[go/sig/egress/sessmon.go, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9aOFdDOo65itSmzv1:-Kw9aOFdDOo65itSmzv2:bef84j9) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/sessmon.go#L128)):*
> ```Go
> }
> 
> func (sm *sessMonitor) updateSig(old *siginfo.Sig) *siginfo.Sig {
> ```

`getNewSig` sounds more intuitive to me (same below for path)

---

*[go/sig/egress/worker.go, line 57 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9g-g1kt_4LelWaoRT:-Kw9g-g1kt_4LelWaoRU:b-ms4qrn) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/worker.go#L57)):*
> ```Go
> type worker struct {
> 	iaString string
> 	pp       *Session
> ```

rename `pp`

---

*[go/sig/egress/worker.go, line 66 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9g6G6tep4J7FNVLqA:-Kw9g6G6tep4J7FNVLqB:b-55i78s) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/worker.go#L66)):*
> ```Go
> }
> 
> func NewEgressWorker(pol *Session) *worker {
> ```

rename `pol`
This can now be called `NewWorker`

---

*[go/sig/egress/worker.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9gK9W3LX5JXbx90Rd:-Kw9gK9W3LX5JXbx90Re:b-fr6sog) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/worker.go#L110)):*
> ```Go
> }
> 
> func (e *worker) processPkt(f *frame, pkt common.RawBytes) error {
> ```

`w* worker`

---

*[go/sig/ingress/dispatcher.go, line 112 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9Nr2lzdQk885TMG99:-Kw9Nr2lzdQk885TMG9A:bc5zd40) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/ingress/dispatcher.go#L112)):*
> ```Go
> // exist yet. Dispatching is done based on source ISD-AS -> source host Addr -> Sess Id.
> func (d *Dispatcher) dispatch(frame *FrameBuf, src *snet.Addr) {
> 	sessId := sigcmn.SessionType((frame.raw[1]))
> ```

Not your fault, but since you are already touching this line change this to `frame.raw[0]`

---

*[go/sig/siginfo/sig.go, line 102 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9Ld2ai9pw44kipnar:-Kw9Ld2bLPChqMRCrc-i:b-8r9pkf) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/siginfo/sig.go#L102)):*
> ```Go
> 	s.statsMutex.Lock()
> 	defer s.statsMutex.Unlock()
> 	if time.Since(s.statsUpdate) > failExpiration {
> ```

The logic is (afaict): if there was no failure with the remote SIG within the last 5 minutes, half the failcount. `ExpireFails` gets called once every second and `maxFail` is 2**16. This means that in ~16 s the failcount would reach 0 again. Imo, if you already wait 5 minutes you might as well reduce the failcount to 0 immediately. It doesn't make a big difference. Alternatively, you start the decay earlier and/or stretch it out over a longer period of time, e.g., you don't half every second but only every 10 s or every minute or so.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

kormat:
Review status: 10 of 25 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/ctrl.go, line 136 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9h9b5OTGOpTl7gwkL:-KwAH3jqmAE2lF8tLnuY:b-awg8j8) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/ctrl/ctrl.go#L136)):*

Alternative:
```
func (p *Pld) PackPld() (common.RawBytes, error) {
	full := make(common.RawBytes, 1 << 16)
	n, err := p.WritePld(full)
	if err != nil {
		return err
	}
	return full[:n], nil
}
```
This allocates too much memory, but avoids the copy. Not sure which one is better.
</blockquote></details>

The big issue is that we can't know ahead of time how large the proto is going to be when encoded :( Also, there's no guarantee that a ctrl payload will fit in 65k, which is why we have a 32bit payload length field.

---

*[go/lib/snet/disp.go, line 25 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9PFIIZlxRY76vf1p9:-KwAHqlbMqL7PE4Bamqk:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L25)):*

We usually call this field `Raw`
</blockquote></details>

Done.

---

*[go/lib/snet/disp.go, line 32 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9PXK6IfjZ7DTDMrOi:-KwAI5c33Bw5nhEwOZnc:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L32)):*

At some point `snet` really needs a `doc.go` explaining the common usage scenarios. A short docstring for `PktDispatcher` would still be nice though.
</blockquote></details>

Done.

---

*[go/sig/base/as.go, line 197 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9UObsZex7k95dkLZ9:-KwAIViWxwevMSgK9GKw:b-lbv2n1) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/base/as.go#L197)):*

Hmm if you acquire `ae.Lock()` here you don't need to copy the entire sig map. Copying the sig map also acquires the lock and `ExpireFails()` is not heavy, so I don't see a reason to copy the entire map.
</blockquote></details>

The copy is pretty cheap, as it's a map of pointers. I _think_ you can get away without needing a copy here, but it's hard to be 100% sure.

---

*[go/sig/base/as.go, line 209 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9SgzvlxH_bqYHN6TL:-KwAItBX-qoYD6GulOyr:bbcb82v) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/base/as.go#L209)):*

The idiomatic way is to close the channel `close(ae.SigMgrStop)`
</blockquote></details>

Yeah, but that makes things more complex when you have goroutines you may want to _restart_, as you then need to ensure that the channel is re-created (and you can't tell locally if a channel is open or not).

---

*[go/sig/disp/disp.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9feFNIYJ1ZGNtoaMB:-KwAJYgwz3zFPtdmHWR_:b-pbdqz3) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L15)):*

A lot of the functionality in this package will be universally useful for a application that need to handle control messages. We should think about how to generalize this and put it into a library (in the future).
</blockquote></details>

Agreed.

---

*[go/sig/disp/disp.go, line 114 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9fLOgvIVRDjp7hscV:-KwAJw6NWWuV0w2Fr24M:bni56nv) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L114)):*

Shouldn't this be handled by whoever registered to receive this message?
</blockquote></details>

If we were trying to make this more generic, quite possibly, but in the meantime it seems to be functionally equivalent.

---

*[go/sig/egress/dispatcher.go, line 48 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9Q9rrtdI1kQvXiC5o:-KwAKFKUJLxjBXAx7U5o:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/dispatcher.go#L48)):*

Rename to `syncSess`
</blockquote></details>

Done.

---

*[go/sig/egress/dispatcher.go, line 62 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9QLdn4xtnpt-sOaJv:-KwAKa1fnJ6tpO3qncED:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/dispatcher.go#L62)):*

Rename to `sess`. Lots of places in this file that still has some form of `pp` in the variable name.
</blockquote></details>

Done.

---

*[go/sig/egress/sessmon.go, line 36 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9_ryy_qIjLjET9cz-:-KwAMJpxdfy_JYzcbQ62:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/sessmon.go#L36)):*

The understandability of the code in here would greatly benefit from having docstrings for the various fields of `sessMonitor`
</blockquote></details>

Done.

---

*[go/sig/egress/sessmon.go, line 70 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9WaBdjV8CiJVPwJSw:-KwAMRQpo6t8ICB473Bs:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/sessmon.go#L70)):*

Not sure what value this logging statement adds if you already have the one at the end of the function.
</blockquote></details>

Done.

---

*[go/sig/egress/sessmon.go, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9aOFdDOo65itSmzv1:-KwAMc4pGvKc61VVkJ5T:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/sessmon.go#L128)):*

`getNewSig` sounds more intuitive to me (same below for path)
</blockquote></details>

Done.

---

*[go/sig/egress/worker.go, line 57 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9g-g1kt_4LelWaoRT:-KwAMopm6OhsyT4z9cJm:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/worker.go#L57)):*

rename `pp`
</blockquote></details>

Done.

---

*[go/sig/egress/worker.go, line 66 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9g6G6tep4J7FNVLqA:-KwAMtpH2_D0CheJaYT5:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/worker.go#L66)):*

rename `pol`
This can now be called `NewWorker`
</blockquote></details>

Done.

---

*[go/sig/egress/worker.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9gK9W3LX5JXbx90Rd:-KwAN2KsIAx0YE5ez2II:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/worker.go#L110)):*

`w* worker`
</blockquote></details>

Done.

---

*[go/sig/ingress/dispatcher.go, line 112 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9Nr2lzdQk885TMG99:-KwAN6rPSiJnlzvpjE2Z:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/ingress/dispatcher.go#L112)):*

Not your fault, but since you are already touching this line change this to `frame.raw[0]`
</blockquote></details>

Done.

---

*[go/sig/siginfo/sig.go, line 102 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9Ld2ai9pw44kipnar:-KwAO-KXlWrkdYRIB3p2:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/siginfo/sig.go#L102)):*

The logic is (afaict): if there was no failure with the remote SIG within the last 5 minutes, half the failcount. `ExpireFails` gets called once every second and `maxFail` is 2**16. This means that in ~16 s the failcount would reach 0 again. Imo, if you already wait 5 minutes you might as well reduce the failcount to 0 immediately. It doesn't make a big difference. Alternatively, you start the decay earlier and/or stretch it out over a longer period of time, e.g., you don't half every second but only every 10 s or every minute or so.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

shitz:
Reviewed 1 of 27 files at r1.
Review status: 10 of 25 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/ctrl.go, line 136 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9h9b5OTGOpTl7gwkL:-KwAOg63S53DqMM5GpC4:b-wrua1w) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/ctrl/ctrl.go#L136)):*

The big issue is that we can't know ahead of time how large the proto is going to be when encoded :( Also, there's no guarantee that a ctrl payload will fit in 65k, which is why we have a 32bit payload length field.
</blockquote></details>

Oh ok . I thought we have 65k max packet size anyway.

---

*[go/sig/base/as.go, line 197 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9UObsZex7k95dkLZ9:-KwAPb5gnfbH9SaJHKbv:bh16fdr) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/base/as.go#L197)):*

The copy is pretty cheap, as it's a map of pointers. I _think_ you can get away without needing a copy here, but it's hard to be 100% sure.
</blockquote></details>

You can definitely get away with it if you acquire the lock (which you do anyway when you copy the map). All I'm saying is if you do this anyway, why still copy the map? Just acquire the lock here, don't copy the map and release the lock at the end of the `case`.

---

*[go/sig/base/as.go, line 209 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9SgzvlxH_bqYHN6TL:-KwAQDiFjnX72Fq574Jb:bs6dr8t) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/base/as.go#L209)):*

Yeah, but that makes things more complex when you have goroutines you may want to _restart_, as you then need to ensure that the channel is re-created (and you can't tell locally if a channel is open or not).
</blockquote></details>

So the same `sigMgr` instance might be restarted after being stopped?

---

*[go/sig/disp/disp.go, line 102 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwATaTNXq1l-mXRJQDe:-KwATaTNXq1l-mXRJQDf:b-l6mwxj) ([raw file](https://github.com/netsec-ethz/scion/blob/88f4ed9cc6d279284255fe35b5cd6d43236a0ade/go/sig/disp/disp.go#L102)):*
> ```Go
> 	dm.Lock()
> 	defer dm.Unlock()
> 	//log.Debug("Got sig ctrl packet", "pld", pld)
> ```

remove

---

*[go/sig/egress/worker.go, line 132 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwASKxCJzBXozXpsVbA:-KwASKxCJzBXozXpsVbB:b-dycq) ([raw file](https://github.com/netsec-ethz/scion/blob/88f4ed9cc6d279284255fe35b5cd6d43236a0ade/go/sig/egress/worker.go#L132)):*
> ```Go
> 
> // Return false if the ringbuf is closed.
> func (w *worker) Read(block bool) bool {
> ```

`Read` and `Write` are currently only used within this file and could be unexported.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

scrye:
Review status: 8 of 27 files reviewed at latest revision, 18 unresolved discussions.

---

*[go/lib/snet/disp.go, line 24 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9m7_E9CJdJd6g933y:-Kw9m7_E9CJdJd6g933z:b-lvnqvn) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L24)):*
> ```Go
> )
> 
> type DispPkt struct {
> ```

This type of functionality is very specific and should not be part of the generic networking library. I think moving it to a messaging library of its own is better. It also needs a short bit of documentation on what it does exactly.

---

*[go/lib/snet/disp.go, line 32 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9PXK6IfjZ7DTDMrOi:-Kw9loDggh0QM91X2sHE:b6eafjh) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L32)):*

Done.
</blockquote></details>

@shitz: `snet` already has some documentation (complete with broken formatting). The scenarios are the same as `net.Conn` except you need to run `Init` first. Is there something you feel is missing right now?

---

*[go/lib/snet/disp.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA0ff-_4GWsWtva7ID:-KwA0ff-_4GWsWtva7IE:b-xunxyt) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L35)):*
> ```Go
> 	defer liblog.LogPanicAndExit()
> 	var err error
> 	var l int
> ```

I'd avoid `l` as variable name, it looks too close to `1`. Maybe `n`?

---

*[go/lib/snet/disp.go, line 36 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA066DkAG8k6mck3yl:-KwA066DkAG8k6mck3ym:bymgk7z) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L36)):*
> ```Go
> 	var err error
> 	var l int
> 	dp := &DispPkt{B: make(common.RawBytes, common.MaxMTU)}
> ```

I find this capacity reset / reslicing pattern to be a bit difficult to follow.

Suggested simplified variant:
```
dp := &DispPkt{}
b := make(common.RawBytes, common.MaxMTU)
for {
    n, a, err := c.ReadFromSCION(b)
    if err != nil {
       log
    }
    dp.B, dp.Addr = b[:n], a
    f(dp)
}
```

---

*[go/sig/disp/disp.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9kcNsujl46Bac432d:-Kw9kcNtvWqNswaUcwK4:b-r3gdkw) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L15)):*
> ```Go
> 
> package disp
> ```

The naming here is a bit confusing because we have a lot of dispatchers. Maybe change to `ctrldisp`?

---

*[go/sig/disp/disp.go, line 38 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwAaUGEvUEPxlJ6Z-Az:-KwAaUGEvUEPxlJ6Z-B-:b-r81kzp) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L38)):*
> ```Go
> 
> const (
> 	RegPollRep = iota
> ```

This isn't the correct type, should be `RegPollRep RegType = iota`.

---

*[go/sig/disp/disp.go, line 58 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwANV0TxGKsTUU6aX4T:-KwANV0TxGKsTUU6aX4U:b-1juulz) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L58)):*
> ```Go
> 
> var (
> 	PollReqC   = make(RegPldChan, 1)
> ```

Is there a reason why this is limited to only one outstanding request?

Also, I'd make this only part of `Dispatcher` as they are tightly connected anyway (so external packages would `disp.Dispatcher.PollReqC`).

---

*[go/sig/disp/disp.go, line 65 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwANrFKo0jocphfPhHh:-KwANrFKo0jocphfPhHi:bnle36b) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L65)):*
> ```Go
> 	sync.RWMutex
> 	pollReq RegPldChan
> 	pollRep map[string]RegPldChan
> ```

Add a comment describing the key format.

---

*[go/sig/egress/session.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA_C5z5gWUV3fwMPHr:-KwA_C5z5gWUV3fwMPHs:b-btpxmx) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/session.go#L123)):*
> ```Go
> }
> 
> func connReader(conn *snet.Conn, name string) {
> ```

Add a comment specifying the purpose of this goroutine.

---

*[go/sig/egress/sessmon.go, line 47 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwAlwsVZ5DiMrX39mYO:-KwAlwsVZ5DiMrX39mYP:b-70zrqt) ([raw file](https://github.com/netsec-ethz/scion/blob/8eaeb469c8647758d0134aedc6e52145bb5d1366/go/sig/egress/sessmon.go#L47)):*
> ```Go
> 	pool *pathmgr.SyncPaths
> 	// the pool of paths this session is currently using, frequently refreshed from pool.
> 	sessPool sessPathPool
> ```

`sessPool` makes it sound like it is a pool of sessions. Maybe use `sessPathPool` or just `paths`?

---

*[go/sig/egress/sessmon.go, line 99 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwAkvMcIXaQb2ih-Q6c:-KwAkvMcIXaQb2ih-Q6d:bjigb77) ([raw file](https://github.com/netsec-ethz/scion/blob/8eaeb469c8647758d0134aedc6e52145bb5d1366/go/sig/egress/sessmon.go#L99)):*
> ```Go
> }
> 
> func (sm *sessMonitor) checkRemote() {
> ```

This name is misleading, because the function those a lot more than just checking.

---

*[go/sig/egress/sessmon.go, line 125 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwAwOeqprQHGrsE5zE5:-KwAwOeqprQHGrsE5zE6:b-z37atk) ([raw file](https://github.com/netsec-ethz/scion/blob/8eaeb469c8647758d0134aedc6e52145bb5d1366/go/sig/egress/sessmon.go#L125)):*
> ```Go
> 			sm.needUpdate = true
> 		}
> 		poolPaths := sm.pool.Load()
> ```

Why are paths reloaded here? Hasn't `sm.sessPool` fetched the exact same data above?

---

*[go/sig/egress/sessmon.go, line 188 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwB-L9R0uLRyW70tvbP:-KwB-L9R0uLRyW70tvbQ:b-dx80kj) ([raw file](https://github.com/netsec-ethz/scion/blob/8eaeb469c8647758d0134aedc6e52145bb5d1366/go/sig/egress/sessmon.go#L188)):*
> ```Go
> 		return
> 	}
> 	_, err = sm.sess.conn.WriteToSCION(raw, sm.smRemote.Sig.CtrlSnetAddr())
> ```

This is dangerous without arming a deadline, perhaps ~ time to next tick.

---

*[go/sig/egress/sesspath.go, line 50 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwAuMXQoZSNwkDlJOlj:-KwAuMXQoZSNwkDlJOlk:b-ogzs9q) ([raw file](https://github.com/netsec-ethz/scion/blob/8eaeb469c8647758d0134aedc6e52145bb5d1366/go/sig/egress/sesspath.go#L50)):*
> ```Go
> 
> func (spp sessPathPool) update(aps pathmgr.AppPathSet) {
> 	entryMap := make(map[pathmgr.PathKey]*sciond.PathReplyEntry)
> ```

Is this `entryMap` needed? It is never written to, and the loops below would look approximately the same with `aps`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

kormat:
Review status: 8 of 27 files reviewed at latest revision, 18 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/ctrl.go, line 136 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9h9b5OTGOpTl7gwkL:-KwEtm1Cl355iLkkTUIq:b-ds9b3c) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/ctrl/ctrl.go#L136)):*

Oh ok . I thought we have 65k max packet size anyway.
</blockquote></details>

Max packet size != max message size, when you start considering stream-oriented protocols.

---

*[go/lib/snet/disp.go, line 24 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9m7_E9CJdJd6g933y:-KwEgceXI8AlC-mnzvJZ:b-59bjuc) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L24)):*

This type of functionality is very specific and should not be part of the generic networking library. I think moving it to a messaging library of its own is better. It also needs a short bit of documentation on what it does exactly.
</blockquote></details>

I agree this isn't the right place for it. The msg dispatcher stuff in go/sig/disp/ also needs to be refactored, and together i think there's a decent message dispatch library to be made in a future PR.

---

*[go/lib/snet/disp.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA0ff-_4GWsWtva7ID:-KwEgvAK1Q1eveq-FVes:bqdm4mg) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L35)):*

I'd avoid `l` as variable name, it looks too close to `1`. Maybe `n`?
</blockquote></details>

`n` is better :) Done.

---

*[go/lib/snet/disp.go, line 36 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA066DkAG8k6mck3yl:-KwEhN5i0DjLn2pO8Ckv:bxbed0g) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L36)):*

I find this capacity reset / reslicing pattern to be a bit difficult to follow.

Suggested simplified variant:
```
dp := &DispPkt{}
b := make(common.RawBytes, common.MaxMTU)
for {
    n, a, err := c.ReadFromSCION(b)
    if err != nil {
       log
    }
    dp.B, dp.Addr = b[:n], a
    f(dp)
}
```

</blockquote></details>

That also works, but the slice/reslice thing is idiomatic in Go. It also makes it clearer that it's the same buffer being re-used, as there's no extra indirection involved.

---

*[go/sig/base/as.go, line 197 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9UObsZex7k95dkLZ9:-KwEiMf-pvHVbN3b8Irz:bevwbnz) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/base/as.go#L197)):*

You can definitely get away with it if you acquire the lock (which you do anyway when you copy the map). All I'm saying is if you do this anyway, why still copy the map? Just acquire the lock here, don't copy the map and release the lock at the end of the `case`.
</blockquote></details>

If sig.ExpireFails (or this case in general) ever becomes expensive, then this will break things, as the sessMon instance will block trying to get an updated map.

---

*[go/sig/base/as.go, line 209 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9SgzvlxH_bqYHN6TL:-KwEihb6eiQ6dNmOuU2D:b-p87lnv) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/base/as.go#L209)):*

So the same `sigMgr` instance might be restarted after being stopped?
</blockquote></details>

Yup. When `DelSession` is added, it will stop the egress worker and sigmgr if there are no sessions left. A later call to `AddSession` will start them both again.

---

*[go/sig/disp/disp.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9kcNsujl46Bac432d:-KwEj0w9TYRXW8UeC670:b-hf7pvg) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L15)):*

The naming here is a bit confusing because we have a lot of dispatchers. Maybe change to `ctrldisp`?
</blockquote></details>

Agreed, i suggest leaving it for this PR and fixing it in a later PR that refactors go/sig/snet/disp.go and this package into something more general and self-contained.

---

*[go/sig/disp/disp.go, line 38 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwAaUGEvUEPxlJ6Z-Az:-KwEjHpFpHvJFkLzKZpZ:b-9ajtt7) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L38)):*

This isn't the correct type, should be `RegPollRep RegType = iota`.
</blockquote></details>

Good catch, thanks :)

---

*[go/sig/disp/disp.go, line 58 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwANV0TxGKsTUU6aX4T:-KwEqCjW-EO-DxsLcc4c:bh6jquh) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L58)):*

Is there a reason why this is limited to only one outstanding request?

Also, I'd make this only part of `Dispatcher` as they are tightly connected anyway (so external packages would `disp.Dispatcher.PollReqC`).
</blockquote></details>

I kept it to 1 to start with as there was enough new complexity that i didn't want buffered channels makings things more complex. Increased it now.

---

*[go/sig/disp/disp.go, line 65 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwANrFKo0jocphfPhHh:-KwEqbRactYUwCLGc1Tn:b-thh16w) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/disp/disp.go#L65)):*

Add a comment describing the key format.
</blockquote></details>

I cheated and made it a type.

---

*[go/sig/disp/disp.go, line 102 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwATaTNXq1l-mXRJQDe:-KwEqfiM08pPPpUA_TRw:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/88f4ed9cc6d279284255fe35b5cd6d43236a0ade/go/sig/disp/disp.go#L102)):*

remove
</blockquote></details>

Done.

---

*[go/sig/egress/session.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA_C5z5gWUV3fwMPHr:-KwErNabYQqRCAO6Nqne:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/session.go#L123)):*

Add a comment specifying the purpose of this goroutine.
</blockquote></details>

Done.

---

*[go/sig/egress/sessmon.go, line 47 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwAlwsVZ5DiMrX39mYO:-KwEseFLi29yFcLVdY7B:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/8eaeb469c8647758d0134aedc6e52145bb5d1366/go/sig/egress/sessmon.go#L47)):*

`sessPool` makes it sound like it is a pool of sessions. Maybe use `sessPathPool` or just `paths`?
</blockquote></details>

Done.

---

*[go/sig/egress/sessmon.go, line 99 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwAkvMcIXaQb2ih-Q6c:-KwEr_NDJv8-KXz1iadM:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/8eaeb469c8647758d0134aedc6e52145bb5d1366/go/sig/egress/sessmon.go#L99)):*

This name is misleading, because the function those a lot more than just checking.
</blockquote></details>

Done.

---

*[go/sig/egress/sessmon.go, line 125 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwAwOeqprQHGrsE5zE5:-KwErfzRu-oyJVYGsxsF:b-k2rwtz) ([raw file](https://github.com/netsec-ethz/scion/blob/8eaeb469c8647758d0134aedc6e52145bb5d1366/go/sig/egress/sessmon.go#L125)):*

Why are paths reloaded here? Hasn't `sm.sessPool` fetched the exact same data above?
</blockquote></details>

Good catch. Fixed.

---

*[go/sig/egress/sesspath.go, line 50 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwAuMXQoZSNwkDlJOlj:-KwEsw6EIWNwXFvSRjoa:bxmiggo) ([raw file](https://github.com/netsec-ethz/scion/blob/8eaeb469c8647758d0134aedc6e52145bb5d1366/go/sig/egress/sesspath.go#L50)):*

Is this `entryMap` needed? It is never written to, and the loops below would look approximately the same with `aps`.
</blockquote></details>

For some reason i thought AppPathSet was a slice. Done.

---

*[go/sig/egress/worker.go, line 132 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwASKxCJzBXozXpsVbA:-KwEtkfX2dqeBRQQmB7e:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/88f4ed9cc6d279284255fe35b5cd6d43236a0ade/go/sig/egress/worker.go#L132)):*

`Read` and `Write` are currently only used within this file and could be unexported.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

shitz:
Review status: 8 of 27 files reviewed at latest revision, 19 unresolved discussions.

---

*[go/border/ifstate/ifstate.go, line 63 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwEgaUxTekUCjZ2PMoQ:-KwEgaUxTekUCjZ2PMoR:b-9307tk) ([raw file](https://github.com/netsec-ethz/scion/blob/225ccada5b016bea345a453cbf4fb82f0bc77e6f/go/border/ifstate/ifstate.go#L63)):*
> ```Go
> 	// Convert IFState infos to map
> 	m := make(map[common.IFIDType]State, len(ifStates.Infos))
> 	/*
> ```

Revert this change. We should 

---

*[go/lib/snet/disp.go, line 24 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9m7_E9CJdJd6g933y:-KwEW8U6hMcPikkF3fx9:b-hnx7c1) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L24)):*

This type of functionality is very specific and should not be part of the generic networking library. I think moving it to a messaging library of its own is better. It also needs a short bit of documentation on what it does exactly.
</blockquote></details>

I agree with this. We can probably also move the generic part of `sig/disp/disp.go` in there.

---

*[go/lib/snet/disp.go, line 32 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-Kw9PXK6IfjZ7DTDMrOi:-KwEVaN_J6sM89cBSLJ9:b-oe3lae) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L32)):*

@shitz: `snet` already has some documentation (complete with broken formatting). The scenarios are the same as `net.Conn` except you need to run `Init` first. Is there something you feel is missing right now?
</blockquote></details>

I was specifically talking about the use-case of the packet dispatcher, since this is a completely new way of using `snet`. You are probably right, though, that this shouldn't be in `snet` but rather in a higher-level lib on top of `snet`.

---

*[go/sig/egress/session.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA_C5z5gWUV3fwMPHr:-KwEWtPJXn0yE39wCQH0:bslm7m2) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/session.go#L123)):*

Add a comment specifying the purpose of this goroutine.
</blockquote></details>

Are we even planning on keeping this around? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

kormat:
Review status: 8 of 26 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/sig/egress/session.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA_C5z5gWUV3fwMPHr:-KwEvZjWq8oGVFdv6F-3:b1qqv6x) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/session.go#L123)):*

Are we even planning on keeping this around? 
</blockquote></details>

@shitz - my plan is to have PktDispatcher handle this use-case too, and use that instead.

---

*[go/border/ifstate/ifstate.go, line 63 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwEgaUxTekUCjZ2PMoQ:-KwEwQm_WJaXcnK8cdka:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/225ccada5b016bea345a453cbf4fb82f0bc77e6f/go/border/ifstate/ifstate.go#L63)):*

Revert this change. We should 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

scrye:
Review status: 8 of 26 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[go/lib/snet/disp.go, line 36 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA066DkAG8k6mck3yl:-KwEuiNGq5I1YVoFOoZS:bkmtu7u) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/lib/snet/disp.go#L36)):*

That also works, but the slice/reslice thing is idiomatic in Go. It also makes it clearer that it's the same buffer being re-used, as there's no extra indirection involved.
</blockquote></details>

I generally tend to shy away from changing the state of an object inside of another object (makes tracking state difficult), but it's a matter of preference I guess.

---

*[go/sig/egress/session.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA_C5z5gWUV3fwMPHr:-KwEwX33fi7EKtqL4N-r:b26z7dm) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/session.go#L123)):*

@shitz - my plan is to have PktDispatcher handle this use-case too, and use that instead.
</blockquote></details>

I think it's safer to have it (or something else doing the same thing). It doesn't really do anything except use up a little bit of memory. During normal operation it will just stay blocked.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

scrye:
Review status: 8 of 26 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[go/sig/egress/session.go, line 123 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwA_C5z5gWUV3fwMPHr:-KwExCYIssoYavb6mi4J:bgfa9cu) ([raw file](https://github.com/netsec-ethz/scion/blob/ee9ff6cc7090146d4e3c253165d75e55dc4131be/go/sig/egress/session.go#L123)):*

I think it's safer to have it (or something else doing the same thing). It doesn't really do anything except use up a little bit of memory. During normal operation it will just stay blocked.
</blockquote></details>

Ugh, got sniped on the last comment. It was in response to @shitz. In any case, handling the use case is the important part, and I'm in favor of keeping it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

kormat:
Review status: 8 of 26 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[go/sig/egress/sessmon.go, line 188 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwB-L9R0uLRyW70tvbP:-KwExRpX8ZpH86H4tp8R:b47aqe5) ([raw file](https://github.com/netsec-ethz/scion/blob/8eaeb469c8647758d0134aedc6e52145bb5d1366/go/sig/egress/sessmon.go#L188)):*

This is dangerous without arming a deadline, perhaps ~ time to next tick.
</blockquote></details>

Missed this earlier, added with a 100ms timeout.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

shitz:
:lgtm:

---

Reviewed 6 of 27 files at r1, 3 of 8 files at r2, 1 of 5 files at r3, 2 of 9 files at r4, 6 of 7 files at r5, 2 of 3 files at r6.

---

*[go/sig/egress/session.go, line 138 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwEuvP1VlZXV_cBRFlq:-KwEuvP1VlZXV_cBRFlr:b-b3u01s) ([raw file](https://github.com/netsec-ethz/scion/blob/3fca74bf7d4d9a5b5e4b18a84739d05c0d328e1a/go/sig/egress/session.go#L138)):*
> ```Go
> 			continue
> 		}
> 		log.Debug("connReader", "name", name, "src", src, "raw", buf[:l])
> ```

Should we actually log this as `Warn`? Since this is unexpected behavior it should probably be more than just debug information.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-:-KwFGEJzdSg9PRM5gwxQ:bnfp4nl)*

shitz:
Reviewed 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

scrye:
:lgtm:

---

Reviewed 1 of 9 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-:-KwFH8wOr-Vy9ic0WA8L:bnfp4nl)*

scrye:
Reviewed 1 of 9 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

kormat:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[go/sig/egress/session.go, line 138 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1282#-KwEuvP1VlZXV_cBRFlq:-KwFGfnG9ofI9YaYmuOg:b606jye) ([raw file](https://github.com/netsec-ethz/scion/blob/3fca74bf7d4d9a5b5e4b18a84739d05c0d328e1a/go/sig/egress/session.go#L138)):*

Should we actually log this as `Warn`? Since this is unexpected behavior it should probably be more than just debug information.
</blockquote></details>

Sure, we can do that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1282)*

PR #1290, opened 2017-10-23
Rewrite path manager

scrye:
Simplified all the convoluted logic in the path manager. Complexity stays the same, but data structures are simpler. Also improved synchronization to have cleaner delimitation of goroutines.

Also:
  - Improve path manager coverage by mocking SCIOND

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1290)


kormat:
Reviewed 16 of 19 files at r1, 4 of 4 files at r2.
Review status: 16 of 19 files reviewed at latest revision, 29 unresolved discussions.

---

*[go/lib/pathmgr/apppath_test.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCRn1m5qG4KW75s-pR:-KxCRn1m5qG4KW75s-pS:bjfoopj) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/apppath_test.go#L15)):*
> ```Go
> 
> package pathmgr
> ```

This file is now ~empty?

---

*[go/lib/pathmgr/cache.go, line 29 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCW_pfYmreL5TnNmK7:-KxCW_pfYmreL5TnNmK8:b-aacn0j) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L29)):*
> ```Go
> )
> 
> type cacheEntry struct {
> ```

It looks like this is specific to a given IA pair? If so it would be good to add a docstring mentioning that.

---

*[go/lib/pathmgr/cache.go, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCWJFep-Z8rvADxIeA:-KxCWJFep-Z8rvADxIeB:b-u971s9) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L34)):*
> ```Go
> 	// Set of watched filters
> 	fs filterSet
> 	// Set to true if the paths are watched (i.e., periodically refreshed from SCIOND)
> ```

This implies that if `registered` is `false`, then `fs` must be nil - is that correct?

---

*[go/lib/pathmgr/cache.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCWEl3oteS2mJYd3md:-KxCWEl3oteS2mJYd3me:b3zh3ji) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L35)):*
> ```Go
> 	fs filterSet
> 	// Set to true if the paths are watched (i.e., periodically refreshed from SCIOND)
> 	registered bool
> ```

Given the change in terms, would `watched` be a better name?

---

*[go/lib/pathmgr/cache.go, line 44 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCWxEN5-b9qRHsSU17:-KxCWxEN5-b9qRHsSU18:bj0fskw) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L44)):*
> ```Go
> 	// Keep one entry for each src-dst pair
> 	m map[IAKey]*cacheEntry
> 	// When an app reads a path older than maxAge, SCIOND is requeried to get fresh paths
> ```

"re-queried" ("required" isn't wrong, it just looks very much like a typo of "required" :)

---

*[go/lib/pathmgr/cache.go, line 86 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCYxW4BwIv6Rj-TfsS:-KxCYxW5-Lvyke238UVq:b-l3k4d3) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L86)):*
> ```Go
> 			sp: NewSyncPaths(),
> 		}
> 		entry.fs["*"] = pf
> ```

This new pathFilter isn't populated.

---

*[go/lib/pathmgr/cache.go, line 113 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCZjMX-DTw6vvRcgt_:-KxCZjMX-DTw6vvRcgta:b-ejhqw5) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L113)):*
> ```Go
> }
> 
> // addFilteredSP adds periodic lookups for paths between src and dst, filtered
> ```

There does not appear to be a way to remove a pathFilter.

---

*[go/lib/pathmgr/cache.go, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCYZ3Vt7nQIvX_qgJd:-KxCYZ3Vt7nQIvX_qgJe:by8nrfe) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L115)):*
> ```Go
> // addFilteredSP adds periodic lookups for paths between src and dst, filtered
> // according to filter.
> func (c *cache) addFilteredSP(src, dst *addr.ISD_AS, filter *PathPredicate) error {
> ```

I'm not clear why there is both `addFilteredSP` and `watch`, with very different semantics.  It looks like a caller has to call `watch` and then `addFilteredSP`. Why not just pass in a `*PathPredicate` into `watch`, and handle both at the same time?

---

*[go/lib/pathmgr/cache.go, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCZ1e6Z0PhdSZq6Eh3:-KxCZ1e6Z0PhdSZq6Eh4:b-8a3qpp) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L128)):*
> ```Go
> 		}
> 		pf.update(entry.aps)
> 		entry.fs[filter.String()] = pf
> ```

As discussed on a previous PR, using a filter string as a key is not a good idea (to recap: if you have an app that has two filters, both of which happen to have the same constraints currently, this code can't distinguish between them. If the app wants to update one of the filters, it would have to keep its own reference counts so as to know when it could remove the pathFilter, which is a lot of complexity. It's far better to use a filter identifier (e.g. name) that's independent of the actual filter contents).

---

*[go/lib/pathmgr/cache.go, line 168 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxC_BQ6MSlZoXG3fvD4:-KxC_BQ6MSlZoXG3fvD5:b-pip80u) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L168)):*
> ```Go
> 	entry, ok := c.getEntry(src, dst)
> 	if !ok {
> 		log.Warn("Attempted to revoke known path, but no path set found", "path",
> ```

s/revoke/remove/ ?

---

*[go/lib/pathmgr/defines.go, line 58 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCTI8N8_T8PL6gQKnl:-KxCTI8N8_T8PL6gQKnm:bvrjy03) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/defines.go#L58)):*
> ```Go
> }
> 
> // Function update changes paths within the SyncPaths object of pf to the ones
> ```

The wording is a bit difficult to follow. Suggestion:
```
update replaces the pathFilter's paths with those from aps, filtered by the pathFilter's PathPredicate (if set).
```

---

*[go/lib/pathmgr/pathmgr.go, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCbSBKg2EVPml-MLBA:-KxCbSBKg2EVPml-MLBB:b-cio5jf) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L115)):*
> ```Go
> 	// expired between the retrieval above and the access (improbable), or no
> 	// paths are available.
> 	if aps, ok := r.cache.getAPS(src, dst); ok {
> ```

I think this can just be `return r.cache.getAPS(src, dst)`

---

*[go/lib/pathmgr/pathmgr.go, line 165 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCcbNrAriGEyDuyfGB:-KxCcbNrAriGEyDuyfGC:b-zf7oft) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L165)):*
> ```Go
> // WatchFilter also adds pair src-dst to the list of tracked paths (if it
> // wasn't already tracked).
> func (r *PR) WatchFilter(src, dst *addr.ISD_AS, filter *PathPredicate) (*SyncPaths, error) {
> ```

I'm not sure that there is a need to have filtered vs non-filtered so separated. If the base `cache` calls accepted filters and DTRT depending on whether one was supplied or not, then this code becomes a lot simpler. `Watch()` becomes `return r.WatchFilter(src, dst, nil)`

---

*[go/lib/pathmgr/pathmgr.go, line 201 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCeEiFm5-9TUo9unxW:-KxCeEiFm5-9TUo9unxX:befb25x) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L201)):*
> ```Go
> 	// Revoke asynchronously to prevent cases where waiting on SCIOND
> 	// blocks the data plane receiver which got the SCMP packet.
> 	go func() {
> ```

Any code run in a go routine should have `defer liblog.LogPanicAndExit()` at the very top.

---

*[go/lib/pathmgr/pathmgr.go, line 204 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCdbyFpx7G61ReS4Kc:-KxCdbyFpx7G61ReS4Kd:b-fyds7) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L204)):*
> ```Go
> 		r.Lock()
> 		defer r.Unlock()
> 		r.revoke(revInfo)
> ```

Is/will `revoke` be called from anywhere else? If not, the lock/unlock could be moved into it, and this simply becomes `go r.revoke(revInfo)`.

---

*[go/lib/pathmgr/pathmgr.go, line 243 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCe1hFDEASWRA2ixvp:-KxCe1hFDEASWRA2ixvq:b-cx44gq) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L243)):*
> ```Go
> }
> 
> // resolver receives requests from PR and answers them by contacting SCIOND.
> ```

This file is already long enough - i would move this to its own file.

---

*[go/lib/pathmgr/pathmgr.go, line 255 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCeUguvI-VNCoozp8e:-KxCeUguvI-VNCoozp8f:befb25x) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L255)):*
> ```Go
> }
> 
> func (r *resolver) run() {
> ```

Any code run in a go routine should have `defer liblog.LogPanicAndExit()` at the very top.

---

*[go/lib/pathmgr/pathmgr.go, line 263 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCfwoe9N5yyx7UvReB:-KxCfwoe9N5yyx7UvReC:b-d510b1) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L263)):*
> ```Go
> 			close(request.done)
> 		case reqMonitor:
> 			if !r.cache.isWatched(request.src, request.dst) {
> ```

This makes it impossible to unwatch, as the resolver will keep re-watching it.

---

*[go/lib/pathmgr/pathmgr.go, line 267 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCgB-cgedWr8IDoNZ7:-KxCgB-cgedWr8IDoNZ8:bijvedl) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L267)):*
> ```Go
> 			}
> 			r.cache.update(request.src, request.dst, aps)
> 			// Create new request, without done channel
> ```

Isn't it simpler to just:
```
if request.done != nil {
   close(request.done)
   request.done = nil
}
// reuse request
```
Note that Watch/Query would need to change so that they keep a direct reference to the chan.

---

*[go/lib/pathmgr/pathmgr.go, line 273 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxClGIj0VFkOJ53DbQ3:-KxClGIj0VFkOJ53DbQ4:b-8qp1qe) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L273)):*
> ```Go
> 				reqType: reqMonitor,
> 			}
> 			time.AfterFunc(r.refireInterval, func() {
> ```

Using a fixed value here is not good. Here's two cases:
- no paths were found (maybe a critical interface got revoked recently). In this case, you would not want to wait for however long `refireInterval` is before trying again (think negative TTL for DNS).
- if some/all of the paths retrieved are going to expire before the next `refireInterval` (or even if they're getting close).

There's really ~4 timers involved here:
- normal refresh interval (when path lookups return paths) - can be a fixed value given at init.
- failure refresh interval (when path lookups don't return paths)  - can be a fixed value given at init.
- the individual TTL for a given path - entirely dependent on the paths you receive.
- optionally a "max-age" for a given path, which is calculated as `min(pathTTL, DefMaxAge)` - combination of both.

---

*[go/lib/pathmgr/pathmgr.go, line 281 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCgU4se9XqqHZMM0qP:-KxCgU4se9XqqHZMM0qQ:b-rrgijs) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L281)):*
> ```Go
> 			}
> 		default:
> 			log.Warn("Unknown query type", "type", request.reqType)
> ```

I'd log the entire request, as debugging this otherwise is going to hurt a lot.

---

*[go/lib/pathmgr/pathmgr.go, line 288 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCggWMGc1U9s0GAwU5:-KxCggWMGc1U9s0GAwU6:b-insmc1) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L288)):*
> ```Go
> // lookup queries SCIOND, blocking while waiting for the response.
> func (r *resolver) lookup(src, dst *addr.ISD_AS) AppPathSet {
> 	reply, err := r.sciondConn.Paths(dst, src, numReqPaths, sciond.PathReqFlags{})
> ```

Hmm. What's the plan for supporting sciond path request flags? Maybe they should be in the `resolverRequest` struct.

---

*[go/lib/pathmgr/pathmgr.go, line 296 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCgyhT4cEVWn692SBJ:-KxCgyhT4cEVWn692SBK:b9h7jca) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L296)):*
> ```Go
> 		// SCIOND internal error, return 0 paths set
> 		log.Error("Unable to find path", "src", src, "dst", dst, "code", reply.ErrorCode)
> 		return make(AppPathSet)
> ```

`return nil` is plenty.

---

*[go/lib/pathmgr/pred.go, line 30 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCOBbnoto4gKn_y0Pj:-KxCOBbnoto4gKn_y0Pk:bi19k98) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pred.go#L30)):*
> ```Go
> // A PathPredicate's Eval method returns true if the slice of interfaces in
> // Match is included in the AppPath parameter. Zero values in Match symbolize
> // wildcard matches.
> ```

If you remove the reference to pktcls/doc.go, then it would be best to add a clearer explanation here of what path predicate does (e.g. order matters, gaps are allowed).

---

*[go/lib/pathmgr/revtable.go, line 82 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCk3Eh_sGqbusLIOEM:-KxCk3Eh_sGqbusLIOEN:b-rxotwr) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/revtable.go#L82)):*
> ```Go
> }
> 
> // RevokeUIFID deletes all the paths that include uifid
> ```

Name update.

---

*[go/lib/sciond/sciond.go, line 60 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCQDKJG5fMCsMUj5XI:-KxCQDKJG5fMCsMUj5XJ:b-jnx7of) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/sciond/sciond.go#L60)):*
> ```Go
> 
> type service struct {
> 	name string
> ```

`name` is a very generic.. name. It looks like maybe this is a path/address?

---

*[go/lib/sciond/sciond.go, line 77 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCQO0dQTevlhounSZp:-KxCQO0dQTevlhounSZq:b-5wj9aa) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/sciond/sciond.go#L77)):*
> ```Go
> }
> 
> // A Connector is used to query SCIOND. The connector maintains an internal
> ```

Converting `go/lib/sciond` to use interfaces should be mentioned in the PR description (along with a brief mention of why).

---

*[go/lib/snet/snet.go, line 102 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-Kx8do3iIuqmwpEd-ON-:-Kx8do3iIuqmwpEd-ON0:b88iqef) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/snet/snet.go#L102)):*
> ```Go
> 		localIA:        ia,
> 	}
> 	sd := sciond.NewService(sPath)
> ```

Hmm. Should snet not keep a reference to this? E.g. if it needs to query sciond to find the address of a given interface.

---

*[go/lib/snet/snet.go, line 103 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-Kx8eN3C7kaNta_LeIWn:-Kx8eN3C7kaNta_LeIWo:b-2n83kn) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/snet/snet.go#L103)):*
> ```Go
> 	}
> 	sd := sciond.NewService(sPath)
> 	pathResolver, err := pathmgr.New(sd, 10*time.Second, time.Minute, log.Root())
> ```

Hmm. The `10*time.Second` value was a mistake i made, it wasn't intended to be included in a PR. These values both look _very_ low. I suggest 1min and 6h.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1290)*

scrye:
Review status: 16 of 19 files reviewed at latest revision, 29 unresolved discussions.

---

*[go/lib/pathmgr/apppath_test.go, line 15 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCRn1m5qG4KW75s-pR:-KxCuSB_C310wIYM6Kep:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/apppath_test.go#L15)):*

This file is now ~empty?
</blockquote></details>

Done.

---

*[go/lib/pathmgr/cache.go, line 29 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCW_pfYmreL5TnNmK7:-KxCub_Yerf84VQMxz2s:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L29)):*

It looks like this is specific to a given IA pair? If so it would be good to add a docstring mentioning that.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/cache.go, line 34 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCWJFep-Z8rvADxIeA:-KxCuc3Os1H6IkeWb8E3:b-ha7inz) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L34)):*

This implies that if `registered` is `false`, then `fs` must be nil - is that correct?
</blockquote></details>

It means `fs` is empty (allocated, but contains nothing).

---

*[go/lib/pathmgr/cache.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCWEl3oteS2mJYd3md:-KxCvL75pReHhDTjBwhT:b-u478au) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L35)):*

Given the change in terms, would `watched` be a better name?
</blockquote></details>

Refactored out.

---

*[go/lib/pathmgr/cache.go, line 44 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCWxEN5-b9qRHsSU17:-KxCvOLCxvjETUdgXksv:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L44)):*

"re-queried" ("required" isn't wrong, it just looks very much like a typo of "required" :)
</blockquote></details>

Done.

---

*[go/lib/pathmgr/cache.go, line 86 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCYxW4BwIv6Rj-TfsS:-KxCvVIDYD_XqzOUDOJE:bqe9kj) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L86)):*

This new pathFilter isn't populated.
</blockquote></details>

It just marks it as watched and initializes the `pathFilter`. Another function (`update`) populates it immediately afterwards (see `resolver.run()`).

---

*[go/lib/pathmgr/cache.go, line 113 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCZjMX-DTw6vvRcgt_:-KxCxCt_eYSflxHZuxmh:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L113)):*

There does not appear to be a way to remove a pathFilter.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/cache.go, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCYZ3Vt7nQIvX_qgJd:-KxCybtrQjJH8LJm2FMy:b-nmny1y) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L115)):*

I'm not clear why there is both `addFilteredSP` and `watch`, with very different semantics.  It looks like a caller has to call `watch` and then `addFilteredSP`. Why not just pass in a `*PathPredicate` into `watch`, and handle both at the same time?
</blockquote></details>

Simplified watches a bit, since now there's really no difference between a filtered and unfiltered one.

---

*[go/lib/pathmgr/cache.go, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCZ1e6Z0PhdSZq6Eh3:-KxD2ZQ7rm31o8GVTxWF:b-p90euq) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L128)):*

As discussed on a previous PR, using a filter string as a key is not a good idea (to recap: if you have an app that has two filters, both of which happen to have the same constraints currently, this code can't distinguish between them. If the app wants to update one of the filters, it would have to keep its own reference counts so as to know when it could remove the pathFilter, which is a lot of complexity. It's far better to use a filter identifier (e.g. name) that's independent of the actual filter contents).
</blockquote></details>

Named path predicates is something I dislike, because we move the problem from one place to another. Imagine the following scenario, you have a predicate `0-0#42`. Two goroutines want to add it to the path manager, they need to decide on different names. How would they do that? Importing from JSON is weird, because there'd be two predicates identical in all but name. So now instead of colliding on predicate string, they collide on name. We reach a point where we need to have the predicate itself, and a named container object containing a reference to it. But this is exactly what a `Class` is.

I'd just let apps handle this. They have all the information needed to make informed decisions here, and the library should not require them to go around hoops like naming just for its internals (i.e., the reference counter should be in app space).

Also, having multiple `SyncPaths` containing the exact same information under different names is odd design.

---

*[go/lib/pathmgr/cache.go, line 168 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxC_BQ6MSlZoXG3fvD4:-KxD5NWY45Au4MIEcsYm:b-r7doob) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L168)):*

s/revoke/remove/ ?
</blockquote></details>

Only `revoke()` calls this function so it's correct, but `remove` also works. Edited.

---

*[go/lib/pathmgr/defines.go, line 58 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCTI8N8_T8PL6gQKnl:-KxD5l-g8BSkX-KBG9us:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/defines.go#L58)):*

The wording is a bit difficult to follow. Suggestion:
```
update replaces the pathFilter's paths with those from aps, filtered by the pathFilter's PathPredicate (if set).
```
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCbSBKg2EVPml-MLBA:-KxD5tPW5OCCiesDn0lU:bcpffz7) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L115)):*

I think this can just be `return r.cache.getAPS(src, dst)`
</blockquote></details>

But that would return a `nil` `AppPathSet`, whereas I want an `AppPathSet` with 0 elements. All methods in this library behave the same way for sets (`AppPathSet` and `filterSet`) -- they return empty sets instead of `nil`. This allows methods with `map` reference receivers to operate without checking if an allocation is needed.

---

*[go/lib/pathmgr/pathmgr.go, line 165 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCcbNrAriGEyDuyfGB:-KxDBxpP8uau02RA6VnF:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L165)):*

I'm not sure that there is a need to have filtered vs non-filtered so separated. If the base `cache` calls accepted filters and DTRT depending on whether one was supplied or not, then this code becomes a lot simpler. `Watch()` becomes `return r.WatchFilter(src, dst, nil)`
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 201 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCeEiFm5-9TUo9unxW:-KxDC4sbFsOxeLp67buN:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L201)):*

Any code run in a go routine should have `defer liblog.LogPanicAndExit()` at the very top.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 204 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCdbyFpx7G61ReS4Kc:-KxDC6Pz53TE9CiB3su-:b7j7ms7) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L204)):*

Is/will `revoke` be called from anywhere else? If not, the lock/unlock could be moved into it, and this simply becomes `go r.revoke(revInfo)`.
</blockquote></details>

It will not, but I like the symmetry it has with the rest of the public functions. It makes it clear that the internal functions do not interact with locking, and as soon as you enter the object the master lock is acquired.

---

*[go/lib/pathmgr/pathmgr.go, line 243 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCe1hFDEASWRA2ixvp:-KxDCiP1eoHAY5A4EIB9:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L243)):*

This file is already long enough - i would move this to its own file.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 255 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCeUguvI-VNCoozp8e:-KxDCwLRDSN02sLUFHVl:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L255)):*

Any code run in a go routine should have `defer liblog.LogPanicAndExit()` at the very top.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 263 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCfwoe9N5yyx7UvReB:-KxDD5Shl-tbC4SstWEE:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L263)):*

This makes it impossible to unwatch, as the resolver will keep re-watching it.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 267 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCgB-cgedWr8IDoNZ7:-KxDDSHSw-GscISghfjg:bzgl61n) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L267)):*

Isn't it simpler to just:
```
if request.done != nil {
   close(request.done)
   request.done = nil
}
// reuse request
```
Note that Watch/Query would need to change so that they keep a direct reference to the chan.
</blockquote></details>

That's elegant, I like it.

---

*[go/lib/pathmgr/pathmgr.go, line 273 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxClGIj0VFkOJ53DbQ3:-KxDhdvVmh9eBtU9Gri0:b-w9u2bg) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L273)):*
> if some/all of the paths retrieved are going to expire before the next refireInterval (or even if they're getting close).

Would `sciond` ever serve paths that are this stale? Refire intervals will tend to be quite aggressive, so it means that SCIOND served paths that it knew will expire in less than a minute (or something small like that).

I've included three of the timers, but the path TTL one is trickier:
  - the path manager needs to completely parse all paths to extract the TTL of the path; I assume it's computed based on InfoField timestamp and HopField ExpTime? And the minimum out of all hopfields is considered the TTL?
  - queries need to check path TTLs to make sure stale paths aren't retrieved from the cache

I've added a fixme for the above, since I'd rather not include it as part of this PR.

---

*[go/lib/pathmgr/pathmgr.go, line 281 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCgU4se9XqqHZMM0qP:-KxDEAL0Tjjl86z6a1-B:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L281)):*

I'd log the entire request, as debugging this otherwise is going to hurt a lot.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 288 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCggWMGc1U9s0GAwU5:-KxDECOpAmY0kqellVPG:b-8irlro) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L288)):*

Hmm. What's the plan for supporting sciond path request flags? Maybe they should be in the `resolverRequest` struct.
</blockquote></details>

I suggest we add them when we need them, because we'll have a clearer idea of the use cases.

---

*[go/lib/pathmgr/pathmgr.go, line 296 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCgyhT4cEVWn692SBJ:-KxDEMO-yvo9gSgUSzcC:b-2hhxoo) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L296)):*

`return nil` is plenty.
</blockquote></details>

I prefer all sets returning empty set instead of nil in this library, they're simpler to handle.

---

*[go/lib/pathmgr/pred.go, line 30 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCOBbnoto4gKn_y0Pj:-KxDddFhXb0EMs1GPMSv:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pred.go#L30)):*

If you remove the reference to pktcls/doc.go, then it would be best to add a clearer explanation here of what path predicate does (e.g. order matters, gaps are allowed).
</blockquote></details>

Done.

---

*[go/lib/pathmgr/revtable.go, line 82 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCk3Eh_sGqbusLIOEM:-KxDEg_pfmll69cgn42q:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/revtable.go#L82)):*

Name update.
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 60 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCQDKJG5fMCsMUj5XI:-KxDcUTKL6fNW4gj2bt-:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/sciond/sciond.go#L60)):*

`name` is a very generic.. name. It looks like maybe this is a path/address?
</blockquote></details>

Done.

---

*[go/lib/sciond/sciond.go, line 77 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCQO0dQTevlhounSZp:-KxDk3SgykJcsCWUNKOS:bordo6d) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/sciond/sciond.go#L77)):*

Converting `go/lib/sciond` to use interfaces should be mentioned in the PR description (along with a brief mention of why).
</blockquote></details>

Will add this to the git commit message when it's merged.

---

*[go/lib/snet/snet.go, line 102 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-Kx8do3iIuqmwpEd-ON-:-KxDcXrrUQFMXjbAG7OJ:b3u38sw) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/snet/snet.go#L102)):*

Hmm. Should snet not keep a reference to this? E.g. if it needs to query sciond to find the address of a given interface.
</blockquote></details>

For the moment `snet` does not need it, although I can see future functionality requiring a connection to SCIOND. I recommend we add it when we need it.

---

*[go/lib/snet/snet.go, line 103 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-Kx8eN3C7kaNta_LeIWn:-KxDd4tYBj0YRhwR3SCG:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/snet/snet.go#L103)):*

Hmm. The `10*time.Second` value was a mistake i made, it wasn't intended to be included in a PR. These values both look _very_ low. I suggest 1min and 6h.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1290)*

kormat:
Reviewed 1 of 19 files at r1, 9 of 11 files at r3, 1 of 1 files at r4.
Review status: 19 of 21 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[go/lib/pathmgr/cache.go, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCZ1e6Z0PhdSZq6Eh3:-KxIK16JdctfA86wiIY0:b-mk167q) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L128)):*

Named path predicates is something I dislike, because we move the problem from one place to another. Imagine the following scenario, you have a predicate `0-0#42`. Two goroutines want to add it to the path manager, they need to decide on different names. How would they do that? Importing from JSON is weird, because there'd be two predicates identical in all but name. So now instead of colliding on predicate string, they collide on name. We reach a point where we need to have the predicate itself, and a named container object containing a reference to it. But this is exactly what a `Class` is.

I'd just let apps handle this. They have all the information needed to make informed decisions here, and the library should not require them to go around hoops like naming just for its internals (i.e., the reference counter should be in app space).

Also, having multiple `SyncPaths` containing the exact same information under different names is odd design.
</blockquote></details>

After Some™ discussion, rather than adding names/ids to predicates, pathFilter will contain a reference counter.

---

*[go/lib/pathmgr/cache.go, line 168 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxC_BQ6MSlZoXG3fvD4:-KxIPdvHtO8q9TLepeJL:br9taxa) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L168)):*

Only `revoke()` calls this function so it's correct, but `remove` also works. Edited.
</blockquote></details>

s/removek/remove/ ? :)

---

*[go/lib/pathmgr/cache.go, line 119 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxIOF3s21V3pUf6lVSu:-KxIOF3s21V3pUf6lVSv:brnx4sa) ([raw file](https://github.com/netsec-ethz/scion/blob/5e382b583bae16355186917d437483a56f2a4edd/go/lib/pathmgr/cache.go#L119)):*
> ```Go
> 
> // getWatch returns a pointer to a thread-safe object that contains paths
> // between src and dst, filtered according to filter. If filter is nil, a
> ```

Hmm. This is a bit misleading. If an unregistered filter is passed in, the response will be nil.

---

*[go/lib/pathmgr/pathmgr.go, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCbSBKg2EVPml-MLBA:-KxIQkbaT6l8huzB41u2:b-ip76f1) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L115)):*

But that would return a `nil` `AppPathSet`, whereas I want an `AppPathSet` with 0 elements. All methods in this library behave the same way for sets (`AppPathSet` and `filterSet`) -- they return empty sets instead of `nil`. This allows methods with `map` reference receivers to operate without checking if an allocation is needed.
</blockquote></details>

Hmm, ok. It feels a bit unidiomatic to create an empty map, but it's probably fine.

---

*[go/lib/pathmgr/pathmgr.go, line 273 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxClGIj0VFkOJ53DbQ3:-KxITTScrio5LXr_CWif:b-8iq8is) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L273)):*

> if some/all of the paths retrieved are going to expire before the next refireInterval (or even if they're getting close).

Would `sciond` ever serve paths that are this stale? Refire intervals will tend to be quite aggressive, so it means that SCIOND served paths that it knew will expire in less than a minute (or something small like that).

I've included three of the timers, but the path TTL one is trickier:
  - the path manager needs to completely parse all paths to extract the TTL of the path; I assume it's computed based on InfoField timestamp and HopField ExpTime? And the minimum out of all hopfields is considered the TTL?
  - queries need to check path TTLs to make sure stale paths aren't retrieved from the cache

I've added a fixme for the above, since I'd rather not include it as part of this PR.
</blockquote></details>

Avoiding getting paths that are close to expiry relies on assumptions about the caching that sciond and the various path servers involved do, which is not great. Also, for things like SIBRA steady paths, they're only valid for a max of ~3 minutes. So i'd rather not rely on avoiding this state.

Re: the path TTL, i think the right answer is to have sciond include it in the path response. So far we've avoided the pathmgr having to care about the raw path bytes, and it would be nice to keep that. Agreed re: not doing this in this PR.

---

*[go/lib/pathmgr/pathmgr.go, line 84 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxIQ-Js13Rz8md0AL9q:-KxIQ-Js13Rz8md0AL9r:b-raxwfc) ([raw file](https://github.com/netsec-ethz/scion/blob/5e382b583bae16355186917d437483a56f2a4edd/go/lib/pathmgr/pathmgr.go#L84)):*
> ```Go
> // refresh the path. New returns with an error if a connection to SCIOND could
> // not be established.
> func New(srvc sciond.Service, timers Timers, logger log.Logger) (*PR, error) {
> ```

One minor suggestion - it would be nice if there was a timers default. How about making this `timers *Timers`, and if its nil, using a default Timers instance?

---

*[go/lib/pathmgr/pathmgr.go, line 113 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxIR-KAROhtF1fYdnXF:-KxIR-KAROhtF1fYdnXG:b-z2zue6) ([raw file](https://github.com/netsec-ethz/scion/blob/5e382b583bae16355186917d437483a56f2a4edd/go/lib/pathmgr/pathmgr.go#L113)):*
> ```Go
> // found in the path resolver's cache, a query to SCIOND is issued and the
> // function blocks until the reply is received.
> func (r *PR) Query(src, dst *addr.ISD_AS) AppPathSet {
> ```

One thing that's not supported is a one-off filtered Query. Is that something you'd add later?

---

*[go/lib/pathmgr/pred.go, line 28 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxISGc437HzNHqr3ngF:-KxISGc437HzNHqr3ngG:bblmqdb) ([raw file](https://github.com/netsec-ethz/scion/blob/5e382b583bae16355186917d437483a56f2a4edd/go/lib/pathmgr/pred.go#L28)):*
> consecutive sequence

That sounds like no gaps are allowed, but IIRC we decided to allow gaps.

---

*[go/lib/pathmgr/pred.go, line 32 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxISaLk0IJyMGe0JcHl:-KxISaLk0IJyMGe0JcHm:b-qpxem0) ([raw file](https://github.com/netsec-ethz/scion/blob/5e382b583bae16355186917d437483a56f2a4edd/go/lib/pathmgr/pred.go#L32)):*
> ```Go
> // with 0. For example, a path filtering predicate that only allows paths which
> // pass through ISD1 can be created with:
> //     pp, err = NewActionFilterPaths("PathsThroughISD1", "1-0#0")
> ```

This is specific to `pktcls`.  `pp, err := NewPathPredicate("1-0#0")` would be the appropriate example here. The marshalling comments below is also pktcls-specific, so it can just be dropped.

---

*[go/lib/pktcls/doc.go, line 33 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxIVaBgpE-rZXleB93y:-KxIVaBgpE-rZXleB93z:b-szqell) ([raw file](https://github.com/netsec-ethz/scion/blob/b16c6ca1ed678ddb7e51d7972b9ed08a37c4d5ae/go/lib/pktcls/doc.go#L33)):*
> ```Go
> // Actions are marshalable objects that describe some process. Currently, the
> // only supported action is Path Filtering (ActionFilterPaths). A path filter
> // is just a container for a path predicate.  The predicate specifies which
> ```

I think most of this is pkgcls-specific, and should be kept here. Maybe: "A path filter is just a container for a pathmgr.PathPredicate", and then continue with the marshalling description.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1290)*

scrye:
Review status: 13 of 21 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/pathmgr/cache.go, line 128 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxCZ1e6Z0PhdSZq6Eh3:-KxNMoo2AzKPIqUccQVk:bv23k6k) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L128)):*

After Some™ discussion, rather than adding names/ids to predicates, pathFilter will contain a reference counter.
</blockquote></details>

Done, it needed very few changes.

---

*[go/lib/pathmgr/cache.go, line 168 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxC_BQ6MSlZoXG3fvD4:-KxNP9qFEdD3vTDf-bsR:bs7ww3u) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/cache.go#L168)):*

s/removek/remove/ ? :)
</blockquote></details>

Donek.

---

*[go/lib/pathmgr/cache.go, line 119 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxIOF3s21V3pUf6lVSu:-KxNKvf9u_r40pS2W4Hp:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5e382b583bae16355186917d437483a56f2a4edd/go/lib/pathmgr/cache.go#L119)):*

Hmm. This is a bit misleading. If an unregistered filter is passed in, the response will be nil.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr.go, line 273 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxClGIj0VFkOJ53DbQ3:-KxNZQNXpEOfZkuFppbR:b-h4nmtu) ([raw file](https://github.com/netsec-ethz/scion/blob/d9db9b42b9f06f408404d54fb3aa55583840ef67/go/lib/pathmgr/pathmgr.go#L273)):*

Avoiding getting paths that are close to expiry relies on assumptions about the caching that sciond and the various path servers involved do, which is not great. Also, for things like SIBRA steady paths, they're only valid for a max of ~3 minutes. So i'd rather not rely on avoiding this state.

Re: the path TTL, i think the right answer is to have sciond include it in the path response. So far we've avoided the pathmgr having to care about the raw path bytes, and it would be nice to keep that. Agreed re: not doing this in this PR.
</blockquote></details>

Yes, I agree re: SCIOND sending the path TTL.

---

*[go/lib/pathmgr/pathmgr.go, line 84 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxIQ-Js13Rz8md0AL9q:-KxNPf4Lal1wFxFOVY6R:bhkxm6r) ([raw file](https://github.com/netsec-ethz/scion/blob/5e382b583bae16355186917d437483a56f2a4edd/go/lib/pathmgr/pathmgr.go#L84)):*

One minor suggestion - it would be nice if there was a timers default. How about making this `timers *Timers`, and if its nil, using a default Timers instance?
</blockquote></details>

To allow for more granular defaults, I've changed the code to initialize 0 timers with the default. Also added comments specifying this behavior.

---

*[go/lib/pathmgr/pathmgr.go, line 113 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxIR-KAROhtF1fYdnXF:-KxNRKqSFBBqdyP8uZow:b-286pf) ([raw file](https://github.com/netsec-ethz/scion/blob/5e382b583bae16355186917d437483a56f2a4edd/go/lib/pathmgr/pathmgr.go#L113)):*

One thing that's not supported is a one-off filtered Query. Is that something you'd add later?
</blockquote></details>

It's trivial to implement (and apps can do it themselves), but it's a nice helper function. Added.

---

*[go/lib/pathmgr/pred.go, line 28 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxISGc437HzNHqr3ngF:-KxN_5fQqxV3C_ZrSzKo:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5e382b583bae16355186917d437483a56f2a4edd/go/lib/pathmgr/pred.go#L28)):*

> consecutive sequence

That sounds like no gaps are allowed, but IIRC we decided to allow gaps.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pred.go, line 32 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxISaLk0IJyMGe0JcHl:-KxN_JxKbHzgiZYWJJWQ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5e382b583bae16355186917d437483a56f2a4edd/go/lib/pathmgr/pred.go#L32)):*

This is specific to `pktcls`.  `pp, err := NewPathPredicate("1-0#0")` would be the appropriate example here. The marshalling comments below is also pktcls-specific, so it can just be dropped.
</blockquote></details>

Done.

---

*[go/lib/pktcls/doc.go, line 33 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxIVaBgpE-rZXleB93y:-KxN_qOc0jCHXm-oSPPs:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/b16c6ca1ed678ddb7e51d7972b9ed08a37c4d5ae/go/lib/pktcls/doc.go#L33)):*

I think most of this is pkgcls-specific, and should be kept here. Maybe: "A path filter is just a container for a pathmgr.PathPredicate", and then continue with the marshalling description.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1290)*

kormat:
Reviewed 6 of 7 files at r5.
Review status: 19 of 21 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1290)*

kormat:
Reviewed 1 of 7 files at r5.
Review status: 20 of 21 files reviewed at latest revision, 7 unresolved discussions.

---

*[go/lib/pathmgr/pathmgr_test.go, line 140 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS4JutSelALHx3MAZj:-KxS4JutSelALHx3MAZk:bftyao1) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L140)):*
> ```Go
> }
> 
> func (m *mockSCIONDConn) Paths(dst, src *addr.ISD_AS, max uint16, f sciond.PathReqFlags) (*sciond.PathReply, error) {
> ```

Line length.

---

*[go/lib/pathmgr/pathmgr_test.go, line 209 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS8kcWaUBdbADjTqEr:-KxS8gSesYidF7MjY3IR:bftyao1) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L209)):*
> ```Go
> 					SoMsg("aps len", len(aps), ShouldEqual, 1)
> 				})
> 				Convey("Wait 2 seconds for paths to expire, then query again. New paths are received", func() {
> ```

Line length.

---

*[go/lib/pathmgr/pathmgr_test.go, line 290 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS5VHJuwrycyvLFLym:-KxS5VHJuwrycyvLFLyn:b-ihuj64) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L290)):*
> ```Go
> 		SoMsg("pm", pm, ShouldNotBeNil)
> 		SoMsg("err", err, ShouldBeNil)
> 		Convey("Register filter IFID42", func() {
> ```

Outdated ifid?

---

*[go/lib/pathmgr/pathmgr_test.go, line 335 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS5zz_3BLCFVTRoYqg:-KxS5zz_3BLCFVTRoYqh:b-mcap6c) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L335)):*
> ```Go
> 			spd := sp.Load()
> 			SoMsg("aps register", spd.APS, ShouldResemble, buildAPS(pathXZ, pathXYZ))
> 			// Needs 0 queries to SCIOND (SyncPaths created for previous query), and the filter keeps only pathZ
> ```

Line length.

`pathZ` - should be `pathXYZ`?

---

*[go/lib/pathmgr/pathmgr_test.go, line 350 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS9wGSdlf7ZZjURP48:-KxS9wGSdlf7ZZjURP49:b-l3rbgd) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L350)):*
> ```Go
> 			})
> 			Convey("Revoke a path that's in Query, but not in Watch/WatchFilter path sets", func() {
> 				// This removes an UIFID from pathY
> ```

Path name needs update.

---

*[go/lib/pathmgr/pathmgr_test.go, line 357 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxSAnCpSixS_fmUfuJZ:-KxSAnCpSixS_fmUfuJ_:bur7geh) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L357)):*
> ```Go
> 			Convey("Revoke a path that's in Watch and WatchFilter, but not in Query", func() {
> 				pm.cache.revoke(uifidFromValues(iaY, 231))
> 				sp, err := pm.Watch(iaX, iaZ)
> ```

Why recreate `sp` and `spf`? They already exist and have identical configurations from above, AFAICT.

---

*[go/lib/pathmgr/pathmgr_test.go, line 360 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS62aR6HSMRlAGOheG:-KxS62aR6HSMRlAGOheH:bftyao1) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L360)):*
> ```Go
> 				SoMsg("err register", err, ShouldBeNil)
> 				spd := sp.Load()
> 				// XYZ and XZ were in the path set, but now that XYZ's been filtered out only XZ remains
> ```

Line length.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1290)*

kormat:
Reviewed 1 of 19 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1290)*

scrye:
Review status: 20 of 21 files reviewed at latest revision, 7 unresolved discussions.

---

*[go/lib/pathmgr/pathmgr_test.go, line 140 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS4JutSelALHx3MAZj:-KxTABD3ap606yVG2FGh:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L140)):*

Line length.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 209 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS8kcWaUBdbADjTqEr:-KxTACYmXAZS3Pno1PBg:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L209)):*

Line length.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 290 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS5VHJuwrycyvLFLym:-KxTAFtd_399fnLJAmAJ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L290)):*

Outdated ifid?
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 335 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS5zz_3BLCFVTRoYqg:-KxTAMcEuQaPA9DfxRRZ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L335)):*

Line length.

`pathZ` - should be `pathXYZ`?
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 350 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS9wGSdlf7ZZjURP48:-KxTATT_hKpLsjSc_VpU:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L350)):*

Path name needs update.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 357 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxSAnCpSixS_fmUfuJZ:-KxTAbpbbeYEXq-XxmsU:bx7n04q) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L357)):*

Why recreate `sp` and `spf`? They already exist and have identical configurations from above, AFAICT.
</blockquote></details>

Good point, Copy paste fail. Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 360 at r5](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-KxS62aR6HSMRlAGOheG:-KxTAqZ608FcVJ7DTUyE:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/d5f1954a9f77e55c9f0bd7277c09e27813cd5e2c/go/lib/pathmgr/pathmgr_test.go#L360)):*

Line length.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1290)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1290#-:-KxTBte56ySTey_o6OEw:bnfp4nl)*

kormat:
Reviewed 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1290)*

PR #1303, opened 2017-10-30
Revert Certificate signature input

Oncilla:
This PR reverts the changes to the signature input of certificates.

Commit ca4bce6 removed base64 encoding of strings during signature input creation of a certificate.
This breaks all certificates in the current deployment.

Such breaking changes will be aggregated and applied in a coordinated manner, in order to 
minimize maintenance  efforts.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1303)


kormat:
From discussions on slack and elsewhere, if commits to master break existing setups, the solution is to have those setups (e.g. scion-lab) pin to a commit/branch other than `HEAD`/`master`.

PR #1312, opened 2017-11-06
Fix a logging race-condition.

kormat:
The existing liblog uses a bufio.Writer to buffer log output, with a
goroutine calling `.Flush()` on it every 5s. However, bufio.Writer is
_not_ thread-safe, so if flush was ever called while a logging statement
was called, the results are undefined.

This issue was discovered while investigating why our go processes
sometimes stop logging permenantly (e.g. a week without logging
anything), and need to be restarted. It's not clear if this is related
to the race-condition, only time will tell.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1312)

### DIFF
--- a/go/lib/log/log.go
+++ b/go/lib/log/log.go
@@ -15,7 +15,6 @@
 package liblog
 
 import (
-	"bufio"
 	"flag"
 	"fmt"
 	"io"
@@ -35,8 +34,9 @@ var logLevel = flag.String("log.level", "debug", "Logging level")
 var logConsole = flag.String("log.console", "crit", "Console logging level")
 var logSize = flag.Int("log.size", 50, "Max size of log file in MiB")
 var logAge = flag.Int("log.age", 7, "Max age of log file in days")
+var logFlush = flag.Int("log.flush", 5, "How frequently to flush to the log file, in seconds")
 
-var logBuf *bufio.Writer
+var logBuf *syncBuf
 
 func init() {
 	os.Setenv("TZ", "UTC")
@@ -45,7 +45,7 @@ func init() {
 
 func Setup(name string) {
 	logLvl, consLvl := parseLvls()
-	logBuf = bufio.NewWriter(mkLogfile(name))
+	logBuf = newSyncBuf(mkLogfile(name))
 	handler := log.MultiHandler(
 		log.LvlFilterHandler(logLvl, log.StreamHandler(logBuf, fmt15.Fmt15Format(nil))),
 		log.LvlFilterHandler(consLvl, log.StreamHandler(os.Stdout,
@@ -53,7 +53,7 @@ func Setup(name string) {
 	)
 	log.Root().SetHandler(handler)
 	go func() {
-		for range time.Tick(5 * time.Second) {
+		for range time.Tick(time.Duration(*logFlush) * time.Second) {
 			Flush()
 		}
 	}()
@@ -73,11 +73,11 @@ func parseLvls() (log.Lvl, log.Lvl) {
 	return logLvl, consLvl
 }
 
-func mkLogfile(name string) io.Writer {
+func mkLogfile(name string) io.WriteCloser {
 	return &lumberjack.Logger{
 		Filename: fmt.Sprintf("%s/%s.log", *logDir, name),
-		MaxSize:  50, // MiB
-		MaxAge:   7,  // days
+		MaxSize:  *logSize, // MiB
+		MaxAge:   *logAge,  // days
 	}
 }
 
--- a/go/lib/log/syncbuf.go
+++ b/go/lib/log/syncbuf.go
@@ -1,0 +1,52 @@
+// Copyright 2017 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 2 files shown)


shitz:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/log/log.go, line 56 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1312#-KyGyapcUK5TUMlYe62I:-KyGyapdI_B2SgyR1yf4:b3nssun) ([raw file](https://github.com/netsec-ethz/scion/blob/1899155cffff8cb313fa80e8509bd304c60bb8f9/go/lib/log/log.go#L56)):*
> ```Go
> 	log.Root().SetHandler(handler)
> 	go func() {
> 		for range time.Tick(time.Duration(*logFlush) * time.Second) {
> ```

Is `time.Duration` really necessary here?

---

*[go/lib/log/syncbuf.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1312#-KyGxvI4cTF-tlNYE7dy:-KyGxvI4cTF-tlNYE7dz:boizez9) ([raw file](https://github.com/netsec-ethz/scion/blob/1899155cffff8cb313fa80e8509bd304c60bb8f9/go/lib/log/syncbuf.go#L35)):*
> ```Go
> }
> 
> func (s *syncBuf) Write(b []byte) (int, error) {
> ```

This should check whether the buffer is full before writing. Suggestion:
```
s.Lock()
defer s.Unlock()
if len(b) > s.buf.Available() {
    s.buf.Flush()
}
return s.buf.Write()
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1312)*

kormat:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/log/log.go, line 56 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1312#-KyGyapcUK5TUMlYe62I:-KyH0C_FxygR_fVPga0a:bp1p6sk) ([raw file](https://github.com/netsec-ethz/scion/blob/1899155cffff8cb313fa80e8509bd304c60bb8f9/go/lib/log/log.go#L56)):*

Is `time.Duration` really necessary here?
</blockquote></details>

Yes. Otherwise it's a type error.

---

*[go/lib/log/syncbuf.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1312#-KyGxvI4cTF-tlNYE7dy:-KyH0EcPjWbomlSOml-J:b-pzzwko) ([raw file](https://github.com/netsec-ethz/scion/blob/1899155cffff8cb313fa80e8509bd304c60bb8f9/go/lib/log/syncbuf.go#L35)):*

This should check whether the buffer is full before writing. Suggestion:
```
s.Lock()
defer s.Unlock()
if len(b) > s.buf.Available() {
    s.buf.Flush()
}
return s.buf.Write()
```
</blockquote></details>

`bufio.Writer` handles this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1312)*

shitz:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/log/log.go, line 56 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1312#-KyGyapcUK5TUMlYe62I:-KyH0dIRMVhT61QjHNPv:b-mxwwm2) ([raw file](https://github.com/netsec-ethz/scion/blob/1899155cffff8cb313fa80e8509bd304c60bb8f9/go/lib/log/log.go#L56)):*

Yes. Otherwise it's a type error.
</blockquote></details>

But `int(*logFlush)` would make more sense. Duration times Duration doesn't make sense.

---

*[go/lib/log/syncbuf.go, line 35 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1312#-KyGxvI4cTF-tlNYE7dy:-KyH1fO7GPpQswJT8vqH:bw3abv) ([raw file](https://github.com/netsec-ethz/scion/blob/1899155cffff8cb313fa80e8509bd304c60bb8f9/go/lib/log/syncbuf.go#L35)):*

`bufio.Writer` handles this.
</blockquote></details>

Indeed it does. Wasn't clear from the docs, but looking at the implementation confirms that it's handled by `bufio.Writer`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1312)*

shitz:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[go/lib/log/log.go, line 56 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1312#-KyGyapcUK5TUMlYe62I:-KyH37ryzsxO0r_llykD:b-hj1fkq) ([raw file](https://github.com/netsec-ethz/scion/blob/f15b1d573eeaaec233e8ce3787fbcf5dcffee9f1/go/lib/log/log.go#L56)):*

But `int(*logFlush)` would make more sense. Duration times Duration doesn't make sense.
</blockquote></details>

Nevermind, it seems this really doesn't make sense on a logical level but is required by the typesystem.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1312#-:-KyH3EecPB_gFcI88YIn:bnfp4nl)*

PR #1315, opened 2017-11-08
Fix SIG seq nr wrapping.

kormat:
As the SIG sequence number is now 24bit, we can't rely on uint32
handling the wrapping for us. This caused the receiving SIG to consider
all frames in an epoch after the first 16mil as "too old", and drop them.

Also:
- Fix a bug that would prevent sigmgr from shutting down.
- Add some extra log statements to mark the start/stop of major go
  routines.
- Don't log timeouts if there is no path.
- Simplify ingress worker start/stop/cleanup.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1315)

### DIFF
--- a/go/sig/base/as.go
+++ b/go/sig/base/as.go
@@ -176,11 +176,13 @@ func (ae *ASEntry) sigMgr() {
 	defer liblog.LogPanicAndExit()
 	ticker := time.NewTicker(sigMgrTick)
 	defer ticker.Stop()
+	log.Info("sigMgr starting")
+Top:
 	for {
 		// TODO(kormat): handle adding new SIGs from discovery, and updating existing ones.
 		select {
 		case <-ae.sigMgrStop:
-			break
+			break Top
 		case <-ticker.C:
 			smap := ae.SigMap()
 			for _, sig := range smap {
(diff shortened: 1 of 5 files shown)


scrye:
:lgtm:

---

Reviewed 5 of 5 files at r1.

---

*[go/sig/ingress/dispatcher.go, line 129 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1315#-KyQd2OT2O3exHXVVQEe:-KyQbbe9Z4yaKv5OBum3:b-god8k2) ([raw file](https://github.com/netsec-ethz/scion/blob/396d3253f6d81e5b97c3eac22ffbb18ad67e1960/go/sig/ingress/dispatcher.go#L129)):*
> ```Go
> 		if worker.markedForCleanup {
> 			delete(d.workers, key)
> 			go worker.Stop()
> ```

It is theoretically possible for `Stop()` to run before `Run()`, but I did not find any issues with that. Looking closer, the old code included a race here, but it's gone now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1315#-:-KyQiJhrua9nXdiBpOvN:bnfp4nl)*

PR #1318, opened 2017-11-10
Fix buffer leaks in egress dispatcher

scrye:
Deleting an AS Entry from the SIG would cause the associated egress dispatcher to continuously raise errors, polluting logs and depleting the free buffer pool for all dispatchers, thus causing the entire SIG to stop forwarding. This fix:
  - releases buffers on errors now, so buffers are never leaked
  - causes the egress dispatcher to cleanly shut down when the tunnel interface is closed

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1318)


kormat:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1318#-:-Ky_owfvdtOZ_Y8vtvob:bnfp4nl)*

PR #1322, opened 2017-11-10
SIG can reload configs

scrye:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1322)

### DIFF
--- a/go/lib/pathmgr/pathmgr.go
+++ b/go/lib/pathmgr/pathmgr.go
@@ -218,8 +218,9 @@ func (r *PR) WatchFilter(src, dst *addr.ISD_AS, filter *PathPredicate) (*SyncPat
 
 // UnwatchFilter deletes a previously registered filter.
 func (r *PR) UnwatchFilter(src, dst *addr.ISD_AS, filter *PathPredicate) error {
-	// FIXME(scrye): Implement this
-	return common.NewCError("Function UnwatchFilter not implemented")
+	r.Lock()
+	defer r.Unlock()
+	return r.cache.removeWatch(src, dst, filter)
 }
 
 // Revoke asynchronously informs SCIOND about a revocation and flushes any
--- a/go/sig/base/as.go
+++ b/go/sig/base/as.go
@@ -27,6 +27,7 @@ import (
 	"github.com/netsec-ethz/scion/go/lib/addr"
 	"github.com/netsec-ethz/scion/go/lib/common"
 	liblog "github.com/netsec-ethz/scion/go/lib/log"
+	"github.com/netsec-ethz/scion/go/sig/config"
 	"github.com/netsec-ethz/scion/go/sig/egress"
 	"github.com/netsec-ethz/scion/go/sig/sigcmn"
 	"github.com/netsec-ethz/scion/go/sig/siginfo"
@@ -67,7 +68,17 @@ func newASEntry(ia *addr.ISD_AS) (*ASEntry, error) {
 	return ae, nil
 }
 
-func (ae *ASEntry) setupNet() error {
+func (ae *ASEntry) ReloadConfig(cfg *config.ASEntry) error {
+	ae.Lock()
+	defer ae.Unlock()
+	ae.addNewSIGS(cfg.Sigs)
+	ae.delOldSIGS(cfg.Sigs)
+	ae.addNewNets(cfg.Nets)
+	ae.delOldNets(cfg.Nets)
+	return nil
+}
+
+func (ae *ASEntry) createTunDevice() error {
 	var err error
 	ae.tunLink, ae.tunIO, err = xnet.ConnectTun(ae.DevName)
 	if err != nil {
@@ -80,13 +91,46 @@ func (ae *ASEntry) setupNet() error {
 	return nil
 }
 
+func (ae *ASEntry) addNewNets(ipnets []*config.IPNet) error {
+	for _, ipnet := range ipnets {
+		err := ae.addNet(ipnet.IPNet())
+		if err != nil {
+			log.Error("Unable to add network", "net", ipnet, "err", err)
+		}
+	}
+	return nil
+}
+
+func (ae *ASEntry) delOldNets(ipnets []*config.IPNet) error {
+	for _, ne := range ae.Nets {
+		found := false
+		for _, ipnet := range ipnets {
+			if ne.Net.String() == ipnet.IPNet().String() {
+				found = true
+				break
+			}
+		}
+		if !found {
+			err := ae.delNet(ne.Net)
+			if err != nil {
+				log.Error("Unable to delete network", "NetEntry", ne, "err", err)
+			}
+		}
+	}
+	return nil
+}
+
 // AddNet idempotently adds a network for the remote IA.
 func (ae *ASEntry) AddNet(ipnet *net.IPNet) error {
 	ae.Lock()
 	defer ae.Unlock()
+	return ae.addNet(ipnet)
+}
+
+func (ae *ASEntry) addNet(ipnet *net.IPNet) error {
 	if ae.tunLink == nil {
 		// Ensure that the network setup is done, as otherwise route entries can't be added.
-		if err := ae.setupNet(); err != nil {
+		if err := ae.createTunDevice(); err != nil {
 			return err
 		}
 	}
@@ -106,6 +150,12 @@ func (ae *ASEntry) AddNet(ipnet *net.IPNet) error {
 // DelIA removes a network for the remote IA.
 func (ae *ASEntry) DelNet(ipnet *net.IPNet) error {
 	ae.Lock()
+	defer ae.Unlock()
+	return ae.delNet(ipnet)
+}
+
+// DelIA removes a network for the remote IA.
+func (ae *ASEntry) delNet(ipnet *net.IPNet) error {
 	key := ipnet.String()
 	ne, ok := ae.Nets[key]
 	if !ok {
@@ -113,16 +163,48 @@ func (ae *ASEntry) DelNet(ipnet *net.IPNet) error {
 		return common.NewCError("DelNet: no network found", "ia", ae.IA, "net", ipnet)
 	}
 	delete(ae.Nets, key)
-	ae.Unlock() // Do cleanup outside the lock.
 	ae.Info("Removed network", "net", ipnet)
 	return ne.Cleanup()
 }
 
+func (ae *ASEntry) addNewSIGS(sigs config.SIGSet) error {
+	for _, sig := range sigs {
+		ctrlPort := int(sig.CtrlPort)
+		if ctrlPort == 0 {
+			ctrlPort = sigcmn.DefaultCtrlPort
+		}
+		encapPort := int(sig.EncapPort)
+		if encapPort == 0 {
+			encapPort = sigcmn.DefaultEncapPort
+		}
+		err := ae.addSig(sig.Id, sig.Addr, ctrlPort, encapPort, true)
+		if err != nil {
+			log.Error("Unable to add SIG", "sig", sig, "err", err)
+		}
+	}
+	return nil
+}
+
+func (ae *ASEntry) delOldSIGS(sigs config.SIGSet) error {
+	for _, sig := range ae.Sigs {
+		if _, ok := sigs[sig.Id]; !ok {
+			err := ae.delSig(sig.Id)
+			if err != nil {
+				log.Error("Unable to delete SIG", "err", err)
+			}
+		}
+	}
+	return nil
+}
+
 // AddSig idempotently adds a SIG for the remote IA.
-func (ae *ASEntry) AddSig(id siginfo.SigIdType, ip net.IP,
-	ctrlPort, encapPort int, static bool) error {
+func (ae *ASEntry) AddSig(id siginfo.SigIdType, ip net.IP, ctrlPort, encapPort int, static bool) error {
 	ae.Lock()
 	defer ae.Unlock()
+	return ae.addSig(id, ip, ctrlPort, encapPort, static)
+}
+
+func (ae *ASEntry) addSig(id siginfo.SigIdType, ip net.IP, ctrlPort, encapPort int, static bool) error {
 	if len(id) == 0 {
 		return common.NewCError("AddSig: SIG id empty", "ia", ae.IA)
 	}
@@ -137,11 +219,11 @@ func (ae *ASEntry) AddSig(id siginfo.SigIdType, ip net.IP,
 		cerr := err.(*common.CError)
 		return cerr.AddCtx(cerr.Ctx, "ia", ae.IA, "id", id)
 	}
-	if _, ok := ae.Sigs[id]; ok {
-		// FIXME(kormat): support updating SIG entry.
-		return nil
-	}
-	ae.Sigs[id] = siginfo.NewSig(ae.IA, id, addr.HostFromIP(ip), ctrlPort, encapPort, static)
+	// The fastpath grabs a copy of the new map after we release the mutex, so
+	// we can edit the new entry (including the reference in the map) without
+	// racing.
+	sig := siginfo.NewSig(ae.IA, id, addr.HostFromIP(ip), ctrlPort, encapPort, static)
+	ae.Sigs[id] = sig
 	ae.Info("Added SIG", "sig", ae.Sigs[id])
 	return nil
 }
@@ -149,13 +231,16 @@ func (ae *ASEntry) AddSig(id siginfo.SigIdType, ip net.IP,
 // DelSIG removes an SIG for the remote IA.
 func (ae *ASEntry) DelSig(id siginfo.SigIdType) error {
 	ae.Lock()
+	defer ae.Unlock()
+	return ae.delSig(id)
+}
+
+func (ae *ASEntry) delSig(id siginfo.SigIdType) error {
 	se, ok := ae.Sigs[id]
 	if !ok {
-		ae.Unlock()
 		return common.NewCError("DelSig: no SIG found", "ia", ae.IA, "id", id)
 	}
 	delete(ae.Sigs, id)
-	ae.Unlock() // Do cleanup outside the lock.
 	ae.Info("Removed SIG", "id", id)
 	return se.Cleanup()
 }
@@ -204,13 +289,8 @@ func (ae *ASEntry) Cleanup() error {
 	}
 	// Clean up sessions, and associated workers.
 	ae.cleanSessions()
-	for _, ne := range ae.Nets {
-		if err := ne.Cleanup(); err != nil {
-			cerr := err.(*common.CError)
-			ae.Error(cerr.Desc, cerr.Ctx...)
-		}
-	}
 	if err := netlink.LinkDel(ae.tunLink); err != nil {
+		// The operating system also removes the tun device and routes
 		// Only return this error, as it's the only critical one.
 		return common.NewCError("Error removing TUN link",
 			"ia", ae.IA, "dev", ae.DevName, "err", err)
--- a/go/sig/base/as.go
+++ b/go/sig/base/as.go
@@ -62,7 +62,7 @@ func newASEntry(ia *addr.ISD_AS) (*ASEntry, error) {
 		sigMgrStop: make(chan struct{}),
 	}
 	var err error
-	if ae.Session, err = egress.NewSession(ia, 0, ae.SigMap, ae.Logger); err != nil {
+	if ae.Session, err = egress.NewSession(ia, 0, ae.Sigs, ae.Logger); err != nil {
 		return nil, err
 	}
 	return ae, nil
@@ -238,11 +238,6 @@ func (ae *ASEntry) DelSig(id siginfo.SigIdType) error {
 	ae.Sigs.Delete(id)
 	ae.Info("Removed SIG", "id", id)
 	return se.Cleanup()
-}
-
-// Internal method to return a *copy* of the ASEntry's SigMap
-func (ae *ASEntry) SigMap() *siginfo.SigMap {
-	return ae.Sigs
 }
 
 // manage the Sig map
--- a/go/sig/base/as.go
+++ b/go/sig/base/as.go
@@ -38,17 +38,23 @@ const sigMgrTick = 10 * time.Second
 
 // ASEntry contains all of the information required to interact with a remote AS.
 type ASEntry struct {
+	// Global lock (used only for exclusive access during cleanup)
 	sync.RWMutex
-	log.Logger
+
+	netLock sync.RWMutex
+	Nets    map[string]*NetEntry
+
+	sigLock sync.RWMutex
+	Sigs    siginfo.SigMap
+
 	IA         *addr.ISD_AS
 	IAString   string
-	Nets       map[string]*NetEntry
-	Sigs       siginfo.SigMap
 	Session    *egress.Session
 	DevName    string
 	tunLink    netlink.Link
 	tunIO      io.ReadWriteCloser
 	sigMgrStop chan struct{}
+	log.Logger
 }
 
 func newASEntry(ia *addr.ISD_AS) (*ASEntry, error) {
@@ -69,8 +75,6 @@ func newASEntry(ia *addr.ISD_AS) (*ASEntry, error) {
 }
 
 func (ae *ASEntry) ReloadConfig(cfg *config.ASEntry) bool {
-	ae.Lock()
-	defer ae.Unlock()
 	// Method calls first to prevent skips due to logical short-circuit
 	s := ae.addNewSIGS(cfg.Sigs)
 	s = ae.delOldSIGS(cfg.Sigs) && s
@@ -80,11 +84,13 @@ func (ae *ASEntry) ReloadConfig(cfg *config.ASEntry) bool {
 
 // addNewNets adds the networks in ipnets that are not currently configured.
 func (ae *ASEntry) addNewNets(ipnets []*config.IPNet) bool {
+	ae.netLock.Lock()
+	defer ae.netLock.Unlock()
 	s := true
 	for _, ipnet := range ipnets {
 		err := ae.addNet(ipnet.IPNet())
 		if err != nil {
-			log.Error("Unable to add network", "net", ipnet, "err", err)
+			ae.Error("Unable to add network", "net", ipnet, "err", err)
 			s = false
 		}
 	}
@@ -93,21 +99,20 @@ func (ae *ASEntry) addNewNets(ipnets []*config.IPNet) bool {
 
 // delOldNets deletes currently configured networks that are not in ipnets.
 func (ae *ASEntry) delOldNets(ipnets []*config.IPNet) bool {
+	ae.netLock.Lock()
+	defer ae.netLock.Unlock()
 	s := true
+Top:
 	for _, ne := range ae.Nets {
-		found := false
 		for _, ipnet := range ipnets {
 			if ne.Net.String() == ipnet.IPNet().String() {
-				found = true
-				break
+				continue Top
 			}
 		}
-		if !found {
-			err := ae.delNet(ne.Net)
-			if err != nil {
-				log.Error("Unable to delete network", "NetEntry", ne, "err", err)
-				s = false
-			}
+		err := ae.delNet(ne.Net)
+		if err != nil {
+			ae.Error("Unable to delete network", "NetEntry", ne, "err", err)
+			s = false
 		}
 	}
 	return s
@@ -115,15 +120,15 @@ func (ae *ASEntry) delOldNets(ipnets []*config.IPNet) bool {
 
 // AddNet idempotently adds a network for the remote IA.
 func (ae *ASEntry) AddNet(ipnet *net.IPNet) error {
-	ae.Lock()
-	defer ae.Unlock()
+	ae.netLock.Lock()
+	defer ae.netLock.Unlock()
 	return ae.addNet(ipnet)
 }
 
 func (ae *ASEntry) addNet(ipnet *net.IPNet) error {
 	if ae.tunLink == nil {
 		// Ensure that the network setup is done, as otherwise route entries can't be added.
-		if err := ae.createTunDevice(); err != nil {
+		if err := ae.setupNet(); err != nil {
 			return err
 		}
 	}
@@ -142,8 +147,8 @@ func (ae *ASEntry) addNet(ipnet *net.IPNet) error {
 
 // DelIA removes a network for the remote IA.
 func (ae *ASEntry) DelNet(ipnet *net.IPNet) error {
-	ae.Lock()
-	defer ae.Unlock()
+	ae.netLock.Lock()
+	defer ae.netLock.Unlock()
 	return ae.delNet(ipnet)
 }
 
@@ -162,6 +167,8 @@ func (ae *ASEntry) delNet(ipnet *net.IPNet) error {
 
 // addNewSIGS adds the SIGs in sigs that are not currently configured.
 func (ae *ASEntry) addNewSIGS(sigs config.SIGSet) bool {
+	ae.sigLock.Lock()
+	defer ae.sigLock.Unlock()
 	s := true
 	for _, sig := range sigs {
 		ctrlPort := int(sig.CtrlPort)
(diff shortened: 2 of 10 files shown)


kormat:
Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[go/sig/base/as.go, line 72 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuW80ijEKWTuXWKdjU:-KyuW80ijEKWTuXWKdjV:b-vt4vh7) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L72)):*
> ```Go
> 
> func (ae *ASEntry) ReloadConfig(cfg *config.ASEntry) bool {
> 	ae.Lock()
> ```

I'm concerned by how coarse the locking is here. E.g. the session monitor calls SigMap every 500ms, and in the future will want to call `AddSIG`/`DelSIG` etc. If reconfiguring is slow, that can cause session timeouts, and potentially data loss.

---

*[go/sig/base/as.go, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuWxa7JrnL6AFYWs6m:-KyuWxa8fTyBN5b49fUN:b6c32wt) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L87)):*
> ```Go
> 		err := ae.addNet(ipnet.IPNet())
> 		if err != nil {
> 			log.Error("Unable to add network", "net", ipnet, "err", err)
> ```

This should use the embedded logger (`ae.Error`).

---

*[go/sig/base/as.go, line 101 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuXXJorMKxh7vTEsXV:-KyuXXJorMKxh7vTEsXW:b-mq2d86) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L101)):*
> ```Go
> 		for _, ipnet := range ipnets {
> 			if ne.Net.String() == ipnet.IPNet().String() {
> 				found = true
> ```

This could be simplifed by doing `continue Top`, to restart at the outter-most for loop. It removes the need for the `found` bool, and the `if` statement below.

---

*[go/sig/base/as.go, line 108 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuX0nzXN7C2k3hBREq:-KyuX0nzXN7C2k3hBREr:buj4p2f) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L108)):*
> ```Go
> 			err := ae.delNet(ne.Net)
> 			if err != nil {
> 				log.Error("Unable to delete network", "NetEntry", ne, "err", err)
> ```

`ae.Error`

---

*[go/sig/base/as.go, line 202 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuZDCmyPvAQTIObWFS:-KyuZDCmyPvAQTIObWFT:b-85jboc) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L202)):*
> ```Go
> }
> 
> // AddSig idempotently adds a SIG for the remote IA.
> ```

This is no longer idempotent.

---

*[go/sig/base/as.go, line 228 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuZWPxoNV99V_fbXS-:-KyuZWPxoNV99V_fbXS0:b52j87u) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L228)):*
> ```Go
> 	// racing.
> 	sig := siginfo.NewSig(ae.IA, id, addr.HostFromIP(ip), ctrlPort, encapPort, static)
> 	ae.Sigs[id] = sig
> ```

This throws away all the existing stats.

---

*[go/sig/base/as.go, line 247 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuZcpLFLhGtk5iLHKp:-KyuZcpLFLhGtk5iLHKq:b-rlpf49) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L247)):*
> ```Go
> 	delete(ae.Sigs, id)
> 	ae.Info("Removed SIG", "id", id)
> 	return se.Cleanup()
> ```

This also puts the cleanup inside the locked section.

---

*[go/sig/base/as.go, line 309 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuYI2xNwo5aITvhadv:-KyuYI2xNwo5aITvhadw:b5mqyh7) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L309)):*
> ```Go
> }
> 
> func (ae *ASEntry) createTunDevice() error {
> ```

This isn't a good name, as it does a lot more than just create a Tun device.

---

*[go/sig/base/map.go, line 42 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuMEz_ZPntQwTdQ6b9:-KyuMEz_ZPntQwTdQ6bA:blauqyg) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L42)):*
> ```Go
> 
> func (am *ASMap) ReloadConfig(cfg *config.Cfg) bool {
> 	// Run this as a single transaction under lock to prevent races while
> ```

What races? The map is only modified from a single place, `loadConfig`. What is it you want to protect against?

---

*[go/sig/base/map.go, line 76 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuNvEB4f3LlhYoL_bQ:-KyuNvEB4f3LlhYoL_bR:b8ayejg) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L76)):*
> ```Go
> 	for iaVal := range am.t {
> 		ia := iaVal.IA()
> 		if _, ok := cfg.ASes[*ia]; !ok {
> ```

You could just use `iaVal` here instead of `*ia`.

---

*[go/sig/base/map.go, line 131 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuKQJIVS_jVNKWC1vP:-KyuKQJIVS_jVNKWC1vQ:b6l0kok) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L131)):*
> ```Go
> 	}
> 	delete(am.t, key)
> 	return ae.Cleanup()
> ```

This change moves the cleanup to _inside_ the lock. Is that on purpose? It defeats the purpose of the existing lock approach.

---

*[go/sig/config/config.go, line 94 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuPVyE4PfQDVWaj3L7:-KyuPVyE4PfQDVWaj3L8:b-3g6ap4) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/config/config.go#L94)):*
> ```Go
> type SIGSet map[siginfo.SigIdType]*SIG
> 
> func (ss *SIGSet) UnmarshalJSON(b []byte) error {
> ```

I've spent a couple of minute trying to figure out what the benefit of this is. AFACT, the only difference it makes is that the SIG id is also set inside the config.SIG entry, is that correct? If so it seems a lot simpler to add 3 lines of code to Parse instead.

---

*[go/sig/egress/sessmon.go, line 98 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuT_A-tbv6fzBKoDB4:-KyuT_A-tbv6fzBKoDB5:bftyao1) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/egress/sessmon.go#L98)):*
> ```Go
> 		}
> 	}
> 	err := disp.Dispatcher.Unregister(disp.RegPollRep, disp.MkRegPollKey(sm.sess.IA, sm.sess.SessId))
> ```

Line length.

---

*[go/sig/egress/sessmon.go, line 100 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuTd6RHIMA9hq2E4jA:-KyuTd6RHIMA9hq2E4jB:b-k264ve) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/egress/sessmon.go#L100)):*
> ```Go
> 	err := disp.Dispatcher.Unregister(disp.RegPollRep, disp.MkRegPollKey(sm.sess.IA, sm.sess.SessId))
> 	if err != nil {
> 		log.Crit("sessMonitor: unable to unregister from ctrl dispatcher", "err", err)
> ```

This should probably be `Error` rather than `Crit`, as it's not a bit enough error to exit the SIG.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1322)*

kormat:
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[go/sig/base/as.go, line 189 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyudWlYkRVIO7siTcQo:-KyudWlYkRVIO7siTcQp:b9dlhza) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L189)):*
> ```Go
> 	for _, sig := range ae.Sigs {
> 		if sig.Static {
> 			continue
> ```

This is the opposite logic. Config reloading should, at most, remove statically configured SIGs that are no longer present (and even then it's questionable, because they may be in use dynamically regardless).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1322)*

scrye:
Review status: 6 of 9 files reviewed at latest revision, 15 unresolved discussions.

---

*[go/sig/base/as.go, line 72 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuW80ijEKWTuXWKdjU:-Kyv0rpFuhI9uKajr-aG:bv1ytyu) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L72)):*

I'm concerned by how coarse the locking is here. E.g. the session monitor calls SigMap every 500ms, and in the future will want to call `AddSIG`/`DelSIG` etc. If reconfiguring is slow, that can cause session timeouts, and potentially data loss.
</blockquote></details>

This taking over 500ms is a big red flag, as there is barely any blocking stuff.

However, it's simple enough to add granular locking so I did it anyway.

---

*[go/sig/base/as.go, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuWxa7JrnL6AFYWs6m:-Kyv3Z4GO141sRikSmeO:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L87)):*

This should use the embedded logger (`ae.Error`).
</blockquote></details>

Done.

---

*[go/sig/base/as.go, line 101 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuXXJorMKxh7vTEsXV:-Kyv4AvJD6Y1GK_8tUX4:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L101)):*

This could be simplifed by doing `continue Top`, to restart at the outter-most for loop. It removes the need for the `found` bool, and the `if` statement below.
</blockquote></details>

Done.

---

*[go/sig/base/as.go, line 108 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuX0nzXN7C2k3hBREq:-Kyv4BjJfp2K0ckogez7:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L108)):*

`ae.Error`
</blockquote></details>

Done.

---

*[go/sig/base/as.go, line 189 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyudWlYkRVIO7siTcQo:-Kyv4c3wfhAcMJbikmAW:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L189)):*

This is the opposite logic. Config reloading should, at most, remove statically configured SIGs that are no longer present (and even then it's questionable, because they may be in use dynamically regardless).
</blockquote></details>

Done.

---

*[go/sig/base/as.go, line 202 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuZDCmyPvAQTIObWFS:-KyvB2K-_P5E33DKH4Lc:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L202)):*

This is no longer idempotent.
</blockquote></details>

Done.

---

*[go/sig/base/as.go, line 228 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuZWPxoNV99V_fbXS-:-Kyv8NVJbg3Ta4VInA7n:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L228)):*

This throws away all the existing stats.
</blockquote></details>

Done.

---

*[go/sig/base/as.go, line 247 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuZcpLFLhGtk5iLHKp:-KyvBJDAdwP8IECR86RV:b283qcu) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L247)):*

This also puts the cleanup inside the locked section.
</blockquote></details>

For now, SIG cleanup is a no-op. If we ever have blocking code in the cleanup we can always move it outside the critical region.

---

*[go/sig/base/as.go, line 309 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuYI2xNwo5aITvhadv:-KyvBo3aD-405Gk97Cd8:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L309)):*

This isn't a good name, as it does a lot more than just create a Tun device.
</blockquote></details>

Done.

---

*[go/sig/base/map.go, line 42 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuMEz_ZPntQwTdQ6b9:-KyvBzqVQWEpziWbdHL5:bgj1bgd) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L42)):*

What races? The map is only modified from a single place, `loadConfig`. What is it you want to protect against?
</blockquote></details>

By design it's a synchronized map, so by definition it should protect against any races. The fact that there's a single writer makes computing the list of ASes to delete safe, but it breaks encapsulation for almost no benefits (I doubt there'll be a lot of contention on this lock).

Come to think of it, why was it locked if there's a single reader/writer?

---

*[go/sig/base/map.go, line 76 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuNvEB4f3LlhYoL_bQ:-KyvDnwz-7jaMaRZmNWH:bxc6dv2) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L76)):*

You could just use `iaVal` here instead of `*ia`.
</blockquote></details>

`iaVal` is the wrong type (`addr.ISD_AS` vs `addr.IAInt`).

---

*[go/sig/base/map.go, line 131 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuKQJIVS_jVNKWC1vP:-KyvEKy7_UUNTqwPybuS:bhk99yu) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L131)):*

This change moves the cleanup to _inside_ the lock. Is that on purpose? It defeats the purpose of the existing lock approach.
</blockquote></details>

It's on purpose, yes. I did not see any issues with holding the lock while reloading for a bit longer. This goroutine has to wait for cleanup anyway, and no other goroutine is trying to grab the lock.

If the need ever arises for faster cleanup, we can always move it out of the lock.

---

*[go/sig/config/config.go, line 94 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuPVyE4PfQDVWaj3L7:-KyvGBlD421_w-5mGQqH:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/config/config.go#L94)):*

I've spent a couple of minute trying to figure out what the benefit of this is. AFACT, the only difference it makes is that the SIG id is also set inside the config.SIG entry, is that correct? If so it seems a lot simpler to add 3 lines of code to Parse instead.
</blockquote></details>

Done.

---

*[go/sig/egress/sessmon.go, line 98 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuT_A-tbv6fzBKoDB4:-KyvGI-EfbElMD16wZnc:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/egress/sessmon.go#L98)):*

Line length.
</blockquote></details>

Done.

---

*[go/sig/egress/sessmon.go, line 100 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuTd6RHIMA9hq2E4jA:-KyvGKzth-Aco0pwKYvu:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/egress/sessmon.go#L100)):*

This should probably be `Error` rather than `Crit`, as it's not a bit enough error to exit the SIG.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1322)*

kormat:
Reviewed 3 of 3 files at r2.

---

*[go/sig/base/as.go, line 247 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuZcpLFLhGtk5iLHKp:-Kyzrdxe9ZeHVISsVZWA:b-4cosg7) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L247)):*

For now, SIG cleanup is a no-op. If we ever have blocking code in the cleanup we can always move it outside the critical region.
</blockquote></details>

Responded in the ASMap cleanup thread.

---

*[go/sig/base/map.go, line 42 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuMEz_ZPntQwTdQ6b9:-Kyzy3c8GCW1DK5Ebkzm:b-7rfu8u) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L42)):*

By design it's a synchronized map, so by definition it should protect against any races. The fact that there's a single writer makes computing the list of ASes to delete safe, but it breaks encapsulation for almost no benefits (I doubt there'll be a lot of contention on this lock).

Come to think of it, why was it locked if there's a single reader/writer?
</blockquote></details>

The intention was that there was a single writer (`loadConfig`), and potentially multiple readers (e.g. calling `ASEntry()`). E.g. some kind of maintenance goroutine that iterates over the map inspecting the ASEntries. If you look at the comment in the type definition, it talks about replacing it with `sync.Map` in go 1.9, which wouldn't have a single lock guarding the whole map.

---

*[go/sig/base/map.go, line 76 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuNvEB4f3LlhYoL_bQ:-KyzrqYwNAtGgdyFjzOn:bn4a889) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L76)):*

`iaVal` is the wrong type (`addr.ISD_AS` vs `addr.IAInt`).
</blockquote></details>

Oh, in that case `iaVal` is misnamed. It should be `iaInt`.

---

*[go/sig/base/map.go, line 131 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuKQJIVS_jVNKWC1vP:-KyzzP-xkLqy3yuNnz77:b-ej1ite) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L131)):*
> It's on purpose, yes. I did not see any issues with holding the lock while reloading for a bit longer. This goroutine has to wait for cleanup anyway, and no other goroutine is trying to grab the lock.

It's true that currently there's no other goroutine, but the intention was to allow that in the future.

> If the need ever arises for faster cleanup, we can always move it out of the lock.

Do you have an approach in mind? The lock is grabbed outside this method.

---

*[go/sig/config/config.go, line 59 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyzsnGThzdGHM6r51vW:-KyzsnGThzdGHM6r51vX:bowkk4f) ([raw file](https://github.com/netsec-ethz/scion/blob/ed4fa05d8c0e939e9687c1a09beaa460e26784fb/go/sig/config/config.go#L59)):*
> ```Go
> 	for _, as := range cfg.ASes {
> 		for id, sig := range as.Sigs {
> 			sig.Id = id
> ```

Just a note that this approach only works because `sig` is a pointer. If it was a value, this would be a no-op. I prefer to use `for id := range as.Sigs { sig := as.Sigs[id]; sig.Id = id }` to avoid this potential trap.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1322)*

scrye:
Review status: 4 of 10 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/sig/base/as.go, line 72 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuW80ijEKWTuXWKdjU:-Kz-gOXlxNS3ikpHGy4l:bx4cu60) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/as.go#L72)):*

This taking over 500ms is a big red flag, as there is barely any blocking stuff.

However, it's simple enough to add granular locking so I did it anyway.
</blockquote></details>

Locking is coarse again, but taking the `SigMap` no longer interacts with the master lock. Only the writer and non-critical readers will wait for the lock, which should be fine.

---

*[go/sig/base/map.go, line 42 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuMEz_ZPntQwTdQ6b9:-Kz-B6MBOVsxzBlJ1rV2:bwazhtf) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L42)):*

The intention was that there was a single writer (`loadConfig`), and potentially multiple readers (e.g. calling `ASEntry()`). E.g. some kind of maintenance goroutine that iterates over the map inspecting the ASEntries. If you look at the comment in the type definition, it talks about replacing it with `sync.Map` in go 1.9, which wouldn't have a single lock guarding the whole map.
</blockquote></details>

Following the offline discussion, changed both `ASMap` and `SigMap` to `sync.Map`.

---

*[go/sig/base/map.go, line 76 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuNvEB4f3LlhYoL_bQ:-Kz-fDko1W0Ppy3WdTdj:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L76)):*

Oh, in that case `iaVal` is misnamed. It should be `iaInt`.
</blockquote></details>

Done.

---

*[go/sig/base/map.go, line 131 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyuKQJIVS_jVNKWC1vP:-Kz-fLmOFvc99CvSLaS0:b-r0yidb) ([raw file](https://github.com/netsec-ethz/scion/blob/727bd165eab0aecbf3dc5dc9cfaccdefbc6c97b3/go/sig/base/map.go#L131)):*

> It's on purpose, yes. I did not see any issues with holding the lock while reloading for a bit longer. This goroutine has to wait for cleanup anyway, and no other goroutine is trying to grab the lock.

It's true that currently there's no other goroutine, but the intention was to allow that in the future.

> If the need ever arises for faster cleanup, we can always move it out of the lock.

Do you have an approach in mind? The lock is grabbed outside this method.
</blockquote></details>

The session monitor no longer grabs any locks, so we can now take the master lock while running `Cleanup()`; generally only writers will grab this lock anyway, of which there is only one.

---

*[go/sig/config/config.go, line 59 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-KyzsnGThzdGHM6r51vW:-Kz-fxS6grFp8VX9a9_b:b-ju8ke2) ([raw file](https://github.com/netsec-ethz/scion/blob/ed4fa05d8c0e939e9687c1a09beaa460e26784fb/go/sig/config/config.go#L59)):*

Just a note that this approach only works because `sig` is a pointer. If it was a value, this would be a no-op. I prefer to use `for id := range as.Sigs { sig := as.Sigs[id]; sig.Id = id }` to avoid this potential trap.
</blockquote></details>

Fair enough, since this has bit us in the past. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1322)*

kormat:
Reviewed 6 of 6 files at r3.

---

*[go/sig/base/as.go, line 243 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-Kz3Ezjs2I_juPeveoxC:-Kz3Ezjs2I_juPeveoxD:b-ptl1pk) ([raw file](https://github.com/netsec-ethz/scion/blob/9db3031486072cc0f8d78c45a98701003111934a/go/sig/base/as.go#L243)):*
> ```Go
> }
> 
> // Internal method to return a *copy* of the ASEntry's SigMap
> ```

Outdated (but as above we can probably just get rid of this method).

---

*[go/sig/base/map.go, line 105 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-Kz3CbU1hEtlJD_37YRd:-Kz3CbU1hEtlJD_37YRe:b-fib7t4) ([raw file](https://github.com/netsec-ethz/scion/blob/9db3031486072cc0f8d78c45a98701003111934a/go/sig/base/map.go#L105)):*
> ```Go
> 				log.Error(cerr.Desc, cerr.Ctx...)
> 				s = false
> 				return true
> ```

This doesn't look right.

---

*[go/sig/egress/session.go, line 42 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-Kz3DJF3kIaQP_liDnrm:-Kz3DJF3kIaQP_liDnrn:bk6fdmu) ([raw file](https://github.com/netsec-ethz/scion/blob/9db3031486072cc0f8d78c45a98701003111934a/go/sig/egress/session.go#L42)):*
> ```Go
> 	pool *pathmgr.SyncPaths
> 	// function pointer to return SigMap from parent ASEntry.
> 	sigMapF func() *siginfo.SigMap
> ```

Hmm. Instead of passing around a callback, we could just pass in a reference to the SigMap directly now that we don't need to copy it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1322)*

scrye:
Review status: 7 of 10 files reviewed at latest revision, 3 unresolved discussions.

---

*[go/sig/base/as.go, line 243 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-Kz3Ezjs2I_juPeveoxC:-Kz3GpzPJSHlyiGj_0HG:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9db3031486072cc0f8d78c45a98701003111934a/go/sig/base/as.go#L243)):*

Outdated (but as above we can probably just get rid of this method).
</blockquote></details>

Done.

---

*[go/sig/base/map.go, line 105 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-Kz3CbU1hEtlJD_37YRd:-Kz3F_a5gN4bYTA9tjED:bj623k0) ([raw file](https://github.com/netsec-ethz/scion/blob/9db3031486072cc0f8d78c45a98701003111934a/go/sig/base/map.go#L105)):*

This doesn't look right.
</blockquote></details>

There are two semantics at work here.

`s` contains the exit (success) status of parent function `delOldIAs`, referenced in the closure.

`return true` is in closure scope and tells `Range` to continue to the next entry.

---

*[go/sig/egress/session.go, line 42 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-Kz3DJF3kIaQP_liDnrm:-Kz3GMW90lbc-xw8mLH_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9db3031486072cc0f8d78c45a98701003111934a/go/sig/egress/session.go#L42)):*

Hmm. Instead of passing around a callback, we could just pass in a reference to the SigMap directly now that we don't need to copy it.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1322)*

kormat:
Reviewed 3 of 3 files at r4.

---

*[go/sig/base/map.go, line 105 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-Kz3CbU1hEtlJD_37YRd:-Kz3JS99JECrarwJABMV:b-4vy6xd) ([raw file](https://github.com/netsec-ethz/scion/blob/9db3031486072cc0f8d78c45a98701003111934a/go/sig/base/map.go#L105)):*

There are two semantics at work here.

`s` contains the exit (success) status of parent function `delOldIAs`, referenced in the closure.

`return true` is in closure scope and tells `Range` to continue to the next entry.
</blockquote></details>

Ohh. Right :)

---

*[go/sig/egress/sessmon.go, line 91 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-Kz3JioUFrS-L_2vfdhh:-Kz3JioUFrS-L_2vfdhi:b-nf5hy6) ([raw file](https://github.com/netsec-ethz/scion/blob/9db3031486072cc0f8d78c45a98701003111934a/go/sig/egress/sessmon.go#L91)):*
> ```Go
> 			// Update paths and sigs
> 			sm.sessPathPool.update(sm.pool.Load().APS)
> 			sm.sigMap = sm.sess.sigMapF()
> ```

`sm.sigMap` could be removed, and replaces with direct calls to `sm.sess.sigMap`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1322)*

scrye:

---

*[go/sig/egress/sessmon.go, line 91 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-Kz3JioUFrS-L_2vfdhh:-Kz3KLujiDoElf-BXvHZ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/9db3031486072cc0f8d78c45a98701003111934a/go/sig/egress/sessmon.go#L91)):*

`sm.sigMap` could be removed, and replaces with direct calls to `sm.sess.sigMap`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1322)*

kormat:
:lgtm:

---

Reviewed 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1322#-:-Kz3KlL2poj8LQQk3vno:bnfp4nl)*

PR #1309, opened 2017-11-01
TRC in go

Oncilla:
This PR adds TRC support to go.

trc/entries.go is used to parse the dictionary entries of the certificate.
trc/trc.go is the TRC implementation.

Additionally,  this PR syncs the TRC format with the book. (See issue #933):
 - add "ExpirationTime"
 - rename "ISDID" -> "ISD"
 - rename "QuorumEEPKI" -> "ThresholdEEPKI"
 - change CertLog entry from public key to cert
 - add dict to RootCAs entries with ("Certificate", "OnlineKeyAlg", "OnlineKey", "TRCSrv", "ARPKISrv","ARPKIKey")
 - rename "RootRainsKey" -> "RAINS"
 - add dict to "RAINS" with ("RootRAINSKey", "OnlineKeyAlg", "OnlineKey", "TRCSrv")

The python code is adopted accordingly.

The code is also being prepared for TRC updates and TRC cross signatures.
Depending on the ISD of the trusted TRC, another verification method has to be used.
This is abstracted away by the verify function, which selects the right method.

For now, the generator creates empty dictionaries for the fields [RootCAs, CertLogs, RAINS],
however, parsing is supported by both python and go code.

Test cases will follow.

Verification based on cross signature and verification of signatures of root CAs and Rains are not implemented yet.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1309)

PR #1325, opened 2017-11-13
Initial Certificate Server in go

Oncilla:
This is an initial commit adding go implementation of the certificate server.
The current version supports certificate chain requests and replies.

Support for TRCs will be added in a future PR.

main.go is the entry point and initializes the certificate server.
io.go handles incoming and outgoing scion packets.
cert.go handles certificate chain requests and replies.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1325)


scrye:
Reviewed 5 of 5 files at r1.

---

*[go/cs/cert.go, line 46 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyulQOTLuNL-QgvSpjX:-KyulQOTLuNL-QgvSpjY:b-xwd2nf) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L46)):*
> ```Go
> 		}
> 	} else if !local || req.CacheOnly {
> 		log.Info("Dropping certificate chain requeset", "addr", addr, "req", req, "err",
> ```

s/requeset/request

---

*[go/cs/cert.go, line 70 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuOAC_NLScrmgmZ1_U:-KyuOAC_NLScrmgmZ1_V:bemmm4c) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L70)):*
> ```Go
> 
> // fetchChain fetches certificate chain from the remote AS.
> func fetchChain(addr *snet.Addr, req *cert_mgmt.ChainReq, conn *snet.Conn, pool *msg.BufPool) error {
> ```

Line length > 100

---

*[go/cs/cert.go, line 71 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyusyaxB2gHQfB4HjGp:-KyusyaxB2gHQfB4HjGq:bea387r) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L71)):*
> ```Go
> // fetchChain fetches certificate chain from the remote AS.
> func fetchChain(addr *snet.Addr, req *cert_mgmt.ChainReq, conn *snet.Conn, pool *msg.BufPool) error {
> 	key := (&cert.Key{IA: *req.IA(), Ver: int(req.Version)}).String()
> ```

This needs to be refactored to somehow call method `cert.Chain.Key`, otherwise a change to one will break the other.

---

*[go/cs/cert.go, line 75 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyun0EqCYLRajyGA7cQ:-Kyun0EqCYLRajyGA7cR:b-6141lo) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L75)):*
> ```Go
> 	if sendReq { // rate limit
> 		return sendChainReq(req, conn, pool)
> 	}
> ```

Log requests that are ignored because there's one already pending and `delta` time hasn't passed.

---

*[go/cs/cert.go, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyun_9psmOvYQNUbIJu:-Kyun_9psmOvYQNUbIJv:b-7itczu) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L87)):*
> ```Go
> 	}
> 	a := &snet.Addr{IA: req.IA(), Host: addr.SvcCS, L4Port: 0}
> 	log.Debug("Send Chain reply", "req", req, "addr", a)
> ```

Message should be "Send chain request"

---

*[go/cs/cs.go, line 39 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuOlUcKcKZvG4v2kpY:-KyuOlUcKcKZvG4v2kpZ:bemmm4c) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L39)):*
> ```Go
> 	id             = flag.String("id", "", "Element ID (Required. E.g. 'cs1-10-1')")
> 	sciondPath     = flag.String("sd", "", "SCIOND socket path (Optional if SCIOND_PATH is set)")
> 	dispatcherPath = flag.String("disp", "/run/shm/dispatcher/default.sock", "SCION Dispatcher path (Optional)")
> ```

Line length > 100

---

*[go/cs/cs.go, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyucgi0B5ihZrm6H1hR:-Kyucgi0B5ihZrm6H1hS:bkxsybk) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L63)):*
> ```Go
> 	}
> 	// initialize snet with retries
> 	if err = initSNET(100, 100*time.Millisecond); err != nil {
> ```

Change these two (`100` and `100*time.Milliseconds`) to constants. Also the timer seems a bit aggressive, maybe try reconnecting once per second?

---

*[go/cs/cs.go, line 78 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuWXi2M8eVVJ4oHFeI:-KyuWXi2M8eVVJ4oHFeJ:b-8ymmsf) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L78)):*
> ```Go
> func checkFlags() {
> 	if *id == "" {
> 		log.Crit("No element ID specified")
> ```

Also print usage details here (`flag.Usage()`).

---

*[go/cs/cs.go, line 80 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuW7AMWU6Ruk5Yl--h:-KyuW7AMWU6Ruk5Yl--i:bng6jm4) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L80)):*
> ```Go
> 		log.Crit("No element ID specified")
> 		os.Exit(1)
> 	}
> ```

It's desirable to initialize logging as soon as possible, so any future messages get sent to the file instead.

---

*[go/cs/cs.go, line 99 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuZzcpbbXghL9t5h7C:-KyuZzcpbbXghL9t5h7D:b-1rand3) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L99)):*
> ```Go
> 		return common.NewCError("Cannot load topology", "err", err)
> 	}
> 	topoId, ok := topo.CS[*id]
> ```

Variable name `topoId` seems a bit misleading, since this contains an address. Wouldn't `topoAddr` also work?

---

*[go/cs/cs.go, line 101 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyubd5QeoVv8vuY2Ib0:-Kyubd5QeoVv8vuY2Ib1:b-s85ed) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L101)):*
> ```Go
> 	topoId, ok := topo.CS[*id]
> 	if !ok {
> 		return common.NewCError("Cannot load BindAddress. Element ID not found", "id", *id)
> ```

Minor point: you are alternating between `Unable` and `Cannot` for error messages; the Go codebase tends to use `Unable`.

---

*[go/cs/cs.go, line 104 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuXyxFlrlT6FOvgbWN:-KyuXyxFlrlT6FOvgbWO:bemmm4c) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L104)):*
> ```Go
> 	}
> 	bindInfo := topoId.BindAddrInfo(topo.Overlay)
> 	local = &snet.Addr{IA: topo.ISD_AS, Host: addr.HostFromIP(bindInfo.IP), L4Port: uint16(bindInfo.L4Port)}
> ```

Line length > 100

---

*[go/cs/cs.go, line 113 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuaqZiqzv5iLgbtuzS:-KyuaqZiqzv5iLgbtuzT:bf2hlta) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L113)):*
> ```Go
> 	// Initialize SCION local networking module
> 	err = snet.Init(local.IA, *sciondPath, *dispatcherPath)
> 	for i := 0; i < attempts; i++ {
> ```

This looks like it runs `Init` `attempts + 1` times before giving up. Why not just include `snet.Init()` at the start of the loop?

---

*[go/cs/cs.go, line 117 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyua_-5HAu1qyjWgURb:-Kyua_-5HAu1qyjWgURc:bq8l87o) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L117)):*
> ```Go
> 			break
> 		}
> 		log.Crit("sleep", "i", i)
> ```

This doesn't say a lot about what's happening. Maybe something like "Unable to initialize snet, retrying in ...s"; include the error and the number of retries left in the context.

---

*[go/cs/io.go, line 49 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyufMdzfHkSUIaFgxSh:-KyufMdzfHkSUIaFgxSi:b-z3sk70) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/io.go#L49)):*
> ```Go
> 		read, addr, err := d.conn.ReadFromSCION(buf.Raw)
> 		if err != nil {
> 			log.Error("Unable to read from External Ingress", "err", err)
> ```

External Ingress is a bit confusing here, since this terminology tends to be used for routers. `Unable to read from network` should be enough, since this is the only place in the code where you read anyway.

---

*[go/cs/io.go, line 96 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuiULg8Z84XH9e_Js0:-KyuiULh24cmI0yDqM4T:be65m96) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/io.go#L96)):*
> ```Go
> 
> // SendPayload is used to send payloads to the specified address using snet.
> func SendPayload(addr *snet.Addr, cpld *ctrl.Pld, conn *snet.Conn, pool *msg.BufPool) error {
> ```

The ordering here feels a bit unintuitive. Maybe conn, cpld, addr, pool?

---

*[go/cs/io.go, line 103 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuhDtSdCehgmLejXyM:-KyuhDtSdCehgmLejXyN:b5l9yhh) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/io.go#L103)):*
> ```Go
> 		return err
> 	}
> 	_, err = conn.WriteToSCION(buf.Raw[:n], addr) // FIXME(roosd): handle incomplete writes
> ```

Can you clarify what you mean by `handle incomplete write` here? Due to the nature of the underlying reliable socket between library and the C dispatcher, all `snet.Write` send everything as a single message.

You might silently send a message that is over the MTU, as `snet` currently does no MTU checks. It's something we've been wanting to solve for a while but haven't gotten to yet.

---

*[go/cs/req_cache.go, line 31 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuLtI5DzlF0oqLboa2:-KyuLtI5DzlF0oqLboa3:b-5t8xw5) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/req_cache.go#L31)):*
Go recommends putting the lock above the fields it protects: 

> Grouping can also indicate relationships between items, such as the fact that a set of variables is protected by a mutex.

(Taken from Effective Go, https://golang.org/doc/effective_go.html)

---

*[go/cs/req_cache.go, line 43 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyumc7xMiLtlF59XsVW:-Kyumc7xMiLtlF59XsVX:b-cssie9) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/req_cache.go#L43)):*
> ```Go
> }
> 
> // Put adds an address to the expiring cache. The return value indicates whether a new request shall be issued.
> ```

Since calling code issues the request, maybe s/shall/should is clearer?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1325)*

Oncilla:
Review status: 1 of 6 files reviewed at latest revision, 19 unresolved discussions.

---

*[go/cs/cert.go, line 46 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyulQOTLuNL-QgvSpjX:-Kyuw47jp2urdxsFlLkg:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L46)):*

s/requeset/request
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 70 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuOAC_NLScrmgmZ1_U:-Kyuw7ktgwDZ7FvSFUGY:bydqeqq) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L70)):*

Line length > 100
</blockquote></details>

I'm not sure how to solve this properly.
A line break seems odd to me here.

---

*[go/cs/cert.go, line 71 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyusyaxB2gHQfB4HjGp:-KyuxoPNxXcaOeQFvs28:b-38c14a) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L71)):*

This needs to be refactored to somehow call method `cert.Chain.Key`, otherwise a change to one will break the other.
</blockquote></details>

I added cert.NewKey in order to have a consistent constructor.

---

*[go/cs/cert.go, line 75 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyun0EqCYLRajyGA7cQ:-KyuzHwnwGsAD87PtGjb:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L75)):*

Log requests that are ignored because there's one already pending and `delta` time hasn't passed.
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyun_9psmOvYQNUbIJu:-KyuzKeJV_PNPtQguffh:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L87)):*

Message should be "Send chain request"
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 39 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuOlUcKcKZvG4v2kpY:-Kyv5fT7_Go-SGIPjYMp:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L39)):*

Line length > 100
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 63 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyucgi0B5ihZrm6H1hR:-Kyuzzjd3WmMGOtf9jyb:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L63)):*

Change these two (`100` and `100*time.Milliseconds`) to constants. Also the timer seems a bit aggressive, maybe try reconnecting once per second?
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 78 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuWXi2M8eVVJ4oHFeI:-Kyv-7iR6gSURnToYYJy:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L78)):*

Also print usage details here (`flag.Usage()`).
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 80 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuW7AMWU6Ruk5Yl--h:-Kyv-htJ-68D0YawFDgz:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L80)):*

It's desirable to initialize logging as soon as possible, so any future messages get sent to the file instead.
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 99 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuZzcpbbXghL9t5h7C:-Kyv-r8DewvxBwW-Rv7h:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L99)):*

Variable name `topoId` seems a bit misleading, since this contains an address. Wouldn't `topoAddr` also work?
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 101 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyubd5QeoVv8vuY2Ib0:-Kyv0769k3lcXbU7f-uM:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L101)):*

Minor point: you are alternating between `Unable` and `Cannot` for error messages; the Go codebase tends to use `Unable`.
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 104 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuXyxFlrlT6FOvgbWN:-Kyv0F3-BRW-l5swoSJn:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L104)):*

Line length > 100
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 113 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuaqZiqzv5iLgbtuzS:-Kyv0Q6rSG8uto9bPMM7:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L113)):*

This looks like it runs `Init` `attempts + 1` times before giving up. Why not just include `snet.Init()` at the start of the loop?
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 117 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyua_-5HAu1qyjWgURb:-Kyv1omwtMcnTeV6LDty:b-m02y80) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cs.go#L117)):*

This doesn't say a lot about what's happening. Maybe something like "Unable to initialize snet, retrying in ...s"; include the error and the number of retries left in the context.
</blockquote></details>

True. Some debugging remains.

---

*[go/cs/io.go, line 49 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyufMdzfHkSUIaFgxSh:-Kyv2P8C8MF0hu41Urfp:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/io.go#L49)):*

External Ingress is a bit confusing here, since this terminology tends to be used for routers. `Unable to read from network` should be enough, since this is the only place in the code where you read anyway.
</blockquote></details>

Done.

---

*[go/cs/io.go, line 96 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuiULg8Z84XH9e_Js0:-Kyv3TDAg2wkzXG6CTRU:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/io.go#L96)):*

The ordering here feels a bit unintuitive. Maybe conn, cpld, addr, pool?
</blockquote></details>

Done.

---

*[go/cs/io.go, line 103 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuhDtSdCehgmLejXyM:-Kyv3hz3LlvF560JBSxC:b-en9dqz) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/io.go#L103)):*

Can you clarify what you mean by `handle incomplete write` here? Due to the nature of the underlying reliable socket between library and the C dispatcher, all `snet.Write` send everything as a single message.

You might silently send a message that is over the MTU, as `snet` currently does no MTU checks. It's something we've been wanting to solve for a while but haven't gotten to yet.
</blockquote></details>

I was assuming conn.WriteToSCION might write less then n bytes. But if that is not the case, then we can drop this comment.
But would the sender be notified through an error or just a number smaller than the number of bytes?

---

*[go/cs/req_cache.go, line 31 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuLtI5DzlF0oqLboa2:-Kyv4kytgCJKBC1VsoM1:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/req_cache.go#L31)):*

Go recommends putting the lock above the fields it protects: 

> Grouping can also indicate relationships between items, such as the fact that a set of variables is protected by a mutex.

(Taken from Effective Go, https://golang.org/doc/effective_go.html)
</blockquote></details>

Done.

---

*[go/cs/req_cache.go, line 43 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyumc7xMiLtlF59XsVW:-Kyv57yR9D2WCU8Funuj:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/req_cache.go#L43)):*

Since calling code issues the request, maybe s/shall/should is clearer?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1325)*

scrye:
Reviewed 5 of 5 files at r2.

---

*[go/cs/cert.go, line 70 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuOAC_NLScrmgmZ1_U:-KyvIPMXkXHlhphxK-w7:b1v5xg3) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L70)):*

I'm not sure how to solve this properly.
A line break seems odd to me here.
</blockquote></details>

Well, it's also a sign that the function's signature is getting a bit fat.

I've noticed that a lot of the functions in this file share the same Conn, maybe wrapping Conn into a new structure and having them as methods on that structure? They could also share the same buffer pool reference.

---

*[go/cs/cert.go, line 75 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyun0EqCYLRajyGA7cQ:-KyvK6gD8N-HaF9LL44p:bqaj753) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L75)):*

Done.
</blockquote></details>

The function should be outside the `if`? (the `if` is for sending new requests, not dropping them).

Also, `Ignoring certificate chain request (same request already pending)` might be a bit clearer on what's happening.

---

*[go/cs/io.go, line 103 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuhDtSdCehgmLejXyM:-KyvLTuEJwKQEMwV2vwH:bfrj8az) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/io.go#L103)):*

I was assuming conn.WriteToSCION might write less then n bytes. But if that is not the case, then we can drop this comment.
But would the sender be notified through an error or just a number smaller than the number of bytes?
</blockquote></details>

Yes, `WriteToSCION` always writes everything.

Regarding MTU, unfortunately, no, you don't get any error -- that's why it's so dangerous. I've filed issue https://github.com/netsec-ethz/scion/issues/1328 so we don't lose track of this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1325)*

Oncilla:
Review status: 4 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[go/cs/cert.go, line 70 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KyuOAC_NLScrmgmZ1_U:-KyvSSpq9-scgkq5Q9nQ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L70)):*

Well, it's also a sign that the function's signature is getting a bit fat.

I've noticed that a lot of the functions in this file share the same Conn, maybe wrapping Conn into a new structure and having them as methods on that structure? They could also share the same buffer pool reference.
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 75 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-Kyun0EqCYLRajyGA7cQ:-KyvT-OKjcq8SuSgAIa-:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7c41e16a1ece010d335df3d1e8fe2f5dec16153e/go/cs/cert.go#L75)):*

The function should be outside the `if`? (the `if` is for sending new requests, not dropping them).

Also, `Ignoring certificate chain request (same request already pending)` might be a bit clearer on what's happening.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1325)*

kormat:
`go/cs` is a bit too cryptic. `go/cert_server`, or `go/cert_srv` would be better. or even `go/certd`.

---

Reviewed 1 of 5 files at r1, 3 of 5 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 25 unresolved discussions, some commit checks failed.

---

*[go/cs/cert.go, line 32 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTyKcFBFrYx1GtA-lA:-KzTyKcFBFrYx1GtA-lB:b63kn0p) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L32)):*
> ```Go
> var (
> 	// chainReqCache is an expiring cache for pending requests of certificate chains.
> 	chainReqCache = NewReqCache(30*time.Second, 10*time.Minute, 100*time.Millisecond)
> ```

100ms is a bit too aggressive for retries. I'd suggest 1s instead.

---

*[go/cs/cert.go, line 48 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTths1o2sVRsklsMNF:-KzTths1o2sVRsklsMNG:b-14nn0b) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L48)):*
> ```Go
> func (h *ChainHandler) HandleReq(addr *snet.Addr, req *cert_mgmt.ChainReq) {
> 	log.Info("Received certificate chain request", "addr", addr, "req", req)
> 	chain := store.GetChain(req.IA(), int(req.Version))
> ```

Ah. This indicates that `store.GetChain` takes an `int`, rather than a fixed-size. As mentioned on 
#1309, i suggest switching to `uint64`. Ah, damn. Just realised that the types in `cert_mgmt.capnp` need to be changed, too. Damn past me.

---

*[go/cs/cert.go, line 49 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTuI_MBKxdwzJS20RP:-KzTuI_MBKxdwzJS20RQ:b-8f9k4h) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L49)):*
> ```Go
> 	log.Info("Received certificate chain request", "addr", addr, "req", req)
> 	chain := store.GetChain(req.IA(), int(req.Version))
> 	local := local.IA.Eq(addr.IA)
> ```

This shadows the global `local` variable, which can cause great confusion. Suggestion: `srcLocal` or similar.

---

*[go/cs/cert.go, line 52 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTucf-iU3oTNcphlVk:-KzTucf-iU3oTNcphlVl:b-5j06cx) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L52)):*
> ```Go
> 	if chain != nil {
> 		if err := h.sendChainRep(addr, chain); err != nil {
> 			log.Error("Unable to send certificate chain reply", "addr", addr, "err", err)
> ```

Log the request.

---

*[go/cs/cert.go, line 59 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTuq_YFL46PUXoRDjk:-KzTuq_YFL46PUXoRDjl:b-5j06cx) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L59)):*
> ```Go
> 	} else {
> 		if err := h.fetchChain(addr, req); err != nil {
> 			log.Error("Unable to fetch certificate chain", "err", err)
> ```

Log the request.

---

*[go/cs/cert.go, line 92 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTwVG0qGPzpHrSBp9z:-KzTwVG0qGPzpHrSBpA-:b-la1hdg) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L92)):*
> ```Go
> // sendChainReq sends a certificate chain request to the specified remote AS.
> func (h *ChainHandler) sendChainReq(req *cert_mgmt.ChainReq) error {
> 	req.CacheOnly = true
> ```

This seems a bit odd. The caller of `sendChainReq` should be able to set the flags it wants. 

---

*[go/cs/cert.go, line 97 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTx3_H8CFLn_BkVSk7:-KzTx3_H8CFLn_BkVSk8:bw726gx) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L97)):*
> ```Go
> 		return err
> 	}
> 	a := &snet.Addr{IA: req.IA(), Host: addr.SvcCS, L4Port: 0}
> ```

You don't need to specify `L4Port` here, the default will be `0` anyway.

---

*[go/cs/cert.go, line 107 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTxhM8gp3YlnFYV7nI:-KzTxhM8gp3YlnFYV7nJ:bleaoxc) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L107)):*
> ```Go
> 	chain, err := rep.Chain()
> 	if err != nil {
> 		log.Error("Unable to parse certificate reply", "err", err)
> ```

Log the cert Key string.

---

*[go/cs/cert.go, line 110 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTxkjs_LoHS-cJ_Cez:-KzTxkjs_LoHS-cJ_Cf-:bleaoxc) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L110)):*
> ```Go
> 	}
> 	if err = store.AddChain(chain, true); err != nil {
> 		log.Error("Unable to store certificate chain", "err", err)
> ```

Log the cert Key string.

---

*[go/cs/cert.go, line 119 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTxlYnG2v-e3rZmZlK:-KzTxlYnG2v-e3rZmZlL:bleaoxc) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L119)):*
> ```Go
> 	cpld, err := ctrl.NewCertMgmtPld(rep)
> 	if err != nil {
> 		log.Error("Unable to create certificate chain reply", "err", err)
> ```

Log the cert Key string.

---

*[go/cs/cert.go, line 124 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTxmc1pXktvLoXd92p:-KzTxmc1pXktvLoXd92q:bleaoxc) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L124)):*
> ```Go
> 	for _, dst := range reqs.Addrs {
> 		if err := SendPayload(h.conn, cpld, dst, h.pool); err != nil {
> 			log.Error("Unable to write certificate chain reply", "err", err)
> ```

Log the cert Key string.

---

*[go/cs/cs.go, line 1 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTBWyBfpd1iOlOv6Nj:-KzTBWyBfpd1iOlOv6Nk:b-5l1kg3) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cs.go#L1)):*
> ```Go
> ```

Name this `main.go`.

---

*[go/cs/cs.go, line 43 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzT_pjf_6OHAykceD_P:-KzT_pjf_6OHAykceD_Q:b-b6flnq) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cs.go#L43)):*
> ```Go
> var (
> 	id             = flag.String("id", "", "Element ID (Required. E.g. 'cs1-10-1')")
> 	sciondPath     = flag.String("sd", "", "SCIOND socket path (Optional if SCIOND_PATH is set)")
> ```

Have a look at the flag names used in `go/sig/sig.go`, consistency would be good.

---

*[go/cs/cs.go, line 48 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTa61z6vIFa8zMWx8g:-KzTa62-IWxuKCdBcV0M:b-4p2d1q) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cs.go#L48)):*
> ```Go
> 	confDir  = flag.String("confd", "", "Configuration directory (Required)")
> 	cacheDir = flag.String("cached", "gen-cache", "Caching directory (Optional)")
> 	prom     = flag.String("prom", "", "prom")
> ```

This needs a description, e.g. `"Address to export prometheus metrics on"` (from `go/border/metrics/metrics.go`)

---

*[go/cs/cs.go, line 93 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTaVcVMdzFRcR5u9kV:-KzTaVcVMdzFRcR5u9kW:b8gunyd) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cs.go#L93)):*
> ```Go
> 			log.Crit("No SCIOND path specified")
> 			flag.Usage()
> 			os.Exit(1)
> ```

Where possible, exit points from the binary should be in `main()`. Otherwise reading `main()` suggests that things like `checkFlags()` can never error.

---

*[go/cs/io.go, line 37 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTgTCVQQPiTehJ--ze:-KzTgTCVQQPiTehJ--zf:b-qvelh1) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/io.go#L37)):*
> ```Go
> // NewDispatcher creates a new dispatcher listening to SCION traffic on the specified address.
> func NewDispatcher(addr *snet.Addr) (*Dispatcher, error) {
> 	conn, err := snet.ListenSCION("udp4", addr)
> ```

Ah. This made me realise that snet/reliable.go don't support SVC or bind addresses yet (https://github.com/netsec-ethz/scion/issues/1334), so it won't be possible to receive queries or function behind nat until that is fixed.

---

*[go/cs/io.go, line 61 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTpglqeZInTKv8S-RQ:-KzTpglqeZInTKv8S-RR:bfl3evo) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/io.go#L61)):*
> ```Go
> 			log.Error("Unable to dispatch", "err", err)
> 		}
> 		d.bufPool.PutBuf(buf)
> ```

This isn't necessarily safe, as references to parts of the msg.Buf may be retained by the dispatched functions, especially if the message is ever passed off to a go routine for processing. I would probably keep it simple - have a large static buffer that you `ReadFromSCION` into, then allocate a new buffer for the size of data received, and pass that into dispatching, and let it be garbage-collected when it's done. While the CS does need to scale, it's not nearly as important to avoid allocations/copies as, say, the BR or SIG.

---

*[go/cs/io.go, line 102 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTsaCM_sb5FTKwfqdj:-KzTsaCM_sb5FTKwfqdk:b-obn3sp) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/io.go#L102)):*
> ```Go
> 	buf := pool.FetchBuf()
> 	defer pool.PutBuf(buf)
> 	n, err := cpld.WritePld(buf.Raw)
> ```

You can simplify things by just doing `cpld.PackPld()`, and avoid having to allocate/free a buffer which may be the wrong size anyway.

---

*[go/cs/req_cache.go, line 29 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTz3yg54p6HjZy6Eir:-KzTz3yg54p6HjZy6Eis:bhte39w) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/req_cache.go#L29)):*
> ```Go
> type ReqCache struct {
> 	// lock is the lock for synchronizing access cache.
> 	lock sync.RWMutex
> ```

It looks like only writing locks are used, so this could just be `sync.Mutex`.

---

*[go/cs/msg/buf.go, line 37 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTiitY1DB0v2aybIKn:-KzTiitY1DB0v2aybIKo:bcq857i) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/msg/buf.go#L37)):*
> ```Go
> 
> // FetchBuf fetches buffer. The caller shall return the buffer when no longer needed.
> func (p *BufPool) FetchBuf() *Buf {
> ```

It's a bit more elegant to do this instead:
```
func (p *BufPool) Get() *Buf {
    msgBuf := p.Pool.Get().(*Buf)
    ...
}
```
and
```
func (p *BufPool) Put(msgBuf *Buf) {
   p.Pool.Put(msgBuf)
}
```
That way you get something that's very close to the `sync.Pool` api, with type safety.

---

*[go/cs/msg/buf.go, line 54 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTlCdzNTAyyStOrTRP:-KzTlCdzNTAyyStOrTRQ:b-4f7i3t) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/msg/buf.go#L54)):*
> ```Go
> }
> 
> func NewBuf() *Buf {
> ```

This probably doesn't need to be exported.

---

*[go/cs/msg/buf.go, line 60 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTlL45DgevCXkpv4b4:-KzTlL45DgevCXkpv4b5:b-rvul78) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/msg/buf.go#L60)):*
> ```Go
> }
> 
> func (mb *Buf) Reset() {
> ```

Same for this re: exported.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-:-KzTSJb7HuyOypxJoR7c:b-mixv9)*

Oncilla:
certd seems cryptic too. I think cert_srv is the best option.
Done.

---

Review status: all files reviewed at latest revision, 22 unresolved discussions, some commit checks failed.

---

*[go/cs/cert.go, line 32 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTyKcFBFrYx1GtA-lA:-KzUX6SaZprTrw_ZgBce:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L32)):*

100ms is a bit too aggressive for retries. I'd suggest 1s instead.
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 48 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTths1o2sVRsklsMNF:-KzYoD7b2Or2JMnx0VJ7:bb16uba) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L48)):*

Ah. This indicates that `store.GetChain` takes an `int`, rather than a fixed-size. As mentioned on 
#1309, i suggest switching to `uint64`. Ah, damn. Just realised that the types in `cert_mgmt.capnp` need to be changed, too. Damn past me.
</blockquote></details>

Right. This is adopted when #1309 is merged.

---

*[go/cs/cert.go, line 49 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTuI_MBKxdwzJS20RP:-KzUXl2qeHhDoNE5PJDd:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L49)):*

This shadows the global `local` variable, which can cause great confusion. Suggestion: `srcLocal` or similar.
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 52 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTucf-iU3oTNcphlVk:-KzUXuUcLB2GRCk3UlYC:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L52)):*

Log the request.
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 59 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTuq_YFL46PUXoRDjk:-KzUXyeIDh2xYUCE9zJg:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L59)):*

Log the request.
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 92 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTwVG0qGPzpHrSBp9z:-KzUY6WZjrnTGXhBeTw0:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L92)):*

This seems a bit odd. The caller of `sendChainReq` should be able to set the flags it wants. 
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 97 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTx3_H8CFLn_BkVSk7:-KzUZPBBGuBy5uAL1nq6:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L97)):*

You don't need to specify `L4Port` here, the default will be `0` anyway.
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 107 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTxhM8gp3YlnFYV7nI:-KzUZh0-TY_M8Aef9Zka:b-xd6i7e) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L107)):*

Log the cert Key string.
</blockquote></details>

What key string?

---

*[go/cs/cert.go, line 110 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTxkjs_LoHS-cJ_Cez:-KzUZvQQhNj48Xhlng6V:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L110)):*

Log the cert Key string.
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 119 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTxlYnG2v-e3rZmZlK:-KzU_YgvEbnLSso-JLdA:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L119)):*

Log the cert Key string.
</blockquote></details>

Done.

---

*[go/cs/cert.go, line 124 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTxmc1pXktvLoXd92p:-KzU_ZQYCNEZyFgCNbU2:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cert.go#L124)):*

Log the cert Key string.
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 1 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTBWyBfpd1iOlOv6Nj:-KzXv9kXc7OOMnVn7SEw:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cs.go#L1)):*

Name this `main.go`.
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 43 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzT_pjf_6OHAykceD_P:-KzXvBkqIp2rZ_8l9usb:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cs.go#L43)):*

Have a look at the flag names used in `go/sig/sig.go`, consistency would be good.
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 48 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTa61z6vIFa8zMWx8g:-KzXvYGRpo-ZsCowl1WY:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cs.go#L48)):*

This needs a description, e.g. `"Address to export prometheus metrics on"` (from `go/border/metrics/metrics.go`)
</blockquote></details>

Done.

---

*[go/cs/cs.go, line 93 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTaVcVMdzFRcR5u9kV:-KzYO5k5C5z5nJ9WE_jO:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/cs.go#L93)):*

Where possible, exit points from the binary should be in `main()`. Otherwise reading `main()` suggests that things like `checkFlags()` can never error.
</blockquote></details>

Done.

---

*[go/cs/io.go, line 37 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTgTCVQQPiTehJ--ze:-KzYOsk3W1GF7NXC5zAx:b-ypsav9) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/io.go#L37)):*

Ah. This made me realise that snet/reliable.go don't support SVC or bind addresses yet (https://github.com/netsec-ethz/scion/issues/1334), so it won't be possible to receive queries or function behind nat until that is fixed.
</blockquote></details>

Okay. Out of interest, is this currently support by the python infrastructure?

---

*[go/cs/io.go, line 61 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTpglqeZInTKv8S-RQ:-KzYRwbvPJUM8j4KEmIa:brfjyw0) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/io.go#L61)):*

This isn't necessarily safe, as references to parts of the msg.Buf may be retained by the dispatched functions, especially if the message is ever passed off to a go routine for processing. I would probably keep it simple - have a large static buffer that you `ReadFromSCION` into, then allocate a new buffer for the size of data received, and pass that into dispatching, and let it be garbage-collected when it's done. While the CS does need to scale, it's not nearly as important to avoid allocations/copies as, say, the BR or SIG.
</blockquote></details>

Okay. The code for MsgBuffer is not needed anymore. Should I remove it?

---

*[go/cs/io.go, line 102 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTsaCM_sb5FTKwfqdj:-KzYSUqyeK97uYxgYbKk:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/io.go#L102)):*

You can simplify things by just doing `cpld.PackPld()`, and avoid having to allocate/free a buffer which may be the wrong size anyway.
</blockquote></details>

Done.

---

*[go/cs/req_cache.go, line 29 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTz3yg54p6HjZy6Eir:-KzYSsJNNc75n6pfp35H:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/req_cache.go#L29)):*

It looks like only writing locks are used, so this could just be `sync.Mutex`.
</blockquote></details>

Done.

---

*[go/cs/msg/buf.go, line 37 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTiitY1DB0v2aybIKn:-KzYTGC0f3yWnK7NODcr:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/msg/buf.go#L37)):*

It's a bit more elegant to do this instead:
```
func (p *BufPool) Get() *Buf {
    msgBuf := p.Pool.Get().(*Buf)
    ...
}
```
and
```
func (p *BufPool) Put(msgBuf *Buf) {
   p.Pool.Put(msgBuf)
}
```
That way you get something that's very close to the `sync.Pool` api, with type safety.
</blockquote></details>

Done.

---

*[go/cs/msg/buf.go, line 54 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTlCdzNTAyyStOrTRP:-KzYTMnVTs-YrG27kSBC:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/msg/buf.go#L54)):*

This probably doesn't need to be exported.
</blockquote></details>

Done.

---

*[go/cs/msg/buf.go, line 60 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTlL45DgevCXkpv4b4:-KzYTN8R0-Rj6QPZUg7W:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cs/msg/buf.go#L60)):*

Same for this re: exported.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-:-KzYm1iao0lcLwvmMx87:b-hn5cjd)*

kormat:
Reviewed 10 of 10 files at r4.

---

*[go/cert_srv/cert.go, line 107 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTxhM8gp3YlnFYV7nI:-KzcN8_zZakxITqL5ckK:b-azumu3) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cert_srv/cert.go#L107)):*

What key string?
</blockquote></details>

Oh, right, there isn't one because parsing failed :)

---

*[go/cert_srv/io.go, line 37 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTgTCVQQPiTehJ--ze:-KzcNKZhI71aNZOov_3G:bfdcdxl) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cert_srv/io.go#L37)):*

Okay. Out of interest, is this currently support by the python infrastructure?
</blockquote></details>

In theory, yes, though i don't know how fully it has been tested.

---

*[go/cert_srv/io.go, line 61 at r3](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzTpglqeZInTKv8S-RQ:-KzcOHg_YJU8V-J8nokp:b-loa1hg) ([raw file](https://github.com/netsec-ethz/scion/blob/5d731ec29cfd33f95b970c2a8110a1a115eb67fb/go/cert_srv/io.go#L61)):*

Okay. The code for MsgBuffer is not needed anymore. Should I remove it?
</blockquote></details>

Yeah, it can be dropped in that case.

---

*[go/cert_srv/main.go, line 49 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzcNcok6JzVY4FOWsSa:-KzcNcok6JzVY4FOWsSb:brpk32g) ([raw file](https://github.com/netsec-ethz/scion/blob/7531fc867634b3e4ffb9563a18b37b20367ddce2/go/cert_srv/main.go#L49)):*
> ```Go
> 	confDir  = flag.String("confd", "", "Configuration directory (Required)")
> 	cacheDir = flag.String("cached", "gen-cache", "Caching directory")
> 	prom     = flag.String("prom", "127.0.0.1:1281", "Address to export prometheus metrics on")
> ```

This is a bad default port, as it happens to be the exact one the SIG uses :) Maybe `1282`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1325)*

Oncilla:

---

*[go/cert_srv/main.go, line 49 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-KzcNcok6JzVY4FOWsSa:-Kzdc2lk_gbVnEAUFvbI:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/7531fc867634b3e4ffb9563a18b37b20367ddce2/go/cert_srv/main.go#L49)):*

This is a bad default port, as it happens to be the exact one the SIG uses :) Maybe `1282`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1325)*

kormat:
:lgtm:

---

Reviewed 1 of 10 files at r4, 1 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1325#-:-KzhvOuZIIbZn1LIhfbg:bnfp4nl)*

kormat:
Reviewed 1 of 10 files at r4, 1 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1325)*

PR #1335, opened 2017-11-21
Make go/proto build cleanly even after proto renaming.

kormat:
Currently if a `proto/*.capnp` file gets renamed, the generated
`go/proto/*.capnp.go` file doesn't get removed, and then compilation
fails as the same type is defined in multiple .go files. This change
allows make to detect and remove these stale files (and without using
shell, so `make -q` has the knowledge).

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1335)

### DIFF
--- a/go/proto/Makefile
+++ b/go/proto/Makefile
@@ -1,12 +1,18 @@
-.PHONY: all clean
+.PHONY: all stale clean
 
 PROTO_DIR = ../../proto
 SRCS := $(wildcard $(PROTO_DIR)/*.capnp)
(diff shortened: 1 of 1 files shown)


klausman:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1335#-:-KzUZaDiTWRh2Mftr5Vb:bnfp4nl)*

PR #1310, opened 2017-11-02
Wrap CtrlPld in a signable outer proto.

kormat:
This PR is the start of a series to overhaul how signing of protos is
handled in SCION.

- The signature is outside the CtrlPld proto, as we otherwise can't
  exclude it from the verification input.
- This approach marshals the CtrlPld to a []byte and signs that
  representation, as we can't depend on language/platform support for
  canonicalisation of capnp protos, especially with unknown fields (for
  forwards/backwards compatibility).

This will allow us to verify ctrl payload messages and senders in
transit. The signature code is in place, but not used/tested.

Note that this approach implies that if later verification of the
CtrlPld is desired, then the original raw representation must be
retained, as it is not possible for a receiver to reconstruct it
identically.

Also:
- Output cerr.Desc in border/gen.go error logs.
- Copy payloads in ScnPkt.Copy()

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1310)

Issue #803, opened 2016-07-05
Major renaming


pszal:
- PCB -> PSB (path segment beacon)
- ASMarking -> ASEntry
- PCBMarking -> HopEntry
- edge router -> border router
- end2end path -> forwarding path
- InfoOpaqueField -> InfoField
- _beaconing_ instead of _dissemination/propagation_
- _registration_ instead of _propagation_ (among PSes)

ref: https://github.com/netsec-ethz/scion-doc/issues/53

shitz:
`PathSegment`: The control-plane data-structure that holds a path segment
`PCB`: A packet that contains a `PathSegment` under construction.

kormat:
This renaming is happening in the go codebase, but we are not going to rename the python codebase, and will just use the new names as services/libraries are implemented in Go.

scrye:
Python servers are no longer under development and will soon be deprecated, so closing this.

PR #1336, opened 2017-11-23
Rework PathSegment signing

kormat:
This is a continuation of the work in #1310, extending that approach to
cover PathSegment signatures in a forwards/backwards
language/platform-agnostic way.

The python part of the change tries to minimize the churn by (mostly)
hiding the difference between PCB and PathSegment. As the IfID field is
no longer in PathSegment's proto, the beacon server still stores PCBs in
zookeeper.

Also:
- Separate out "PCB" and "PathSegment" concepts and protos. The PCB
  proto is used during beaconing and contains fields that are only used
  during that process. PathSegment is used everywhere else (path
  registration/replies).
- Renamed protos following #803.
- Add enum for segment type (up/down/core).
- build go libs that aren't currently used, to make code breakage
  obvious.
- Switch to 64-bit TRC and Cert versions (matching #1309)
- Fix a bug in scion.sh where python tests failing would stop go tests
  from being run.
- Remove SIBRA support from beacon service/generator, as it hasn't
  worked in a long time, and complicates things.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1336)


shitz:
Reviewed 15 of 34 files at r1.
Review status: 15 of 34 files reviewed at latest revision, 6 unresolved discussions.

---

*[go/lib/ctrl/seg/misc.go, line 1 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxMHBK-YPZODh51kQR:-KzxMHBK-YPZODh51kQS:be6ilhr) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/misc.go#L1)):*
> ```Go
> ```

Call this file `meta.go`?

---

*[go/lib/ctrl/seg/seg.go, line 69 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxI6aYWgssVDpuFbRm:-KzxI6aYWgssVDpuFbRn:b-oy2mle) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L69)):*
> ```Go
> }
> 
> func (ps *PathSegment) ID() (common.RawBytes, error) {
> ```

Not related to this PR, but can we cache the result of `ID()`? I know this is tricky for the `BS`, since it appends a new `ASEntry` which changes the `ID`, but in that case `AddASEntry` could update the internal `ID` variable. `ID()` is and will be used in many places that this optimization would make sense imo. (Can also be part of a different PR, but then I would file an issue if you agree)

---

*[go/lib/ctrl/seg/seg.go, line 101 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxKoFZAAhOm-es1elT:-KzxKoFZAAhOm-es1elU:b-m5oopg) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L101)):*
> ```Go
> }
> 
> func (ps *PathSegment) SignASEntry(key common.RawBytes) error {
> ```

Hmm, this always signs the last `ASEntry`. It should have at least a docstring saying so. Alternatively, rename this to `SignLastASEntry` or give this method an `idx` parameter.

---

*[go/lib/ctrl/seg/seg.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxLISsGTGWtGwgYvNa:-KzxLISsGTGWtGwgYvNb:b-soztzf) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L110)):*
> ```Go
> }
> 
> func (ps *PathSegment) VerifyASEntry(key common.RawBytes, idx int) error {
> ```

Future optimization: If a verifier verifies all ASEntries one after the other this method could take a parameter that already contains the sigpacked version up to the previous ASEntry and then only append the current packed ASEntry (would also need a change to not call sigpack here in that case). Maybe add a `NOTE` or `TODO` that we don't forget about this.

---

*[go/proto/sign.go, line 72 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxF03b8-BFIConBCF5:-KzxF03b8-BFIConBCF6:b-w6u15p) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/proto/sign.go#L72)):*
> ```Go
> func (s *SignS) Pack() common.RawBytes {
> 	var raw common.RawBytes
> 	raw = append(raw, common.RawBytes(s.Type.String())...)
> ```

Hmm, afaics this only uses the string representation of the enum because pycapnp represents enums as strings. Imo, this should use the integer type and have a string -> int compatibility method in python, since the Python code will be replaced over time anyway at which point it wouldn't make sense anymore to use the string representation here.

---

*[proto/path_seg.capnp, line 17 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxH6b4uc1Ly33n1fR8:-KzxH6b4uc1Ly33n1fR9:b-9vn28w) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/proto/path_seg.capnp#L17)):*
> ```
> 
> # Contains all top-level signed data for PathSegment
> struct PathSegmentSigned {
> ```

Not a big fan of this name. The name suggests that this is a signed path segment. What if we call the toplevel elements of the path segment the `PathSegmentHeader`? Alternatively, `PathSegmentMeta` would be applicable, but that name-clashes with the `PathSegMeta` that we already have.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

shitz:
Reviewed 19 of 34 files at r1.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[python/lib/crypto/asymcrypto.py, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxU-TYjX2gM-SbEq-_:-KzxU-TYjX2gM-SbEq-a:b-edp0pq) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/python/lib/crypto/asymcrypto.py#L87)):*
> ```Python
>     Generate Curve25519 keypair
> 
>     :returns tuple: A byte pair containing the public key and the private key.
> ```

"... used for encryption."

---

*[python/lib/crypto/certificate_chain.py, line 52 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxTj9f-GRwkLhMdE_q:-KzxTj9f-GRwkLhMdE_r:b-ylykhm) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/python/lib/crypto/certificate_chain.py#L52)):*
> ```Python
> 
> 
> def verify_sig_chain_trc(msg, sig, subject, chain, trc):
> ```

This method could now make use of `verify_chain_trc`.

---

*[python/lib/packet/pcb.py, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxP5lJeoA8rpecUI3N:-KzxP4w6Ko0SkFgqSZwK:bxeh84e) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/python/lib/packet/pcb.py#L115)):*
> ```Python
> 
> 
> class PathSegment(Cerealizable):
> ```

So just to make sure I understand correctly: In the Python code 

---

*[python/lib/packet/pcb.py, line 333 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxSeR-bzytMnN8JjOT:-KzxSeR-bzytMnN8JjOU:bq8iwkx) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/python/lib/packet/pcb.py#L333)):*
> ```Python
> 
> 
> class PCB(Cerealizable):
> ```

This should still contain `ifID` as a member even though it is also included in `PathSegment` (to minimize changes needed, correct?) for completeness.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

kormat:
Review status: 31 of 34 files reviewed at latest revision, 10 unresolved discussions.

---

*[go/lib/ctrl/seg/seg.go, line 69 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxI6aYWgssVDpuFbRm:-L-0vVNsQQ9P1-vC2Qil:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L69)):*

Not related to this PR, but can we cache the result of `ID()`? I know this is tricky for the `BS`, since it appends a new `ASEntry` which changes the `ID`, but in that case `AddASEntry` could update the internal `ID` variable. `ID()` is and will be used in many places that this optimization would make sense imo. (Can also be part of a different PR, but then I would file an issue if you agree)
</blockquote></details>

Done.

---

*[go/lib/ctrl/seg/seg.go, line 101 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxKoFZAAhOm-es1elT:-L-0vZfbjmo0h3Ew8T-X:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L101)):*

Hmm, this always signs the last `ASEntry`. It should have at least a docstring saying so. Alternatively, rename this to `SignLastASEntry` or give this method an `idx` parameter.
</blockquote></details>

Done.

---

*[go/lib/ctrl/seg/seg.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxLISsGTGWtGwgYvNa:-L-0viMc3HAEMeWGFbuW:b11ten9) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L110)):*

Future optimization: If a verifier verifies all ASEntries one after the other this method could take a parameter that already contains the sigpacked version up to the previous ASEntry and then only append the current packed ASEntry (would also need a change to not call sigpack here in that case). Maybe add a `NOTE` or `TODO` that we don't forget about this.
</blockquote></details>

This gets kinda messy, as the last ASEntry gets handled differently in `sigPack`. 

---

*[go/proto/sign.go, line 72 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxF03b8-BFIConBCF5:-L-0vytX7Y7cVWeKpCyO:bspvjgz) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/proto/sign.go#L72)):*
I had initially looked at using an integer, but unfortunately there is no documentation to say what size the enum int field _is_. It seems to be `uint16`, but also the [docs say](https://capnproto.org/language.html#enums):
> In languages where enums have numeric values, these numbers will be used, but in general Cap’n Proto enums should not be considered numeric.

---

*[proto/path_seg.capnp, line 17 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxH6b4uc1Ly33n1fR8:-L-0yBkaBKR-s1RD4Zzu:b-irda91) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/proto/path_seg.capnp#L17)):*
> Not a big fan of this name.

You could have mentioned when you looked at this in a previous PR :P

> The name suggests that this is a signed path segment.

I've tried to use a soft convention of `SignedX` being the signed version of `X` (c.f. `SignedCtrlPld`/`SignedBlob`).

> What if we call the toplevel elements of the path segment the PathSegmentHeader? 

I don't know what you're suggesting here. Do you mean rename `PathSegment` to `PathSegmentHeader`?

> Alternatively, PathSegmentMeta would be applicable, but that name-clashes with the PathSegMeta that we already have.

Yup.

So, i agree the naming is non-ideal, but i don't currently see a better idea.

---

*[python/lib/crypto/asymcrypto.py, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxU-TYjX2gM-SbEq-_:-L-0ym4-KYacgdvL_dpa:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/python/lib/crypto/asymcrypto.py#L87)):*

"... used for encryption."
</blockquote></details>

Done.

---

*[python/lib/crypto/certificate_chain.py, line 52 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxTj9f-GRwkLhMdE_q:-L-1-DS_OWlzVBMKCbwi:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/python/lib/crypto/certificate_chain.py#L52)):*

This method could now make use of `verify_chain_trc`.
</blockquote></details>

Done.

---

*[python/lib/packet/pcb.py, line 333 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxSeR-bzytMnN8JjOT:-L-1-Fvaxiy5yY_uG4Br:b-6btzww) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/python/lib/packet/pcb.py#L333)):*

This should still contain `ifID` as a member even though it is also included in `PathSegment` (to minimize changes needed, correct?) for completeness.
</blockquote></details>

Mostly. This must contain `ifID`, as it's the only proto that cerealises that field. It's also included in PathSegment to reduce churn.

---

*[go/lib/ctrl/seg/misc.go, line 1 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxMHBK-YPZODh51kQR:-L-0uNK56v5cJ8zgMGzQ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/misc.go#L1)):*

Call this file `meta.go`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

shitz:
Reviewed 2 of 6 files at r2.
Review status: 30 of 34 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/lib/ctrl/seg/seg.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxLISsGTGWtGwgYvNa:-L-16V7NCiV2OAyRkrgz:b31juwe) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L110)):*

This gets kinda messy, as the last ASEntry gets handled differently in `sigPack`. 
</blockquote></details>

I meant something like this (if a caller wants to verify the entire path segment in one call):
```
func (ps *PathSegment) Verify(key common.RawBytes) error {
	var data common.RawBytes
	for i := 0; i < ps.MaxAEIdx() - 1; i++ {
		data = append(data, ps.RawASEntries[i].Blob...)
		if err := ps.RawASEntries[i].Sign.Verify(key, data); err != nil {
			return err
		}
		data = append(data ps.RawASEntries[i].Sign.Pack()...)
	}
	idx := ps.MaxAEIdx() - 1
	data = append(data, ps.RawASEntries[idx].Blob...)
	return ps.RawASEntries[idx].Sign.Verify(key, data)
}
```

---

*[go/proto/sign.go, line 72 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxF03b8-BFIConBCF5:-L-13AHWIp_Eon6K_UVB:b-l9cn09) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/proto/sign.go#L72)):*

I had initially looked at using an integer, but unfortunately there is no documentation to say what size the enum int field _is_. It seems to be `uint16`, but also the [docs say](https://capnproto.org/language.html#enums):
> In languages where enums have numeric values, these numbers will be used, but in general Cap’n Proto enums should not be considered numeric.
</blockquote></details>

Still dislike having to use the string representation here, but I guess I don't care enough.

---

*[proto/path_seg.capnp, line 17 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxH6b4uc1Ly33n1fR8:-L-14SkZS2gTPKlczza1:bqt5t5p) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/proto/path_seg.capnp#L17)):*

> Not a big fan of this name.

You could have mentioned when you looked at this in a previous PR :P

> The name suggests that this is a signed path segment.

I've tried to use a soft convention of `SignedX` being the signed version of `X` (c.f. `SignedCtrlPld`/`SignedBlob`).

> What if we call the toplevel elements of the path segment the PathSegmentHeader? 

I don't know what you're suggesting here. Do you mean rename `PathSegment` to `PathSegmentHeader`?

> Alternatively, PathSegmentMeta would be applicable, but that name-clashes with the PathSegMeta that we already have.

Yup.

So, i agree the naming is non-ideal, but i don't currently see a better idea.

</blockquote></details>

No I mean renaming `PathSegmentSigned` to `PathSegmentHeader` since it contains the top level entries for the path segment.

---

*[python/lib/packet/pcb.py, line 115 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxP5lJeoA8rpecUI3N:-L-15tN6e9CFd7SAljbP:b-7x2vin) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/python/lib/packet/pcb.py#L115)):*

So just to make sure I understand correctly: In the Python code 
</blockquote></details>

This was an unfinished brainfart.

---

*[python/lib/packet/pcb.py, line 333 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxSeR-bzytMnN8JjOT:-L-164S04ODUlDXVejOF:b-b5sdqa) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/python/lib/packet/pcb.py#L333)):*

Mostly. This must contain `ifID`, as it's the only proto that cerealises that field. It's also included in PathSegment to reduce churn.
</blockquote></details>

What I meant is to have `ifID` on the `PCB` object, but I recall that in the python code we access simple proto members via `p`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

shitz:
Reviewed 4 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

kormat:

---

*[go/lib/ctrl/seg/seg.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxLISsGTGWtGwgYvNa:-L-1DzGm3wcOz0ISDb8t:brjanx) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L110)):*

I meant something like this (if a caller wants to verify the entire path segment in one call):
```
func (ps *PathSegment) Verify(key common.RawBytes) error {
	var data common.RawBytes
	for i := 0; i < ps.MaxAEIdx() - 1; i++ {
		data = append(data, ps.RawASEntries[i].Blob...)
		if err := ps.RawASEntries[i].Sign.Verify(key, data); err != nil {
			return err
		}
		data = append(data ps.RawASEntries[i].Sign.Pack()...)
	}
	idx := ps.MaxAEIdx() - 1
	data = append(data, ps.RawASEntries[idx].Blob...)
	return ps.RawASEntries[idx].Sign.Verify(key, data)
}
```
</blockquote></details>

That doesn't make sense - the key would be different for each ASEntry. Making that work would add a lot of complication to the caller.

---

*[proto/path_seg.capnp, line 17 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxH6b4uc1Ly33n1fR8:-L-1ENqHVkXOlZ1rb6Cu:b5shtle) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/proto/path_seg.capnp#L17)):*

No I mean renaming `PathSegmentSigned` to `PathSegmentHeader` since it contains the top level entries for the path segment.
</blockquote></details>

That's not true. The "logical" top-level entries are: infoF, asEntries, and potential future unsigned data like some extensions. "Header" implies that the data in there applies to the rest of the PathSegment, but the actual criteria for putting something into the current `PathSegmentSigned` is "should this be included in all signatures".

---

*[python/lib/packet/pcb.py, line 333 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxSeR-bzytMnN8JjOT:-L-1F517zQ_jiBmgOlOi:b-sskemy) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/python/lib/packet/pcb.py#L333)):*

What I meant is to have `ifID` on the `PCB` object, but I recall that in the python code we access simple proto members via `p`.
</blockquote></details>

Yes, we do. If a value is useful "raw", we access it directly via the `p` attribute. If it is not useful raw, then we either have getters, or set a top-level attribute at init time. In this case, ifID is perfectly useful raw, as there is no translation to be done. Otherwise we'd have to add a getter _and_ a setter.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

shitz:

---

*[go/lib/ctrl/seg/seg.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxLISsGTGWtGwgYvNa:-L-1VI8LLSeIUNNMCryp:b1nn0sv) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L110)):*

That doesn't make sense - the key would be different for each ASEntry. Making that work would add a lot of complication to the caller.
</blockquote></details>

Good point. Then this could take a slice of keys. I feel like a common pattern to verify a path segment would be:
1) Get all the ASes from the `ASEntries`.
2) Get/Request all keys for them from the trust store.
3) Call `pseg.Verify(keys)`.

---

*[proto/path_seg.capnp, line 17 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxH6b4uc1Ly33n1fR8:-L-1MziV7-gX_WTvgwue:b-3ckpsw) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/proto/path_seg.capnp#L17)):*

That's not true. The "logical" top-level entries are: infoF, asEntries, and potential future unsigned data like some extensions. "Header" implies that the data in there applies to the rest of the PathSegment, but the actual criteria for putting something into the current `PathSegmentSigned` is "should this be included in all signatures".
</blockquote></details>

Right. `PathSegmentSigned` still sounds like this is a singed path segment object. `PathSegmentSignedMeta`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

kormat:
Review status: 29 of 34 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/ctrl/seg/seg.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxLISsGTGWtGwgYvNa:-L-1mQ8RN2Ehy4ohlTol:b-ifuq4t) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L110)):*

Good point. Then this could take a slice of keys. I feel like a common pattern to verify a path segment would be:
1) Get all the ASes from the `ASEntries`.
2) Get/Request all keys for them from the trust store.
3) Call `pseg.Verify(keys)`.
</blockquote></details>

If you look at how the python code handles this, it separates these out a lot. When a path segment is received, the first thing is that 1. + 2. are done. If the trust store doesn't have all of the keys, then requests are sent, and the seg is put into limbo. This approach is basically required by the nature of our cert system, and won't change from python to go.

When all keys are available, then verification happens. Having to do 2 loops again for verification seems overly complex to me (1 loop to get all the ASes, 1 loop get to all the keys, finally call Verify() with all keys). A single loop to request the key for each ASEntry and then verify it makes for a single loop, and simpler logic, at the cost of being more inefficient.

---

*[proto/path_seg.capnp, line 17 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxH6b4uc1Ly33n1fR8:-L-1pVrJWzbTEJZw1y23:bcl4biz) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/proto/path_seg.capnp#L17)):*

Right. `PathSegmentSigned` still sounds like this is a singed path segment object. `PathSegmentSignedMeta`?
</blockquote></details>

Inspired by your fearless leadership, I've gone with `PathSegmentSignedData`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

shitz:
Reviewed 1 of 5 files at r3.
Review status: 30 of 34 files reviewed at latest revision, 3 unresolved discussions.

---

*[go/lib/ctrl/seg/seg.go, line 110 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxLISsGTGWtGwgYvNa:-L-1r3HyF44hUe9VZejJ:b-2nfwsw) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/go/lib/ctrl/seg/seg.go#L110)):*

If you look at how the python code handles this, it separates these out a lot. When a path segment is received, the first thing is that 1. + 2. are done. If the trust store doesn't have all of the keys, then requests are sent, and the seg is put into limbo. This approach is basically required by the nature of our cert system, and won't change from python to go.

When all keys are available, then verification happens. Having to do 2 loops again for verification seems overly complex to me (1 loop to get all the ASes, 1 loop get to all the keys, finally call Verify() with all keys). A single loop to request the key for each ASEntry and then verify it makes for a single loop, and simpler logic, at the cost of being more inefficient.
</blockquote></details>

I agree that it's simpler logic, I just fear that for a path server that has to do a lot of verification (think core path server of a big ISP) it will actually matter, however, we might revisit this when the time comes. Might also be premature optimization and turn out to not be a bottleneck.

---

*[go/lib/ctrl/seg/seg.go, line 34 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-L-1s4mNeITxT3E270a-:-L-1s4mNeITxT3E270a0:bndbvoj) ([raw file](https://github.com/netsec-ethz/scion/blob/e9f37d8520bd17b97aeeb6927e06c0b7491739ba/go/lib/ctrl/seg/seg.go#L34)):*
> ```Go
> type PathSegment struct {
> 	RawSigned    common.RawBytes      `capnp:"signed"`
> 	Signed       *PathSegmentSigned   `capnp:"-"`
> ```

This type should also be renamed and while you are at it maybe also rename the `signed` to `signedData` everywhere (code and proto).

---

*[proto/path_seg.capnp, line 17 at r1](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-KzxH6b4uc1Ly33n1fR8:-L-1rZrVN7aM64wno5Cd:btoh0qi) ([raw file](https://github.com/netsec-ethz/scion/blob/1ae9864b9acc27411fc7fa4f3936950b0f8999bc/proto/path_seg.capnp#L17)):*

Inspired by your fearless leadership, I've gone with `PathSegmentSignedData`
</blockquote></details>

Grand!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

kormat:
Review status: 29 of 34 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/lib/ctrl/seg/seg.go, line 34 at r2](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-L-1s4mNeITxT3E270a-:-L-1tRTDemNlFZrGHcuH:b-9dcgjj) ([raw file](https://github.com/netsec-ethz/scion/blob/e9f37d8520bd17b97aeeb6927e06c0b7491739ba/go/lib/ctrl/seg/seg.go#L34)):*

This type should also be renamed and while you are at it maybe also rename the `signed` to `signedData` everywhere (code and proto).
</blockquote></details>

The type _is_ renamed. Renamed the member too now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

shitz:
:lgtm:

---

Reviewed 1 of 3 files at r4.
Review status: 30 of 34 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/lib/ctrl/seg/seg.go, line 56 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-L-1uQES7XLqYD8Hsb5B:-L-1uQES7XLqYD8Hsb5C:b-bfh525) ([raw file](https://github.com/netsec-ethz/scion/blob/684a37589f5fbff62ef9eeaa502b6848d97ccd3a/go/lib/ctrl/seg/seg.go#L56)):*
> ```Go
> 		return nil, err
> 	}
> 	ps.SData, err = NewPathSegmentSignedFromRaw(ps.RawSData)
> ```

nit - this should be `NewPathSegmentSignedDataFromRaw`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-:-L-1uawR9H1s1myD85mX:bnfp4nl)*

kormat:
Review status: 30 of 34 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/ctrl/seg/seg.go, line 56 at r4](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-L-1uQES7XLqYD8Hsb5B:-L-1ux8_van65spw0MO-:b-896fix) ([raw file](https://github.com/netsec-ethz/scion/blob/684a37589f5fbff62ef9eeaa502b6848d97ccd3a/go/lib/ctrl/seg/seg.go#L56)):*

nit - this should be `NewPathSegmentSignedDataFromRaw`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

shitz:
:lgtm:

---

Reviewed 1 of 2 files at r5.
Review status: 30 of 34 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336#-:-L-1vAFHAeMxV4ThkexH:bnfp4nl)*

shitz:
Reviewed 1 of 5 files at r3, 1 of 3 files at r4, 1 of 2 files at r5, 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/scion/1336)*

PR #1337, opened 2017-11-24
Unit tests for pathmgr/cache.go and rpkt/rpkt.go and various others

andviane:
The cache implementation pathmgr/cache.go currently has 10 methods without obvious test coverage. The proposed change adds testing suite for 7 methods. The 3 remaining methods are explicitly marked as "internal". 

rpkt seems the central unit also without much coverage. I propose two test suites, one based on parsing the real world logged package from the binary representation and another based on human readable constants rather than the embedded binary dump. At the time of making this request it is not obvious which approach does the community prefers here, both help against regressions.

I also added tests for UDP structure, not much code there but better to have it covered. 

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1337)


scrye:
Thank you for you contribution! I did a pass through the code and included some remarks below. If you have any questions, please let me know.

Also, please rebase on master, as some Go paths have changed since the last version (the repository has moved to https://github.com/scionproto/scion).

Regarding empty lines, we tend to use fewer in the codebase. Notably:
  - testing assertions that are related to the same call should be placed in a compact block;
  - structure initializations should not include newlines.

Also, try to include as much code as possible in the `Convey` blocks. When the testing framework runs, it explores each branch of nested `Convey` blocks starting from the topmost one. This means that state external to the outer one gets shared between tests that should be independent. This doesn't affect tests that only read from that state, but it can hide traps.

---

Review status: 0 of 4 files reviewed at latest revision, 18 unresolved discussions, some commit checks failed.

---

*[go/border/rpkt/rpkt_hook_test.go, line 64 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-gwm4fXqsdteBJBV1y:-L-gwm4fXqsdteBJBV1z:bs1fj9i) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/border/rpkt/rpkt_hook_test.go#L64)):*
> ```Go
> 
> 		ia, err = rpkt.DstIA()
> 		So(ia, ShouldEqual, dstIA)
> ```

Calling `SoMsg` (defined in the fork of the `goconvey` library we're using) can be used here to determine which of the `So` checks failed. For an example of `SoMsg`, please see `go/lib/snet/integration_test.go`. All `So` calls should be replaced with `SoMsg` calls.

---

*[go/border/rpkt/rpkt_hook_test.go, line 146 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-gzwNRlWtRs3PMJzWn:-L-gzwNRlWtRs3PMJzWo:b-dni79z) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/border/rpkt/rpkt_hook_test.go#L146)):*
> ```Go
> 		So(info.ISD, ShouldEqual, infof.ISD)
> 		So(info.TsInt, ShouldEqual, infof.TsInt)
> 		So(info.Hops, ShouldEqual, infof.Hops)
> ```

A `ShouldResemble` here would achieve the same effect as these three statements.

---

*[go/border/rpkt/rpkt_hook_test.go, line 153 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-h-GvcZBwMDuQRoyww:-L-h-GvcZBwMDuQRoywx:b-dni79z) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/border/rpkt/rpkt_hook_test.go#L153)):*
> ```Go
> 		So(info.ISD, ShouldEqual, infof.ISD)
> 		So(info.TsInt, ShouldEqual, infof.TsInt)
> 		So(info.Hops, ShouldEqual, infof.Hops)
> ```

A `ShouldResemble` here would achieve the same effect as these three statements.

---

*[go/border/rpkt/rpkt_test.go, line 51 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-h0RqWebciVy8NA7sm:-L-h0RqWebciVy8NA7sn:b-cpcnoj) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/border/rpkt/rpkt_test.go#L51)):*
> ```Go
> 	var config = &conf.Conf{
> 		IA: &addr.ISD_AS{I: 1, A: 2},
> 		Net: &netconf.NetConf{IFs: map[common.IFIDType]*netconf.Interface{
> ```

Begin a new line starting with `IFs` here.

---

*[go/border/rpkt/rpkt_test.go, line 73 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-h0blGvDrHOeym-TWX:-L-h0blGvDrHOeym-TWY:b7suqa1) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/border/rpkt/rpkt_test.go#L73)):*
> ```Go
> 		dstHost, _ := r.DstHost()
> 
> 		So(srcIA.String(), ShouldEqual, addr.ISD_AS{I: 1, A: 14}.String())
> ```

`ShouldResemble` would eliminate the need to call `String()` here. If one being a pointer and one a value is a problem, reference/dereference one of the operands.

---

*[go/border/rpkt/rpkt_test.go, line 77 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-h1s1u1OZyFurhQyAj:-L-h1s1u1OZyFurhQyAk:b-3fmy2r) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/border/rpkt/rpkt_test.go#L77)):*
> ```Go
> 
> 		So(srcHost.String(), ShouldEqual,
> 			addr.HostFromIP(net.IPv4(127, 0, 0, 26)).String())
> ```

`ShouldResemble` can be used here to compare directly between IPs.

---

*[go/border/rpkt/rpkt_test.go, line 98 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-h21pn9pIw5CHFipDb:-L-h21pn9pIw5CHFipDc:bby80sb) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/border/rpkt/rpkt_test.go#L98)):*
> ```Go
> 		// Verify additional fields
> 		l4hdr, _ := r.L4Hdr(false)
> 		So(l4hdr.String(), ShouldEqual,
> ```

`ShouldResemble` can be used here to prevent updates to `String()` from breaking the test:

```
SoMsg("udp", l4hdr, ShouldResemble, &l4.UDP{
    SrcPort: 50000, 
    DstPort: 50000,
    TotalLen: 39,
    Checksum: common.RawBytes{0x0a, 0x16},
}
```

---

*[go/lib/l4/udp_test.go, line 32 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-gaWDwa6g2ba5dX9f5:-L-gaWDwa6g2ba5dX9f6:b-t153gm) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/l4/udp_test.go#L32)):*
> ```Go
> }
> 
> func Test_UDPFromRaw(t *testing.T) {
> ```

We generally use pure camelcase here, without `_`.

---

*[go/lib/l4/udp_test.go, line 34 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-gYJv9p4BnwPPIGo5c:-L-gR14ezZKYi3XLqJiN:br7032j) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/l4/udp_test.go#L34)):*
> ```Go
> func Test_UDPFromRaw(t *testing.T) {
> 	raw := common.RawBytes{
> 		0x12, 0x34, 0x56, 0x78, 0, 0xA, 0, 0}
> ```

We tend to line wrap much more sparingly in the code base:
  - we always wrap lines longer than 100 characters;
  - we wrap below 100 characters only if it significantly improves readability (e.g., when initializing complex structures with lots of members).

In this case (and some other initializations in this file) line wrapping s is not needed.

---

*[go/lib/l4/udp_test.go, line 35 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-gRQS8yPjC0_YME2YV:-L-gRQS8yPjC0_YME2YW:bagvona) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/l4/udp_test.go#L35)):*
> ```Go
> 	raw := common.RawBytes{
> 		0x12, 0x34, 0x56, 0x78, 0, 0xA, 0, 0}
> 	fromRaw, err := UDPFromRaw(raw)
> ```

We generally include the tested function call inside the `Convey` block (right before the `So` checks). Here it is not necessarily a problem, but it is especially helpful in larger `Convey` trees, as state doesn't get shared between `Convey` branch explorations. For more examples, please check the official `goconvey` documentation: https://github.com/smartystreets/goconvey.

The same observation for the rest of the functions in this file.

For example, moving `x` outside the topmost `Convey` block causes the second test to fail:

```
Convey("Base", t, func() {
    x := 1
    Convey("Inc #1", func() {
        x++
        So(x, ShouldEqual, 2)
    })
    Convey("Inc #2 starts from the original x again", func() {
        x++
        So(x, ShouldEqual, 2)
    })
})
```

---

*[go/lib/l4/udp_test.go, line 45 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-gcGu55r0J6zyYAoCE:-L-gcGu55r0J6zyYAoCF:b-y0j5se) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/l4/udp_test.go#L45)):*
> ```Go
> }
> 
> func Test_Validate(t *testing.T) {
> ```

Since in this package multiple types might expose the same methods in the future, please include the type in function names (e.g., `TestUDPValidate`).

---

*[go/lib/l4/udp_test.go, line 76 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-gUTzUkjYy-YUvd7km:-L-gUTzV6aTjuC7_GIW1:b96i6hr) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/l4/udp_test.go#L76)):*
> ```Go
> 	Convey("Binary content must match", t, func() {
> 		So(err, ShouldBeNil)
> 		So(raw.String(), ShouldEqual, "12345678000a0000")
> ```

This is an immediate consequence of the check below, so it can be skipped.

---

*[go/lib/l4/udp_test.go, line 94 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-gd3SKzo2UJ5IWNF2J:-L-gd3SKzo2UJ5IWNF2K:b-u2sa5u) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/l4/udp_test.go#L94)):*
> ```Go
> }
> 
> func Test_Checksum(t *testing.T) {
> ```

Getter/setter tests like this one will probably never fail (as tested functionality is very basic and they are very tightly coupled with the implementation). We tend to not include such tests at all to keep the code base smaller. Coverage can be improved in this case through better integration testing.

---

*[go/lib/l4/udp_test.go, line 104 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-ghbGCrSq-n56BZllo:-L-ghbGDqVgFkff92hMw:bkrnhav) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/l4/udp_test.go#L104)):*
> ```Go
> }
> 
> func Test_PldLen(t *testing.T) {
> ```

Same as `Test_Checksum`.

---

*[go/lib/l4/udp_test.go, line 113 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-ghcU7TCXsYC3Cb6yi:-L-ghcU8yv32GZxz1BO7:bkrnhav) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/l4/udp_test.go#L113)):*
> ```Go
> }
> 
> func Test_Copy(t *testing.T) {
> ```

Same as `Test_Checksum`.

---

*[go/lib/l4/udp_test.go, line 122 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-ghkLi58tcnmCjeuYM:-L-ghkLi58tcnmCjeuYN:bkrnhav) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/l4/udp_test.go#L122)):*
> ```Go
> }
> 
> func Test_Reverse(t *testing.T) {
> ```

Same as `Test_Checksum`.

---

*[go/lib/l4/udp_test.go, line 135 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-ghmsGUixhv8dpDf7H:-L-ghmsGUixhv8dpDf7I:bkrnhav) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/l4/udp_test.go#L135)):*
> ```Go
> }
> 
> func Test_String(t *testing.T) {
> ```

Same as `Test_Checksum`.

---

*[go/lib/pathmgr/cache_test.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1337#-L-glnlyMaDvYIn0rwUQ:-L-glnlz7Ar6HYfgfL14:b-llq1y6) ([raw file](https://github.com/scionproto/scion/blob/00f5f5fca32fba30705e5f2059955ce621d15f94/go/lib/pathmgr/cache_test.go#L1)):*
> ```Go
> // Copyright 2017 Audrius Meskauskas with all possible permissions granted
> ```

We decided against having unit tests for the cache itself because it is part of package internals, only used by the API in `pathmgr.go`. In general, we do not test internals.

The API tests in `pathmgr_test.go` already test the functionality of the cache itself (76% coverage). Testing the cache on its own adds a lot of code to the codebase (code that needs to be maintained), without adding significant new information.

The only function missing from that coverage is `removeWatch`, but it should be included by writing an API test and calling the public method `Unwatch`. Doing that would be a really useful contribution here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1337#-:-L-gblIfIH_M2HY9fr3H:bininq9)*

andviane:
I agree with all proposals apart from the complete removal of the test suite for the cache. Did you consider that having separate tests for this class allows to work on it as on the independent unit? I planned more work on this class (to implement the expiration of old entries), and wanted an easy proof that the cache unit still behaves correctly after refactoring. This is not the same as testing arbitrary internals. I will fix other issues and remove the cache test from this pull request so that it could be discussed separately.

andviane:
I have applied (I hope) all review findings:
* _SoMsg_ is now used everywhere with the reason explaining the test failure
* The number of blank lines is now much reduced.
* Basically everything is now moved into _Convey_
* ShouldResemble is consistently used everywhere.
* Underscores are removed from names. 
* Cache test is removed from this pull request.

I hope I addressed all issues unless overlooked something by chance.

andviane:
I am sorry I needed to re-fork repository and re-apply changes in order to recover from the repository rename. Closing this pull request, opening another. I hope it should be visible that all changes are applied.

PR #1356, opened 2017-12-08
Add send/recv timeouts for the dispatcher on app sockets.

kormat:
The current dispatcher design will deadlock if an application is failing
to read from its socket and that socket's buffer fills up.
https://github.com/scionproto/scion/issues/1278 is the relevant issue.

This patch adds a 2s read/write timeout for app sockets, after which the
dispatcher will log an error and close the socket. This doesn't fully
fix the issue, but it does mean the dispatcher can recover, and there
will be clear logging to ease debugging, while hopefully avoiding any
packet loss during normal operation.

Also:
- Fix a memory leak in parse_request.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1356)

### DIFF
--- a/c/dispatcher/dispatcher.c
+++ b/c/dispatcher/dispatcher.c
@@ -152,6 +152,7 @@ void handle_send(int index);
 void cleanup_socket(int sock, int index, int err);
 
 int send_data(uint8_t *buf, int len, HostAddr *first_hop);
+int deliver_data(int sock, HostAddr *from, uint8_t *buf, int len);
 
 #ifndef UNIX_PATH_MAX
 #define UNIX_PATH_MAX 108
@@ -463,6 +464,23 @@ void handle_app()
         return;
     }
     zlog_info(zc, "new socket created: %d", sock);
+    // Set send/recv timeouts on socket, so that an unresponsive application
+    // doesn't block the dispatcher indefinitely.
+    struct timeval timeout;
+    timeout.tv_sec = 2;
+    timeout.tv_usec = 0;
+    errno = 0;
+    if (setsockopt(sock, SOL_SOCKET, SO_SNDTIMEO, (char *)&timeout, sizeof(timeout)) < 0) {
+            zlog_error(zc, "failed to set send timeout on app socket (fd: %d): %s", sock, strerror(errno));
+            close(sock);
+            return;
+    }
+    errno = 0;
+    if (setsockopt(sock, SOL_SOCKET, SO_RCVTIMEO, (char *)&timeout, sizeof(timeout)) < 0) {
+            zlog_error(zc, "failed to set recv timeout on app socket (fd: %d): %s", sock, strerror(errno));
+            close(sock);
+            return;
+    }
     /*
      * Application message format:
      * cookie (8B) | addr_type (1B) | packet_len (4B) | addr (?B) | port (2B) | msg (?B)
@@ -552,6 +570,7 @@ Entry * parse_request(uint8_t *buf, int len, int proto, int sock)
     if (type < ADDR_IPV4_TYPE || type > ADDR_IPV6_TYPE) {
         zlog_error(zc, "Invalid address type: %d", type);
         close(sock);
+        free(e);
         return NULL;
     }
 
@@ -684,7 +703,14 @@ void reply(int sock, int port)
     uint8_t buf[DP_HEADER_LEN + 2];
     write_dp_header(buf, NULL, 2);
     *(uint16_t *)(buf + DP_HEADER_LEN) = htons(port);
-    send_all(sock, buf, sizeof(buf));
+    int len = sizeof(buf);
+    int sent = send_all(sock, buf, len);
+    if (sent != len) {
+        zlog_warn(zc, "Failed to send registration reply on fd %d, expected:%dB got:%dB (errno: %s)",
+                sock, len, sent, strerror(errno));
+        close(sock);
+        return;
+    }
     zlog_debug(zc, "sent reply %d on socket %d", port, sock);
 }
 
@@ -823,7 +849,6 @@ void deliver_udp(uint8_t *buf, int len, HostAddr *from, HostAddr *dst)
     uint8_t *l4ptr = buf;
     get_l4_proto(&l4ptr);
     SCIONUDPHeader *udp = (SCIONUDPHeader *)l4ptr;
-    int sock;
 
     uint16_t checksum = scion_udp_checksum(buf, chk_udp_input);
     if (checksum != udp->checksum) {
@@ -855,13 +880,10 @@ void deliver_udp(uint8_t *buf, int len, HostAddr *from, HostAddr *dst)
             return;
         }
     }
-    sock = e->sock;
-    send_dp_header(sock, from, len);
-    send_all(sock, buf, len);
+    deliver_data(e->sock, from, buf, len);
 }
 
 void deliver_udp_svc(uint8_t *buf, int len, HostAddr *from, HostAddr *dst) {
-    int sock;
     SVCKey svc_key;
     memset(&svc_key, 0, sizeof(SVCKey));
     uint16_t addr = ntohs(*(uint16_t *)get_dst_addr(buf));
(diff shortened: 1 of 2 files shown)


scrye:
Reviewed 2 of 2 files at r1.

---

*[c/dispatcher/dispatcher.c, line 472 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qC2dfx8PSr375FVot:-L-qC2dgxa_X_62argQk:benyjit) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L472)):*
> ```C
>     timeout.tv_sec = 2;
>     timeout.tv_usec = 0;
>     errno = 0;
> ```

You don't need to set `errno`, functions that fail will set it anyway.

---

*[c/dispatcher/dispatcher.c, line 478 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qC6dvVlMGWW7jHvqI:-L-qC6dvVlMGWW7jHvqJ:b2nreh6) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L478)):*
> ```C
>             return;
>     }
>     errno = 0;
> ```

Same as above.

---

*[c/dispatcher/dispatcher.c, line 479 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qCgpVtSkbQrBZPvH4:-L-qCgpVtSkbQrBZPvH5:bdz513p) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L479)):*
> ```C
>     }
>     errno = 0;
>     if (setsockopt(sock, SOL_SOCKET, SO_RCVTIMEO, (char *)&timeout, sizeof(timeout)) < 0) {
> ```

This actually solves a really nasty issue where a truncated packet would cause the dispatcher to hang.

---

*[c/dispatcher/dispatcher.c, line 1200 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qEq9xHL-iK4mu08T1:-L-qEq9xHL-iK4mu08T2:benyjit) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L1200)):*
> ```C
> int deliver_data(int sock, HostAddr *from, uint8_t *buf, int len)
> {
>     errno = 0;
> ```

You don't need to set `errno`, functions that fail will set it anyway.

---

*[c/dispatcher/dispatcher.c, line 1202 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qF0sekuvT7Cb_8F9w:-L-qF0sfsyBVvTldPfEi:by6o30b) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L1202)):*
> ```C
>     errno = 0;
>     if (send_dp_header(sock, from, len) != 0) {
>         zlog_warn(zc, "Failed to send dp header to app on fd %d (errno: %s)", sock, strerror(errno));
> ```

`(errno: %s)` should be rephrased, as it's not printing the error number.

---

*[c/dispatcher/dispatcher.c, line 1206 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qEuJZRy18N64yUL_S:-L-qEuJZRy18N64yUL_T:benyjit) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L1206)):*
> ```C
>         return -1;
>     }
>     errno = 0;
> ```

You don't need to set `errno`, functions that fail will set it anyway.

---

*[c/dispatcher/dispatcher.c, line 1209 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qFEcbomi1XEb1gNLL:-L-qFEcbomi1XEb1gNLM:by6o30b) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L1209)):*
> ```C
>     int sent = send_all(sock, buf, len);
>     if (sent != len) {
>         zlog_warn(zc, "Failed to send all data to app on fd %d, expected:%dB got:%dB (errno: %s)",
> ```

`(errno: %s)` should be rephrased, as it's not printing the error number.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1356)*

kormat:

---

*[c/dispatcher/dispatcher.c, line 472 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qC2dfx8PSr375FVot:-L04KZTlQk_Eeecj-hbg:bfkocc5) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L472)):*

You don't need to set `errno`, functions that fail will set it anyway.
</blockquote></details>

It's not quite that simple. See https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152351 for some background. I've ran into the problem too many times where a stale non-zero errno caused a program (or humans, reading the logs) to believe an error had occurred. So as a matter of principle i try to set it to 0 before a call if it's going to be used in the error handling/reporting after the call. That way it is unambiguous where/when the errno was set, and there can't be confusion about stale values.

---

*[c/dispatcher/dispatcher.c, line 1200 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qEq9xHL-iK4mu08T1:-L04LOXeKv2pI9XsqH4y:b7gevo4) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L1200)):*

You don't need to set `errno`, functions that fail will set it anyway.
</blockquote></details>

That's not guaranteed, `send_dp_header` isn't a syscall. It's possible for it to fail due to a logic error, without any syscall failing, in which case a stale non-zero errno value would give a completely bogus reason in the logs.

---

*[c/dispatcher/dispatcher.c, line 1202 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qF0sekuvT7Cb_8F9w:-L04M-T1TZI5eEGu3-LO:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L1202)):*

`(errno: %s)` should be rephrased, as it's not printing the error number.
</blockquote></details>

Done.

---

*[c/dispatcher/dispatcher.c, line 1209 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qFEcbomi1XEb1gNLL:-L04M05g4-AZSwRIl8c3:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L1209)):*

`(errno: %s)` should be rephrased, as it's not printing the error number.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1356)*

scrye:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---

*[c/dispatcher/dispatcher.c, line 472 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qC2dfx8PSr375FVot:-L04mLesEK38yixZMxpg:b-ts3rdy) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L472)):*

It's not quite that simple. See https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152351 for some background. I've ran into the problem too many times where a stale non-zero errno caused a program (or humans, reading the logs) to believe an error had occurred. So as a matter of principle i try to set it to 0 before a call if it's going to be used in the error handling/reporting after the call. That way it is unambiguous where/when the errno was set, and there can't be confusion about stale values.
</blockquote></details>

The link refers to C standard library calls; those are generally more complex, and do fancier stuff with `errno`.

Nearly all syscall wrappers from the Linux API (like the one below) have the same behavior, returning `<0` on errors (or a `-1` pointer) and setting `errno` (as specified in their man pages). In this case setting `errno` beforehand is not needed. For examples, see `coreutils` sources: http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/.

I find `errno` resets a bit confusing, because it hints to me that the program is trying to do something unnatural with it (e.g., using it during decision-making).

In any case, we can leave it as is.

---

*[c/dispatcher/dispatcher.c, line 1200 at r1](https://reviewable.io:443/reviews/scionproto/scion/1356#-L-qEq9xHL-iK4mu08T1:-L04pUGH9TmIr1SD323r:b-y3etdw) ([raw file](https://github.com/scionproto/scion/blob/ccfd7f6ccf703233b323c51e303f333d7648b066/c/dispatcher/dispatcher.c#L1200)):*

That's not guaranteed, `send_dp_header` isn't a syscall. It's possible for it to fail due to a logic error, without any syscall failing, in which case a stale non-zero errno value would give a completely bogus reason in the logs.
</blockquote></details>

Hmm, then those logic errors cannot be told apart based on the logging message below.

I'm fine with leaving them as is; the dispatcher's error handling is quite flaky in multiple places and spending time on it isn't a priority.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1356#-:-L04qYBFIw-r4nA9QfZ2:bnfp4nl)*

PR #1359, opened 2017-12-11
UDP with ACKs for SCION infra

scrye:
Implements a protocol for messages with ACKs on top of a net.PacketConn object.

Also:
  - A dummy loopback connection object for simplified testing;
  - An elastic free buffer pool that multiple goroutines can share (when batching is not a performance concern yet).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1359)

### DIFF
--- a/go/lib/infra/messaging/messaging.go
+++ b/go/lib/infra/messaging/messaging.go
@@ -1,0 +1,39 @@
+// Copyright 2017 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package messaging
+
+import (
+	"context"
+	"net"
+
+	"github.com/scionproto/scion/go/lib/common"
+)
+
+var (
+	ErrClosed      = common.NewCError("Messaging transport closed")
+	ErrContextDone = common.NewCError("Context expired while waiting")
+)
+
+// Transport layers must be safe for concurrent use by multiple goroutines.
+type Transport interface {
+	// Send a message without expecting an ACK.
+	SendUnreliableMsgTo(context.Context, common.RawBytes, net.Addr) error
+	// Send a message, waiting for it to be ACK'd.
+	SendMsgTo(context.Context, common.RawBytes, net.Addr) error
+	// Receive a message.
+	RecvFrom(context.Context) (common.RawBytes, net.Addr, error)
+	// Clean up.
+	Close(context.Context) error
+}
--- a/go/lib/infra/messaging/messaging.go
+++ b/go/lib/infra/messaging/messaging.go
@@ -28,9 +28,12 @@ var (
 
 // Transport layers must be safe for concurrent use by multiple goroutines.
 type Transport interface {
-	// Send a message without expecting an ACK.
+	// Send an unreliable message. Unreliable transport layers do not request
+	// an ACK. For reliable transport layers, this is the same as SendMsgTo.
 	SendUnreliableMsgTo(context.Context, common.RawBytes, net.Addr) error
-	// Send a message, waiting for it to be ACK'd.
+	// Send a reliable message. Unreliable transport layers block here waiting
+	// for the message to be ACK'd. Reliable transport layers return
+	// immediately.
 	SendMsgTo(context.Context, common.RawBytes, net.Addr) error
 	// Receive a message.
 	RecvFrom(context.Context) (common.RawBytes, net.Addr, error)
--- a/go/lib/infra/messaging/types.go
+++ b/go/lib/infra/messaging/types.go
@@ -22,7 +22,7 @@ import (
 )
 
 // ackTable maps packet IDs to channels. Goroutines waiting for an ACK for a
-// session ID create a channel, store it in the map at that ID and wait for the
+// packet ID create a channel, store it in the map at that ID and wait for the
 // channel to be closed. The background receiving goroutine closes the channel
 // when it receives the corresponding ack.
 type ackTable sync.Map
@@ -64,7 +64,7 @@ type uint56 uint64
 func (u *uint56) Inc() uint56 {
 	for {
 		old := atomic.LoadUint64((*uint64)(u))
-		new := (old + 1) % (1<<56 - 1)
+		new := (old + 1) % (1 << 56)
 		swapped := atomic.CompareAndSwapUint64((*uint64)(u), old, new)
 		if swapped {
 			return uint56(new)
@@ -72,19 +72,21 @@ func (u *uint56) Inc() uint56 {
 	}
 }
 
+// putUint56 writes a to b in network byte order.
 func (a uint56) putUint56(b common.RawBytes) {
 	_ = b[6]
-	b[0] = byte(a)
-	b[1] = byte(a >> 8)
-	b[2] = byte(a >> 16)
+	b[0] = byte(a >> 48)
+	b[1] = byte(a >> 40)
+	b[2] = byte(a >> 32)
 	b[3] = byte(a >> 24)
-	b[4] = byte(a >> 32)
-	b[5] = byte(a >> 40)
-	b[6] = byte(a >> 48)
+	b[4] = byte(a >> 16)
+	b[5] = byte(a >> 8)
+	b[6] = byte(a)
 }
 
+// getUint56 returns the number in b, read in network byte order.
 func getUint56(b common.RawBytes) uint56 {
 	_ = b[6]
-	return uint56(b[0]) | uint56(b[1])<<8 | uint56(b[2])<<16 | uint56(b[3])<<24 |
-		uint56(b[4])<<32 | uint56(b[5])<<40 | uint56(b[6])<<48
+	return uint56(b[0])<<48 | uint56(b[1])<<40 | uint56(b[2])<<32 | uint56(b[3])<<24 |
+		uint56(b[4])<<16 | uint56(b[5])<<8 | uint56(b[6])
 }
--- a/go/lib/infra/messaging/types.go
+++ b/go/lib/infra/messaging/types.go
@@ -60,6 +60,10 @@ func (m *ackTable) Store(key uint56, value chan struct{}) {
 // Supports atomic increments and wraps on 7 bytes.
 type uint56 uint64
 
+const (
+	maxUint56 = (1 << 56) - 1
+)
+
 // Inc atomically increments u, and returns the new value.
 func (u *uint56) Inc() uint56 {
 	for {
--- a/go/lib/infra/messaging/types.go
+++ b/go/lib/infra/messaging/types.go
@@ -1,0 +1,90 @@
+// Copyright 2017 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package messaging
+
+import (
+	"sync"
+	"sync/atomic"
+
+	"github.com/scionproto/scion/go/lib/common"
+)
+
+// ackTable maps packet IDs to channels. Goroutines waiting for an ACK for a
+// session ID create a channel, store it in the map at that ID and wait for the
+// channel to be closed. The background receiving goroutine closes the channel
+// when it receives the corresponding ack.
+type ackTable sync.Map
+
+func (m *ackTable) Delete(key uint56) {
+	(*sync.Map)(m).Delete(key)
+}
+
+func (m *ackTable) Load(key uint56) (chan struct{}, bool) {
+	value, loaded := (*sync.Map)(m).Load(key)
+	if value == nil {
+		return nil, loaded
+	}
+	return value.(chan struct{}), loaded
+}
+
+func (m *ackTable) LoadOrStore(key uint56, value chan struct{}) (chan struct{}, bool) {
+	newValue, loaded := (*sync.Map)(m).LoadOrStore(key, value)
+	if newValue == nil {
+		return nil, loaded
+	}
+	return newValue.(chan struct{}), loaded
+}
+
+func (m *ackTable) Range(f func(uint56, chan struct{}) bool) {
+	(*sync.Map)(m).Range(func(k, v interface{}) bool {
+		return f(k.(uint56), v.(chan struct{}))
+	})
+}
+
+func (m *ackTable) Store(key uint56, value chan struct{}) {
+	(*sync.Map)(m).Store(key, value)
+}
+
+// Supports atomic increments and wraps on 7 bytes.
+type uint56 uint64
+
+// Inc atomically increments u, and returns the new value.
+func (u *uint56) Inc() uint56 {
+	for {
+		old := atomic.LoadUint64((*uint64)(u))
+		new := (old + 1) % (1<<56 - 1)
+		swapped := atomic.CompareAndSwapUint64((*uint64)(u), old, new)
+		if swapped {
+			return uint56(new)
+		}
+	}
+}
+
+func (a uint56) putUint56(b common.RawBytes) {
+	_ = b[6]
+	b[0] = byte(a)
+	b[1] = byte(a >> 8)
+	b[2] = byte(a >> 16)
+	b[3] = byte(a >> 24)
+	b[4] = byte(a >> 32)
+	b[5] = byte(a >> 40)
+	b[6] = byte(a >> 48)
+}
+
+func getUint56(b common.RawBytes) uint56 {
+	_ = b[6]
+	return uint56(b[0]) | uint56(b[1])<<8 | uint56(b[2])<<16 | uint56(b[3])<<24 |
+		uint56(b[4])<<32 | uint56(b[5])<<40 | uint56(b[6])<<48
+}
--- a/go/lib/infra/messaging/udp.go
+++ b/go/lib/infra/messaging/udp.go
@@ -60,6 +60,11 @@ const (
 	maxReadEvents = 1 << 8
 )
 
+var (
+	// Used to initialize the first packet ID
+	generator = rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
+)
+
 var _ Transport = (*RUDP)(nil)
 
 // RUDP (Reliable UDP) implements a simple UDP protocol with ACKs on top of SCION/UDP.
@@ -107,7 +112,7 @@ type RUDP struct {
 func NewRUDP(conn net.PacketConn, logger log.Logger) *RUDP {
 	t := &RUDP{
 		conn:           conn,
-		nextPktID:      uint56(rand.Intn(maxUint56 + 1)),
+		nextPktID:      uint56(generator.Intn(maxUint56 + 1)),
 		readEvents:     make(chan *readEventDesc, maxReadEvents),
 		closed:         make(chan struct{}),
 		backgroundDone: make(chan struct{}),
--- a/go/lib/infra/messaging/udp.go
+++ b/go/lib/infra/messaging/udp.go
@@ -103,7 +103,7 @@ type RUDP struct {
 //
 // NewRUDP also spawns a background receiving goroutine that continuously reads
 // from conn and keeps track of ACKs and messages.
-func NewUDPTransport(conn net.PacketConn, logger log.Logger) *RUDP {
+func NewRUDP(conn net.PacketConn, logger log.Logger) *RUDP {
 	t := &RUDP{
 		conn:           conn,
 		readEvents:     make(chan *readEventDesc, maxReadEvents),
@@ -118,8 +118,12 @@ func NewUDPTransport(conn net.PacketConn, logger log.Logger) *RUDP {
 // SendUnreliableMsgTo sends a message and returns without waiting for an ACK.
 func (t *RUDP) SendUnreliableMsgTo(ctx context.Context, b common.RawBytes, a net.Addr) error {
 	id := t.nextPktID.Inc()
-	msg := t.putHeader(id, flagsNone, b)
-	return t.send(ctx, msg, a)
+	buffer, err := t.putHeader(id, flagsNone, b)
+	if err != nil {
+		return err
+	}
+	defer freepool.Put(buffer)
+	return t.send(ctx, buffer.B, a)
 }
 
 // SendMsgTo sends a message and waits for an ACK. If no ACK is received for a
@@ -127,7 +131,11 @@ func (t *RUDP) SendUnreliableMsgTo(ctx context.Context, b common.RawBytes, a net
 // ctx is not canceled.
 func (t *RUDP) SendMsgTo(ctx context.Context, b common.RawBytes, a net.Addr) error {
 	id := t.nextPktID.Inc()
-	msg := t.putHeader(id, flagNeedACK, b)
+	buffer, err := t.putHeader(id, flagNeedACK, b)
+	if err != nil {
+		return err
+	}
+	defer freepool.Put(buffer)
 	// Store the channel in the shared table s.t. the background receiver can
 	// close it when it gets the ACK
 	ackChannel := make(chan struct{})
@@ -139,7 +147,7 @@ func (t *RUDP) SendMsgTo(ctx context.Context, b common.RawBytes, a net.Addr) err
 
 	defer t.ackTable.Delete(id)
 	for {
-		if err := t.send(ctx, msg, a); err != nil {
+		if err := t.send(ctx, buffer.B, a); err != nil {
 			return err
 		}
 		select {
@@ -147,7 +155,7 @@ func (t *RUDP) SendMsgTo(ctx context.Context, b common.RawBytes, a net.Addr) err
 			// Received ack and can return successfully
 			return nil
 		case <-ctx.Done():
-			// Context was canceled and we are out of time, return with failure
+			// Context was canceled or we are out of time, return with failure
 			return ErrContextDone
 		case <-time.After(rudpRetryTimeout):
 			// Did not get ACK and context is not canceled yet, so do nothing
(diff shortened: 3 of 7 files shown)


scrye:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/infra/messaging/types.go, line 67 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L04hvh3uLeIj1zJ74_f:-L04hvh3uLeIj1zJ74_g:b6uz9yk) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/types.go#L67)):*
> ```Go
> 	for {
> 		old := atomic.LoadUint64((*uint64)(u))
> 		new := (old + 1) % (1<<56 - 1)
> ```

Note to self: drop the `-1` here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

scrye:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/messaging/udp.go, line 205 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L04j1SS5BN-EUQ9KvBC:-L04j1ST4W5SHDH8K_wV:bnwywx3) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L205)):*
> ```Go
> 	case event := <-t.readEvents:
> 		// Propagate message payload to caller
> 		return event.buffer.CloneB(), event.address, nil
> ```

Note to self: return buffer to free pool here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

scrye:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/messaging/udp.go, line 281 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L04jOgiZijJw53bFJz1:-L04jOgiZijJw53bFJz2:b-8qigly) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L281)):*
> ```Go
> 				// (if requested)
> 				if flags.isSet(flagNeedACK) {
> 					t.sendACK(id, address)
> ```

Note to self: catch errors here (probably ctx timeout).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

kormat:
Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/messaging/messaging.go, line 33 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05RnynQcI5MBp5_aJ9:-L05RnynQcI5MBp5_aJA:b3vet96) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/messaging.go#L33)):*
> ```Go
> 	// Send a message without expecting an ACK.
> 	SendUnreliableMsgTo(context.Context, common.RawBytes, net.Addr) error
> 	// Send a message, waiting for it to be ACK'd.
> ```

The semantics here are clearly written for unreliable protocols, but unclear for reliable protocols.

---

*[go/lib/infra/messaging/types.go, line 25 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L052hXjwGR0KPSUeN3y:-L052hXjwGR0KPSUeN3z:bmqwtat) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/types.go#L25)):*
> ```Go
> 
> // ackTable maps packet IDs to channels. Goroutines waiting for an ACK for a
> // session ID create a channel, store it in the map at that ID and wait for the
> ```

"packet ID" vs "session ID"?

---

*[go/lib/infra/messaging/types.go, line 75 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05BErQwBXMRBOj-gpW:-L05BErQwBXMRBOj-gpX:bl8fyhr) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/types.go#L75)):*
> ```Go
> }
> 
> func (a uint56) putUint56(b common.RawBytes) {
> ```

This is using little-endian encoding... Had to write a test program to confirm:
`0x1234567890abcd` becomes `cdab9078563412`.

---

*[go/lib/infra/messaging/udp.go, line 106 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05BilI35q87qOQEXky:-L05BilJ-GLuXc5n495H:b-1b5iof) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L106)):*
> ```Go
> // NewRUDP also spawns a background receiving goroutine that continuously reads
> // from conn and keeps track of ACKs and messages.
> func NewUDPTransport(conn net.PacketConn, logger log.Logger) *RUDP {
> ```

Name and comment do not match.

---

*[go/lib/infra/messaging/udp.go, line 150 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05GruOjS0l9Wps7x9A:-L05GruOjS0l9Wps7x9B:bj8k8s3) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L150)):*
> ```Go
> 			return nil
> 		case <-ctx.Done():
> 			// Context was canceled and we are out of time, return with failure
> ```

s/and/or/

---

*[go/lib/infra/messaging/udp.go, line 164 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05FhzFaTvAoL_iG8dN:-L05FhzFaTvAoL_iG8dO:b-xl4nlr) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L164)):*
go vet says:
> the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak

From what i can tell, i think the fix is to do:
```
ctx, cancel := ...
defer cancel()
return t.send(...
```

---

*[go/lib/infra/messaging/udp.go, line 193 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05CNnaS6_AC41Sv9dd:-L05CNnaS6_AC41Sv9de:bajycwx) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L193)):*
> ```Go
> // putHeader returns a new buffer containing the Reliable UDP header and b.
> func (t *RUDP) putHeader(id uint56, flags rudpFlag, b common.RawBytes) common.RawBytes {
> 	packet := make(common.RawBytes, rudpHdrLen)
> ```

If you're going to copy the whole packet anyway, you might as well allocate the full length to begin with. `make(common.RawBytes, rudpHdrLen, rudpHdrLen + len(b))`

---

*[go/lib/infra/messaging/udp.go, line 230 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05O-mp72foY6D5-DSE:-L05O-mp72foY6D5-DSF:blsf2x6) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L230)):*
> ```Go
> func (t *RUDP) goBackgroundReceiver() {
> 	go func() {
> 		liblog.LogPanicAndExit()
> ```

Noo, this needs to be `defer ...`, otherwise it doesn't do anything.

---

*[go/lib/infra/messaging/udp.go, line 242 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05PSnC2mmJ7ghDesJd:-L05PSnC2mmJ7ghDesJe:b18v0y6) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L242)):*
> ```Go
> 				if opErr, ok := err.(*snet.OpError); ok && opErr.SCMP() != nil {
> 					t.log.Warn("Received SCMP message", "msg", opErr.SCMP())
> 					continue
> ```

I think this is missing a `freepool.Put`

---

*[go/lib/infra/messaging/udp.go, line 266 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05Q1esm490lRWWXwUB:-L05Q1etj34deTh2sgks:bb6duqs) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L266)):*
> ```Go
> 				ackChannel, loaded := t.ackTable.Load(id)
> 				if !loaded {
> 					t.log.Warn("Received ACK, but no one is waiting fot it", "id", id)
> ```

*for

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

scrye:
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/messaging/messaging.go, line 33 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05RnynQcI5MBp5_aJ9:-L05cTSLP9mrrES2cN9k:b-wtxirv) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/messaging.go#L33)):*

The semantics here are clearly written for unreliable protocols, but unclear for reliable protocols.
</blockquote></details>

I think we still need ACKs, even for reliable protocols. E.g., for TCP, it gives a strong guarantee that the receiving host has received up to a certain sequence number, but that information is difficult to see from the sending app. For a long lived connection, you can keep sending hundreds of messages without any actually getting processed by the remote end, just because the local kernel's buffer has the space so `send` keeps returning.

---

*[go/lib/infra/messaging/types.go, line 25 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L052hXjwGR0KPSUeN3y:-L05V4qW6SnXSmtJD0pw:bmzz9xm) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/types.go#L25)):*

"packet ID" vs "session ID"?
</blockquote></details>

Packet ID, done.

---

*[go/lib/infra/messaging/types.go, line 75 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05BErQwBXMRBOj-gpW:-L05V6T_HOB8WG6X9s5Q:bhk4qjp) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/types.go#L75)):*

This is using little-endian encoding... Had to write a test program to confirm:
`0x1234567890abcd` becomes `cdab9078563412`.
</blockquote></details>

Whoops, you're right. Now it's always network byte order and added a docstring to say this.

---

*[go/lib/infra/messaging/udp.go, line 106 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05BilI35q87qOQEXky:-L05W9joJGK2komXIM0U:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L106)):*

Name and comment do not match.
</blockquote></details>

Done.

---

*[go/lib/infra/messaging/udp.go, line 150 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05GruOjS0l9Wps7x9A:-L05WHT41063affmSO-Z:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L150)):*

s/and/or/
</blockquote></details>

Done.

---

*[go/lib/infra/messaging/udp.go, line 164 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05FhzFaTvAoL_iG8dN:-L05WIIzW44ZIy3LZCHx:b-x6okux) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L164)):*

go vet says:
> the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak

From what i can tell, i think the fix is to do:
```
ctx, cancel := ...
defer cancel()
return t.send(...
```
</blockquote></details>

Yes, cancel functions were missing everywhere (a lot in the test file). Added them now.

---

*[go/lib/infra/messaging/udp.go, line 193 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05CNnaS6_AC41Sv9dd:-L05YQlGuCRHcieudWVm:b9xacs) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L193)):*

If you're going to copy the whole packet anyway, you might as well allocate the full length to begin with. `make(common.RawBytes, rudpHdrLen, rudpHdrLen + len(b))`
</blockquote></details>

Changed the code to use the free buffer pool.

---

*[go/lib/infra/messaging/udp.go, line 230 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05O-mp72foY6D5-DSE:-L05YplTcwVWGo60DLPi:b-kgkoda) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L230)):*

Noo, this needs to be `defer ...`, otherwise it doesn't do anything.
</blockquote></details>

Ouch, right. Done.

---

*[go/lib/infra/messaging/udp.go, line 242 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05PSnC2mmJ7ghDesJd:-L05ZbkCDAwFt_DtPy-Z:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L242)):*

I think this is missing a `freepool.Put`
</blockquote></details>

Done.

---

*[go/lib/infra/messaging/udp.go, line 266 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05Q1esm490lRWWXwUB:-L05Ze08TMMpNEmQ7L_f:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/udp.go#L266)):*

*for
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

kormat:
Reviewed 3 of 3 files at r2.

---

*[go/lib/infra/messaging/messaging.go, line 33 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05RnynQcI5MBp5_aJ9:-L08uubMEH5zSiuvPwjW:bc9o09j) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/messaging.go#L33)):*

I think we still need ACKs, even for reliable protocols. E.g., for TCP, it gives a strong guarantee that the receiving host has received up to a certain sequence number, but that information is difficult to see from the sending app. For a long lived connection, you can keep sending hundreds of messages without any actually getting processed by the remote end, just because the local kernel's buffer has the space so `send` keeps returning.
</blockquote></details>

The real answer, as you mentioned elsewhere, is that we need app-level ACKs. But that's a different layer. This layer is just the transport, so long as it provides retries i think that's all it needs to do, because app-level ACKs are needed regardless.

One thing that makes me a bit uncomfortable is the division of the API here. The choices for the user are:
- `non-blocking && unreliable`
- `blocking && reliable`
There's no way to do `non-blocking && reliable` (`blocking && unreliable` probably doesn't make sense :). This is why i had been suggesting that the caller optionally supplies a channel to be used to signal completion, so that it is in control of how this is handled.

---

*[go/lib/infra/messaging/udp.go, line 213 at r2](https://reviewable.io:443/reviews/scionproto/scion/1359#-L08xrdBHiwMq4F3oplQ:-L08xrdC82rUvXv2OOHv:b-wmd53g) ([raw file](https://github.com/scionproto/scion/blob/aeb210741aa27ed9901d25c5ef1691928b7fee23/go/lib/infra/messaging/udp.go#L213)):*
> ```Go
> func (t *RUDP) putHeader(id uint56, flags rudpFlag, b common.RawBytes) (*freepool.Buffer, error) {
> 	buffer := freepool.Get()
> 	if rudpHdrLen+len(b) > len(buffer.B) {
> ```

Buffer should be returned.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

scrye:
Review status: 4 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/infra/messaging/messaging.go, line 33 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05RnynQcI5MBp5_aJ9:-L091j3mMLfwaizwrZfr:b-ad107q) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/messaging.go#L33)):*

The real answer, as you mentioned elsewhere, is that we need app-level ACKs. But that's a different layer. This layer is just the transport, so long as it provides retries i think that's all it needs to do, because app-level ACKs are needed regardless.

One thing that makes me a bit uncomfortable is the division of the API here. The choices for the user are:
- `non-blocking && unreliable`
- `blocking && reliable`
There's no way to do `non-blocking && reliable` (`blocking && unreliable` probably doesn't make sense :). This is why i had been suggesting that the caller optionally supplies a channel to be used to signal completion, so that it is in control of how this is handled.
</blockquote></details>

I'm not sure we'll need another layer of ACKs on top of this messaging layer, but we'll see when we get there. RUDP provides a different feature compared to TCP, I don't think they're directly comparable.

Go's idiomatic way of doing nonblocking is to spawn a goroutine in the upper layer and have that block (I've seen a few RPC libraries work this way). The app gets full control of its inter-goroutine synchronization without propagating sync primitives down the stack (the context itself suffices for signaling, so everything is cleaner). (For us, this will be supported at the API level of the stack, the app itself doesn't need to see this).

---

*[go/lib/infra/messaging/udp.go, line 213 at r2](https://reviewable.io:443/reviews/scionproto/scion/1359#-L08xrdBHiwMq4F3oplQ:-L0934-HDDEVhClZXJ9P:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/aeb210741aa27ed9901d25c5ef1691928b7fee23/go/lib/infra/messaging/udp.go#L213)):*

Buffer should be returned.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

shitz:
Review status: 4 of 6 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/lib/infra/messaging/udp.go, line 106 at r2](https://reviewable.io:443/reviews/scionproto/scion/1359#-L096_vGrLl9B899Y3RR:-L096_vGrLl9B899Y3RS:b-7f4q1v) ([raw file](https://github.com/scionproto/scion/blob/aeb210741aa27ed9901d25c5ef1691928b7fee23/go/lib/infra/messaging/udp.go#L106)):*
> ```Go
> // NewRUDP also spawns a background receiving goroutine that continuously reads
> // from conn and keeps track of ACKs and messages.
> func NewRUDP(conn net.PacketConn, logger log.Logger) *RUDP {
> ```

This should initialize `nextPkdId` to a random valid starting value to reduce the risk of unintended associations.

---

*[go/lib/infra/messaging/udp.go, line 304 at r3](https://reviewable.io:443/reviews/scionproto/scion/1359#-L099QYC1kYXV2uLFIeO:-L099QYC1kYXV2uLFIeP:bbogdzw) ([raw file](https://github.com/scionproto/scion/blob/df49b00982576335413faf6f3c1c71b509eb2cb8/go/lib/infra/messaging/udp.go#L304)):*
> ```Go
> 				}
> 			default:
> 				t.log.Warn("Internal queue full, dropped message", "msg_len", n)
> ```

Also log the rudp header, so we can correlate packet ids.

---

*[go/lib/util/freepool/freepool.go, line 26 at r2](https://reviewable.io:443/reviews/scionproto/scion/1359#-L099wgmj0jif_2EbKEk:-L099wgn_spjie4OTL4j:b-9r65f0) ([raw file](https://github.com/scionproto/scion/blob/aeb210741aa27ed9901d25c5ef1691928b7fee23/go/lib/util/freepool/freepool.go#L26)):*
> ```Go
> // non-neglibile (e.g., per packet processing in a router), package ringbuf (and
> // manual management of free buffers) should be used instead.
> package freepool
> ```

I would prefer the name `bufpool`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

kormat:
Review status: 4 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/lib/util/freepool/freepool.go, line 26 at r2](https://reviewable.io:443/reviews/scionproto/scion/1359#-L099wgmj0jif_2EbKEk:-L09ALJR3rSI6UwEhrQq:b-8fx2u7) ([raw file](https://github.com/scionproto/scion/blob/aeb210741aa27ed9901d25c5ef1691928b7fee23/go/lib/util/freepool/freepool.go#L26)):*

I would prefer the name `bufpool`
</blockquote></details>

:+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

kormat:
Reviewed 1 of 2 files at r3.
Review status: 5 of 6 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

shitz:
Reviewed 2 of 6 files at r1, 2 of 3 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

scrye:

---

*[go/lib/infra/messaging/udp.go, line 106 at r2](https://reviewable.io:443/reviews/scionproto/scion/1359#-L096_vGrLl9B899Y3RR:-L09EBRtygEw-tC0lsEj:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/aeb210741aa27ed9901d25c5ef1691928b7fee23/go/lib/infra/messaging/udp.go#L106)):*

This should initialize `nextPkdId` to a random valid starting value to reduce the risk of unintended associations.
</blockquote></details>

Done.

---

*[go/lib/infra/messaging/udp.go, line 304 at r3](https://reviewable.io:443/reviews/scionproto/scion/1359#-L099QYC1kYXV2uLFIeO:-L09CEdUJF6NRFi9ibHR:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/df49b00982576335413faf6f3c1c71b509eb2cb8/go/lib/infra/messaging/udp.go#L304)):*

Also log the rudp header, so we can correlate packet ids.
</blockquote></details>

Done.

---

*[go/lib/util/freepool/freepool.go, line 26 at r2](https://reviewable.io:443/reviews/scionproto/scion/1359#-L099wgmj0jif_2EbKEk:-L09DnOH9HHKhqxfdnj1:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/aeb210741aa27ed9901d25c5ef1691928b7fee23/go/lib/util/freepool/freepool.go#L26)):*

:+1: 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

kormat:
Reviewed 1 of 2 files at r3.
Review status: 3 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/lib/infra/messaging/messaging.go, line 33 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05RnynQcI5MBp5_aJ9:-L09AU3Huxp_ksYwGYBA:b-fvvpq3) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/messaging.go#L33)):*
> I'm not sure we'll need another layer of ACKs on top of this messaging layer, but we'll see when we get there. 

It's one of the big things we're currently missing, and it's hurting us. E.g. when you register a path segment, you don't get a response, so you can't tell if it succeeded or not. Same for querying for a path/cert/etc and the server doesn't have an answer, it just drops the query.

> RUDP provides a different feature compared to TCP, I don't think they're directly comparable.

Which is why i'm not trying to compare them :P

We currently have two problems that we need to fix:
A. packets can get lost/dropped and they won't be re-sent
B. there's no response to most messages, so we can't tell if they pass the application logic even if they are delivered.

The way i see it is that this `Transport` interface is to fix A. Reliable protocols intrinsically take care of A in most cases, unreliable protocols don't take care of it at all. By putting ACKs into RUDP, it can then cover A (in slightly more cases than the reliable protocol's basic guarantees, but still not all cases).

Case B still needs to be handled at a higher level. This is where the RPC layer we've been planning comes in. As that is going to happen _anyway_, i don't see a need to try to make reliable protocols slightly more reliable by adding `Transport`-level ACKs to them.

[Ah, having read the updated descriptions, it seems you're taking the same approach as i had in mind anyway, so perhaps this was just a miscommunication :)]

> Go's idiomatic way of doing nonblocking is to spawn a goroutine in the upper layer and have that block (I've seen a few RPC libraries work this way). The app gets full control of its inter-goroutine synchronization without propagating sync primitives down the stack (the context itself suffices for signaling, so everything is cleaner). (For us, this will be supported at the API level of the stack, the app itself doesn't need to see this).

It doesn't get _full_ control, because non-blocking now _requires_ spawning a goroutine, whereas with the chan method is actually has a choice. But, ok, let's go with it for now at least.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

kormat:
:lgtm:

---

Reviewed 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359#-:-L09HVI3NNnySLeb32_w:bnfp4nl)*

scrye:
Review status: 5 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/infra/messaging/messaging.go, line 33 at r1](https://reviewable.io:443/reviews/scionproto/scion/1359#-L05RnynQcI5MBp5_aJ9:-L09JLhtViv9gH0npOkZ:bljut0v) ([raw file](https://github.com/scionproto/scion/blob/8584810483ce78be47e15c31fa8804754510fa83/go/lib/infra/messaging/messaging.go#L33)):*
> It doesn't get full control, because non-blocking now requires spawning a goroutine, whereas with the chan method is actually has a choice. But, ok, let's go with it for now at least.

The API will give it a choice, as it will expose methods for all app requirements (blocking, non-blocking, sync, async, etc.). I think the transport layer can be used to implement all of those, but we'll see later. We can change it if we need to.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

kormat:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359)*

shitz:
:lgtm:

---

Reviewed 3 of 4 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1359#-:-L09RrmB8u5QkS1iJX-d:bnfp4nl)*

PR #1364, opened 2017-12-13
Fix posixOutput out of bounds error.

kormat:
If WriteBatch returns an error, pktsWritten can be -1. This causes the
copy() at the end of the loop to panic with an out of bounds error.

This patch changes the logic when an error occurs:
- If _some_ packets were sent, process those as normal.
- If no packets were sent, and the error is known to be temporary,
  retry.
- If no packets were sent, and the error is not known to be temporary,
  drop the current batch of packets, and move on.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1364)

### DIFF
--- a/go/border/io.go
+++ b/go/border/io.go
@@ -29,6 +29,7 @@ import (
 	"github.com/scionproto/scion/go/border/rctx"
 	"github.com/scionproto/scion/go/border/rpkt"
 	"github.com/scionproto/scion/go/lib/assert"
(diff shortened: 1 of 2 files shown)


shitz:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1364#-:-L0EeRR49yn0e_a3jaFd:bnfp4nl)*

PR #1365, opened 2017-12-13
Add SessionId label to metrics and optimize metric instantiation

shitz:


kormat:
A top-level comment:
- This needs to include the remote IA, as session has no meaning outside of that scope.

---

Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/sig/ingress/dispatcher.go, line 101 at r1](https://reviewable.io:443/reviews/scionproto/scion/1365#-L0FQQpBAQXmiyn4v5RP:-L0FQQpBAQXmiyn4v5RQ:behpel5) ([raw file](https://github.com/scionproto/scion/blob/4ba3f94247548365843da6ee2c75e9ffba24158c/go/sig/ingress/dispatcher.go#L101)):*
> ```Go
> 				frame.sessId = mgmt.SessionType((frame.raw[0]))
> 				d.dispatch(frame, src)
> 				counters, ok := framesRecvCounters[frame.sessId]
> ```

This is illegal - you can't access `frame` after calling `dispatch` on it. Naughty.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1365#-:-L0EzJjWG3MhJCDsfTx5:bp8gata)*

shitz:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/sig/ingress/dispatcher.go, line 101 at r1](https://reviewable.io:443/reviews/scionproto/scion/1365#-L0FQQpBAQXmiyn4v5RP:-L4q5nI67Cb31VUlgmri:b-o7vuhc) ([raw file](https://github.com/scionproto/scion/blob/4ba3f94247548365843da6ee2c75e9ffba24158c/go/sig/ingress/dispatcher.go#L101)):*

This is illegal - you can't access `frame` after calling `dispatch` on it. Naughty.
</blockquote></details>

Good catch! Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1365)*

kormat:
Reviewed 7 of 7 files at r2.

---

*[go/sig/egress/dispatcher.go, line 21 at r2](https://reviewable.io:443/reviews/scionproto/scion/1365#-L4urNArqRHlWgeFi9BD:-L4urNArqRHlWgeFi9BE:bi8a4d3) ([raw file](https://github.com/scionproto/scion/blob/314a9ffb99446e3ea8399244b980e360b3b61852/go/sig/egress/dispatcher.go#L21)):*
> ```Go
> 	"os"
> 
> 	"github.com/scionproto/scion/go/sig/mgmt"
> ```

This should be in the scion import block.

---

*[go/sig/egress/dispatcher.go, line 111 at r2](https://reviewable.io:443/reviews/scionproto/scion/1365#-L4urv2bZ1evxp2sF3mP:-L4urv2bZ1evxp2sF3mQ:bblli2b) ([raw file](https://github.com/scionproto/scion/blob/314a9ffb99446e3ea8399244b980e360b3b61852/go/sig/egress/dispatcher.go#L111)):*
> ```Go
> 			if !ok {
> 				counters = metrics.CtrPair{
> 					Pkts:  metrics.PktsSent.WithLabelValues(ed.devName, sess.SessId.String()),
> ```

As the params for both metrics are the same, you could just make a `NewCtrPair(intf, sessId string) *CtrPair` constructor.

---

*[go/sig/ingress/dispatcher.go, line 48 at r2](https://reviewable.io:443/reviews/scionproto/scion/1365#-L4us_aJC74stHYiJ2P0:-L4us_aJC74stHYiJ2P1:bbz9y7j) ([raw file](https://github.com/scionproto/scion/blob/314a9ffb99446e3ea8399244b980e360b3b61852/go/sig/ingress/dispatcher.go#L48)):*
> ```Go
> 	freeFrames         *ringbuf.Ring
> 	pktsSentCounters   map[mgmt.SessionType]metrics.CtrPair
> 	framesRecvCounters map[mgmt.SessionType]metrics.CtrPair
> ```

This doesn't look right. A session Id is only valid when scoped to a specific AS, but these counters look to be global.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1365)*

shitz:
Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[go/sig/egress/dispatcher.go, line 21 at r2](https://reviewable.io:443/reviews/scionproto/scion/1365#-L4urNArqRHlWgeFi9BD:-L4uwYZXi0rtPRNvjfEX:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/314a9ffb99446e3ea8399244b980e360b3b61852/go/sig/egress/dispatcher.go#L21)):*

This should be in the scion import block.
</blockquote></details>

Done.

---

*[go/sig/egress/dispatcher.go, line 111 at r2](https://reviewable.io:443/reviews/scionproto/scion/1365#-L4urv2bZ1evxp2sF3mP:-L4uxL-rs0QRVfYulNmh:b-5jtfky) ([raw file](https://github.com/scionproto/scion/blob/314a9ffb99446e3ea8399244b980e360b3b61852/go/sig/egress/dispatcher.go#L111)):*

As the params for both metrics are the same, you could just make a `NewCtrPair(intf, sessId string) *CtrPair` constructor.
</blockquote></details>

The metric names themselves are also different each time so `NewCtrPair` either would need to take two additional params or it would need to be redefined in each package. Both I think is worse than the current thing.

---

*[go/sig/ingress/dispatcher.go, line 48 at r2](https://reviewable.io:443/reviews/scionproto/scion/1365#-L4us_aJC74stHYiJ2P0:-L4ztAy9DuHoRimf6e3d:beqt5ip) ([raw file](https://github.com/scionproto/scion/blob/314a9ffb99446e3ea8399244b980e360b3b61852/go/sig/ingress/dispatcher.go#L48)):*

This doesn't look right. A session Id is only valid when scoped to a specific AS, but these counters look to be global.
</blockquote></details>

All metrics (frames(sent/recv) and packets(sent/recv)) now take the remote IA and the session Id as labels. The packet metrics currently contained the tun device on which they were sent/received which implicitly contains the same information as remote IA for the pktsRecv metrics. For the pktsSent metrics this label was always `scion-local`, however, I think what we really want to have is the remoteIA that sent the frame(s) from which a packet got reassembled.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1365)*

kormat:
Reviewed 7 of 7 files at r3.

---

*[go/sig/egress/dispatcher.go, line 111 at r2](https://reviewable.io:443/reviews/scionproto/scion/1365#-L4urv2bZ1evxp2sF3mP:-L592MOLFjtEPTktI6Aw:b-jmuen0) ([raw file](https://github.com/scionproto/scion/blob/314a9ffb99446e3ea8399244b980e360b3b61852/go/sig/egress/dispatcher.go#L111)):*

The metric names themselves are also different each time so `NewCtrPair` either would need to take two additional params or it would need to be redefined in each package. Both I think is worse than the current thing.
</blockquote></details>

Ok. How about just moving this block into a seperate method `getCounters`? Or, even better, `updateMetrics`, and pass in sessId and length. Same for the ingress dispatcher.

---

*[go/sig/egress/dispatcher.go, line 42 at r3](https://reviewable.io:443/reviews/scionproto/scion/1365#-L592Be53BJxZYEK_uWP:-L592Be53BJxZYEK_uWQ:bbrqr1m) ([raw file](https://github.com/scionproto/scion/blob/cd935c5ee002670b474ba2fc6bbea5bd5ca2a282/go/sig/egress/dispatcher.go#L42)):*
> ```Go
> )
> 
> type metricKey struct {
> ```

Defining this type twice seems.. odd :) How about just moving it into `sig/metrics`?

---

*[go/sig/egress/dispatcher.go, line 113 at r3](https://reviewable.io:443/reviews/scionproto/scion/1365#-L58Cvno5J4BaT7sEL9F:-L58Cvno5J4BaT7sEL9G:bfs953v) ([raw file](https://github.com/scionproto/scion/blob/cd935c5ee002670b474ba2fc6bbea5bd5ca2a282/go/sig/egress/dispatcher.go#L113)):*
> ```Go
> 			}
> 			sess.ring.Write(ringbuf.EntryList{buf}, true)
> 			key := metricKey{sess.IA.IAInt(), sess.SessId}
> ```

As `sess.IA.IAInt()` will be a constant for a given `egressDispatcher`, this could be calculated at the top, and re-used.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1365)*

shitz:
Review status: 4 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/sig/egress/dispatcher.go, line 42 at r3](https://reviewable.io:443/reviews/scionproto/scion/1365#-L592Be53BJxZYEK_uWP:-L5IWWTRIUm6pMLVGlgh:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cd935c5ee002670b474ba2fc6bbea5bd5ca2a282/go/sig/egress/dispatcher.go#L42)):*

Defining this type twice seems.. odd :) How about just moving it into `sig/metrics`?
</blockquote></details>

Done.

---

*[go/sig/egress/dispatcher.go, line 113 at r3](https://reviewable.io:443/reviews/scionproto/scion/1365#-L58Cvno5J4BaT7sEL9F:-L5IT38IxWY9QN17a0uE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cd935c5ee002670b474ba2fc6bbea5bd5ca2a282/go/sig/egress/dispatcher.go#L113)):*

As `sess.IA.IAInt()` will be a constant for a given `egressDispatcher`, this could be calculated at the top, and re-used.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1365)*

kormat:
Reviewed 3 of 3 files at r4.

---

*[go/sig/ingress/dispatcher.go, line 98 at r4](https://reviewable.io:443/reviews/scionproto/scion/1365#-L5IYij3Ul56UFJ_LVDA:-L5IYij3Ul56UFJ_LVDB:b-5tg191) ([raw file](https://github.com/scionproto/scion/blob/44c335ea453e589138800e1bcf6a9e53e0b560b6/go/sig/ingress/dispatcher.go#L98)):*
> ```Go
> 				frame.frameLen = read
> 				frame.sessId = mgmt.SessionType((frame.raw[0]))
> 				key := metrics.CtrPairKey{RemoteIA: src.IA.IAInt(), SessId: frame.sessId}
> ```

This is a large block of code just to handle metrics, it ~doubles the size of this method. I suggest moving it out to a dedicated method, and call it with `IAInt`, `SessId`, and `read`.

The same applies to `egressDispatcher.Run`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1365)*

shitz:

---

*[go/sig/ingress/dispatcher.go, line 98 at r4](https://reviewable.io:443/reviews/scionproto/scion/1365#-L5IYij3Ul56UFJ_LVDA:-L5IbsS6Dc6yx0Zoy8_T:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/44c335ea453e589138800e1bcf6a9e53e0b560b6/go/sig/ingress/dispatcher.go#L98)):*

This is a large block of code just to handle metrics, it ~doubles the size of this method. I suggest moving it out to a dedicated method, and call it with `IAInt`, `SessId`, and `read`.

The same applies to `egressDispatcher.Run`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1365)*

kormat:
:lgtm:

---

Reviewed 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1365#-:-L5IcJ2ShYx7dmPNz-RY:bnfp4nl)*

PR #1366, opened 2017-12-18
Message dispatcher for SCION infra

scrye:
Also:
  - A package implementing a simple point-to-point pair of net.PacketConn connections that can be used for testing

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1366)


shitz:
Reviewed 2 of 9 files at r1, 4 of 7 files at r2.
Review status: 6 of 13 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/infra.go, line 1 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jDLEve9PTv2Vf3sfx:-L0jDLEve9PTv2Vf3sfy:bnj4iyw) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/infra.go#L1)):*
> ```Go
> ```

This should be called `error.go`

---

*[go/lib/infra/infra.go, line 44 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jeWxvSiqfjV8wDdxI:-L0jeWxw78QsqtGffan-:bazxwrs) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/infra.go#L44)):*
> ```Go
> }
> 
> func WrapAdapterError(err error, ctx ...interface{}) error {
> ```

I'd prefer replacing all instances of `Wrap` with `New`

---

*[go/lib/infra/disp/disp.go, line 24 at r1](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0j7z02hiB9SBH2zOdC:-L0j7z03wTWLfbs_PiSi:b-670t2k) ([raw file](https://github.com/scionproto/scion/blob/d04b34e4d1689530c675db5e61dd80b283b97e14/go/lib/infra/disp/disp.go#L24)):*
> ```Go
> //
> // A Dispatcher can be customized by implementing interface MessageAdapter. The
> // inteface instructs the dispatcher how to convert a message to its raw
> ```

interface

---

*[go/lib/infra/disp/disp.go, line 194 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0j_LPIEjGp6VCN85TH:-L0j_LPIEjGp6VCN85TI:b68b10c) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/disp/disp.go#L194)):*
> ```Go
> 		return
> 	}
> 	if d.adapter.MsgIsReply(msg) {
> ```

I'm not 100% convinced if `MsgIsReply` is useful. I here I would just check the `waitTable` and if no outstanding request is there for this message just pass it on to the app. I'm not sure if we can always make a very clear distinction between a reply and some other notification (but maybe it would be worth it to be able to do that). The app anyway needs code to drop/ignore unwanted messages thus I also don't think you'd save a lot of complexity in the app by not passing it along. 

---

*[go/lib/infra/disp/wtable.go, line 45 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jdz-bCxGFIBfa0jeX:-L0jdz-bCxGFIBfa0jeY:b-9p4nzu) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/disp/wtable.go#L45)):*
> ```Go
> 		// Continue below
> 	}
> 	// Destroy can be called between now and when we exit this function.
> ```

Remove this comment and the blank line everywhere.

---

*[go/lib/infra/disp/wtable.go, line 73 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jeFdFOhGMMPcP4pox:-L0jeFdGV-dO854v3tgB:bqx6ynr) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/disp/wtable.go#L73)):*
> ```Go
> 		return nil, common.NewCError("Key not found", "key", wt.keyF(object))
> 	}
> ```

remove blank line

---

*[go/lib/infra/disp/wtable.go, line 97 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jeCu5QVAdsAZYSXEY:-L0jeCu5QVAdsAZYSXEZ:bco1vef) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/disp/wtable.go#L97)):*
> ```Go
> 		return common.NewCError("Reply received, but no one is waiting", "key", wt.keyF(object))
> 	}
> ```

Remove blank line.

---

*[go/lib/infra/disp/wtable_test.go, line 27 at r1](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jiDaqtYT72cX6z6Ts:-L0jiDaqtYT72cX6z6Tt:bhfjkc7) ([raw file](https://github.com/scionproto/scion/blob/d04b34e4d1689530c675db5e61dd80b283b97e14/go/lib/infra/disp/wtable_test.go#L27)):*
> ```Go
> )
> 
> func Test_FunctionWT_Wait(t *testing.T) {
> ```

Add a testcase for a destroyed wait table as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

shitz:
Reviewed 2 of 9 files at r1, 3 of 7 files at r2.
Review status: 11 of 13 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/messaging/udp.go, line 29 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jiOy8aNmGy67diH1U:-L0jiOy8aNmGy67diH1V:b-ap0li9) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/messaging/udp.go#L29)):*
> ```Go
> 	logext "github.com/inconshreveable/log15/ext"
> 
> 	. "github.com/scionproto/scion/go/lib/common"
> ```

You have to get @kormat's blessing for this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

scrye:
Review status: 7 of 13 files reviewed at latest revision, 9 unresolved discussions.

---

*[go/lib/infra/disp/disp.go, line 24 at r1](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0j7z02hiB9SBH2zOdC:-L0yFBChFKbpBUe2aoXf:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d04b34e4d1689530c675db5e61dd80b283b97e14/go/lib/infra/disp/disp.go#L24)):*

interface
</blockquote></details>

Done.

---

*[go/lib/infra/disp/disp.go, line 194 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0j_LPIEjGp6VCN85TH:-L0yFCh5Qd6D7Ktv6vUf:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/disp/disp.go#L194)):*

I'm not 100% convinced if `MsgIsReply` is useful. I here I would just check the `waitTable` and if no outstanding request is there for this message just pass it on to the app. I'm not sure if we can always make a very clear distinction between a reply and some other notification (but maybe it would be worth it to be able to do that). The app anyway needs code to drop/ignore unwanted messages thus I also don't think you'd save a lot of complexity in the app by not passing it along. 
</blockquote></details>

Done.

---

*[go/lib/infra/disp/wtable.go, line 45 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jdz-bCxGFIBfa0jeX:-L0yFLmngCnprPv7EGdE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/disp/wtable.go#L45)):*

Remove this comment and the blank line everywhere.
</blockquote></details>

Done.

---

*[go/lib/infra/disp/wtable.go, line 73 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jeFdFOhGMMPcP4pox:-L0yGtH3HPJgFNUKdXCv:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/disp/wtable.go#L73)):*

remove blank line
</blockquote></details>

Done.

---

*[go/lib/infra/disp/wtable.go, line 97 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jeCu5QVAdsAZYSXEY:-L0yH-YsnNMilXBsZ7ZE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/disp/wtable.go#L97)):*

Remove blank line.
</blockquote></details>

Done.

---

*[go/lib/infra/disp/wtable_test.go, line 27 at r1](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jiDaqtYT72cX6z6Ts:-L0yIZZrNtb9mJzVTHeq:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d04b34e4d1689530c675db5e61dd80b283b97e14/go/lib/infra/disp/wtable_test.go#L27)):*

Add a testcase for a destroyed wait table as well.
</blockquote></details>

Done.

---

*[go/lib/infra/infra.go, line 1 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jDLEve9PTv2Vf3sfx:-L0yEv-9xpZX_dVjwY3p:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/infra.go#L1)):*

This should be called `error.go`
</blockquote></details>

Done.

---

*[go/lib/infra/infra.go, line 44 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jeWxvSiqfjV8wDdxI:-L0yF7n-IpKsMBHKDv8P:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/infra.go#L44)):*

I'd prefer replacing all instances of `Wrap` with `New`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

kormat:
Reviewed 2 of 8 files at r2, 5 of 7 files at r3.
Review status: 9 of 13 files reviewed at latest revision, 24 unresolved discussions.

---

*[go/lib/infra/errors.go, line 22 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wBQlk746AS9GTfaAl:-L1wBQlk746AS9GTfaAm:b-50vyo3) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L22)):*
> ```Go
> 	"fmt"
> 
> 	. "github.com/scionproto/scion/go/lib/common"
> ```

Please don't do this: https://github.com/golang/go/wiki/CodeReviewComments#import-dot
It's also not supported by some standard tools (e.g. https://github.com/rogpeppe/godef/issues/19)

---

*[go/lib/infra/errors.go, line 26 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wXlV-gGG9pULcyBcH:-L1wXlV-gGG9pULcyBcI:bltb10e) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L26)):*
> ```Go
> 
> const (
> 	strCtxDoneError   = "context canceled"
> ```

Making these unexported mostly defeats the purpose of having them defined as constants.

---

*[go/lib/infra/errors.go, line 39 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1w_jSXQbeNgPqiM53x:-L1w_jSXQbeNgPqiM53y:b-yccr97) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L39)):*
> ```Go
> func NewCtxDoneError(ctx ...interface{}) error {
> 	meta := &errMeta{
> 		timeout: true,
> ```

This isn't necessarily true. If the caller calls its own cancelF, for example, this case will happen without it being a timeout.

---

*[go/lib/infra/errors.go, line 66 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wYFmKGtoaincXPAAh:-L1wYFmKGtoaincXPAAi:b-1mhrae) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L66)):*
> ```Go
> // Returns true if err was
> func IsTimeout(err error) bool {
> 	cmnErr, ok := err.(*CError)
> ```

`cerr` is the naming convention for this.

---

*[go/lib/infra/errors.go, line 79 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wYZu4MhxU-W0vV1K6:-L1wYZu4MhxU-W0vV1K7:bsm3xlu) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L79)):*
> ```Go
> // Mutates the state of err to contain additional context, while preserving the
> // original metadata.
> func WrapError(err error, desc string, ctx ...interface{}) error {
> ```

This seems to completely lose access to the original err, making it unuseful for programmatic consumption. I have a plan to change `CError.Cerr` to `CError.Err` (type `error`) to make nesting better, but currently the convention we have is to just do `common.NewCError(mydesc, "my", ctx, "err", olderr)` (which happens to format badly, and is also not useful for programmatic consumption...). 

Hmm. I should really prioritise the error nesting stuff.

---

*[go/lib/infra/errors.go, line 83 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wYcdOouthJ9w-lzGB:-L1wYcdOouthJ9w-lzGC:b-f2t8gn) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L83)):*
> ```Go
> 	if !ok {
> 		// Convert standard error to CError
> 		cmnErr = NewCError(err.Error()).(*CError)
> ```

This doesn't do anything. `err` is what's returned, which is unmodified.

---

*[go/lib/infra/disp/disp.go, line 21 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wK_6LvzDx_nB9l8so:-L1wK_6LvzDx_nB9l8sp:b-wsbzmc) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L21)):*
> ```Go
> // a reply message with the same key is received), Notify (send a message,
> // block until an ACK is received from the remote Transport protocol),
> // NotifyUnreliable (send a message, return immediately).
> ```

This needs a big notice to clarify the meaning of "reliable" vs "unreliable", as it's completely misleading otherwise if the transport layer is actually reliable. Referencing the `messaging` definition would help a lot.

---

*[go/lib/infra/disp/disp.go, line 50 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wLAlBKJRZLa_W_xjF:-L1wLAlBKJRZLa_W_xjG:bu0r7bx) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L50)):*
> ```Go
> 	"github.com/scionproto/scion/go/lib/infra"
> 	liblog "github.com/scionproto/scion/go/lib/log"
> ```

Shouldn't be a blank line here.

---

*[go/lib/infra/disp/disp.go, line 71 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wU4dikeLWewio6kCC:-L1wU4dikeLWewio6kCD:b-3jfocq) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L71)):*
> ```Go
> 	adapter MessageAdapter
> 	// Closed when background goroutine shuts down
> 	stoppedC chan struct{}
> ```

The equivalent channels in `messaging.RUDP` are:
```
    // Closed when Close() starts to run
    closed chan struct{}
    // Closed when background goroutine finishes shutting down
    backgroundDone chan struct{}
```
Some naming consistency would be good. `git grep 'C .*chan'` shows no hits outside of this PR.

---

*[go/lib/infra/disp/disp.go, line 81 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wLk57otW00o_2ifne:-L1wLk57otW00o_2ifnf:b-2bvh23) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L81)):*
> ```Go
> }
> 
> // New creates a new dispatcher backed by transport t, and using adapter to
> ```

Func name doesn't match.

---

*[go/lib/infra/disp/disp.go, line 84 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wMFGbJ1mFH6U7DMRX:-L1wMFGbJ1mFH6U7DMRY:b-ioja5j) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L84)):*
> ```Go
> // convert generic Message objects to and from their raw representation.
> //
> // All methods guarantee to return immediately once their context is canceled.
> ```

"immediately" isn't true or even possible. E.g. if `Request` is currently blocked on writing to the connection, then the context cancellation can only be noticed once the deadline is exceeded.

---

*[go/lib/infra/disp/disp.go, line 115 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wNsS7jMJAt66z7I48:-L1wNsS7jMJAt66z7I49:bceqr2r) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L115)):*
> ```Go
> 	b, err := d.adapter.MsgToRaw(msg)
> 	if err != nil {
> 		return nil, infra.NewAdapterError(err, "op", "waitTable.MsgToRaw")
> ```

This isn't `waitTable`.

---

*[go/lib/infra/disp/disp.go, line 128 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wOmuzJ7qc7l91ax7j:-L1wOmuzJ7qc7l91ax7k:boht6ew) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L128)):*
> ```Go
> }
> 
> // Notify sends msg to address, and returns once the send has been ACK'd by the
> ```

This is only true for some possible transports, not for others.

---

*[go/lib/infra/disp/disp.go, line 187 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wbyozwY7OhbFoM11V:-L1wbyozwY7OhbFoM11W:b-6nd8lu) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L187)):*
> ```Go
> func (d *Dispatcher) recvNext() {
> 	// Once the transport is closed, RecvFrom returns immediately.
> 	b, address, err := d.transport.RecvFrom(context.TODO())
> ```

I think this should be https://golang.org/pkg/context/#Background, it seems to be specifically for this kind of case.

---

*[go/lib/infra/disp/wtable.go, line 88 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wWHAbS37BZutF2qRn:-L1wWHAbS37BZutF2qRo:bszb3t7) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/wtable.go#L88)):*
> ```Go
> 	default:
> 	}
> 	replyChannel, loaded := wt.replyMap.Load(wt.keyF(object))
> ```

`ok` is a more conventional name for this bool (and what's used in the Go docs for the func).

---

*[go/lib/infra/messaging/udp.go, line 29 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jiOy8aNmGy67diH1U:-L1wTuOO0E5QFvdYSvxq:b-e6cstr) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/messaging/udp.go#L29)):*

You have to get @kormat's blessing for this.
</blockquote></details>

As commented on `errors.go`, :-1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

scrye:
Review status: 4 of 13 files reviewed at latest revision, 24 unresolved discussions.

---

*[go/lib/infra/errors.go, line 22 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wBQlk746AS9GTfaAl:-L2K0ce_cCfToc_w8Lue:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L22)):*

Please don't do this: https://github.com/golang/go/wiki/CodeReviewComments#import-dot
It's also not supported by some standard tools (e.g. https://github.com/rogpeppe/godef/issues/19)
</blockquote></details>

Done.

---

*[go/lib/infra/errors.go, line 26 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wXlV-gGG9pULcyBcH:-L2K0lCERn48PlYINaUT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L26)):*

Making these unexported mostly defeats the purpose of having them defined as constants.
</blockquote></details>

Done.

---

*[go/lib/infra/errors.go, line 39 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1w_jSXQbeNgPqiM53x:-L2K0xLUif0jBDMVGzps:b-45bakb) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L39)):*

This isn't necessarily true. If the caller calls its own cancelF, for example, this case will happen without it being a timeout.
</blockquote></details>

Well, a canceled context is still a timeout (the operation didn't complete before its assigned context was destroyed). In any case, the owner knows it called the cancel function already and doesn't need to inspect for timeouts.

---

*[go/lib/infra/errors.go, line 66 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wYFmKGtoaincXPAAh:-L2K1UbQNohJu7EtfTcL:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L66)):*

`cerr` is the naming convention for this.
</blockquote></details>

Done.

---

*[go/lib/infra/errors.go, line 83 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wYcdOouthJ9w-lzGB:-L2K1lsenW-yKdf9-lVZ:b3dp2yl) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L83)):*

This doesn't do anything. `err` is what's returned, which is unmodified.
</blockquote></details>

The wrong error was returned, fixed now.

---

*[go/lib/infra/disp/disp.go, line 21 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wK_6LvzDx_nB9l8so:-L2K4WXnbQDHhEz-EFWK:b-uikz3s) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L21)):*

This needs a big notice to clarify the meaning of "reliable" vs "unreliable", as it's completely misleading otherwise if the transport layer is actually reliable. Referencing the `messaging` definition would help a lot.
</blockquote></details>

Ok, changed it a bit to hopefully make it clearer.

---

*[go/lib/infra/disp/disp.go, line 50 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wLAlBKJRZLa_W_xjF:-L2K4nipyzBi6BcoEisc:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L50)):*

Shouldn't be a blank line here.
</blockquote></details>

Done.

---

*[go/lib/infra/disp/disp.go, line 71 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wU4dikeLWewio6kCC:-L2K5C-aafOt_38dyFq6:bjy2fn9) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L71)):*

The equivalent channels in `messaging.RUDP` are:
```
    // Closed when Close() starts to run
    closed chan struct{}
    // Closed when background goroutine finishes shutting down
    backgroundDone chan struct{}
```
Some naming consistency would be good. `git grep 'C .*chan'` shows no hits outside of this PR.
</blockquote></details>

It's inspired by the go standard libs (e.g., `net/http`). I was planning on doing a refactoring pass on `messaging` after this gets merged (I have a list of things I need to update there). I've updated the channel names there now.

---

*[go/lib/infra/disp/disp.go, line 81 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wLk57otW00o_2ifne:-L2K5td88EWhe_t5J3R2:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L81)):*

Func name doesn't match.
</blockquote></details>

Done.

---

*[go/lib/infra/disp/disp.go, line 84 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wMFGbJ1mFH6U7DMRX:-L2K5vI7hbbj1Z9rjnYF:b7y442h) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L84)):*

"immediately" isn't true or even possible. E.g. if `Request` is currently blocked on writing to the connection, then the context cancellation can only be noticed once the deadline is exceeded.
</blockquote></details>

Ah, yes, you're right. I forgot about manual cancellation.

I've clarified the issue (although having the ability to cancel stuff at any time would be useful; this means propagating contexts through snet all the way to `go/lib/sock/reliable`, which is a lot of work).

---

*[go/lib/infra/disp/disp.go, line 115 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wNsS7jMJAt66z7I48:-L2K86ChCjOsoSI8jqo-:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L115)):*

This isn't `waitTable`.
</blockquote></details>

Done.

---

*[go/lib/infra/disp/disp.go, line 128 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wOmuzJ7qc7l91ax7j:-L2K8Uo8IFSfJfpdAgbq:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L128)):*

This is only true for some possible transports, not for others.
</blockquote></details>

Done.

---

*[go/lib/infra/disp/disp.go, line 187 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wbyozwY7OhbFoM11V:-L2K8dFzKX-xPU8y8DZA:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L187)):*

I think this should be https://golang.org/pkg/context/#Background, it seems to be specifically for this kind of case.
</blockquote></details>

Done.

---

*[go/lib/infra/disp/wtable.go, line 88 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wWHAbS37BZutF2qRn:-L2K8kg90utqP0nQrnPw:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/wtable.go#L88)):*

`ok` is a more conventional name for this bool (and what's used in the Go docs for the func).
</blockquote></details>

Done.

---

*[go/lib/infra/messaging/udp.go, line 29 at r2](https://reviewable.io:443/reviews/scionproto/scion/1366#-L0jiOy8aNmGy67diH1U:-L2K928kPrLuzyGC8YUn:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb9ff3a56f5cbfb2fa21820db178389526c15e31/go/lib/infra/messaging/udp.go#L29)):*

As commented on `errors.go`, :-1: 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

kormat:
Reviewed 2 of 9 files at r1, 1 of 7 files at r3, 4 of 5 files at r4.
Review status: 9 of 13 files reviewed at latest revision, 11 unresolved discussions.

---

*[go/lib/infra/errors.go, line 79 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wYZu4MhxU-W0vV1K6:-L2KGcGiJASfOjPqWVqW:b6s35zd) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L79)):*

This seems to completely lose access to the original err, making it unuseful for programmatic consumption. I have a plan to change `CError.Cerr` to `CError.Err` (type `error`) to make nesting better, but currently the convention we have is to just do `common.NewCError(mydesc, "my", ctx, "err", olderr)` (which happens to format badly, and is also not useful for programmatic consumption...). 

Hmm. I should really prioritise the error nesting stuff.
</blockquote></details>

#1369 out for review to make all of this a lot better.

---

*[go/lib/infra/disp/disp.go, line 71 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wU4dikeLWewio6kCC:-L2KIiP8JIg1mWinfFwO:b-b8cmy1) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L71)):*
> It's inspired by the go standard libs (e.g., net/http).

Which "it"? I've done a grep over the go stdlib and don't see a major convention of a trailing `C` for channel name. (`grep -R '\SC .*chan' /usr/local/go/src` shows two instances, both in examples)
There are no channels named `stopped` in the stdlib.  So i'm quite lost what you're referring to.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

scrye:
Review status: 9 of 13 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/errors.go, line 79 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wYZu4MhxU-W0vV1K6:-L2LZvZINxSpCV1JtSzU:bdo7ewu) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/errors.go#L79)):*

#1369 out for review to make all of this a lot better.
</blockquote></details>

Done, the new error nesting stuff plugged in really nicely.

---

*[go/lib/infra/disp/disp.go, line 71 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wU4dikeLWewio6kCC:-L2L9QLlFjR4RfY0h0Dk:b5jipcn) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L71)):*

> It's inspired by the go standard libs (e.g., net/http).

Which "it"? I've done a grep over the go stdlib and don't see a major convention of a trailing `C` for channel name. (`grep -R '\SC .*chan' /usr/local/go/src` shows two instances, both in examples)
There are no channels named `stopped` in the stdlib.  So i'm quite lost what you're referring to.
</blockquote></details>

Ah, I admit, I shortened their version a bit :stuck_out_tongue:.

Try `grep -R '\SChan .*chan' /usr/local/go/src`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

kormat:
Reviewed 1 of 7 files at r3, 5 of 8 files at r5.
Review status: 9 of 13 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/infra/disp/disp.go, line 71 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wU4dikeLWewio6kCC:-L2Uw7UbqOJ3O_FkPwwW:bqea714) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L71)):*

Ah, I admit, I shortened their version a bit :stuck_out_tongue:.

Try `grep -R '\SChan .*chan' /usr/local/go/src`.
</blockquote></details>

Ah. So, i'd be a lot happier with `stoppedCh` than `stoppedC`. It's a bit hungarian-notationy for my taste, especially in a strongly-type language, but i'd be ok with it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

kormat:
Reviewed 1 of 7 files at r3, 3 of 8 files at r5.

---

*[go/lib/infra/disp/disp_test.go, line 1 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2V17WpDndFqXT2pxXK:-L2V17WpDndFqXT2pxXL:b-7fm8ui) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/disp_test.go#L1)):*
> ```Go
> ```

All the timeouts in this test make it very slow to run - it takes 15s just for this file (out of 22s for all the go unit tests). Also, as all the tests are bundled into large test functions, it's also really difficult to see exactly how long each individual test takes (which makes it hard to confirm assumptions about where the delays are coming from). Having ~one function per test case would make this a lot easier to work with.

---

*[go/lib/infra/disp/disp_test.go, line 30 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2UzriXmMDeqDK3gaKq:-L2UzriXmMDeqDK3gaKr:b-i4f2k0) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/disp_test.go#L30)):*
> ```Go
> 	"github.com/scionproto/scion/go/lib/xtest/p2p"
> 
> 	. "github.com/smartystreets/goconvey/convey"
> ```

The import block looks like:
- stdlib imports
- third-party imports
- local imports
with blank lines separating the sub-blocks. goconvey should be in the middle section.

---

*[go/lib/infra/disp/disp_test.go, line 35 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2V0HfnbX1bbJPaHcau:-L2V0HfnbX1bbJPaHcav:b-7zquuz) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/disp_test.go#L35)):*
> ```Go
> func Setup() (*Dispatcher, *Dispatcher, *customObject, *customObject) {
> 	a2b, b2a := p2p.New()
> 	dispA := New(messaging.NewRUDP(a2b, log.Root()), testAdapter, log.Root())
> ```

Passing in an actual logger makes `go test -v`'s output very hard to read. I suggest doing this in `init()`:
```
l := log.Root()
l.SetHandler(log.DiscardHandler())
```
If the logs are ever desired for debugging, then this can be commented out.

---

*[go/lib/infra/disp/disp_test.go, line 68 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2V-rP73shNGXB5G8ry:-L2V-rP73shNGXB5G8rz:bnuhbo4) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/disp_test.go#L68)):*
> ```Go
> 			SoMsg("a request err timeout", common.IsTimeoutErr(err), ShouldBeTrue)
> 			SoMsg("a request reply", recvReply, ShouldBeNil)
> 			fmt.Println(err)
> ```

Debugging line?

---

*[go/lib/infra/disp/wtable_test.go, line 33 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2V453UfvdvneB_xe0R:-L2V453UfvdvneB_xe0S:bkb6qk4) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/wtable_test.go#L33)):*
> ```Go
> 
> 		Convey("Normal request/reply", func() {
> 			// channel to force reply after request is added to table
> ```

It looks like it would be simpler to just move the `addRequest` line to just before `Convey("Parallel"...`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

scrye:

---

*[go/lib/infra/disp/disp.go, line 71 at r3](https://reviewable.io:443/reviews/scionproto/scion/1366#-L1wU4dikeLWewio6kCC:-L2ViuFmXTG03BConp8B:bfczfbd) ([raw file](https://github.com/scionproto/scion/blob/e50bf58255badb9ebb065083d607932aee7e5be3/go/lib/infra/disp/disp.go#L71)):*

Ah. So, i'd be a lot happier with `stoppedCh` than `stoppedC`. It's a bit hungarian-notationy for my taste, especially in a strongly-type language, but i'd be ok with it.
</blockquote></details>

From what I've noticed, they tend to use it for members where there's no good way to express that it's a channel (when I see `stopped`, I think of a `bool`; adding the suffix `Chan` is a bit more intuitive, even if it's in Hungarian notation style). I went with the full `Chan` suffix since I'm not against it, but I can change it to only `Ch` if you'd prefer.

---

*[go/lib/infra/disp/disp_test.go, line 1 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2V17WpDndFqXT2pxXK:-L2Vli8fwsXbkmAs0Dxh:b-5bq0jz) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/disp_test.go#L1)):*

All the timeouts in this test make it very slow to run - it takes 15s just for this file (out of 22s for all the go unit tests). Also, as all the tests are bundled into large test functions, it's also really difficult to see exactly how long each individual test takes (which makes it hard to confirm assumptions about where the delays are coming from). Having ~one function per test case would make this a lot easier to work with.
</blockquote></details>

The timeouts for contexts were way over what they needed to be. I've greatly decreased them making the test run much faster. Also split to smaller functions.

---

*[go/lib/infra/disp/disp_test.go, line 30 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2UzriXmMDeqDK3gaKq:-L2VlzehBY50TRkFDsv9:b3crr22) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/disp_test.go#L30)):*

The import block looks like:
- stdlib imports
- third-party imports
- local imports
with blank lines separating the sub-blocks. goconvey should be in the middle section.
</blockquote></details>

My `goimports` rewrite keeps messing up these and I never notice (even though it's set up with the order you suggest). I think it tries to conserve the order when I sometimes paste an import, which breaks stuff.

---

*[go/lib/infra/disp/disp_test.go, line 35 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2V0HfnbX1bbJPaHcau:-L2VnBpc6cTJW7aH77so:b1xf6tz) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/disp_test.go#L35)):*

Passing in an actual logger makes `go test -v`'s output very hard to read. I suggest doing this in `init()`:
```
l := log.Root()
l.SetHandler(log.DiscardHandler())
```
If the logs are ever desired for debugging, then this can be commented out.
</blockquote></details>

Done; also changed in `messaging/`.

---

*[go/lib/infra/disp/disp_test.go, line 68 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2V-rP73shNGXB5G8ry:-L2VnF4jYnGNdSf3tyHY:b-tm0elr) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/disp_test.go#L68)):*

Debugging line?
</blockquote></details>

Yup, removed.

---

*[go/lib/infra/disp/wtable_test.go, line 33 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2V453UfvdvneB_xe0R:-L2Vo3kvoyIidw2B78ql:bljk1tp) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/wtable_test.go#L33)):*

It looks like it would be simpler to just move the `addRequest` line to just before `Convey("Parallel"...`.
</blockquote></details>

Ah, right, good point. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

kormat:
Reviewed 6 of 6 files at r6.

---

*[go/lib/infra/disp/wtable_test.go, line 33 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2V453UfvdvneB_xe0R:-L2ZW17-TsYuthvl63UW:b-uj96yq) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/wtable_test.go#L33)):*

Ah, right, good point. Done.
</blockquote></details>

The same applies in other tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

scrye:
Review status: 12 of 13 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/infra/disp/wtable_test.go, line 33 at r5](https://reviewable.io:443/reviews/scionproto/scion/1366#-L2V453UfvdvneB_xe0R:-L2ZXHaNpT2U6fdAIscM:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/575483e30ff6e84df2ba78b7e08cfc0cbdba816e/go/lib/infra/disp/wtable_test.go#L33)):*

The same applies in other tests.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1366#-:-L2ZZCbDfw8NkYUmEg3v:bnfp4nl)*

PR #1368, opened 2017-12-22
API layer for SCION infra services

scrye:
Rebased on https://github.com/scionproto/scion/pull/1366.

Message send/receive API for SCION infrastructure processes. Only Cert and TRC messages are supported for now.

Also includes a generic server for infrastructure processes, and a Trust Cache stub that shows how message handlers can be created and used.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1368)


scrye:
This is submitted for reference purposes. High level documentation is available in the godocs for `go/lib/infra/api`, `go/lib/infra/server` and `go/lib/infra/components/trc`.

This is based on https://github.com/scionproto/scion/pull/1366. A lot of the changes Reviewable is showing are part of that PR and should be reviewed/merged first.

---

Review status: 0 of 19 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368#-:-L0yO-uu_oasifndj6DJ:b-g27g1s)*

kormat:
Reviewed 17 of 19 files at r1.
Review status: 17 of 19 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/infra/api/adapter.go, line 54 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-TeJMJE_wR_Zvxs0b:-L2-TeJMJE_wR_Zvxs0c:b-q3mvig) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/adapter.go#L54)):*
> ```Go
> 		return nil, common.NewCError("Unable to parse signed pld", "err", err)
> 	}
> 	return signedPld.Pld()
> ```

This throws away all signing information.

---

*[go/lib/infra/api/api.go, line 81 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RcamfFS_PitvY0NH:-L2-RcamfFS_PitvY0NI:b-98k1sb) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L81)):*
> ```Go
> 
> // RecvMsg reads a new message from the dispatcher.
> func (m *Messenger) RecvMsg(ctx context.Context) (interface{}, net.Addr, error) {
> ```

This should be `proto.Cerealizable`, not `interface{}` (same applies elsewhere).

---

*[go/lib/infra/api/api.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RstrrsFw3j-gPLc2:-L2-RstrrsFw3j-gPLc3:b-x5gr6q) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L86)):*
> ```Go
> 		return nil, nil, err
> 	}
> 	msg, err := extractMessage(ctrlPldMsg)
> ```

This approach breaks as soon as there's non-union data at any level, as it can't expose it. It also greatly complicates signature verification, as the caller has no access to the signature or the signed message. Doing this verification in `api` is also dubious, because api doesn't have enough context.

While it's more work for the caller, i think the only sustainable approach is to just return the `ctrl.SignedPld` as-is, and potentially provide some helper functions that the caller than use to operate on it.

---

*[go/lib/infra/api/api.go, line 96 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-SdihGwx28kb8510Q:-L2-SdihGwx28kb8510R:b8hejnz) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L96)):*
> ```Go
> // reply and returns the reply.
> func (m *Messenger) GetTRC(ctx context.Context, msg *cert_mgmt.TRCReq, a net.Addr) (*cert_mgmt.TRC, error) {
> 	ctrlPldMsg, err := ctrl.NewCertMgmtPld(msg)
> ```

Creating the `ctrl.Pld` inside `api` also greatly complicates signing of messages. As above, i think the only sustainable approach is to have the caller provide the fully-formed `ctrl.SignedPld`.

---

*[go/lib/infra/api/api.go, line 175 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-QxGXek38g8Y3kiHe:-L2-QxGXek38g8Y3kiHf:bkd9ng1) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L175)):*
> ```Go
> 	case proto.CtrlPld_Which_certMgmt:
> 		// Extract cert mgmt message
> 		innerPld := ctrlPld.CertMgmt
> ```

Maybe i'm missing something, but why not just `return ctrlPld.CertMgmt.Union()`?

---

*[go/lib/infra/components/trust/trust.go, line 15 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-To-eSO24Tr0OH8v2:-L2-To-eSO24Tr0OH8v3:b-roz7fq) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/components/trust/trust.go#L15)):*
> ```Go
> 
> package trust
> ```

This needs a module docstring.

---

*[go/lib/infra/server/server.go, line 17 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-U9wA01ORJKNebAaK:-L2-U9wA01ORJKNebAaL:bg9uvim) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/server/server.go#L17)):*
> ```Go
> // Package server contains a generic server framework for CtrlPld messages.
> //
> // A server runs on top of a *infra.Messenger:
> ```

Should this be `*api.Messenger`? If so, the name probably needs to be more fully qualified, as `api` is very generic.

---

*[go/lib/infra/server/server.go, line 29 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-UagU4uFVGjuBomYp:-L2-UagU4uFVGjuBomYq:b1mtnbz) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/server/server.go#L29)):*
> ```Go
> // process messages, handlers need to be registered. One handler can be
> // registered for each message type, identified by its capnp Proto ID:
> //  server.AddHandler(proto.TRCReq_TypeID, MyCustomHandlerConstructor)
> ```

This is not a valid namespace. A given proto can appear multiple times in the CtrlPld tree. E.g. PathMgmt's union has two members of type `SegRecs` (it used to have 3, but one of them was changed relatively recently).

Looking at how the `Union()` api looks for CtrlPld, this is actually already an issue there. We need to change it from:
`func (p *Pld) Union() (p proto.Cerealizable, err error) {`
to
`func (p *Pld) Union() (p proto.Cerealizable, which string, err error) {`
to allow callers to distinguish between multiple union entries of the same type. I guess we didn't find this before now because we haven't implemented the path server in go yet :)

So, the proper approach i think is:
```
server.AddHandler(proto.CtrlPld_Which_pcb, PCBHandlerConstructor)
server.AddHandlerNest(proto.CtrlPld_Which_pathMgmt, proto.PathMgmt_Which_segReg, SegRegHandler)
server.AddHandlerNest(proto.CtrlPld_Which_pathMgmt, proto.PathMgmt_Which_segSync, SegSyncHandler)
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368)*

scrye:
Review status: 17 of 19 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/infra/api/api.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RstrrsFw3j-gPLc2:-L2KA2BeFeSO1SEhvcVC:b9dp2wg) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L86)):*

This approach breaks as soon as there's non-union data at any level, as it can't expose it. It also greatly complicates signature verification, as the caller has no access to the signature or the signed message. Doing this verification in `api` is also dubious, because api doesn't have enough context.

While it's more work for the caller, i think the only sustainable approach is to just return the `ctrl.SignedPld` as-is, and potentially provide some helper functions that the caller than use to operate on it.
</blockquote></details>

The alternative of having it outside just creates another layer on top of the API, which  handles all the complexity (signing, verification, payload extraction and creation). At that point, the API isn't doing anything except forwarding function calls to the dispatcher.

The API is very simple right now. We can give it the crypto it needs to handle signatures and addresses.

The end goal here is to hide`ctrl.SignedPld` and all serialization/deserialization headaches from apps. They should only operate with native Go structs, and never care about anything else. All the capnproto stuff is very complex, and should be under the hood as much as possible.

---

*[go/lib/infra/api/api.go, line 96 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-SdihGwx28kb8510Q:-L2KBQp9GXHrunZWocYb:b-edh3k2) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L96)):*

Creating the `ctrl.Pld` inside `api` also greatly complicates signing of messages. As above, i think the only sustainable approach is to have the caller provide the fully-formed `ctrl.SignedPld`.
</blockquote></details>

As stated above, that means being aware of the messaging scheme and doing low level payload manipulation in the caller. I think we can hide everything under the hood, which will give us better decoupling and lead to very clean apps (e.g., the client/server example in `go/lib/infra/server/server_test.go`).

---

*[go/lib/infra/server/server.go, line 29 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-UagU4uFVGjuBomYp:-L2KCZCqcaQJ0t_FVDRu:b-1dk4jn) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/server/server.go#L29)):*

This is not a valid namespace. A given proto can appear multiple times in the CtrlPld tree. E.g. PathMgmt's union has two members of type `SegRecs` (it used to have 3, but one of them was changed relatively recently).

Looking at how the `Union()` api looks for CtrlPld, this is actually already an issue there. We need to change it from:
`func (p *Pld) Union() (p proto.Cerealizable, err error) {`
to
`func (p *Pld) Union() (p proto.Cerealizable, which string, err error) {`
to allow callers to distinguish between multiple union entries of the same type. I guess we didn't find this before now because we haven't implemented the path server in go yet :)

So, the proper approach i think is:
```
server.AddHandler(proto.CtrlPld_Which_pcb, PCBHandlerConstructor)
server.AddHandlerNest(proto.CtrlPld_Which_pathMgmt, proto.PathMgmt_Which_segReg, SegRegHandler)
server.AddHandlerNest(proto.CtrlPld_Which_pathMgmt, proto.PathMgmt_Which_segSync, SegSyncHandler)
```
</blockquote></details>

Ah, this is the exact reason why I flattened the namespace. The intention was to flatten everything so that apps don't need to care about such details.

From an app's point of view, we can simplify even more here. We can define _aliases_ (basically, app-friendly string names) for each message type (e.g., `SegSync` and `SegReg`), and each alias is associated to its own capnp unique name (so, `SegSync -> 'CtrlPld.PathMgmt.segSync`). Under the hood, the correct payload is constructed. This means apps never have to care about `proto.Xxx_Which_Xxx` (which already looks unintuitive for apps) or encapsulation hierarchies.

The API will mix and match between capnp stuff and Go stuff. So apps just register a handler for `SegRec`, and client just do `SendSegRec(segRec)`, and never care about anything else.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368)*

kormat:
Reviewed 1 of 19 files at r1.
Review status: 18 of 19 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/infra/api/api.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RstrrsFw3j-gPLc2:-L2KMAbLzJIeAd_hOXW5:bqnx663) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L86)):*
You can't separate them like that. E.g. for a `PCB`, it's illegal for the signer of the `ctrl.SignedPld` to differ from the signer of the last `ASEntry` inside the `PCB` (while it's perfectly legal for the signers to differ for `PathSegment` if it's inside CtrlPld.pathMgmt.segSync, for example). It should also be possible for an application to inspect the signers of a bunch of `ctrl.SignedPld` so it can do batch requests of certificates etc. Neither of these are possible if the api does not expose the `SignedPld`.

I don't think it's difficult to provide a few helper functions layered on top of `api` to handle `ctrl.SignedPld` that apps can use, and leave `api` simple, and deal take/return `ctrl.SignedPld`'s.

> This approach breaks as soon as there's non-union data at any level, as it can't expose it. 
This is still an issue ^.

---

*[go/lib/infra/api/api.go, line 96 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-SdihGwx28kb8510Q:-L2KQhdjhlUNBmIzdblw:b9mjmbj) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L96)):*
> As stated above, that means being aware of the messaging scheme
What do you mean? Capnp?

---

*[go/lib/infra/server/server.go, line 29 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-UagU4uFVGjuBomYp:-L2KTKGtVSan3Dt--9Lz:bm8ctcs) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/server/server.go#L29)):*
> From an app's point of view, we can simplify even more here. We can define aliases (basically, app-friendly string names) for each message type (e.g., SegSync and SegReg), and each alias is associated to its own capnp unique name (so, SegSync -> 'CtrlPld.PathMgmt.segSync). Under the hood, the correct payload is constructed. This means apps never have to care about proto.Xxx_Which_Xxx (which already looks unintuitive for apps) or encapsulation hierarchies.

Ok, then the api needs to change to not use `proto.ProtoIdType` to be useful. But this still has the issue that it can't handle non-union data.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368)*

scrye:
Review status: 18 of 19 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/infra/api/api.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RstrrsFw3j-gPLc2:-L2KoWSAXe1qEhIur_po:b7rdtr7) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L86)):*
Well, for more complex crypto requirements we can have the API take additional keys to/return additional values from the respective calls. We'll have a better idea of these as we start writing on top of the API, some things might change to be more expressive but I think the design is flexible enough to handle them.

If I'm understanding correctly, the main point is that signing/verification of `CtrlPld`s can't be fully transparent, so we need to also return crypto metadata in addition to the message itself. This can be done within the API, it just needs changes to method signatures (e.g., `return Message, Signer, error`).

> This approach breaks as soon as there's non-union data at any level, as it can't expose it.
I assumed there will always be some "top-level" message when exchanging messages between services.

---

*[go/lib/infra/api/api.go, line 96 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-SdihGwx28kb8510Q:-L2L3iPoxj3JjTQ6Ovy8:b-7dpemk) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L96)):*

> As stated above, that means being aware of the messaging scheme
What do you mean? Capnp?
</blockquote></details>

Yes, my intention is to hide away all the complexities of the schema -- in essence, have the infra API behave like a type-safe RPC with 0 serialization/deserialization awareness. This means hiding `proto` IDs, writes and parses. I think it would be great if server code looks like:
```
msg := &cert_mgmt.TRCReq{ISD: 42, Version: 1234, CacheOnly: true}
trc, err := messenger.GetTRC(ctx, msg, address)
```
like in `go/lib/infra/server/server_test.go`. Sure, there might be places where some glue code needs to be added, but I feel like net application complexity still goes down. It makes for code that is very easy to follow, and moves all the complexity behind the scenes leaving only application logic.

---

*[go/lib/infra/server/server.go, line 29 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-UagU4uFVGjuBomYp:-L2L5otE1FtyPzvJxaFk:b-doj57w) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/server/server.go#L29)):*

> From an app's point of view, we can simplify even more here. We can define aliases (basically, app-friendly string names) for each message type (e.g., SegSync and SegReg), and each alias is associated to its own capnp unique name (so, SegSync -> 'CtrlPld.PathMgmt.segSync). Under the hood, the correct payload is constructed. This means apps never have to care about proto.Xxx_Which_Xxx (which already looks unintuitive for apps) or encapsulation hierarchies.

Ok, then the api needs to change to not use `proto.ProtoIdType` to be useful. But this still has the issue that it can't handle non-union data.

</blockquote></details>

Yes, we'd probably change from `proto.ProtoIdType` to some prettier strings.

When you refer to non-union data, do you have something specific in mind? Like an actual non-union message that is now sent/received by Python infrastructure services?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368)*

kormat:
Review status: 18 of 19 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/api/api.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RstrrsFw3j-gPLc2:-L2LFWSBelPkHkQxzf3D:b4qofyi) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L86)):*
> If I'm understanding correctly, the main point is that signing/verification of CtrlPlds can't be fully transparent, so we need to also return crypto metadata in addition to the message itself. This can be done within the API, it just needs changes to method signatures (e.g., return Message, Signer, error).

I'm not sure how the semantics of this work. Here's a sketch of what i'm thinking of:

```
type verifyCallback func(data common.RawBytes, sign *proto.SignS)

func (m *Messenger) RecvMsg(ctx context.Context, verifier verifyCallback) (interface{}, data, net.Addr, error) {
  signedPld, address, err := m.dispatcher.RecvFrom(ctx)
  ...
  if err = verifier(signedPld.Blob, signedPld.Sign); err != nil {
    return signedPld.Pld(), signedPld.Sign, address, err
  }
  ...
  return signedPld.Pld(), signedPld.Sign, address, nil
}
```
The `verifyCallback` would need to figure out which key to use, and return a specific error if it isn't available. The caller of `RecvMsg` would have to check if there was an error and if the error means "you're missing the key", and if so use the signer info to get the key and... Oh. Then it would have to independently verify the signature anyway. It can't just call `RecvMsg` again, that gets a new message. So RecvMsg _must_ return a full signedPld as otherwise it's not possible to verify signatures if the key isn't already on hand. I don't think that `RecvMsg` is the right place to do this. The caller has to handle signature verification regardless.

---

*[go/lib/infra/api/api.go, line 96 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-SdihGwx28kb8510Q:-L2LIohcJ6QkcKy8dEWL:b-aokmz3) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L96)):*

Yes, my intention is to hide away all the complexities of the schema -- in essence, have the infra API behave like a type-safe RPC with 0 serialization/deserialization awareness. This means hiding `proto` IDs, writes and parses. I think it would be great if server code looks like:
```
msg := &cert_mgmt.TRCReq{ISD: 42, Version: 1234, CacheOnly: true}
trc, err := messenger.GetTRC(ctx, msg, address)
```
like in `go/lib/infra/server/server_test.go`. Sure, there might be places where some glue code needs to be added, but I feel like net application complexity still goes down. It makes for code that is very easy to follow, and moves all the complexity behind the scenes leaving only application logic.

</blockquote></details>

Hmm, ok, i see where you're going with this, but I'm not sure how much of the complexity it's possible to hide.

---

*[go/lib/infra/server/server.go, line 29 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-UagU4uFVGjuBomYp:-L2LCtMtiJsM9UbYPN_Z:b-i03mql) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/server/server.go#L29)):*
> Yes, we'd probably change from proto.ProtoIdType to some prettier strings.

Your phrasing sounds like this is sometime in the future. It should happen in _this_ PR.

> When you refer to non-union data, do you have something specific in mind? Like an actual non-union message that is now sent/received by Python infrastructure services?

Currently there is no non-union data under `CtrlPld`, but there is under `SIGCtrl` and `SCIONDMsg`. We have talked about putting a zipkin (or equivalent) field in `CtrlPld`. You had also suggested that CtrlPld extensions would live outside the union. If something like that happens down the line, the api surface area will have to grow for every change, which really sucks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368)*

scrye:
Review status: 18 of 19 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/api/api.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RstrrsFw3j-gPLc2:-L2LMj7mopvxhtk-fdI8:bwgpjhg) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L86)):*

> If I'm understanding correctly, the main point is that signing/verification of CtrlPlds can't be fully transparent, so we need to also return crypto metadata in addition to the message itself. This can be done within the API, it just needs changes to method signatures (e.g., return Message, Signer, error).

I'm not sure how the semantics of this work. Here's a sketch of what i'm thinking of:

```
type verifyCallback func(data common.RawBytes, sign *proto.SignS)

func (m *Messenger) RecvMsg(ctx context.Context, verifier verifyCallback) (interface{}, data, net.Addr, error) {
  signedPld, address, err := m.dispatcher.RecvFrom(ctx)
  ...
  if err = verifier(signedPld.Blob, signedPld.Sign); err != nil {
    return signedPld.Pld(), signedPld.Sign, address, err
  }
  ...
  return signedPld.Pld(), signedPld.Sign, address, nil
}
```
The `verifyCallback` would need to figure out which key to use, and return a specific error if it isn't available. The caller of `RecvMsg` would have to check if there was an error and if the error means "you're missing the key", and if so use the signer info to get the key and... Oh. Then it would have to independently verify the signature anyway. It can't just call `RecvMsg` again, that gets a new message. So RecvMsg _must_ return a full signedPld as otherwise it's not possible to verify signatures if the key isn't already on hand. I don't think that `RecvMsg` is the right place to do this. The caller has to handle signature verification regardless.
</blockquote></details>

I believe we're thinking at different abstraction layers. If the API doesn't do signatures and type encapsulation/decapsulation, it no longer does anything. It becomes an empty layer, and from that point on might as well call the dispatcher underneath directly.

I'm saying that all the complexity you reference (some of which I hadn't considered) _is_ this layer. This is the layer that makes app code simple and robust, and where all the complexity should happen. If `RecvMsg` needs to issue requests before returning the result to the app, I think we can let it do just that. We give the API a reference to a trust store object and let it construct certificate request handlers. The certificates are so essential to messaging that it makes sense to have this reference.

The API is customized for SCION Infra, and contains a good piece of the functionality of `scion_elem`, split into various objects like the trust store.

---

*[go/lib/infra/api/api.go, line 96 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-SdihGwx28kb8510Q:-L2LOpeqMAuGAis4WuMo:btuq2dk) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L96)):*

Hmm, ok, i see where you're going with this, but I'm not sure how much of the complexity it's possible to hide.
</blockquote></details>

I guess we'll see once we start implementing actual infrastructure pieces, but I'm pretty confident we can hide all of it.

---

*[go/lib/infra/server/server.go, line 29 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-UagU4uFVGjuBomYp:-L2LLI7sh7va0sKlahkg:bngwkrz) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/server/server.go#L29)):*
> Your phrasing sounds like this is sometime in the future. It should happen in this PR.

Ah, sorry I wasn't clearer. This PR isn't proposed for merging as it doesn't really implement anything. It's more a reference/proof of concept for how the API would work in the future. I mentioned this in the comment at the top, but granted, it's not exactly clear:

> This is submitted for reference purposes.

> Currently there is no non-union data under CtrlPld, but there is under SIGCtrl and SCIONDMsg.

This API is supposed to contain `CtrlPld` specific logic. `SigCtrl` and `SCIONDMsg` have no uses for it, and can call the dispatcher layer directly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368)*

shitz:
Review status: 18 of 19 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/api/api.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RstrrsFw3j-gPLc2:-L2LaBlq9V7vcFwbAQci:b-glndzl) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L86)):*

I believe we're thinking at different abstraction layers. If the API doesn't do signatures and type encapsulation/decapsulation, it no longer does anything. It becomes an empty layer, and from that point on might as well call the dispatcher underneath directly.

I'm saying that all the complexity you reference (some of which I hadn't considered) _is_ this layer. This is the layer that makes app code simple and robust, and where all the complexity should happen. If `RecvMsg` needs to issue requests before returning the result to the app, I think we can let it do just that. We give the API a reference to a trust store object and let it construct certificate request handlers. The certificates are so essential to messaging that it makes sense to have this reference.

The API is customized for SCION Infra, and contains a good piece of the functionality of `scion_elem`, split into various objects like the trust store.
</blockquote></details>

I'm with @scrye on this. In my mind, the API layer was the one implementing the entire verification logic, since that is by far the most complex part (in terms of messaging flow) for all the various lookups and is also shared between many/all infra elements. At least the common case should be hideable behind the API.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368)*

kormat:
Review status: 18 of 19 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/api/api.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RstrrsFw3j-gPLc2:-L2LfOnrrYvFTFAQpDR3:b-kswr11) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L86)):*
> I believe we're thinking at different abstraction layers. If the API doesn't do signatures and type encapsulation/decapsulation, it no longer does anything. It becomes an empty layer, and from that point on might as well call the dispatcher underneath directly.

Oh. Hum. My issue is that by trying to make this a nice tidy API, you're creating a lot of unnecessary complexity by pushing what is essentially application logic into the API.

> We give the API a reference to a trust store object and let it construct certificate request handlers.

This only works if the required certs/trcs are already present. You can't block reading messages from the network while waiting for a cert/trc reply. Hell, _those_ are messages that need to be read/processed. Doing signature verification has to be handled in an async fashion in general, otherwise it just doesn't work. An API like this cannot hide that behind a single call in a sane fashion.

Do you spawn off a goroutine for every cert you don't have? How long/frequently do you retry for? How can the application decide to give up on a given message? How does the app know how many outstanding certs/trcs it's waiting on, and how many messages are depending on those? It gets horrifically complex.

There's a lot of complexity in getting all the certs/trcs to verify a single PCB, for example. There's a lot of code in scion_elem dedicated to it. It's done asynchronously, with queues and retries, because that's the only way it can work.

> At least the common case should be hideable behind the API.

If the common case is defined to be one without any signatures, then yes. Otherwise, i can't see how.

---

*[go/lib/infra/server/server.go, line 29 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-UagU4uFVGjuBomYp:-L2LglyPOFbVwI_ia6GQ:b-da9ofh) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/server/server.go#L29)):*
> > Currently there is no non-union data under CtrlPld, but there is under SIGCtrl and SCIONDMsg.

> This API is supposed to contain CtrlPld specific logic. SigCtrl and SCIONDMsg have no uses for it, and can call the dispatcher layer directly.

You trimmed the rest of the paragraph that was specifically about CtrlPld...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368)*

shitz:
Review status: 18 of 19 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/api/api.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RstrrsFw3j-gPLc2:-L2VQgjJtNL2ujpXX_S8:b-u2dsml) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L86)):*

> I believe we're thinking at different abstraction layers. If the API doesn't do signatures and type encapsulation/decapsulation, it no longer does anything. It becomes an empty layer, and from that point on might as well call the dispatcher underneath directly.

Oh. Hum. My issue is that by trying to make this a nice tidy API, you're creating a lot of unnecessary complexity by pushing what is essentially application logic into the API.

> We give the API a reference to a trust store object and let it construct certificate request handlers.

This only works if the required certs/trcs are already present. You can't block reading messages from the network while waiting for a cert/trc reply. Hell, _those_ are messages that need to be read/processed. Doing signature verification has to be handled in an async fashion in general, otherwise it just doesn't work. An API like this cannot hide that behind a single call in a sane fashion.

Do you spawn off a goroutine for every cert you don't have? How long/frequently do you retry for? How can the application decide to give up on a given message? How does the app know how many outstanding certs/trcs it's waiting on, and how many messages are depending on those? It gets horrifically complex.

There's a lot of complexity in getting all the certs/trcs to verify a single PCB, for example. There's a lot of code in scion_elem dedicated to it. It's done asynchronously, with queues and retries, because that's the only way it can work.

> At least the common case should be hideable behind the API.

If the common case is defined to be one without any signatures, then yes. Otherwise, i can't see how.
</blockquote></details>

I don't see why the the API cannot spawn a separate go-routine to fetch the necessary TRCs/Certs (if needed) and do the verification. All the points you are making about the complexity, the app would also need to handle those as well. Of course it needs to be done asynchronously and with retries, but I don't see how the app is a better place for it than the API itself. If you want to support some rather exotic use-cases, e.g., you want to batch verify PCBs or not verify them at all, then separate API calls that skip the verification or return the signed payload instead of doing the verification can always be added (although I don't think this is necessary for the infra for which the API is made in the first place).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368)*

scrye:
Superseded by https://github.com/scionproto/scion/pull/1376.

---

Review status: 18 of 19 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/api/adapter.go, line 54 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-TeJMJE_wR_Zvxs0b:-L3XEcyndR0UxPd546_-:bbggyj7) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/adapter.go#L54)):*

This throws away all signing information.
</blockquote></details>

In https://github.com/scionproto/scion/pull/1376 the Adapter serializes/deserializes full Signed Payloads without verification, thus giving the higher layer complete control.

---

*[go/lib/infra/api/api.go, line 81 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RcamfFS_PitvY0NH:-L3XG8vnGexkW6Az_MFH:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L81)):*

This should be `proto.Cerealizable`, not `interface{}` (same applies elsewhere).
</blockquote></details>

Done.

---

*[go/lib/infra/api/api.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-RstrrsFw3j-gPLc2:-L3XGTymP5P8VPPX-x2f:b7sfgtw) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L86)):*

I don't see why the the API cannot spawn a separate go-routine to fetch the necessary TRCs/Certs (if needed) and do the verification. All the points you are making about the complexity, the app would also need to handle those as well. Of course it needs to be done asynchronously and with retries, but I don't see how the app is a better place for it than the API itself. If you want to support some rather exotic use-cases, e.g., you want to batch verify PCBs or not verify them at all, then separate API calls that skip the verification or return the signed payload instead of doing the verification can always be added (although I don't think this is necessary for the infra for which the API is made in the first place).
</blockquote></details>

@kormat : All retry timers can be enforced via `context.Context` deadlines. It's simple and intuitive -- lower layers should just retry until the context expires. From the point of view of the app, it just waits a maximum amount of time for some functionality, and if it doesn't happen, decides on that. From this perspective, each API call is a handler with a lifetime, and can spawn off goroutines that inherit that lifetime.

I think all verification use cases above can be handled by the current architecture proposed in https://github.com/scionproto/scion/pull/1376.

We also need proper state machines (at least in documentation) for these behaviors, to have a better idea of what the architecture should support. Only very complex architectures can support _everything_ (and to support some arcane app behaviors they usually pay for that in development time and maintenance costs). Also, coding complex protocols/timeouts and verification without state machines is stumbling in the dark.

If there are limitations in the approach proposed in https://github.com/scionproto/scion/pull/1376, we should discuss them separately in that commit.

---

*[go/lib/infra/api/api.go, line 175 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-QxGXek38g8Y3kiHe:-L3XIPBPVOfI8GQWpCEl:b-fjpmpu) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/api/api.go#L175)):*

Maybe i'm missing something, but why not just `return ctrlPld.CertMgmt.Union()`?
</blockquote></details>

This function is no longer part of the new PR.

---

*[go/lib/infra/components/trust/trust.go, line 15 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-To-eSO24Tr0OH8v2:-L3XI_tFqQUGIgCz0Oq7:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/components/trust/trust.go#L15)):*

This needs a module docstring.
</blockquote></details>

Done.

---

*[go/lib/infra/server/server.go, line 17 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-U9wA01ORJKNebAaK:-L3XIc6CWSCxdnzHI_jW:b-pxbobc) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/server/server.go#L17)):*

Should this be `*api.Messenger`? If so, the name probably needs to be more fully qualified, as `api` is very generic.
</blockquote></details>

The package has been renamed.

---

*[go/lib/infra/server/server.go, line 29 at r1](https://reviewable.io:443/reviews/scionproto/scion/1368#-L2-UagU4uFVGjuBomYp:-L3XIr3o-m4__wCvx2Sb:b-qzkde8) ([raw file](https://github.com/scionproto/scion/blob/037c5a303b0f3f43e3bd7ba5e88c536462c50c1f/go/lib/infra/server/server.go#L29)):*

> > Currently there is no non-union data under CtrlPld, but there is under SIGCtrl and SCIONDMsg.

> This API is supposed to contain CtrlPld specific logic. SigCtrl and SCIONDMsg have no uses for it, and can call the dispatcher layer directly.

You trimmed the rest of the paragraph that was specifically about CtrlPld...
</blockquote></details>

The API is very close to infra apps (it's basically the common infra app layer). Expanding it to support extensions/other capnp modifications should be ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1368#-:-L3XJ9CHF3oPvMNgfEMN:b-c1fqq5)*

scrye:
Superseded by https://github.com/scionproto/scion/pull/1376.

PR #1370, opened 2018-01-08
Fix trc req path bug.

kormat:
If a CS needs to send a trc request to a remote AS, it uses `ISD-0` as
the destination AS for the path lookup, which sciond supports. However,
the current code uses the same IA for the data-plane address, which is
illegal.

This patch adds two helpers to `FwdPathMeta` to allow an app to easily
determine the source/destination ASes, and updates the CS to use this
information.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1370)

### DIFF
--- a/python/cert_server/main.py
+++ b/python/cert_server/main.py
@@ -361,10 +361,10 @@ class CertServer(SCIONElement):
 
     def _send_trc_request(self, isd, ver):
         trc_req = TRCRequest.from_values(isd, ver, cache_only=True)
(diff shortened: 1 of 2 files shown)


shitz:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1370#-:-L2LXTy9_-_2pW-BsVCC:bnfp4nl)*

PR #1371, opened 2018-01-08
Add KeyConfig

Oncilla:
Add key configuration to go lib.
The key configuration holds:
- AS decryption key
- AS signing key
- AS online key (for core AS)
- AS offline key (for core AS)

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1371)

### DIFF
--- a/go/cert_srv/main.go
+++ b/go/cert_srv/main.go
@@ -51,6 +51,7 @@ var (
 	store    *trust.Store
 	bind     *snet.Addr
 	public   *snet.Addr
+	keyConf  *trust.KeyConf
 )
 
 // main initializes the certificate server and starts the dispatcher.
@@ -69,6 +70,9 @@ func main() {
 		fatal(err.Error())
 	}
 	if err = loadTopo(); err != nil {
+		fatal(err.Error())
+	}
+	if keyConf, err = trust.LoadKeyConf(filepath.Join(*confDir, "keys"), topo.Core); err != nil {
 		fatal(err.Error())
 	}
 	// initialize Trust Store
--- a/go/lib/trust/key_conf.go
+++ b/go/lib/trust/key_conf.go
@@ -1,0 +1,88 @@
+// Copyright 2018 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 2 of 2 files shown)


shitz:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/trust/key_conf.go, line 51 at r1](https://reviewable.io:443/reviews/scionproto/scion/1371#-L2ecr28I-61TXnIO_rP:-L2ecr28I-61TXnIO_rQ:b2g6kx8) ([raw file](https://github.com/scionproto/scion/blob/7fe902baf8632c3777e606f22b0a15617447d433/go/lib/trust/key_conf.go#L51)):*
> ```Go
> // LoadKeyConf loads key configuration from specified path. The loaded config is put into
> // CurrKeyConf.
> func LoadKeyConf(path string, core bool) (*KeyConf, error) {
> ```

Instead of passing a bool here `LoadKeyConf` should just check if a offline/online root key is present and if yes load it. The application using the keystore should then check itself whether there _must_ be an offline/online root key. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1371)*

Oncilla:
*[go/lib/trust/key_conf.go, line 51 at r1](https://reviewable.io:443/reviews/scionproto/scion/1371#-L2ecr28I-61TXnIO_rP:-L2elXCpffuSygADs48Y:b-iqmij6) ([raw file](https://github.com/scionproto/scion/blob/7fe902baf8632c3777e606f22b0a15617447d433/go/lib/trust/key_conf.go#L51)):*

Instead of passing a bool here `LoadKeyConf` should just check if a offline/online root key is present and if yes load it. The application using the keystore should then check itself whether there _must_ be an offline/online root key. 
</blockquote></details>

This seems a bit odd to me. How about we rename core to loadRootKeys instead.
Other wise, the client has to check for error and additionally do a nil check for both online and offline root keys

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1371)*

shitz:
Reviewed 2 of 2 files at r1.

---

*[go/lib/trust/key_conf.go, line 51 at r1](https://reviewable.io:443/reviews/scionproto/scion/1371#-L2ecr28I-61TXnIO_rP:-L3-11dQpHMizSBRBMX7:b2nhll9) ([raw file](https://github.com/scionproto/scion/blob/7fe902baf8632c3777e606f22b0a15617447d433/go/lib/trust/key_conf.go#L51)):*

This seems a bit odd to me. How about we rename core to loadRootKeys instead.
Other wise, the client has to check for error and additionally do a nil check for both online and offline root keys
</blockquote></details>

Ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1371)*

shitz:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1371#-:-L33PbvAzmssD7xk_qNT:bnfp4nl)*

PR #1379, opened 2018-01-22
Adds sciond request for segments used to construct paths

### LINKED ISSUE #1299: Add core and path type to sciond.get_paths
Currently sciond api users are not able to see if the ASes of the returned paths are core or not and what type of link the paths are traversing.
- Modify get_paths() to return the isCore boolean for each ISDAS returned.
- Modify get_paths() to return the link type (peer, parent, child, ...) for each interface pair returned.
- Potentially the topology file if the best place for this the daemon has access to?

mwfarb:
- Adds ability to request detailed segment data after get_paths succeeds.
- Returns requested segments data (core, up, down).
- Requested segment data can be used to calculate core ASes in paths.
- Requested segment data can be used to calculate path types.
- Useful for visualizations to display core and path types for every AS routed through.
- Closes #1299.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1379)


mwfarb:
@kormat @shitz Let me know if you have time to take a look at this or not. This adds get_segments() to sciond to solve some of the needs of our path visualizations and could be used for other apps as well.

mwfarb:
I need to setup another branch. And learn to read better. *shudders*

PR #1380, opened 2018-01-24
Add customer mapping to CS

Oncilla:
Add customer mapping during topology generation.

Core ASes are responsible for all ASes they issue a certificate for.
Core ASes store verifying keys associated with their customers to verify certificate reissuance requests.
The naming convention is ISDx-ASy-Vz.key, where z denotes the first certificate version of AS x-y which authenticates the key. The keys are stored in cs_config_dir/customers.
For each AS, the key with the highest associated version is loaded.
New key files are only written, when the verifying key changes.

If the core AS needs to remove a customer or change the verifying key manually (e.g. key compromise), the key files
are adjusted as desired and a SIG_HUP is sent to the CS-go process to force reload the customer mapping.

 Additionally:
  - move trust store from cert_srv/main to conf
  - wrap keyConfig in functions to guarantee synchronized acces

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1380)

### DIFF
--- a/go/cert_srv/cert.go
+++ b/go/cert_srv/cert.go
@@ -45,9 +45,9 @@ func (h *ChainHandler) HandleReq(addr *snet.Addr, req *cert_mgmt.ChainReq) {
 	log.Info("Received certificate chain request", "addr", addr, "req", req)
 	var chain *cert.Chain
 	if req.Version == cert_mgmt.NewestVersion {
-		chain = store.GetNewestChain(req.IA())
+		chain = config.Store.GetNewestChain(req.IA())
 	} else {
-		chain = store.GetChain(req.IA(), req.Version)
+		chain = config.Store.GetChain(req.IA(), req.Version)
 	}
 	srcLocal := config.PublicAddr.IA.Eq(addr.IA)
 	if chain != nil {
@@ -109,7 +109,7 @@ func (h *ChainHandler) HandleRep(addr *snet.Addr, rep *cert_mgmt.Chain) {
 	if err != nil {
 		log.Error("Unable to parse certificate reply", "err", err)
 	}
-	if err = store.AddChain(chain, true); err != nil {
+	if err = config.Store.AddChain(chain, true); err != nil {
 		log.Error("Unable to store certificate chain", "key", chain.Key(), "err", err)
 		return
 	}
--- a/go/cert_srv/conf/conf.go
+++ b/go/cert_srv/conf/conf.go
@@ -52,11 +52,11 @@ type Conf struct {
 	customers Customers
 	// customersLock guards the customers map.
 	customersLock sync.RWMutex
-	// CacheDir is the configuration directory.
+	// CacheDir is the cache directory.
 	CacheDir string
 	// ConfDir is the configuration directory.
 	ConfDir string
-	// StateDir is the configuration directory.
+	// StateDir is the state directory.
 	StateDir string
 }
 
@@ -66,7 +66,8 @@ func Load(id string, confDir string, cacheDir string, stateDir string) (*Conf, e
 	conf := &Conf{
 		ConfDir:  confDir,
 		CacheDir: cacheDir,
-		StateDir: stateDir}
+		StateDir: stateDir,
+	}
 	// load topology
 	path := filepath.Join(confDir, topology.CfgName)
 	if conf.Topo, err = topology.LoadFromFile(path); err != nil {
--- a/go/cert_srv/conf/conf.go
+++ b/go/cert_srv/conf/conf.go
@@ -16,6 +16,7 @@ package conf
 
 import (
 	"path/filepath"
+	"sync"
 
 	"github.com/scionproto/scion/go/lib/addr"
 	"github.com/scionproto/scion/go/lib/common"
@@ -25,9 +26,11 @@ import (
 )
 
 const (
-	ErrorTopo    = "Unable to load topology"
-	ErrorAddr    = "Unable to load addresses"
-	ErrorKeyConf = "Unable to load KeyConf"
+	ErrorAddr      = "Unable to load addresses"
+	ErrorKeyConf   = "Unable to load KeyConf"
+	ErrorStore     = "Unable to load TrustStore"
+	ErrorTopo      = "Unable to load topology"
+	ErrorCustomers = "Unable to load Customers"
 )
 
 type Conf struct {
@@ -38,14 +41,23 @@ type Conf struct {
 	BindAddr *snet.Addr
 	// PublicAddr is the public address.
 	PublicAddr *snet.Addr
-	// KeyConf contains the AS level keys used for signing and decrypting.
-	KeyConf *trust.KeyConf
+	// Store is the trust store.
+	Store *trust.Store
+	// keyConf contains the AS level keys used for signing and decrypting.
+	keyConf *trust.KeyConf
+	// keyConfLock guards KeyConf, CertVer and TRCVer.
+	keyConfLock sync.RWMutex
+	// customers is a mapping from non-core ASes assigned to this core AS to their public
+	// verifying key.
+	customers Customers
+	// customersLock guards the customers map.
+	customersLock sync.RWMutex
 	// Dir is the configuration directory.
 	Dir string
 }
 
 // Load initializes the configuration by loading it from confDir.
-func Load(id string, confDir string) (*Conf, error) {
+func Load(id string, confDir string, cacheDir string) (*Conf, error) {
 	var err error
 	conf := &Conf{Dir: confDir}
 	// load topology
@@ -68,10 +80,56 @@ func Load(id string, confDir string) (*Conf, error) {
 	if !tmpBind.EqAddr(conf.PublicAddr) {
 		conf.BindAddr = tmpBind
 	}
-	// load keyConf
-	path = filepath.Join(confDir, "keys")
-	if conf.KeyConf, err = trust.LoadKeyConf(path, conf.Topo.Core); err != nil {
+	// load trust store
+	conf.Store, err = trust.NewStore(filepath.Join(confDir, "certs"), cacheDir, id)
+	if err != nil {
+		return nil, common.NewBasicError(ErrorStore, err)
+	}
+	// load key configuration
+	if conf.keyConf, err = conf.loadKeyConf(); err != nil {
 		return nil, common.NewBasicError(ErrorKeyConf, err)
 	}
+	// load customers
+	if conf.customers, err = conf.LoadCustomers(); err != nil {
+		return nil, err
+	}
 	return conf, nil
+}
+
+// ReloadCustomers reloads the mapping from customer to verifying key.
+func (c *Conf) ReloadCustomers() error {
+	c.customersLock.Lock()
+	defer c.customersLock.Unlock()
+	cust, err := c.LoadCustomers()
+	if err != nil {
+		return common.NewBasicError(ErrorCustomers, err)
+	}
+	c.customers = cust
+	return nil
+}
+
+// loadKeyConf loads key configuration.
+func (c *Conf) loadKeyConf() (*trust.KeyConf, error) {
+	return trust.LoadKeyConf(filepath.Join(c.Dir, "keys"), c.Topo.Core)
+}
+
+// GetSigningKey returns the signing key of the current key configuration.
+func (c *Conf) GetSigningKey() common.RawBytes {
+	c.keyConfLock.RLock()
+	defer c.keyConfLock.RUnlock()
+	return c.keyConf.SignKey
+}
+
+// GetDecryptKey returns the decryption key of the current key configuration.
+func (c *Conf) GetDecryptKey() common.RawBytes {
+	c.keyConfLock.RLock()
+	defer c.keyConfLock.RUnlock()
+	return c.keyConf.DecryptKey
+}
+
+// GetOnRootKey returns the online root key of the current key configuration.
+func (c *Conf) GetOnRootKey() common.RawBytes {
+	c.keyConfLock.RLock()
+	defer c.keyConfLock.RUnlock()
+	return c.keyConf.OnRootKey
 }
--- a/go/cert_srv/conf/conf.go
+++ b/go/cert_srv/conf/conf.go
@@ -52,14 +52,21 @@ type Conf struct {
 	customers Customers
 	// customersLock guards the customers map.
 	customersLock sync.RWMutex
-	// Dir is the configuration directory.
-	Dir string
+	// CacheDir is the configuration directory.
+	CacheDir string
+	// ConfDir is the configuration directory.
+	ConfDir string
+	// StateDir is the configuration directory.
+	StateDir string
 }
 
 // Load initializes the configuration by loading it from confDir.
-func Load(id string, confDir string, cacheDir string) (*Conf, error) {
+func Load(id string, confDir string, cacheDir string, stateDir string) (*Conf, error) {
 	var err error
-	conf := &Conf{Dir: confDir}
+	conf := &Conf{
+		ConfDir:  confDir,
+		CacheDir: cacheDir,
+		StateDir: stateDir}
 	// load topology
 	path := filepath.Join(confDir, topology.CfgName)
 	if conf.Topo, err = topology.LoadFromFile(path); err != nil {
@@ -98,6 +105,7 @@ func Load(id string, confDir string, cacheDir string) (*Conf, error) {
 
 // ReloadCustomers reloads the mapping from customer to verifying key.
 func (c *Conf) ReloadCustomers() error {
+	// Makes sure no new files can be written by SetVerifyingKey in the meantime
 	c.customersLock.Lock()
 	defer c.customersLock.Unlock()
 	cust, err := c.LoadCustomers()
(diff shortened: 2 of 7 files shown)


kormat:
Reviewed 3 of 6 files at r1, 4 of 4 files at r2.

---

*[go/cert_srv/main.go, line 122 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MhkIGWPaGTJdmFVu4:-L6MhkIGWPaGTJdmFVu5:blu4ddc) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/main.go#L122)):*
> ```Go
> 		os.Exit(1)
> 	}()
> 	sighup := make(chan os.Signal, 1)
> ```

This is racey - if SIGHUP arrives just after startup but before the signal handler is registered, it will cause the process to exit. It's not possible to fully avoid this, but have a look at `go/border/router.go:41` to see the best way i've found to make that window as small as possible.

---

*[go/cert_srv/conf/customers.go, line 41 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6Mc57KD1cbQLi6TShc:-L6Mc57KD1cbQLi6TShd:b6dhkl2) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L41)):*
> ```Go
> func (c *Conf) LoadCustomers() (Customers, error) {
> 	// Makes sure no new files can be written by SetVerifyingKey in the meantime
> 	path := filepath.Join(c.Dir, "customers")
> ```

I'm trying to move us away from writing to the `gen` folder. Ideally a process would have:
- a config dir: read-only, only updated by external processes, think /etc/blah
- a cache dir: writeable, wipeable but persistent across boots, think /var/cache/blah
- a state dir:  writeable, persistent, think /var/lib/blah
- a logs dir: think /var/log/blah
Currently we can configure the config/cache/logs dirs, but not a state dir. This use-case falls clearly into the state dir category. So, could you add a `-state-dir` flag (that defaults to the config dir if not set), and put `"customers"` inside it, please?

---

*[go/cert_srv/conf/customers.go, line 50 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MetpYAa3ExlN9y72o:-L6MetpYAa3ExlN9y72p:bkl0nxw) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L50)):*
> ```Go
> 	for _, file := range files {
> 		re := regexp.MustCompile(`ISD(\d+)-AS(\d+)-V(\d+)`)
> 		s := re.FindAllStringSubmatch(file, -1)
> ```

For this kind of regex, i think you can use the simpler `FindStringSubmatch` method.

---

*[go/cert_srv/conf/customers.go, line 62 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MfZ0NtH7Wztkvjr9j:-L6MfZ0NtH7Wztkvjr9k:b4v97sr) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L62)):*
> ```Go
> 	}
> 	customers := make(map[addr.ISD_AS]common.RawBytes)
> 	for ia, file := range activeKeys {
> ```

One thing we'll need to consider at some point is not loading keys into memory unless they're needed, and clearing them when they're not needed any more.

---

*[go/cert_srv/conf/customers.go, line 67 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MfSlSGalWI9fnM1ST:-L6MfSlTenm6zlQ5qMao:b5xv43x) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L67)):*
> ```Go
> 			return nil, err
> 		}
> 		key, err := base64.StdEncoding.DecodeString(string(b))
> ```

`.Decode(b)`?

---

*[go/cert_srv/conf/customers.go, line 76 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MgHDFcxND3dWLWftD:-L6MgHDFcxND3dWLWftE:bg0s6o0) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L76)):*
> ```Go
> }
> 
> // GetVerifyingKey returns the verifying key from the requested AS and true if it is in the mapping.
> ```

This doesn't match the return types.

---

*[go/cert_srv/conf/customers.go, line 103 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MgbSTTfviTwevcqJe:-L6MgbSUnkHCX8RXQOXw:b-gom0wf) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L103)):*
> ```Go
> 	if !bytes.Equal(newKey, currKey) {
> 		var err error
> 		path := filepath.Join(c.Dir, fmt.Sprintf("ISD%d-AS%d-V%d.key", ia.I, ia.A, ver))
> ```

This is not writing to the same dir as the above method reads from - no `customers` subdir.

---

*[go/cert_srv/conf/customers.go, line 105 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6Mgs1dEx7S-rMUq9cK:-L6Mgs1dEx7S-rMUq9cL:b-3hw0xa) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L105)):*
> ```Go
> 		path := filepath.Join(c.Dir, fmt.Sprintf("ISD%d-AS%d-V%d.key", ia.I, ia.A, ver))
> 		if _, err = os.Stat(path); os.IsNotExist(err) {
> 			b := base64.StdEncoding.EncodeToString(newKey)
> ```

`Encode(newKey)`?

---

*[go/cert_srv/conf/customers.go, line 114 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6Mh0_W6l5e_nnpo_Fu:-L6Mh0_W6l5e_nnpo_Fv:bondj6k) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L114)):*
> ```Go
> 			return nil
> 		}
> 		return err
> ```

The Go convention is to handle errors first. So you'd have:
```
if _, err = os.Stat(path); !os.IsNotExist(err) {
   return err
}
b := base64.StdEncoding...
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1380)*

Oncilla:
Review status: 3 of 7 files reviewed at latest revision, 9 unresolved discussions.

---

*[go/cert_srv/main.go, line 122 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MhkIGWPaGTJdmFVu4:-L6QXKrDsOEU-7ki8M1Q:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/main.go#L122)):*

This is racey - if SIGHUP arrives just after startup but before the signal handler is registered, it will cause the process to exit. It's not possible to fully avoid this, but have a look at `go/border/router.go:41` to see the best way i've found to make that window as small as possible.
</blockquote></details>

Done.

---

*[go/cert_srv/conf/customers.go, line 41 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6Mc57KD1cbQLi6TShc:-L6Qb5jDIy6DJlEECEep:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L41)):*

I'm trying to move us away from writing to the `gen` folder. Ideally a process would have:
- a config dir: read-only, only updated by external processes, think /etc/blah
- a cache dir: writeable, wipeable but persistent across boots, think /var/cache/blah
- a state dir:  writeable, persistent, think /var/lib/blah
- a logs dir: think /var/log/blah
Currently we can configure the config/cache/logs dirs, but not a state dir. This use-case falls clearly into the state dir category. So, could you add a `-state-dir` flag (that defaults to the config dir if not set), and put `"customers"` inside it, please?
</blockquote></details>

Done.

---

*[go/cert_srv/conf/customers.go, line 50 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MetpYAa3ExlN9y72o:-L6Qb7CJA0aBqnyZG8DT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L50)):*

For this kind of regex, i think you can use the simpler `FindStringSubmatch` method.
</blockquote></details>

Done.

---

*[go/cert_srv/conf/customers.go, line 62 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MfZ0NtH7Wztkvjr9j:-L6Qeua1TJkTtKqOrVsW:b-q04gpg) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L62)):*

One thing we'll need to consider at some point is not loading keys into memory unless they're needed, and clearing them when they're not needed any more.
</blockquote></details>

Right. Maybe they can be database backed in the future.

---

*[go/cert_srv/conf/customers.go, line 67 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MfSlSGalWI9fnM1ST:-L6QfW1q8pWo5KSXquRg:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L67)):*

`.Decode(b)`?
</blockquote></details>

Done.

---

*[go/cert_srv/conf/customers.go, line 76 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MgHDFcxND3dWLWftD:-L6QgU09aOf3taIBcR1P:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L76)):*

This doesn't match the return types.
</blockquote></details>

Done.

---

*[go/cert_srv/conf/customers.go, line 103 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6MgbSTTfviTwevcqJe:-L6QgYTkF1PqErFG6scm:be421sj) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L103)):*

This is not writing to the same dir as the above method reads from - no `customers` subdir.
</blockquote></details>

Good catch.
Done.

---

*[go/cert_srv/conf/customers.go, line 105 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6Mgs1dEx7S-rMUq9cK:-L6QhggJnLLKoLmF44rE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L105)):*

`Encode(newKey)`?
</blockquote></details>

Done.

---

*[go/cert_srv/conf/customers.go, line 114 at r2](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6Mh0_W6l5e_nnpo_Fu:-L6Qhh8IhmNnU7kCncnK:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/56620a21eab27ec14e1425c390b7d1af4ca155b3/go/cert_srv/conf/customers.go#L114)):*

The Go convention is to handle errors first. So you'd have:
```
if _, err = os.Stat(path); !os.IsNotExist(err) {
   return err
}
b := base64.StdEncoding...
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1380)*

kormat:
Reviewed 4 of 4 files at r3.

---

*[go/cert_srv/conf/conf.go, line 59 at r3](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6QoXl4WTeCQ4UMQZnY:-L6QoXl4WTeCQ4UMQZnZ:b3ekuqo) ([raw file](https://github.com/scionproto/scion/blob/7cd4bcb2ca46d0842e5b8ad8df6b4d541ecd881b/go/cert_srv/conf/conf.go#L59)):*
> ```Go
> 	// ConfDir is the configuration directory.
> 	ConfDir string
> 	// StateDir is the configuration directory.
> ```

No, it isn't :P

---

*[go/cert_srv/conf/conf.go, line 69 at r3](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6QoaSiAw9SToqjDCWg:-L6QoaSjVNKlvqc_aV7M:baj28l1) ([raw file](https://github.com/scionproto/scion/blob/7cd4bcb2ca46d0842e5b8ad8df6b4d541ecd881b/go/cert_srv/conf/conf.go#L69)):*
> ```Go
> 		ConfDir:  confDir,
> 		CacheDir: cacheDir,
> 		StateDir: stateDir}
> ```

In a multi-line literal like this, move the closing `}` to a new line.

---

*[go/cert_srv/conf/customers.go, line 110 at r3](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6QpHeSzjFUPrGU6AqL:-L6QpHeSzjFUPrGU6AqM:b-pxw76y) ([raw file](https://github.com/scionproto/scion/blob/7cd4bcb2ca46d0842e5b8ad8df6b4d541ecd881b/go/cert_srv/conf/customers.go#L110)):*
> ```Go
> 		buf := make([]byte, base64.StdEncoding.EncodedLen(len(newKey)))
> 		base64.StdEncoding.Encode(buf, newKey)
> 		if err = ioutil.WriteFile(path, buf, 0644); err != nil {
> ```

This should be 0600. This is secret key material, right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1380)*

Oncilla:

---

*[go/cert_srv/conf/conf.go, line 59 at r3](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6QoXl4WTeCQ4UMQZnY:-L6Qqks0BMOm9kVBMDjN:bk4pd1j) ([raw file](https://github.com/scionproto/scion/blob/7cd4bcb2ca46d0842e5b8ad8df6b4d541ecd881b/go/cert_srv/conf/conf.go#L59)):*

No, it isn't :P
</blockquote></details>

copy-paste ftw.
Done.

---

*[go/cert_srv/conf/conf.go, line 69 at r3](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6QoaSiAw9SToqjDCWg:-L6QqpknH5hvGzSbykq_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/7cd4bcb2ca46d0842e5b8ad8df6b4d541ecd881b/go/cert_srv/conf/conf.go#L69)):*

In a multi-line literal like this, move the closing `}` to a new line.
</blockquote></details>

Done.

---

*[go/cert_srv/conf/customers.go, line 110 at r3](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6QpHeSzjFUPrGU6AqL:-L6QqquvxwfgtIJ66Yvi:bq2yi7s) ([raw file](https://github.com/scionproto/scion/blob/7cd4bcb2ca46d0842e5b8ad8df6b4d541ecd881b/go/cert_srv/conf/customers.go#L110)):*

This should be 0600. This is secret key material, right?
</blockquote></details>

Actually, it is the verifying key (public value).
Should we change the file ending to make this more clear (e.g. `.vkey`)

In the case of key compromise, the customer will notify the core AS out-of-band, and a new verifying key is provided.
Using sighup the mapping is updated.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1380)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r4.

---

*[go/cert_srv/conf/customers.go, line 110 at r3](https://reviewable.io:443/reviews/scionproto/scion/1380#-L6QpHeSzjFUPrGU6AqL:-L6Qx-jTGYRNJ0GwOPeB:b9qzs4t) ([raw file](https://github.com/scionproto/scion/blob/7cd4bcb2ca46d0842e5b8ad8df6b4d541ecd881b/go/cert_srv/conf/customers.go#L110)):*

Actually, it is the verifying key (public value).
Should we change the file ending to make this more clear (e.g. `.vkey`)

In the case of key compromise, the customer will notify the core AS out-of-band, and a new verifying key is provided.
Using sighup the mapping is updated.

</blockquote></details>

It might be a good idea to stop using `.key` as a suffix, and instead of `.pubkey` and `.privkey`, because "key" is utterly ambiguous. But in any case, we can leave that for a future PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1380#-:-L6QxB2fQXOdmf8TRkl7:bnfp4nl)*

kormat:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1380)*

PR #1382, opened 2018-01-25
Fix 2 issues with sciond revocation

kormat:
1. _remove_revoked_pcbs() now only removes segments if they have
   revoked upstream PCBMarking.
2. handle_revocation now caches directly-received revocations instead of
   just using them once and then dropping them.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1382)

### DIFF
--- a/python/sciond/sciond.py
+++ b/python/sciond/sciond.py
@@ -392,6 +392,7 @@ class SCIONDaemon(SCIONElement):
                 % (rev_info.p.epoch, ConnectedHashTree.get_current_epoch()))
             return SCIONDRevReplyStatus.STALE
 
+        self.peer_revs.add(rev_info)
         # Go through all segment databases and remove affected segments.
         removed_up = self._remove_revoked_pcbs(self.up_segments, rev_info)
         removed_core = self._remove_revoked_pcbs(self.core_segments, rev_info)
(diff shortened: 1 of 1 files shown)


shitz:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[python/sciond/sciond.py, line 395 at r1](https://reviewable.io:443/reviews/scionproto/scion/1382#-L3gtJMM_RpQKlGmVJwz:-L3gtJMM_RpQKlGmVJx-:b-niauv4) ([raw file](https://github.com/scionproto/scion/blob/4480cbefcb7c81438804b0ae0d8b6bfcbf5dc5b9/python/sciond/sciond.py#L395)):*
> ```Python
>             return SCIONDRevReplyStatus.STALE
> 
>         self.peer_revs.add(rev_info)
> ```

As briefly discussed offline, `peer_revs` should become a general revocation cache that stores all revocations received by SCIOND. In `handle_path_reply` each segment should be checked against existing revocations and not added should it contain a revoked up/down-hop interface. If we extend `RevInfo` to contain a flag indicating whether the revocation is for a peering interface or not, we can have a `peer_rev` cache and one for non-peer revocations, which would make the validation cheaper.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1382)*

kormat:

---

*[python/sciond/sciond.py, line 395 at r1](https://reviewable.io:443/reviews/scionproto/scion/1382#-L3gtJMM_RpQKlGmVJwz:-L46WxDdvj1U4KK3yRBy:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/4480cbefcb7c81438804b0ae0d8b6bfcbf5dc5b9/python/sciond/sciond.py#L395)):*

As agreed, will be done in the rewrite.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1382)*

PR #1383, opened 2018-01-25
Adds sciond request for segments used to construct paths

### LINKED ISSUE #1299: Add core and path type to sciond.get_paths

mwfarb:
- Adds ability to request detailed segment data after get_paths
succeeds.
- Returns requested segments data (core, up, down).
- Requested segment data can be used to calculate core ASes in paths.
- Requested segment data can be used to calculate path types.
- Useful for visualizations to display core and path types for every AS
routed through.
- Closes #1299.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1383)

### DIFF
--- a/proto/sciond.capnp
+++ b/proto/sciond.capnp
@@ -4,6 +4,7 @@ $Go.package("proto");
 $Go.import("github.com/scionproto/scion/go/proto");
 
 using RevInfo = import "rev_info.capnp";
+using PSeg = import "path_seg.capnp";
 
 struct SCIONDMsg {
     id @0 :UInt64;  # Request ID
@@ -19,6 +20,8 @@ struct SCIONDMsg {
         serviceInfoRequest @9 :ServiceInfoRequest;
         serviceInfoReply @10 :ServiceInfoReply;
         revReply @11 :RevReply;
+        segTypeHopReq @12 :SegTypeHopReq;
+        segTypeHopReply @13 :SegTypeHopReply;
     }
 }
 
@@ -116,4 +119,18 @@ struct ServiceInfoReplyEntry {
     serviceType @0 :ServiceInfoRequest.ServiceType;  # The service ID of the service.
     ttl @1 :UInt32;  # The TTL for the service record in seconds (currently unused).
     hostInfos @2 :List(HostInfo);  # The host infos of the service.
+}
+
+struct SegTypeHopReq {
+    type @0 :PSeg.PathSegType;  # The path segments type: up, down, core.
+}
+
+struct SegTypeHopReply {
+    entries @0 :List(SegTypeHopReplyEntry);  # List of path segments matching type request, if any
+}
+
+struct SegTypeHopReplyEntry {
+    interfaces @0 :List(PathInterface);  # List of interfaces for the segment
+    timestamp @1 :UInt64;  # Creation timestamp, seconds since Unix Epoch
+    expTime @2 :UInt64;  # Expiration timestamp, seconds since Unix Epoch
 }
--- a/python/lib/app/sciond.py
+++ b/python/lib/app/sciond.py
@@ -42,9 +42,14 @@ from lib.sciond_api.host_info import HostInfo
 from lib.sciond_api.if_req import SCIONDIFInfoRequest
 from lib.sciond_api.path_req import SCIONDPathReplyError, SCIONDPathRequest
 from lib.sciond_api.revocation import SCIONDRevNotification
+from lib.sciond_api.segment_req import SCIONDSegTypeHopRequest
 from lib.sciond_api.service_req import SCIONDServiceInfoRequest
 from lib.socket import ReliableSocket
-from lib.types import AddrType, SCIONDMsgType as SMT
+from lib.types import (
+    AddrType,
+    PathSegmentType as PST,
+    SCIONDMsgType as SMT,
+)
 
 
 # TTL for the ASInfo object (1 hour)
@@ -196,6 +201,17 @@ class SCIONDConnector:
         if if_id:
             return self._resolve_ifid(if_id)
         return self._resolve_dst_addr(spkt.addrs.src, spkt.addrs.dst)
+
+    def get_segtype_hops(self, seg_type):
+        req_id = self._req_id.inc()
+        valid_pst = set([PST.CORE, PST.UP, PST.DOWN])
+        assert seg_type in valid_pst, 'PathSegmentType %s is unrecognized.' % seg_type
+        request = SCIONDMsg(SCIONDSegTypeHopRequest.from_values(seg_type), req_id)
+        with closing(self._create_socket()) as socket:
+            if not socket.send(request.pack()):
+                raise SCIONDRequestError
+            response = self._get_response(socket, req_id, SMT.SEGTYPEHOP_REPLY)
+            return list(response.iter_entries())
 
     def _resolve_ifid(self, if_id):  # pragma: no cover
         if_infos = self.get_if_info([if_id])
@@ -394,3 +410,18 @@ def send_rev_notification(rev_info, connector=None):  # pragma: no cover
     if not connector:
         raise SCIONDLibNotInitializedError
     connector.send_rev_notification(rev_info)
+
+
+def get_segtype_hops(seg_type, connector=None):  # pragma: no cover
+    """
+    Request list of segments by type used to construct paths.
+
+    :param seg_type: The type of PathSegmentType requested.
+    :returns: List of SCIONDSegTypeHopReplyEntry objects.
+    """
+    global _connector
+    if not connector:
+        connector = _connector
+    if not connector:
+        raise SCIONDLibNotInitializedError
+    return connector.get_segtype_hops(seg_type)
--- a/python/lib/sciond_api/base.py
+++ b/python/lib/sciond_api/base.py
@@ -26,6 +26,7 @@ from lib.sciond_api.as_req import SCIONDASInfoRequest, SCIONDASInfoReply
 from lib.sciond_api.if_req import SCIONDIFInfoReply, SCIONDIFInfoRequest
 from lib.sciond_api.path_req import SCIONDPathRequest, SCIONDPathReply
 from lib.sciond_api.revocation import SCIONDRevNotification, SCIONDRevReply
+from lib.sciond_api.segment_req import SCIONDSegTypeHopRequest, SCIONDSegTypeHopReply
 from lib.sciond_api.service_req import SCIONDServiceInfoReply, SCIONDServiceInfoRequest
 from lib.types import SCIONDMsgType
 
@@ -44,6 +45,8 @@ class SCIONDMsg(CerealBox):  # pragma: no cover
         SCIONDIFInfoReply: SCIONDMsgType.IF_REPLY,
         SCIONDServiceInfoRequest: SCIONDMsgType.SERVICE_REQUEST,
         SCIONDServiceInfoReply: SCIONDMsgType.SERVICE_REPLY,
+        SCIONDSegTypeHopRequest: SCIONDMsgType.SEGTYPEHOP_REQUEST,
+        SCIONDSegTypeHopReply: SCIONDMsgType.SEGTYPEHOP_REPLY,
     }
 
     def __init__(self, union, id):
--- a/python/lib/sciond_api/segment_req.py
+++ b/python/lib/sciond_api/segment_req.py
@@ -1,0 +1,95 @@
+# Copyright 2018 ETH Zurich
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+:mod:`segment_req` --- SCIOND segments by type requests and replies
+====================================================
+"""
+# External
+import capnp  # noqa
+from datetime import timedelta
+
+# SCION
+import proto.sciond_capnp as P
+from lib.packet.packet_base import Cerealizable
+from lib.sciond_api.path_meta import PathInterface
+from lib.util import iso_timestamp
+
+
+class SCIONDSegTypeHopRequest(Cerealizable):
+    NAME = "SCIONDSegTypeHopRequest"
+    P_CLS = P.SegTypeHopReq
+
+    @classmethod
+    def from_values(cls, seg_type):
+        p = cls.P_CLS.new_message(type=seg_type)
+        return cls(p)
+
+    def short_desc(self):
+        return "%s: type: %s" % (self.NAME, self.p.type)
+
+
+class SCIONDSegTypeHopReply(Cerealizable):
+    NAME = "SCIONDSegTypeHopReply"
+    P_CLS = P.SegTypeHopReply
+
+    @classmethod
+    def from_values(cls, entries):
+        p = cls.P_CLS.new_message()
+        entry_list = p.init("entries", len(entries))
+        for i, entry in enumerate(entries):
+            entry_list[i] = entry.p
+        return cls(p)
+
+    def entry(self, idx):
+        return SCIONDSegTypeHopReplyEntry(self.p.entries[idx])
+
+    def iter_entries(self):
+        for entry in self.p.entries:
+            yield SCIONDSegTypeHopReplyEntry(entry)
+
+    def short_desc(self):
+        return "\n".join([entry.short_desc() for entry in self.iter_entries()])
+
+
+class SCIONDSegTypeHopReplyEntry(Cerealizable):
+    NAME = "SCIONDSegTypeHopReplyEntry"
+    P_CLS = P.SegTypeHopReplyEntry
+
+    @classmethod
+    def from_values(cls, interfaces, timestamp, expTime):
+        p = cls.P_CLS.new_message(timestamp=timestamp, expTime=expTime)
+        ifs = p.init("interfaces", len(interfaces))
+        for i, if_ in enumerate(interfaces):
+            ifs[i] = if_.p
+        return cls(p)
+
+    def iter_ifs(self):
+        for if_ in self.p.interfaces:
+            yield PathInterface(if_)
+
+    def short_desc(self):
+        desc = []
+        remain = self.p.expTime - self.p.timestamp
+        desc.append("%s, %s, " % (iso_timestamp(
+            self.p.timestamp), timedelta(seconds=remain)))
+        desc.append(", ".join([if_.short_desc() for if_ in self.iter_ifs()]))
+        return "".join(desc)
+
+    def __str__(self):
+        desc = ["%s:" % self.NAME]
+        if_str = ", ".join([if_.short_desc() for if_ in self.iter_ifs()])
+        desc.append("  Interfaces: %s " % if_str)
+        desc.append("  Timestamp: %s" % self.p.timestamp)
+        desc.append("  Expiration: %s" % self.p.expTime)
+        return "\n".join(desc)
--- a/python/lib/types.py
+++ b/python/lib/types.py
@@ -196,6 +196,8 @@ class SCIONDMsgType(TypeBase):
     SERVICE_REPLY = "serviceInfoReply"
     DRKEY_REQUEST = "drkeyRequest"
     DRKEY_REPLY = "drkeyReply"
+    SEGTYPEHOP_REQUEST = "segTypeHopReq"
+    SEGTYPEHOP_REPLY = "segTypeHopReply"
 
 
 #######################
--- a/python/sciond/sciond.py
+++ b/python/sciond/sciond.py
@@ -52,7 +52,7 @@ from lib.sciond_api.revocation import SCIONDRevReply, SCIONDRevReplyStatus
 from lib.sciond_api.host_info import HostInfo
 from lib.sciond_api.if_req import SCIONDIFInfoReply, SCIONDIFInfoReplyEntry, SCIONDIFInfoRequest
 from lib.sciond_api.base import SCIONDMsg
-from lib.sciond_api.path_meta import FwdPathMeta
+from lib.sciond_api.path_meta import FwdPathMeta, PathInterface
 from lib.sciond_api.path_req import (
     SCIONDPathRequest,
     SCIONDPathReplyError,
@@ -60,6 +60,11 @@ from lib.sciond_api.path_req import (
     SCIONDPathReplyEntry,
 )
 from lib.sciond_api.revocation import SCIONDRevNotification
+from lib.sciond_api.segment_req import (
+    SCIONDSegTypeHopReply,
+    SCIONDSegTypeHopReplyEntry,
+    SCIONDSegTypeHopRequest,
+)
 from lib.sciond_api.service_req import (
     SCIONDServiceInfoReply,
     SCIONDServiceInfoReplyEntry,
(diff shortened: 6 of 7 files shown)


mwfarb:
@kormat @shitz This PR is updated for the current master after I somehow forgot everything I learned in the past year WRT our git process. It essentially allows the return of segments used to construct paths during sciond's get_paths() for consumption by visualizations and other apps.

kormat:
Reviewed 7 of 7 files at r1.

---

*[proto/sciond.capnp, line 23 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4okMulflJka8QozH3f:-L4okMulflJka8QozH3g:b-erz3y4) ([raw file](https://github.com/scionproto/scion/blob/7b0588e1c3c0f684784b50d855cb9599c11ccb4b/proto/sciond.capnp#L23)):*
> ```
>         serviceInfoReply @10 :ServiceInfoReply;
>         revReply @11 :RevReply;
>         segmentRequest @12 :SegmentRequest;
> ```

I'm not thrilled with the naming here. To me, this name implies you tell sciond you want segments from A to B, whereas this asks sciond for all segments of a given type. `segTypeReq`/`segTypeReply` maybe?

---

*[proto/sciond.capnp, line 125 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oreXknf2mNLE_0qlI:-L4oreXknf2mNLE_0qlJ:b-1e01nh) ([raw file](https://github.com/scionproto/scion/blob/7b0588e1c3c0f684784b50d855cb9599c11ccb4b/proto/sciond.capnp#L125)):*
> ```
> 
> struct SegmentRequest {
>     segmentType @0 :PSeg.PathSegType;
> ```

"type" would sufficient, as the context is already clear.

---

*[proto/sciond.capnp, line 129 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oqMIQ3-eMaOqh4Rnk:-L4oqMIQ3-eMaOqh4Rnl:bysrt43) ([raw file](https://github.com/scionproto/scion/blob/7b0588e1c3c0f684784b50d855cb9599c11ccb4b/proto/sciond.capnp#L129)):*
> ```
> 
> struct SegmentReply {
>     entries @0 :List(SegmentReplyEntry);
> ```

Why not simply `:List(PSeg.PathSegment)`?

---

*[python/lib/sciond_api/segment_req.py, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oqySTMbni_TWeZe64:-L4oqySTMbni_TWeZe65:bb6ht2i) ([raw file](https://github.com/scionproto/scion/blob/7b0588e1c3c0f684784b50d855cb9599c11ccb4b/python/lib/sciond_api/segment_req.py#L1)):*
> ```Python
> # Copyright 2017 ETH Zurich
> ```

2018

---

*[python/lib/sciond_api/segment_req.py, line 31 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4orFn8V-qqpKLLVDT6:-L4orFn8V-qqpKLLVDT7:b-tx5ohy) ([raw file](https://github.com/scionproto/scion/blob/7b0588e1c3c0f684784b50d855cb9599c11ccb4b/python/lib/sciond_api/segment_req.py#L31)):*
> ```Python
> 
>     @classmethod
>     def from_values(cls, seg_type=None):
> ```

This is not an optional argument, there's no sane default. And this will cause an exception if the arg isn't supplied.

---

*[python/lib/sciond_api/segment_req.py, line 33 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4orMql9sEULmb45Lt7:-L4orMql9sEULmb45Lt8:buguhie) ([raw file](https://github.com/scionproto/scion/blob/7b0588e1c3c0f684784b50d855cb9599c11ccb4b/python/lib/sciond_api/segment_req.py#L33)):*
> ```Python
>     def from_values(cls, seg_type=None):
>         p.segmentType = seg_type
> ```

p = cls.P_CLS.new_message(segmentType=seg_type)

---

*[python/lib/sciond_api/segment_req.py, line 39 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oroFSHpiDNFvsz-RC:-L4oroFSHpiDNFvsz-RD:bgxi6mn) ([raw file](https://github.com/scionproto/scion/blob/7b0588e1c3c0f684784b50d855cb9599c11ccb4b/python/lib/sciond_api/segment_req.py#L39)):*
> ```Python
>         desc = ["%s:" % self.NAME]
>         desc.append("  segmentType: %s" % self.p.segmentType)
>         return "\n".join(desc)
> ```

This is a very simple request, with no repeating fields, so this can just be
```
return "%s: type: %s" % (self.NAME, self.p.segmentType)
```

---

*[python/sciond/sciond.py, line 378 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oyffUEz57sFQu_L5b:-L4oyffUEz57sFQu_L5c:b-9wvbqx) ([raw file](https://github.com/scionproto/scion/blob/7b0588e1c3c0f684784b50d855cb9599c11ccb4b/python/sciond/sciond.py#L378)):*
> ```Python
>             db = self.up_segments
>         elif segmentType == PST.DOWN:
>             db = self.down_segments
> ```

This should log an error if the segmentType is anything else.

---

*[python/sciond/sciond.py, line 382 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oyMA8vg9NvvcDoq4i:-L4oyMA8vg9NvvcDoq4j:bhbu0a1) ([raw file](https://github.com/scionproto/scion/blob/7b0588e1c3c0f684784b50d855cb9599c11ccb4b/python/sciond/sciond.py#L382)):*
> ```Python
>         total_segs = len(self.core_segments) + \
>             len(self.up_segments) + len(self.down_segments)
>         assert total_segs != 0, 'No segments found, did you call get_paths() first?'
> ```

This is a bad idea - if sciond had just restarted, for example, this will kill it. There's no need to check for this case. The response will just have no entries in it. That's perfectly legal.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1383)*

kormat:
Always put a blank line in the git commit after the description, otherwise git will do things like:

`* pubpr/1383        7b0588e [pubup/pubpr/1383] Adds sciond request for segments used to construct paths - Adds ability to request detailed segment data after get_paths succeeds. - Returns requested segments data (core, up, down). - Requested segment data can be used to calculate core ASes in paths. - Requested segment data can be used to calculate path types. - Useful for visualizations to display core and path types for every AS routed through. - Closes #1299.`

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1383#-:-L4pJVmOOmbn8ON14jiC:bs7unmn)*

mwfarb:
Updated original commit message.

---

Review status: 0 of 7 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[proto/sciond.capnp, line 23 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4okMulflJka8QozH3f:-L4wFilwLjHeM1Lb7PQK:b-mzl0m6) ([raw file](https://github.com/scionproto/scion/blob/d0fdecd5c9cc7919bd31b2c0e4cec0ec87a4152e/proto/sciond.capnp#L23)):*

I'm not thrilled with the naming here. To me, this name implies you tell sciond you want segments from A to B, whereas this asks sciond for all segments of a given type. `segTypeReq`/`segTypeReply` maybe?
</blockquote></details>

Good point. Better to not confuse with other messaging in our protocols. I like your suggestion. I've propagated that naming throughout.

---

*[proto/sciond.capnp, line 125 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oreXknf2mNLE_0qlI:-L4wEx8iLc0pbS_hZ9uS:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d0fdecd5c9cc7919bd31b2c0e4cec0ec87a4152e/proto/sciond.capnp#L125)):*

"type" would sufficient, as the context is already clear.
</blockquote></details>

Done.

---

*[proto/sciond.capnp, line 129 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oqMIQ3-eMaOqh4Rnk:-L4wEuHxeLaEsjhvk-ks:b-w10y5l) ([raw file](https://github.com/scionproto/scion/blob/d0fdecd5c9cc7919bd31b2c0e4cec0ec87a4152e/proto/sciond.capnp#L129)):*

Why not simply `:List(PSeg.PathSegment)`?
</blockquote></details>

True. I've removed `SegmentReplyEntry`.

---

*[python/lib/sciond_api/segment_req.py, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oqySTMbni_TWeZe64:-L4wEzv15ckJAzd_omJO:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d0fdecd5c9cc7919bd31b2c0e4cec0ec87a4152e/python/lib/sciond_api/segment_req.py#L1)):*

2018
</blockquote></details>

Done.

---

*[python/lib/sciond_api/segment_req.py, line 31 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4orFn8V-qqpKLLVDT6:-L4wFROfJMx9BkdEOCoB:buv0nem) ([raw file](https://github.com/scionproto/scion/blob/d0fdecd5c9cc7919bd31b2c0e4cec0ec87a4152e/python/lib/sciond_api/segment_req.py#L31)):*

This is not an optional argument, there's no sane default. And this will cause an exception if the arg isn't supplied.
</blockquote></details>

Yep. Removed default.

---

*[python/lib/sciond_api/segment_req.py, line 33 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4orMql9sEULmb45Lt7:-L4wF_Db3Vdn53Gct9Ei:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d0fdecd5c9cc7919bd31b2c0e4cec0ec87a4152e/python/lib/sciond_api/segment_req.py#L33)):*

p = cls.P_CLS.new_message(segmentType=seg_type)
</blockquote></details>

Done.

---

*[python/lib/sciond_api/segment_req.py, line 39 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oroFSHpiDNFvsz-RC:-L4wFacC_qNBaSBkDvrC:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d0fdecd5c9cc7919bd31b2c0e4cec0ec87a4152e/python/lib/sciond_api/segment_req.py#L39)):*

This is a very simple request, with no repeating fields, so this can just be
```
return "%s: type: %s" % (self.NAME, self.p.segmentType)
```
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 378 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oyffUEz57sFQu_L5b:-L4wFc0SrXsuvmXTJaC8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d0fdecd5c9cc7919bd31b2c0e4cec0ec87a4152e/python/sciond/sciond.py#L378)):*

This should log an error if the segmentType is anything else.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 382 at r1](https://reviewable.io:443/reviews/scionproto/scion/1383#-L4oyMA8vg9NvvcDoq4i:-L4wFgm50CV97xrTR3Gi:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d0fdecd5c9cc7919bd31b2c0e4cec0ec87a4152e/python/sciond/sciond.py#L382)):*

This is a bad idea - if sciond had just restarted, for example, this will kill it. There's no need to check for this case. The response will just have no entries in it. That's perfectly legal.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1383#-:-L4wNChmtHJcKVXzcPMS:b-heaefp)*

kormat:
Reviewed 7 of 7 files at r2.

---

*[python/lib/app/sciond.py, line 205 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58uIqDrxULgcAQ5_R1:-L58uIqDrxULgcAQ5_R2:b5th81e) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/lib/app/sciond.py#L205)):*
> ```Python
>         return self._resolve_dst_addr(spkt.addrs.src, spkt.addrs.dst)
> 
>     def get_segments(self, seg_type):
> ```

This should be renamed, too. `get_type_segs`?

---

*[python/lib/app/sciond.py, line 420 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58uNgdNaDShsJiA7Lf:-L58uNgdNaDShsJiA7Lg:b-twzawc) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/lib/app/sciond.py#L420)):*
> ```Python
> 
>     :param seg_type: The type of PathSegmentType requested.
>     :returns: List of SCIONDSegTypeReplyEntry objects.
> ```

This is a list of PathSegments now.

---

*[python/lib/sciond_api/segment_req.py, line 45 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58vAG9xNTPBqsPFMXt:-L58vAG9xNTPBqsPFMXu:b-47amfk) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/lib/sciond_api/segment_req.py#L45)):*
> ```Python
>     @classmethod
>     def from_values(cls, entries):
>         p = cls.P_CLS.new_message(entries=entries)
> ```

I don't think this works (or if it does, what exactly the result is :). The previous code was correct (using `p.init` etc).

---

*[python/lib/sciond_api/segment_req.py, line 49 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58vZdvjCGzMa6sJVQ0:-L58vZdvjCGzMa6sJVQ1:bb99lhe) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/lib/sciond_api/segment_req.py#L49)):*
> ```Python
> 
>     def entry(self, idx):
>         return self.p.entries[idx]
> ```

This should return PathSegments, not raw capnp objects. Same for `iter_entries`.

---

*[python/sciond/sciond.py, line 379 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58vzNr9j_9knhDLD5q:-L58vzNr9j_9knhDLD5r:brd7uza) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/sciond/sciond.py#L379)):*
> ```Python
>             db = self.down_segments
>         else:
>             logging.error("Requesting segment type %s unrecognized.\n%s" % segmentType)
> ```

This has a trailing `\n%s`, that will cause a run-time error.

---

*[python/sciond/sciond.py, line 384 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58w93KwE0mLwzhS9KS:-L58w93KwE0mLwzhS9KT:bkqxv49) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/sciond/sciond.py#L384)):*
> ```Python
>         for segment in db(full=True):
>             seg_entries.append(segment.p)
>         seg_reply = SCIONDMsg(SCIONDSegTypeReply.from_values(seg_entries), pld.id)
> ```

Hmm. One thing to consider here is that you can run out of space pretty quickly. The max message size is ~65k. A sciond that has a lot of segments could easily exceed this, with segments ranging from maybe 750B to 10KB. This makes this API only really usable in a controlled/test environment, which in turn raises the question of whether it's a good idea to put this into the main codebase. I'm leaning towards no.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1383)*

mwfarb:
Review status: 2 of 7 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[python/lib/app/sciond.py, line 205 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58uIqDrxULgcAQ5_R1:-L5EjGMB66ibidfbIVVt:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/lib/app/sciond.py#L205)):*

This should be renamed, too. `get_type_segs`?
</blockquote></details>

Done.

---

*[python/lib/app/sciond.py, line 420 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58uNgdNaDShsJiA7Lf:-L5EkU5nNSX_TRaePVma:b-9ly4cf) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/lib/app/sciond.py#L420)):*

This is a list of PathSegments now.
</blockquote></details>

Reverted to list of `SCIONDSegTypeReplyEntry`

---

*[python/lib/sciond_api/segment_req.py, line 45 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58vAG9xNTPBqsPFMXt:-L5EknYv4wIDzinE60SR:be4ptly) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/lib/sciond_api/segment_req.py#L45)):*

I don't think this works (or if it does, what exactly the result is :). The previous code was correct (using `p.init` etc).
</blockquote></details>

Reverted to list of `SCIONDSegTypeReplyEntry` and restored. :-)

---

*[python/lib/sciond_api/segment_req.py, line 49 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58vZdvjCGzMa6sJVQ0:-L5EkvaDsLXSMLdkZOHd:b-9ly4cf) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/lib/sciond_api/segment_req.py#L49)):*

This should return PathSegments, not raw capnp objects. Same for `iter_entries`.
</blockquote></details>

Reverted to list of `SCIONDSegTypeReplyEntry`

---

*[python/sciond/sciond.py, line 379 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58vzNr9j_9knhDLD5q:-L5EcjwG2MejfTtfLT_A:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/sciond/sciond.py#L379)):*

This has a trailing `\n%s`, that will cause a run-time error.
</blockquote></details>

Done.

---

*[python/sciond/sciond.py, line 384 at r2](https://reviewable.io:443/reviews/scionproto/scion/1383#-L58w93KwE0mLwzhS9KS:-L5EjSwJHAPSMtFZ7r-G:budocm0) ([raw file](https://github.com/scionproto/scion/blob/08ee160f642415ada86e51c89db0445bf9b36336/python/sciond/sciond.py#L384)):*

Hmm. One thing to consider here is that you can run out of space pretty quickly. The max message size is ~65k. A sciond that has a lot of segments could easily exceed this, with segments ranging from maybe 750B to 10KB. This makes this API only really usable in a controlled/test environment, which in turn raises the question of whether it's a good idea to put this into the main codebase. I'm leaning towards no.
</blockquote></details>

I'd like to make this useful and at least come under the size `get_paths()` currently returns. So in general, I've reverted the reply from a list of `PathSegment` back to a list of `SCIONDSegTypeReplyEntry` with modifications. The return objects should contain the minimum data we need, including timestamps and all interfaces, rather than the larger PathSegment which included a number of sub-objects not currently needed. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1383)*

kormat:
Reviewed 5 of 5 files at r3.

---

*[proto/sciond.capnp, line 124 at r3](https://reviewable.io:443/reviews/scionproto/scion/1383#-L5mfK-C68kcwK3da4cs:-L5mfK-DLOi8IsgvZsUU:b-57bzje) ([raw file](https://github.com/scionproto/scion/blob/192ec1f82f16e29c8b5eef98904a8c4570d86815/proto/sciond.capnp#L124)):*
> ```
> }
> 
> struct SegTypeReq {
> ```

So, now we're back to a naming issue :) This no longer requests segments. My best suggestion right now is `SegTypeHopReq` - how does that sound?

Relatedly, one thing that comes to mind is that we're planning to rewrite sciond in the ~near future, in which case it will use the new sqlite3-backed pathdb (https://github.com/scionproto/scion/tree/master/go/lib/pathdb/sqlite). This means that if you want to do advanced querying such as this use-case, you could simply query the database directly for anything you like, without needing sciond protocol/code changes.

With that in mind, while i'm ok with adding this functionality to the existing python sciond, i don't plan to have the new go version support it. Just something to be aware of.

---

*[proto/sciond.capnp, line 134 at r3](https://reviewable.io:443/reviews/scionproto/scion/1383#-L5mioj_qSkcNYkuD9LV:-L5mioj_qSkcNYkuD9LW:b-deun8x) ([raw file](https://github.com/scionproto/scion/blob/192ec1f82f16e29c8b5eef98904a8c4570d86815/proto/sciond.capnp#L134)):*
> ```
> struct SegTypeReplyEntry {
>     interfaces @0 :List(PathInterface);  # List of interfaces for the segment
>     timestamp @1 :UInt64;  # Creation timestamp
> ```

This should specify the unit/epoch. E.g. `Creation timestamp. Seconds since Unix Epoch.`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1383)*

mwfarb:
Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[proto/sciond.capnp, line 124 at r3](https://reviewable.io:443/reviews/scionproto/scion/1383#-L5mfK-C68kcwK3da4cs:-L5oODsWcBWv088LOCVY:b-fkat7y) ([raw file](https://github.com/scionproto/scion/blob/192ec1f82f16e29c8b5eef98904a8c4570d86815/proto/sciond.capnp#L124)):*

So, now we're back to a naming issue :) This no longer requests segments. My best suggestion right now is `SegTypeHopReq` - how does that sound?

Relatedly, one thing that comes to mind is that we're planning to rewrite sciond in the ~near future, in which case it will use the new sqlite3-backed pathdb (https://github.com/scionproto/scion/tree/master/go/lib/pathdb/sqlite). This means that if you want to do advanced querying such as this use-case, you could simply query the database directly for anything you like, without needing sciond protocol/code changes.

With that in mind, while i'm ok with adding this functionality to the existing python sciond, i don't plan to have the new go version support it. Just something to be aware of.
</blockquote></details>

Sure! I've updated the naming across all files. I'm glad to know the work on the pathdb is forthcoming. This kind of data is quite useful.

---

*[proto/sciond.capnp, line 134 at r3](https://reviewable.io:443/reviews/scionproto/scion/1383#-L5mioj_qSkcNYkuD9LV:-L5nz0YuOiVqxgt1A-46:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/192ec1f82f16e29c8b5eef98904a8c4570d86815/proto/sciond.capnp#L134)):*

This should specify the unit/epoch. E.g. `Creation timestamp. Seconds since Unix Epoch.`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1383)*

mwfarb:
Screenshot result combined with https://github.com/netsec-ethz/scion-viz/pull/40.

screenshot from 2018-02-19 16-46-02

kormat:
:lgtm:

Please rebase it on master, and then we can merge it.

---

Reviewed 7 of 7 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1383#-:-L6MlQEuuyFNMX9f7eV-:b1c6ri2)*

kormat:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1383)*

PR #1385, opened 2018-01-30
Simplify handling of signed ctrl messages.

kormat:
This patch reworks how scion ctrl payloads are created/signed/verified.
By using callbacks, an app can create and sign a ctrl payload and only
have a single error value to check, greatly reducing boiler-plate. It
also becomes much more obvious when a message is not being signed.

This patch also adds support for non-union data at the CtrlPld, CertMgmt
and PathMgmt levels. This allows for a unified approach to things like
request IDs, tracing, and error reporting.

Also:
- Cache the parsed CtrlPld in SignedCtrlPld, to reduce the overhead of
  extracting it multiple times.
- Add request ID and trace ID to CtrlPld, though they are currently not
  used.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1385)

### DIFF
--- a/go/border/gen.go
+++ b/go/border/gen.go
@@ -109,7 +109,8 @@ func (r *Router) genIFIDPkt(ifID common.IFIDType, ctx *rctx.Ctx) {
 	logger := log.New("ifid", ifID)
 	intf := ctx.Conf.Net.IFs[ifID]
 	srcAddr := intf.IFAddr.PublicAddrInfo(intf.IFAddr.Overlay)
-	scpld, err := ctrl.NewSignedPldFromUnion(&ifid.IFID{OrigIfID: uint64(ifID)})
+	scpld, err := ctrl.MkSPld(ctrl.NewPldF(&ifid.IFID{OrigIfID: uint64(ifID)}, nil),
+		ctrl.NullSigner)
 	if err != nil {
 		logger.Error("Error generating IFID payload", "err", common.FmtError(err))
 		return
@@ -141,7 +142,8 @@ func (r *Router) genIFStateReq() {
 	ctx := rctx.Get()
 	// Pick first local address from topology as source.
 	srcAddr := ctx.Conf.Net.LocAddr[0].PublicAddrInfo(ctx.Conf.Net.LocAddr[0].Overlay)
-	scpld, err := ctrl.NewSignedPathMgmtPld(&path_mgmt.IFStateReq{})
+	scpld, err := ctrl.MkSPld(ctrl.NewPathMgmtPldF(&path_mgmt.IFStateReq{}, nil, nil),
+		ctrl.NullSigner)
 	if err != nil {
 		log.Error("Error generating IFStateReq payload", "err", common.FmtError(err))
 		return
--- a/go/border/gen.go
+++ b/go/border/gen.go
@@ -109,10 +109,14 @@ func (r *Router) genIFIDPkt(ifID common.IFIDType, ctx *rctx.Ctx) {
 	logger := log.New("ifid", ifID)
 	intf := ctx.Conf.Net.IFs[ifID]
 	srcAddr := intf.IFAddr.PublicAddrInfo(intf.IFAddr.Overlay)
-	scpld, err := ctrl.MkSPld(ctrl.NewPldF(&ifid.IFID{OrigIfID: uint64(ifID)}, nil),
-		ctrl.NullSigner)
+	cpld, err := ctrl.NewPld(&ifid.IFID{OrigIfID: uint64(ifID)}, nil)
 	if err != nil {
-		logger.Error("Error generating IFID payload", "err", common.FmtError(err))
+		logger.Error("Error generating IFID Ctrl payload", "err", common.FmtError(err))
+		return
+	}
+	scpld, err := cpld.SignedPld(ctrl.NullSigner)
+	if err != nil {
+		logger.Error("Error generating IFID signed Ctrl payload", "err", common.FmtError(err))
 		return
 	}
 	if err := r.genPkt(intf.RemoteIA, addr.HostFromIP(intf.RemoteAddr.IP),
@@ -142,10 +146,14 @@ func (r *Router) genIFStateReq() {
 	ctx := rctx.Get()
 	// Pick first local address from topology as source.
 	srcAddr := ctx.Conf.Net.LocAddr[0].PublicAddrInfo(ctx.Conf.Net.LocAddr[0].Overlay)
-	scpld, err := ctrl.MkSPld(ctrl.NewPathMgmtPldF(&path_mgmt.IFStateReq{}, nil, nil),
-		ctrl.NullSigner)
+	cpld, err := ctrl.NewPathMgmtPld(&path_mgmt.IFStateReq{}, nil, nil)
 	if err != nil {
-		log.Error("Error generating IFStateReq payload", "err", common.FmtError(err))
+		log.Error("Error generating IFStateReq Ctrl payload", "err", common.FmtError(err))
+		return
+	}
+	scpld, err := cpld.SignedPld(ctrl.NullSigner)
+	if err != nil {
+		log.Error("Error generating IFStateReq signed Ctrl payload", "err", common.FmtError(err))
 		return
 	}
 	if err := r.genPkt(ctx.Conf.IA, addr.SvcBS.Multicast(), 0, srcAddr, scpld); err != nil {
--- a/go/border/revinfo.go
+++ b/go/border/revinfo.go
@@ -57,9 +57,14 @@ func (r *Router) fwdRevInfo(revInfo *path_mgmt.RevInfo, dstHost addr.HostAddr) {
 	ctx := rctx.Get()
 	// Pick first local address from topology as source.
 	srcAddr := ctx.Conf.Net.LocAddr[0].PublicAddrInfo(ctx.Conf.Topo.Overlay)
-	scpld, err := ctrl.MkSPld(ctrl.NewPldF(revInfo, nil), ctrl.NullSigner)
+	cpld, err := ctrl.NewPld(revInfo, nil)
 	if err != nil {
-		log.Error("Error generating RevInfo payload", "err", common.FmtError(err))
+		log.Error("Error generating RevInfo Ctrl payload", "err", common.FmtError(err))
+		return
+	}
+	scpld, err := cpld.SignedPld(ctrl.NullSigner)
+	if err != nil {
+		log.Error("Error generating RevInfo signed Ctrl payload", "err", common.FmtError(err))
 		return
 	}
 	if err = r.genPkt(ctx.Conf.IA, *dstHost.(*addr.HostSVC), 0, srcAddr, scpld); err != nil {
--- a/go/border/revinfo.go
+++ b/go/border/revinfo.go
@@ -57,7 +57,7 @@ func (r *Router) fwdRevInfo(revInfo *path_mgmt.RevInfo, dstHost addr.HostAddr) {
 	ctx := rctx.Get()
 	// Pick first local address from topology as source.
 	srcAddr := ctx.Conf.Net.LocAddr[0].PublicAddrInfo(ctx.Conf.Topo.Overlay)
-	cpld, err := ctrl.NewPld(revInfo, nil)
+	cpld, err := ctrl.NewPathMgmtPld(revInfo, nil, nil)
 	if err != nil {
 		log.Error("Error generating RevInfo Ctrl payload", "err", common.FmtError(err))
 		return
--- a/go/border/revinfo.go
+++ b/go/border/revinfo.go
@@ -57,7 +57,7 @@ func (r *Router) fwdRevInfo(revInfo *path_mgmt.RevInfo, dstHost addr.HostAddr) {
 	ctx := rctx.Get()
 	// Pick first local address from topology as source.
 	srcAddr := ctx.Conf.Net.LocAddr[0].PublicAddrInfo(ctx.Conf.Topo.Overlay)
-	scpld, err := ctrl.NewSignedPathMgmtPld(revInfo)
+	scpld, err := ctrl.MkSPld(ctrl.NewPldF(revInfo, nil), ctrl.NullSigner)
 	if err != nil {
 		log.Error("Error generating RevInfo payload", "err", common.FmtError(err))
 		return
--- a/go/border/rpkt/process.go
+++ b/go/border/rpkt/process.go
@@ -159,7 +159,11 @@ func (rp *RtrPkt) processDestSelf() (HookResult, error) {
 func (rp *RtrPkt) processIFID(ifid *ifid.IFID) (HookResult, error) {
 	// Set the RelayIF field in the payload to the current interface ID.
 	ifid.RelayIfID = uint64(*rp.ifCurr)
-	scpld, err := ctrl.MkSPld(ctrl.NewPldF(ifid, nil), ctrl.NullSigner)
+	cpld, err := ctrl.NewPld(ifid, nil)
+	if err != nil {
+		return HookError, err
+	}
+	scpld, err := cpld.SignedPld(ctrl.NullSigner)
 	if err != nil {
 		return HookError, err
 	}
--- a/go/border/rpkt/process.go
+++ b/go/border/rpkt/process.go
@@ -159,7 +159,7 @@ func (rp *RtrPkt) processDestSelf() (HookResult, error) {
 func (rp *RtrPkt) processIFID(ifid *ifid.IFID) (HookResult, error) {
 	// Set the RelayIF field in the payload to the current interface ID.
 	ifid.RelayIfID = uint64(*rp.ifCurr)
-	scpld, err := ctrl.NewSignedPldFromUnion(ifid)
+	scpld, err := ctrl.MkSPld(ctrl.NewPldF(ifid, nil), ctrl.NullSigner)
 	if err != nil {
 		return HookError, err
 	}
--- a/go/cert_srv/cert.go
+++ b/go/cert_srv/cert.go
@@ -72,7 +72,7 @@ func (h *ChainHandler) sendChainRep(addr *snet.Addr, chain *cert.Chain) error {
 	if err != nil {
 		return err
 	}
-	cpld, err := ctrl.NewCertMgmtPld(&cert_mgmt.Chain{RawChain: raw})
+	cpld, err := ctrl.NewCertMgmtPld(&cert_mgmt.Chain{RawChain: raw}, nil, nil)
 	if err != nil {
 		return err
 	}
@@ -94,7 +94,7 @@ func (h *ChainHandler) fetchChain(addr *snet.Addr, req *cert_mgmt.ChainReq) erro
 
 // sendChainReq sends a certificate chain request to the specified remote AS.
 func (h *ChainHandler) sendChainReq(req *cert_mgmt.ChainReq) error {
-	cpld, err := ctrl.NewCertMgmtPld(req)
+	cpld, err := ctrl.NewCertMgmtPld(req, nil, nil)
 	if err != nil {
 		return err
 	}
@@ -123,7 +123,7 @@ func (h *ChainHandler) HandleRep(addr *snet.Addr, rep *cert_mgmt.Chain) {
 	if reqVer == nil && reqNew == nil { // No pending requests
 		return
 	}
-	cpld, err := ctrl.NewCertMgmtPld(rep)
+	cpld, err := ctrl.NewCertMgmtPld(rep, nil, nil)
 	if err != nil {
 		log.Error("Unable to create certificate chain reply", "key", key,
 			"err", common.FmtError(err))
--- a/go/cert_srv/trc.go
+++ b/go/cert_srv/trc.go
@@ -71,7 +71,7 @@ func (h *TRCHandler) sendTRCRep(addr *snet.Addr, t *trc.TRC) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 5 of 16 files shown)


shitz:
Reviewed 16 of 16 files at r1.

---

*[go/lib/ctrl/ctrl.go, line 47 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46UeHVrrU2xGtOr92A:-L46UeHVrrU2xGtOr92B:b-p0dr9e) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl.go#L47)):*
> ```Go
> 
> // NewPldF returns a Plder that calls NewPld
> func NewPldF(u proto.Cerealizable, d *Data) Plder {
> ```

I'm not a fan of this pattern. The only thing you gain is that the caller doesn't have to first create the payload, check for an error, then call `Request` and check for an error and instead can just check the error of `Request`. However,  it took me a while to understand this. 

---

*[go/lib/ctrl/signed_util.go, line 24 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46cKNxpBXBnrbM1tsA:-L46cKNxpBXBnrbM1tsB:b2u3wqe) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L24)):*
> ```Go
> )
> 
> type Signer func(*Pld) (*SignedPld, error)
> ```

I wonder if all of these callbacks and closures should instead be `interfaces`. I learned that you can implement an interface with purely a method so this would allow your approach, but would also allow an object-based implementation plus I think interfaces are more intuitive to understand than closures.

---

*[go/lib/ctrl/signed_util.go, line 36 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46ZFwREQAVTWea447E:-L46ZFwREQAVTWea447F:blbu460) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L36)):*
> ```Go
> }
> 
> type trustStore struct{} // TODO(kormat): replace this with the actual trust store.
> ```

I _think_ this will lead to a circular dependency. Having the trust store here means importing `go/lib/trust`. The trust store itself will make use of `go/lib/ctrl`, either directly or through the messenger, which will then lead to a circular import. I'm not sure if these utils can be in `ctrl`

---

*[go/lib/ctrl/signed_util.go, line 41 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46Y9sqGyunQWaOPPon:-L46Y9sr-VWQ63z7XtE2:b-6mrmnx) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L41)):*
> ```Go
> type SigCheckerInner func(*SignedPld, *trustStore) error
> 
> func MkSigchecker(sigCin SigCheckerInner, tStore *trustStore) SigChecker {
> ```

`MkSigChecker`

---

*[go/lib/ctrl/signed_util.go, line 94 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46_ct33uPO-Mklx2QI:-L46_ct33uPO-Mklx2QJ:bdorwpo) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L94)):*
> ```Go
> 
> // MkSPld creates a SignedPld from the provided Plder and Signer.
> func MkSPld(plder Plder, signer Signer) (*SignedPld, error) {
> ```

This is a bit too acronym-y for my tasted. `MkSignedPld`is reasonably short. Also, given how often I see `ctrl.MkSPld(..., ctrl.NullSigner)` it probably would make sense to have something like `MkUnsignedPld`, even though it might be confusing that it returns an object of `SignedPld`. I guess the correct term would be `MkUnsignedSignedPld`, and the correct name for `SignedPld` would be `PossiblySignedPld` ;)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1385)*

scrye:
Reviewed 1 of 16 files at r1.
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[go/lib/ctrl/ctrl.go, line 47 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46Kt6B4Az_jQEs6myW:-L46Kt6B4Az_jQEs6myX:b-dzzw9h) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl.go#L47)):*
> ```Go
> 
> // NewPldF returns a Plder that calls NewPld
> func NewPldF(u proto.Cerealizable, d *Data) Plder {
> ```

The state of the closure returned by this function is immutable and bound to variables `u` and `d`. This means that prior to each call to `Request`, the callback needs to be constructed to bind the variables:

```
plder = NewPldF(u, d)
ctrl_msg.Request(ctx, plder, ...)
```

At that point, why not just:
```
pld, err = NewPld(u, d)
// Handle err
ctrl_msg.Request(ctx, pld, ...)
```

The closure callback just moves error handling from the outside to inside `Request`, but it looks like it does nothing else.

More complex implementations of `Plder` are also subject to the same limitation, because their whole state needs to be constructed prior to each call to `Request`.

---

*[go/lib/ctrl/ctrl.go, line 96 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46IZHbPrnFQWeUz0SP:-L46IZHcNIdKQggXQ94v:b-l5uz5s) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl.go#L96)):*
> ```Go
> }
> 
> func (p *Pld) GetCertMgmt() (*cert_mgmt.Pld, *Data, error) {
> ```

The returned values here probably need a docstring.

---

*[go/lib/ctrl/ctrl.go, line 109 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46Ic75uhSZ362SPfBN:-L46Ic75uhSZ362SPfBO:b-l5uz5s) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl.go#L109)):*
> ```Go
> }
> 
> func (p *Pld) GetPathMgmt() (*path_mgmt.Pld, *Data, error) {
> ```

The returned values here probably need a docstring.

---

*[go/lib/ctrl/signed_util.go, line 38 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46d5-XNrQMYisbuVl5:-L46d5-XNrQMYisbuVl6:bkwhaf4) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L38)):*
> ```Go
> type trustStore struct{} // TODO(kormat): replace this with the actual trust store.
> 
> type SigChecker func(*SignedPld) error
> ```

Suggest changing `SigChecker` to `Verifier`, so it's more intuitive for external packages to refer to `ctrl.Signer` and `ctrl.Verifier`.

---

*[go/lib/ctrl/signed_util.go, line 39 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46VjmPZIKOB3x-cLdZ:-L46VjmPZIKOB3x-cLd_:bg7qw0b) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L39)):*
> ```Go
> 
> type SigChecker func(*SignedPld) error
> type SigCheckerInner func(*SignedPld, *trustStore) error
> ```

I'm not sure I understand what the meaning behind this type is. To me it looks like it is a variant of `SigChecker` with a bound trust store, but it is used after some checks in `MkSigchecker`. Where does the job of `SigChecker` internals stop, and the job of `SigCheckerInner` start?

---

*[go/lib/ctrl/signed_util.go, line 41 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46ScDoga6GYcq8adXe:-L46ScDpp0agUJsonL0y:b-bvl0u2) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L41)):*
> ```Go
> type SigCheckerInner func(*SignedPld, *trustStore) error
> 
> func MkSigchecker(sigCin SigCheckerInner, tStore *trustStore) SigChecker {
> ```

The purpose of `sigCin` is not evident to me. A very detailed docstring is needed here, s.t. users of package `ctrl` have an intuition on the differences between `SigChecker` and `SigCheckerInner`.

---

*[go/lib/ctrl/ctrl_msg/msg.go, line 15 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46eDH0041pSOqiXHPq:-L46eDH0041pSOqiXHPr:bviwckr) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl_msg/msg.go#L15)):*
> ```Go
> 
> package ctrl_msg
> ```

This needs a docstring, as it's not immediately clear what the package offers (it's more related to signing and verifying, than actually messages).

---

*[go/lib/ctrl/ctrl_msg/msg.go, line 78 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46ffeeiplQEa1sBu1I:-L46ffeeiplQEa1sBu1J:b-g166ew) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl_msg/msg.go#L78)):*
> ```Go
> }
> 
> func (r *Requester) notify(ctx context.Context, plder ctrl.Plder, a net.Addr, f notifyF) error {
> ```

I think the callback pattern is a bit more confusing than just calling a helper method to set up `spld` at the start of both `Notify` and `NotifyUnreliable`, and then calling the correct function. The reason being that to understand`Notify` and `NotifyUnreliable` a programmer needs to look into `notify`, and the other way around. In any case, I guess it's about preference here, and I prefer only using callbacks only when there are no other solutions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1385)*

kormat:
Review status: 8 of 16 files reviewed at latest revision, 13 unresolved discussions.

---

*[go/lib/ctrl/ctrl.go, line 47 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46Kt6B4Az_jQEs6myW:-L47-ImywwLGHS8v-9Tx:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl.go#L47)):*

The state of the closure returned by this function is immutable and bound to variables `u` and `d`. This means that prior to each call to `Request`, the callback needs to be constructed to bind the variables:

```
plder = NewPldF(u, d)
ctrl_msg.Request(ctx, plder, ...)
```

At that point, why not just:
```
pld, err = NewPld(u, d)
// Handle err
ctrl_msg.Request(ctx, pld, ...)
```

The closure callback just moves error handling from the outside to inside `Request`, but it looks like it does nothing else.

More complex implementations of `Plder` are also subject to the same limitation, because their whole state needs to be constructed prior to each call to `Request`.
</blockquote></details>

Done.

---

*[go/lib/ctrl/ctrl.go, line 47 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46UeHVrrU2xGtOr92A:-L47-J5P8w7Hlq5Ymkq_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl.go#L47)):*

I'm not a fan of this pattern. The only thing you gain is that the caller doesn't have to first create the payload, check for an error, then call `Request` and check for an error and instead can just check the error of `Request`. However,  it took me a while to understand this. 
</blockquote></details>

Done.

---

*[go/lib/ctrl/ctrl.go, line 96 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46IZHbPrnFQWeUz0SP:-L47-abwzBHdwTm2zvkE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl.go#L96)):*

The returned values here probably need a docstring.
</blockquote></details>

Done.

---

*[go/lib/ctrl/ctrl.go, line 109 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46Ic75uhSZ362SPfBN:-L47-ayP5_XFgTho6sFG:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl.go#L109)):*

The returned values here probably need a docstring.
</blockquote></details>

Done.

---

*[go/lib/ctrl/signed_util.go, line 24 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46cKNxpBXBnrbM1tsA:-L47-buu2-1LofWj2FMs:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L24)):*

I wonder if all of these callbacks and closures should instead be `interfaces`. I learned that you can implement an interface with purely a method so this would allow your approach, but would also allow an object-based implementation plus I think interfaces are more intuitive to understand than closures.
</blockquote></details>

Done.

---

*[go/lib/ctrl/signed_util.go, line 36 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46ZFwREQAVTWea447E:-L47-coMEFocyjaXwYov:b-cy24yv) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L36)):*

I _think_ this will lead to a circular dependency. Having the trust store here means importing `go/lib/trust`. The trust store itself will make use of `go/lib/ctrl`, either directly or through the messenger, which will then lead to a circular import. I'm not sure if these utils can be in `ctrl`
</blockquote></details>

Good point, the PR with the trust store can instead put an interface into `go/lib/infra` as @scrye suggested, and just change the type in this file.

---

*[go/lib/ctrl/signed_util.go, line 38 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46d5-XNrQMYisbuVl5:-L47-kYymRQgVn_y-ads:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L38)):*

Suggest changing `SigChecker` to `Verifier`, so it's more intuitive for external packages to refer to `ctrl.Signer` and `ctrl.Verifier`.
</blockquote></details>

Done.

---

*[go/lib/ctrl/signed_util.go, line 39 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46VjmPZIKOB3x-cLdZ:-L47-lLUYLGHJjdmZy1v:b-271o07) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L39)):*

I'm not sure I understand what the meaning behind this type is. To me it looks like it is a variant of `SigChecker` with a bound trust store, but it is used after some checks in `MkSigchecker`. Where does the job of `SigChecker` internals stop, and the job of `SigCheckerInner` start?
</blockquote></details>

I've rewritten this, hopefully it's clearer now.

---

*[go/lib/ctrl/signed_util.go, line 41 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46ScDoga6GYcq8adXe:-L47-purcdE4U6ntylcq:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L41)):*

The purpose of `sigCin` is not evident to me. A very detailed docstring is needed here, s.t. users of package `ctrl` have an intuition on the differences between `SigChecker` and `SigCheckerInner`.
</blockquote></details>

Done.

---

*[go/lib/ctrl/signed_util.go, line 41 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46Y9sqGyunQWaOPPon:-L47-qEy70Kowgjf-MZp:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L41)):*

`MkSigChecker`
</blockquote></details>

Done.

---

*[go/lib/ctrl/signed_util.go, line 94 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46_ct33uPO-Mklx2QI:-L47-qw7ee0zcA3Lxivg:b-9x38du) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/signed_util.go#L94)):*

This is a bit too acronym-y for my tasted. `MkSignedPld`is reasonably short. Also, given how often I see `ctrl.MkSPld(..., ctrl.NullSigner)` it probably would make sense to have something like `MkUnsignedPld`, even though it might be confusing that it returns an object of `SignedPld`. I guess the correct term would be `MkUnsignedSignedPld`, and the correct name for `SignedPld` would be `PossiblySignedPld` ;)
</blockquote></details>

I realised that if Plder is no longer an interface, there's no need for this to be a function, and so i just added a `SignedPld` method to `Pld` instead that takes a Signer`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1385)*

shitz:
:lgtm:

---

Reviewed 8 of 8 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1385#-:-L47z1Gs6DxB06uIpaiA:bnfp4nl)*

kormat:

---

*[go/lib/ctrl/ctrl_msg/msg.go, line 15 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46eDH0041pSOqiXHPq:-L4AbFsx5ZLjci6rrBC0:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl_msg/msg.go#L15)):*

This needs a docstring, as it's not immediately clear what the package offers (it's more related to signing and verifying, than actually messages).
</blockquote></details>

Done.

---

*[go/lib/ctrl/ctrl_msg/msg.go, line 78 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46ffeeiplQEa1sBu1I:-L4AbK7G7-O5zxvpPxWb:bjllu3h) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl_msg/msg.go#L78)):*

I think the callback pattern is a bit more confusing than just calling a helper method to set up `spld` at the start of both `Notify` and `NotifyUnreliable`, and then calling the correct function. The reason being that to understand`Notify` and `NotifyUnreliable` a programmer needs to look into `notify`, and the other way around. In any case, I guess it's about preference here, and I prefer only using callbacks only when there are no other solutions.
</blockquote></details>

Given that the entire 3 methods are 15 lines in total, i don't really see much room for confusion in this case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1385)*

scrye:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---

*[go/lib/ctrl/ctrl_msg/msg.go, line 78 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46ffeeiplQEa1sBu1I:-L4AcBbDt-FSrt7vGpej:bb26lb4) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl_msg/msg.go#L78)):*

Given that the entire 3 methods are 15 lines in total, i don't really see much room for confusion in this case.
</blockquote></details>

Exactly, using function pointers (and additional types) for something so simple feels overkill. It's a matter of preference/coding style, so it doesn't need to block the merge.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1385#-:-L4Aep6AyK6ZuJ_xRW26:bnfp4nl)*

scrye:

---

*[go/lib/ctrl/ctrl_msg/msg.go, line 78 at r1](https://reviewable.io:443/reviews/scionproto/scion/1385#-L46ffeeiplQEa1sBu1I:-L4AfBKnbhZtmBK06WRo:bm2ydin) ([raw file](https://github.com/scionproto/scion/blob/68b91a74ea7b71ac4145767ad77f2b3ebfe500da/go/lib/ctrl/ctrl_msg/msg.go#L78)):*

Exactly, using function pointers (and additional types) for something so simple feels overkill. It's a matter of preference/coding style, so it doesn't need to block the merge.
</blockquote></details>

FYI Switch test

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1385)*

scrye:
:lgtm:

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1385#-:-L4Akgn-jbtrhdhhZknG:bnfp4nl)*

PR #1387, opened 2018-01-31
Don't use a reference to a loop var

kormat:
This caused a very subtle bug in the SIG that only manifests once
you have multiple remote ASes in the config. All ASes configured in a
single config load would end up with the same IA in some parts of the
logic, leading to utter sanity loss.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1387)


scrye:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1387#-:-L4C8b5A-DU9NdJ95KXi:bnfp4nl)*

PR #1388, opened 2018-02-01
Include address and path header lengths in SIG frame size calculation.

kormat:
The current SIG code assumes that the address+path headers are <= 40B
long. This only happens to work because the existing environmets have a
single hop involving a core AS, and using IPv4 addresses. This fixes the
SIG to calculate the actual address+path header sizes, and use those.

N.b. it still does not include any extensions that might be in use. That
will eventually depend on snet providing that information.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1388)

### DIFF
--- a/go/lib/spkt/spkt.go
+++ b/go/lib/spkt/spkt.go
@@ -85,8 +85,7 @@ func (s *ScnPkt) Reverse() error {
 }
 
 func (s *ScnPkt) AddrLen() int {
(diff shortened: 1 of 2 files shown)


shitz:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1388#-:-L4Gn6Xskll7KQFpxtCe:bnfp4nl)*

PR #1398, opened 2018-02-03
Handle partial IFState updates in the BR.

### LINKED ISSUE #1395: IFStateInfo handling in BR needs update
Currently the BR assumes that every IFStateInfos packet from the BS contains all interfaces. However, the BS currently does not include all IFs in every update, and neither would this actually scale (an AS with say 3k interfaces would be sending a Huge amount of traffic in IFStateInfos to BRs). This makes the BR's view of interface state inconsistent/unreliable.

To fix this the following changes are needed:
- The BR no longer assumes that IFStateInfos are complete, and instead just applies the state they contain, rather than replacing all existing state.
- When the BR reloads the topology and notices an interface being removed in the local AS, also remove it from the IFState map the BR maintains. This _probably_ isn't needed when an interface is added to the AS, as what the BR really cares about is whether an interface is revoked or not.

shitz:
Fixes #1395

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1398)

### DIFF
--- a/go/border/ifstate/ifstate.go
+++ b/go/border/ifstate/ifstate.go
@@ -108,8 +108,7 @@ func Process(ifStates *path_mgmt.IFStateInfos) {
 		s, ok := states.Load(ifid)
 		if !ok {
 			log.Info("IFState: intf added", "ifid", ifid, "active", info.Active)
-			s = &state{}
-			atomic.StorePointer(&s.info, unsafe.Pointer(stateInfo))
+			s = &state{info: unsafe.Pointer(stateInfo)}
 			states.Store(ifid, s)
 			continue
 		}
--- a/go/border/ifstate/ifstate.go
+++ b/go/border/ifstate/ifstate.go
@@ -28,6 +28,8 @@ import (
 	"sync/atomic"
 	"unsafe"
 
+	"github.com/prometheus/client_golang/prometheus"
+
 	log "github.com/inconshreveable/log15"
 
 	"github.com/scionproto/scion/go/border/metrics"
@@ -58,15 +60,35 @@ func (s *ifStates) Store(key common.IFIDType, val *state) {
 var states ifStates
 
 type state struct {
+	// info is a pointer to an Info object.
 	info unsafe.Pointer
 }
 
 // Info stores state information, as well as the raw revocation info for a given interface.
 type Info struct {
-	IfID    common.IFIDType
-	Active  bool
-	RevInfo *path_mgmt.RevInfo
-	RawRev  common.RawBytes
+	IfID         common.IFIDType
+	Active       bool
+	RevInfo      *path_mgmt.RevInfo
+	RawRev       common.RawBytes
+	ActiveMetric prometheus.Gauge
+}
+
+func NewInfo(ifID common.IFIDType, active bool, rev *path_mgmt.RevInfo,
+	rawRev common.RawBytes) *Info {
+	i := &Info{
+		IfID:         ifID,
+		Active:       active,
+		RevInfo:      rev,
+		RawRev:       rawRev,
+		ActiveMetric: metrics.IFState.WithLabelValues(fmt.Sprintf("intf:%d", ifID)),
+	}
+	var isActive float64
+	if active {
+		isActive = 1
+	}
+	i.ActiveMetric.Set(isActive)
+
+	return i
 }
 
 // Process processes Interface State updates from the beacon service.
@@ -82,25 +104,23 @@ func Process(ifStates *path_mgmt.IFStateInfos) {
 				return
 			}
 		}
-		stateInfo := &Info{IfID: ifid, Active: info.Active, RevInfo: info.RevInfo, RawRev: rawRev}
-		gauge := metrics.IFState.WithLabelValues(fmt.Sprintf("intf:%d", ifid))
+		stateInfo := NewInfo(ifid, info.Active, info.RevInfo, rawRev)
 		s, ok := states.Load(ifid)
 		if !ok {
 			log.Info("IFState: intf added", "ifid", ifid, "active", info.Active)
 			s = &state{}
+			atomic.StorePointer(&s.info, unsafe.Pointer(stateInfo))
 			states.Store(ifid, s)
+			continue
+		}
+		oldInfo := (*Info)(atomic.LoadPointer(&s.info))
+		if stateInfo.Active {
+			if !oldInfo.Active {
+				log.Info("IFState: intf activated", "ifid", ifid)
+			}
 		} else {
-			oldInfo := (*Info)(atomic.LoadPointer(&s.info))
-			if stateInfo.Active {
-				if !oldInfo.Active {
-					log.Info("IFState: intf activated", "ifid", ifid)
-				}
-				gauge.Set(1)
-			} else {
-				if oldInfo.Active {
-					log.Info("IFState: intf deactivated", "ifid", ifid)
-				}
-				gauge.Set(0)
+			if oldInfo.Active {
+				log.Info("IFState: intf deactivated", "ifid", ifid)
 			}
 		}
 		atomic.StorePointer(&s.info, unsafe.Pointer(stateInfo))
--- a/go/border/ifstate/ifstate.go
+++ b/go/border/ifstate/ifstate.go
@@ -25,6 +25,8 @@ package ifstate
 import (
 	"fmt"
 	"sync"
+	"sync/atomic"
+	"unsafe"
 
 	log "github.com/inconshreveable/log15"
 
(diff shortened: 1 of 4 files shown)


kormat:
Reviewed 3 of 3 files at r1.

---

*[go/border/rpkt/path.go, line 114 at r1](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4adLA1b13sq1hgTRGf:-L4adLA1b13sq1hgTRGg:bn5qjj9) ([raw file](https://github.com/scionproto/scion/blob/64191ac9b24c5757cb478eb9a0abbb6ab3b2512f/go/border/rpkt/path.go#L114)):*
> ```Go
> 		// If the BR does not have a revocation for the current epoch, it considers
> 		// the interface as active until it receives a new revocation.
> 		state.Info.Active = true
> ```

Oh, damn. This is racy (and in the original version, too). I'm not sure what a good fix is, here. We might need to reconsider the design.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1398)*

kormat:
This does not address the second item listed in #1395:

> When the BR reloads the topology and notices an interface being removed in the local AS, also remove it from the IFState map the BR maintains. This probably isn't needed when an interface is added to the AS, as what the BR really cares about is whether an interface is revoked or not.

It's not as important as the first item, but it's still good for state hygiene. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1398#-:-L4ajBtm5aJGO0UgJuuf:b-35v778)*

shitz:
Interfaces are now removed from the ifstate map.

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/border/rpkt/path.go, line 114 at r1](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4adLA1b13sq1hgTRGf:-L4bf95GHhRK2ZRgi7eF:bl3xnao) ([raw file](https://github.com/scionproto/scion/blob/64191ac9b24c5757cb478eb9a0abbb6ab3b2512f/go/border/rpkt/path.go#L114)):*

Oh, damn. This is racy (and in the original version, too). I'm not sure what a good fix is, here. We might need to reconsider the design.
</blockquote></details>

Updated as discussed offline.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1398#-:-L4bf06jNkpEQHZDozZq:b-iaeuxg)*

kormat:
Reviewed 4 of 4 files at r2.

---

*[go/border/setup.go, line 217 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4eYodx5jvfs7TsmAYw:-L4eYodyYKz63LomMQdv:b-gu3qwr) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/setup.go#L217)):*
> ```Go
> 				sock.Stop()
> 				// Remove the interface from the ifstate map.
> 				ifstate.DeleteState(ifid)
> ```

This only removes state for interfaces local to the BR, not for other interfaces in the same AS.

---

*[go/border/ifstate/ifstate.go, line 61 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4exFjfSoj0lWA6ZHDj:-L4exFjfSoj0lWA6ZHDk:bk07k9o) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/ifstate/ifstate.go#L61)):*
> ```Go
> 
> type state struct {
> 	info unsafe.Pointer
> ```

I would use `*Info`, as it's more obvious what the intended contents are. Yes, this means doing the unsafe.Pointer casts while using the `atomic.*Pointer` functions, but i think it's worth it.

---

*[go/border/ifstate/ifstate.go, line 86 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4eyweF0dMjuiIbjPlO:-L4eyweG2-zlad33sfCS:b-2s0yzy) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/ifstate/ifstate.go#L86)):*
> ```Go
> 		}
> 		stateInfo := &Info{IfID: ifid, Active: info.Active, RevInfo: info.RevInfo, RawRev: rawRev}
> 		gauge := metrics.IFState.WithLabelValues(fmt.Sprintf("intf:%d", ifid))
> ```

Now that we have a dedicated `Info` type, let's add the metric to it, so we don't have to construct it every time.

---

*[go/border/ifstate/ifstate.go, line 90 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4ewQyaU4-Asa-07B8B:-L4ewQybXmlSyyaGeXEy:bitpuva) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/ifstate/ifstate.go#L90)):*
> ```Go
> 		if !ok {
> 			log.Info("IFState: intf added", "ifid", ifid, "active", info.Active)
> 			s = &state{}
> ```

Why not initialise this fully? Otherwise you're adding an entry to the map with no information in it, and then adding the information later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1398)*

shitz:
Review status: 1 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/border/setup.go, line 217 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4eYodx5jvfs7TsmAYw:-L4gjSO-hcsD_Sn6D9jl:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/setup.go#L217)):*

This only removes state for interfaces local to the BR, not for other interfaces in the same AS.
</blockquote></details>

Done.

---

*[go/border/ifstate/ifstate.go, line 61 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4exFjfSoj0lWA6ZHDj:-L4fYRfbIUPvPL4Qo8_R:bjvyhdu) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/ifstate/ifstate.go#L61)):*

I would use `*Info`, as it's more obvious what the intended contents are. Yes, this means doing the unsafe.Pointer casts while using the `atomic.*Pointer` functions, but i think it's worth it.
</blockquote></details>

The thing is that I cannot do the casting inline. It ends up being
```
p := unsafe.Pointer(s.info)
i := (Info*)(atomic.LoadPointer(&p))
```
which isn't worthwhile imho. Added a docstring  to info instead.

---

*[go/border/ifstate/ifstate.go, line 86 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4eyweF0dMjuiIbjPlO:-L4fhmLasaKj7J5DnB6U:b-axckyk) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/ifstate/ifstate.go#L86)):*

Now that we have a dedicated `Info` type, let's add the metric to it, so we don't have to construct it every time.
</blockquote></details>

The thing is that we have to construct a new `Info` object anyway, since you don't want to modify an existing `Info` in place. Anyway, it made me realise that the metric should always be kept in sync with the `Active` flag.

---

*[go/border/ifstate/ifstate.go, line 90 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4ewQyaU4-Asa-07B8B:-L4flHSR2ChBd1mz3cRz:baqqkmr) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/ifstate/ifstate.go#L90)):*

Why not initialise this fully? Otherwise you're adding an entry to the map with no information in it, and then adding the information later.
</blockquote></details>

To save one line of code I guess ;) Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1398)*

kormat:
Reviewed 3 of 3 files at r3.

---

*[go/border/ifstate/ifstate.go, line 61 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4exFjfSoj0lWA6ZHDj:-L4kzBLeFuIR7VPzzzeH:b-khe2z9) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/ifstate/ifstate.go#L61)):*

The thing is that I cannot do the casting inline. It ends up being
```
p := unsafe.Pointer(s.info)
i := (Info*)(atomic.LoadPointer(&p))
```
which isn't worthwhile imho. Added a docstring  to info instead.
</blockquote></details>

Yup, agreed. I hadn't thought about it being an `*unsafe.Pointer`, and not just an `unsafe.Pointer`. I'm not even sure that workaround is safe.

:+1: 

---

*[go/border/ifstate/ifstate.go, line 90 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4ewQyaU4-Asa-07B8B:-L4l-SCTuYTDwhnKMtDy:bwqg9un) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/ifstate/ifstate.go#L90)):*

To save one line of code I guess ;) Done.
</blockquote></details>

You don't need to use `atomic` for non-shared addresses. `s = &state{info: unsafe.Pointer(stateInfo)}` is perfectly valid (and stops people like me from wondering what they're missing :).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1398)*

shitz:
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/border/ifstate/ifstate.go, line 90 at r2](https://reviewable.io:443/reviews/scionproto/scion/1398#-L4ewQyaU4-Asa-07B8B:-L4q3wb6FCB2DY_FbZmT:b-g7u1a8) ([raw file](https://github.com/scionproto/scion/blob/9cb1e6a1676159c47c6525abf7f9d591923f425c/go/border/ifstate/ifstate.go#L90)):*

You don't need to use `atomic` for non-shared addresses. `s = &state{info: unsafe.Pointer(stateInfo)}` is perfectly valid (and stops people like me from wondering what they're missing :).
</blockquote></details>

Silly me. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1398)*

kormat:
:lgtm:

---

Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1398#-:-L4utpWNl7uBSSACxu3k:bnfp4nl)*

shitz:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1398)*

PR #1401, opened 2018-02-06
Add cert reissue req/rep

Oncilla:
Add certificate chain reissuance request and response.
The requests consists of the desired leaf certificate.
The response consists of the reissued certificate chain.

Request signatures are added to signed payload.
Code for signing an verifying requests will follow in a future PR.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1401)

### DIFF
--- a/go/cert_srv/io.go
+++ b/go/cert_srv/io.go
@@ -93,9 +93,12 @@ func (d *Dispatcher) dispatch(addr *snet.Addr, buf common.RawBytes) error {
 			d.trcHandler.HandleRep(addr, pld.(*cert_mgmt.TRC))
 		case proto.TRCReq_TypeID:
 			d.trcHandler.HandleReq(addr, pld.(*cert_mgmt.TRCReq))
+		default:
+			return common.NewBasicError("Handler for cert_mgmt.pld not implemented", nil,
+				"protoID", pld.ProtoId())
 		}
 	default:
-		return common.NewBasicError("Not implemented", nil, "protoID", c.ProtoId())
+		return common.NewBasicError("Handler for cpld not implemented", nil, "protoID", c.ProtoId())
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


kormat:
Reviewed 5 of 5 files at r1.

---

*[go/cert_srv/io.go, line 87 at r1](https://reviewable.io:443/reviews/scionproto/scion/1401#-L4pUzhetCGaKtRnQRec:-L4pUzhetCGaKtRnQRed:b-gtwhzg) ([raw file](https://github.com/scionproto/scion/blob/06a394e42d0ec086d8632901f601e87ff366823a/go/cert_srv/io.go#L87)):*
> ```Go
> 			return err
> 		}
> 		switch pld.ProtoId() {
> ```

I only just caught this, but this isn't a valid approach. It's legal to have multiple entries in a union which have the same proto ID, but different names. E.g. `PathMgmt.segReg` vs `PathMgmt.segSync`. Instead, our go code has a seperate _go_ type per union member. Have a look at `go/lib/ctrl/path_mgmt/seg_recs.go` as an example. This means that instead of switching on `pld.ProtoId()`, you should switch on `pld.(type)`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1401)*

Oncilla:

---

*[go/cert_srv/io.go, line 87 at r1](https://reviewable.io:443/reviews/scionproto/scion/1401#-L4pUzhetCGaKtRnQRec:-L6GXqYO9PHLLKJBwvjW:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/06a394e42d0ec086d8632901f601e87ff366823a/go/cert_srv/io.go#L87)):*

I only just caught this, but this isn't a valid approach. It's legal to have multiple entries in a union which have the same proto ID, but different names. E.g. `PathMgmt.segReg` vs `PathMgmt.segSync`. Instead, our go code has a seperate _go_ type per union member. Have a look at `go/lib/ctrl/path_mgmt/seg_recs.go` as an example. This means that instead of switching on `pld.ProtoId()`, you should switch on `pld.(type)`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1401)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1401#-:-L6Mp03DpsP3OhQiYabI:bnfp4nl)*

PR #1403, opened 2018-02-06
Update python infra to use CtrlPld.reqId

kormat:
Also:
- Obsolete SegReq.id, as it's no longer useful.
- Removed a now-unsed random_uint64() function from util, and made a
  similar one in ctrl_pld.py (which specifically avoids 0 as a value).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1403)

PR #1404, opened 2018-02-07
Default Messenger for SCION Infra

scrye:
Fixes some issues with Reviewable in https://github.com/scionproto/scion/pull/1402.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1404)


scrye:
Review status: 0 of 16 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/lib/infra/common.go, line 34 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jlZ1jZwgUZyjxr-SS:-L4jlZ1jZwgUZyjxr-ST:b1annts) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L34)):*
> @kormat: I would put this immediately after Handler, rather than after Request.

Done.

---

*[go/lib/infra/common.go, line 44 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jlJpbU9yOvx3kCyff:-L4jlJpcPmavjECW95rf:bloma7s) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L44)):*
> @kormat: I'm not sure that namespace-flattening is a descriptive term. I think this is more likely the inner capnp message. It would be good to give an example, in any case.

Done.

---

*[go/lib/infra/common.go, line 48 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jlNMuEB4VBhJEj-ox:-L4jlNMuEB4VBhJEj-oy:b-n4vbld) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L48)):*
> @kormat: Is that SignedCtrlPld, or CtrlPld?

SignedCtrlPld. Changed the docstring to reflect this.

---

*[go/lib/infra/common.go, line 68 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jlT46TDrIVQw3btui:-L4jlT46TDrIVQw3btuj:b-rql7za) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L68)):*
> @kormat: Why a getter, rather than exporting the context?

This makes the context read-only. It prevents callers from trying to manipulate the context while it is passed around (e.g., by handlers), either intentionally or by mistake.

---

*[go/lib/infra/messenger/adapter.go, line 67 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jldrOpNKZp7Nx_tnS:-L4jldrOpNKZp7Nx_tnT:b27h4hw) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/messenger/adapter.go#L67)):*
> @kormat: Now that Ctrl.Pld.ReqId exists, this can be updated (even though there's nothing that sets that field, currently).

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

kormat:
Reviewed 18 of 20 files at r1, 1 of 1 files at r2.
Review status: 15 of 16 files reviewed at latest revision, 18 unresolved discussions.

---

*[go/lib/ctrl/signed_util.go, line 196 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kFDqfgPytyxmhaING:-L4kFDqfgPytyxmhaINH:b2iuo1o) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/ctrl/signed_util.go#L196)):*
> ```Go
> 
> func (_ *nullSigVerifier) Verify(p *SignedPld) error {
> 	// Run the payload extraction step only
> ```

Is this really needed? The only reason that `BasicSigVerifier` extracts the payload is so it can avoid cert_mgmt.Chain/TRC . The null verifier doesn't even care about that.

---

*[go/lib/infra/common.go, line 68 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jlT46TDrIVQw3btui:-L4k9hTuqTp9ea9SdZhd:b7tlyci) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L68)):*

> @kormat: Why a getter, rather than exporting the context?

This makes the context read-only. It prevents callers from trying to manipulate the context while it is passed around (e.g., by handlers), either intentionally or by mistake.
</blockquote></details>

Contexts _are_ read-only. They have no mutating methods.

---

*[go/lib/infra/common.go, line 102 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kAZ8wAEvFfsdw1Mg5:-L4kAZ8wAEvFfsdw1Mg6:bftyao1) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L102)):*
> ```Go
> 	NewPushTRCHandler() Handler
> 	NewPushChainHandler() Handler
> 	GetCertificate(ctx context.Context, trail []Descriptor, hint net.Addr) (*cert.Certificate, error)
> ```

Line length.

---

*[go/lib/infra/common.go, line 105 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kAm65wyhFEpDVpi0b:-L4kAm65wyhFEpDVpi0c:b-5f7ixu) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L105)):*
> ```Go
> }
> 
> type Descriptor struct {
> ```

This name is very generic, i have no idea what it's for.

---

*[go/lib/infra/disp/common_test.go, line 66 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kAy6AxlDS7cOTpAH5:-L4kAy6AxlDS7cOTpAH6:blj8jn4) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/disp/common_test.go#L66)):*
> ```Go
> }
> 
> func (c *customObject) String() string {
> ```

It's probably worth putting an interface assertion that `customObject` implements `proto.Cerealizable`.

---

*[go/lib/infra/messenger/adapter.go, line 39 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kBV54kgVihmiUqwLD:-L4kBV54kgVihmiUqwLE:bftyao1) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/messenger/adapter.go#L39)):*
> ```Go
> 	pld, ok := msg.(*ctrl.SignedPld)
> 	if !ok {
> 		return nil, common.NewBasicError("Unable to type assert proto.Cerealizable to ctrl.SignedPld",
> ```

Line length.

---

*[go/lib/infra/messenger/messenger.go, line 20 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kG8uneC9kX84Ulzn8:-L4kG8uneC9kX84Ulzn9:b-kov8aw) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L20)):*
> ```Go
> //
> // The following message types are valid messages. For messages marked (nv), no
> // signature verification is performed on the SignedPld.
> ```

Afaict, the message type is irrelevant, as `Messenger` hard-codes the `NullSigVerifier` anyway.

---

*[go/lib/infra/messenger/messenger.go, line 48 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kGXXFpWDr0d7L8yyt:-L4kGXXFpWDr0d7L8yyu:biioqra) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L48)):*
> ```Go
> // shared functionality. One handler can be registered for each message type,
> // identified by its msgType string:
> //   msger.AddHandler("ChainRequest", MyCustomHandler)
> ```

This should use the string constants.

---

*[go/lib/infra/messenger/messenger.go, line 67 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJwP_knaWq5-dzYTs:-L4kJwP_knaWq5-dzYTt:b7a7pn) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L67)):*
> ```Go
> //
> // CloseServer() does not do graceful shutdown (all handlers are canceled
> // immediately) and does not close the Messenger itself.
> ```

s/immediately//, as there's no guarantee it's actually _immediate_. Their context is cancelled immediately, but it can take a while before a handler notices.

---

*[go/lib/infra/messenger/messenger.go, line 113 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kH7m3J0go-sOCt9vY:-L4kH7m3J0go-sOCt9vZ:bhzmuve) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L113)):*
> ```Go
> 	handlers map[string]infra.Handler
> 
> 	lock      sync.Mutex
> ```

It looks like this only protects `closeChan`; renaming it to `closeLock` would make it more obvious.

---

*[go/lib/infra/messenger/messenger.go, line 157 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kILNUcOc9xYZuUyAk:-L4kILNUcOc9xYZuUyAl:b-5sn37j) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L157)):*
> ```Go
> 		return nil, err
> 	}
> ```

Unneeded blank line.

---

*[go/lib/infra/messenger/messenger.go, line 190 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kIP97GNX3gaI2D6um:-L4kIP97GNX3gaI2D6un:b-5sn37j) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L190)):*
> ```Go
> 		return nil, err
> 	}
> ```

Unneeded blank line.

---

*[go/lib/infra/messenger/messenger.go, line 223 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kIZWQjD-Xrafqh0UR:-L4kIZWQjD-Xrafqh0US:b-5sn37j) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L223)):*
> ```Go
> 		return nil, err
> 	}
> ```

Unneeded blank line.

---

*[go/lib/infra/messenger/messenger.go, line 310 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJ5khzfIJPMQ7BfAk:-L4kJ5khzfIJPMQ7BfAl:blrkyxq) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L310)):*
> ```Go
> 	}
> 	serveCtx := context.WithValue(m.ctx, infra.MessengerContextKey, m)
> 	// XXX(scrye): The handler might perform additional verifications; for
> ```

What's the purpose of this comment? A handler can presumably do pretty much anything, so singling this out seems a bit odd to me.

---

*[go/lib/infra/messenger/messenger.go, line 313 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJNpEP3DuKFd-XlSp:-L4kJNpEP3DuKFd-XlSq:b-o257jh) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L313)):*
> ```Go
> 	// example, a PCB handler will probably verify additional signatures.
> 	request := infra.NewRequest(serveCtx, msg, pld, address)
> 	go handler.Handle(request)
> ```

This can just become `go handler.Handle(infra.NewRequest(...))`.

---

*[go/lib/infra/messenger/messenger.go, line 318 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJVH4W-cD2lK9AI4B:-L4kJVH4W-cD2lK9AI4C:b-vuxla6) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L318)):*
> the object containing only the payload

Might be clearer to say "the inner proto.Cerealizable object"?

---

*[go/lib/infra/messenger/messenger.go, line 336 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJfmuyH83jan_-baD:-L4kJfmuyH83jan_-baE:bmtyb19) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L336)):*
> ```Go
> 		default:
> 			return "", nil,
> 				common.NewBasicError("Unsupported SignedPld.CtrlPld.CertMgmt.Xxx message type", nil)
> ```

This should include the type.

---

*[go/lib/infra/messenger/messenger.go, line 339 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJjM70b1LXKznzCn8:-L4kJjM8lzvZbdWx3xfD:bmtyb19) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L339)):*
> ```Go
> 		}
> 	default:
> 		return "", nil, common.NewBasicError("Unsupported SignedPld.Pld.Xxx message type", nil)
> ```

This should include the type.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

kormat:
Review status: 15 of 16 files reviewed at latest revision, 19 unresolved discussions.

---

*[go/lib/infra/messenger/messenger.go, line 27 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kKQCbwv7Bs7ARQVu7:-L4kKQCbwv7Bs7ARQVu8:bthfc5j) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L27)):*
> ```Go
> //  PathRequest  -> ctrl.SignedPld/ctrl.Pld/path_mgmt.SegReq   (nv)
> //
> // Unsupported messages are returned by Messenger.RecvMsg(), but if the
> ```

`RecvMsg` no longer exists (same 2 paragraphs below).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

kormat:
Review status: 15 of 16 files reviewed at latest revision, 20 unresolved discussions.

---

*[go/lib/infra/messenger/messenger.go, line 174 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kKwtJNdH4-Lfmzv8S:-L4kKwtJNdH4-Lfmzv8T:b-7a626j) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L174)):*
> ```Go
> 
> // SendTRC sends a reliable cert_mgmt.TRC to address a.
> func (m *Messenger) SendTRC(ctx context.Context, msg *cert_mgmt.TRC, a net.Addr) error {
> ```

This doesn't have a way of setting the CtrlPld.ReqId, meaning it can't be used to respond to queries. (Same for `SendCertChain`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

kormat:
Reviewed 1 of 20 files at r1.
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

kormat:
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[go/lib/infra/messenger/messenger_test.go, line 78 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kMAGwMmtn9GmXWqZ2:-L4kMAGwMmtn9GmXWqZ3:b-ed6xw9) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger_test.go#L78)):*
> ```Go
> 			// The client sends a TRC request to the server, and receives the
> 			// TRC.
> 			ctx, cancelF := context.WithTimeout(context.Background(), 3*time.Second)
> ```

There's something wrong with timeouts here, somewhere. If i comment out the `AddHandler` below, the unit test never exits.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

kormat:
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[go/lib/infra/messenger/messenger_test.go, line 78 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kMAGwMmtn9GmXWqZ2:-L4kMg5cFtpL-O-HUusk:bia556n) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger_test.go#L78)):*

There's something wrong with timeouts here, somewhere. If i comment out the `AddHandler` below, the unit test never exits.
</blockquote></details>

Ah. If you move the `CloseServer` to directly after `GetTRC`, then it works.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

scrye:
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[go/lib/ctrl/signed_util.go, line 196 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kFDqfgPytyxmhaING:-L4kWM8thDhopKIMykww:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/ctrl/signed_util.go#L196)):*

Is this really needed? The only reason that `BasicSigVerifier` extracts the payload is so it can avoid cert_mgmt.Chain/TRC . The null verifier doesn't even care about that.
</blockquote></details>

Done.

---

*[go/lib/infra/common.go, line 68 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jlT46TDrIVQw3btui:-L4kMocLN1penWSTr7bu:b-3bcjyc) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L68)):*

Contexts _are_ read-only. They have no mutating methods.
</blockquote></details>

If you expose the member, it can get rewritten. This prevents that. It's the same reason why `http.Request.Context()` exists.

---

*[go/lib/infra/common.go, line 102 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kAZ8wAEvFfsdw1Mg5:-L4kNAB5Z--y-0XytICH:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L102)):*

Line length.
</blockquote></details>

Done.

---

*[go/lib/infra/common.go, line 105 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kAm65wyhFEpDVpi0b:-L4kN8hwZ-qrNGjdJQqZ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L105)):*

This name is very generic, i have no idea what it's for.
</blockquote></details>

Done.

---

*[go/lib/infra/disp/common_test.go, line 66 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kAy6AxlDS7cOTpAH5:-L4kNO2oR5u2bGmsApC6:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/disp/common_test.go#L66)):*

It's probably worth putting an interface assertion that `customObject` implements `proto.Cerealizable`.
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/adapter.go, line 39 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kBV54kgVihmiUqwLD:-L4kNT_TRSNDgKaTKjwY:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/messenger/adapter.go#L39)):*

Line length.
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 20 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kG8uneC9kX84Ulzn8:-L4kNZx-s2CzBrXbcr4-:b5mbm17) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L20)):*

Afaict, the message type is irrelevant, as `Messenger` hard-codes the `NullSigVerifier` anyway.
</blockquote></details>

It's hard-coded for now, but it is not guaranteed to stay that way. I think it's useful if the documentation allows for the possibility of it changing at some point in the future.

---

*[go/lib/infra/messenger/messenger.go, line 27 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kKQCbwv7Bs7ARQVu7:-L4kNo0bgJLhJvXKd0MK:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L27)):*

`RecvMsg` no longer exists (same 2 paragraphs below).
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 48 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kGXXFpWDr0d7L8yyt:-L4kNtSVGFu83cvxxtOb:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L48)):*

This should use the string constants.
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 67 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJwP_knaWq5-dzYTs:-L4kO1gtbVYgSv11ERcY:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L67)):*

s/immediately//, as there's no guarantee it's actually _immediate_. Their context is cancelled immediately, but it can take a while before a handler notices.
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 113 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kH7m3J0go-sOCt9vY:-L4kO57qkFRnYTp74QXP:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L113)):*

It looks like this only protects `closeChan`; renaming it to `closeLock` would make it more obvious.
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 157 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kILNUcOc9xYZuUyAk:-L4kODTpw7XDsIbAahR5:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L157)):*

Unneeded blank line.
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 174 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kKwtJNdH4-Lfmzv8S:-L4kOFwIpMRYXbCbC9AZ:b-mkz71u) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L174)):*

This doesn't have a way of setting the CtrlPld.ReqId, meaning it can't be used to respond to queries. (Same for `SendCertChain`.
</blockquote></details>

Yes, this is by design. This method is used for initiating an exchange, so it can choose a `CtrlPld.ReqId` internally. `ListenAndServe` registered handlers are the ones that have to care about responding with the same ID.

Also added support for request IDs.

For explicit sends, apps should use the helper functions directly (since they're basically doing manual management of responses at that point).

---

*[go/lib/infra/messenger/messenger.go, line 190 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kIP97GNX3gaI2D6um:-L4kSbN3gnCF05-roJHU:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L190)):*

Unneeded blank line.
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 223 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kIZWQjD-Xrafqh0UR:-L4kSd5cjtZqw2cLvOWa:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L223)):*

Unneeded blank line.
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 310 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJ5khzfIJPMQ7BfAk:-L4kSk4xmhu7TkI7WL4P:br3cd0a) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L310)):*

What's the purpose of this comment? A handler can presumably do pretty much anything, so singling this out seems a bit odd to me.
</blockquote></details>

Deprecated, removed.

---

*[go/lib/infra/messenger/messenger.go, line 313 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJNpEP3DuKFd-XlSp:-L4kSsHdmDg-vQZqitsi:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L313)):*

This can just become `go handler.Handle(infra.NewRequest(...))`.
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 318 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJVH4W-cD2lK9AI4B:-L4kT64U33jk-YCoSdYE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L318)):*

> the object containing only the payload

Might be clearer to say "the inner proto.Cerealizable object"?
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 336 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJfmuyH83jan_-baD:-L4kTl_DOZmnjapGQotW:b-qldeod) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L336)):*

This should include the type.
</blockquote></details>

Because the code determines the type based on `which` values, I've included that in the error.

---

*[go/lib/infra/messenger/messenger.go, line 339 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kJjM70b1LXKznzCn8:-L4kU35uHcdyonTmpnwl:b2nreh6) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L339)):*

This should include the type.
</blockquote></details>

Same as above.

---

*[go/lib/infra/messenger/messenger_test.go, line 78 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kMAGwMmtn9GmXWqZ2:-L4kUVkvOduKtU50SkKk:b7iqwub) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger_test.go#L78)):*

Ah. If you move the `CloseServer` to directly after `GetTRC`, then it works.
</blockquote></details>

The anomaly is due to `sc.SoMsg` failing the first assertion and giving up on continuing the goroutine, so `CloseServer` never gets called.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

kormat:
Reviewed 6 of 6 files at r3.

---

*[go/lib/ctrl/signed_util.go, line 196 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kFDqfgPytyxmhaING:-L4kX8z-z_rbXFWLxcKc:b-ck8yt9) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/ctrl/signed_util.go#L196)):*

Done.
</blockquote></details>

The type docstring needs to be updated too.

---

*[go/lib/infra/common.go, line 68 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jlT46TDrIVQw3btui:-L4kXDdC2GGg6dXXk2yD:b-p8rn52) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L68)):*

If you expose the member, it can get rewritten. This prevents that. It's the same reason why `http.Request.Context()` exists.
</blockquote></details>

The same applies to all members we expose in pretty much every struct. Either we say this isn't a big concern, or we add lots and lots of getters (and maybe even setters) and a lot of noise to the code. I'm in the former camp on this one.

---

*[go/lib/infra/messenger/adapter.go, line 39 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kBV54kgVihmiUqwLD:-L4kYhsXhzgb6uKYz7Mf:byd2ex1) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/messenger/adapter.go#L39)):*

Done.
</blockquote></details>

Ah, it's much better to move the entire string to a new line rather than splitting it like that.

---

*[go/lib/infra/messenger/messenger.go, line 20 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kG8uneC9kX84Ulzn8:-L4kZ2Bba0oLm45Z7JBM:bnjzx69) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L20)):*

It's hard-coded for now, but it is not guaranteed to stay that way. I think it's useful if the documentation allows for the possibility of it changing at some point in the future.
</blockquote></details>

It would be more straight-forward to just mention in the documentation that currently no signing or signature verification is done. When that changes, the docs can be updated.

---

*[go/lib/infra/messenger/messenger.go, line 174 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kKwtJNdH4-Lfmzv8S:-L4kalBinyxHYq3SHe-B:b-pebpkj) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L174)):*

Yes, this is by design. This method is used for initiating an exchange, so it can choose a `CtrlPld.ReqId` internally. `ListenAndServe` registered handlers are the ones that have to care about responding with the same ID.

Also added support for request IDs.

For explicit sends, apps should use the helper functions directly (since they're basically doing manual management of responses at that point).
</blockquote></details>

Ok, at this point i'm really not seeing what use this library is. An endhost app _might_ use it, but i don't see why any infra server would use it. If you look at PR #1403 , you'll see that it updates most places that send/receive requests and send replies to use explicit request IDs so that they can be logged by the app, and correlated between requestor/requestee. This makes it possible to debug issues across systems, and is something we want all infra services to be able to do (whether that logging is on by default or not is a separate question). Hiding even this basic functionality means we should not use this library.

---

*[go/lib/infra/messenger/messenger.go, line 352 at r3](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4keMs1g1Km_P7QFvZ-:-L4keMs1g1Km_P7QFvZ0:b86t8sr) ([raw file](https://github.com/scionproto/scion/blob/15fedf5f22e9266073d975e622386ed3530d184b/go/lib/infra/messenger/messenger.go#L352)):*
> ```Go
> 
> // nextID returns a unique CtrlPld Request ID.
> func (m *Messenger) nextID() uint64 {
> ```

If this is going to be a sequential ID, can it at least be initialised to a random value? Otherwise it will be very hard to use for correlating requests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

scrye:

---

*[go/lib/ctrl/signed_util.go, line 196 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kFDqfgPytyxmhaING:-L4kjKOeEAtvPHYmbEZk:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/ctrl/signed_util.go#L196)):*

The type docstring needs to be updated too.
</blockquote></details>

Done.

---

*[go/lib/infra/common.go, line 68 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jlT46TDrIVQw3btui:-L4kjP3GlHVIJYNcRGIk:b1zb51a) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L68)):*

The same applies to all members we expose in pretty much every struct. Either we say this isn't a big concern, or we add lots and lots of getters (and maybe even setters) and a lot of noise to the code. I'm in the former camp on this one.
</blockquote></details>

Contexts are designed around getting passed around a lot, and interacting with them. Handlers might think, "oh, I need to pass this Request along, I'll just put my own context in it, and change it back when I get control again". This just stops all of this. They are also designed to be passed by value to functions, and this mimics that access model.

It's not something we need to do for every member, but it makes sense for this one.

---

*[go/lib/infra/messenger/messenger.go, line 20 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kG8uneC9kX84Ulzn8:-L4kj8n7wXe5OK9NHo_H:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L20)):*

It would be more straight-forward to just mention in the documentation that currently no signing or signature verification is done. When that changes, the docs can be updated.
</blockquote></details>

Done.

---

*[go/lib/infra/messenger/messenger.go, line 174 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kKwtJNdH4-Lfmzv8S:-L4kg3NYRpE4L4Kw6QlB:bm44ent) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L174)):*

Ok, at this point i'm really not seeing what use this library is. An endhost app _might_ use it, but i don't see why any infra server would use it. If you look at PR #1403 , you'll see that it updates most places that send/receive requests and send replies to use explicit request IDs so that they can be logged by the app, and correlated between requestor/requestee. This makes it possible to debug issues across systems, and is something we want all infra services to be able to do (whether that logging is on by default or not is a separate question). Hiding even this basic functionality means we should not use this library.
</blockquote></details>

But the correlation is already done for you, behind the scenes, by the Dispatcher. The app will see "I've sent this out, and I got this answer to my exact request". It shouldn't need to juggle request IDs for this. That's messy, and it leads to more useless bookkeeping by each app.

I've added `id` parameters (for now), because it was simpler to have the server respond through the same calls. But I think that finding a solution in the future where this no longer needs any intervention from the app would be ideal.

---

*[go/lib/infra/messenger/messenger.go, line 352 at r3](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4keMs1g1Km_P7QFvZ-:-L4kgPj-Krh_iMs5wEJ4:b-7dpnla) ([raw file](https://github.com/scionproto/scion/blob/15fedf5f22e9266073d975e622386ed3530d184b/go/lib/infra/messenger/messenger.go#L352)):*

If this is going to be a sequential ID, can it at least be initialised to a random value? Otherwise it will be very hard to use for correlating requests.
</blockquote></details>

IDs are now specified by the app.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

scrye:

---

*[go/lib/infra/messenger/adapter.go, line 39 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kBV54kgVihmiUqwLD:-L4krXq7llZpy5WIC8AZ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/messenger/adapter.go#L39)):*

Ah, it's much better to move the entire string to a new line rather than splitting it like that.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

kormat:
Reviewed 5 of 5 files at r4.

---

*[go/lib/infra/common.go, line 68 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jlT46TDrIVQw3btui:-L4ktwtq0_S3K91X8crr:blfpqk0) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L68)):*

Contexts are designed around getting passed around a lot, and interacting with them. Handlers might think, "oh, I need to pass this Request along, I'll just put my own context in it, and change it back when I get control again". This just stops all of this. They are also designed to be passed by value to functions, and this mimics that access model.

It's not something we need to do for every member, but it makes sense for this one.
</blockquote></details>

I really don't see it happening, and if it does, we have far bigger problems regarding the quality of our code.

If a handler changes the Id, or the Peer, things also break. I don't get why context is special.

I don't think that http.Request's handling of this is all that relevant, either. It's `Context()` method is a getter because it has to be; it returns `context.Background()` if there is no context. Plus there's some cleverness going on with `WithContext`, where the context is actually replaced. Neither of those apply in our case.

---

*[go/lib/infra/messenger/messenger.go, line 174 at r2](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4kKwtJNdH4-Lfmzv8S:-L4ksnKTehgeDxFTE-5N:b-yvkxf0) ([raw file](https://github.com/scionproto/scion/blob/1311d8760923ce2c45bd29d6e5d3642c35d42ec0/go/lib/infra/messenger/messenger.go#L174)):*

But the correlation is already done for you, behind the scenes, by the Dispatcher. The app will see "I've sent this out, and I got this answer to my exact request". It shouldn't need to juggle request IDs for this. That's messy, and it leads to more useless bookkeeping by each app.

I've added `id` parameters (for now), because it was simpler to have the server respond through the same calls. But I think that finding a solution in the future where this no longer needs any intervention from the app would be ideal.
</blockquote></details>

That's not what i was talking about. I'm talking about a human (or later, something like logdog) being able to look at a log file for the requestor, and a log file for the requestee, and see when the request was sent, when it was received, any logging info generated while handling the request, and the same for the response. The infra.dispatcher knowing to correlate things is completely orthogonal for this.

---

*[go/lib/infra/messenger/messenger_test.go, line 43 at r4](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4ksbrGci6dpSs3-ttc:-L4ksbrGci6dpSs3-ttd:b-w188va) ([raw file](https://github.com/scionproto/scion/blob/64fd92c62e18e7f235ec8e13f7754f991d3e311f/go/lib/infra/messenger/messenger_test.go#L43)):*
> ```Go
> 
> func MockTRCHandler(request *infra.Request) {
> 	fmt.Println("id = ", request.ID)
> ```

Forgotten debug statement? Unit tests shouldn't write to stdout.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

scrye:
Review status: 15 of 16 files reviewed at latest revision, 3 unresolved discussions.

---

*[go/lib/infra/common.go, line 68 at r1](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4jlT46TDrIVQw3btui:-L4kwDSxU0wv0Xp516Bt:b-xmbrhw) ([raw file](https://github.com/scionproto/scion/blob/9216e3908917512427462d268987194f9f7fa5f3/go/lib/infra/common.go#L68)):*

I really don't see it happening, and if it does, we have far bigger problems regarding the quality of our code.

If a handler changes the Id, or the Peer, things also break. I don't get why context is special.

I don't think that http.Request's handling of this is all that relevant, either. It's `Context()` method is a getter because it has to be; it returns `context.Background()` if there is no context. Plus there's some cleverness going on with `WithContext`, where the context is actually replaced. Neither of those apply in our case.
</blockquote></details>

Their `WithContext` is a workaround, s.t. when somebody changes the `context`, they better have a damn good reason.

Using a getter for this keeps the data access model similar to how it is when calling a method. You cannot change the context, it is always passed by value. I think this type of consistency is useful. It enforces a clean propagation of the context.

---

*[go/lib/infra/messenger/messenger_test.go, line 43 at r4](https://reviewable.io:443/reviews/scionproto/scion/1404#-L4ksbrGci6dpSs3-ttc:-L4kzaFzSvf_gaZMr0XE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/64fd92c62e18e7f235ec8e13f7754f991d3e311f/go/lib/infra/messenger/messenger_test.go#L43)):*

Forgotten debug statement? Unit tests shouldn't write to stdout.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1404#-:-L4l4wfW6CQc8ZmFskjm:bnfp4nl)*

PR #1413, opened 2018-02-21
pingpong: Accept multiple clients

FR4NK-W:
Accept each client in a separate go routine
Make read errors not LogFatal:
prevents a single '[CRIT] Unable to read err=EOF' from bringing down the server
only log it the faulty read and exit for the current routine

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1413)


FR4NK-W:
@sgmonroy Made the pingpong example server accept multiple clients

kormat:
Reviewed 1 of 1 files at r2.

---

*[go/examples/pingpong/pingpong.go, line 150 at r2](https://reviewable.io:443/reviews/scionproto/scion/1413#-L6QlBUKw7DRiGi-23wn:-L6QlBUKw7DRiGi-23wo:b-fs6i3r) ([raw file](https://github.com/scionproto/scion/blob/e9e4f47f88f79ce0768fc38c953de3b7c3f23ff7/go/examples/pingpong/pingpong.go#L150)):*
> ```Go
> 			}
> 			log.Error("Unable to read", "err", err)
> 			continue
> ```

This should probably kill the client session, too. If the server goes away, the client never recovers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1413)*

FR4NK-W:
@kormat Updated PR to add an optional timeout flag, keeping the same default.
Close client session on `NetworkIdleTimeout` as suggested.

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1413#-:-L6R_gHAStl0oNeXMQny:bnfp4nl)*

PR #1418, opened 2018-02-26
Implement infra Transport for normal UDP

scrye:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1418)


scrye:
Git diff is quite confused; `udp.go` is completely new, while `rudp.go` is an unmodified move of old `udp.go`.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1418#-:-L6HEpTthb-Hoz8uawQJ:b-ajj8f6)*

kormat:
Reviewed 5 of 5 files at r1.

---

*[go/lib/infra/transport/udp.go, line 35 at r1](https://reviewable.io:443/reviews/scionproto/scion/1418#-L6HH_PsXEeQSFb_TkFG:-L6HH_PsXEeQSFb_TkFH:b-5wqwv) ([raw file](https://github.com/scionproto/scion/blob/2cef448cff214835476283b791b6b508e5d283e2/go/lib/infra/transport/udp.go#L35)):*
> ```Go
> 	conn net.PacketConn
> 	// While conn is safe for use from multiple goroutines, deadlines are
> 	// global so it is not safe to enforce two at the same time. Thus, to
> ```

Can you clarify this a bit? An example scenario would help. It looks like if a short deadline read comes in while a long deadline read is pending, the new read would just block unconditionally until after the existing read finishes, which seems a bit self-defeating.

---

*[go/lib/infra/transport/udp.go, line 67 at r1](https://reviewable.io:443/reviews/scionproto/scion/1418#-L6HI1UDwXZxzsRTOhr_:-L6HI1UDwXZxzsRTOhra:bn3zbwn) ([raw file](https://github.com/scionproto/scion/blob/2cef448cff214835476283b791b6b508e5d283e2/go/lib/infra/transport/udp.go#L67)):*
> ```Go
> 		return nil, nil, err
> 	}
> 	b := bufpool.NewBytes()
> ```

What is the reason for using this function here? It means the reader needs to go look at another module, only to find out that this is effectively just calling `make(common.RawBytes, common.MaxMTU)`. I started reading the module docs for `bufpool` before i realised it was actually irrelevant.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1418)*

scrye:

---

*[go/lib/infra/transport/udp.go, line 35 at r1](https://reviewable.io:443/reviews/scionproto/scion/1418#-L6HH_PsXEeQSFb_TkFG:-L6HN2p6RJP292KrwAfK:b-c7gml3) ([raw file](https://github.com/scionproto/scion/blob/2cef448cff214835476283b791b6b508e5d283e2/go/lib/infra/transport/udp.go#L35)):*

Can you clarify this a bit? An example scenario would help. It looks like if a short deadline read comes in while a long deadline read is pending, the new read would just block unconditionally until after the existing read finishes, which seems a bit self-defeating.
</blockquote></details>

Damn, this means the same bug is present in `RUDP`. Hopefully fixed both now.

Also, neat discussion re: this issue of mixing old-school primitives and contexts: https://github.com/golang/go/issues/16620

---

*[go/lib/infra/transport/udp.go, line 67 at r1](https://reviewable.io:443/reviews/scionproto/scion/1418#-L6HI1UDwXZxzsRTOhr_:-L6HP_hnZbK-op_A7w9U:b6oor3e) ([raw file](https://github.com/scionproto/scion/blob/2cef448cff214835476283b791b6b508e5d283e2/go/lib/infra/transport/udp.go#L67)):*

What is the reason for using this function here? It means the reader needs to go look at another module, only to find out that this is effectively just calling `make(common.RawBytes, common.MaxMTU)`. I started reading the module docs for `bufpool` before i realised it was actually irrelevant.
</blockquote></details>

I'm thinking of having all packet buffer allocations going through the same library, to (1) avoid explicitly specifying the size everywhere, and (2) maybe amortize in the future (e.g., by actually grabbing the buffer from the bufpool itself).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1418)*

kormat:
Reviewed 2 of 2 files at r2, 1 of 1 files at r3.

---

*[go/lib/infra/transport/udp.go, line 67 at r1](https://reviewable.io:443/reviews/scionproto/scion/1418#-L6HI1UDwXZxzsRTOhr_:-L6HRtkYKp4E48yBsnOu:blb5872) ([raw file](https://github.com/scionproto/scion/blob/2cef448cff214835476283b791b6b508e5d283e2/go/lib/infra/transport/udp.go#L67)):*

I'm thinking of having all packet buffer allocations going through the same library, to (1) avoid explicitly specifying the size everywhere, and (2) maybe amortize in the future (e.g., by actually grabbing the buffer from the bufpool itself).
</blockquote></details>

The size _shouldn't_ be the same everywhere, depending on what you mean by "packet buffer". E.g. the BR uses 9k buffers, not 65k, and even that's only until we can support setting it based on the actual MTUs in use. It used to be 65k, but that wastes an obscene amount of memory. Using a buffer pool in the future is definitely something we can look at, but it would be cleaner to not add arbitrary references to that module until we're actually _using_ it for this. At that point we can do a sweep of the codebase and transition use-cases to it, which we'll have to do anyway.

---

*[go/lib/infra/transport/udp.go, line 58 at r2](https://reviewable.io:443/reviews/scionproto/scion/1418#-L6HTucugD3oDa36ry13:-L6HTucugD3oDa36ry14:bmo5sj) ([raw file](https://github.com/scionproto/scion/blob/7db182ff8b611c2a7d70fdbcf22a8cc47b06337d/go/lib/infra/transport/udp.go#L58)):*
> ```Go
> 		return err
> 	}
> 	_, err := u.conn.WriteTo(b, address)
> ```

This doesn't verify that the entire message was sent.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1418)*

scrye:
Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/infra/transport/udp.go, line 67 at r1](https://reviewable.io:443/reviews/scionproto/scion/1418#-L6HI1UDwXZxzsRTOhr_:-L6HYx_vg9kVLsY-eTk4:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/2cef448cff214835476283b791b6b508e5d283e2/go/lib/infra/transport/udp.go#L67)):*

The size _shouldn't_ be the same everywhere, depending on what you mean by "packet buffer". E.g. the BR uses 9k buffers, not 65k, and even that's only until we can support setting it based on the actual MTUs in use. It used to be 65k, but that wastes an obscene amount of memory. Using a buffer pool in the future is definitely something we can look at, but it would be cleaner to not add arbitrary references to that module until we're actually _using_ it for this. At that point we can do a sweep of the codebase and transition use-cases to it, which we'll have to do anyway.
</blockquote></details>

Done.

---

*[go/lib/infra/transport/udp.go, line 58 at r2](https://reviewable.io:443/reviews/scionproto/scion/1418#-L6HTucugD3oDa36ry13:-L6HYkRu_aKtcUUuvQlu:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/7db182ff8b611c2a7d70fdbcf22a8cc47b06337d/go/lib/infra/transport/udp.go#L58)):*

This doesn't verify that the entire message was sent.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1418)*

kormat:
:lgtm:

---

Reviewed 1 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1418#-:-L6Ha2AZLYHGFsgFUQfe:bnfp4nl)*

PR #1434, opened 2018-03-05
Export SIG config version as uint

scrye:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1434)

### DIFF
--- a/go/sig/config/config.go
+++ b/go/sig/config/config.go
@@ -29,7 +29,7 @@ import (
 // Cfg is a direct Go representation of the JSON file format.
 type Cfg struct {
 	ASes          map[addr.ISD_AS]*ASEntry
-	ConfigVersion float64
+	ConfigVersion uint64
 }
 
 // Load a JSON config file from path and parse it into a Cfg struct.
--- a/go/sig/metrics/metrics.go
+++ b/go/sig/metrics/metrics.go
@@ -18,6 +18,7 @@ package metrics
 
 import (
 	"flag"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
@@ -38,7 +39,7 @@ var promAddr = flag.String("prom", "127.0.0.1:1281", "Address to export promethe
 
 // Declare prometheus metrics to export.
 var (
-	ConfigVersion      prometheus.Gauge
+	ConfigVersion      uint64 // write this atomically
 	PktsRecv           *prometheus.CounterVec
 	PktsSent           *prometheus.CounterVec
 	PktBytesRecv       *prometheus.CounterVec
@@ -65,18 +66,12 @@ func Init(elem string) {
 		prometheus.MustRegister(v)
 		return v
 	}
-	newG := func(name, help string) prometheus.Gauge {
-		v := prom.NewGauge(namespace, "", name, help, constLabels)
-		prometheus.MustRegister(v)
-		return v
-	}
 	newCVec := func(name, help string, lNames []string) *prometheus.CounterVec {
 		v := prom.NewCounterVec(namespace, "", name, help, constLabels, lNames)
 		prometheus.MustRegister(v)
 		return v
 	}
 	// FIXME(kormat): these metrics should probably have more informative labels
-	ConfigVersion = newG("config_version", "Version number of the current config")
 	PktsRecv = newCVec("pkts_recv_total", "Number of packets received.", iaLabels)
 	PktsSent = newCVec("pkts_sent_total", "Number of packets sent.", iaLabels)
 	PktBytesRecv = newCVec("pkt_bytes_recv_total", "Number of packet bytes received.", iaLabels)
@@ -108,6 +103,9 @@ func Start() error {
 		return common.NewBasicError("Unable to bind prometheus metrics port", err)
 	}
 	log.Info("Exporting prometheus metrics", "addr", *promAddr)
+	http.HandleFunc("/version", func(w http.ResponseWriter, req *http.Request) {
+		fmt.Fprint(w, ConfigVersion)
+	})
 	go http.Serve(ln, nil)
 	return nil
 }
--- a/go/sig/metrics/metrics.go
+++ b/go/sig/metrics/metrics.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"sync/atomic"
 
 	log "github.com/inconshreveable/log15"
 
@@ -39,7 +40,7 @@ var promAddr = flag.String("prom", "127.0.0.1:1281", "Address to export promethe
 
 // Declare prometheus metrics to export.
 var (
-	ConfigVersion      uint64 // write this atomically
+	ConfigVersion      uint64 // atomic
 	PktsRecv           *prometheus.CounterVec
 	PktsSent           *prometheus.CounterVec
 	PktBytesRecv       *prometheus.CounterVec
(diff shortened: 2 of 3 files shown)


kormat:
Reviewed 3 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/sig/sig.go, line 151 at r1](https://reviewable.io:443/reviews/scionproto/scion/1434#-L6qHDyDs0Txq2YvcI5_:-L6qHDyDs0Txq2YvcI5a:bwyngvv) ([raw file](https://github.com/scionproto/scion/blob/cd01323f1d31fb1239c56e57095388e3612ae489/go/sig/sig.go#L151)):*
> ```Go
> 		return false
> 	}
> 	atomic.StoreUint64(&metrics.ConfigVersion, cfg.ConfigVersion)
> ```

There's a disconnect here between "last config version read from disk", and "config applied". The latter is more likely to be useful, which implies doing `ReloadConfig` and checking it succeeded before updating the exported version.

---

*[go/sig/metrics/metrics.go, line 42 at r1](https://reviewable.io:443/reviews/scionproto/scion/1434#-L6qHYmKEdESfCpQYmTP:-L6qHYmKEdESfCpQYmTQ:bjevrgu) ([raw file](https://github.com/scionproto/scion/blob/cd01323f1d31fb1239c56e57095388e3612ae489/go/sig/metrics/metrics.go#L42)):*
> ```Go
> // Declare prometheus metrics to export.
> var (
> 	ConfigVersion      uint64 // write this atomically
> ```

This is not a prometheus metric, so at a minimum it shouldn't be in this `var` block.

---

*[go/sig/metrics/metrics.go, line 106 at r1](https://reviewable.io:443/reviews/scionproto/scion/1434#-L6qHhktwXJ3GcyZuvuZ:-L6qHhktwXJ3GcyZuvu_:b9rgo77) ([raw file](https://github.com/scionproto/scion/blob/cd01323f1d31fb1239c56e57095388e3612ae489/go/sig/metrics/metrics.go#L106)):*
> ```Go
> 	}
> 	log.Info("Exporting prometheus metrics", "addr", *promAddr)
> 	http.HandleFunc("/version", func(w http.ResponseWriter, req *http.Request) {
> ```

This url is a bit generic, it would be easy to assume it refers to the _software_ version, rather than config version.

---

*[go/sig/metrics/metrics.go, line 107 at r1](https://reviewable.io:443/reviews/scionproto/scion/1434#-L6qIs4P9bpgc5zcldOh:-L6qIs4P9bpgc5zcldOi:b-u5nusu) ([raw file](https://github.com/scionproto/scion/blob/cd01323f1d31fb1239c56e57095388e3612ae489/go/sig/metrics/metrics.go#L107)):*
> ```Go
> 	log.Info("Exporting prometheus metrics", "addr", *promAddr)
> 	http.HandleFunc("/version", func(w http.ResponseWriter, req *http.Request) {
> 		fmt.Fprint(w, ConfigVersion)
> ```

Add a newline to the end, to make it a little easier to work with by hand.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1434)*

scrye:
Review status: 2 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[go/sig/sig.go, line 151 at r1](https://reviewable.io:443/reviews/scionproto/scion/1434#-L6qHDyDs0Txq2YvcI5_:-L6qJP1do7epfZIl080u:b-a17m43) ([raw file](https://github.com/scionproto/scion/blob/cd01323f1d31fb1239c56e57095388e3612ae489/go/sig/sig.go#L151)):*

There's a disconnect here between "last config version read from disk", and "config applied". The latter is more likely to be useful, which implies doing `ReloadConfig` and checking it succeeded before updating the exported version.
</blockquote></details>

There's no rollback though, so you end up with some changes from the new config, but the number is the old one.

---

*[go/sig/metrics/metrics.go, line 42 at r1](https://reviewable.io:443/reviews/scionproto/scion/1434#-L6qHYmKEdESfCpQYmTP:-L6qJVHMjtmmc3A9Njam:b-1e3w2l) ([raw file](https://github.com/scionproto/scion/blob/cd01323f1d31fb1239c56e57095388e3612ae489/go/sig/metrics/metrics.go#L42)):*

This is not a prometheus metric, so at a minimum it shouldn't be in this `var` block.
</blockquote></details>

`http` server runs from here, so I left it in the file.

---

*[go/sig/metrics/metrics.go, line 106 at r1](https://reviewable.io:443/reviews/scionproto/scion/1434#-L6qHhktwXJ3GcyZuvuZ:-L6qJjfUxzIUkQRykLYx:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cd01323f1d31fb1239c56e57095388e3612ae489/go/sig/metrics/metrics.go#L106)):*

This url is a bit generic, it would be easy to assume it refers to the _software_ version, rather than config version.
</blockquote></details>

Done.

---

*[go/sig/metrics/metrics.go, line 107 at r1](https://reviewable.io:443/reviews/scionproto/scion/1434#-L6qIs4P9bpgc5zcldOh:-L6qJjXhk3SfLkJscoUG:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cd01323f1d31fb1239c56e57095388e3612ae489/go/sig/metrics/metrics.go#L107)):*

Add a newline to the end, to make it a little easier to work with by hand.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1434)*

kormat:
Reviewed 1 of 1 files at r2.

---

*[go/sig/sig.go, line 151 at r1](https://reviewable.io:443/reviews/scionproto/scion/1434#-L6qHDyDs0Txq2YvcI5_:-L6qJu_srMwjHlFztdnw:b-xb7x2p) ([raw file](https://github.com/scionproto/scion/blob/cd01323f1d31fb1239c56e57095388e3612ae489/go/sig/sig.go#L151)):*

There's no rollback though, so you end up with some changes from the new config, but the number is the old one.
</blockquote></details>

Right. But this is being used to check if the new config version was deployed succesfully. In the case of a partial update, displaying the old number would at least signal there was a problem, and prompt an investigation. Updating the number regardless of whether the config loaded or not means it's not useful for that purpose. Alternatively, we export two pieces of data: the last config version loaded, and a bool indicating loading success. A json format would make this easy to work with.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1434)*

scrye:
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/sig/sig.go, line 151 at r1](https://reviewable.io:443/reviews/scionproto/scion/1434#-L6qHDyDs0Txq2YvcI5_:-L6qKhP6cYLAp-CLOdEG:blq4bpy) ([raw file](https://github.com/scionproto/scion/blob/cd01323f1d31fb1239c56e57095388e3612ae489/go/sig/sig.go#L151)):*

Right. But this is being used to check if the new config version was deployed succesfully. In the case of a partial update, displaying the old number would at least signal there was a problem, and prompt an investigation. Updating the number regardless of whether the config loaded or not means it's not useful for that purpose. Alternatively, we export two pieces of data: the last config version loaded, and a bool indicating loading success. A json format would make this easy to work with.
</blockquote></details>

Having a single number is simpler. It's now only updated on success.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1434)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1434#-:-L6qMC2NYl_WbfGS6qBu:bnfp4nl)*

PR #1436, opened 2018-03-05
Add support for Certificates to trustdb

Oncilla:
Add support for Certificates to `trustDB`:
 - Add `LeafCerts` table to database
 - Add `IssuerCerts` table to database

Modify `Chains` table:
 - Certificate chains are stored as links to leaf and issuer certificate to reduce space overhead

Fix Get*MaxVersion:
 - Before, Get*MaxVersion returned the overall max version, instead of max version for this IA or ISD.

Expose affected rows:
 - Ignore inserts of duplicate cert/chain/TRC
 - Instead, expose the number of affected rows. 

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1436)


shitz:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 53 at r1](https://reviewable.io:443/reviews/scionproto/scion/1436#-L81oCfWnlYhsxcHNuNW:-L81oCfXD5Vr1rQbg4nv:bawunu7) ([raw file](https://github.com/scionproto/scion/blob/022186144e58b76f135e12226ab56415389dbeaa/go/lib/infra/modules/trust/trustdb/db.go#L53)):*
> ```Go
> 	);
> 
> 	CREATE TABLE Certs (
> ```

This effectively leads to potentially a lot of duplicate information within the DB. My suggestion is to not store the chains explicitly. The schema would then look something like the following:
```
CREATE TABLE Certs (
    RowId INTEGER PRIMARY KEY AUTOINCREMENT,
    IsdID INTEGER NOT NULL,
    AsID INTEGER NOT NULL,
    Version INTEGER NOT NULL,
    Data TEXT NOT NULL,
    CONSTRAINT iav_unique UNIQUE (IsdID, AsID, Version)
);

CREATE TABLE Chains (
    IsdID INTEGER NOT NULL,
    AsID INTEGER NOT NULL,
    Version INTEGER NOT NULL,
    OrderKey INTEGER NOT NULL,
    CertsRowId INTEGER NOT NULL,
    PRIMARY KEY (IsdID, AsID, Version),
    FOREIGN KEY (CertsRowID) REFERENCES Certs(RowID),
);
```
`OrderKey` is used to encode the order of the cert in the chain. This will also allow for longer chains than 2, e.g., if we ever have non-core issuers.
Of course this will change the way you insert and retrieve chains. For an example how to use `RowID` as a foreign key you can look at `go/lib/pathdb/sqlite/sqlite.go`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

shitz:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 53 at r1](https://reviewable.io:443/reviews/scionproto/scion/1436#-L81oCfWnlYhsxcHNuNW:-L82D_DEkU8b94CFMPuO:b-rd92vq) ([raw file](https://github.com/scionproto/scion/blob/022186144e58b76f135e12226ab56415389dbeaa/go/lib/infra/modules/trust/trustdb/db.go#L53)):*

This effectively leads to potentially a lot of duplicate information within the DB. My suggestion is to not store the chains explicitly. The schema would then look something like the following:
```
CREATE TABLE Certs (
    RowId INTEGER PRIMARY KEY AUTOINCREMENT,
    IsdID INTEGER NOT NULL,
    AsID INTEGER NOT NULL,
    Version INTEGER NOT NULL,
    Data TEXT NOT NULL,
    CONSTRAINT iav_unique UNIQUE (IsdID, AsID, Version)
);

CREATE TABLE Chains (
    IsdID INTEGER NOT NULL,
    AsID INTEGER NOT NULL,
    Version INTEGER NOT NULL,
    OrderKey INTEGER NOT NULL,
    CertsRowId INTEGER NOT NULL,
    PRIMARY KEY (IsdID, AsID, Version),
    FOREIGN KEY (CertsRowID) REFERENCES Certs(RowID),
);
```
`OrderKey` is used to encode the order of the cert in the chain. This will also allow for longer chains than 2, e.g., if we ever have non-core issuers.
Of course this will change the way you insert and retrieve chains. For an example how to use `RowID` as a foreign key you can look at `go/lib/pathdb/sqlite/sqlite.go`
</blockquote></details>

Just to clarify: a chain is then effectively encoded as a number of rows in the `Chains` table and the corresponding certs are stored in the `Certs` table. Currently, the number of rows is always 2, with the row for the leaf cert having `OrderKey` 0 and the row for the issuer cert having `OrderKey` 1, but this could change in the future.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

Oncilla:
*[go/lib/infra/modules/trust/trustdb/db.go, line 53 at r1](https://reviewable.io:443/reviews/scionproto/scion/1436#-L81oCfWnlYhsxcHNuNW:-L86qogcziHKE_sZSEJY:b6ew797) ([raw file](https://github.com/scionproto/scion/blob/022186144e58b76f135e12226ab56415389dbeaa/go/lib/infra/modules/trust/trustdb/db.go#L53)):*

Just to clarify: a chain is then effectively encoded as a number of rows in the `Chains` table and the corresponding certs are stored in the `Certs` table. Currently, the number of rows is always 2, with the row for the leaf cert having `OrderKey` 0 and the row for the issuer cert having `OrderKey` 1, but this could change in the future.
</blockquote></details>

I do not really get how this is supposed to work.

Assuming the chain has length 3, how would you reference the intermediate certificate?

Also, I assume you use (IsdID,AsID,Ver) to implicitly fetch the leaf certificate. With this approach we cannot ensure at a sql instruction level, that the leaf certificate is actually in the db, using an non-null foreign key, this is possible.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

shitz:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 53 at r1](https://reviewable.io:443/reviews/scionproto/scion/1436#-L81oCfWnlYhsxcHNuNW:-L87SmfRCIYkI36JXfZp:b-k4d596) ([raw file](https://github.com/scionproto/scion/blob/022186144e58b76f135e12226ab56415389dbeaa/go/lib/infra/modules/trust/trustdb/db.go#L53)):*

I do not really get how this is supposed to work.

Assuming the chain has length 3, how would you reference the intermediate certificate?

Also, I assume you use (IsdID,AsID,Ver) to implicitly fetch the leaf certificate. With this approach we cannot ensure at a sql instruction level, that the leaf certificate is actually in the db, using an non-null foreign key, this is possible.



</blockquote></details>

The chain is uniquely identified by the (Isd, As, Version) triple of the leaf certificate. When you insert a new chain into the database, you'd first insert the individual certs into the `Certs` table. This will give you the `rowId` for each cert. Then you'd insert 3 rows into the `Chains` table, which all have the same (Isd, AS, Version) triple (this is the 'chain-key'). The `OrderKey` would be the relative order in the chain, i.e., the leaf is has `OrderKey` 0, the intermediate issuer `OrderKey` 1, and the root issuer `OrderKey` 2. The `CertsRowId` field would be set to the corresponding rowIds you got from inserting the certs into the `Certs` table.

So to answer your question you would refer to the intermediate certificate with the chain-key ((Isd, AS, Version) of leaf cert) and `OrderKey` 1. If you want to fetch the corresponding cert you can then use `CertsRowId` to fetch it from the `Certs` table.

Does this clarify it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

Oncilla:
*[go/lib/infra/modules/trust/trustdb/db.go, line 53 at r1](https://reviewable.io:443/reviews/scionproto/scion/1436#-L81oCfWnlYhsxcHNuNW:-L87ZW1-QYxF4O_d-kOo:bahpfgx) ([raw file](https://github.com/scionproto/scion/blob/022186144e58b76f135e12226ab56415389dbeaa/go/lib/infra/modules/trust/trustdb/db.go#L53)):*

The chain is uniquely identified by the (Isd, As, Version) triple of the leaf certificate. When you insert a new chain into the database, you'd first insert the individual certs into the `Certs` table. This will give you the `rowId` for each cert. Then you'd insert 3 rows into the `Chains` table, which all have the same (Isd, AS, Version) triple (this is the 'chain-key'). The `OrderKey` would be the relative order in the chain, i.e., the leaf is has `OrderKey` 0, the intermediate issuer `OrderKey` 1, and the root issuer `OrderKey` 2. The `CertsRowId` field would be set to the corresponding rowIds you got from inserting the certs into the `Certs` table.

So to answer your question you would refer to the intermediate certificate with the chain-key ((Isd, AS, Version) of leaf cert) and `OrderKey` 1. If you want to fetch the corresponding cert you can then use `CertsRowId` to fetch it from the `Certs` table.

Does this clarify it?
</blockquote></details>

The default way would be to have a third table which is used to keep the relation between `Chains` and `Certs` to reduce redundancy:
`
CREATE TABLE Chains (
		RowID INTEGER PRIMARY KEY AUTOINCREMENT,
		IsdID INTEGER NOT NULL,
		AsID INTEGER NOT NULL,
		Version INTEGER NOT NULL,
		CONSTRAINT iav_unique UNIQUE (IsdID, AsID, Version)
	);

	CREATE TABLE Certs (
		RowID INTEGER PRIMARY KEY AUTOINCREMENT,
		IsdID INTEGER NOT NULL,
		AsID INTEGER NOT NULL,
		Version INTEGER NOT NULL, 
		Type INTEGER NOT NULL,
		Data TEXT NOT NULL, 
		CONSTRAINT iavt_unique UNIQUE (IsdID, AsID, Version, Type)
	);

	CREATE TABLE ChainsCertsMap (
		ChainRowID INTEGER NOT NULL,
		CertRowID INTEGER NOT NULL,
		ChainIndex INTEGER NOT NULL,
		PRIMARY KEY (ChainRowID, CertRowID, ChainIndex)
		FOREIGN KEY (ChainRowID) REFERENCES Chains(RowID),
		FOREIGN KEY (CertRowID) REFERENCES Certs(RowID)
	);
`

What would be the advantage of your approach over this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

shitz:
Reviewed 2 of 7 files at r1, 3 of 4 files at r2.
Review status: 5 of 7 files reviewed at latest revision, 18 unresolved discussions.

---

*[go/cert_srv/conf/conf.go, line 31 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MTq2t8I_QgzuSGrvi:-L8MTq2t8I_QgzuSGrvj:b-p88g76) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/conf.go#L31)):*
> ```Go
> const (
> 	ErrorAddr      = "Unable to load addresses"
> 	ErrorCoreCert  = "Unable to load core certificate"
> ```

issuer

---

*[go/cert_srv/conf/conf.go, line 67 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MU0uoevhun83gYoxs:-L8MU0uoevhun83gYoxt:b-sc2u10) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/conf.go#L67)):*
> ```Go
> 	StateDir string
> 	// CoreCertStore holds issued core certificate.
> 	CoreCertStore *CoreCertStore
> ```

core cert should be renamed to issuer cert everywhere

---

*[go/cert_srv/conf/core_cert.go, line 15 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MUwATj0a5vNRA-m1e:-L8MUwATj0a5vNRA-m1f:b-s37m0q) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/core_cert.go#L15)):*
> ```Go
> 
> package conf
> ```

I don't think the `conf` package is a good fit for this store.

---

*[go/cert_srv/conf/core_cert.go, line 27 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MVM5vzwimKsNm03KK:-L8MVM5vzwimKsNm03KL:b-dmc0sz) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/core_cert.go#L27)):*
> ```Go
> 
> const (
> 	getCoreCertVersionStr = `
> ```

These statements shouldn't be needed anymore.

---

*[go/cert_srv/conf/core_cert.go, line 57 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MVgZeWYzt3g1qS3xa:-L8MVgZeWYzt3g1qS3xb:bsrhzpi) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/core_cert.go#L57)):*
> ```Go
> 		return nil, common.NewBasicError("Unable to prepare insertCoreCert", err)
> 	}
> 	var ver uint64 = 0
> ```

No need to initialize to 0

---

*[go/cert_srv/conf/core_cert.go, line 81 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MZueG4btINc6sf9FQ:-L8MZueG4btINc6sf9FR:b-glj932) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/core_cert.go#L81)):*
> ```Go
> // Set sets the core certificate. An error is returned, if there already exists a core cert with
> // higher version number.
> func (s *CoreCertStore) Set(crt *cert.Certificate) error {
> ```

Why do you do custom SQL here? Just get the issuer cert with the max version. If `crt.Version` is higher then you insert that one.

---

*[go/cert_srv/conf/testdata/ISD1-AS10-V1.crt, line 20 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MTZnFrK5zOUj6EobR:-L8MTZnGYtFlhEL9EMLc:bybznam) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/testdata/ISD1-AS10-V1.crt#L20)):*
> ```
>         "Version": 1,
>         "SubjectSignKey": "DDn+pZzqqaMtpg94vAXa1vkJubnyOVquMNQ2KeyYL7w=",
>         "Comment": "Core AS Certificate",
> ```

Comment outdated

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 60 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MPUWhpW8CjdpUeZOW:-L8MPUWhpW8CjdpUeZOX:b-e3fft1) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L60)):*
> ```Go
> 		AsID INTEGER NOT NULL,
> 		Version INTEGER NOT NULL,
> 		Data TEXT NOT NULL, 
> ```

trailing ws

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 70 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8ID7kz8TALcw5cc5SP:-L8ID7kz8TALcw5cc5SQ:b-hxngnh) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L70)):*
> ```Go
> 		Version INTEGER NOT NULL,
> 		Data TEXT NOT NULL, 
> 		CONSTRAINT iavt_unique UNIQUE (IsdID, AsID, Version)
> ```

`iav_unique`

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 105 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MP20f7RQVeVTVQvli:-L8MP20f7RQVeVTVQvlj:bswtufu) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L105)):*
> ```Go
> 			SELECT Data FROM LeafCerts WHERE IsdID=? AND AsID=? AND Version=?
> 			UNION 
> 			SELECT Data FROM IssuerCerts WHERE ROWID IN (
> ```

`RowId` here and in other places.

Also, this doesn't include any information about the ordering. My suggestion would be to sort the result according to the `OrderKey`. Then you already know everything is in the correct order when you iterate over the rows.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 114 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MRZZzKD-VMwS4mfvU:-L8MRZZzKD-VMwS4mfvV:b-1mw4g0) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L114)):*
> ```Go
> 			SELECT Data FROM IssuerCerts WHERE ROWID IN (
> 				SELECT IssCertsRowID FROM Chains WHERE IsdID=? AND AsID=? AND Version = (
> 					SELECT MAX(Version) FROM Chains WHERE IsdID=? AND AsID=?
> ```

You have to reuse the version you get from the first query, otherwise this might not be the same.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 160 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8IFOGZfoZCAe_3ghqQ:-L8IFOGZfoZCAe_3ghqR:b3bxcjk) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L160)):*
> ```Go
> 		return nil, err
> 	}
> 	if _, err = db.Exec(enableForeignKeysStr); err != nil {
> ```

This is already done in `sqlite.New`

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 238 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8IHF_BVpJJqYY_NLor:-L8IHF_BVpJJqYY_NLos:b-kigq1n) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L238)):*
> ```Go
> }
> 
> func parseIssCert(raw common.RawBytes, ia addr.IA, v uint64, err error) (*cert.Certificate, error) {
> ```

needed?

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 286 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8IGEgHNhjdLU8BFox4:-L8IGEgHNhjdLU8BFox5:b-58so6z) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L286)):*
> ```Go
> }
> 
> func parseLeafCert(raw common.RawBytes, ia addr.IA, v uint64, err error) (*cert.Certificate, error) {
> ```

needed?

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 367 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8ILPm6Q7mXzejFTfdw:-L8ILPm6Q7mXzejFTfdx:ba4jqsl) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L367)):*
> ```Go
> 		return nil, nil
> 	}
> 	if len(raw) != 2 {
> ```

Add a `FIXME` to remove this when chains longer than 2 are supported.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 378 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8ILnpvlSezJHmMYp6-:-L8ILnpvlSezJHmMYp60:b-sbwp2q) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L378)):*
> ```Go
> 		return nil, err
> 	}
> 	return &cert.Chain{Leaf: leaf, Core: iss}, nil
> ```

While we are at it, could you rename `Core` to `Issuer` in `cert.Chain`?

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 393 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8IS5BtaD8-ZYpG4dZz:-L8IS5BtaD8-ZYpG4d_-:b58hx2q) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L393)):*
> ```Go
> 		return common.NewBasicError("Unable to convert to JSON", err)
> 	}
> 	res, err := db.insertIssCertStmt.Exec(chain.Core.Subject.I, chain.Core.Subject.A,
> ```

Should use `ExecContext`. Same below.

Actually, now that inserting chains requires multiple statements and touches multiple tables, it should all be part of the a transaction that either gets committed completely in case nothing went wrong, or is rolled back. Check `pathdb/sqlite.go` for examples.

---

*[go/lib/infra/modules/trust/trustdb/db_test.go, line 70 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MS0dCIuIzv8_1pjQD:-L8MS0dCIuIzv8_1pjQE:b-sfwayz) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db_test.go#L70)):*
> ```Go
> 
> 		chain, err := cert.ChainFromFile("testdata/ISD1-AS10-V1.crt", false)
> 		SoMsg("err chain", err, ShouldBeNil)
> ```

You are not testing the functionality of `cert.ChainFromFile` here. You should `t.Fatal` for any error that is unexpected and has nothing to do with the thing that you actually want to test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

Oncilla:
Review status: 1 of 14 files reviewed at latest revision, 18 unresolved discussions.

---

*[go/cert_srv/conf/conf.go, line 31 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MTq2t8I_QgzuSGrvi:-L8WH6xX99yRX7DiUhba:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/conf.go#L31)):*

issuer
</blockquote></details>

Done.

---

*[go/cert_srv/conf/conf.go, line 67 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MU0uoevhun83gYoxs:-L8WPY7KAL0dMPnkS6dj:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/conf.go#L67)):*

core cert should be renamed to issuer cert everywhere
</blockquote></details>

Done.

---

*[go/cert_srv/conf/testdata/ISD1-AS10-V1.crt, line 20 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MTZnFrK5zOUj6EobR:-L8WVjFgCxZKMo-uT7u_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/testdata/ISD1-AS10-V1.crt#L20)):*

Comment outdated
</blockquote></details>

Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 60 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MPUWhpW8CjdpUeZOW:-L8WK3y13Qu0r1S3y8ic:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L60)):*

trailing ws
</blockquote></details>

Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 70 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8ID7kz8TALcw5cc5SP:-L8WK6f6DfVqLTH-lWSZ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L70)):*

`iav_unique`
</blockquote></details>

Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 105 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MP20f7RQVeVTVQvli:-L8WOLnV3o5Gj9yLy2hA:btaxp4n) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L105)):*

`RowId` here and in other places.

Also, this doesn't include any information about the ordering. My suggestion would be to sort the result according to the `OrderKey`. Then you already know everything is in the correct order when you iterate over the rows.
</blockquote></details>

For length of 2 it is sorted.
Sorting it would require a lot more complex sql statement. If this functionality is ever needed, we can implement it then.
Also, i think sorting this in code is far easier and faster, since in sql this will require a join.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 114 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MRZZzKD-VMwS4mfvU:-L8WLY181iAltpsWhX-Z:b-vd2xfm) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L114)):*

You have to reuse the version you get from the first query, otherwise this might not be the same.
</blockquote></details>

good point.
Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 160 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8IFOGZfoZCAe_3ghqQ:-L8WLi-w7WbYZKbOBBjN:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L160)):*

This is already done in `sqlite.New`
</blockquote></details>

Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 238 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8IHF_BVpJJqYY_NLor:-L8WLuE67nxvLRESNUqc:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L238)):*

needed?
</blockquote></details>

Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 286 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8IGEgHNhjdLU8BFox4:-L8WLuqT-g8dSOH_fg7q:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L286)):*

needed?
</blockquote></details>

Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 367 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8ILPm6Q7mXzejFTfdw:-L8WMGGPC8ILdg8baaI5:b-lbitrh) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L367)):*

Add a `FIXME` to remove this when chains longer than 2 are supported.
</blockquote></details>

I'm not sure if that should be a fixme. Currently I don't think chains longer than 2 should be supported.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 378 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8ILnpvlSezJHmMYp6-:-L8WNCjOC_XZwqOQNN8m:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L378)):*

While we are at it, could you rename `Core` to `Issuer` in `cert.Chain`?
</blockquote></details>

Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 393 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8IS5BtaD8-ZYpG4dZz:-L8WNXhm8yKwiSyfEJVi:blh8o0k) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L393)):*

Should use `ExecContext`. Same below.

Actually, now that inserting chains requires multiple statements and touches multiple tables, it should all be part of the a transaction that either gets committed completely in case nothing went wrong, or is rolled back. Check `pathdb/sqlite.go` for examples.
</blockquote></details>

I was thinking about that. But why should we rollback certificates inserts?
They are still legitimate, even though the chain was not added. 

---

*[go/lib/infra/modules/trust/trustdb/db_test.go, line 70 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MS0dCIuIzv8_1pjQD:-L8WOCq3DkaXv1gGqTZN:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db_test.go#L70)):*

You are not testing the functionality of `cert.ChainFromFile` here. You should `t.Fatal` for any error that is unexpected and has nothing to do with the thing that you actually want to test.
</blockquote></details>

Done.

---

*[go/cert_srv/conf/core_cert.go, line 15 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MUwATj0a5vNRA-m1e:-L8WHUlhDpGPoAQQFxOD:b-o880po) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/core_cert.go#L15)):*

I don't think the `conf` package is a good fit for this store.
</blockquote></details>

I'm open for suggestions. I did not find a more fitting place. 

---

*[go/cert_srv/conf/core_cert.go, line 27 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MVM5vzwimKsNm03KK:-L8WHftl-h4wQLpOEbf3:b-9iswmu) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/core_cert.go#L27)):*

These statements shouldn't be needed anymore.
</blockquote></details>

Why not?
first statement: 
Either we have this statement or we need to fetch the actual certificate data, parse it and then only access the version number.
I think it is more sane to just add another statement.

second statement: 
The update is done atomically, thus avoids possible race conditions.

---

*[go/cert_srv/conf/core_cert.go, line 57 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MVgZeWYzt3g1qS3xa:-L8WJ90l983NZpcRhFDo:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/core_cert.go#L57)):*

No need to initialize to 0
</blockquote></details>

Done.

---

*[go/cert_srv/conf/core_cert.go, line 81 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MZueG4btINc6sf9FQ:-L8WJBWp9dJGCA2OTypl:brde3d6) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/core_cert.go#L81)):*

Why do you do custom SQL here? Just get the issuer cert with the max version. If `crt.Version` is higher then you insert that one.
</blockquote></details>

The custom sql statement is atomic and avoids possible race condition. Alternatively I could add locks, but this seams a lot more sane.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1436)*

Oncilla:
*[go/cert_srv/conf/core_cert.go, line 81 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MZueG4btINc6sf9FQ:-L8W_MYpAyaSZnYT3FSq:b-i90j0i) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/cert_srv/conf/core_cert.go#L81)):*

The custom sql statement is atomic and avoids possible race condition. Alternatively I could add locks, but this seams a lot more sane.
</blockquote></details>

Or it could be done in a transaction, also requiring custom sql statement.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1436)*

shitz:
Reviewed 12 of 14 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[go/lib/crypto/cert/chain.go, line 64 at r3](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8XmKOLmjFVq6LOZXQv:-L8XmKOLmjFVq6LOZXQw:bxlymjm) ([raw file](https://github.com/scionproto/scion/blob/cdaf2335abefcafd9739ef5e0ed72ab2192d9757/go/lib/crypto/cert/chain.go#L64)):*
> fore
for

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 105 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MP20f7RQVeVTVQvli:-L8XnlHSytoDhCGooaqz:b-u8tvkm) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L105)):*

For length of 2 it is sorted.
Sorting it would require a lot more complex sql statement. If this functionality is ever needed, we can implement it then.
Also, i think sorting this in code is far easier and faster, since in sql this will require a join.
</blockquote></details>

Without the `OrderKey` you won't be able to sort anything in code, thus the join is necessary anyway. I quickly tested the following query and it seems to work:
```
SELECT Data, 0 FROM LeafCerts WHERE IsdID=? AND AsID=? AND Version=?
UNION
SELECT ic.Data, ch.OrderKey FROM IssuerCerts ic, Chains ch 
WHERE icRowID IN (
    SELECT IssCertsRowID FROM Chains WHERE IsdID=? AND AsID=? AND Version=?
)
ORDER BY ch.OrderKey
```

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 367 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8ILPm6Q7mXzejFTfdw:-L8XyaGAL6ldF6E90ZGB:b-j47kam) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L367)):*

I'm not sure if that should be a fixme. Currently I don't think chains longer than 2 should be supported.
</blockquote></details>

If that's the case we also don't need an order key. I think our decision was to support this in the future.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 393 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8IS5BtaD8-ZYpG4dZz:-L8XzfVYNUmFdZfuL6M2:b-frkt76) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L393)):*

I was thinking about that. But why should we rollback certificates inserts?
They are still legitimate, even though the chain was not added. 
</blockquote></details>

Hmm ok that is true I guess. As soon as you can have more than one issuer cert, those should be inserted as part of a transaction, otherwise you will have incomplete chains. I'm ok if you add a fixme.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 104 at r3](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8XnvHk1H9KF3o-dPF-:-L8XnvHk1H9KF3o-dPF0:b-e3fft1) ([raw file](https://github.com/scionproto/scion/blob/cdaf2335abefcafd9739ef5e0ed72ab2192d9757/go/lib/infra/modules/trust/trustdb/db.go#L104)):*
> ```Go
> 	getChainVersionStr = `
> 			SELECT Data FROM LeafCerts WHERE IsdID=? AND AsID=? AND Version=?
> 			UNION 
> ```

trailing ws

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 416 at r3](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8XzCbylLEHhcIDZsmR:-L8XzCbylLEHhcIDZsmS:b-e3fe1l) ([raw file](https://github.com/scionproto/scion/blob/cdaf2335abefcafd9739ef5e0ed72ab2192d9757/go/lib/infra/modules/trust/trustdb/db.go#L416)):*
> ```Go
> 		return rowsAffected, err
> 	}
> 	return rowsAffected + tmpRows, nil
> ```

`InsertChain` should probably only return as either 1 or 0 depending on the success of the operation to insert rows in the `Chains` table.

---

*[go/lib/trust/key_conf.go, line 40 at r3](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8Y08_8nzS49PB8z0Z-:-L8Y08_8nzS49PB8z0Z0:b-uml5p2) ([raw file](https://github.com/scionproto/scion/blob/cdaf2335abefcafd9739ef5e0ed72ab2192d9757/go/lib/trust/key_conf.go#L40)):*
> ```Go
> 
> const (
> 	IssSigKeyFile = "iss-sig.key"
> ```

Instead of actually changing the value, add a TODO to update it in a later PR, such that this PR does not become a breaking PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

Oncilla:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 105 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8MP20f7RQVeVTVQvli:-L8Y5YzN5BEBQX_a5ubY:b-4tofdn) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L105)):*

Without the `OrderKey` you won't be able to sort anything in code, thus the join is necessary anyway. I quickly tested the following query and it seems to work:
```
SELECT Data, 0 FROM LeafCerts WHERE IsdID=? AND AsID=? AND Version=?
UNION
SELECT ic.Data, ch.OrderKey FROM IssuerCerts ic, Chains ch 
WHERE icRowID IN (
    SELECT IssCertsRowID FROM Chains WHERE IsdID=? AND AsID=? AND Version=?
)
ORDER BY ch.OrderKey
```
</blockquote></details>

You can use the `Issuer` field in the certificate itself.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 367 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8ILPm6Q7mXzejFTfdw:-L8Y4Ga-8ZxqZ583KsNa:b-816tgq) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L367)):*

If that's the case we also don't need an order key. I think our decision was to support this in the future.
</blockquote></details>

If there ever will be the need for this, we the DB schema supports it already (modulo a sql statement change) .
I think we should leave it as is in that regard.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 393 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8IS5BtaD8-ZYpG4dZz:-L8Y4gkq6KY2kYFJfryz:b-978jqv) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L393)):*

Hmm ok that is true I guess. As soon as you can have more than one issuer cert, those should be inserted as part of a transaction, otherwise you will have incomplete chains. I'm ok if you add a fixme.
</blockquote></details>

This is not true, if the issuing certs are inserted in order. Or am I missing something?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1436)*

shitz:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 105 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MP20f7RQVeVTVQvli:-L8aLxtZCcn1IGmWcOAx:b7zd690) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L105)):*

You can use the `Issuer` field in the certificate itself.
</blockquote></details>

Then the `OrderKey` is not needed at all and should be removed. DBs are very smart about inner joins and you have to do a lookup in `Chains` anyway. If you feel strongly about using the `issuer` and do the reassembly purely in code, then please remove `OrderKey` completely.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 367 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8ILPm6Q7mXzejFTfdw:-L8aNm4P66IlvRLKR-0l:b-6grlbc) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L367)):*

If there ever will be the need for this, we the DB schema supports it already (modulo a sql statement change) .
I think we should leave it as is in that regard.
</blockquote></details>

If you want to use the issuer field to reassemble the chain, then `OrderKey` is not needed anymore and nothing changes in terms of DB schema or SQL statement. I would still like to add a `TODO: Add support for chains longer than 2`.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 393 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8IS5BtaD8-ZYpG4dZz:-L8aMwfdzkY5VBBrPuN9:bxin18w) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L393)):*

This is not true, if the issuing certs are inserted in order. Or am I missing something?
</blockquote></details>

I'm not talking about the certs. You could still do this outside of a transaction, one-by-one. The entries to the `Chains` table, however, need to either be all there or none of them, i.e., those should be done using a transaction.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

Oncilla:
Review status: 6 of 9 files reviewed at latest revision, 7 unresolved discussions.

---

*[go/lib/crypto/cert/chain.go, line 64 at r3](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8XmKOLmjFVq6LOZXQv:-L8aNaRLEuIwYCMRqUbz:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cdaf2335abefcafd9739ef5e0ed72ab2192d9757/go/lib/crypto/cert/chain.go#L64)):*

> fore
for
</blockquote></details>

Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 105 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MP20f7RQVeVTVQvli:-L8a_CulgEdxMlyfObqg:b-ljgugf) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L105)):*

Then the `OrderKey` is not needed at all and should be removed. DBs are very smart about inner joins and you have to do a lookup in `Chains` anyway. If you feel strongly about using the `issuer` and do the reassembly purely in code, then please remove `OrderKey` completely.
</blockquote></details>

It is still needed since `PRIMARY KEY (IsdID, AsID, Version)` does not work for multiple entries of the same chain.
But this might be a premature optimization on my part.
I went with you're solution. If it turns out that the query is to slow, we can always come back to this and avoid joins.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 367 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8ILPm6Q7mXzejFTfdw:-L8a_xObC0xuqJZi6nkX:b-pxdwjb) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L367)):*

If you want to use the issuer field to reassemble the chain, then `OrderKey` is not needed anymore and nothing changes in terms of DB schema or SQL statement. I would still like to add a `TODO: Add support for chains longer than 2`.
</blockquote></details>

Offloaded `len(chain) != 2` to `lib/crypto/cert.Chain` such that it is in one place.
I don't think we need a `FIXME/TODO` anymore, since supporting chains longer 2 will require a structural change in cert.Chain. 

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 393 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8IS5BtaD8-ZYpG4dZz:-L8aYi-UqPDiDwBEgrks:b-9bj8nm) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L393)):*

I'm not talking about the certs. You could still do this outside of a transaction, one-by-one. The entries to the `Chains` table, however, need to either be all there or none of them, i.e., those should be done using a transaction.
</blockquote></details>

I see. Doing this as a transaction should be implemented in case we move to longer chains.
At this point I think it is fine as is. 

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 104 at r3](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8XnvHk1H9KF3o-dPF-:-L8aOAlo6S3osrSZ5BsG:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cdaf2335abefcafd9739ef5e0ed72ab2192d9757/go/lib/infra/modules/trust/trustdb/db.go#L104)):*

trailing ws
</blockquote></details>

Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 416 at r3](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8XzCbylLEHhcIDZsmR:-L8aZaCbfiNahrVbTbeY:b-lkcnoj) ([raw file](https://github.com/scionproto/scion/blob/cdaf2335abefcafd9739ef5e0ed72ab2192d9757/go/lib/infra/modules/trust/trustdb/db.go#L416)):*

`InsertChain` should probably only return as either 1 or 0 depending on the success of the operation to insert rows in the `Chains` table.
</blockquote></details>

True.
Done.

---

*[go/lib/trust/key_conf.go, line 40 at r3](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8Y08_8nzS49PB8z0Z-:-L8a_AxrGiw5IcnvkynI:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cdaf2335abefcafd9739ef5e0ed72ab2192d9757/go/lib/trust/key_conf.go#L40)):*

Instead of actually changing the value, add a TODO to update it in a later PR, such that this PR does not become a breaking PR.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

shitz:
Reviewed 3 of 4 files at r4.

---

*[go/lib/crypto/cert/chain.go, line 106 at r4](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8b9lFM-jDWG9qG620b:-L8b9lFM-jDWG9qG620c:bmlg2tq) ([raw file](https://github.com/scionproto/scion/blob/063db999152e79a7ce55fc7d380d164f5b09ad96/go/lib/crypto/cert/chain.go#L106)):*
> ```Go
> }
> 
> func ChainFromSlice(certs []*Certificate) (*Chain, error) {
> ```

Add a docstring saying that `certs[0]` needs to be the leaf cert and `certs[1]` the issuer cert.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 105 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8MP20f7RQVeVTVQvli:-L8b4AYEinYF0idEvayj:b-bahg1p) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L105)):*

It is still needed since `PRIMARY KEY (IsdID, AsID, Version)` does not work for multiple entries of the same chain.
But this might be a premature optimization on my part.
I went with you're solution. If it turns out that the query is to slow, we can always come back to this and avoid joins.
</blockquote></details>

Actually, you don't need to provide a primary key. Sqlite will automatically use a `rowId` autoincrement behind the scenes as a primary key, so if you want to use the issuer-based approach, feel free.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 393 at r2](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8IS5BtaD8-ZYpG4dZz:-L8b97KsGpqncF6tHvkA:beqc5x) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L393)):*

I see. Doing this as a transaction should be implemented in case we move to longer chains.
At this point I think it is fine as is. 
</blockquote></details>

Yes, I agree it's fine as it is, but it should have some note that we don't forget about it when the time comes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

Oncilla:

---

*[go/lib/crypto/cert/chain.go, line 106 at r4](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8b9lFM-jDWG9qG620b:-L8bM3q53c_FTMVQw5w8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/063db999152e79a7ce55fc7d380d164f5b09ad96/go/lib/crypto/cert/chain.go#L106)):*

Add a docstring saying that `certs[0]` needs to be the leaf cert and `certs[1]` the issuer cert.
</blockquote></details>

Done.

---

*[go/lib/infra/modules/trust/trustdb/db.go, line 393 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1436#-L8IS5BtaD8-ZYpG4dZz:-L8bO7kI386nfb2NJBIF:b-eam1tf) ([raw file](https://github.com/scionproto/scion/blob/728911f1c5b089dc02377422037921d9074763a7/go/lib/infra/modules/trust/trustdb/db.go#L393)):*

Yes, I agree it's fine as it is, but it should have some note that we don't forget about it when the time comes.
</blockquote></details>

`if` the time comes.
Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1436)*

shitz:
Reviewed 2 of 2 files at r5.

---

*[go/lib/crypto/cert/chain.go, line 113 at r5](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8bhK1UVqRwE2W4VHQu:-L8bhK1UVqRwE2W4VHQv:b-ra4wjs) ([raw file](https://github.com/scionproto/scion/blob/da28b0450cacc9d74223640e45f60d423f96f4ce/go/lib/crypto/cert/chain.go#L113)):*
> ```Go
> 		return nil, common.NewBasicError("Unsupported chain length", nil, "len", len(certs))
> 	}
> 	if certs[0] == nil && certs[1] == nil {
> ```

This should be `||` instead of `&&`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

Oncilla:
Review status: 8 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/lib/crypto/cert/chain.go, line 113 at r5](https://reviewable.io:443/reviews/scionproto/scion/1436#-L8bhK1UVqRwE2W4VHQu:-L8bkI3MULTrpLprzEKZ:be421sj) ([raw file](https://github.com/scionproto/scion/blob/da28b0450cacc9d74223640e45f60d423f96f4ce/go/lib/crypto/cert/chain.go#L113)):*

This should be `||` instead of `&&`
</blockquote></details>

Good catch.
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436)*

shitz:
:lgtm:

---

Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1436#-:-L8bkz_W6JI0CLkJuIdX:bnfp4nl)*

PR #1437, opened 2018-03-05
Fix topology generator

Oncilla:
Under some specific circumstances the validity period of some
certificates is outside of the signer's certificate. This renders them
non-verifiable.

This fix makes sure that signed certificates are inside of the signer's
certificates validity period.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1437)


Oncilla:
Closed, since scion-pki will replace generator soon™.

PR #1019, opened 2017-03-17
Reorder the fields in the common header.

kormat:
For:
https://github.com/netsec-ethz/scion/issues/980:
- "Move HdrLen just after TotalLen"
https://github.com/netsec-ethz/scion/issues/982:
- "Change address header order and padding"

N.B.: this change only affects the common header. The address header
will be changed in an upcoming PR instead.

Also:
- Add dedicated Go HostAddrType type.
- Unit tests for go/lib/spkt/cmnhdr.go
- Increase max python line length to 100 to match Go.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1019)

PR #1443, opened 2018-03-08
Implement mock sciond for testing

scrye:
Also includes miscellaneous unit test cleanup/improvements.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1443)

### DIFF
--- a/go/lib/common/defs.go
+++ b/go/lib/common/defs.go
@@ -16,6 +16,7 @@ package common
 
 import (
 	"reflect"
+	"strconv"
 )
 
 const (
@@ -38,6 +39,10 @@ const (
 
 // Interface ID
 type IFIDType uint64
+
+func (ifid IFIDType) String() string {
+	return strconv.FormatUint(uint64(ifid), 10)
+}
 
 const IFIDBytes = 8
 
--- a/go/lib/pathmgr/main_test.go
+++ b/go/lib/pathmgr/main_test.go
@@ -1,0 +1,28 @@
+// Copyright 2017 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pathmgr
+
+import (
+	"os"
+	"testing"
+
+	log "github.com/inconshreveable/log15"
+)
+
+// Disable logging in all tests
+func TestMain(m *testing.M) {
+	log.Root().SetHandler(log.DiscardHandler())
+	os.Exit(m.Run())
+}
--- a/go/lib/pathmgr/main_test.go
+++ b/go/lib/pathmgr/main_test.go
@@ -1,4 +1,4 @@
-// Copyright 2017 ETH Zurich
+// Copyright 2018 ETH Zurich
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/go/lib/pathmgr/pathmgr_test.go
+++ b/go/lib/pathmgr/pathmgr_test.go
@@ -149,6 +149,7 @@ func TestRevoke(t *testing.T) {
 			"[1-18#1815 1-15#1518 1-15#1512 1-12#1215 1-12#1222 2-22#2212]")
 
 		Convey("Revoke a path that's not part of any path set", func() {
+			g.RemoveLink(1311)
 			pm.cache.revoke(uifidFromValues(MustParseIA("1-13"), 1311))
 			aps := pm.Query(querySrc, queryDst)
 			apsCheckPaths("path", aps,
@@ -164,10 +165,10 @@ func TestRevoke(t *testing.T) {
 			// does not inform sciond. This means that the path manager
 			// reaches 0 paths after the revocation, thus forcing a requery
 			// to sciond behind the scenes, which gets back the same path.
+			g.RemoveLink(1019)
 			pm.cache.revoke(uifidFromValues(MustParseIA("1-10"), 1019))
 			aps := pm.Query(querySrc, queryDst)
-			apsCheckPaths("path", aps,
-				"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
+			apsCheckPaths("path", aps)
 			apsCheckPaths("watch", sp.Load().APS,
 				"[1-18#1815 1-15#1518 1-15#1512 1-12#1215 1-12#1222 2-22#2212]")
 			apsCheckPaths("watch filter", spf.Load().APS,
@@ -178,6 +179,7 @@ func TestRevoke(t *testing.T) {
 			// The revoke below only invalidates the cache. Because watches
 			// do not requery sciond automatically (like the previous
 			// test), they will be left with 0 paths.
+			g.RemoveLink(1815)
 			pm.cache.revoke(uifidFromValues(watchSrc, 1815))
 			apsCheckPaths("path", aps,
 				"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
--- a/go/lib/pathmgr/pathmgr_test.go
+++ b/go/lib/pathmgr/pathmgr_test.go
@@ -15,6 +15,7 @@
 package pathmgr
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
@@ -23,197 +24,187 @@ import (
 	. "github.com/smartystreets/goconvey/convey"
 
 	"github.com/scionproto/scion/go/lib/sciond"
-	"github.com/scionproto/scion/go/lib/sciond/graph"
+	"github.com/scionproto/scion/go/lib/xtest/graph"
 )
 
 func TestQuery(t *testing.T) {
-	Convey("Create path manager (path set max age = 1 second)", t, func() {
-		g := graph.NewFromDescription(graph.DefaultGraphDescription)
-		pm, err := New(
-			sciond.NewMockService(g),
-			&Timers{
-				NormalRefire: 5 * time.Second,
-				ErrorRefire:  5 * time.Second,
-				MaxAge:       time.Second,
-			},
-			log.Root(),
-		)
-		SoMsg("pm", pm, ShouldNotBeNil)
-		SoMsg("err", err, ShouldBeNil)
+	Convey("Query, we have 0 paths and SCIOND is asked again, receive 1 path", t, func() {
+		g := graph.NewDefaultGraph()
+		pm := NewPR(t, g, 5, 5, 1)
+		srcIA := MustParseIA("1-10")
+		dstIA := MustParseIA("1-16")
 
-		Convey("Query, we have 0 paths and SCIOND is asked again, receive 1 path", func() {
-			srcIA := MustParseIA("1-10")
-			dstIA := MustParseIA("1-16")
+		aps := pm.Query(srcIA, dstIA)
+		SoMsg("aps len", len(aps), ShouldEqual, 1)
+		Convey("Query immediately, same path is read from cache", func() {
 			aps := pm.Query(srcIA, dstIA)
 			SoMsg("aps len", len(aps), ShouldEqual, 1)
-			Convey("Query immediately, same path is read from cache", func() {
-				aps := pm.Query(srcIA, dstIA)
-				SoMsg("aps len", len(aps), ShouldEqual, 1)
-			})
-			Convey("Wait 2 seconds for paths to expire, then query and get new paths", func() {
-				// Add new path between 1-10 and 1-16
-				g.Connect("1-10", 101902, "1-19", 191002)
-				// Wait for two seconds to guarantee that the pathmgr refreshes the paths
-				<-time.After(2 * time.Second)
-				aps := pm.Query(srcIA, dstIA)
-				SoMsg("aps len", len(aps), ShouldEqual, 2)
-			})
+			SoMsg("path", getPathStrings(aps), ShouldContain,
+				"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
+		})
+		Convey("Wait 2 seconds for paths to expire, then query and get new paths", func() {
+			// Add new path between 1-10 and 1-16
+			g.AddLink("1-10", 101902, "1-19", 191002)
+			// Wait for two seconds to guarantee that the pathmgr refreshes the paths
+			<-time.After(2 * time.Second)
+			aps := pm.Query(srcIA, dstIA)
+			SoMsg("aps len", len(aps), ShouldEqual, 2)
+			SoMsg("path #1", getPathStrings(aps), ShouldContain,
+				"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
+			SoMsg("path #2", getPathStrings(aps), ShouldContain,
+				"[1-10#101902 1-19#191002 1-19#1916 1-16#1619]")
 		})
 	})
 }
 
 func TestQueryFilter(t *testing.T) {
-	Convey("Create path manager", t, func() {
-		g := graph.NewFromDescription(graph.DefaultGraphDescription)
-		pm, err := New(sciond.NewMockService(g), &Timers{}, log.Root())
-		SoMsg("pm", pm, ShouldNotBeNil)
+	Convey("Query with filter, only one path should remain", t, func() {
+		g := graph.NewDefaultGraph()
+		pm := NewPR(t, g, 0, 0, 0)
+		srcIA := MustParseIA("1-10")
+		dstIA := MustParseIA("1-16")
+
+		pp, err := NewPathPredicate("1-19#0")
 		SoMsg("err", err, ShouldBeNil)
+		SoMsg("pp", pp, ShouldNotBeNil)
 
-		Convey("Query with filter, only one path should remain", func() {
-			srcIA := MustParseIA("1-10")
-			dstIA := MustParseIA("1-16")
-
-			pp, err := NewPathPredicate("1-19#0")
-			SoMsg("err", err, ShouldBeNil)
-			SoMsg("pp", pp, ShouldNotBeNil)
-
-			aps := pm.QueryFilter(srcIA, dstIA, pp)
-			SoMsg("aps len", len(aps), ShouldEqual, 1)
-		})
+		aps := pm.QueryFilter(srcIA, dstIA, pp)
+		SoMsg("aps len", len(aps), ShouldEqual, 1)
+		SoMsg("path", getPathStrings(aps), ShouldContain,
+			"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
 	})
 }
 
 func TestRegister(t *testing.T) {
-	Convey("Create path manager", t, func() {
-		g := graph.NewFromDescription(graph.DefaultGraphDescription)
+
+	Convey("Register for path, receive 0 responses", t, func() {
+		g := graph.NewDefaultGraph()
 		// Remove link between 1-19 and 1-16 so that the initial path set is
 		// nil
-		g.Disconnect(1019)
+		g.RemoveLink(1019)
+		pm := NewPR(t, g, 1, 1, 1)
+		srcIA := MustParseIA("1-10")
+		dstIA := MustParseIA("1-16")
 
-		pm, err := New(
-			sciond.NewMockService(g),
-			&Timers{
-				NormalRefire: time.Second,
-				ErrorRefire:  time.Second,
-				MaxAge:       time.Second,
-			},
-			log.Root(),
-		)
-		SoMsg("pm", pm, ShouldNotBeNil)
+		sp, err := pm.Watch(srcIA, dstIA)
 		SoMsg("err", err, ShouldBeNil)
+		SoMsg("aps", len(sp.Load().APS), ShouldEqual, 0)
 
-		Convey("Register for path, receive 0 responses", func() {
-			srcIA := MustParseIA("1-10")
-			dstIA := MustParseIA("1-16")
-
-			sp, err := pm.Watch(srcIA, dstIA)
-			SoMsg("err", err, ShouldBeNil)
-			SoMsg("aps", len(sp.Load().APS), ShouldEqual, 0)
-
-			Convey("Wait 5 seconds, the APS should contain fresh paths", func() {
-				// Readd the link between 1-19 and 1-16; the path manager will
-				// update APS behind the scenes (after a normal refire of one
-				// second), so it should contain the path after 4 seconds.
-				g.Connect("1-10", 1019, "1-19", 1910)
-				<-time.After(4 * time.Second)
-				SoMsg("aps", len(sp.Load().APS), ShouldEqual, 1)
-			})
+		Convey("Wait 5 seconds, the APS should contain fresh paths", func() {
+			// Readd the link between 1-19 and 1-16; the path manager will
+			// update APS behind the scenes (after a normal refire of one
+			// second), so it should contain the path after 4 seconds.
+			g.AddLink("1-10", 1019, "1-19", 1910)
+			<-time.After(4 * time.Second)
+			SoMsg("aps", len(sp.Load().APS), ShouldEqual, 1)
+			SoMsg("path", getPathStrings(sp.Load().APS), ShouldContain,
+				"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
 		})
 	})
 }
 
 func TestRegisterFilter(t *testing.T) {
-	Convey("Create path manager", t, func() {
-		g := graph.NewFromDescription(graph.DefaultGraphDescription)
+	Convey("Register filter 1-19#1910", t, func() {
+		g := graph.NewDefaultGraph()
+		pm := NewPR(t, g, 1, 1, 1)
+		srcIA := MustParseIA("1-10")
+		dstIA := MustParseIA("1-16")
 
-		pm, err := New(
-			sciond.NewMockService(g),
-			&Timers{
-				NormalRefire: time.Second,
-				ErrorRefire:  time.Second,
-				MaxAge:       time.Second,
-			},
-			log.Root(),
-		)
-		SoMsg("pm", pm, ShouldNotBeNil)
+		pp, err := NewPathPredicate("1-19#1910")
+		SoMsg("pp", pp, ShouldNotBeNil)
 		SoMsg("err", err, ShouldBeNil)
 
-		Convey("Register filter 1-19#1910", func() {
-			srcIA := MustParseIA("1-10")
-			dstIA := MustParseIA("1-16")
-
-			pp, err := NewPathPredicate("1-19#1910")
-			SoMsg("pp", pp, ShouldNotBeNil)
-			SoMsg("err", err, ShouldBeNil)
-
-			sp, err := pm.WatchFilter(srcIA, dstIA, pp)
-			SoMsg("err", err, ShouldBeNil)
-			SoMsg("len aps", len(sp.Load().APS), ShouldEqual, 1)
-		})
+		sp, err := pm.WatchFilter(srcIA, dstIA, pp)
+		SoMsg("err", err, ShouldBeNil)
+		SoMsg("len aps", len(sp.Load().APS), ShouldEqual, 1)
+		SoMsg("path", getPathStrings(sp.Load().APS), ShouldContain,
+			"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
 	})
 }
 
 func TestRevoke(t *testing.T) {
-	Convey("Create path manager", t, func() {
-		g := graph.NewFromDescription(graph.DefaultGraphDescription)
+	Convey("Populate cache via Query, Watch, WatchFilter for different destinations", t, func() {
+		g := graph.NewDefaultGraph()
+		pm := NewPR(t, g, 60, 60, 60)
+		// Query: 1-10 -> 1-16
+		querySrc := MustParseIA("1-10")
+		queryDst := MustParseIA("1-16")
+		// Watch/WatchFilter: 1-18 -> 2-22
+		watchSrc := MustParseIA("1-18")
+		watchDst := MustParseIA("2-22")
 
-		pm, err := New(
-			sciond.NewMockService(g),
-			&Timers{
-				NormalRefire: time.Minute,
-				ErrorRefire:  time.Minute,
-				MaxAge:       time.Minute,
-			},
-			log.Root(),
-		)
-		SoMsg("pm", pm, ShouldNotBeNil)
-		SoMsg("err", err, ShouldBeNil)
+		aps := pm.Query(querySrc, queryDst)
+		SoMsg("len(aps)", len(aps), ShouldEqual, 1)
+		SoMsg("path", getPathStrings(aps), ShouldContain,
+			"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
 
-		Convey("Populate cache via Query, Watch, WatchFilter for different destinations", func() {
-			// Query: 1-10 -> 1-16
-			querySrc := MustParseIA("1-10")
-			queryDst := MustParseIA("1-16")
-			// Watch/WatchFilter: 1-18 -> 2-22
-			watchSrc := MustParseIA("1-18")
-			watchDst := MustParseIA("2-22")
+		sp, err := pm.Watch(watchSrc, watchDst)
+		SoMsg("err watch", err, ShouldBeNil)
+		SoMsg("len(aps) watch", len(sp.Load().APS), ShouldEqual, 1)
+		SoMsg("path watch", getPathStrings(sp.Load().APS), ShouldContain,
+			"[1-18#1815 1-15#1518 1-15#1512 1-12#1215 1-12#1222 2-22#2212]")
 
+		pp, err := NewPathPredicate("1-15#1518")
+		SoMsg("err predicate", err, ShouldBeNil)
+		spf, err := pm.WatchFilter(watchSrc, watchDst, pp)
+		SoMsg("err watch filter", err, ShouldBeNil)
+		SoMsg("len(aps) watch filter", len(spf.Load().APS), ShouldEqual, 1)
+		SoMsg("path watch filter", getPathStrings(spf.Load().APS), ShouldContain,
+			"[1-18#1815 1-15#1518 1-15#1512 1-12#1215 1-12#1222 2-22#2212]")
+
+		Convey("Revoke a path that's not part of any path set", func() {
+			g.RemoveLink(1311)
 			aps := pm.Query(querySrc, queryDst)
 			SoMsg("len(aps)", len(aps), ShouldEqual, 1)
-
-			sp, err := pm.Watch(watchSrc, watchDst)
-			SoMsg("err watch", err, ShouldBeNil)
-			SoMsg("len(aps) watch", len(sp.Load().APS), ShouldEqual, 1)
-
-			pp, err := NewPathPredicate("1-15#1518")
-			SoMsg("err predicate", err, ShouldBeNil)
-			spf, err := pm.WatchFilter(watchSrc, watchDst, pp)
-			SoMsg("err watch filter", err, ShouldBeNil)
-			SoMsg("len(aps) watch filter", len(spf.Load().APS), ShouldEqual, 1)
-
-			Convey("Revoke a path that's not part of any path set", func() {
-				g.Disconnect(1311)
-				aps := pm.Query(querySrc, queryDst)
-				SoMsg("len(aps)", len(aps), ShouldEqual, 1)
-			})
-			Convey("Revoke a path that's in Query, but not in Watch/WatchFilter path sets", func() {
-				// Disconnect #1619
-				// Note that the revoke below only invalidates the cache and
-				// does not inform sciond. This means that the path manager
-				// reaches 0 paths after the revocation, thus forcing a requery
-				// to sciond behind the scenes, which gets back the same path.
-				pm.cache.revoke(uifidFromValues(MustParseIA("1-10"), 1019))
-				aps := pm.Query(querySrc, queryDst)
-				SoMsg("len(aps)", len(aps), ShouldEqual, 1)
-			})
-			Convey("Revoke a path that's in Watch and WatchFilter, but not in Query", func() {
-				// Disconnect #1815
-				// The revoke below only invalidates the cache. Because watches
-				// do not requery sciond automatically (like the previous
-				// test), they will be left with 0 paths.
-				pm.cache.revoke(uifidFromValues(watchSrc, 1815))
-				SoMsg("len(aps) watch", len(sp.Load().APS), ShouldEqual, 0)
-				SoMsg("len(aps) watch filter", len(spf.Load().APS), ShouldEqual, 0)
-			})
+			SoMsg("path", getPathStrings(aps), ShouldContain,
+				"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
+		})
+		Convey("Revoke a path that's in Query, but not in Watch/WatchFilter path sets", func() {
+			// Disconnect #1619
+			// Note that the revoke below only invalidates the cache and
+			// does not inform sciond. This means that the path manager
+			// reaches 0 paths after the revocation, thus forcing a requery
+			// to sciond behind the scenes, which gets back the same path.
+			pm.cache.revoke(uifidFromValues(MustParseIA("1-10"), 1019))
+			aps := pm.Query(querySrc, queryDst)
+			SoMsg("len(aps)", len(aps), ShouldEqual, 1)
+			SoMsg("path watch", getPathStrings(aps), ShouldContain,
+				"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
+		})
+		Convey("Revoke a path that's in Watch and WatchFilter, but not in Query", func() {
+			// Disconnect #1815
+			// The revoke below only invalidates the cache. Because watches
+			// do not requery sciond automatically (like the previous
+			// test), they will be left with 0 paths.
+			pm.cache.revoke(uifidFromValues(watchSrc, 1815))
+			SoMsg("len(aps) watch", len(sp.Load().APS), ShouldEqual, 0)
+			SoMsg("len(aps) watch filter", len(sp.Load().APS), ShouldEqual, 0)
 		})
 	})
+}
+
+func NewPR(t *testing.T, g *graph.Graph, normalRefire, errorRefire, maxAge time.Duration) *PR {
+	t.Helper()
+
+	pm, err := New(
+		sciond.NewMockService(g),
+		&Timers{
+			NormalRefire: normalRefire * time.Second,
+			ErrorRefire:  errorRefire * time.Second,
+			MaxAge:       maxAge * time.Second,
+		},
+		log.Root(),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return pm
+}
+
+func getPathStrings(aps AppPathSet) []string {
+	var ss []string
+	for _, v := range aps {
+		ss = append(ss, fmt.Sprintf("%v", v.Entry.Path.Interfaces))
+	}
+	return ss
 }
--- a/go/lib/pathmgr/pathmgr_test.go
+++ b/go/lib/pathmgr/pathmgr_test.go
@@ -30,7 +30,7 @@ import (
 func TestQuery(t *testing.T) {
 	Convey("Query, we have 0 paths and SCIOND is asked again, receive 1 path", t, func() {
 		g := graph.NewDefaultGraph()
-		pm := NewPR(t, g, 5, 5, 1)
+		pm := NewPR(t, g, 250, 250, 100)
 		srcIA := MustParseIA("1-10")
 		dstIA := MustParseIA("1-16")
 
@@ -42,11 +42,11 @@ func TestQuery(t *testing.T) {
 			SoMsg("path", getPathStrings(aps), ShouldContain,
 				"[1-10#1019 1-19#1910 1-19#1916 1-16#1619]")
 		})
-		Convey("Wait 2 seconds for paths to expire, then query and get new paths", func() {
+		Convey("Wait 200ms for paths to expire, then query and get new paths", func() {
 			// Add new path between 1-10 and 1-16
 			g.AddLink("1-10", 101902, "1-19", 191002)
 			// Wait for two seconds to guarantee that the pathmgr refreshes the paths
-			<-time.After(2 * time.Second)
+			<-time.After(200 * time.Millisecond)
 			aps := pm.Query(srcIA, dstIA)
 			SoMsg("aps len", len(aps), ShouldEqual, 2)
 			SoMsg("path #1", getPathStrings(aps), ShouldContain,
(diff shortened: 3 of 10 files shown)


shitz:
Review status: 0 of 5 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[go/lib/pathmgr/main_test.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L75uWmV_WMYH16XI_co:-L75uWmWjVaEqMoh76En:bb6ht2i) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/pathmgr/main_test.go#L1)):*
> ```Go
> ```

2018

---

*[go/lib/pathmgr/pathmgr_test.go, line 41 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L75wV2fZwgA0H_B7qzp:-L75wV2fZwgA0H_B7qzq:b-ye40b2) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/pathmgr/pathmgr_test.go#L41)):*
> ```Go
> 			log.Root(),
> 		)
> 		SoMsg("pm", pm, ShouldNotBeNil)
> ```

You should have a separate unit test for `New` (if you really want to test that behavior) and always call `t.Fatal` for all errors that are not expected during the test. Applies to other places as well.

---

*[go/lib/pathmgr/pathmgr_test.go, line 51 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L75x1ukN-peFAD8864L:-L75x1ukN-peFAD8864M:bzqqn9) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/pathmgr/pathmgr_test.go#L51)):*
> ```Go
> 			Convey("Query immediately, same path is read from cache", func() {
> 				aps := pm.Query(srcIA, dstIA)
> 				SoMsg("aps len", len(aps), ShouldEqual, 1)
> ```

Technically, you should also test that the paths themselves are the same.

---

*[go/lib/pathmgr/pathmgr_test.go, line 67 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7Uj7gxWmgNYr3qBOGU:-L7Uj7gxWmgNYr3qBOGV:by6dr4l) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/pathmgr/pathmgr_test.go#L67)):*
> ```Go
> func TestQueryFilter(t *testing.T) {
> 	Convey("Create path manager", t, func() {
> 		g := graph.NewFromDescription(graph.DefaultGraphDescription)
> ```

Instead of exporting `DefaultGraphDescription` you should export `NewDefaultGraph()` 

---

*[go/lib/sciond/graph/graph.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7Uf_LnmABSPs9WjsJ-:-L7Uf_LnmABSPs9WjsJ0:b-7oe6p4) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L1)):*
> ```Go
> ```

I would move this package to `go/lib/xtest/graph/`

---

*[go/lib/sciond/graph/graph.go, line 35 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UdnuEnqNp4yG4NxWR:-L7UdnuEnqNp4yG4NxWS:bixuupj) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L35)):*
> ```Go
> type Graph struct {
> 	// maps IFIDs to the other IFID of the edge
> 	ifids map[uint64]uint64
> ```

would call this `links`

---

*[go/lib/sciond/graph/graph.go, line 77 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UgIR5EFWv7h109KOA:-L7UgIR5EFWv7h109KOB:bvyvimc) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L77)):*
> ```Go
> // xIFID in xIA and yIFID in yIA. If xIA or yIA are not valid string
> // representations of an ISD-AS, Connect panics.
> func (g *Graph) Connect(xIA string, xIFID uint64, yIA string, yIFID uint64) {
> ```

`AddLink`, `RemoveLink` would be more intuitive

---

*[go/lib/sciond/graph/graph.go, line 97 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UgbTDeduaC3WZzge3:-L7UgbTDeduaC3WZzge4:bla39sy) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L97)):*
> ```Go
> 
> // Disconnect deletes the edge containing ifid from the graph.
> func (g *Graph) Disconnect(ifid uint64) {
> ```

This assumes that `ifid` is unique in the graph. It should at least take the `IA` owning that interface.

---

*[go/lib/sciond/graph/graph.go, line 113 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UgxZD0P5f7Qw7faTM:-L7UgxZD0P5f7Qw7faTN:b-aggeyy) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L113)):*
> ```Go
> 
> // GetParent returns the parent AS of ifid.
> func (g *Graph) GetParent(ifid uint64) addr.IA {
> ```

Never called. Do you have a use-case in mind for this?

---

*[go/lib/sciond/graph/graph.go, line 128 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7Uhh1z6CwK9Nk5Rt2B:-L7Uhh2-f7_vgHNoQM2n:b-wdv066) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L128)):*
> ```Go
> 	dst := MustParseIA(yIA)
> 	solutionLength := 1000 // Infinity
> 	visitTrailQueue := []*testVisitTrail{
> ```

just `queue` would be enough

---

*[go/lib/sciond/graph/graph.go, line 131 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UhlRmrgpeSNvoSXY1:-L7UhlRmrgpeSNvoSXY2:b-2vw4n4) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L131)):*
> ```Go
> 		newTestVisitTrail(src),
> 	}
> 	solutionTrails := []*testVisitTrail{}
> ```

this could directly be a `[][]uint64` then you don't need to extract it at the end. Also call it `solutions`

---

*[go/lib/sciond/graph/graph.go, line 178 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UiBer2_jixZtkqEk7:-L7UiBer2_jixZtkqEk8:b-n5b7xd) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L178)):*
> ```Go
> // AS contains a list of all the IFIDs in an AS.
> type AS struct {
> 	IFIDs []uint64
> ```

Deleting would be easier if this was just `IFIDs map[uint64]struct{}`. Since you don't care about iteration order it should work fine.

---

*[go/lib/sciond/graph/graph.go, line 194 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UiQXooi9IuhoMBIKT:-L7UiQXooi9IuhoMBIKU:b-kkl3h0) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L194)):*
> ```Go
> // testVisitTrail tracks the state of a candidate solution for the graph
> // exploration in graph.GetPaths.
> type testVisitTrail struct {
> ```

Call this `Solution`

---

*[go/lib/sciond/graph/graph.go, line 221 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UiaizFa51qm4ZcxEm:-L7UiaizFa51qm4ZcxEn:bb086b) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L221)):*
> ```Go
> }
> 
> func (trail *testVisitTrail) Visited(ia addr.IA) bool {
> ```

Not a fan of these small getters and setters. `Add` and `Len` are ok I guess.

---

*[go/lib/sciond/graph/graph.go, line 259 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UioF0aNIgMXmC2f5d:-L7UioF0aNIgMXmC2f5e:b-ezj5hl) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L259)):*
> ```Go
> 
> // DescriptionEdge is used in Descriptions to describe the links between ASes.
> type DescriptionEdge struct {
> ```

`EdgeDesc`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1443)*

kormat:
Review status: 0 of 5 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[go/lib/sciond/graph/graph.go, line 35 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UghlKg856ni0n6JGZ:-L7UghlKg856ni0n6JG_:b2yg6jf) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L35)):*
> ```Go
> type Graph struct {
> 	// maps IFIDs to the other IFID of the edge
> 	ifids map[uint64]uint64
> ```

Might as well use `common.IFIDType` instead of `uint64`.

---

*[go/lib/sciond/graph/graph.go, line 122 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UiZY9E8Aom_fAGlPc:-L7UiZYAxzl06bsH1h29:bcef5le) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L122)):*
> ```Go
> // slice containing an empty path is returned. If no path exists between xIA
> // and yIA, a 0-length slice is returned.
> func (g *Graph) GetPaths(xIA string, yIA string) [][]uint64 {
> ```

This does not seem to have anything to do with scion path construction, which makes me wonder as to its purpose.  It will happily cross multiple peering links, cross a peering link and go _up_, take a downsegment and then an upsegment if that's shorter than going up, etc, all of which are illegal.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1443)*

shitz:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[go/lib/sciond/graph/graph.go, line 122 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UiZY9E8Aom_fAGlPc:-L7YaQ1b6kZABMaHzfKB:bjh2trm) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L122)):*

This does not seem to have anything to do with scion path construction, which makes me wonder as to its purpose.  It will happily cross multiple peering links, cross a peering link and go _up_, take a downsegment and then an upsegment if that's shorter than going up, etc, all of which are illegal.
</blockquote></details>

That is certainly true (and also noted as such in `sciond/mock.go`, but there are sufficiently many test cases that do not care about the paths themselves and whether they are valid SCION paths that I think this is a useful abstraction to have. Once we have path combinator in Go maybe this is not needed anymore, but for now I think it's useful.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1443)*

kormat:
Reviewed 1 of 5 files at r1.
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[go/lib/sciond/graph/graph.go, line 97 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UgbTDeduaC3WZzge3:-L7ZgUoxFdUxRvc3_DN6:bx0gur6) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L97)):*

This assumes that `ifid` is unique in the graph. It should at least take the `IA` owning that interface.
</blockquote></details>

That's documented in the assumptions at the top. It threw me too, but i think it's ok in this instance. But, it would be nice to ensure that duplicate IFIDs can't be used. Putting a check into `Connect` would solve that.

---

*[go/lib/sciond/graph/graph.go, line 113 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UgxZD0P5f7Qw7faTM:-L7ZguJKQvcqm2yite7A:bwb1dqr) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L113)):*

Never called. Do you have a use-case in mind for this?
</blockquote></details>

It's used in `sciond/mock.go`

---

*[go/lib/sciond/graph/graph.go, line 122 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UiZY9E8Aom_fAGlPc:-L7ZY98M2aKigPsuisj7:bejdjx6) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L122)):*

That is certainly true (and also noted as such in `sciond/mock.go`, but there are sufficiently many test cases that do not care about the paths themselves and whether they are valid SCION paths that I think this is a useful abstraction to have. Once we have path combinator in Go maybe this is not needed anymore, but for now I think it's useful.
</blockquote></details>

As discussed offline: let's go with this approach, just move the "these paths aren't real SCION paths, the value of these paths may fall as well as rise" disclaimer to to the godoc in this file.

---

*[go/lib/sciond/graph/graph.go, line 141 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7Zh8IbwGLjA40xL2l3:-L7Zh8IbwGLjA40xL2l4:bf18qqj) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L141)):*
> ```Go
> 		visitTrailQueue = visitTrailQueue[1:]
> 
> 		if currentTrail.Len() > solutionLength {
> ```

This doesn't look right. If i understand correctly, if you've already found a solution that's length 3, if you encounter _any_ solutions that are longer you stop looking?

---

*[go/lib/sciond/graph/graph.go, line 198 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7ZiRQEXr66_GdCzG8W:-L7ZiRQEXr66_GdCzG8X:b-ros97k) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L198)):*
> ```Go
> 	currentIA addr.IA
> 	// whether the AS has already been visited by this path, to avoid loops
> 	visited map[addr.IA]bool
> ```

This can also be `map[addr.Ia]struct{}`, and just check for presence.

---

*[go/lib/sciond/graph/graph.go, line 272 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7ZjFz2AN5oSu2tF9F9:-L7ZjFz2AN5oSu2tF9FA:b-v6dsrw) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L272)):*
> ```Go
> 	Edges: []DescriptionEdge{
> 		// Non-core ISD 1
> 		{"1-10", 1019, "1-19", 1910},
> ```

The order of these is ~random, making it very hard to follow the topology. The direction is also inconsistent (e.g. the first line has a child -> parent link, but the second line has parent -> child link). As a counter example, you could follow the order in `topology/Default.topo`'s `links` section.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1443)*

scrye:
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[go/lib/pathmgr/main_test.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L75uWmV_WMYH16XI_co:-L7ZpgIh9UH5VpN2_Mt_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/pathmgr/main_test.go#L1)):*

2018
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 41 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L75wV2fZwgA0H_B7qzp:-L7ZuM64kOl2AGpA6so9:bigrwbr) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/pathmgr/pathmgr_test.go#L41)):*

You should have a separate unit test for `New` (if you really want to test that behavior) and always call `t.Fatal` for all errors that are not expected during the test. Applies to other places as well.
</blockquote></details>

I don't think `New` deserves a test, so I only implemented the helper method.

---

*[go/lib/pathmgr/pathmgr_test.go, line 51 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L75x1ukN-peFAD8864L:-L7ZwUcpgHx0V3--U2is:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/pathmgr/pathmgr_test.go#L51)):*

Technically, you should also test that the paths themselves are the same.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 67 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7Uj7gxWmgNYr3qBOGU:-L7_7VhrIcdkp2xNYDmJ:b-3wwgtm) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/pathmgr/pathmgr_test.go#L67)):*

Instead of exporting `DefaultGraphDescription` you should export `NewDefaultGraph()` 
</blockquote></details>

Hah, I actually had exactly this in a prior version. I'm going to export both, because I want `DefaultGraphDescription` to remain public, s.t. users can see its contents in the `godoc`.

---

*[go/lib/sciond/graph/graph.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7Uf_LnmABSPs9WjsJ-:-L7_8p044gZg0nBtay6f:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L1)):*

I would move this package to `go/lib/xtest/graph/`
</blockquote></details>

Done.

---

*[go/lib/sciond/graph/graph.go, line 35 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UdnuEnqNp4yG4NxWR:-L7_8xNgUgceT98DYCK4:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L35)):*

would call this `links`
</blockquote></details>

Done.

---

*[go/lib/sciond/graph/graph.go, line 35 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UghlKg856ni0n6JGZ:-L7_9U0SmqoXiF7xPGzx:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L35)):*

Might as well use `common.IFIDType` instead of `uint64`.
</blockquote></details>

Done.

---

*[go/lib/sciond/graph/graph.go, line 77 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UgIR5EFWv7h109KOA:-L7_9cTNkuQYdq6909Vn:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L77)):*

`AddLink`, `RemoveLink` would be more intuitive
</blockquote></details>

Done.

---

*[go/lib/sciond/graph/graph.go, line 97 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UgbTDeduaC3WZzge3:-L7_9hnnOmDggWSJnqc4:blztyfy) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L97)):*

That's documented in the assumptions at the top. It threw me too, but i think it's ok in this instance. But, it would be nice to ensure that duplicate IFIDs can't be used. Putting a check into `Connect` would solve that.
</blockquote></details>

They're unique to keep mocking stuff (somewhat) simpler.

I've changed `Connect` (aka `AddLink`) to panic on duplicates.

---

*[go/lib/sciond/graph/graph.go, line 113 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UgxZD0P5f7Qw7faTM:-L7_AS79wDztAEGhfgux:b-5wt5df) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L113)):*

It's used in `sciond/mock.go`
</blockquote></details>

Paths in the graph consist only of IFIDs; this is used to fill in AS information in the SCIOND mock, while still providing concurrent-use safety.

---

*[go/lib/sciond/graph/graph.go, line 122 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UiZY9E8Aom_fAGlPc:-L7_BG2YiI5uO31fsoxr:b-sbsbg) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L122)):*

As discussed offline: let's go with this approach, just move the "these paths aren't real SCION paths, the value of these paths may fall as well as rise" disclaimer to to the godoc in this file.
</blockquote></details>

Also added a comment here to stress that they aren't necessarily SCION-valid paths.

---

*[go/lib/sciond/graph/graph.go, line 128 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7Uhh1z6CwK9Nk5Rt2B:-L7_BSniy0UwUuwHcwTS:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L128)):*

just `queue` would be enough
</blockquote></details>

Done.

---

*[go/lib/sciond/graph/graph.go, line 131 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UhlRmrgpeSNvoSXY1:-L7_CMwtdF4pk0fx2nZI:bcus9ab) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L131)):*

this could directly be a `[][]uint64` then you don't need to extract it at the end. Also call it `solutions`
</blockquote></details>

Nice idea, done.

---

*[go/lib/sciond/graph/graph.go, line 141 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7Zh8IbwGLjA40xL2l3:-L7_COexdjG9QVuVvWwC:bw5e6p7) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L141)):*

This doesn't look right. If i understand correctly, if you've already found a solution that's length 3, if you encounter _any_ solutions that are longer you stop looking?
</blockquote></details>

Yes. Because this is BFS, we explore solutions in increasing length. So as soon as we reach a candidate solution of length 4, we are guaranteed to no longer have any candidate solution of length 3.

---

*[go/lib/sciond/graph/graph.go, line 178 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UiBer2_jixZtkqEk7:-L7_DA4bzC1eI07Ig_Ej:b-fznogt) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L178)):*

Deleting would be easier if this was just `IFIDs map[uint64]struct{}`. Since you don't care about iteration order it should work fine.
</blockquote></details>

Good idea, done.

---

*[go/lib/sciond/graph/graph.go, line 194 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UiQXooi9IuhoMBIKT:-L7_E3_LVpbshXyqhdtU:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L194)):*

Call this `Solution`
</blockquote></details>

Done.

---

*[go/lib/sciond/graph/graph.go, line 198 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7ZiRQEXr66_GdCzG8W:-L7_EcTNcfkxOJcNlsmp:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L198)):*

This can also be `map[addr.Ia]struct{}`, and just check for presence.
</blockquote></details>

Done.

---

*[go/lib/sciond/graph/graph.go, line 221 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UiaizFa51qm4ZcxEm:-L7_ElHST09xAUoEbxC1:b6iu1qc) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L221)):*

Not a fan of these small getters and setters. `Add` and `Len` are ok I guess.
</blockquote></details>

I find it easier to reason about a collection if I can interact with it without ever caring about internals. It allows me to not care about coherency (not an issue here, specifically) and plumbing.

Granted, `currentIA` doesn't need this treatment since it'll never cause a side-effect, so I made it public to show this and deleted its getter/setter.

`Visited` is nastier and I'd prefer it stayed a method.

---

*[go/lib/sciond/graph/graph.go, line 259 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7UioF0aNIgMXmC2f5d:-L7_FxI1Uf1exHgQz4YP:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L259)):*

`EdgeDesc`
</blockquote></details>

Done.

---

*[go/lib/sciond/graph/graph.go, line 272 at r1](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7ZjFz2AN5oSu2tF9F9:-L7_GqxIxFyaa9x-LKcl:bz4gwza) ([raw file](https://github.com/scionproto/scion/blob/1160f9f7eec1b6aa2a25d326848e9e32877b1248/go/lib/sciond/graph/graph.go#L272)):*

The order of these is ~random, making it very hard to follow the topology. The direction is also inconsistent (e.g. the first line has a child -> parent link, but the second line has parent -> child link). As a counter example, you could follow the order in `topology/Default.topo`'s `links` section.
</blockquote></details>

It wasn't random, it was sorted by as number within the commented subcategories.

I've mirrored `topology/Default.topo`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1443)*

kormat:
Reviewed 2 of 5 files at r1, 4 of 6 files at r2.
Review status: 3 of 5 files reviewed at latest revision, 16 unresolved discussions.

---

*[go/lib/sciond/mock.go, line 91 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7_QijTEkhWDL55FA7y:-L7_QijTEkhWDL55FA7z:bfmeu0j) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/sciond/mock.go#L91)):*
> ```Go
> 				PathInterface{
> 					RawIsdas: m.g.GetParent(ifid).IAInt(),
> 					IfID:     uint64(ifid),
> ```

Ah. This indicates a type issue. Can you also change `PathInterface.IfID` to use `comon.IFIDType`, please? Same goes for the other uses of `uint64` as Ifid in `sciond/types.go`.

---

*[go/lib/xtest/graph/graph.go, line 132 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7_OLop5jv2Sh0wftpM:-L7_OLoq5AATGouXYc3C:b-88mnr0) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L132)):*
> ```Go
> // and yIA, a 0-length slice is returned.
> //
> // Note that this always returns shortest length paths, even if they might not
> ```

I think this caveat is too important to put on a method, rather than in the module docstring (or at least the type docstring). If it's in the module/type docstring then here can be "As noted above, this does not return valid SCION paths".

---

*[go/lib/xtest/graph/graph.go, line 207 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7_PdKU0QBBaVOXbwV8:-L7_PdKU0QBBaVOXbwV9:baflc31) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L207)):*
> ```Go
> }
> 
> func newTestVisitTrail(start addr.IA) *solution {
> ```

name vs type mismatch.

---

*[go/lib/xtest/graph/graph.go, line 214 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7_PfpEDaNoRrGlqrFv:-L7_PfpEDaNoRrGlqrFw:b-od0hxi) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L214)):*
> ```Go
> }
> 
> func (trail *solution) Copy() *solution {
> ```

receiver vs type name mismatch.

---

*[go/lib/xtest/graph/graph.go, line 231 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7_PxB_0ZILy5ytrsGd:-L7_PxB_0ZILy5ytrsGe:b-vw0jdr) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L231)):*
> ```Go
> 
> // Add appends localIFID and nextIFID to the trail, and advances to nextIA.
> func (trail *solution) Add(localIFID, nextIFID common.IFIDType, nextIA addr.IA) {
> ```

`nextIA` isn't _required_, as it can be fetched easily from `nextIFID`.

---

*[go/lib/xtest/graph/graph.go, line 250 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7_Q3iU2tMFhOy_OtlM:-L7_Q3iU2tMFhOy_OtlN:b-83q4fk) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L250)):*
> ```Go
> // FooBarDesc contains the entire specification of a graph. It is useful for
> // one shot initilizations.
> type FooBarDesc struct {
> ```

No :P

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1443)*

scrye:
Review status: 1 of 9 files reviewed at latest revision, 22 unresolved discussions.

---

*[go/lib/sciond/mock.go, line 91 at r2](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7_QijTEkhWDL55FA7y:-L7dTQHvWlnMY9U2kmrV:bjcxqbh) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/sciond/mock.go#L91)):*

Ah. This indicates a type issue. Can you also change `PathInterface.IfID` to use `comon.IFIDType`, please? Same goes for the other uses of `uint64` as Ifid in `sciond/types.go`.
</blockquote></details>

Done, hope I got them all.

---

*[go/lib/xtest/graph/graph.go, line 132 at r2](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7_OLop5jv2Sh0wftpM:-L7dVSTxEerIhz7HLhaY:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L132)):*

I think this caveat is too important to put on a method, rather than in the module docstring (or at least the type docstring). If it's in the module/type docstring then here can be "As noted above, this does not return valid SCION paths".
</blockquote></details>

Done.

---

*[go/lib/xtest/graph/graph.go, line 207 at r2](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7_PdKU0QBBaVOXbwV8:-L7dVWknRRnQiv-cWoc2:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L207)):*

name vs type mismatch.
</blockquote></details>

Done.

---

*[go/lib/xtest/graph/graph.go, line 214 at r2](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7_PfpEDaNoRrGlqrFv:-L7dVd8QnWude3RN5WFd:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L214)):*

receiver vs type name mismatch.
</blockquote></details>

Done.

---

*[go/lib/xtest/graph/graph.go, line 231 at r2](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7_PxB_0ZILy5ytrsGd:-L7dVrEUTiA3a1qNPUtu:bban9xd) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L231)):*

`nextIA` isn't _required_, as it can be fetched easily from `nextIFID`.
</blockquote></details>

We don't have a reference to the graph in this scope (and `solution` probably doesn't need to see the whole graph). I think it's more intuitive if the caller just supplies `nextIA`.

---

*[go/lib/xtest/graph/graph.go, line 250 at r2](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7_Q3iU2tMFhOy_OtlM:-L7dWJxEvwcrwc-R3fMP:b-rbzbuk) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L250)):*

No :P
</blockquote></details>

Yes :P

(also, fixed)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1443)*

shitz:
:lgtm:

---

Reviewed 4 of 6 files at r2, 6 of 6 files at r3.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1443#-:-L7eWu4C09RdhrienB8M:bnfp4nl)*

kormat:
Reviewed 2 of 6 files at r2, 6 of 6 files at r3.

---

*[go/lib/pathmgr/pathmgr_test.go, line 79 at r3](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7iRdIM0Mm9QtcYVrJL:-L7iRdIM0Mm9QtcYVrJM:bnhfb1l) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L79)):*
> ```Go
> 
> func TestRegister(t *testing.T) {
> ```

Blank line.

---

*[go/lib/pathmgr/pathmgr_test.go, line 94 at r3](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7iQG9YgOFxmzDhsc9D:-L7iQG9YgOFxmzDhsc9E:bjg1q2c) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L94)):*
> Re-add

(easier to parse)

---

*[go/lib/pathmgr/pathmgr_test.go, line 98 at r3](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7iQs8ktZTKcGCxsR4u:-L7iQs8ktZTKcGCxsR4v:bc3hyo6) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L98)):*
> ```Go
> 			// second), so it should contain the path after 4 seconds.
> 			g.AddLink("1-10", 1019, "1-19", 1910)
> 			<-time.After(4 * time.Second)
> ```

These sort of tests add quite a large amount of time to the go unit tests. One thing that would make it a lot less painful is if they honoured the `-short` flag to `go test`. There's an example in the Overview of the `testing` package: https://golang.org/pkg/testing/#pkg-overview. (CI would obviously run without `-short`)

---

*[go/lib/pathmgr/pathmgr_test.go, line 143 at r3](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7iTyITRkOdY1EEQz2S:-L7iTyITRkOdY1EEQz2T:b-f99co3) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L143)):*
> ```Go
> 		sp, err := pm.Watch(watchSrc, watchDst)
> 		SoMsg("err watch", err, ShouldBeNil)
> 		SoMsg("len(aps) watch", len(sp.Load().APS), ShouldEqual, 1)
> ```

It would be better to have a function for these two checks:
```
def apsCheck(desc string, sp *SyncPaths, expCount int, expValue string) {}
```
That way you can check all three syncPaths easily whenever the state changes. E.g. `"Revoke a path that's not part of any path seg"` would be better if it checked that none of the syncPath's were affected, etc.

---

*[go/lib/pathmgr/pathmgr_test.go, line 163 at r3](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7iV62c2GaWFni7HJ5S:-L7iV62c2GaWFni7HJ5T:b2eza5q) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L163)):*
> ```Go
> 		})
> 		Convey("Revoke a path that's in Query, but not in Watch/WatchFilter path sets", func() {
> 			// Disconnect #1619
> ```

`#1019`?

---

*[go/lib/pathmgr/pathmgr_test.go, line 164 at r3](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7iVOy-swd4gb3UScWR:-L7iVOy-swd4gb3UScWS:b-op5m34) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L164)):*
> ```Go
> 		Convey("Revoke a path that's in Query, but not in Watch/WatchFilter path sets", func() {
> 			// Disconnect #1619
> 			// Note that the revoke below only invalidates the cache and
> ```

The caveats make it sound like these last two tests aren't actually useful. I guess you _could_ call revoke directly on the graph, to simulate this more realistically.

---

*[go/lib/xtest/graph/graph.go, line 231 at r2](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7_PxB_0ZILy5ytrsGd:-L7iM2lB6SzkaSoXaAs6:b-b65bqw) ([raw file](https://github.com/scionproto/scion/blob/6ed65eca3670a0af2fe8fe98a7154dc51f127479/go/lib/xtest/graph/graph.go#L231)):*

We don't have a reference to the graph in this scope (and `solution` probably doesn't need to see the whole graph). I think it's more intuitive if the caller just supplies `nextIA`.
</blockquote></details>

Ah, yes. Good point.

---

*[go/lib/xtest/graph/graph.go, line 137 at r3](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7iLoaS9Uas5cJvQJF5:-L7iLoaS9Uas5cJvQJF6:b-vytjix) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/xtest/graph/graph.go#L137)):*
> paths

---

*[go/lib/xtest/graph/graph.go, line 154 at r3](https://reviewable.io:443/reviews/scionproto/scion/1443#-L7iMKEOcUIHWekseMed:-L7iMKEOcUIHWekseMee:b-ead28z) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/xtest/graph/graph.go#L154)):*
> ```Go
> 		}
> 		// Explore the next element in the queue.
> 		currentTrail := queue[0]
> ```

`curSolution`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1443)*

scrye:
Review status: 7 of 9 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/pathmgr/pathmgr_test.go, line 79 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7iRdIM0Mm9QtcYVrJL:-L7iZIMp2BFpKuS1r8X2:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L79)):*

Blank line.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 94 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7iQG9YgOFxmzDhsc9D:-L7iZKLS-wkAJl_VAO_t:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L94)):*

> Re-add

(easier to parse)
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 98 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7iQs8ktZTKcGCxsR4u:-L7iaPRR7AjhIxv1RSmN:bsr4sdn) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L98)):*

These sort of tests add quite a large amount of time to the go unit tests. One thing that would make it a lot less painful is if they honoured the `-short` flag to `go test`. There's an example in the Overview of the `testing` package: https://golang.org/pkg/testing/#pkg-overview. (CI would obviously run without `-short`)
</blockquote></details>

With the sciond mock we have now I can reduce the timers a lot (hopefully) without introducing flakiness. If we see failures in CI, we can refactor the tests and try other solutions.

---

*[go/lib/pathmgr/pathmgr_test.go, line 143 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7iTyITRkOdY1EEQz2S:-L7ifftvCa_go6MyOY3z:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L143)):*

It would be better to have a function for these two checks:
```
def apsCheck(desc string, sp *SyncPaths, expCount int, expValue string) {}
```
That way you can check all three syncPaths easily whenever the state changes. E.g. `"Revoke a path that's not part of any path seg"` would be better if it checked that none of the syncPath's were affected, etc.
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 163 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7iV62c2GaWFni7HJ5S:-L7ifilXBCktGiH3Z7xN:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L163)):*

`#1019`?
</blockquote></details>

Done.

---

*[go/lib/pathmgr/pathmgr_test.go, line 164 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7iVOy-swd4gb3UScWR:-L7ifjHTEYlBuLyZ6x8q:b-ck0kut) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L164)):*

The caveats make it sound like these last two tests aren't actually useful. I guess you _could_ call revoke directly on the graph, to simulate this more realistically.
</blockquote></details>

The two tests simulate receiving an SCMP revocation from the network (which is why they call private method `revoke`). Removing the link will make the pathmgr update the paths when it requeries, but will not actually test the revocation code.

---

*[go/lib/xtest/graph/graph.go, line 137 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7iLoaS9Uas5cJvQJF5:-L7igfwE0mqIv4mUMOkd:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/xtest/graph/graph.go#L137)):*

> paths
</blockquote></details>

Done.

---

*[go/lib/xtest/graph/graph.go, line 154 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7iMKEOcUIHWekseMed:-L7igkDCCZvYVFX0RBRN:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/xtest/graph/graph.go#L154)):*

`curSolution`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1443)*

kormat:
Reviewed 2 of 2 files at r4.

---

*[go/lib/pathmgr/pathmgr_test.go, line 164 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1443#-L7iVOy-swd4gb3UScWR:-L7iiF-c3_Qhkew4id4O:bplx4y0) ([raw file](https://github.com/scionproto/scion/blob/097c4ea700f1db24f7f3a32a75a633b5337a0aa7/go/lib/pathmgr/pathmgr_test.go#L164)):*

The two tests simulate receiving an SCMP revocation from the network (which is why they call private method `revoke`). Removing the link will make the pathmgr update the paths when it requeries, but will not actually test the revocation code.
</blockquote></details>

Ah, i was suggesting doing _both_. Revoke from the pm.cache as well as from the graph, and then do the tests.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1443)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1443#-:-L7jLtUA4Y7dHPRMWJgp:bnfp4nl)*

kormat:
Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1443)*

PR #1447, opened 2018-03-09
SCIOND and e2etest skeleton

scrye:
Very rough around the edges for now, basically just a proposal draft for how services can be tested and be built to share as much code as possible.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)


scrye:
:warning: PR frozen until #1455 is merged.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1447#-:-L7jFhsz-UzGxadbIOW1:bsya4hj)*

shitz:
Reviewed 2 of 8 files at r1.
Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[go/lib/infra/env/env.go, line 42 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1447#-L9GPwTE-Emn_eFxZDnm:-L9GPwTE-Emn_eFxZDnn:b-23a65x) ([raw file](https://github.com/scionproto/scion/blob/7cf9b14d7aaa6771bf8571cd640888e72beb6a9b/go/lib/infra/env/env.go#L42)):*
> ```Go
> 	// FIXME(scrye): Enable this when trust store becomes available.
> 	//databasePath = flag.String("trustdb", "trust.db", "Trust database file")
> 	prom = flag.String("prom", "127.0.0.1:1282", "Address to export prometheus metrics on")
> ```

I would leave this one empty by default and not start a prometheus endpoint if not set.

---

*[go/sciond/e2etest/server_test.go, line 39 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1447#-L9GTxJZ0uKFdYXfZHyr:-L9GTxJZ0uKFdYXfZHys:b-th19ml) ([raw file](https://github.com/scionproto/scion/blob/7cf9b14d7aaa6771bf8571cd640888e72beb6a9b/go/sciond/e2etest/server_test.go#L39)):*
> ```Go
> 	defer deleteDirTree()
> 	sock := xtest.MustTempFileName(dir, "rsock")
> 	stopServer := StartServer(t, dir, sock)
> ```

Lets got with `defer StartServer(t, dir, sock)()`. Once you get over the first bump it's super slick syntax

---

*[go/sciond/e2etest/server_test.go, line 102 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1447#-L9GPgxRE8EuANYP4Vz2:-L9GPgxRE8EuANYP4Vz3:b-iq8asa) ([raw file](https://github.com/scionproto/scion/blob/7cf9b14d7aaa6771bf8571cd640888e72beb6a9b/go/sciond/e2etest/server_test.go#L102)):*
> ```Go
> 		"-reliable", file,
> 		"-log.console", "debug",
> 		"-prom", fmt.Sprintf("127.0.0.1:%d", 30000+r.Intn(1000)),
> ```

I don't think we should run a prometheus endpoint as part of this test.

---

*[go/sciond/internal/servers/api.go, line 43 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1447#-L9KeUv32XVe7Y73lHxT:-L9KeUv32XVe7Y73lHxU:b-buvywy) ([raw file](https://github.com/scionproto/scion/blob/7cf9b14d7aaa6771bf8571cd640888e72beb6a9b/go/sciond/internal/servers/api.go#L43)):*
> ```Go
> 
> 	// State for request handlers
> 	handlers *handlers
> ```

I think I'd prefer if the there would be a generic handler interface and a map of handlers. Something like this
```
type handler interface {
    Handle(pld *sciond.Pld, src net.Addr)
}

type API struct {
    T transport.Transport
    handlers map[proto.SCIONDMsg_which]handler
}

func (srv *API) Handle(b common.RawBytes, addr net.Addr) error {
    ...
    go srv.handlers[p.Which].Handle(pld, addr)
}

...
func (h *PathRequestHandler) Handle(pld *sciond.Pld, src net.Addr) {
    req := pld.PathReq
    ...
} 
```

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1447)*

shitz:
Review status: 0 of 7 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[go/sciond/main.go, line 166 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1447#-L9eUUz35QAdk0-tShKV:-L9eUUz35QAdk0-tShKW:bl2oxwq) ([raw file](https://github.com/scionproto/scion/blob/7cf9b14d7aaa6771bf8571cd640888e72beb6a9b/go/sciond/main.go#L166)):*
> ```Go
> 	msger := messenger.New(dispatcher, trustStore, env.Log)
> 	// Include the below once we TrustStore is merged
> 	// msger.AddHandler(messenger.ChainRequest, trustStore.NewChainReqHandler())
> ```

As discussed on the design doc, this won't be needed anymore.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1447)*

scrye:
Review status: 0 of 7 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/lib/infra/env/env.go, line 42 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1447#-L9GPwTE-Emn_eFxZDnm:-LANoid5A-8X1c5_Kfau:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/7cf9b14d7aaa6771bf8571cd640888e72beb6a9b/go/lib/infra/env/env.go#L42)):*

I would leave this one empty by default and not start a prometheus endpoint if not set.
</blockquote></details>

Done.

---

*[go/sciond/main.go, line 166 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1447#-L9eUUz35QAdk0-tShKV:-LANowwND_YxJN1XTDpj:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/7cf9b14d7aaa6771bf8571cd640888e72beb6a9b/go/sciond/main.go#L166)):*

As discussed on the design doc, this won't be needed anymore.
</blockquote></details>

Done.

---

*[go/sciond/e2etest/server_test.go, line 39 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1447#-L9GTxJZ0uKFdYXfZHyr:-LANpQ7B57YuQg9A6_-B:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/7cf9b14d7aaa6771bf8571cd640888e72beb6a9b/go/sciond/e2etest/server_test.go#L39)):*

Lets got with `defer StartServer(t, dir, sock)()`. Once you get over the first bump it's super slick syntax
</blockquote></details>

Done.

---

*[go/sciond/e2etest/server_test.go, line 102 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1447#-L9GPgxRE8EuANYP4Vz2:-LANpVMkDWPpemwrmjA5:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/7cf9b14d7aaa6771bf8571cd640888e72beb6a9b/go/sciond/e2etest/server_test.go#L102)):*

I don't think we should run a prometheus endpoint as part of this test.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/api.go, line 43 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1447#-L9KeUv32XVe7Y73lHxT:-LAO1TUN02L2ljfvh-s3:b-4wrsay) ([raw file](https://github.com/scionproto/scion/blob/7cf9b14d7aaa6771bf8571cd640888e72beb6a9b/go/sciond/internal/servers/api.go#L43)):*

I think I'd prefer if the there would be a generic handler interface and a map of handlers. Something like this
```
type handler interface {
    Handle(pld *sciond.Pld, src net.Addr)
}

type API struct {
    T transport.Transport
    handlers map[proto.SCIONDMsg_which]handler
}

func (srv *API) Handle(b common.RawBytes, addr net.Addr) error {
    ...
    go srv.handlers[p.Which].Handle(pld, addr)
}

...
func (h *PathRequestHandler) Handle(pld *sciond.Pld, src net.Addr) {
    req := pld.PathReq
    ...
} 
```
</blockquote></details>

Nice, clean idea. Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1447)*

shitz:
Reviewed 1 of 8 files at r1, 6 of 6 files at r3.

---

*[go/sciond/internal/servers/api.go, line 89 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAnKkHV01rfIEIs7VDe:-LAnKkHV01rfIEIs7VDf:b-2rc6tp) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L89)):*
> ```Go
> 		return
> 	}
> 	go srv.handlers[p.Which].Handle(p, address)
> ```

This should check whether there is a handler for the message type instead of crashing.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1447)*

kormat:
Reviewed 1 of 8 files at r1, 6 of 6 files at r3.

---

*[go/lib/infra/env/env.go, line 34 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAn8XQ334RvKCtregED:-LAn8XQ334RvKCtregEE:b-vrz0gk) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/lib/infra/env/env.go#L34)):*
> ```Go
> 	"github.com/scionproto/scion/go/lib/common"
> 	liblog "github.com/scionproto/scion/go/lib/log"
> )
> ```

There needs to be a sighup chan setup ASAP. See https://github.com/scionproto/scion/blob/master/go/border/router.go#L41 for an example and the reasoning.

---

*[go/lib/infra/env/env.go, line 37 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAn5y-E80B8mvaUMNGv:-LAn5y-E80B8mvaUMNGw:bwjyjyw) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/lib/infra/env/env.go#L37)):*
> ```Go
> 
> var (
> 	id = flag.String("id", "", "Element ID (e.g., 'cs1-10-1'). (Required)")
> ```

Use the example AS: `cs4-ff00:0:2f`.

---

*[go/lib/infra/env/env.go, line 90 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAn8vQC7hzvGNdCysr6:-LAn8vQC7hzvGNdCysr7:b-vrfsi4) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/lib/infra/env/env.go#L90)):*
> ```Go
> func (env *Env) setupSignals() {
> 	env.AppShutdownSignal = make(chan struct{})
> 	sig := make(chan os.Signal, 1)
> ```

As a rule of thumb the channel should be sized to the number of signals it receives (so 2, here). In this case it doesn't make a huge amount of difference as both signals will exit the process anyway, but it's good to do for consistency.

---

*[go/sciond/main.go, line 60 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAnA2Q05jI3yJcy5cZv:-LAnA2Q05jI3yJcy5cZw:bg0rk14) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/main.go#L60)):*
> ```Go
> }
> 
> func realMain() int {
> ```

Any panics should be logged before exiting.

---

*[go/sciond/main.go, line 86 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAqlo0c8U12yXrN2c-f:-LAqlo0c8U12yXrN2c-g:b-6y9axk) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/main.go#L86)):*
> ```Go
> 		server, shutdownF := NewRSockServer(*reliableSockPath, Env)
> 		defer shutdownF()
> 		go func() {
> ```

Any panics should be logged. Same for the other go routines.

---

*[go/sciond/main.go, line 153 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAqm3vFD2KvQNW5u3B7:-LAqm3vFD2KvQNW5u3B8:b-t3inor) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/main.go#L153)):*
> ```Go
> }
> 
> func NewMessenger(scionAddress string, env *env.Env) (infra.Messenger, error) {
> ```

As this (would be) is called before the NewXServer functions, it should appear above them.

---

*[go/sciond/e2etest/server_test.go, line 1 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LArP46k70y_DE62fjGg:-LArP46k70y_DE62fjGh:b-w0imjk) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/e2etest/server_test.go#L1)):*
> ```Go
> ```

Why is this whole thing so slow to run?
```
$ go test ./go/sciond/e2etest/
ok      github.com/scionproto/scion/go/sciond/e2etest   8.104s
$ go test ./go/sciond/e2etest/
ok      github.com/scionproto/scion/go/sciond/e2etest   8.331s
```

---

*[go/sciond/e2etest/server_test.go, line 15 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAqmVZM49GbWP9qY8rT:-LAqmVZM49GbWP9qY8rU:b-a2ad0e) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/e2etest/server_test.go#L15)):*
> ```Go
> 
> // End to end test for SCIOND. This is a separate package because it is
> ```

I'm not sure what "end to end" means in this context. It seems like it's maybe an integration test?

`python/integration/end2end.py` is named like that because it, in theory, tests BS/PS/CS/SD/BR.

---

*[go/sciond/e2etest/server_test.go, line 43 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LArO0di4c0cqH3cqDo2:-LArO0di4c0cqH3cqDo3:bkyr2xm) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/e2etest/server_test.go#L43)):*
> ```Go
> 
> 	Convey("Send and receive ASInfo", t, func() {
> 		reply, err := conn.ASInfo(addr.IA{I: 1, A: 1})
> ```

Use ASes from the documentation range.

---

*[go/sciond/internal/servers/api.go, line 50 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LArKXEPEAPMU9w7kwsr:-LArKXEPEAPMU9w7kwss:bio9m2e) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L50)):*
> ```Go
> }
> 
> func (srv *API) Serve() error {
> ```

As this is called in a go routine, it should log any panics.

---

*[go/sciond/internal/servers/api.go, line 52 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAqnLtFEieFLu4JVTbI:-LAqnLtFEieFLu4JVTbJ:bt1j4ca) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L52)):*
> ```Go
> func (srv *API) Serve() error {
> 	// Initialize handler state if first time calling Serve
> 	if srv.handlers == nil {
> ```

This is what a constructor is for. I don't see a benefit to trying to hide this.

---

*[go/sciond/internal/servers/api.go, line 77 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAqnrTqA1ivApGFDPz4:-LAqnrTqA1ivApGFDPz5:b-hvzc4u) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L77)):*
> ```Go
> 			return err
> 		}
> 		go func() {
> ```

This should log any panics.

---

*[go/sciond/internal/servers/api.go, line 89 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAqo1bU27MjRKklLcPG:-LAqo1bU27MjRKklLcPH:bm5ycul) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L89)):*
> ```Go
> 		return
> 	}
> 	go srv.handlers[p.Which].Handle(p, address)
> ```

Why a second goroutine? `Server()` already runs `Handle()` in a separate goroutine per message.

(If they are in another go routine, then each handler must also log panics)

---

*[go/sciond/internal/servers/handlers.go, line 37 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAqoTx7Cuxs7oH0jrWC:-LAqoTx7Cuxs7oH0jrWD:b-9pilgq) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/handlers.go#L37)):*
> ```Go
> 
> 	// For network traffic (i.e., sending out path requests) include reference
> 	// to Messenger or maybe a throttler
> ```

I'm not sure how to parse this.

---

*[go/sciond/internal/servers/rsock.go, line 37 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAqqyN9DV2RbXO8ld_5:-LAqqyN9DV2RbXO8ld_6:b-xhcc0c) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/rsock.go#L37)):*
> ```Go
> 	Address   string
> 	Messenger infra.Messenger
> 	Log       log.Logger
> ```

These fields don't look like they need to be exported.

---

*[go/sciond/internal/servers/rsock.go, line 39 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LAqpctOB6YORV3NWd0W:-LAqpctOB6YORV3NWd0X:b-7ydgb3) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/rsock.go#L39)):*
> ```Go
> 	Log       log.Logger
> 
> 	mu       sync.Mutex // protect access to listener during init/close
> ```

There's no need to do this. You have a constructor for this type, but it's in `main.go` instead of here. If you move it to here, then you can do all the setup properly, and simplify things a bit.

---

*[go/sciond/internal/servers/rsock.go, line 64 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LArLGRq86S4iEmfK3PB:-LArLGRq86S4iEmfK3PC:bgcgqga) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/rsock.go#L64)):*
> ```Go
> 		}
> 
> 		// Launch SCIONDMsg server on the accepted conn
> ```

Is "SCIONDMsg" an old name?

---

*[go/sciond/internal/servers/unix.go, line 46 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LArMpXb2jfhmtGim30P:-LArMpXb2jfhmtGim30Q:b-orkwot) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L46)):*
> ```Go
> // address. The server handles SCIOND messages coming from clients, intepreting
> // data as raw capnp. Each capnp SCIOND message must be entirely contained in a
> // single UNIX datagram. If a datagram contains data after the capnp message,
> ```

Extra data should be treated as an error.

---

*[go/sciond/internal/servers/unix.go, line 49 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1447#-LArNhE12QKTSjfjQNF5:-LArNhE12QKTSjfjQNF6:ber2vcy) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L49)):*
> ```Go
> // the extra data is ignored. Messages received from clients that have not
> // bound to a local UNIX address are logged and discarded, as there is no way
> // to answer to them (see man 7 unix for more info).
> ```

This sounds.. odd, and i can't find a mention in the manpage. Can you be more specific?

(Also i don't see any code that actually does this)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1447)*

scrye:
Review status: 0 of 7 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/env/env.go, line 34 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAn8XQ334RvKCtregED:-LArR7AV9A-FgszWoqjw:b2fcect) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/lib/infra/env/env.go#L34)):*
Done.

Note that a possibly better approach here is to start the process (from the shell, or wherever) with `SIGHUP` ignored. As soon as `signal.Notify` is called, the signal will be handled again.

From https://golang.org/pkg/os/signal/:
> If the program was started with SIGHUP or SIGINT ignored, and Notify is called for either signal, a signal handler will be installed for that signal and it will no longer be ignored. 

---

*[go/lib/infra/env/env.go, line 37 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAn5y-E80B8mvaUMNGv:-LArVIhw-IyI7JQn7AEF:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/lib/infra/env/env.go#L37)):*

Use the example AS: `cs4-ff00:0:2f`.
</blockquote></details>

Done.

---

*[go/lib/infra/env/env.go, line 90 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAn8vQC7hzvGNdCysr6:-LArTL8940caa4LmcOjD:b-lgvzh0) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/lib/infra/env/env.go#L90)):*

As a rule of thumb the channel should be sized to the number of signals it receives (so 2, here). In this case it doesn't make a huge amount of difference as both signals will exit the process anyway, but it's good to do for consistency.
</blockquote></details>

What is the reasoning behind the size of the channel? It doesn't guarantee that both will have space in the buffer.

The process can still receive two `SIGTERM`s (and the go runtime will happily put them in the channel), and the other signal won't fit.

---

*[go/sciond/main.go, line 60 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAnA2Q05jI3yJcy5cZv:-LArX9NDDHg0y6a3FWdB:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/main.go#L60)):*

Any panics should be logged before exiting.
</blockquote></details>

Done.

---

*[go/sciond/main.go, line 86 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqlo0c8U12yXrN2c-f:-LArXNaT3fSaeoSV3hSP:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/main.go#L86)):*

Any panics should be logged. Same for the other go routines.
</blockquote></details>

Done.

---

*[go/sciond/main.go, line 153 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqm3vFD2KvQNW5u3B7:-LArXTb50owQX66rwfpq:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/main.go#L153)):*

As this (would be) is called before the NewXServer functions, it should appear above them.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/api.go, line 50 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArKXEPEAPMU9w7kwsr:-LAr_ZxcCTkiGtt9S2nS:b-2tj1f9) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L50)):*

As this is called in a go routine, it should log any panics.
</blockquote></details>

Done. I've put the panic logs outside, because some callers spawn a goroutine (rsock server), some don't (unix sock_dgram server).

---

*[go/sciond/internal/servers/api.go, line 52 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqnLtFEieFLu4JVTbI:-LAr_GukBkKWFHplUKxJ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L52)):*

This is what a constructor is for. I don't see a benefit to trying to hide this.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/api.go, line 77 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqnrTqA1ivApGFDPz4:-LArep-q9kLMkphxttd5:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L77)):*

This should log any panics.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/api.go, line 89 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqo1bU27MjRKklLcPG:-LArfWtm5iZTTV7BtWTu:buerdxo) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L89)):*

Why a second goroutine? `Server()` already runs `Handle()` in a separate goroutine per message.

(If they are in another go routine, then each handler must also log panics)
</blockquote></details>

Ah, yes, the extra goroutine is an oversight. Done.

---

*[go/sciond/internal/servers/api.go, line 89 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAnKkHV01rfIEIs7VDe:-LArfZQ3CooMnaCOra33:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L89)):*

This should check whether there is a handler for the message type instead of crashing.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/handlers.go, line 37 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqoTx7Cuxs7oH0jrWC:-LArg94k7yM1IAjquC-s:bs7k03a) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/handlers.go#L37)):*

I'm not sure how to parse this.
</blockquote></details>

Outdated, deleted.

---

*[go/sciond/internal/servers/rsock.go, line 37 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqqyN9DV2RbXO8ld_5:-LArkyA9-6xcI0JZR2iC:bxzwe0z) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/rsock.go#L37)):*

These fields don't look like they need to be exported.
</blockquote></details>

Since in this approach the constructor only sets some fields, it feels cleaner for users of the package to do that directly.

---

*[go/sciond/internal/servers/rsock.go, line 39 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqpctOB6YORV3NWd0W:-LArl1ymBvbn4IP8memx:bgxv32l) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/rsock.go#L39)):*

There's no need to do this. You have a constructor for this type, but it's in `main.go` instead of here. If you move it to here, then you can do all the setup properly, and simplify things a bit.
</blockquote></details>

It's a clean pattern  to have the constructor only set up the fields and never fail, and have `ListenAndServe` do the actual listening that might fail. In this approach, the mutex is needed to prevent races between `Listen` and `Close`.

I don't think the logic gets simpler in the other approach, as the API works a bit clunkier (my constructor can fail now, since it also listens, and then Serve can probably fail again, so I need to handle errors twice).

---

*[go/sciond/internal/servers/rsock.go, line 64 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArLGRq86S4iEmfK3PB:-LArly_aAm096FYY-Tm_:b-kunt6b) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/rsock.go#L64)):*

Is "SCIONDMsg" an old name?
</blockquote></details>

It refers to the type of capnp message that is exchanged, not the server name. Reworded.

---

*[go/sciond/internal/servers/unix.go, line 46 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArMpXb2jfhmtGim30P:-LAroURTA_SCCZJjFx6R:b6v5wic) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L46)):*

Extra data should be treated as an error.
</blockquote></details>

Hmm, `proto.ParseFromReader` doesn't give any information about how much data it consumes from the reader. How can we detect trailing data?

---

*[go/sciond/internal/servers/unix.go, line 49 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArNhE12QKTSjfjQNF5:-LAromrZ7xQUldL9CjH2:b-qyb5hj) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L49)):*

This sounds.. odd, and i can't find a mention in the manpage. Can you be more specific?

(Also i don't see any code that actually does this)
</blockquote></details>

The `man 7 unix` bit was a hint on where to find more documentation, it doesn't explain why it's impossible to answer to those messages. Reworded.

This is a connection-less UNIX socket, and `n` parties can communicate on it. They can send messages without a source address, but in that case the receiver cannot know where to send back the response.

Because we don't have any library support for clients on UNIX SOCK_DGRAM, part of the functionality isn't implemented yet. I've added a FIXME.

---

*[go/sciond/e2etest/server_test.go, line 1 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArP46k70y_DE62fjGg:-LArXrrg0R0zkhroj04c:b-topyn5) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/e2etest/server_test.go#L1)):*

Why is this whole thing so slow to run?
```
$ go test ./go/sciond/e2etest/
ok      github.com/scionproto/scion/go/sciond/e2etest   8.104s
$ go test ./go/sciond/e2etest/
ok      github.com/scionproto/scion/go/sciond/e2etest   8.331s
```
</blockquote></details>

It rebuilds `sciond` for every test to keep everything self-contained.

Since it's a bit expensive, we can skip it in CI with a build flag.

---

*[go/sciond/e2etest/server_test.go, line 15 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqmVZM49GbWP9qY8rT:-LArYRrX9TAECtWXqSAb:b9hqka9) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/e2etest/server_test.go#L15)):*

I'm not sure what "end to end" means in this context. It seems like it's maybe an integration test?

`python/integration/end2end.py` is named like that because it, in theory, tests BS/PS/CS/SD/BR.
</blockquote></details>

It's a full application test, so it's higher level compared to an integration test. When I chose the name I was thinking that from SCIOND's perspective it will interact with other servers and clients, so it will be pretty close to an e2e test.

---

*[go/sciond/e2etest/server_test.go, line 43 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArO0di4c0cqH3cqDo2:-LArZShk9Zr9NpGgiPju:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/e2etest/server_test.go#L43)):*

Use ASes from the documentation range.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)*

kormat:
Reviewed 7 of 9 files at r4.

---

*[go/lib/infra/env/env.go, line 90 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAn8vQC7hzvGNdCysr6:-LAsNGiy5s8ZDFVSkMoH:bp4ty06) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/lib/infra/env/env.go#L90)):*

What is the reasoning behind the size of the channel? It doesn't guarantee that both will have space in the buffer.

The process can still receive two `SIGTERM`s (and the go runtime will happily put them in the channel), and the other signal won't fit.
</blockquote></details>

If you're expecting a lot of signals, then you'd have to make it bigger again of course. But if the expectation is that you may receive "one of these signals occasionally", the extra buffer means that if both signals happen at the ~same time that one isn't lost.

---

*[go/sciond/internal/servers/api.go, line 77 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqnrTqA1ivApGFDPz4:-LAsO9x0-9V-q2K-cJMQ:b-b951pt) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L77)):*

Done.
</blockquote></details>

Missing `defer`

---

*[go/sciond/internal/servers/rsock.go, line 37 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqqyN9DV2RbXO8ld_5:-LAwQ-D4C_NXgwOeTKae:b-9igb8m) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/rsock.go#L37)):*
> the constructor only sets some fields

That's entirely normal, so i don't get why this has to be treated differently. Exposing fields that should only be set at construction and should never be accessed from outside doesn't seem clean at all, to me.

---

*[go/sciond/internal/servers/rsock.go, line 39 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqpctOB6YORV3NWd0W:-LAwQZGv-TQ-ceBAjsol:b-riikef) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/rsock.go#L39)):*
> It's a clean pattern to have the constructor only set up the fields and never fail, and have ListenAndServe do the actual listening that might fail. In this approach, the mutex is needed to prevent races between Listen and Close.

Ok, that's fair. :+1: 

---

*[go/sciond/internal/servers/rsock.go, line 67 at r4](https://reviewable.io/reviews/scionproto/scion/1447#-LAvi9_A3p-cLjAfbIzj:-LAvi9_A3p-cLjAfbIzk:b-b951pt) ([raw file](https://github.com/scionproto/scion/blob/42eb9d306f40b574cf198c715318c984ac566656/go/sciond/internal/servers/rsock.go#L67)):*
> ```Go
> 		// Launch server for SCIONDMsg messages on the accepted conn
> 		go func() {
> 			liblog.LogPanicAndExit()
> ```

Missing `defer`

---

*[go/sciond/internal/servers/unix.go, line 46 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArMpXb2jfhmtGim30P:-LAviixEDNS5lgPFcv2O:bhiuauo) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L46)):*

Hmm, `proto.ParseFromReader` doesn't give any information about how much data it consumes from the reader. How can we detect trailing data?
</blockquote></details>

Ah. Good point. The underlying `capnp.Decoder` assumes it's operating on a stream, so it doesn't actually care. It is something we could do in `proto.ReadRootFromRaw` though:
```
buf := bytes.NewBuffer(b)
s, err := ReadRootFromReader(buf)
if err != nil {
   return capnp.Struct{}, err
}
if buf.Len() > 0 {
   // return error about trailing data
}
return s, nil
```
But i think it's probably not worth tackling now.

---

*[go/sciond/internal/servers/unix.go, line 49 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArNhE12QKTSjfjQNF5:-LAwKc8MEkz7-Hgj77Wj:b-kmqhld) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L49)):*

The `man 7 unix` bit was a hint on where to find more documentation, it doesn't explain why it's impossible to answer to those messages. Reworded.

This is a connection-less UNIX socket, and `n` parties can communicate on it. They can send messages without a source address, but in that case the receiver cannot know where to send back the response.

Because we don't have any library support for clients on UNIX SOCK_DGRAM, part of the functionality isn't implemented yet. I've added a FIXME.
</blockquote></details>

What about using UNIX SOCK_SEQPACKET instead? It still handles the framing, allows clients to cleanly have multiple connections, doesn't leave socket files that need to be cleaned up after a crash, reliable, guarantees ordering etc (which is not guaranteed on all platforms with SOCK_DGRAM). It's supported at linux and freebsd, at a minimum.

---

*[go/sciond/e2etest/server_test.go, line 1 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArP46k70y_DE62fjGg:-LAwOemD62xBr-TilCP7:b-lcxx1t) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/e2etest/server_test.go#L1)):*

It rebuilds `sciond` for every test to keep everything self-contained.

Since it's a bit expensive, we can skip it in CI with a build flag.
</blockquote></details>

This really doesn't seem like a good approach. I would far rather lose "self-containment", and have a test that can be usefully run in a reasonable amount of time. Having to edit test files to change build-flags is also an anti-goal, in my mind.

This approach, by its nature, hard-codes build flags. If you change the actual project's build flags (in `go/Makefile`) and don't remember to update all test files appropriately, then you'll be testing something different than what is actually run. If you want to manually test with a given build flag, you have to edit all the relevant test files too. This seems like a real step backwards.

---

*[go/sciond/e2etest/server_test.go, line 15 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqmVZM49GbWP9qY8rT:-LAwOn7UCXQu6q9kAWHA:btdbrr4) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/e2etest/server_test.go#L15)):*

It's a full application test, so it's higher level compared to an integration test. When I chose the name I was thinking that from SCIOND's perspective it will interact with other servers and clients, so it will be pretty close to an e2e test.
</blockquote></details>

Having had a quick google, i see that e2e is a more general term than i had in mind :) :+1: 

---

*[go/sciond/server_test.go, line 17 at r4](https://reviewable.io/reviews/scionproto/scion/1447#-LAwPDq64ox1bm8q0q7o:-LAwPDq64ox1bm8q0q7p:bsuo2kr) ([raw file](https://github.com/scionproto/scion/blob/42eb9d306f40b574cf198c715318c984ac566656/go/sciond/server_test.go#L17)):*
> ```Go
> // End to end test for SCIOND. This is a separate package because it is
> // expected to grow. If it doesn't, we just merge it into go/sciond.
> package main_test
> ```

I've never seen a test file be in a different package than the rest of the dir. While it seems to work, it would be better to just use `package main`, unless there's a reason i'm missing for this.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)*

scrye:
Review status: 1 of 7 files reviewed at latest revision, 10 unresolved discussions.

---

*[go/lib/infra/env/env.go, line 90 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAn8vQC7hzvGNdCysr6:-LAx66tW-9VqrUaB1PbP:badwltb) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/lib/infra/env/env.go#L90)):*

If you're expecting a lot of signals, then you'd have to make it bigger again of course. But if the expectation is that you may receive "one of these signals occasionally", the extra buffer means that if both signals happen at the ~same time that one isn't lost.
</blockquote></details>

Eh, 1, 2 or 42 makes no difference here. Done.

---

*[go/sciond/internal/servers/api.go, line 77 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqnrTqA1ivApGFDPz4:-LAx5qMA4vH9cDeDnlfv:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/api.go#L77)):*

Missing `defer`
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/rsock.go, line 37 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LAqqyN9DV2RbXO8ld_5:-LAxGVvt2ubRGJmBfTt9:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/rsock.go#L37)):*

> the constructor only sets some fields

That's entirely normal, so i don't get why this has to be treated differently. Exposing fields that should only be set at construction and should never be accessed from outside doesn't seem clean at all, to me.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/rsock.go, line 67 at r4](https://reviewable.io/reviews/scionproto/scion/1447#-LAvi9_A3p-cLjAfbIzj:-LAx3vnS7WYUn5jNgznl:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/42eb9d306f40b574cf198c715318c984ac566656/go/sciond/internal/servers/rsock.go#L67)):*

Missing `defer`
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/unix.go, line 49 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArNhE12QKTSjfjQNF5:-LAwyMAz3thN34WfX20G:b-8evvos) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L49)):*

What about using UNIX SOCK_SEQPACKET instead? It still handles the framing, allows clients to cleanly have multiple connections, doesn't leave socket files that need to be cleaned up after a crash, reliable, guarantees ordering etc (which is not guaranteed on all platforms with SOCK_DGRAM). It's supported at linux and freebsd, at a minimum.
</blockquote></details>

While SOCK_SEQPACKET is definitely a major improvement over SOCK_STREAM, I don't think connection oriented protocols are the best fit for this type of communication.

Advantages of SOCK_SEQPACKET over SOCK_DGRAM:
  - every exchange is its own connection, so no client address is needed (and bind is replaced with connect);
  - once the connection is closed, sock files get removed.

Advantages of SOCK_DGRAM over SOCK_SEQPACKET:
  - no connect/accept calls needed; this is especially useful on sciond restarts, which would no longer need reconnect logic on clients. If sciond goes down, clients send datagrams but no one is there to receive them so requests time out. Once it goes back up, newly sent datagrams reach the new process.

Non-issues:
  - mux/demux - we already include exchange demux information in the capnp ID fields, and clients would need to do the matching anyway if multiple goroutines want to communicate over the same SEQPACKET socket (unless strict ordering is assumed, but this would need queues and just feels messier in general)
  - ordering - application to sciond communication is single-shot, so ordering is not important

---

*[go/sciond/e2etest/server_test.go, line 1 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArP46k70y_DE62fjGg:-LAx6gZI2swj-2ejsNik:b9gsadl) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/e2etest/server_test.go#L1)):*

This really doesn't seem like a good approach. I would far rather lose "self-containment", and have a test that can be usefully run in a reasonable amount of time. Having to edit test files to change build-flags is also an anti-goal, in my mind.

This approach, by its nature, hard-codes build flags. If you change the actual project's build flags (in `go/Makefile`) and don't remember to update all test files appropriately, then you'll be testing something different than what is actually run. If you want to manually test with a given build flag, you have to edit all the relevant test files too. This seems like a real step backwards.
</blockquote></details>

Changed the test to use the existing binary file.

---

*[go/sciond/server_test.go, line 17 at r4](https://reviewable.io/reviews/scionproto/scion/1447#-LAwPDq64ox1bm8q0q7o:-LAxGQiG-JO_cdQZ_MaI:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/42eb9d306f40b574cf198c715318c984ac566656/go/sciond/server_test.go#L17)):*

I've never seen a test file be in a different package than the rest of the dir. While it seems to work, it would be better to just use `package main`, unless there's a reason i'm missing for this.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)*

shitz:
Review status: 1 of 7 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[go/sciond/server_test.go, line 17 at r4](https://reviewable.io/reviews/scionproto/scion/1447#-LAwPDq64ox1bm8q0q7o:-LAxMAJS1tClG0pyBTgV:b-3px05c) ([raw file](https://github.com/scionproto/scion/blob/42eb9d306f40b574cf198c715318c984ac566656/go/sciond/server_test.go#L17)):*

Done.
</blockquote></details>

`*_test` packages are used for tests that only test the public API of a package and nothing else. I guess it's to ensure that the developer of a package needs to work solely with exported fields and methods. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)*

kormat:
Reviewed 6 of 7 files at r5.

---

*[go/sciond/internal/servers/unix.go, line 49 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArNhE12QKTSjfjQNF5:-LAxPHjM-zyjnw6sRr6S:briyqdg) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L49)):*
For SCIOND the connectionless approach might work, but the dispatcher definitely needs to be connection-oriented because it's stateful, and it would be really good to use the same approach for both.

> If sciond goes down, clients send datagrams but no one is there to receive them so requests time out. 
> Once it goes back up, newly sent datagrams reach the new process.

I don't think this works. sciond would need to remove the socket file on shutdown (or if it was unclean, systemd etc would need to remove the socket file on start), otherwise `bind()` fails (as bind _creates_ the file, and fails if it already exists).

It's not a timeout btw: you get connection reset if the listener goes away, and of course connection refused if either the socket isn't there, or is isn't being listened on. You can probably work around this by treating all/most errors as a timeout.

(I made a tiny toy server/client to test with: https://gist.github.com/kormat/250df8762718719f2bc7944f6f1322aa)

---

*[go/sciond/server_test.go, line 17 at r4](https://reviewable.io/reviews/scionproto/scion/1447#-LAwPDq64ox1bm8q0q7o:-LAxNtq0EmgMrzsjGFN3:bgrxfpx) ([raw file](https://github.com/scionproto/scion/blob/42eb9d306f40b574cf198c715318c984ac566656/go/sciond/server_test.go#L17)):*

`*_test` packages are used for tests that only test the public API of a package and nothing else. I guess it's to ensure that the developer of a package needs to work solely with exported fields and methods. 
</blockquote></details>

@shitz : oh, neat idea.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)*

scrye:

---

*[go/sciond/internal/servers/unix.go, line 49 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArNhE12QKTSjfjQNF5:-LAxXfOS5iGbW8qdYlmV:b-l7awzr) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L49)):*
The dispatcher is a different beast, and there `SOCK_SEQPACKET` makes perfect sense. I don't think having the same socket type helps though. For example, in Go the only difference is changing `unixgram` to `unixpacket`, and not running the connect and accept loop. It leads to much cleaner code for sciond.

>  don't think this works. sciond would need to remove the socket file on shutdown (or if it was unclean, systemd etc would need to remove the socket file on start), otherwise bind() fails (as bind creates the file, and fails if it already exists).

sciond doesn't need to remove the socket on shutdown. As you point out, both not having the file and nobody listening on it yields the same error, "connection refused". Most apps just delete the socket before trying to bind (e.g., `syslogd` does this). I tested this out and it works ok.

Go toy example here: https://gist.github.com/scrye/dda0565722d88bdbc9dbde29812c553b. Server can restart multiple times, and the client can just keeps sending.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)*

kormat:

---

*[go/sciond/internal/servers/unix.go, line 49 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArNhE12QKTSjfjQNF5:-LBVTE_05k6G1rtArnLF:b-npjzf5) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L49)):*

The dispatcher is a different beast, and there `SOCK_SEQPACKET` makes perfect sense. I don't think having the same socket type helps though. For example, in Go the only difference is changing `unixgram` to `unixpacket`, and not running the connect and accept loop. It leads to much cleaner code for sciond.

>  don't think this works. sciond would need to remove the socket file on shutdown (or if it was unclean, systemd etc would need to remove the socket file on start), otherwise bind() fails (as bind creates the file, and fails if it already exists).

sciond doesn't need to remove the socket on shutdown. As you point out, both not having the file and nobody listening on it yields the same error, "connection refused". Most apps just delete the socket before trying to bind (e.g., `syslogd` does this). I tested this out and it works ok.

Go toy example here: https://gist.github.com/scrye/dda0565722d88bdbc9dbde29812c553b. Server can restart multiple times, and the client can just keeps sending.
</blockquote></details>

After some offline discussion, the plan is to use the same `SOCK_SEQPACKET` approach for both sciond's local API and the dispatcher (in a future rewrite). This way we get the same low-level interface for both.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)*

scrye:
Review status: 3 of 9 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/sciond/internal/servers/unix.go, line 49 at r3](https://reviewable.io/reviews/scionproto/scion/1447#-LArNhE12QKTSjfjQNF5:-LBeoLcK9eKjnzidqze6:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/39fc12e22e6da11f3602e9573fce5c607a663d4c/go/sciond/internal/servers/unix.go#L49)):*

After some offline discussion, the plan is to use the same `SOCK_SEQPACKET` approach for both sciond's local API and the dispatcher (in a future rewrite). This way we get the same low-level interface for both.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)*

kormat:
Reviewed 7 of 10 files at r6.

---

*[go/sciond/internal/servers/server.go, line 60 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBuwVnw2j9tPecz5EHm:-LBuwVnw2j9tPecz5EHn:bro41yt) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L60)):*
> ```Go
> // ListenAndServe listens on the UNIX stream socket at address, and
> // repeatedly accepts connections from clients. Stream data is interpreted
> // according to the Reliable Socket protocol described in go/lib/sock/reliable.
> ```

Needs updating for unixpacket.

---

*[go/sciond/internal/servers/server.go, line 69 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBuwb_Z9Gfo7i0stUr3:-LBuwb_Z9Gfo7i0stUr4:bro41yt) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L69)):*
> ```Go
> 	srv.mu.Unlock()
> 	if err != nil {
> 		return common.NewBasicError("unable to listen on reliable socket", nil,
> ```

Needs updating for unixpacket.

---

*[go/sciond/internal/servers/server.go, line 77 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBuwdjuERAhlomwRO0I:-LBuwdjuERAhlomwRO0J:bro41yt) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L77)):*
> ```Go
> 		conn, err := srv.listener.Accept()
> 		if err != nil {
> 			srv.log.Warn("unable to accept reliable socket conn", "err", err)
> ```

Needs updating for unixpacket.

---

*[go/sciond/internal/servers/server.go, line 105 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBux-Su6J3s0jnOPhuz:-LBux-Su6J3s0jnOPhv-:blgt1r4) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L105)):*
> ```Go
> }
> 
> // Close makes the ReliableSockServer stop listening for new reliable socket
> ```

Needs updating for unixpacket and rename.

---

*[go/sciond/internal/servers/server.go, line 113 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBux6AgDcuVypdelupQ:-LBux6AgDcuVypdelupR:b-v2mu46) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L113)):*
> ```Go
> 
> 	if srv.listener == nil {
> 		return common.NewBasicError("unitialized RSockServer", nil)
> ```

Needs update for rename.

---

*[go/sciond/internal/servers/server.go, line 118 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBux97WEBub3U1OqN-1:-LBux97WEBub3U1OqN-2:blgt1r4) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L118)):*
> ```Go
> }
> 
> // Shutdown makes the ReliableSockServer stop listening for new reliable socket
> ```

Needs updating for unixpacket and rename.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)*

scrye:

---

*[go/sciond/internal/servers/server.go, line 60 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBuwVnw2j9tPecz5EHm:-LC-D_Kv202NYTOQCXda:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L60)):*

Needs updating for unixpacket.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/server.go, line 69 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBuwb_Z9Gfo7i0stUr3:-LC-DdQl85pdG7XHYWt2:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L69)):*

Needs updating for unixpacket.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/server.go, line 77 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBuwdjuERAhlomwRO0I:-LC-Dpe4DRhidBQbB7ob:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L77)):*

Needs updating for unixpacket.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/server.go, line 105 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBux-Su6J3s0jnOPhuz:-LC-E2zjBUaHGREjOxeR:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L105)):*

Needs updating for unixpacket and rename.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/server.go, line 113 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBux6AgDcuVypdelupQ:-LC-EBJhDLy1KuI28uaW:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L113)):*

Needs update for rename.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/server.go, line 118 at r6](https://reviewable.io/reviews/scionproto/scion/1447#-LBux97WEBub3U1OqN-1:-LC-EGSS3O-THWj4j7DB:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ded48bdf0e849fcfe1de22f41516c3eaaaa06a3a/go/sciond/internal/servers/server.go#L118)):*

Needs updating for unixpacket and rename.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447)*

kormat:
:lgtm:

---

Reviewed 1 of 2 files at r7.

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1447#-:-LC-G-uI3DIzPt7PsNIh:bnfp4nl)*

PR #1449, opened 2018-03-12
Make Go SCIOND client use request dispatcher

scrye:
Also:
  - delete some tests that no longer made sense;
  - move `transport.Transport` to `infra` for simpler importing;
  - move `rudp` implementation under `lib/snet` since it depends on it;
  - move channel lock implementation under `lib/util` since it's used in multiple packages now;

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1449)


shitz:
Reviewed 14 of 18 files at r1.
Review status: 11 of 14 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[go/lib/sciond/infra_test.go, line 31 at r1](https://reviewable.io:443/reviews/scionproto/scion/1449#-L7i7BFq0Pg_RqdItbDa:-L7i7BFq0Pg_RqdItbDb:bikg354) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/sciond/infra_test.go#L31)):*
> ```Go
> 
> var (
> 	token = "\x10\x51\x50\x03\x04\x01\x57\x0f" +
> ```

Lets not have these hard-coded bytestrings in test files anymore. For now it's enough to just move it to some `testdata/foo.bin` file and later we can have the test suite automatically generate a new one.

---

*[go/lib/snet/rudp/rudp.go, line 15 at r1](https://reviewable.io:443/reviews/scionproto/scion/1449#-L7i8hdz9tstn8hoJvL2:-L7i8hdz9tstn8hoJvL3:b9xiqvy) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/snet/rudp/rudp.go#L15)):*
> ```Go
> 
> package rudp
> ```

I'm not sure about moving `rudp` to `snet`. `RUDP` is specific to the infra, while `snet` should not depend on anything in `infra`. This introduces a dependency from `snet` to `infra`, which should not be the case (the other way round is fine).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1449)*

shitz:
Reviewed 4 of 18 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1449)*

scrye:
Review status: 13 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/sciond/infra_test.go, line 31 at r1](https://reviewable.io:443/reviews/scionproto/scion/1449#-L7i7BFq0Pg_RqdItbDa:-L8XQDMY_DobY-7ZLy2R:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/sciond/infra_test.go#L31)):*

Lets not have these hard-coded bytestrings in test files anymore. For now it's enough to just move it to some `testdata/foo.bin` file and later we can have the test suite automatically generate a new one.
</blockquote></details>

Done.

---

*[go/lib/snet/rudp/rudp.go, line 15 at r1](https://reviewable.io:443/reviews/scionproto/scion/1449#-L7i8hdz9tstn8hoJvL2:-L8XG5n4oDW7KqR9gozH:br4r0av) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/snet/rudp/rudp.go#L15)):*

I'm not sure about moving `rudp` to `snet`. `RUDP` is specific to the infra, while `snet` should not depend on anything in `infra`. This introduces a dependency from `snet` to `infra`, which should not be the case (the other way round is fine).
</blockquote></details>

I agree re: the inelegant dependencies, but on the flip side it resides near `squic` which is somewhat similar in function.

I don't have strong feelings either way. What is the best place for it, in your opinion?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1449)*

kormat:
Reviewed 16 of 22 files at r1, 4 of 4 files at r2.

---

*[go/lib/sciond/adapter.go, line 34 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L8loKQd9k9yu8xI0Jzx:-L8loKQd9k9yu8xI0Jzy:brg1cq5) ([raw file](https://github.com/scionproto/scion/blob/4c07debfbcc6319a45c0ef3339b2982cb5b2ea42/go/lib/sciond/adapter.go#L34)):*
> ```Go
> }
> 
> func (a *Adapter) RawToMsg(b common.RawBytes) (proto.Cerealizable, error) {
> ```

`return NewPldFromRaw(b)`?

---

*[go/lib/sciond/infra_test.go, line 31 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L7i7BFq0Pg_RqdItbDa:-L8lowCz4DqJq98I8g__:b-y7i70i) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/sciond/infra_test.go#L31)):*

Done.
</blockquote></details>

I'd prefer to use a more meaningful suffix. I used `.bin` for the hpkt and rpkt binary testdata, and have my editor configured to open those in a useful way, so it would be nice to be consistent.

---

*[go/lib/sciond/testdata/ignore.goconvey, line 1 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L8lpSXq5577YmMsaQd_:-L8lpSXr50fggTr_26ji:bvw74i0) ([raw file](https://github.com/scionproto/scion/blob/4c07debfbcc6319a45c0ef3339b2982cb5b2ea42/go/lib/sciond/testdata/ignore.goconvey#L1)):*
> ```
> IGNORE
> ```

This can be dropped now.

---

*[go/lib/snet/rudp/rudp.go, line 15 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L7i8hdz9tstn8hoJvL2:-L8lqvJi12-QoM-LEM70:b-jmgp7n) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/snet/rudp/rudp.go#L15)):*

I agree re: the inelegant dependencies, but on the flip side it resides near `squic` which is somewhat similar in function.

I don't have strong feelings either way. What is the best place for it, in your opinion?
</blockquote></details>

As this does not seem tied to UDP at all, i would suggest renaming it and putting it back in `go/lib/infra/transport`. `ReliablePacketTransport` for example.

---

*[go/lib/snet/rudp/rudp.go, line 110 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L8lq5iZBoZM-Y_Y5AJx:-L8lq5iZBoZM-Y_Y5AJy:bc2bf8s) ([raw file](https://github.com/scionproto/scion/blob/4c07debfbcc6319a45c0ef3339b2982cb5b2ea42/go/lib/snet/rudp/rudp.go#L110)):*
> ```Go
> }
> 
> // New creates a new RUDP connection by wrapping around a PacketConn.
> ```

Ah. As this is wrapping a `PacketConn` and not a `UDPConn`, then `RUDP` is a misnomer, and should be renamed, because it's no longer specific to UDP.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1449)*

scrye:
Review status: 8 of 16 files reviewed at latest revision, 6 unresolved discussions.

---

*[go/lib/sciond/adapter.go, line 34 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L8loKQd9k9yu8xI0Jzx:-L9ABdWAABD54NfrQxw2:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/4c07debfbcc6319a45c0ef3339b2982cb5b2ea42/go/lib/sciond/adapter.go#L34)):*

`return NewPldFromRaw(b)`?
</blockquote></details>

Done.

---

*[go/lib/sciond/infra_test.go, line 31 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L7i7BFq0Pg_RqdItbDa:-L9AGafp5xqiWHznlGOu:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/sciond/infra_test.go#L31)):*

I'd prefer to use a more meaningful suffix. I used `.bin` for the hpkt and rpkt binary testdata, and have my editor configured to open those in a useful way, so it would be nice to be consistent.
</blockquote></details>

Done.

---

*[go/lib/sciond/infra_test.go, line 58 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L8lv6lACsWFnpmdiuK1:-L9AGa5EBySObUqbu4OC:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/4c07debfbcc6319a45c0ef3339b2982cb5b2ea42/go/lib/sciond/infra_test.go#L58)):*

This looks like something that could be added to our test helper collection in `xtest`
</blockquote></details>

Done.

---

*[go/lib/sciond/testdata/ignore.goconvey, line 1 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L8lpSXq5577YmMsaQd_:-L9ADgTU53WSk3XHgQw8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/4c07debfbcc6319a45c0ef3339b2982cb5b2ea42/go/lib/sciond/testdata/ignore.goconvey#L1)):*

This can be dropped now.
</blockquote></details>

Done.

---

*[go/lib/snet/rudp/rudp.go, line 15 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L7i8hdz9tstn8hoJvL2:-L9ADdBY6T8ODY4VCKrW:bd03tax) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/snet/rudp/rudp.go#L15)):*

As this does not seem tied to UDP at all, i would suggest renaming it and putting it back in `go/lib/infra/transport`. `ReliablePacketTransport` for example.
</blockquote></details>

It is tied to `snet` for SCMP stuff (if running on top of `snet`). This means it imports pathmgr and sciond, which in turn means sciond can no longer use transports due to circular imports. At the minimum, this means `ReliablePacketTransport` cannot share the same package with `PacketTransport`.

---

*[go/lib/snet/rudp/rudp.go, line 110 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L8lq5iZBoZM-Y_Y5AJx:-L9AME-xEWRnHJl0W-js:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/4c07debfbcc6319a45c0ef3339b2982cb5b2ea42/go/lib/snet/rudp/rudp.go#L110)):*

Ah. As this is wrapping a `PacketConn` and not a `UDPConn`, then `RUDP` is a misnomer, and should be renamed, because it's no longer specific to UDP.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1449)*

kormat:
Reviewed 9 of 16 files at r3.

---

*[go/lib/snet/rudp/rudp.go, line 15 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L7i8hdz9tstn8hoJvL2:-L9ATK4H7a9rJ7NSlLUC:b-pdq1du) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/snet/rudp/rudp.go#L15)):*

It is tied to `snet` for SCMP stuff (if running on top of `snet`). This means it imports pathmgr and sciond, which in turn means sciond can no longer use transports due to circular imports. At the minimum, this means `ReliablePacketTransport` cannot share the same package with `PacketTransport`.
</blockquote></details>

This implies to me that we need to change our approach for packages that are imported by lots of packages and import lots of packages (e.g. `go/lib/snet` and `go/lib/sciond`). In these cases we should start having  separate package for interfaces and basic types. That will remove most possible import loops, and make things cleaner.

The go stdlib does roughly the same by having "top-level" packages define interfaces, and have the caller import implementations. E.g. https://golang.org/pkg/crypto/. In this model `snet` wouldn't import `pathmgr` or `sciond` implementations. Instead it would import interfaces for them. That would greatly reduce the inter-dependencies between the libraries.

An opposite example:
If you look at `go/lib/common` the only scion package it imports is `go/lib/asserts`, which is a leaf (it imports zero scion packages). In those sorts of packages it's fine to have base types and functionality in the same package, as there's no possibility of a dependency loop.

---

*[go/lib/snet/rpt/rpt_test.go, line 36 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L9AfSLmDby3ooyKHdOx:-L9AfSLmDby3ooyKHdOy:b-rlhw5i) ([raw file](https://github.com/scionproto/scion/blob/62b02a5bcc17ab1fa81b07910888fbdfdac4658b/go/lib/snet/rpt/rpt_test.go#L36)):*
> ```Go
> 	Convey("Create RPT, send unreliable message, and receive same message", t, func() {
> 		conn := loopback.New()
> 		udp := New(conn, log.Root())
> ```

"udp" should be renamed.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1449)*

scrye:

---

*[go/lib/snet/rudp/rudp.go, line 15 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L7i8hdz9tstn8hoJvL2:-L9B5p7CE79L6HobO32f:b-g980by) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/snet/rudp/rudp.go#L15)):*

This implies to me that we need to change our approach for packages that are imported by lots of packages and import lots of packages (e.g. `go/lib/snet` and `go/lib/sciond`). In these cases we should start having  separate package for interfaces and basic types. That will remove most possible import loops, and make things cleaner.

The go stdlib does roughly the same by having "top-level" packages define interfaces, and have the caller import implementations. E.g. https://golang.org/pkg/crypto/. In this model `snet` wouldn't import `pathmgr` or `sciond` implementations. Instead it would import interfaces for them. That would greatly reduce the inter-dependencies between the libraries.

An opposite example:
If you look at `go/lib/common` the only scion package it imports is `go/lib/asserts`, which is a leaf (it imports zero scion packages). In those sorts of packages it's fine to have base types and functionality in the same package, as there's no possibility of a dependency loop.
</blockquote></details>

Since this has evolved to a large discussion about moving stuff around, I think we should merge this as is and move it to `infra/transport` as soon as the main libs are sorted out. Added a FIXME for this.

---

*[go/lib/snet/rpt/rpt_test.go, line 36 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L9AfSLmDby3ooyKHdOx:-L9BBStf-kZ7z4-_7khE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/62b02a5bcc17ab1fa81b07910888fbdfdac4658b/go/lib/snet/rpt/rpt_test.go#L36)):*

"udp" should be renamed.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1449)*

kormat:
Reviewed 1 of 2 files at r4.

---

*[go/lib/snet/rudp/rudp.go, line 15 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L7i8hdz9tstn8hoJvL2:-L9EvzQ0Es-B47ow1hrr:bmcrqw0) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/snet/rudp/rudp.go#L15)):*

Since this has evolved to a large discussion about moving stuff around, I think we should merge this as is and move it to `infra/transport` as soon as the main libs are sorted out. Added a FIXME for this.
</blockquote></details>

Yeah, that sounds reasonable. Just please move the FIXME to `rpt.go`, rather than a test file so that it's more visible.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1449)*

scrye:
Review status: 14 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/snet/rudp/rudp.go, line 15 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1449#-L7i8hdz9tstn8hoJvL2:-L9GImnDE70zwVv5Nqrq:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/44d08df30208bab019fe8fa95620c33674429d95/go/lib/snet/rudp/rudp.go#L15)):*

Yeah, that sounds reasonable. Just please move the FIXME to `rpt.go`, rather than a test file so that it's more visible.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1449)*

kormat:
:lgtm:

---

Reviewed 2 of 4 files at r5.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1449#-:-L9GOAGCC7fPYICX13tB:bnfp4nl)*

shitz:
:lgtm:

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1449#-:-L9GOjYt7kiWEo2xhncN:bnfp4nl)*

PR #1450, opened 2018-03-13
Add timestamp to messages

FR4NK-W:
Fixes ReadDeadline handling: previously a missed deadline invalidated all further measurements, since unconsumed packets from previous deadlines were used to infer the RTT.

Only treat errors as Fatal when necessary. Avoid Fatal errors after the initialization on the server since (a broken connection to a client should not bring down the server)
Close sessions also on the server side.
Sending and receiving messages is no longer lock-stepped.

Thanks for the suggestions in #dev

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1450)

### DIFF
--- a/go/examples/pingpong/pingpong.go
+++ b/go/examples/pingpong/pingpong.go
@@ -31,6 +31,7 @@ import (
 	//"github.com/lucas-clemente/quic-go/qerr"
 
 	"github.com/scionproto/scion/go/lib/addr"
+	"github.com/scionproto/scion/go/lib/common"
 	liblog "github.com/scionproto/scion/go/lib/log"
 	sd "github.com/scionproto/scion/go/lib/sciond"
 	"github.com/scionproto/scion/go/lib/snet"
@@ -42,8 +43,9 @@ const (
 	DefaultInterval = 1 * time.Second
 	DefaultTimeout  = 2 * time.Second
 	MaxPings        = 1 << 16
-	ReqMsg          = "ping!"
+	ReqMsg          = "ping!" // ReqMsg and ReplyMsg length need to be the same
 	ReplyMsg        = "pong!"
+	TSLen           = 8
 )
 
 func GetDefaultSCIONDPath(ia addr.IA) string {
@@ -64,6 +66,7 @@ var (
 	timeout = flag.Duration("timeout", DefaultTimeout,
 		"Timeout for the ping response")
 	interval = flag.Duration("interval", DefaultInterval, "time between pings")
+	verbose  = flag.Bool("v", false, "sets verbose output")
 )
 
 func init() {
@@ -144,8 +147,14 @@ func Client() {
 	}
 	defer qstream.Close()
 	log.Debug("Quic stream opened", "local", &local, "remote", &remote)
+	go Send(qstream)
+	Read(qstream)
+}
 
-	b := make([]byte, 1<<12)
+func Send( /* qstream quic.Stream */ ) {
+	reqMsgLen := len(ReqMsg)
+	payload := make([]byte, reqMsgLen+TSLen)
+	copy(payload[0:], ReqMsg)
 	for i := 0; i < *count || *count == 0; i++ {
 		if i != 0 && *interval != 0 {
 			time.Sleep(*interval)
@@ -153,7 +162,8 @@ func Client() {
 
 		// Send ping message to destination
 		before := time.Now()
-		written, err := qstream.Write([]byte(ReqMsg))
+		common.Order.PutUint64(payload[reqMsgLen:], uint64(before.UnixNano()))
+		written, err := qstream.Write(payload[:])
 		if err != nil {
 			//qer := qerr.ToQuicError(err)
 			//if qer.ErrorCode == qerr.NetworkIdleTimeout {
@@ -163,18 +173,26 @@ func Client() {
 			log.Error("Unable to write", "err", err)
 			continue
 		}
-		if written != len(ReqMsg) {
-			log.Error("Wrote incomplete message", "expected", len(ReqMsg),
+		if written != len(ReqMsg)+TSLen {
+			log.Error("Wrote incomplete message", "expected", len(ReqMsg)+TSLen,
 				"actual", written)
 			continue
 		}
+	}
+	// After sending the last ping, set a ReadDeadline on the stream
+	err := qstream.SetReadDeadline(time.Now().Add(*timeout))
+	if err != nil {
+		LogFatal("SetReadDeadline failed", "err", err)
+	}
+}
 
-		// Receive pong message with timeout
-		err = qstream.SetReadDeadline(time.Now().Add(*timeout))
-		if err != nil {
-			LogFatal("SetReadDeadline failed", "err", err)
-		}
+func Read( /* qstream quic.Stream */ ) {
+	// Receive pong message (with final timeout)
+	b := make([]byte, 1<<12)
+	replyMsgLen := len(ReplyMsg)
+	for i := 0; i < *count || *count == 0; i++ {
 		read, err := qstream.Read(b)
+		after := time.Now()
 		if err != nil {
 			//qer := qerr.ToQuicError(err)
 			//if qer.ErrorCode == qerr.PeerGoingAway {
@@ -183,18 +201,32 @@ func Client() {
 			//}
 			if nerr, ok := err.(net.Error); ok && nerr.Timeout() {
 				log.Debug("ReadDeadline missed", "err", err)
-				continue
+				// ReadDeadline is only set after we are done writing
+				// and we don't want to wait indefinitely for the remaining responses
+				break
 			}
 			log.Error("Unable to read", "err", err)
 			continue
 		}
-		if string(b[:read]) != ReplyMsg {
+		if read < replyMsgLen || string(b[:replyMsgLen]) != ReplyMsg {
 			fmt.Println("Received bad message", "expected", ReplyMsg,
 				"actual", string(b[:read]))
+			continue
 		}
-		after := time.Now()
+		if read < replyMsgLen+TSLen {
+			fmt.Println("Received bad message missing timestamp",
+				"actual", string(b[:read]))
+			continue
+		}
+		before := time.Unix(0, int64(common.Order.Uint64(b[replyMsgLen:replyMsgLen+TSLen])))
 		elapsed := after.Sub(before)
-		fmt.Printf("%d bytes from %v: seq=%d time=%s\n", read, &remote, i, elapsed)
+		if *verbose {
+			fmt.Printf("[%s]\tReceived %d bytes from %v: seq=%d RTT=%s\n",
+				before.Unix(), read, &remote, i, elapsed)
+		} else {
+			fmt.Printf("Received %d bytes from %v: seq=%d RTT=%s\n",
+				read, &remote, i, elapsed)
+		}
 	}
 }
 
@@ -212,7 +244,8 @@ func Server() {
 	for {
 		qsess, err := qsock.Accept()
 		if err != nil {
-			LogFatal("Unable to accept quic session", "err", err)
+			log.Error("Unable to accept quic session", "err", err)
+			continue
 		}
 		log.Debug("Quic session accepted", "src", qsess.RemoteAddr())
 		go handleClient(qsess)
@@ -231,13 +264,17 @@ func initNetwork() {
 	log.Debug("QUIC/SCION successfully initialized")
 }
 
-func handleClient( /*qsess quic.Session*/ ) {
+func handleClient( /* qsess quic.Session */ ) {
+	defer qsess.Close(nil)
 	qstream, err := qsess.AcceptStream()
+	defer qstream.Close()
 	if err != nil {
-		LogFatal("Unable to accept quic stream", "err", err)
+		log.Error("Unable to accept quic stream", "err", err)
+		return
 	}
 
 	b := make([]byte, 1<<12)
+	reqMsgLen := len(ReqMsg)
 	for {
 		// Receive ping message
 		read, err := qstream.Read(b)
@@ -250,17 +287,25 @@ func handleClient( /*qsess quic.Session*/ ) {
 			log.Error("Unable to read", "err", err)
 			break
 		}
-		if string(b[:read]) != ReqMsg {
+		if string(b[:reqMsgLen]) != ReqMsg {
 			fmt.Println("Received bad message", "expected", ReqMsg,
-				"actual", string(b[:read]))
+				"actual", string(b[:reqMsgLen]), "full", string(b[:read]))
 		}
+		// extract timestamp
+		ts := common.Order.PutUint64(b[reqMsgLen:])
 
 		// Send pong message
-		written, err := qstream.Write([]byte(ReplyMsg))
+		replyMsgLen := len(ReplyMsg)
+		copy(b[:replyMsgLen], ReplyMsg)
+		common.Order.PutUint64(b[replyMsgLen:], ts)
+		written, err := qstream.Write(b[:replyMsgLen+TSLen])
 		if err != nil {
-			LogFatal("Unable to write", "err", err)
-		} else if written != len(ReplyMsg) {
-			LogFatal("Wrote incomplete message", "expected", len(ReplyMsg), "actual", written)
+			log.Error("Unable to write", "err", err)
+			continue
+		} else if written != len(ReplyMsg)+TSLen {
+			log.Error("Wrote incomplete message",
+				"expected", len(ReplyMsg)+TSLen, "actual", written)
+			continue
 		}
 	}
 }


FR4NK-W:
@scrye @sgmonroy
Fixes the issue where unrealistic RTTs are computed because after a missed read deadline, in the next loop iteration a packet from the previous exchange gets used to compute the time elapsed since the send.

kormat:
Reviewed 1 of 1 files at r1.

---

*[go/examples/pingpong/pingpong.go, line 45 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7csQ9-SzQjNKDpuyX7:-L7csQ9-SzQjNKDpuyX8:byiicd) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L45)):*
> ```Go
> 	ReqMsg          = "ping!" // ReqMsg and ReplyMsg length need to be the same
> 	ReplyMsg        = "pong!"
> 	TSLen           = 10 // length of unix timestamp == len(strconv.FormatInt(time.Now().Unix(), 10))
> ```

It would be simpler to just encode the timestamp as binary. Marshal:
```
common.Order.PutUint64(b, time.Now().UnixNano())
...
time.Unix(0, common.Order.Uint64(b))
```

---

*[go/examples/pingpong/pingpong.go, line 131 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7csXvPbYzX1mgmtd_5:-L7csXvPbYzX1mgmtd_6:b-rq4irc) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L131)):*
> ```Go
> 	defer qstream.Close()
> 	log.Debug("Quic stream opened", "local", &local, "remote", &remote)
> 	wg := new(sync.WaitGroup)
> ```

I don't think you need a waitgroup for this:
```
go Send(..)
Read()
```
should do the right thing. If the read loop is finished, then there's no point in waiting for the sending loop.

---

*[go/examples/pingpong/pingpong.go, line 189 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7ctI3rQnZFaODF1i3P:-L7ctI3rQnZFaODF1i3Q:b-2en18h) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L189)):*
> ```Go
> 			if nerr, ok := err.(net.Error); ok && nerr.Timeout() {
> 				log.Debug("ReadDeadline missed", "err", err)
> 				// ReadDeadline is only set after we are done writting
> ```

`writing`

---

*[go/examples/pingpong/pingpong.go, line 215 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cuZ4AHoaQ5j3I8clX:-L7cuZ4AHoaQ5j3I8clY:b-7lj47f) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L215)):*
> ```Go
> 		before := time.Unix(unixTs, 0)
> 		elapsed := after.Sub(before)
> 		fmt.Printf("[%v] Received %d bytes from %v: seq=%d RTT=%s\n", before.Unix(), read, &remote, i, elapsed)
> ```

I don't think it's useful to print the unix timestamp. The RTT is the relevant part.

I would also round the elapsed time: `elapsed.Round(time.Microsecond)`, so that it's less noisy, while still sufficiently precise.

---

*[go/examples/pingpong/pingpong.go, line 234 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cvA1B4yfVc2te7mMa:-L7cvA1B4yfVc2te7mMb:bivf07a) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L234)):*
> ```Go
> 		if err != nil {
> 			log.Error("Unable to accept quic session", "err", err)
> 			qsess.Close(nil)
> ```

Are you sure that `qsess` is non-nil if `Accept()` returns an error? That would not be what i'd expect.

---

*[go/examples/pingpong/pingpong.go, line 254 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cvIv7kxWEjz6ISzli:-L7cvIv8zNIIUFtx1XzW:b-5pk5qt) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L254)):*
> ```Go
> }
> 
> func handleClient(qsess quic.Session) {
> ```

This breaks our lint process if you happen to have `quic-go` in your GOPATH. That's why all the other `quic` references are commented out.

---

*[go/examples/pingpong/pingpong.go, line 285 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cw5vjuxssIuFnipwg:-L7cw5vjuxssIuFnipwh:bkbgfuw) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L285)):*
That doesn't sound right. https://golang.org/pkg/builtin/#copy:

> Copy returns the number of elements copied, which will be the minimum of len(src) and len(dst).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1450)*

FR4NK-W:

---

*[go/examples/pingpong/pingpong.go, line 215 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cuZ4AHoaQ5j3I8clX:-L7cxyAl8nFekh7xtLam:byt0mof) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L215)):*

I don't think it's useful to print the unix timestamp. The RTT is the relevant part.

I would also round the elapsed time: `elapsed.Round(time.Microsecond)`, so that it's less noisy, while still sufficiently precise.
</blockquote></details>

I added timestamp to the output because the remote end goes down sometimes during longer runs. So having a timestamp helps correlating events.

I also have a version with an additional format flag, where the user can pass the format string (would also give control over truncating the RTT) as a flag at runtime (with a default one similar to the current one). 
Would that be a better solution? Or just log the more detailed output and only print the uncluttered view?

---

*[go/examples/pingpong/pingpong.go, line 234 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cvA1B4yfVc2te7mMa:-L7d2WYUp_K_EmMorlDO:b-h32ab0) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L234)):*

Are you sure that `qsess` is non-nil if `Accept()` returns an error? That would not be what i'd expect.
</blockquote></details>

No you are right, underneath it calls net.Listener.Accept() which returns a nil qsess when err is non-nil

---

*[go/examples/pingpong/pingpong.go, line 285 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cw5vjuxssIuFnipwg:-L7d-JddS5JvhOdBcpQh:boe3lc7) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L285)):*

That doesn't sound right. https://golang.org/pkg/builtin/#copy:

> Copy returns the number of elements copied, which will be the minimum of len(src) and len(dst).
</blockquote></details>

This I agree with. And here len(src)==len(dst) by definition (replyMsgLen:=len(ReplyMsg)), the thing is that len(ReqMsg) might be smaller than len(ReplyMsg).
So is your suggestion to use `copy(b[:len(ReqMsg)], ReplyMsg)` and possibly have ReplyMsg truncated?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1450)*

FR4NK-W:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[go/examples/pingpong/pingpong.go, line 131 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7csXvPbYzX1mgmtd_5:-L7dCg6KkV7Zxfl_iVPk:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L131)):*

I don't think you need a waitgroup for this:
```
go Send(..)
Read()
```
should do the right thing. If the read loop is finished, then there's no point in waiting for the sending loop.
</blockquote></details>

Done.

---

*[go/examples/pingpong/pingpong.go, line 189 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7ctI3rQnZFaODF1i3P:-L7dCgspi8iS6PkEMusb:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L189)):*

`writing`
</blockquote></details>

Done.

---

*[go/examples/pingpong/pingpong.go, line 234 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cvA1B4yfVc2te7mMa:-L7dCiT_eLeT0o_5bn47:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L234)):*

No you are right, underneath it calls net.Listener.Accept() which returns a nil qsess when err is non-nil
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1450)*

kormat:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[go/examples/pingpong/pingpong.go, line 45 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7csQ9-SzQjNKDpuyX7:-L7dIiuKGNVhFf9kpjnB:b-82oqa1) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L45)):*

It would be simpler to just encode the timestamp as binary. Marshal:
```
common.Order.PutUint64(b, time.Now().UnixNano())
...
time.Unix(0, common.Order.Uint64(b))
```
</blockquote></details>

Ping.

---

*[go/examples/pingpong/pingpong.go, line 215 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cuZ4AHoaQ5j3I8clX:-L7dIVmH8PSkuWn1C8wh:b-lfu7y8) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L215)):*

I added timestamp to the output because the remote end goes down sometimes during longer runs. So having a timestamp helps correlating events.

I also have a version with an additional format flag, where the user can pass the format string (would also give control over truncating the RTT) as a flag at runtime (with a default one similar to the current one). 
Would that be a better solution? Or just log the more detailed output and only print the uncluttered view?

</blockquote></details>

I'd suggest in this case a `-v` (verbose) flag that, if included, prints the full timestamp at the end of the line.

---

*[go/examples/pingpong/pingpong.go, line 254 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cvIv7kxWEjz6ISzli:-L7dIwA7CFvGOMdba_GH:b-82oqa1) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L254)):*

This breaks our lint process if you happen to have `quic-go` in your GOPATH. That's why all the other `quic` references are commented out.
</blockquote></details>

Ping.

---

*[go/examples/pingpong/pingpong.go, line 285 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cw5vjuxssIuFnipwg:-L7dJ0aVeOd0CjoNkqJ9:b-7oq9kv) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L285)):*

This I agree with. And here len(src)==len(dst) by definition (replyMsgLen:=len(ReplyMsg)), the thing is that len(ReqMsg) might be smaller than len(ReplyMsg).
So is your suggestion to use `copy(b[:len(ReqMsg)], ReplyMsg)` and possibly have ReplyMsg truncated?
</blockquote></details>

The simplest thing is to extract the timestamp, and then write a new payload including the sender's timestamp. That way the relative lengths of `ReqMsg` and `ReplyMsg` aren't relevant.

---

*[go/examples/pingpong/pingpong.go, line 134 at r2](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7dIqVilBWf2NJL8YkX:-L7dIqVilBWf2NJL8YkY:b-nbtfa4) ([raw file](https://github.com/scionproto/scion/blob/131e03d7f3563e931c9bab9c87f2149c9cd101aa/go/examples/pingpong/pingpong.go#L134)):*
> ```Go
> }
> 
> func Send(qstream quic.Stream) {
> ```

Breaks lint.

---

*[go/examples/pingpong/pingpong.go, line 169 at r2](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7dIs_-ie9PhuhbIQXh:-L7dIs_-ie9PhuhbIQXi:b-nbtfa4) ([raw file](https://github.com/scionproto/scion/blob/131e03d7f3563e931c9bab9c87f2149c9cd101aa/go/examples/pingpong/pingpong.go#L169)):*
> ```Go
> }
> 
> func Read(qstream quic.Stream) {
> ```

Breaks lint.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1450)*

FR4NK-W:
Reviewed 1 of 1 files at r3.

---

*[go/examples/pingpong/pingpong.go, line 45 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7csQ9-SzQjNKDpuyX7:-L7drGjqMOaboEzOsMs2:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L45)):*

Ping.
</blockquote></details>

Done.

---

*[go/examples/pingpong/pingpong.go, line 215 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cuZ4AHoaQ5j3I8clX:-L7dwMNF1nSoAdQKVc-8:bk72ui1) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L215)):*

I'd suggest in this case a `-v` (verbose) flag that, if included, prints the full timestamp at the end of the line.
</blockquote></details>

Added the flag, having the RTT at the end of the line gives a quick grasp of the RTT variance

---

*[go/examples/pingpong/pingpong.go, line 254 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cvIv7kxWEjz6ISzli:-L7e4iewPxhXvkpT9YKA:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L254)):*

Ping.
</blockquote></details>

Done.

---

*[go/examples/pingpong/pingpong.go, line 285 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cw5vjuxssIuFnipwg:-L7dqYZwiQdNoBLTKpz5:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L285)):*

The simplest thing is to extract the timestamp, and then write a new payload including the sender's timestamp. That way the relative lengths of `ReqMsg` and `ReplyMsg` aren't relevant.
</blockquote></details>

Done.

---

*[go/examples/pingpong/pingpong.go, line 134 at r2](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7dIqVilBWf2NJL8YkX:-L7e4n3a-y5gaA0kyUj1:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/131e03d7f3563e931c9bab9c87f2149c9cd101aa/go/examples/pingpong/pingpong.go#L134)):*

Breaks lint.
</blockquote></details>

Done.

---

*[go/examples/pingpong/pingpong.go, line 169 at r2](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7dIs_-ie9PhuhbIQXh:-L7e4nhiU0lHFHFWHiRk:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/131e03d7f3563e931c9bab9c87f2149c9cd101aa/go/examples/pingpong/pingpong.go#L169)):*

Breaks lint.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1450)*

kormat:
Reviewed 1 of 1 files at r3.

---

*[go/examples/pingpong/pingpong.go, line 215 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cuZ4AHoaQ5j3I8clX:-L7e6IzuhA4HOuRCp94U:bju1tr) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L215)):*

Added the flag, having the RTT at the end of the line gives a quick grasp of the RTT variance
</blockquote></details>

Alrighty. One minor thing - use `%s` for the timestamp, as it implements Stringer.

---

*[go/examples/pingpong/pingpong.go, line 275 at r3](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7e6X7QOc_twb7J70LW:-L7e6X7QOc_twb7J70LX:b91nbat) ([raw file](https://github.com/scionproto/scion/blob/25247547b785819f8cd56b3404ac64d3371c926e/go/examples/pingpong/pingpong.go#L275)):*
> ```Go
> 		// extract timestamp
> 		ts := make([]byte, 8)
> 		copy(ts, b[reqMsgLen:reqMsgLen+TSLen])
> ```

This is slightly more work than just:
```
ts := common.Order.Uint64(b[reqMsgLen:])
...
common.Order.PutUint64(b[replyMsgLen:])
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1450)*

FR4NK-W:
Reviewed 1 of 1 files at r4.

---

*[go/examples/pingpong/pingpong.go, line 215 at r1](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7cuZ4AHoaQ5j3I8clX:-L7e6yk76tPobw-lEUT4:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bb88f188e76d37b32b0e4a195f3d47e06daac8c2/go/examples/pingpong/pingpong.go#L215)):*

Alrighty. One minor thing - use `%s` for the timestamp, as it implements Stringer.
</blockquote></details>

Done.

---

*[go/examples/pingpong/pingpong.go, line 275 at r3](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7e6X7QOc_twb7J70LW:-L7e8MjsYeROBfet3zZw:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/25247547b785819f8cd56b3404ac64d3371c926e/go/examples/pingpong/pingpong.go#L275)):*

This is slightly more work than just:
```
ts := common.Order.Uint64(b[reqMsgLen:])
...
common.Order.PutUint64(b[replyMsgLen:])
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1450)*

kormat:
Reviewed 1 of 1 files at r4.

---

*[go/examples/pingpong/pingpong.go, line 205 at r4](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7eGQs4loVlsQpPMB-s:-L7eGQs4loVlsQpPMB-t:b-f4m0io) ([raw file](https://github.com/scionproto/scion/blob/c96c33bf966d1216b1d664633f9f08ce1c977bdd/go/examples/pingpong/pingpong.go#L205)):*
> ```Go
> 		elapsed := after.Sub(before)
> 		if *verbose {
> 			fmt.Printf("[%s]\tReceived %d bytes from %v: seq=%d RTT=%s\n", before.Unix(), read, &remote, i, elapsed)
> ```

Line length (code should be wrapped before reaching 100 lines, assuming tabs are 4 spaces).

---

*[go/examples/pingpong/pingpong.go, line 285 at r4](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7eGbGpX8QFN0NqFAti:-L7eGbGpX8QFN0NqFAtj:bftyao1) ([raw file](https://github.com/scionproto/scion/blob/c96c33bf966d1216b1d664633f9f08ce1c977bdd/go/examples/pingpong/pingpong.go#L285)):*
> ```Go
> 			continue
> 		} else if written != len(ReplyMsg)+TSLen {
> 			log.Error("Wrote incomplete message", "expected", len(ReplyMsg)+TSLen, "actual", written)
> ```

Line length.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1450)*

FR4NK-W:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[go/examples/pingpong/pingpong.go, line 205 at r4](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7eGQs4loVlsQpPMB-s:-L7eH1qqjGuzjKLU72QL:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c96c33bf966d1216b1d664633f9f08ce1c977bdd/go/examples/pingpong/pingpong.go#L205)):*

Line length (code should be wrapped before reaching 100 lines, assuming tabs are 4 spaces).
</blockquote></details>

Done.

---

*[go/examples/pingpong/pingpong.go, line 285 at r4](https://reviewable.io:443/reviews/scionproto/scion/1450#-L7eGbGpX8QFN0NqFAti:-L7eHVZV5-u29v8-s-4-:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c96c33bf966d1216b1d664633f9f08ce1c977bdd/go/examples/pingpong/pingpong.go#L285)):*

Line length.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1450)*

kormat:
Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1450)*

PR #1452, opened 2018-03-14
Implement Deduper

scrye:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1452)


kormat:
Reviewed 3 of 4 files at r1.
Review status: 3 of 4 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/dedupe/chanset.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_2TETX9NA79yrajVv:-L7_2TETX9NA79yrajVw:b-fnfgy) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L1)):*
> ```Go
> ```

This isn't a good name for a file which is almost entirely made up of `notificationTable`.

---

*[go/lib/infra/dedupe/chanset.go, line 26 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_81w1q3oWPlhHhQgL:-L7_81w1q3oWPlhHhQgM:b-bnuxpn) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L26)):*
> ```Go
> // map to themselves. Each channel corresponds to one client call to
> // Deduper.Request.
> type notifyList map[ResponseChannel]ResponseChannel
> ```

Why map to itself? I tested changing this to `map[ResponseChannel]struct{}`, and it compiled, and the tests passed. I also couldn't find a place where the value was used.

---

*[go/lib/infra/dedupe/chanset.go, line 30 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_3V0L7hNMzoNBtloY:-L7_3V0L7hNMzoNBtloZ:b-ue1i62) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L30)):*
> ```Go
> // cacheEntry describes a value obtained via a successful network request.
> //
> // FIXME(scrye): Currently cacheEntries leak memory; they expire, but are not
> ```

What about just using `github.com/patrickmn/go-cache` as we do elsewhere? 

---

*[go/lib/infra/dedupe/chanset.go, line 80 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_5ehirfTSfVNF-CLw:-L7_5ehirfTSfVNF-CLx:b-q7o0xw) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L80)):*
> ```Go
> // Add registers ch with the dedupe and broadcast key maps. If a network
> // request needs to be issued, the returneed context is non-nil.
> func (table *notificationTable) Add(object Request, ch ResponseChannel) context.Context {
> ```

`object` is not a great name. It's also used for the _response_. I would suggest `req`.

---

*[go/lib/infra/dedupe/chanset.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_5nApLMNxXMJr2kDF:-L7_5nApLMNxXMJr2kDG:b-lr15hs) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L86)):*
> ```Go
> 	// If the answer is cached and in the grace period, do not bother sending
> 	// out a request and answer immediately.
> 	if entry, ok := table.cache[object.BroadcastKey()]; ok {
> ```

There are 3 separate calls to `BroadcastKey()` in this method. It's possible that might be expensive to compute. As it's always used at least once, you might as well just calculate it at the top.

---

*[go/lib/infra/dedupe/chanset.go, line 147 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_9RoB6inuy7AE8C0_:-L7_9RoB6inuy7AE8C0a:b9wtg9w) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L147)):*
> ```Go
> 
> // removeLocked is the acquired-lock variant of Remove. It should only be
> // called after writing the response to the to-be removed channel ch.
> ```

`to-be-removed` would be easier to parse.

---

*[go/lib/infra/dedupe/chanset.go, line 151 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_DOU3r04Pk1W0XjNi:-L7_DOU3r04Pk1W0XjNj:bhpr2ta) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L151)):*
> ```Go
> 	dedupeKey := object.DedupeKey()
> 	broadcastKey := object.BroadcastKey()
> 	delete(table.broadcast[broadcastKey], ch)
> ```

Shouldn't `table.broadcast[bkey]` be cleaned up, too, if it's also empty?

---

*[go/lib/infra/dedupe/chanset.go, line 154 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_E5i3wDqEnUT_5xda:-L7_E5i3wDqEnUT_5xdb:b-h5xf7s) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L154)):*
> ```Go
> 	delete(table.dedupe[dedupeKey], ch)
> 	// If there are no more channels on the dedupeKey, it means no goroutine is
> 	// still waiting for the request and we can cancel it.
> ```

I don't follow. `f` is only useful for stopping an existing request goroutine as far as i can tell, but this says that such a goroutine doesn't exist, in which case what does calling `f` do?

---

*[go/lib/infra/dedupe/chanset.go, line 204 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_L1id-UphKPtGi-Pz:-L7_L1id-UphKPtGi-Q-:b-6tai9t) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L204)):*
> ```Go
> 			delete(table.dedupe, dedupeKey)
> 			delete(table.goroutines, ch)
> 			// We just nuked all the notification list for this dedupeKey, so
> ```

This looks racey. If an outstanding request gets cancelled, it will call `BroadcastError`, which will then panic, because the entry for its dedupe key has been removed.

---

*[go/lib/infra/dedupe/dedupe.go, line 26 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_-Qr9gUgksy4iyIZ2:-L7_-QrAKDo7qE3wbuMO:b-wvlh5i) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L26)):*
> ```Go
> //   dd.Request(ctx.TODO(), objectB)
> //
> // If objectA and objectB have the sam DedupeKey() (and arrive at dd at
> ```

"sam" indeed :)

---

*[go/lib/infra/dedupe/dedupe.go, line 48 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_6v1EZJVrOJDydVqY:-L7_6v1EZJVrOJDydVqZ:b-jjv32t) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L48)):*
> ```Go
> // CancelFunc can be called to cancel a request ahead of time, freeing up
> // internal resources (notification lists, goroutines, etc.).
> type CancelFunc func()
> ```

Why a custom type? It looks like this is just `context.CancelFunc` in practice.

---

*[go/lib/infra/dedupe/dedupe.go, line 115 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_1PPJh_xdw58L8ERN:-L7_1PPJh_xdw58L8ERO:b-fxvvae) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L115)):*
> ```Go
> func (dd *Deduper) Request(ctx context.Context, object Request) (<-chan Response, CancelFunc) {
> 	dd.mu.Lock()
> 	if dd.notifications == nil {
> ```

This looks like it belongs in a constructor, rather than in the main method for the type. That also removes the need to have a mutex at all. It _also_ removes the need to have unnecessary `DedupeLifetime` and `Graceperiod` attributes on the `Deduper` type - they are only used during construction, and changing them later has no effect.

---

*[go/lib/infra/dedupe/dedupe.go, line 150 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_AsPrj1GCj70aO33J:-L7_AsPrj1GCj70aO33K:bnly2vn) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L150)):*
> ```Go
> func (dd *Deduper) handler(ctx context.Context, object Request) {
> 	ch := make(chan Response, 1)
> 	dd.RequestFunc(ctx, object, ch)
> ```

The docs don't say that `RequestFunc` must return immediately. If it supports `context.Context`, then it can handle the timeout behaviour and it doesn't need to be handled here as well. If it doesn't support context, then it could block indefinitely, and the `select` statement is ~useless.

If we can specifiy the RequestFunc must supports contexts, then the logic here becomes much simpler. You wouldn't even need to use a channel.

```
type RequestFunc func(ctx context.Context, request Request) Response

func .. handler(..) {
rep := dd.RequestFunc(ctx, object)
if rep.err != nil {
  BroadcastError
} else {
  BroadcastSuccess
  Cache
}
```

---

*[go/lib/infra/dedupe/dedupe.go, line 163 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_BQMFHCFaWnEiEd7e:-L7_BQMFHCFaWnEiEd7f:bdms2gq) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L163)):*
> ```Go
> 		dd.notifications.BroadcastError(object, response)
> 	case response := <-ch:
> 		// Peek for context expiration
> ```

I don't follow the logic here. If the context times out at ~exactly the same time as a response is received, why not just use the response?

---

*[go/lib/infra/dedupe/example_test.go, line 15 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_LEMELCu35Dho-2TD:-L7_LEMELCu35Dho-2TE:b-t71mbe) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/example_test.go#L15)):*
> ```Go
> 
> package dedupe
> ```

Blank example?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

kormat:
Review status: 3 of 4 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/dedupe/chanset.go, line 204 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_L1id-UphKPtGi-Pz:-L7cs5vppYwdOYdQca9T:bo5uned) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L204)):*

This looks racey. If an outstanding request gets cancelled, it will call `BroadcastError`, which will then panic, because the entry for its dedupe key has been removed.
</blockquote></details>

Aaand disregard this. I keep forgetting that looking up a non-existent key in a go map doesn't cause a panic, it just returns the zero-value, which in this case would be an empty list (in the form of a typed `nil`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

scrye:
Review status: 3 of 4 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/dedupe/chanset.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_2TETX9NA79yrajVv:-L7_MOJBLrXsrfl2WIPm:b-sc65an) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L1)):*

This isn't a good name for a file which is almost entirely made up of `notificationTable`.
</blockquote></details>

It was the pre-grand-renaming name, when the world was young and stuff had other names. Changed to `notification.go`.

---

*[go/lib/infra/dedupe/chanset.go, line 26 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_81w1q3oWPlhHhQgL:-L7_MgOVzzNWgKBf3XHZ:be5pnua) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L26)):*

Why map to itself? I tested changing this to `map[ResponseChannel]struct{}`, and it compiled, and the tests passed. I also couldn't find a place where the value was used.
</blockquote></details>

It was used in some older one-liners, but it's a good idea to change it.

---

*[go/lib/infra/dedupe/chanset.go, line 30 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_3V0L7hNMzoNBtloY:-L7_N77BnPmu2dI_J5yB:b-akx5dv) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L30)):*

What about just using `github.com/patrickmn/go-cache` as we do elsewhere? 
</blockquote></details>

I wanted to avoid any long-running goroutines for this package, and `go-cache` spawns a periodic janitor. I guess we could just use it anyway?

---

*[go/lib/infra/dedupe/chanset.go, line 80 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_5ehirfTSfVNF-CLw:-L7_OR2MB54s89pIQUGs:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L80)):*

`object` is not a great name. It's also used for the _response_. I would suggest `req`.
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/chanset.go, line 86 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_5nApLMNxXMJr2kDF:-L7_OfsMn1avG_-mhXnO:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L86)):*

There are 3 separate calls to `BroadcastKey()` in this method. It's possible that might be expensive to compute. As it's always used at least once, you might as well just calculate it at the top.
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/chanset.go, line 147 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_9RoB6inuy7AE8C0_:-L7_Oj-Tb3xApCv_p-27:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L147)):*

`to-be-removed` would be easier to parse.
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/chanset.go, line 151 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_DOU3r04Pk1W0XjNi:-L7_OxE0DEjpHSED9ema:bfbq99i) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L151)):*

Shouldn't `table.broadcast[bkey]` be cleaned up, too, if it's also empty?
</blockquote></details>

Good point, done.

---

*[go/lib/infra/dedupe/chanset.go, line 154 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_E5i3wDqEnUT_5xda:-L7_PSDYEK3MNPUotJBI:bs1z9mk) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L154)):*

I don't follow. `f` is only useful for stopping an existing request goroutine as far as i can tell, but this says that such a goroutine doesn't exist, in which case what does calling `f` do?
</blockquote></details>

Ah, the naming is ambiguous. There a difference between client goroutine (the app using this package) and the network goroutine (the function that is actually running the network exchanges). The comment says "no more client goroutines are running, so just cancel the network goroutine". Reworded the comment to hopefully make this clearer.

---

*[go/lib/infra/dedupe/chanset.go, line 204 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_L1id-UphKPtGi-Pz:-L7_Q-QwlwrTSyNLq22d:b-y2fg86) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/chanset.go#L204)):*

This looks racey. If an outstanding request gets cancelled, it will call `BroadcastError`, which will then panic, because the entry for its dedupe key has been removed.
</blockquote></details>

It is intended behavior (and actually tested in test number 3 - `multiple requests with same broadcast key, data comes from first`). Goroutines `a-x` and `a-y` get cancelled, they enter `BroadcastError`, range on an empty map (the default map value when a value is used as a map in another map) and exit cleanly.

---

*[go/lib/infra/dedupe/dedupe.go, line 26 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_-Qr9gUgksy4iyIZ2:-L7_R_Jlu2RO3S3JEudb:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L26)):*

"sam" indeed :)
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/dedupe.go, line 48 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_6v1EZJVrOJDydVqY:-L7_R_rm8iHewCT5gpqL:b-gx77mq) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L48)):*

Why a custom type? It looks like this is just `context.CancelFunc` in practice.
</blockquote></details>

Semantics, mostly. It's not always a context `CancelFunc` (e.g., `Deduper.Request` returns a cancellation function that does something else). I didn't want to associate it in any way with `context`'s cancellation functions.

---

*[go/lib/infra/dedupe/dedupe.go, line 115 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_1PPJh_xdw58L8ERN:-L7_Rs_caZWyTmwrl24K:b-khkftx) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L115)):*

This looks like it belongs in a constructor, rather than in the main method for the type. That also removes the need to have a mutex at all. It _also_ removes the need to have unnecessary `DedupeLifetime` and `Graceperiod` attributes on the `Deduper` type - they are only used during construction, and changing them later has no effect.
</blockquote></details>

I'm trying to move away from constructors and towards "zero value is valid", as it allows for flexible behavior of objects without tying it to a constructor that needs to be refactored. In my opinion it leads to cleaner code.

I've added a comment on the `type` to specify that the fields should only be set during initialization and never changed afterwards.

---

*[go/lib/infra/dedupe/dedupe.go, line 150 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_AsPrj1GCj70aO33J:-L7czccSi70IvspR2k1t:bgoy2an) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L150)):*

The docs don't say that `RequestFunc` must return immediately. If it supports `context.Context`, then it can handle the timeout behaviour and it doesn't need to be handled here as well. If it doesn't support context, then it could block indefinitely, and the `select` statement is ~useless.

If we can specifiy the RequestFunc must supports contexts, then the logic here becomes much simpler. You wouldn't even need to use a channel.

```
type RequestFunc func(ctx context.Context, request Request) Response

func .. handler(..) {
rep := dd.RequestFunc(ctx, object)
if rep.err != nil {
  BroadcastError
} else {
  BroadcastSuccess
  Cache
}
```
</blockquote></details>

`RequestFunc` doesn't need to return immediately, and it usually can't as internally it probably blocks on some network messaging. It is supposed to take contexts into account, so it blocking forever is a programming error and will always leak a goroutine (the one running the RequestFunc itself). I've added a comment to `RequestFunc` to stress this aspect.

I'm not 100% sure, but your proposal looks equivalent to the current one. It might also return a valid answer/success if `RequestFunc` overextends past its context to return success.

Now, one improvement we can make here is to make the `Deduper` not care at all about these goroutine leaks (and be immune to context misuse in RequestFunc). The solution is to have the `Deduper` select on the context and the channel _while_ the `RequestFunc` is running. This allows `Deduper` to correctly enforce expirations without ever caring if `RequestFunc`takes `ctx` into account correctly. The leak is also completely ignored during future use of the Deduper.

---

*[go/lib/infra/dedupe/dedupe.go, line 163 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_BQMFHCFaWnEiEd7e:-L7d3URCXq_NaMql1Je9:bo6j7c9) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L163)):*

I don't follow the logic here. If the context times out at ~exactly the same time as a response is received, why not just use the response?
</blockquote></details>

The response can contain an error, one being "my context expired". That needs to translate to `BroadcastError`. I've refactored the bit to hopefully make its behavior clearer.

---

*[go/lib/infra/dedupe/example_test.go, line 15 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_LEMELCu35Dho-2TD:-L7d5hx3Ygqk7EkUW9wb:b2aahuo) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/example_test.go#L15)):*

Blank example?
</blockquote></details>

Deprecated, since now the test and docs show the package is used. Removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

scrye:
*[go/lib/infra/dedupe/notification.go, line 204 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_L1id-UphKPtGi-Pz:-L7dOz-M5NJotBsRbtsz:bmd5b5h) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/notification.go#L204)):*

It is intended behavior (and actually tested in test number 3 - `multiple requests with same broadcast key, data comes from first`). Goroutines `a-x` and `a-y` get cancelled, they enter `BroadcastError`, range on an empty map (the default map value when a value is used as a map in another map) and exit cleanly.
</blockquote></details>

Ah, didn't notice the new comment before posting mine. Ack.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

kormat:
Reviewed 4 of 5 files at r2.
Review status: 3 of 4 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/infra/dedupe/dedupe.go, line 115 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_1PPJh_xdw58L8ERN:-L7dM5IyhiwLfPq0gTeR:b-350agx) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L115)):*

I'm trying to move away from constructors and towards "zero value is valid", as it allows for flexible behavior of objects without tying it to a constructor that needs to be refactored. In my opinion it leads to cleaner code.

I've added a comment on the `type` to specify that the fields should only be set during initialization and never changed afterwards.
</blockquote></details>

It's quite a limited approach, however. It only really works when you create an object and use it ~immediately with a single call, e.g. `http.ListenAndServe()`. Otherwise if you have invalid configuration etc you don't find out until potentially Much later when you try to _use_ it. Plus it adds extra complication to the type that isn't needed (exported struct members that are only used once on a persistent object, a mutex that's only used during init), To avoid refactoring of constructors, it's pretty simple to have a config type. E.g.:

```
type DeDupeOpts struct {
  RequestFunc RequestFunc
  DedupeLifetime time.Duration
  GracePeriod time.Duration
}

type DeDuper {
   notifications *notificationTable
}

// specifically passing opts by value rather than reference, so there's no confusion about the intended lifetime of the opts object.
func NewDeduper(opts DeDupeOpts) (*DeDuper, error) {
  dd := DeDuper{RequestFunc: opts.RequestFunc}
  ...
  dd.notifications = &notificationTable{
     DedupeLifetime: opts.DeDupeLifetime,
     GracePeriod: opts:Graceperiod
     ...
}

// usage:
dd := NewDeDuper(DeDupeOpts{RequestFunc: f, DeDupeLifetime: 3 * time.Second})
```

The constructor doesn't need refactoring, it's simple and pretty intuitive to use, allows errors to be caught at construction rather than at use sometime later. It's also applicable in pretty much every situation (e.g. you need to setup a connection on startup, for later use).

---

*[go/lib/infra/dedupe/dedupe.go, line 150 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_AsPrj1GCj70aO33J:-L7dPmoJFY9dQvUgWmoA:b-ykqsrz) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L150)):*

`RequestFunc` doesn't need to return immediately, and it usually can't as internally it probably blocks on some network messaging. It is supposed to take contexts into account, so it blocking forever is a programming error and will always leak a goroutine (the one running the RequestFunc itself). I've added a comment to `RequestFunc` to stress this aspect.

I'm not 100% sure, but your proposal looks equivalent to the current one. It might also return a valid answer/success if `RequestFunc` overextends past its context to return success.

Now, one improvement we can make here is to make the `Deduper` not care at all about these goroutine leaks (and be immune to context misuse in RequestFunc). The solution is to have the `Deduper` select on the context and the channel _while_ the `RequestFunc` is running. This allows `Deduper` to correctly enforce expirations without ever caring if `RequestFunc`takes `ctx` into account correctly. The leak is also completely ignored during future use of the Deduper.
</blockquote></details>

Given that you've now specified that `RequestFunc` must handle context correctly, then you don't need to run it in another goroutine, or even check for `ctx.Done()`. It just simplifies the logic.

---

*[go/lib/infra/dedupe/dedupe.go, line 162 at r2](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7dQ0WNSNK7Dw8BTtrT:-L7dQ0WNSNK7Dw8BTtrU:b-6r176p) ([raw file](https://github.com/scionproto/scion/blob/43aa483463161117c066be57078d2ae930fd8082/go/lib/infra/dedupe/dedupe.go#L162)):*
> ```Go
> 	// object. During errors, both the context is marked as Done (via external
> 	// cancellation or expiry) and ch contains an error. Because the top-level
> 	// cases in the select below are nondeterministic, we also check for context
> ```

That is not correct. You check for _errors_, of which context expiration is one possible error.

---

*[go/lib/infra/dedupe/notification.go, line 30 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_3V0L7hNMzoNBtloY:-L7dJhSQQdUWektMSPOA:bbov8ln) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/notification.go#L30)):*

I wanted to avoid any long-running goroutines for this package, and `go-cache` spawns a periodic janitor. I guess we could just use it anyway?
</blockquote></details>

I think the choice is between a single long-running maintenance thread, vs potentially many short-lived cleanup threads. I think the former is easier to work with. If you're looking at the list of goroutines, it's a lot less noise to deal with.

---

*[go/lib/infra/dedupe/notification.go, line 80 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_5ehirfTSfVNF-CLw:-L7dQbygT5qGWBMo78Yb:b6dr5zc) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/notification.go#L80)):*

Done.
</blockquote></details>

Same applies to `Deduper.Request()` and `Deduper.handler()`.

---

*[go/lib/infra/dedupe/notification.go, line 122 at r2](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7dQsPTMTFlTBxcbvuZ:-L7dQsPTMTFlTBxcbvu_:b-gf5tcx) ([raw file](https://github.com/scionproto/scion/blob/43aa483463161117c066be57078d2ae930fd8082/go/lib/infra/dedupe/notification.go#L122)):*
> ```Go
> }
> 
> // Cache saves object in the cache, using the specified key. The current time
> ```

s/object/response/

---

*[go/lib/infra/dedupe/notification.go, line 169 at r2](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7dRVID0pEUeeH4Fy9V:-L7dRVID0pEUeeH4Fy9W:b-adut7k) ([raw file](https://github.com/scionproto/scion/blob/43aa483463161117c066be57078d2ae930fd8082/go/lib/infra/dedupe/notification.go#L169)):*
> ```Go
> 
> // BroadcastError writes response to all the channels waiting on
> // object's dedupe key. Response should contain an error and nil data.
> ```

s/object/req/

---

*[go/lib/infra/dedupe/notification.go, line 188 at r2](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7dRanZ-YbO3edk9Ijr:-L7dRanZ-YbO3edk9Ijs:b-eo25iu) ([raw file](https://github.com/scionproto/scion/blob/43aa483463161117c066be57078d2ae930fd8082/go/lib/infra/dedupe/notification.go#L188)):*
> ```Go
> }
> 
> // BroadcastSuccess writes response to all the channels waiting on object's
> ```

"on a request's broadcast key"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

scrye:
Review status: 1 of 4 files reviewed at latest revision, 8 unresolved discussions.

---

*[go/lib/infra/dedupe/dedupe.go, line 115 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_1PPJh_xdw58L8ERN:-L7dndBtUH_43gxiXPnD:btfyhfx) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L115)):*
> It only really works when you create an object and use it ~immediately with a single call, e.g. `http.ListenAndServe()`

There's no reason why that has to be the case, proof in case, this package where it works with multiple calls to `Request`.

>  Otherwise if you have invalid configuration etc you don't find out until potentially Much later when you try to use it. 

The trick to designing something this way is to make sure that most configurations are valid; this is the case here, as every value for `GracePeriod` and `DedupeLifetime` is valid.

> Plus it adds extra complication to the type that isn't needed (exported struct members that are only used once on a persistent object, a mutex that's only used during init),

All that complexity is hidden away, though. The simplicity of the API is more important, and this provides the simplest API as it eliminates the need for (1) one additional type, (2) one constructor.

> `// code snippet`

That to me looks more involved than just initializing the object.

> The constructor doesn't need refactoring, it's simple and pretty intuitive to use, allows errors to be caught at construction rather than at use sometime later

The current solution also doesn't need refactoring, and it cannot throw errors on later use because there are no errors to throw.

On a separate note, there also wasn't any reason why the snapshot of the values had to be there. Now callers can actually change behavior (in a concurrency-_unsafe_ way, which is same as package `http`) after creation.

---

*[go/lib/infra/dedupe/dedupe.go, line 150 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_AsPrj1GCj70aO33J:-L7dsUTdF55Mh84xeqpd:b-vgkf08) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L150)):*

Given that you've now specified that `RequestFunc` must handle context correctly, then you don't need to run it in another goroutine, or even check for `ctx.Done()`. It just simplifies the logic.
</blockquote></details>

I like the idea of `Deduper` reacting correctly to incorrectly programmed `RequestFuncs`, because it will be more intuitive for package users. They can see "oh, I got a time out, I need to fix this" even though their function blocks forever (and they can investigate it and fix the bug). I think it's useful to have it. It also gives some consistency regarding `DedupeLifetime`, as that will always work correctly and it guarantees that you will see some response after `DedupeLifetime` from starting your `Request`, even though the submitted `RequestFunc` is buggy.

---

*[go/lib/infra/dedupe/dedupe.go, line 162 at r2](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7dQ0WNSNK7Dw8BTtrT:-L7dsCd4uR4wimqX783v:byqy032) ([raw file](https://github.com/scionproto/scion/blob/43aa483463161117c066be57078d2ae930fd8082/go/lib/infra/dedupe/dedupe.go#L162)):*

That is not correct. You check for _errors_, of which context expiration is one possible error.
</blockquote></details>

By checking the error we also propagate that the context expired. I don't see why the statement is incorrect.

Maybe the sentence is confusing, I reworded it a bit.

---

*[go/lib/infra/dedupe/notification.go, line 30 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_3V0L7hNMzoNBtloY:-L7dvcy6hMLcwD4r6mbJ:b-eyy60z) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/notification.go#L30)):*

I think the choice is between a single long-running maintenance thread, vs potentially many short-lived cleanup threads. I think the former is easier to work with. If you're looking at the list of goroutines, it's a lot less noise to deal with.
</blockquote></details>

Sure, I have no strong feelings one way or another. Done.

---

*[go/lib/infra/dedupe/notification.go, line 80 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_5ehirfTSfVNF-CLw:-L7eFvUYAFLm8G310JE8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/notification.go#L80)):*

Same applies to `Deduper.Request()` and `Deduper.handler()`.
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/notification.go, line 122 at r2](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7dQsPTMTFlTBxcbvuZ:-L7eG7sKBHShXUIKN9Lk:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/43aa483463161117c066be57078d2ae930fd8082/go/lib/infra/dedupe/notification.go#L122)):*

s/object/response/
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/notification.go, line 169 at r2](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7dRVID0pEUeeH4Fy9V:-L7eG6NoKaJUWzkeyGEp:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/43aa483463161117c066be57078d2ae930fd8082/go/lib/infra/dedupe/notification.go#L169)):*

s/object/req/
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/notification.go, line 188 at r2](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7dRanZ-YbO3edk9Ijr:-L7eGFn8b3iT8GAQsqO8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/43aa483463161117c066be57078d2ae930fd8082/go/lib/infra/dedupe/notification.go#L188)):*

"on a request's broadcast key"
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

shitz:
Review status: 1 of 4 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[go/lib/infra/dedupe/dedupe.go, line 115 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_1PPJh_xdw58L8ERN:-L7ebcP5g_Z2pMKk0Vmi:bgl0qhq) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L115)):*

> It only really works when you create an object and use it ~immediately with a single call, e.g. `http.ListenAndServe()`

There's no reason why that has to be the case, proof in case, this package where it works with multiple calls to `Request`.

>  Otherwise if you have invalid configuration etc you don't find out until potentially Much later when you try to use it. 

The trick to designing something this way is to make sure that most configurations are valid; this is the case here, as every value for `GracePeriod` and `DedupeLifetime` is valid.

> Plus it adds extra complication to the type that isn't needed (exported struct members that are only used once on a persistent object, a mutex that's only used during init),

All that complexity is hidden away, though. The simplicity of the API is more important, and this provides the simplest API as it eliminates the need for (1) one additional type, (2) one constructor.

> `// code snippet`

That to me looks more involved than just initializing the object.

> The constructor doesn't need refactoring, it's simple and pretty intuitive to use, allows errors to be caught at construction rather than at use sometime later

The current solution also doesn't need refactoring, and it cannot throw errors on later use because there are no errors to throw.

On a separate note, there also wasn't any reason why the snapshot of the values had to be there. Now callers can actually change behavior (in a concurrency-_unsafe_ way, which is same as package `http`) after creation.
</blockquote></details>

I  agree with @kormat in this case. The way to use the `Deduper` is to configure it once and keep using it for requests. If you want to change some of the parameters you want to use another `Deduper` anyway, since it's dicey at best to change freshness and timeout parameter of a running deduper. I wouldn't export those values in the first place, since it's dangerous to change them.

In general, I'm all for avoiding unnecessary constructors, but in this case it seems that
1) The zero-value is actually not usable, since `notifications` needs to be initialized
2) There are parameters that shouldn't be changed after initialization, but need to be configurable during initialization.

These two points strongly suggest a constructor. @kormat was just providing a way that avoids the need to change the API later. Whether you want to implement it now is up to you.

---

*[go/lib/infra/dedupe/dedupe.go, line 84 at r3](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7eaA_q4VkEjNhW4TeJ:-L7eaA_q4VkEjNhW4TeK:bl5h7n9) ([raw file](https://github.com/scionproto/scion/blob/9ca8c03e5788f64f706b48d9f29014988ebfcd39/go/lib/infra/dedupe/dedupe.go#L84)):*
> ```Go
> // set during initialization; setting them after the first Request is undefined
> // behavior.
> type Deduper struct {
> ```

Personally, I think `Aggregator` or `RequestAggregator` would be a more fitting name. Moreover, whenever the word 'deduplicate` is used I think it could be replaced with 'aggregate` and it would sound better to me. I don't have a strong opinion about this, but I wanted to throw it out there.

---

*[go/lib/infra/dedupe/dedupe.go, line 98 at r3](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7e_dvC1bakPgixjfGC:-L7e_dvC1bakPgixjfGD:b-raqlfp) ([raw file](https://github.com/scionproto/scion/blob/9ca8c03e5788f64f706b48d9f29014988ebfcd39/go/lib/infra/dedupe/dedupe.go#L98)):*
> ```Go
> 	// for the same broadcast key are sent out. The result is immediately
> 	// returned from an internal cache for this period.
> 	GracePeriod time.Duration
> ```

`ResponseValidity` would be a better name imho.

---

*[go/lib/infra/dedupe/notification.go, line 30 at r3](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7ehxMK2fu0-4Xz1uJf:-L7ehxMK2fu0-4Xz1uJg:b-14zrgg) ([raw file](https://github.com/scionproto/scion/blob/9ca8c03e5788f64f706b48d9f29014988ebfcd39/go/lib/infra/dedupe/notification.go#L30)):*
> ```Go
> 
> // notifyList maintains a set of channels for disseminating responses. The keys
> // map to themselves. Each channel corresponds to one client call to
> ```

Comment outdated.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

shitz:
Reviewed 3 of 5 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

scrye:
Review status: 1 of 4 files reviewed at latest revision, 11 unresolved discussions.

---

*[go/lib/infra/dedupe/dedupe.go, line 115 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_1PPJh_xdw58L8ERN:-L7i2cKrtnQljE5Wt_CA:b-jsyygf) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L115)):*

I  agree with @kormat in this case. The way to use the `Deduper` is to configure it once and keep using it for requests. If you want to change some of the parameters you want to use another `Deduper` anyway, since it's dicey at best to change freshness and timeout parameter of a running deduper. I wouldn't export those values in the first place, since it's dangerous to change them.

In general, I'm all for avoiding unnecessary constructors, but in this case it seems that
1) The zero-value is actually not usable, since `notifications` needs to be initialized
2) There are parameters that shouldn't be changed after initialization, but need to be configurable during initialization.

These two points strongly suggest a constructor. @kormat was just providing a way that avoids the need to change the API later. Whether you want to implement it now is up to you.
</blockquote></details>

I disagree. Re: 1, initializing on demand is a useful pattern when it simplifies the API. Re 2: In the spirit of good Go practices, only method `Request` is marked as safe for use from multiple goroutines; anything outside of this is in the danger zone.

In any case, I don't want this to block the PR, so I've changed it to an explicit constructor. While I think the `Option` struct might be useful, I'm a bit scared by the number of similar structures that we would have in the codebase (one for possibly every server). To have some consistency and minimize API surface, I'll just use constructors.

---

*[go/lib/infra/dedupe/dedupe.go, line 84 at r3](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7eaA_q4VkEjNhW4TeJ:-L7i2-bu6cP3YUO7meJT:banzaj0) ([raw file](https://github.com/scionproto/scion/blob/9ca8c03e5788f64f706b48d9f29014988ebfcd39/go/lib/infra/dedupe/dedupe.go#L84)):*

Personally, I think `Aggregator` or `RequestAggregator` would be a more fitting name. Moreover, whenever the word 'deduplicate` is used I think it could be replaced with 'aggregate` and it would sound better to me. I don't have a strong opinion about this, but I wanted to throw it out there.
</blockquote></details>

The problem with aggregator is that I've seen the same name used for the behavior of bundling (aggregating) multiple requests into a single network message. So every request is put on the wire, but with fewer messages. I wanted to avoid that confusion, hence the weirder naming.

---

*[go/lib/infra/dedupe/dedupe.go, line 98 at r3](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7e_dvC1bakPgixjfGC:-L7i2GzIafJPvi2C3zF2:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9ca8c03e5788f64f706b48d9f29014988ebfcd39/go/lib/infra/dedupe/dedupe.go#L98)):*

`ResponseValidity` would be a better name imho.
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/notification.go, line 30 at r3](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7ehxMK2fu0-4Xz1uJf:-L7i6grUuDUGiu5cSHiq:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9ca8c03e5788f64f706b48d9f29014988ebfcd39/go/lib/infra/dedupe/notification.go#L30)):*

Comment outdated.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

shitz:
Review status: 1 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/lib/infra/dedupe/dedupe.go, line 115 at r1](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7_1PPJh_xdw58L8ERN:-L7iDbuNhVXfQnMa-kEm:bt927bx) ([raw file](https://github.com/scionproto/scion/blob/c26bf0d92cb013076fec6429fa20d8ca72c8b494/go/lib/infra/dedupe/dedupe.go#L115)):*

I disagree. Re: 1, initializing on demand is a useful pattern when it simplifies the API. Re 2: In the spirit of good Go practices, only method `Request` is marked as safe for use from multiple goroutines; anything outside of this is in the danger zone.

In any case, I don't want this to block the PR, so I've changed it to an explicit constructor. While I think the `Option` struct might be useful, I'm a bit scared by the number of similar structures that we would have in the codebase (one for possibly every server). To have some consistency and minimize API surface, I'll just use constructors.
</blockquote></details>

The options pattern is very common and I used it extensively in a large project/codebase in the past. Nothing to be afraid of there ;)

---

*[go/lib/infra/dedupe/dedupe.go, line 84 at r3](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7eaA_q4VkEjNhW4TeJ:-L7iB7uMMDBIZAwpm4rZ:b-4liqwm) ([raw file](https://github.com/scionproto/scion/blob/9ca8c03e5788f64f706b48d9f29014988ebfcd39/go/lib/infra/dedupe/dedupe.go#L84)):*

The problem with aggregator is that I've seen the same name used for the behavior of bundling (aggregating) multiple requests into a single network message. So every request is put on the wire, but with fewer messages. I wanted to avoid that confusion, hence the weirder naming.
</blockquote></details>

Ok, that's a good point.

---

*[go/lib/infra/dedupe/dedupe.go, line 107 at r4](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7iCzD4m0Z-NvqFfbjr:-L7iCzD4m0Z-NvqFfbjs:b-o5vbqp) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe.go#L107)):*
> ```Go
> // immediately returned from an internal cache for this period. If 0,
> // responseValidity defaults to DefaultResponseValidity.
> func New(f RequestFunc, dedupeLifetime, responseValidity time.Duration) *Deduper {
> ```

Btw, what if a client wants no caching at all? Shouldn't `responseValidity == 0` be allowed and valid?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

kormat:
Reviewed 1 of 5 files at r2, 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[go/lib/infra/dedupe/dedupe.go, line 162 at r2](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7dQ0WNSNK7Dw8BTtrT:-L7iB_PovKA0TbYR36vU:b-5xbffc) ([raw file](https://github.com/scionproto/scion/blob/43aa483463161117c066be57078d2ae930fd8082/go/lib/infra/dedupe/dedupe.go#L162)):*

By checking the error we also propagate that the context expired. I don't see why the statement is incorrect.

Maybe the sentence is confusing, I reworded it a bit.
</blockquote></details>

I think the whole comment can be dropped at this point. It's obvious what the code does, there's no race-condition that needs to be discussed, etc. It's more misleading than anything else.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 40 at r4](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7iHOUQWPx4aXfMfyyk:-L7iHOUQWPx4aXfMfyyl:bnxweri) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L40)):*
> ```Go
> 	// Data and Error are the cooked request response (simulates network response)
> 	Data  string
> 	Error error
> ```

There is no test that actually sets `Error` to non-nil. It would be good to test that if a non-timeout error happens in RequestFunc that it is returned appropriately.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 44 at r4](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7iEiLTSKFlGdNd2fpD:-L7iEiLTSKFlGdNd2fpE:b5vgqs6) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L44)):*
> ```Go
> 	// Delay is the time to wait before the Request runs in the test
> 	Delay time.Duration
> 	// Latency is the time while Request runs (simulates network and server
> ```

"the time while Request" - i can't parse that.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 72 at r4](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7iEwT0Cdl1kvDskuZ1:-L7iEwT0Cdl1kvDskuZ2:b-6jxp0w) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L72)):*
> ```Go
> 	srv.mu.Unlock()
> 	select {
> 	case <-time.After(trequest.Latency * time.Millisecond):
> ```

Wait, `Latency` is already `time.Duration`. Either the `* time.Millisecond` should be done when a new `TRequest` is made, or `Latency` should be an int, and documented as being treated as milliseconds.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 97 at r4](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7iGb_R7lHvzpbyCHGt:-L7iGb_R7lHvzpbyCHGu:b-m2fpfu) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L97)):*
> ```Go
> 			Name: "one request",
> 			requests: []TRequest{
> 				{DKey: "a-x", BKey: "a", Data: "foo", Error: nil, Latency: 0},
> ```

It would be nice to have a case where the `DKey` and `BKey` were the same, as i imagine that will be quite common.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 156 at r4](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7iH3XnlUJrGKX9Kk-r:-L7iH3XnlUJrGKX9Kk-s:b-jsdisi) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L156)):*
> ```Go
> 			requests: []TRequest{
> 				{DKey: "a-x", BKey: "a", Data: "foo", Error: nil, Latency: 0},
> 				{DKey: "a-x", BKey: "a", Data: "foo", Error: nil, Delay: 500, Latency: 0},
> ```

It would be nice to also have a `a-y` Request here.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 215 at r4](https://reviewable.io:443/reviews/scionproto/scion/1452#-L7iFchs06Sy4vLzfVAO:-L7iFchs06Sy4vLzfVAP:bnhfb1l) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L215)):*
> ```Go
> 		}
> 	})
> ```

Blank line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

scrye:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[go/lib/infra/dedupe/dedupe.go, line 162 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1452#-L7dQ0WNSNK7Dw8BTtrT:-L7iI9ih940Py82k_Whp:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/43aa483463161117c066be57078d2ae930fd8082/go/lib/infra/dedupe/dedupe.go#L162)):*

I think the whole comment can be dropped at this point. It's obvious what the code does, there's no race-condition that needs to be discussed, etc. It's more misleading than anything else.
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/dedupe.go, line 107 at r4](https://beta.reviewable.io/reviews/scionproto/scion/1452#-L7iCzD4m0Z-NvqFfbjr:-L7iLwYvEGQzYX4bVkTm:beoe75d) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe.go#L107)):*

Btw, what if a client wants no caching at all? Shouldn't `responseValidity == 0` be allowed and valid?
</blockquote></details>

Putting it on 0 makes the package behave in a somewhat racey way (you might get stuck with a network request going out and the caller waiting for it while the data is actually there from a prior request), so I'm not too fond about supporting it. There's no way around this, as the race is external to the package. E.g., if A runs the below first and B evaluates the `if` condition just as the Request is finishing:

```
    if smthMissing() {
       // race is here
       <- d.Request()
    }
```

`responseValidity` decreases the probability of this race happening (although it is best effort, and does not eliminate it completely).

---

*[go/lib/infra/dedupe/dedupe_test.go, line 40 at r4](https://beta.reviewable.io/reviews/scionproto/scion/1452#-L7iHOUQWPx4aXfMfyyk:-L7iJ2jNDyTz61BBfbF4:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L40)):*

There is no test that actually sets `Error` to non-nil. It would be good to test that if a non-timeout error happens in RequestFunc that it is returned appropriately.
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 44 at r4](https://beta.reviewable.io/reviews/scionproto/scion/1452#-L7iEiLTSKFlGdNd2fpD:-L7iJ38QADPOicNB15vD:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L44)):*

"the time while Request" - i can't parse that.
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 72 at r4](https://beta.reviewable.io/reviews/scionproto/scion/1452#-L7iEwT0Cdl1kvDskuZ1:-L7iJMc4302cR8BvikEE:b-aq656j) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L72)):*

Wait, `Latency` is already `time.Duration`. Either the `* time.Millisecond` should be done when a new `TRequest` is made, or `Latency` should be an int, and documented as being treated as milliseconds.
</blockquote></details>

Good point, it was misleading. Done.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 97 at r4](https://beta.reviewable.io/reviews/scionproto/scion/1452#-L7iGb_R7lHvzpbyCHGt:-L7iJPl34muM1P2Qw3wy:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L97)):*

It would be nice to have a case where the `DKey` and `BKey` were the same, as i imagine that will be quite common.
</blockquote></details>

Done.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 156 at r4](https://beta.reviewable.io/reviews/scionproto/scion/1452#-L7iH3XnlUJrGKX9Kk-r:-L7iJZZx-yu60bjaIWj-:b-k0bt3b) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L156)):*

It would be nice to also have a `a-y` Request here.
</blockquote></details>

Yes, that's useful to see. Done.

---

*[go/lib/infra/dedupe/dedupe_test.go, line 215 at r4](https://beta.reviewable.io/reviews/scionproto/scion/1452#-L7iFchs06Sy4vLzfVAO:-L7iJgMg1hu6UUbvhT-W:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5c23c4b7a583adc6f51d38f9d1c4ae193f997ba2/go/lib/infra/dedupe/dedupe_test.go#L215)):*

Blank line.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1452)*

shitz:
Reviewed 1 of 5 files at r2, 1 of 3 files at r4, 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452)*

kormat:
One minor comment left, otherwise LGTM.

---

Reviewed 2 of 2 files at r5.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1452#-:-L7iZeX0DphbmoESgqtl:b-u2fge0)*

shitz:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452#-:-L7iZmcJnePqnYAO35v2:bnfp4nl)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1452#-:-L7jLhVum6d8E66lw88W:bnfp4nl)*

PR #1453, opened 2018-03-15
Switch to 64-bit ISD-ASes

### LINKED ISSUE #1451: Switch to 64b ISD-AS numbers
The existing 32-bit ISD-AS numbers are divided into 12bits for ISD (~4k ISDs) and 20bits for AS (~1m ASes). The latter is likely to be too small in reality. It's possible that many existing BGP ASes will split into multiple scion ASes, and it's important we don't make the mistake of making the namespace too small.

My proposal is as follows:
- Switch to 64b ISD-AS numbers
- The top 16b will be for ISDs, of which the top 4 will be reserved. This retains the existing ~4k limit on ISDs for the moment, which is desirable because ISD proliferation is something we want to avoid for scaling reasons.
- The bottom 48b will be for ASes. These AS identifiers will be *global*, not scoped to the ISD as they are in the current design.

The reasons for going for global AS identifiers are:
-  It allows multi-homed ASes (e.g. a business that is in multiple ISDs) to have a clear connection between its multiple ISD-AS identitier, as the AS number will be the same.
- It means that existing BGP ASes can have their AS number ported directly to SCION, which makes things easier for everyone, and is easier to explain. The current BGP AS namespace is 32b, with first 16b of values fairly densely populated, but the remainder of the space is very sparsely used at the moment. Any _new_ SCION AS ids would be provisioned above the 32b boundary, to avoid any potential clashes.
- The major scaling issue in core routing for SCION is the number of core ASes. With the current system an AS that is a core in multiple ISDs must propagate separate beacons for each ISD. This leads to many more beacons than are needed, especially in the case of existing tier1 ISPs which will probably be core ASes in many ISDs. By switching to a global AS identifier, beacons could potentially have a single AS id, but a list of ISDs that the AS is a member of, and thereby have a single beacon instead of one per ISD. This needs to be investigated more closely to see if it would actually work, but the global AS choice at least makes it something we can explore.

kormat:
This is ready for review, but i still need to update some tests that use hard-coded packets.

This addresses part of #1451 (the globally uniqueness is not included).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1453)


scrye:
Reviewed 27 of 32 files at r1.
Review status: 27 of 32 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[c/lib/scion/address.h, line 35 at r1](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7eMl7aWGNXfej56Mdk:-L7eMl7b1FVYJ__ZZHkb:bswq5cj) ([raw file](https://github.com/scionproto/scion/blob/f178cf777b23cfeb605bf97c6a061a3e44cbb737/c/lib/scion/address.h#L35)):*
> ```objc
> #define AS_MASK ((1L << AS_BITS) - 1)
> 
> #define ISD(isd_as) ((uint16_t)(isd_as >> AS_BITS))
> ```

Since we're here, add parentheses around `isd_as` here and on the line below to avoid some odd behaviors like:
```
x = 1<<48;
// ISD(x & x) != ISD((x & x))
```

---

*[c/lib/scion/address.h, line 69 at r1](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7ePi7FjNddAi9hUbc5:-L7ePi7FjNddAi9hUbc6:b-wm6qmx) ([raw file](https://github.com/scionproto/scion/blob/f178cf777b23cfeb605bf97c6a061a3e44cbb737/c/lib/scion/address.h#L69)):*
> ```objc
> 
> #define SADDR_ISD(saddr) ISD(ntohs(*(isdas_t *)(saddr.addr)))
> #define SADDR_AS(saddr) AS(ntohs(*(isdas_t *)(saddr.addr)))
> ```

This looks like it silently truncates; `*(isdas_t *)(saddr.addr)` might contain an AS number on 48 bits, but `ntohs` truncates it to the lower 16 bits. I think this breaks for large AS numbers  that are greater or equal to `1>>16`.

Consider adding `-Wconversion` to the compile line flags to catch other issues like this.

---

*[go/lib/addr/isdas.go, line 34 at r1](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7eTd_3Y__CERFF6pN2:-L7eTd_3Y__CERFF6pN3:bjqivcv) ([raw file](https://github.com/scionproto/scion/blob/f178cf777b23cfeb605bf97c6a061a3e44cbb737/go/lib/addr/isdas.go#L34)):*
> ```Go
> )
> 
> type ISDInt uint16
> ```

Why not just `ISD` and `AS`? `addr.ISD` and `addr.AS` seem explicit enough.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1453)*

kormat:
Review status: 26 of 40 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[c/lib/scion/address.h, line 69 at r1](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7ePi7FjNddAi9hUbc5:-L7eY04TujUMWupM5mS_:bey6x4t) ([raw file](https://github.com/scionproto/scion/blob/f178cf777b23cfeb605bf97c6a061a3e44cbb737/c/lib/scion/address.h#L69)):*

This looks like it silently truncates; `*(isdas_t *)(saddr.addr)` might contain an AS number on 48 bits, but `ntohs` truncates it to the lower 16 bits. I think this breaks for large AS numbers  that are greater or equal to `1>>16`.

Consider adding `-Wconversion` to the compile line flags to catch other issues like this.
</blockquote></details>

Good catch! Fixed.

I tried `-Wconversion`, there's faaar to many errors just in `libscion`, unfortunately.

---

*[go/lib/addr/isdas.go, line 34 at r1](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7eTd_3Y__CERFF6pN2:-L7eYvmfneDJ8igztzFA:b-tg96fy) ([raw file](https://github.com/scionproto/scion/blob/f178cf777b23cfeb605bf97c6a061a3e44cbb737/go/lib/addr/isdas.go#L34)):*

Why not just `ISD` and `AS`? `addr.ISD` and `addr.AS` seem explicit enough.
</blockquote></details>

Sure, why not :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1453)*

kormat:
Review status: 18 of 40 files reviewed at latest revision, 3 unresolved discussions.

---

*[c/lib/scion/address.h, line 35 at r1](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7eMl7aWGNXfej56Mdk:-L7eYyolXWOaXsGum8LZ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/f178cf777b23cfeb605bf97c6a061a3e44cbb737/c/lib/scion/address.h#L35)):*

Since we're here, add parentheses around `isd_as` here and on the line below to avoid some odd behaviors like:
```
x = 1<<48;
// ISD(x & x) != ISD((x & x))
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1453)*

shitz:
:lgtm:

---

Reviewed 23 of 32 files at r1, 20 of 20 files at r2, 1 of 2 files at r3.
Review status: 43 of 44 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[go/lib/pathdb/query/query.go, line 29 at r2](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7hnqYYeBLjgxL431ND:-L7hnqYZWcFHbVTjOATK:b2277xd) ([raw file](https://github.com/scionproto/scion/blob/42895a86ef36195d15732ba0d4ec6859519e30f0/go/lib/pathdb/query/query.go#L29)):*
> ```Go
> }
> 
> func NewHPCfgID(ia addr.IA, ID uint64) *HPCfgID {
> ```

This could be removed. Imo, constructors that simply set exported fields based on parameters passed in are not needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1453#-:-L7hp54j6HQK2vtgr9f7:bnfp4nl)*

scrye:
Reviewed 5 of 32 files at r1, 19 of 20 files at r2, 2 of 2 files at r3, 1 of 1 files at r4.

---

*[c/lib/scion/address.h, line 69 at r1](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7ePi7FjNddAi9hUbc5:-L7iBItfIwVOjDZbvSss:b-kpzncl) ([raw file](https://github.com/scionproto/scion/blob/f178cf777b23cfeb605bf97c6a061a3e44cbb737/c/lib/scion/address.h#L69)):*

Good catch! Fixed.

I tried `-Wconversion`, there's faaar to many errors just in `libscion`, unfortunately.
</blockquote></details>

Sorry, I meant both lines truncate.

The upper one is subtler; because `ntohs` returns a `uint16`, the shift right in macro `ISD` always puts the ISD to 0. There is also a different issue here with the lower 4 bits of the 2-byte ISD, but it is also fixed by `ntohs` -> `be64toh`.

I don't think `SADDR_ISD` and SADDR_AS` are used anywhere though, which is why these issues never cropped up. You might consider just deleting them.

---

*[python/lib/packet/scion_addr.py, line 33 at r4](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7iEYy0Ce8fEiy3ylR1:-L7iEYy0Ce8fEiy3ylR2:b-dbb0ix) ([raw file](https://github.com/scionproto/scion/blob/9420b1a229336c79ca67e7818b8370c3590b3ceb/python/lib/packet/scion_addr.py#L33)):*
> ```Python
> class ISD_AS(Serializable):
>     """
>     Class for representing ISD-AS pair. The underyling type is a 64bit int; ISD is represented by
> ```

underlying

64bit unsigned int

---

*[python/lib/packet/scion_addr.py, line 57 at r4](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7iExYfNCxTp8gw_n2q:-L7iExYgYnSo4X1HpqIi:bm8mj2w) ([raw file](https://github.com/scionproto/scion/blob/9420b1a229336c79ca67e7818b8370c3590b3ceb/python/lib/packet/scion_addr.py#L57)):*
> ```Python
>     def _parse_bytes(self, raw):
>         """
>         :param bytes raw: a byte string containing a 64-bit integer.
> ```

unsigned integer

---

*[python/lib/packet/scion_addr.py, line 79 at r4](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7iF0ZtjsaH016LFQvh:-L7iF0ZtjsaH016LFQvi:bm8mj2w) ([raw file](https://github.com/scionproto/scion/blob/9420b1a229336c79ca67e7818b8370c3590b3ceb/python/lib/packet/scion_addr.py#L79)):*
> ```Python
>     def _parse_int(self, raw):
>         """
>         :param int raw: a 64-bit integer
> ```

unsigned integer

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1453)*

kormat:

---

*[c/lib/scion/address.h, line 69 at r1](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7ePi7FjNddAi9hUbc5:-L7iJMmWoaUiQUWW3GP_:b-elh40o) ([raw file](https://github.com/scionproto/scion/blob/f178cf777b23cfeb605bf97c6a061a3e44cbb737/c/lib/scion/address.h#L69)):*

Sorry, I meant both lines truncate.

The upper one is subtler; because `ntohs` returns a `uint16`, the shift right in macro `ISD` always puts the ISD to 0. There is also a different issue here with the lower 4 bits of the 2-byte ISD, but it is also fixed by `ntohs` -> `be64toh`.

I don't think `SADDR_ISD` and SADDR_AS` are used anywhere though, which is why these issues never cropped up. You might consider just deleting them.
</blockquote></details>

Another good point re: ISD :) Removed in any case as they're not used, as you pointed out.

---

*[go/lib/pathdb/query/query.go, line 29 at r2](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7hnqYYeBLjgxL431ND:-L7iIPwHBybyB43l1fF6:b2a524a) ([raw file](https://github.com/scionproto/scion/blob/42895a86ef36195d15732ba0d4ec6859519e30f0/go/lib/pathdb/query/query.go#L29)):*

This could be removed. Imo, constructors that simply set exported fields based on parameters passed in are not needed.
</blockquote></details>

I think in this case i probably agree, but for more complex structs they can be useful as a way to make sure that all the relevant fields are specified, and to greatly reduce the amount of boilerplate for the caller.

---

*[python/lib/packet/scion_addr.py, line 33 at r4](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7iEYy0Ce8fEiy3ylR1:-L7iIHu7kxzy4roXHJ5Q:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9420b1a229336c79ca67e7818b8370c3590b3ceb/python/lib/packet/scion_addr.py#L33)):*

underlying

64bit unsigned int
</blockquote></details>

Done.

---

*[python/lib/packet/scion_addr.py, line 57 at r4](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7iExYfNCxTp8gw_n2q:-L7iIHkNFRlOiXUqeYGD:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9420b1a229336c79ca67e7818b8370c3590b3ceb/python/lib/packet/scion_addr.py#L57)):*

unsigned integer
</blockquote></details>

Done.

---

*[python/lib/packet/scion_addr.py, line 79 at r4](https://reviewable.io:443/reviews/scionproto/scion/1453#-L7iF0ZtjsaH016LFQvh:-L7iILAdGSTxTnoXFHta:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9420b1a229336c79ca67e7818b8370c3590b3ceb/python/lib/packet/scion_addr.py#L79)):*

unsigned integer
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1453)*

scrye:
:lgtm:

---

Reviewed 3 of 3 files at r5.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1453#-:-L7iiCOL4I1eJmew00t1:bnfp4nl)*

shitz:
:lgtm:

---

Reviewed 1 of 2 files at r3, 1 of 1 files at r4, 3 of 3 files at r5, 9 of 9 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1453#-:-L7jDaT2tGVWCFARUqlz:bnfp4nl)*

PR #1454, opened 2018-03-16
Add SCMP Trace Route

sgmonroy:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1454)


kormat:
Reviewed 14 of 14 files at r1.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[go/border/error.go, line 35 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XcsFP4nx7OGnMDrCm:-L8XcsFP4nx7OGnMDrCn:b5fudpq) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/error.go#L35)):*
> ```Go
> }
> 
> // handlePacketError is called to enqueue packets with protocol-level errors
> ```

`handlePktError`

---

*[go/border/error.go, line 41 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XdVD7_fKSszYSpBkt:-L8XdVD7_fKSszYSpBku:b-5bfcxn) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/error.go#L41)):*
> ```Go
> 	// perr = common.NewBasicError("Raw packet", perr, "raw", rp.Raw)
> 	rp.Error(desc, "err", perr)
> 	rp.RefInc(1)
> ```

For every `rp.RefInc` there must be a `rp.Release()`

---

*[go/border/error.go, line 57 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XdB2_HGMRyt7SJltc:-L8XdB2_HGMRyt7SJltd:bnhfb1l) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/error.go#L57)):*
> ```Go
> 		r.doPktError(args.rp, args.perr)
> 	}
> ```

Blank line.

---

*[go/border/error.go, line 85 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XQ3Roln_K7q-HLAUF:-L8XQ3Roln_K7q-HLAUG:brs2u8j) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/error.go#L85)):*
> ```Go
> 	rp.Debug("", "srcIA", srcIA, "rp", rp)
> 	rp.Debug("", "rp.Ctx", rp.Ctx)
> 	rp.Debug("", "rp.Ctx.Conf", rp.Ctx.Conf)
> ```

Debugging statements.

---

*[go/border/setup.go, line 65 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XPONWyun95gRIu4sz:-L8XPONWyun95gRIu4t-:bs0t4an) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/setup.go#L65)):*
> ```Go
> 	}, "free", prometheus.Labels{"ringId": "freePkts"})
> 	r.revInfoQ = make(chan rpkt.RevTokenCallbackArgs)
> 	r.pktErrorQ = make(chan pktErrorArgs)
> ```

This should probably be buffered, at least a bit. I suggest `16` to begin with.

---

*[go/border/rpkt/create.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XPWh-jXtXM2ZtyDB1:-L8XPWh-jXtXM2ZtyDB2:brzz8rv) ([raw file](https://github.com/scionproto/scion/blob/002c9e013a3076e65836fad48f268a35118b8c06/go/border/rpkt/create.go#L1)):*
> ```Go
> // Copyright 2016 ETH Zurich
> ```

Whoa, you can't remove that.

---

*[go/border/rpkt/create.go, line 164 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8Xbv_Mwm9Zu7r06jw8:-L8Xbv_Mwm9Zu7r06jw9:b-ihit2c) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/rpkt/create.go#L164)):*
> ```Go
> }
> 
> // createReplyScnPkt creates a generic ScnPkt reply, by converting the RtrPkt
> ```

`CreateReplyScnPkt`.

---

*[go/border/rpkt/create.go, line 232 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XcfdZjRY3lvECdEzK:-L8XcfdZjRY3lvECdEzL:b-ux63cm) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/rpkt/create.go#L232)):*
> ```Go
> // replyEgress calculates the corresponding egress function and destination
> // address to use when replying to a packet.
> func (rp *RtrPkt) ReplyEgress() (EgressPair, error) {
> ```

This is only used inside `rpkt`, so i think it can be unexported.

---

*[go/border/rpkt/process.go, line 219 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XPjKw5mMWsVRJWF8O:-L8XPjKw5mMWsVRJWF8P:b4wwmd8) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/rpkt/process.go#L219)):*
> ```Go
> 	hdr := rp.l4.(*scmp.Hdr)
> 	switch {
> 	case hdr.Class == scmp.C_General && hdr.Type == scmp.T_G_TraceRouteReply:
> ```

Hmm. It occurs to me that traceroute (and recordpath) replies shouldn't be HBH. There's no reason the BR has to process them on the way back.

---

*[go/border/rpkt/process.go, line 259 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XVczXigsXNPTU9pUm:-L8XVczXigsXNPTU9pUn:b-mbs3hb) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/rpkt/process.go#L259)):*
> ```Go
> 		return nil
> 	}
> 	if *rp.upFlag != infoTrace.In && rp.hopF.Ingress != *rp.ifCurr {
> ```

The logic here is melting my brain. From testing, it does seem to do the right thing, but...
I think this can be simplified by using directions:
```
// If In is set and the packet came from the local AS,
// or if In is false and the packet came from outside,
// then stop processing.
if infoTrace.In != (rp.DirFrom == rcmn.DirExternal) {
  return nil
}
```
This also has the advantage that it doesn't reply on the path header (e.g. if a traceroute was sent using a OneHopPath extension, then the hop field might actually be blank).

---

*[go/lib/scmp/info_traceroute.go, line 44 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XQTYuJVBD3EqdHrMl:-L8XQTYuJVBD3EqdHrMm:b-mz4m4e) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/lib/scmp/info_traceroute.go#L44)):*
> ```Go
> // the IA and IfID fields.
> // If the HopOff is 0, the packet is meant for the endhost and BR should just
> // forward the packet.
> ```

In that case it shouldn't be a HBH packet, so the BR shouldn't be seeing the payload.

---

*[go/lib/scmp/scmp.go, line 54 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XP0zzATVFTKW1Fe5C:-L8XP0zzATVFTKW1Fe5D:bf6k7nx) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/lib/scmp/scmp.go#L54)):*
> ```Go
> 	T_G_TraceRouteReply
> 	T_G_RecordPathRequest
> 	T_G_RecordPathReply
> ```

The traceroute and recordpath types are missing from the python lib (python/lib/packet/scmp/types.py).

---

*[go/tools/scmp/types.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XXfYVBdr51eLKrPpi:-L8XXfYVBdr51eLKrPpj:b87fe6n) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/tools/scmp/types.go#L1)):*
> ```Go
> package main
> ```

The timing logic for traceroute is different than echo. It looks like this at a high level:

1. send a request
1. wait for reply or timeout, increment the hop offset goto 1.

Basically, the interval should default to `0`, and sending packets should be gated on receiving packets (or a timeout). If i try to emulate this by setting `-interval` to a very small number, i sometimes get replies out of sequence (and `-interval 0` causes a panic).

My suggestion would be to modularise the code of `scmp`, and have `scmp echo`, `scmp recordpath`, `scmp traceroute` etc subcommands.

Also, there's no mention of traceroute in the output of `scmp -h`.

---

*[go/tools/scmp/types.go, line 107 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XRacGMyyElhAA-4Tw:-L8XRacGMyyElhAA-4Tx:b3pyiul) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/tools/scmp/types.go#L107)):*
> ```Go
> 	s.ctR = scmp.ClassType{Class: scmp.C_General, Type: scmp.T_G_TraceRouteReply}
> 	// Send packet
> 	ext := &scmp.Extn{Error: false, HopByHop: true}
> ```

I think this should only be included if there is a path.

---

*[go/tools/scmp/types.go, line 193 at r1](https://reviewable.io:443/reviews/scionproto/scion/1454#-L8XTkf8FvcVRhHoWcCg:-L8XTkf8FvcVRhHoWcCh:b-yj8869) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/tools/scmp/types.go#L193)):*
> ```Go
> 	case scmp.ClassType{Class: scmp.C_General, Type: scmp.T_G_EchoReply}:
> 		s.infoR, ok = scmpPld.Info.(*scmp.InfoEcho)
> 		if ok == false {
> ```

`if !ok` is more idiomatic.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1454)*

sgmonroy:
*[go/border/rpkt/process.go, line 219 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XPjKw5mMWsVRJWF8O:-L8h7Oc7EHBobbYjsF6A:b-op0vc5) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/rpkt/process.go#L219)):*

Hmm. It occurs to me that traceroute (and recordpath) replies shouldn't be HBH. There's no reason the BR has to process them on the way back.
</blockquote></details>

Right, need to update the dispatcher to remove HBH extension when replying

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

sgmonroy:
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[go/border/error.go, line 35 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XcsFP4nx7OGnMDrCm:-L8gLQbM-nlI6zAwn0h3:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/error.go#L35)):*

`handlePktError`
</blockquote></details>

Done.

---

*[go/border/error.go, line 41 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XdVD7_fKSszYSpBkt:-L8gLXvT-yZ3VCdfEwbw:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/error.go#L41)):*

For every `rp.RefInc` there must be a `rp.Release()`
</blockquote></details>

Done.

---

*[go/border/error.go, line 57 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XdB2_HGMRyt7SJltc:-L8gLhPy4hqo9oPrzaZO:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/error.go#L57)):*

Blank line.
</blockquote></details>

Done.

---

*[go/border/error.go, line 85 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XQ3Roln_K7q-HLAUF:-L8gLkfLD2xublciTbCX:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/error.go#L85)):*

Debugging statements.
</blockquote></details>

Done.

---

*[go/border/setup.go, line 65 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XPONWyun95gRIu4sz:-L8gMe2B7tDHvDGyb51R:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/setup.go#L65)):*

This should probably be buffered, at least a bit. I suggest `16` to begin with.
</blockquote></details>

Done.

---

*[go/border/rpkt/create.go, line 1 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XPWh-jXtXM2ZtyDB1:-L8h5T1a5ZDdZ8uZ7p1K:b-uv87ed) ([raw file](https://github.com/scionproto/scion/blob/002c9e013a3076e65836fad48f268a35118b8c06/go/border/rpkt/create.go#L1)):*

Whoa, you can't remove that.
</blockquote></details>

Right the copyright, that was not intended.
done.

---

*[go/border/rpkt/create.go, line 164 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8Xbv_Mwm9Zu7r06jw8:-L8h5wGhCjI-6t1dHynD:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/rpkt/create.go#L164)):*

`CreateReplyScnPkt`.
</blockquote></details>

Done.

---

*[go/border/rpkt/create.go, line 232 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XcfdZjRY3lvECdEzK:-L8h639CCGpdAIhvoBac:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/rpkt/create.go#L232)):*

This is only used inside `rpkt`, so i think it can be unexported.
</blockquote></details>

Done.

---

*[go/border/rpkt/process.go, line 259 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XVczXigsXNPTU9pUm:-L8h7M0b74TvzLrXlxt_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/border/rpkt/process.go#L259)):*

The logic here is melting my brain. From testing, it does seem to do the right thing, but...
I think this can be simplified by using directions:
```
// If In is set and the packet came from the local AS,
// or if In is false and the packet came from outside,
// then stop processing.
if infoTrace.In != (rp.DirFrom == rcmn.DirExternal) {
  return nil
}
```
This also has the advantage that it doesn't reply on the path header (e.g. if a traceroute was sent using a OneHopPath extension, then the hop field might actually be blank).
</blockquote></details>

Done.

---

*[go/lib/scmp/info_traceroute.go, line 44 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XQTYuJVBD3EqdHrMl:-L8hOCSK-nfNKz6HlBaG:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/lib/scmp/info_traceroute.go#L44)):*

In that case it shouldn't be a HBH packet, so the BR shouldn't be seeing the payload.
</blockquote></details>

Done.

---

*[go/lib/scmp/scmp.go, line 54 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XP0zzATVFTKW1Fe5C:-L8kwYWa64_G-cfbfmpb:b-10bqt8) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/lib/scmp/scmp.go#L54)):*

The traceroute and recordpath types are missing from the python lib (python/lib/packet/scmp/types.py).
</blockquote></details>

Done.kkkkkkk

---

*[go/tools/scmp/types.go, line 1 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XXfYVBdr51eLKrPpi:-L95szRe9n_INAvHmS63:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/tools/scmp/types.go#L1)):*

The timing logic for traceroute is different than echo. It looks like this at a high level:

1. send a request
1. wait for reply or timeout, increment the hop offset goto 1.

Basically, the interval should default to `0`, and sending packets should be gated on receiving packets (or a timeout). If i try to emulate this by setting `-interval` to a very small number, i sometimes get replies out of sequence (and `-interval 0` causes a panic).

My suggestion would be to modularise the code of `scmp`, and have `scmp echo`, `scmp recordpath`, `scmp traceroute` etc subcommands.

Also, there's no mention of traceroute in the output of `scmp -h`.
</blockquote></details>

Done.

---

*[go/tools/scmp/types.go, line 107 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XRacGMyyElhAA-4Tw:-L8kwMjn-2bOdNKwBNAK:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/tools/scmp/types.go#L107)):*

I think this should only be included if there is a path.
</blockquote></details>

Done.

---

*[go/tools/scmp/types.go, line 193 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L8XTkf8FvcVRhHoWcCg:-L8kwWOd3tMTaAcysok2:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ecf6e95d5d3c0806d1896cc5049c9ca29a56c2a7/go/tools/scmp/types.go#L193)):*

`if !ok` is more idiomatic.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

sgmonroy:
*[go/tools/scmp/main.go, line 85 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AQYUl--iSD17PdnoR:-L9AQYUl--iSD17PdnoS:bh2wclm) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L85)):*
> ```Go
> 		//	cmn.mtu = setLocalMtu()
> 		cmn.Mtu = 1500
> 		pathStr = "None"
> ```

Above is left over from debugging, will be removed.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

kormat:
Reviewed 9 of 15 files at r2.
Review status: 11 of 17 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[c/lib/scion/packet.c, line 433 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjiUE9PtFJglfj0Zs:-L9AjiUF7-D4UFNJDdLi:b-k9oq5p) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L433)):*
> ```C
> }
> 
> struct extn_h {
> ```

I think this needs to be packed in order to be portable.

This should also be in `extensions.h`.

---

*[c/lib/scion/packet.c, line 447 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjcBR-l_2gpwKVVVA:-L9AjcBSA5YJn7RDtTVI:bj30dgh) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L447)):*
> ```C
>     uint16_t total_len = ntohs(spkt.sch->total_len);
>     uint16_t hdr_len = spkt.sch->header_len * LINE_LEN;
>     if (nh != HOP_BY_HOP || total_len < hdr_len + LINE_LEN) {
> ```

Why `+ LINE_LEN`?

---

*[go/border/error.go, line 102 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AlVMy2UJ92vEiWSc5:-L9AlVMy2UJ92vEiWSc6:b-ulozre) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/border/error.go#L102)):*
> ```Go
> 		return
> 	}
> 	rp.Release()
> ```

This will only get called if no errors happens, but it should be called unconditionally. You can either have `defer rp.Release()` at the top of this function, or just `args.rp.Release()` in the loop in `PacketError`.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

kormat:
Reviewed 5 of 15 files at r2.
Review status: 16 of 17 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[go/tools/scmp/main.go, line 63 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B33s6EkUVz405G1_3:-L9B33s6EkUVz405G1_4:bop4pby) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L63)):*
> ```Go
> 	// Initialize default SCION networking context
> 	if err := snet.Init(cmn.Local.IA, *sciondPath, *dispatcher); err != nil {
> 		//		cmn.Fatal("Unable to initialize SCION network\nerr=%v", err)
> ```

Uncomment?

---

*[go/tools/scmp/main.go, line 71 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B1U22CKI2kkFRE7oC:-L9B1U22CKI2kkFRE7oD:bftyao1) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L71)):*
> ```Go
> 		bindAddress = &reliable.AppAddr{Addr: cmn.Bind.Host}
> 	}
> 	cmn.Conn, _, err = reliable.Register(*dispatcher, cmn.Local.IA, address, bindAddress, addr.SvcNone)
> ```

Line length.

---

*[go/tools/scmp/main.go, line 73 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B3EM0AwqqQil7h76p:-L9B3EM0AwqqQil7h76q:b-6vtqul) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L73)):*
> ```Go
> 	cmn.Conn, _, err = reliable.Register(*dispatcher, cmn.Local.IA, address, bindAddress, addr.SvcNone)
> 	if err != nil {
> 		cmn.Fatal("Unable to register with the dispatcher addr=%s\nerr=%v", cmn.Local.String(), err)
> ```

This can be just `cmn.Local`, as it implements `fmt.Stringer`.

---

*[go/tools/scmp/main.go, line 83 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B3aHO3cqNwBx39j48:-L9B3aHO3cqNwBx39j49:bop4pby) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L83)):*
> ```Go
> 		pathStr = cmn.PathEntry.Path.String()
> 	} else {
> 		//	cmn.mtu = setLocalMtu()
> ```

Uncomment?

---

*[go/tools/scmp/main.go, line 106 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B3Qam2pwC1QJg2IAi:-L9B3Qam2pwC1QJg2IAj:b-3lkx6a) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L106)):*
> ```Go
> 		recordpath.Run()
> 	default:
> 		fmt.Fprintf(os.Stderr, "ERROR: Invalid command\n")
> ```

This should include the supplied command.

---

*[go/tools/scmp/cmn/common.go, line 34 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B2mKt5Nub4iubI28K:-L9B2mKt5Nub4iubI28L:b3x2ite) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/cmn/common.go#L34)):*
> ```Go
> 
> var (
> 	Interactive bool
> ```

I'd separate out the flag vars from the rest.


```
var (
   // non-flag vars
)

var (
  // flag vars
)

func init() {
  // set up flag vars
}
```

---

*[go/tools/scmp/cmn/common.go, line 91 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B20Gs2-WZZMaPvzCQ:-L9B20Gs2-WZZMaPvzCR:b5pvfoo) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/cmn/common.go#L91)):*
> ```Go
> 	}
> 	if Interval == 0 {
> 		Interval = 1
> ```

This is 1 nanosecond - is that the intention?

---

*[go/tools/scmp/echo/echo.go, line 38 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B66Kl8LlmyHi2kEDD:-L9B66Kl8LlmyHi2kEDE:bnhfb1l) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L38)):*
> ```Go
> 
> func Run() {
> ```

Blank line.

---

*[go/tools/scmp/echo/echo.go, line 44 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B6ElTA2DEJQHiboOJ:-L9B6ElTA2DEJQHiboOK:b-h8th06) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L44)):*
> ```Go
> 	go sendPkts()
> 
> 	wg.Wait()
> ```

There's no need for a wait group:
```
go sendPkts()
recvPkts()
```

---

*[go/tools/scmp/echo/echo.go, line 55 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B6rJf0o7NSY2-I7Am:-L9B6rJf0o7NSY2-I7An:bi5etpa) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L55)):*
> ```Go
> 
> 	b := make(common.RawBytes, cmn.Mtu)
> ```

The blank line usage here is a bit arbitrary. If you want to separate setup from running, i would put info/pkt/b/nhAddr in a block, and then nextPktTS/ticker/for in another.

---

*[go/tools/scmp/echo/echo.go, line 109 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B7i05CYrI8MMZ7Hud:-L9B7i0660EYnQYOTWLZ:b-rso5hd) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L109)):*
> ```Go
> 		if err != nil {
> 			e, ok := err.(*net.OpError)
> 			if ok && e.Timeout() {
> ```

This can be simplified to `if common.IsTimeoutErr(err) { continue } else { ... }`

---

*[go/tools/scmp/echo/echo.go, line 153 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B8LoJEjDdkgoZqPqv:-L9B8LoJEjDdkgoZqPqw:b-t526l1) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L153)):*
> ```Go
> 			common.NewBasicError("Not an Info Echo", nil, "type", common.TypeOf(info))
> 	}
> 	return scmpHdr, info, nil
> ```

This should also validate that the echo Id and Seq are correct.

---

*[go/tools/scmp/recordpath/recordpath.go, line 42 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B8gL06C4QTHSDPNFu:-L9B8gL06C4QTHSDPNFv:bnhfb1l) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/recordpath/recordpath.go#L42)):*
> ```Go
> 	info := &scmp.InfoRecordPath{Id: cmn.Rand(), Entries: entries}
> 	pkt := cmn.NewSCMPPkt(scmp.T_G_RecordPathRequest, info, ext)
> ```

Blank line.

---

*[go/tools/scmp/recordpath/recordpath.go, line 118 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B9BPhE7LTjmbUWFWS:-L9B9BPhE7LTjmbUWFWT:b-mcd4pr) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/recordpath/recordpath.go#L118)):*
> ```Go
> 		return nil, nil,
> 			common.NewBasicError("Not an Info RecordPath", nil, "type", common.TypeOf(info))
> 	}
> ```

This should also check the req Id.

---

*[go/tools/scmp/traceroute/traceroute.go, line 46 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B9JYh6Y8VtWmm-brf:-L9B9JYh6Y8VtWmm-brg:btc45ku) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/traceroute/traceroute.go#L46)):*
> ```Go
> 		ext = &scmp.Extn{Error: false, HopByHop: true}
> 	}
> 	// Send packet
> ```

Old comment?

---

*[go/tools/scmp/traceroute/traceroute.go, line 49 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B9MAg-Tc7qjCHxn2n:-L9B9MAg-Tc7qjCHxn2o:bnhfb1l) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/traceroute/traceroute.go#L49)):*
> ```Go
> 	info := &scmp.InfoTraceRoute{Id: cmn.Rand(), HopOff: hopOff}
> 	pkt := cmn.NewSCMPPkt(scmp.T_G_TraceRouteRequest, info, ext)
> ```

Blank line.

---

*[go/tools/scmp/traceroute/traceroute.go, line 55 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B9ikS8ZkxzISxwtqf:-L9B9ikS8ZkxzISxwtqg:bxi47j4) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/traceroute/traceroute.go#L55)):*
> ```Go
> 	ts := time.Now()
> 	ticker := time.NewTicker(cmn.Interval)
> 	for ; true; ts = <-ticker.C {
> ```

This shouldn't be on a ticker. It should send as soon as either the previous request gets a reply, or the request times out.

---

*[go/tools/scmp/traceroute/traceroute.go, line 79 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B9Z0m6w0S0CJozRAD:-L9B9Z0m6w0S0CJozRAE:bxiy2ix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/traceroute/traceroute.go#L79)):*
> ```Go
> 		if err != nil {
> 			e, ok := err.(*net.OpError)
> 			if ok && e.Timeout() {
> ```

`common.IsTimeoutErr()`

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

kormat:
Review status: 16 of 17 files reviewed at latest revision, 25 unresolved discussions, some commit checks failed.

---

*[go/tools/scmp/traceroute/traceroute.go, line 33 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9EplySB9H2LjikuuOb:-L9EplySB9H2LjikuuOc:bg5a1rn) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/traceroute/traceroute.go#L33)):*
> ```Go
> )
> 
> func Run() {
> ```

There are a few issues with the logic and behaviour here:
- a hop that doesn't respond will cause an infinite loop (the timeout->continue branch is unconditional)
- if there are any timeouts then it's possible for the loop to not terminate either (`cmd.Stats.Sent == total` will fail to match if the timeouts happen near the end of the run, as `cmd.Stats.Sent` can end up more than `total`)
- There's no indication to the user that there _was_ a timeout

If you look at classic IP traceroute, it sends 3 packets to every hop, and prints either the rtt if it gets a response, or a `*` if it times out. E.g.:
```
...
 8  72.14.195.4 (72.14.195.4)  1.330 ms  1.320 ms  1.307 ms
 9  209.85.247.126 (209.85.247.126)  1.522 ms 209.85.250.150 (209.85.250.150)  1.561 ms 209.85.243.127 (209.85.243.127)  1.539 ms
10  * google-public-dns-a.google.com (8.8.8.8)  1.221 ms  1.219 ms
```

We don't need to worry about hops changing identity as in the IP example above, as in SCION the path is pre-decided, but it would be good to emulate the other behaviour.

A simple way to test timeout behaviour is to do something like this:
```
diff --git a/go/border/rpkt/process.go b/go/border/rpkt/process.go
index 5411529..f62ea06 100644
--- a/go/border/rpkt/process.go
+++ b/go/border/rpkt/process.go
@@ -18,6 +18,7 @@ package rpkt
 
 import (
        "fmt"
+       "math/rand"
        "time"
 
        //log "github.com/inconshreveable/log15"
@@ -252,6 +253,9 @@ func (rp *RtrPkt) processSCMPTraceRoute() error {
        if infoTrace.HopOff != rp.CmnHdr.CurrHopF {
                return nil
        }
+       if rand.Int()%2 != 0 {
+               return common.NewBasicError("Randomly not replying to traceroute", nil)
+       }
        // If In is set and the packet came from the local AS,
        // or if In is false and the packet came from outside,
        // then stop processing.
```

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

sgmonroy:
Review status: 16 of 17 files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[c/lib/scion/packet.c, line 433 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjiUE9PtFJglfj0Zs:-L9FlAqTA87iWDCgt69Z:b-u5taz) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L433)):*

I think this needs to be packed in order to be portable.

This should also be in `extensions.h`.
</blockquote></details>

I don't think it applies in this case. All the fields are 1 byte each, there is no padding 

---

*[c/lib/scion/packet.c, line 447 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjcBR-l_2gpwKVVVA:-L9FlgpY-S47-iX1ebgs:b-by9xy7) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L447)):*

Why `+ LINE_LEN`?
</blockquote></details>

LINE_LEN would be the extension header itself, right? otherwise it cannot be an extension header
I think I got that from somewhere else, cannot remember now where from exactly (from our source code)

---

*[go/border/error.go, line 102 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AlVMy2UJ92vEiWSc5:-L9Fms005nCk_zp6K7DH:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/border/error.go#L102)):*

This will only get called if no errors happens, but it should be called unconditionally. You can either have `defer rp.Release()` at the top of this function, or just `args.rp.Release()` in the loop in `PacketError`.
</blockquote></details>

Done.

---

*[go/tools/scmp/main.go, line 63 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B33s6EkUVz405G1_3:-L9Fn1ddAGlmzjVqXX46:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L63)):*

Uncomment?
</blockquote></details>

Done.

---

*[go/tools/scmp/main.go, line 71 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B1U22CKI2kkFRE7oC:-L9FnAfwDSy1xNlZOEST:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L71)):*

Line length.
</blockquote></details>

Done.

---

*[go/tools/scmp/main.go, line 73 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B3EM0AwqqQil7h76p:-L9FnFr_1Jb03eQH61Fj:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L73)):*

This can be just `cmn.Local`, as it implements `fmt.Stringer`.
</blockquote></details>

Done.

---

*[go/tools/scmp/main.go, line 83 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B3aHO3cqNwBx39j48:-L9FnI_MCj3t6zjqqRlR:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L83)):*

Uncomment?
</blockquote></details>

Done.

---

*[go/tools/scmp/main.go, line 85 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AQYUl--iSD17PdnoR:-L9FnIyo3qN7ALNZa96F:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L85)):*

Above is left over from debugging, will be removed.
</blockquote></details>

Done.

---

*[go/tools/scmp/main.go, line 106 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B3Qam2pwC1QJg2IAi:-L9FnQ8o-Mq7orWAc9-P:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/main.go#L106)):*

This should include the supplied command.
</blockquote></details>

Done.

---

*[go/tools/scmp/cmn/common.go, line 34 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B2mKt5Nub4iubI28K:-L9Fo2qr2YM_Wumm1X-l:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/cmn/common.go#L34)):*

I'd separate out the flag vars from the rest.


```
var (
   // non-flag vars
)

var (
  // flag vars
)

func init() {
  // set up flag vars
}
```
</blockquote></details>

Done.

---

*[go/tools/scmp/cmn/common.go, line 91 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B20Gs2-WZZMaPvzCQ:-L9Fo3zw84NfcN8LKQZS:b-sgo2pf) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/cmn/common.go#L91)):*

This is 1 nanosecond - is that the intention?
</blockquote></details>

Yes, ticker interval needs to be greater than 0

---

*[go/tools/scmp/echo/echo.go, line 38 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B66Kl8LlmyHi2kEDD:-L9FoY9l26uPln9SO6Pq:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L38)):*

Blank line.
</blockquote></details>

Done.

---

*[go/tools/scmp/echo/echo.go, line 44 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B6ElTA2DEJQHiboOJ:-L9Foimj4TtxT0wjABlk:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L44)):*

There's no need for a wait group:
```
go sendPkts()
recvPkts()
```
</blockquote></details>

Done.

---

*[go/tools/scmp/echo/echo.go, line 55 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B6rJf0o7NSY2-I7Am:-L9FovB61NnKHq4Lp5A4:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L55)):*

The blank line usage here is a bit arbitrary. If you want to separate setup from running, i would put info/pkt/b/nhAddr in a block, and then nextPktTS/ticker/for in another.
</blockquote></details>

Done.

---

*[go/tools/scmp/echo/echo.go, line 109 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B7i05CYrI8MMZ7Hud:-L9FpGtDBSsC4BtW_0hJ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L109)):*

This can be simplified to `if common.IsTimeoutErr(err) { continue } else { ... }`
</blockquote></details>

Done.

---

*[go/tools/scmp/echo/echo.go, line 153 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B8LoJEjDdkgoZqPqv:-L9FpJaDANN9ExpZXNFW:b-xhkbp3) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L153)):*

This should also validate that the echo Id and Seq are correct.
</blockquote></details>

The ID has to be correct, otherwise we would not get the packet at all.
I have not change the logic from what we had before,  echo is not keeping track of seq etc not sure what you want to accomplish by checking that?

---

*[go/tools/scmp/recordpath/recordpath.go, line 42 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B8gL06C4QTHSDPNFu:-L9FpwI81f39n_2ze0VY:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/recordpath/recordpath.go#L42)):*

Blank line.
</blockquote></details>

Done.

---

*[go/tools/scmp/recordpath/recordpath.go, line 118 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B9BPhE7LTjmbUWFWS:-L9Fpw_f9ssSiCU-x3Bz:b-ctrph2) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/recordpath/recordpath.go#L118)):*

This should also check the req Id.
</blockquote></details>

See above comment

---

*[go/tools/scmp/traceroute/traceroute.go, line 33 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9EplySB9H2LjikuuOb:-L9FkpcP3GA_h5F2SENh:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/traceroute/traceroute.go#L33)):*

There are a few issues with the logic and behaviour here:
- a hop that doesn't respond will cause an infinite loop (the timeout->continue branch is unconditional)
- if there are any timeouts then it's possible for the loop to not terminate either (`cmd.Stats.Sent == total` will fail to match if the timeouts happen near the end of the run, as `cmd.Stats.Sent` can end up more than `total`)
- There's no indication to the user that there _was_ a timeout

If you look at classic IP traceroute, it sends 3 packets to every hop, and prints either the rtt if it gets a response, or a `*` if it times out. E.g.:
```
...
 8  72.14.195.4 (72.14.195.4)  1.330 ms  1.320 ms  1.307 ms
 9  209.85.247.126 (209.85.247.126)  1.522 ms 209.85.250.150 (209.85.250.150)  1.561 ms 209.85.243.127 (209.85.243.127)  1.539 ms
10  * google-public-dns-a.google.com (8.8.8.8)  1.221 ms  1.219 ms
```

We don't need to worry about hops changing identity as in the IP example above, as in SCION the path is pre-decided, but it would be good to emulate the other behaviour.

A simple way to test timeout behaviour is to do something like this:
```
diff --git a/go/border/rpkt/process.go b/go/border/rpkt/process.go
index 5411529..f62ea06 100644
--- a/go/border/rpkt/process.go
+++ b/go/border/rpkt/process.go
@@ -18,6 +18,7 @@ package rpkt
 
 import (
        "fmt"
+       "math/rand"
        "time"
 
        //log "github.com/inconshreveable/log15"
@@ -252,6 +253,9 @@ func (rp *RtrPkt) processSCMPTraceRoute() error {
        if infoTrace.HopOff != rp.CmnHdr.CurrHopF {
                return nil
        }
+       if rand.Int()%2 != 0 {
+               return common.NewBasicError("Randomly not replying to traceroute", nil)
+       }
        // If In is set and the packet came from the local AS,
        // or if In is false and the packet came from outside,
        // then stop processing.
```
</blockquote></details>

Done.

---

*[go/tools/scmp/traceroute/traceroute.go, line 46 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B9JYh6Y8VtWmm-brf:-L9Fqo6CBRNORDWeOWNl:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/traceroute/traceroute.go#L46)):*

Old comment?
</blockquote></details>

Done.

---

*[go/tools/scmp/traceroute/traceroute.go, line 49 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B9MAg-Tc7qjCHxn2n:-L9FqC7i6GeRJdzxKjds:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/traceroute/traceroute.go#L49)):*

Blank line.
</blockquote></details>

Done.

---

*[go/tools/scmp/traceroute/traceroute.go, line 55 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B9ikS8ZkxzISxwtqf:-L9FqD9oDRD7vzxcAqfU:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/traceroute/traceroute.go#L55)):*

This shouldn't be on a ticker. It should send as soon as either the previous request gets a reply, or the request times out.
</blockquote></details>

Done.

---

*[go/tools/scmp/traceroute/traceroute.go, line 79 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B9Z0m6w0S0CJozRAD:-L9FqmsmBQn8HDyJ80DX:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/traceroute/traceroute.go#L79)):*

`common.IsTimeoutErr()`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

kormat:
Reviewed 1 of 15 files at r2.
Review status: 11 of 17 files reviewed at latest revision, 22 unresolved discussions.

---

*[c/lib/scion/packet.c, line 433 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjiUE9PtFJglfj0Zs:-L9FrbwK4DODry5jZJsH:b-8z1q5z) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L433)):*

I don't think it applies in this case. All the fields are 1 byte each, there is no padding 
</blockquote></details>

Is that guaranteed by the C standard?

---

*[c/lib/scion/packet.c, line 447 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjcBR-l_2gpwKVVVA:-L9FrmDjDO7aR9UvnfhH:bxmnyoz) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L447)):*

LINE_LEN would be the extension header itself, right? otherwise it cannot be an extension header
I think I got that from somewhere else, cannot remember now where from exactly (from our source code)
</blockquote></details>

But you're trying to detect the _absense_ of a hbh extension. There are 3 possible cases here:

- nh != HBH - there's an e2e or l4 header next.
- nh == HBH and total_len == hdr_len: there are no extensions _or_ l4 header
- nh == HBH and total_len != hdr_len: there are hbh extensions.

`nh != HOP_BY_HOP` covers the first case, but the second case needs a `total_len == hdr_len` to be correct. (This is ignoring a malformed packet where total_len could be < hdr_len, but we have ~zero error handling in this library anyway, so there's no real way to deal with that :(  )

---

*[go/tools/scmp/echo/echo.go, line 153 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B8LoJEjDdkgoZqPqv:-L9FrDc80OYCWzk8WvjJ:b-avc6eq) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L153)):*
> The ID has to be correct, otherwise we would not get the packet at all.

This assumes there isn't a bug in the dispatcher, which i would prefer not to assume :)

Checking seq means you can tell if packets are reordered, or if a request that has timed out gets a response anyway.

I'm ok not to fix these issues in this PR, if you want to put a FIXME in instead.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

sgmonroy:
Review status: 11 of 18 files reviewed at latest revision, 22 unresolved discussions.

---

*[c/lib/scion/packet.c, line 433 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjiUE9PtFJglfj0Zs:-L9JXR2W6BHD-4sJjzrI:bpj2bea) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L433)):*

Is that guaranteed by the C standard?
</blockquote></details>

AFAIK the standard does not guaranty anything regarding internal struct padding, hence why compilers are free to insert padding for performance/alignment reasons. That said, the padding is determined by self-alignment of basic types rules and given that all members of this structs are 1B there is no alignment/padding needed as all fields are self-aligned to 1B.

---

*[c/lib/scion/packet.c, line 447 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjcBR-l_2gpwKVVVA:-L9FxW1bCIrET4ApsaEm:b-t9i9wc) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L447)):*

But you're trying to detect the _absense_ of a hbh extension. There are 3 possible cases here:

- nh != HBH - there's an e2e or l4 header next.
- nh == HBH and total_len == hdr_len: there are no extensions _or_ l4 header
- nh == HBH and total_len != hdr_len: there are hbh extensions.

`nh != HOP_BY_HOP` covers the first case, but the second case needs a `total_len == hdr_len` to be correct. (This is ignoring a malformed packet where total_len could be < hdr_len, but we have ~zero error handling in this library anyway, so there's no real way to deal with that :(  )
</blockquote></details>

I still don't see why the current logic is wrong. If for whatever reason we get such a packet, it is malformed, so why not drop it?

---

*[go/tools/scmp/echo/echo.go, line 153 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9B8LoJEjDdkgoZqPqv:-L9Ja6al2WylF5i0Abcj:bt6qiti) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/go/tools/scmp/echo/echo.go#L153)):*

> The ID has to be correct, otherwise we would not get the packet at all.

This assumes there isn't a bug in the dispatcher, which i would prefer not to assume :)

Checking seq means you can tell if packets are reordered, or if a request that has timed out gets a response anyway.

I'm ok not to fix these issues in this PR, if you want to put a FIXME in instead.
</blockquote></details>

I have added the ID check, although I don't agree with your reason. To catch this kind of bugs you use a test bench instead of spurious checks in different applications/tools.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

kormat:
Reviewed 7 of 7 files at r3.

---

*[go/lib/scmp/hdr.go, line 135 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9LJGuO7k74adCqTDlD:-L9LJGuPDltcY2pNCG7Q:b-44s6vc) ([raw file](https://github.com/scionproto/scion/blob/9cbf9ab06dd049112bc2f2a57105b5072466d3d8/go/lib/scmp/hdr.go#L135)):*
> ```Go
> 
> func (h *Hdr) SetTime(t time.Time) {
> 	h.Timestamp = uint64(t.Truncate(time.Microsecond).UnixNano()) / 1000
> ```

The `Truncate` isn't necessary, the `/ 1000` will do that anyway.

---

*[go/tools/scmp/echo/echo.go, line 166 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9LMtxqBbmpNL6J1WXO:-L9LMtxqBbmpNL6J1WXP:b-n8w0w9) ([raw file](https://github.com/scionproto/scion/blob/9cbf9ab06dd049112bc2f2a57105b5072466d3d8/go/tools/scmp/echo/echo.go#L166)):*
> ```Go
> 	}
> 	expectedTS = expectedTS.Truncate(time.Microsecond)
> 	if pktTS.Before(expectedTS) {
> ```

This logic doesn't look right. From a test:
```
176 bytes from 2-26,[127.0.0.2] scmp_seq=1 time=48.356ms
176 bytes from 2-26,[127.0.0.2] scmp_seq=2 time=1.013506s
176 bytes from 2-26,[127.0.0.2] scmp_seq=3 time=1.02184s
176 bytes from 2-26,[127.0.0.2] scmp_seq=5 time=1.047638s  Out of Order
176 bytes from 2-26,[127.0.0.2] scmp_seq=6 time=1.077863s  Out of Order
176 bytes from 2-26,[127.0.0.2] scmp_seq=7 time=1.093544s  Out of Order
```
Those last 3 packets _aren't_ out of order. I think you have to track the last received seq number, and check if the current seq is > than it. If it's less, then it's an out of order packet.

---

*[go/tools/scmp/traceroute/traceroute.go, line 109 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9LP0GvE6f55mumYPHD:-L9LP0GvE6f55mumYPHE:bchk9jx) ([raw file](https://github.com/scionproto/scion/blob/9cbf9ab06dd049112bc2f2a57105b5072466d3d8/go/tools/scmp/traceroute/traceroute.go#L109)):*
> ```Go
> 		if err != nil {
> 			fmt.Fprintf(os.Stderr, "ERROR: SCMP validation error: %v\n", err)
> 			// We continue as one bad reply does not mean all replies would be bad ones
> ```

Out-dated comment.

---

*[go/tools/scmp/traceroute/traceroute.go, line 134 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9LPMrA9rI6erCVtBE5:-L9LPMrA9rI6erCVtBE6:ba3dfna) ([raw file](https://github.com/scionproto/scion/blob/9cbf9ab06dd049112bc2f2a57105b5072466d3d8/go/tools/scmp/traceroute/traceroute.go#L134)):*
> ```Go
> 		fmt.Printf(" *")
> 	} else {
> 		//	fmt.Printf("DBG: itmeout %v, rtt %v\n", cmn.Timeout, rtt)
> ```

Remove?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

sgmonroy:
Review status: 13 of 19 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[go/lib/scmp/hdr.go, line 135 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9LJGuO7k74adCqTDlD:-L9LQu_k5PlxbDUoI1U_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9cbf9ab06dd049112bc2f2a57105b5072466d3d8/go/lib/scmp/hdr.go#L135)):*

The `Truncate` isn't necessary, the `/ 1000` will do that anyway.
</blockquote></details>

Done.

---

*[go/tools/scmp/echo/echo.go, line 166 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9LMtxqBbmpNL6J1WXO:-L9LV1wU5PXlmAbhYDaz:b-8zy032) ([raw file](https://github.com/scionproto/scion/blob/9cbf9ab06dd049112bc2f2a57105b5072466d3d8/go/tools/scmp/echo/echo.go#L166)):*

This logic doesn't look right. From a test:
```
176 bytes from 2-26,[127.0.0.2] scmp_seq=1 time=48.356ms
176 bytes from 2-26,[127.0.0.2] scmp_seq=2 time=1.013506s
176 bytes from 2-26,[127.0.0.2] scmp_seq=3 time=1.02184s
176 bytes from 2-26,[127.0.0.2] scmp_seq=5 time=1.047638s  Out of Order
176 bytes from 2-26,[127.0.0.2] scmp_seq=6 time=1.077863s  Out of Order
176 bytes from 2-26,[127.0.0.2] scmp_seq=7 time=1.093544s  Out of Order
```
Those last 3 packets _aren't_ out of order. I think you have to track the last received seq number, and check if the current seq is > than it. If it's less, then it's an out of order packet.
</blockquote></details>

Yeah, it is easier to just keep track of the last seq. 

---

*[go/tools/scmp/traceroute/traceroute.go, line 109 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9LP0GvE6f55mumYPHD:-L9LQSze0KRxnmcb4_yx:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9cbf9ab06dd049112bc2f2a57105b5072466d3d8/go/tools/scmp/traceroute/traceroute.go#L109)):*

Out-dated comment.
</blockquote></details>

Done.

---

*[go/tools/scmp/traceroute/traceroute.go, line 134 at r3](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9LPMrA9rI6erCVtBE5:-L9LQZ2h9DHK4Guyzw1B:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9cbf9ab06dd049112bc2f2a57105b5072466d3d8/go/tools/scmp/traceroute/traceroute.go#L134)):*

Remove?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

kormat:
Reviewed 6 of 6 files at r4.

---

*[c/lib/scion/packet.c, line 447 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjcBR-l_2gpwKVVVA:-L9eKRlg4Mqpkc5qTV2B:b-63ci7v) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L447)):*

I still don't see why the current logic is wrong. If for whatever reason we get such a packet, it is malformed, so why not drop it?
</blockquote></details>

Silently dropping malformed packets is not desirable behaviour, and this isn't the right place to check for that in any case.

---

*[go/tools/scmp/echo/echo.go, line 96 at r4](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9eEGcs061YaghOdukk:-L9eEGcs061YaghOdukl:b-qt2nq9) ([raw file](https://github.com/scionproto/scion/blob/51073446a6c98efa9850e4885d1e2752a061dd15/go/tools/scmp/echo/echo.go#L96)):*
> ```Go
> 		if err != nil {
> 			if common.IsTimeoutErr(err) {
> 				fmt.Printf("Timeout on scmp_seq=%d\n", expectedSeq)
> ```

I don't think we need this. It also makes the output confusing in some cases:
```
176 bytes from 2-4_295_002_022,[127.0.0.2] scmp_seq=0 time=3.578ms
176 bytes from 2-4_295_002_022,[127.0.0.2] scmp_seq=2 time=2.974ms
Timeout on scmp_seq=1
176 bytes from 2-4_295_002_022,[127.0.0.2] scmp_seq=4 time=3.382ms
Timeout on scmp_seq=3
```

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

sgmonroy:

---

*[c/lib/scion/packet.c, line 447 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjcBR-l_2gpwKVVVA:-L9eMJWG516MH0wFFHwS:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L447)):*

Silently dropping malformed packets is not desirable behaviour, and this isn't the right place to check for that in any case.
</blockquote></details>

Done.

---

*[go/tools/scmp/echo/echo.go, line 96 at r4](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9eEGcs061YaghOdukk:-L9eOBplDnAWGZOoMCxD:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/51073446a6c98efa9850e4885d1e2752a061dd15/go/tools/scmp/echo/echo.go#L96)):*

I don't think we need this. It also makes the output confusing in some cases:
```
176 bytes from 2-4_295_002_022,[127.0.0.2] scmp_seq=0 time=3.578ms
176 bytes from 2-4_295_002_022,[127.0.0.2] scmp_seq=2 time=2.974ms
Timeout on scmp_seq=1
176 bytes from 2-4_295_002_022,[127.0.0.2] scmp_seq=4 time=3.382ms
Timeout on scmp_seq=3
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

kormat:
Reviewed 2 of 2 files at r5.

---

*[c/lib/scion/packet.c, line 447 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1454#-L9AjcBR-l_2gpwKVVVA:-L9eQSJZ6iPgYqx355uo:b-49dwol) ([raw file](https://github.com/scionproto/scion/blob/d51fbd350b70551128aed16f765bf523ceebca69/c/lib/scion/packet.c#L447)):*

Done.
</blockquote></details>

The correct check is: `if (nh != HOP_BY_HOP || total_len <= hdr_len) {`, otherwise a packet with no extensions or l4 header (which is technically valid, just.. unusual) will be treated as having a hbh extension.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1454)*

PR #1455, opened 2018-03-16
Package for building Go apps during tests

scrye:
For tests that cover more functionality than unit tests we currently have to resort to bash scripts, or to have the SCION test infrastructure running. This either (1) does not integrate with our unit testing framework, (2) is more brittle and (3) requires more busywork/is harder to automate.

This PR provides basic building blocks for running Go applications from within the Go testing framework. We can use this in the future to test utilities (like the `scion-pki` tool) and servers (like `SCIOND`) while still using the same patterns and environment we run in our unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1455)


shitz:
Reviewed 1 of 3 files at r1.
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[go/lib/xtest/loader/binary.go, line 59 at r1](https://reviewable.io:443/reviews/scionproto/scion/1455#-L81UIm5fjJplVyvLolu:-L81UIm5fjJplVyvLolv:bf7esnx) ([raw file](https://github.com/scionproto/scion/blob/c3d87115e35fbef8537025b1777ff50d552448f4/go/lib/xtest/loader/binary.go#L59)):*
> ```Go
> // Build compiles package target, and saves the resulting binary in directory
> // dir with a file name starting with prefix, and a randomly generated suffix.
> func Build(target, dir, prefix string) string {
> ```

This does not allow for any build flags to be passed in. I suggest you add build flags to `Binary` and have this method take an additional parameter.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1455)*

scrye:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/lib/xtest/loader/binary.go, line 59 at r1](https://reviewable.io:443/reviews/scionproto/scion/1455#-L81UIm5fjJplVyvLolu:-L831mwCPclQRopKxl8W:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c3d87115e35fbef8537025b1777ff50d552448f4/go/lib/xtest/loader/binary.go#L59)):*

This does not allow for any build flags to be passed in. I suggest you add build flags to `Binary` and have this method take an additional parameter.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1455)*

shitz:
Reviewed 1 of 2 files at r2.
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/lib/xtest/loader/binary.go, line 45 at r2](https://reviewable.io:443/reviews/scionproto/scion/1455#-L8DJ750s-8f1H50I-Y9:-L8DJ750s-8f1H50I-YA:b-9m4fyp) ([raw file](https://github.com/scionproto/scion/blob/53a4e5b673ddcd7ab31d4258fe2a9fab521558ed/go/lib/xtest/loader/binary.go#L45)):*
> ```Go
> 	if name, err := Build(b.Target, b.Dir, b.Prefix, b.BuildFlags...); err != nil {
> 		return err
> 	} else {
> ```

I really prefer when there code adheres to the so called 'line-of-sight of code', i.e., remove the `else`. See https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 for more info. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1455)*

scrye:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/lib/xtest/loader/binary.go, line 45 at r2](https://reviewable.io:443/reviews/scionproto/scion/1455#-L8DJ750s-8f1H50I-Y9:-L8HpfQWOIs8ZH-m8h47:bbkqzvz) ([raw file](https://github.com/scionproto/scion/blob/53a4e5b673ddcd7ab31d4258fe2a9fab521558ed/go/lib/xtest/loader/binary.go#L45)):*

I really prefer when there code adheres to the so called 'line-of-sight of code', i.e., remove the `else`. See https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 for more info. 
</blockquote></details>

I generally adhere to what the article is advocating; the only time I choose not to do that is if it interferes with my preference for aggressive scoping (e.g., when I need to declare some variable in the function's top scope, even though its lifetime is only relevant for an if/else block). It's something I've also seen on the Golang slack as causing quite a lot of debate :-).

In this case, since the function is very short so there is no risk of the top-level scope getting polluted, I think your suggestion is cleaner. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1455)*

shitz:
:lgtm:

---

Reviewed 6 of 8 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1455#-:-L8HvyUdT5EwdGuLw0ju:bnfp4nl)*

PR #1464, opened 2018-03-26
Change SIG to only use a single TUN device.

kormat:
This PR adds a new goroutine, egress.Reader. It is responsible for:
- reading packets from the TUN device
- consulting egress.NetMap to determine the correct ringbuf to use
- writing packets to the ringbuf

egress.Networks is a basic implementation of the egress.NetMapI
interface, which defines how an IP can be mapped to an allocation
belonging to a remote AS.

As the sig now only needs CAP_NET_ADMIN at start, capabilities are
cleared once the TUN device is opened and routes injected.

Also:
- Rename sig.go to main.go for consistency.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1464)


scrye:
Reviewed 12 of 12 files at r1.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[go/sig/egress/nets.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YRaE3AA9QPWvp5XsC:-L8YRaE3AA9QPWvp5XsD:b-lxueo) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L1)):*
> ```Go
> ```

The mixed IPv4/IPv6 routing table means prefixes of one type might be flagged as being part of the other, e.g., if I try `Test_Networks_Add` with `{[]string{"0.0.0.0/0", "::ffff:0000:0000/96"}, 2, true}`, it yields an error. This is due to `net`'s approach to interpreting `::ffff:0:0/96` IPv6 addresses as IPv4 addresses. I don't think this is a problem, but it's something I found interesting.

It also means lookup is slower than it needs to be, since the code tests IPv6 addresses for inclusion in IPv4 networks, even though there's no chance of a match.

---

*[go/sig/egress/nets.go, line 28 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YCZmP935PhjpVruSC:-L8YCZmP935PhjpVruSD:b-qq9gv3) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L28)):*
> ```Go
> var NetMap NetMapI = &Networks{}
> 
> type NetMapI interface {
> ```

I don't think an interface is needed here, `NetMap` can be used directly as a `*Networks` object.

---

*[go/sig/egress/nets.go, line 43 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YItHo-SIeAMcDKb4U:-L8YItHo-SIeAMcDKb4V:bunw3o7) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L43)):*
> ```Go
> 
> func (ns *Networks) Add(ipnet *net.IPNet, ia addr.IA, ring *ringbuf.Ring) error {
> 	cnet := newCanonNet(ipnet)
> ```

This can be moved after the sanity checks, since there's no point allocating it they fail at the start.

---

*[go/sig/egress/nets.go, line 53 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YFiYqAcalqd1fT78i:-L8YFiYqAcalqd1fT78j:b2rp9y) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L53)):*
> ```Go
> 	defer ns.m.Unlock()
> 	for i := range ns.nets {
> 		exnet := ns.nets[i]
> ```

Make `exnet` the same type as `ipnet` to simplify the code a bit and have them have the same type during logging (where arguably we don't care about the ring anyway):

```
exnet := ns.Nets.net
```

---

*[go/sig/egress/nets.go, line 54 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YEzFHDLsKrG6jQ0Og:-L8YEzFHDLsKrG6jQ0Oh:b-2twx4u) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L54)):*
> ```Go
> 	for i := range ns.nets {
> 		exnet := ns.nets[i]
> 		if exnet.net.Equal(cnet) {
> ```

These three checks can probably be simplified to something like:
```
if exnet.net.Contains(cnet.IP) || cnet.Contains(exnet.net.IP) {
    return common.NewBasicError("Networks overlap...")
}
```

The `if` should also be in its own function, to make the intent clearer.

---

*[go/sig/egress/nets.go, line 79 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YJEDVB8wukhenTVlR:-L8YJEDVB8wukhenTVlS:bps4bp6) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L79)):*
> ```Go
> 		return common.NewBasicError("Networks.Delete(): IPNet entry not present", nil, "net", ipnet)
> 	}
> 	// Fast delete, as it doesn't preserve order.
> ```

Nice trick!

It might improve clarity a bit to move the code to a helper function, so it looks like `ns.nets = del(ns.nets, idx)` (there are probably better names than my suggestion, though).

---

*[go/sig/egress/nets.go, line 88 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YGjPH3HjYX3D4TDe7:-L8YGjPH3HjYX3D4TDe8:b-p8cp7p) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L88)):*
> ```Go
> }
> 
> func (ns *Networks) Lookup(ip net.IP) (addr.IA, *ringbuf.Ring) {
> ```

Using a bool return value here is more idiomatic for searches in go.

---

*[go/sig/egress/nets.go, line 91 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YIYUq4dupjjTMiZb3:-L8YIYUs1dVGOO0zbFpP:b-6259is) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L91)):*
> ```Go
> 	ns.m.RLock()
> 	defer ns.m.RUnlock()
> 	for i := range ns.nets {
> ```

This can be simplified a bit:
```
for i, n := range ns.nets {
  if n.net.Contains...
}
```

---

*[go/sig/egress/nets.go, line 100 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YGvCL-d-sXqRtoBiR:-L8YGvCL-d-sXqRtoBiS:b-v0qswi) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L100)):*
> ```Go
> }
> 
> func (ns *Networks) getIdxL(cnet *canonNet) int {
> ```

Using an `ok bool` as the second return value is more idiomatic for go collections.

---

*[go/sig/egress/nets.go, line 115 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YDerPBfT66-VQAk2L:-L8YDerPBfT66-VQAk2M:bemnxlj) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L115)):*
> ```Go
> }
> 
> // canonNet contains a canonicalized version of net.IPNet, which allows it to be tested for equality.
> ```

I don't think a new type is needed for this (since it's still just an `*net.IPNet`). I think just having two helper functions (one for masking, one for equality) is enough.

---

*[go/sig/egress/nets_test.go, line 37 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YLRkT-TKciwqB-0tN:-L8YLRkT-TKciwqB-0tO:b-6vb7v0) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets_test.go#L37)):*
> ```Go
> )
> 
> func parseNet(t *testing.T, s string) *net.IPNet {
> ```

Mark this with `t.Helper()` s.t. it prints line and file information from the parent test (to pinpoint the failure more quickly).

---

*[go/sig/egress/nets_test.go, line 45 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YL8NZEa_Fu3Ji7SH0:-L8YL8NZEa_Fu3Ji7SH1:b-6vb7v0) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets_test.go#L45)):*
> ```Go
> }
> 
> func defNetworks(t *testing.T) *Networks {
> ```

Mark this with `t.Helper()` s.t. it prints line and file information from the parent test (to pinpoint the failure more quickly).

---

*[go/sig/egress/nets_test.go, line 137 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YMHIk1GIw9IguO8dl:-L8YMHIk1GIw9IguO8dm:bwvymww) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets_test.go#L137)):*
> ```Go
> 					SoMsg("Number of nets should have reduced",
> 						len(nets.nets), ShouldEqual, numNets-1)
> 					//for i := range nets.nets {
> ```

Delete commented code?

---

*[go/sig/egress/nets_test.go, line 204 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8bCHl4AI7xJTQ2Jedm:-L8bCHl4AI7xJTQ2Jedn:bmypwk0) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets_test.go#L204)):*
> ```Go
> 				netA := newCanonNet(parseNet(t, tc.netA))
> 				netB := newCanonNet(parseNet(t, tc.netB))
> 				if tc.ok {
> ```

This can be simplified a bit:
```
SoMsg("netA == netB", netA.Equal(netB), ShouldEqual, tc.ok)
SoMsg("netB == netA", netB.Equal(netA), ShouldEqual, tc.ok)
```

---

*[go/sig/egress/reader.go, line 105 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8bCvCp3JYBxfG8tWFB:-L8bCvCp3JYBxfG8tWFC:bgzi6x) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/reader.go#L105)):*
> ```Go
> 				continue
> 			}
> 			dstRing.Write(ringbuf.EntryList{buf}, true)
> ```

If the reader on the other end of this specific ringbuffer is very slow for some reason, this can block forever, taking the rest of the data plane forwarding down with it. In that case, it may be better to try a nonblocking write, and if it fails log the event and discard the packet.

---

*[go/sig/sigcmn/common.go, line 48 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8bDvaveyf0dOQanUHw:-L8bDvaveyf0dOQanUHx:b-ph5cp3) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/sigcmn/common.go#L48)):*
> ```Go
> 
> var (
> 	DefV4Net = &net.IPNet{IP: net.IPv4zero, Mask: make(net.IPMask, net.IPv4len)}
> ```

`Mask: net.CIDRMask(0, net.IPv4len)` hides internals a bit better, maybe.

---

*[go/sig/sigcmn/common.go, line 49 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8bEWTzP2OiznX3QOn0:-L8bEWTzP2OiznX3QOn1:b2nreh6) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/sigcmn/common.go#L49)):*
> ```Go
> var (
> 	DefV4Net = &net.IPNet{IP: net.IPv4zero, Mask: make(net.IPMask, net.IPv4len)}
> 	DefV6Net = &net.IPNet{IP: net.IPv6zero, Mask: make(net.IPMask, net.IPv6len)}
> ```

Same as above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1464)*

shitz:
Reviewed 10 of 12 files at r1.

---

*[go/sig/egress/nets.go, line 1 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8c12cINmptdEDWGMey:-L8c12cINmptdEDWGMez:bb6ht2i) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L1)):*
> ```Go
> ```

2018

---

*[go/sig/egress/nets.go, line 91 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8YIYUq4dupjjTMiZb3:-L8c3F0PkHiFsY5ay6Cy:bm72fmg) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L91)):*

This can be simplified a bit:
```
for i, n := range ns.nets {
  if n.net.Contains...
}
```
</blockquote></details>

No need for the `i` in this case.

---

*[go/sig/egress/nets.go, line 101 at r1](https://reviewable.io:443/reviews/scionproto/scion/1464#-L8c3Kt0IAPsSc1fMj8K:-L8c3Kt0IAPsSc1fMj8L:b-6yu8nc) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L101)):*
> ```Go
> 
> func (ns *Networks) getIdxL(cnet *canonNet) int {
> 	for i := range ns.nets {
> ```

`for i, n := range ns.nets { if n.net.Equal(cnet) { return i } }`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/scionproto/scion/1464)*

kormat:
Review status: 7 of 11 files reviewed at latest revision, 19 unresolved discussions.

---

*[go/sig/egress/nets.go, line 1 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8c12cINmptdEDWGMey:-L8gv5Gr0vwlizfkRH4L:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L1)):*

2018
</blockquote></details>

Done.

---

*[go/sig/egress/nets.go, line 1 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YRaE3AA9QPWvp5XsC:-L8gwAHR4_add8CbsBu5:bllpdb3) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L1)):*

The mixed IPv4/IPv6 routing table means prefixes of one type might be flagged as being part of the other, e.g., if I try `Test_Networks_Add` with `{[]string{"0.0.0.0/0", "::ffff:0000:0000/96"}, 2, true}`, it yields an error. This is due to `net`'s approach to interpreting `::ffff:0:0/96` IPv6 addresses as IPv4 addresses. I don't think this is a problem, but it's something I found interesting.

It also means lookup is slower than it needs to be, since the code tests IPv6 addresses for inclusion in IPv4 networks, even though there's no chance of a match.
</blockquote></details>

We could have separate v4 and v6 mappings, but then there's also the question of how to handle v4-in-v6 addresses, so i decided it was simpler to start with a single mapping at least.

---

*[go/sig/egress/nets.go, line 28 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YCZmP935PhjpVruSC:-L8gwXYn57RmpXOaD5_-:b-vuvvfv) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L28)):*

I don't think an interface is needed here, `NetMap` can be used directly as a `*Networks` object.
</blockquote></details>

Using an interface allows the implementation to be pluggable.

---

*[go/sig/egress/nets.go, line 43 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YItHo-SIeAMcDKb4U:-L8gwhXZ8B_lPjHmiC59:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L43)):*

This can be moved after the sanity checks, since there's no point allocating it they fail at the start.
</blockquote></details>

Done.

---

*[go/sig/egress/nets.go, line 53 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YFiYqAcalqd1fT78i:-L8gxH13EsLVT7FWt5gV:b-p7cydv) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L53)):*

Make `exnet` the same type as `ipnet` to simplify the code a bit and have them have the same type during logging (where arguably we don't care about the ring anyway):

```
exnet := ns.Nets.net
```
</blockquote></details>

(i'm not 100% sure what you mean, that code doesn't compile)

We probably don't care about the ring, but we do care about the network and the IA. Logging the `exnet` as-is gives us that. Otherwise you need two vars, and careful naming of the logging params. Changed the logging in any case for the next comment.

---

*[go/sig/egress/nets.go, line 54 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YEzFHDLsKrG6jQ0Og:-L8h-M6ZDyNLyd_umcUH:b7yorys) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L54)):*

These three checks can probably be simplified to something like:
```
if exnet.net.Contains(cnet.IP) || cnet.Contains(exnet.net.IP) {
    return common.NewBasicError("Networks overlap...")
}
```

The `if` should also be in its own function, to make the intent clearer.
</blockquote></details>

Simplified. I don't get what you mean about making the intent clearer though. I can add a comment, but it seems redundant.

---

*[go/sig/egress/nets.go, line 88 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YGjPH3HjYX3D4TDe7:-L8h-fHH6_ORym1C708l:b-rkh4us) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L88)):*

Using a bool return value here is more idiomatic for searches in go.
</blockquote></details>

As far as i can tell, that idiom is for when a key may be present but with a zero-valued value. That is explicitly disallowed in this type, hence i removed the `bool`s that i had initially used. Otherwise calling code should really test both the bool _and_ the ringbuf pointer, for no gain other than boilerplate.

---

*[go/sig/egress/nets.go, line 91 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YIYUq4dupjjTMiZb3:-L8h0DKe-ZWzrGWkKXw6:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L91)):*

No need for the `i` in this case.
</blockquote></details>

Done.

---

*[go/sig/egress/nets.go, line 100 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YGvCL-d-sXqRtoBiR:-L8h07YR0REqjPteCRTJ:bdkwe3s) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L100)):*

Using an `ok bool` as the second return value is more idiomatic for go collections.
</blockquote></details>

Using -1 when you're working on a linear "collection" is also widely used. There's a large number of functions in `strings` that do this, for example: https://golang.org/pkg/strings/#Index

---

*[go/sig/egress/nets.go, line 101 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8c3Kt0IAPsSc1fMj8K:-L8h0aAKBUPDk3vH9IcF:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L101)):*

`for i, n := range ns.nets { if n.net.Equal(cnet) { return i } }`
</blockquote></details>

Done.

---

*[go/sig/egress/nets.go, line 115 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YDerPBfT66-VQAk2L:-L8h0r210XapX6_YMK3-:b4r8hh9) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L115)):*

I don't think a new type is needed for this (since it's still just an `*net.IPNet`). I think just having two helper functions (one for masking, one for equality) is enough.
</blockquote></details>

The advantage of it is that you can make assumptions in the code that uses this type that if it's a `canonNet` it can rely on it having being masked. It just makes it a bit easier to reason about.

---

*[go/sig/egress/nets_test.go, line 37 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YLRkT-TKciwqB-0tN:-L8h2Kri7VVj8gcCSqFi:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets_test.go#L37)):*

Mark this with `t.Helper()` s.t. it prints line and file information from the parent test (to pinpoint the failure more quickly).
</blockquote></details>

Done.

---

*[go/sig/egress/nets_test.go, line 45 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YL8NZEa_Fu3Ji7SH0:-L8h2KYnEp0DPszNWl8l:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets_test.go#L45)):*

Mark this with `t.Helper()` s.t. it prints line and file information from the parent test (to pinpoint the failure more quickly).
</blockquote></details>

Done.

---

*[go/sig/egress/nets_test.go, line 137 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YMHIk1GIw9IguO8dl:-L8h6BZHAGSSTmZY36xc:b7knd3u) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets_test.go#L137)):*

Delete commented code?
</blockquote></details>

Forgot to finish. Done.

---

*[go/sig/egress/nets_test.go, line 204 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8bCHl4AI7xJTQ2Jedm:-L8h2H0D5f38qNamYFB6:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets_test.go#L204)):*

This can be simplified a bit:
```
SoMsg("netA == netB", netA.Equal(netB), ShouldEqual, tc.ok)
SoMsg("netB == netA", netB.Equal(netA), ShouldEqual, tc.ok)
```
</blockquote></details>

Done.

---

*[go/sig/egress/reader.go, line 105 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8bCvCp3JYBxfG8tWFB:-L8h755K46nZasTNf59s:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/reader.go#L105)):*

If the reader on the other end of this specific ringbuffer is very slow for some reason, this can block forever, taking the rest of the data plane forwarding down with it. In that case, it may be better to try a nonblocking write, and if it fails log the event and discard the packet.
</blockquote></details>

Done.

---

*[go/sig/sigcmn/common.go, line 48 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8bDvaveyf0dOQanUHw:-L8hBIEo445HKrG7vy-Z:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/sigcmn/common.go#L48)):*

`Mask: net.CIDRMask(0, net.IPv4len)` hides internals a bit better, maybe.
</blockquote></details>

Done.

---

*[go/sig/sigcmn/common.go, line 49 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8bEWTzP2OiznX3QOn0:-L8hBIpyBhm5997cj7rx:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/sigcmn/common.go#L49)):*

Same as above.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1464)*

shitz:
:lgtm:

---

Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1464#-:-L8hHVQm9UMOIdDt7Ny7:bnfp4nl)*

scrye:

---

*[go/sig/egress/nets.go, line 28 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YCZmP935PhjpVruSC:-L8hLQTm98gCHlbc6xAZ:b-itkz8b) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L28)):*

Using an interface allows the implementation to be pluggable.
</blockquote></details>

Sure, but do we intend to support more than one routing table?

---

*[go/sig/egress/nets.go, line 53 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YFiYqAcalqd1fT78i:-L8hLjgHD1Cab8eegdvy:b-ivn0on) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L53)):*

(i'm not 100% sure what you mean, that code doesn't compile)

We probably don't care about the ring, but we do care about the network and the IA. Logging the `exnet` as-is gives us that. Otherwise you need two vars, and careful naming of the logging params. Changed the logging in any case for the next comment.
</blockquote></details>

Right, I meant `exnet := ns.nets[i].net`, but the current version no longer needs it anyway.

---

*[go/sig/egress/nets.go, line 54 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YEzFHDLsKrG6jQ0Og:-L8hLreHE3sVz7kPp6Ir:b-t552fa) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L54)):*

Simplified. I don't get what you mean about making the intent clearer though. I can add a comment, but it seems redundant.
</blockquote></details>

I mean giving it a name via moving it to a function. But it's simple enough to follow anyway.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1464)*

scrye:
:lgtm:

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1464#-:-L8hMrCD3aB0GjL1L93S:bnfp4nl)*

kormat:

---

*[go/sig/egress/nets.go, line 28 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YCZmP935PhjpVruSC:-L8kSJhsBoUVpvkZrozI:b-x0paxf) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L28)):*

Sure, but do we intend to support more than one routing table?
</blockquote></details>

Yes :)

---

*[go/sig/egress/nets.go, line 79 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1464#-L8YJEDVB8wukhenTVlR:-L8kT-DA-5X41smVGIwV:b-ehkxch) ([raw file](https://github.com/scionproto/scion/blob/3d6ef0eb50d35a857114a9ae3d28a6cafbf5aaba/go/sig/egress/nets.go#L79)):*

Nice trick!

It might improve clarity a bit to move the code to a helper function, so it looks like `ns.nets = del(ns.nets, idx)` (there are probably better names than my suggestion, though).
</blockquote></details>

It's not a lot of code, it's commented, it's the main point of `Delete()`, and the functionality is only needed by `Delete()`, so i'm inclined to leave it here.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1464)*

PR #1466, opened 2018-03-28
IPv6 topology/integration support

sgmonroy:
This change adds IPv6 topology generation and integration test support along with required IPv6 changes to the overlay for the BRs.

./scion.sh topology -6
./integration/integration_test.sh -6

It also merges revocation_test into integration_test

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1466)


kormat:
Reviewed 11 of 11 files at r1.

---

*[scion.sh, line 47 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mGhESERmu7MzqbATs:-L8mGhESERmu7MzqbATt:b5tlmgw) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/scion.sh#L47)):*
> ```Shell
>         bash gen/zk_datalog_dirs.sh || exit 1
>     fi
>     sudo PYTHONPATH=$PYTHONPATH python/integration/set_ipv6_addr.py -a
> ```

In general we try to only use sudo if required. Otherwise it gets very annoying for people who stay with the more secure default of being prompted for a password. In this case, the `sudo` should be part of the `os.system()` call in `set_ipv6_addr.py`. If someone is using a v4 topo, no addrs are being added/removed, and they shouldn't be prompted for a password.

Assuming a v6 topo, if a user is doing `./scion.sh stop; ./scion.sh start` fairly frequently, they will end up getting prompted for their sudo password a Lot. If the approach is instead changed to:
- if a v6 addr is missing during `./scion.sh start`, add it using sudo.
- don't remove v6 address during `./scion.sh stop` unless a `clean` keyword is passed. (`./scion.sh stop clean`)

Another complication is that if a user deletes a gen folder (e.g. by generating a new topology) without cleaning up the existing ipv6 addresses, they will have to manually remove all the extra IPs. So i suggest that `set_ipv6_addr` will remove _all_ ips in the `::FFFF:127.0.0.0/104` subnet on shutdown, ignoring the `gen/networks.yml` file completely.

---

*[go/lib/overlay/conn/conn.go, line 100 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mLp1REkwTsUbzqCoH:-L8mLp1REkwTsUbzqCoI:b-jbl8bu) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/go/lib/overlay/conn/conn.go#L100)):*
> ```Go
> 		metas[i].Reset()
> 	}
> 	//msgs = msgs[0].(ipv4.Message)
> ```

Can be dropped.

---

*[go/lib/overlay/conn/conn.go, line 174 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mK3VgE9hbZMA94tQp:-L8mK3VgE9hbZMA94tQq:bnhfb1l) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/go/lib/overlay/conn/conn.go#L174)):*
> ```Go
> 
> func (cc *connUDPBase) initConnUDP(network string, listen, remote *topology.AddrInfo) error {
> ```

Blank line.

---

*[go/vendor/vendor.json, line 386 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8lu65s42LsYlIqalv9:-L8lu65s42LsYlIqalvA:b-5h1xpu) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/go/vendor/vendor.json#L386)):*
> ```JSON
> 			"checksumSHA1": "8SyMsOsCcamqXm+5xBQBlvSmm+g=",
> 			"path": "golang.org/x/net/ipv6",
> 			"revision": "892bf7b0c6e2f93b51166bf3882e50277fa5afc6",
> ```

Ah, this is a trap. All deps from a single repo must use the same commit id. In this case, we have at least two existing deps for commit `c7086645de248775cbf2373cf5ca4d2fa664b8c1`, so we need to use that for the ipv6 dep as well.

---

*[integration/revocation_test.sh, line 1 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mEQDuAi4MvgLI5Tvt:-L8mEQDuAi4MvgLI5Tvu:b-u7katl) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/integration/revocation_test.sh#L1)):*
> ```Shell
> #!/bin/bash
> ```

I'm not convinced by moving this into `integration_test.sh`. All the integration tests should be easily runnable independently. As an example, `./python/integration/end2end_test.py` fails with a v6 topology, because it binds to v4 addresses by default. A ~simple fix is:
- Change `python/integration/set_ipv6_addr.py` to always add `::FFFF:127.0.0.2` and `::FFFF:127.0.0.3` if the topo is v6.
- In `base_cli_srv:setup_main()` if no address has been specified, check if the topo is v6 and if so use the above default v6 addresses.

---

*[python/integration/set_ipv6_addr.py, line 35 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8lw0SO0cxlu4TjlM93:-L8lw0SO0cxlu4TjlM94:bsm8sf4) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/integration/set_ipv6_addr.py#L35)):*
> ```Python
>                 address = l.split("= ")[1]
>                 addr = IPv6Address(address[:-1])
>                 if addr in IPv6Network('::127:0:0:0/112'):
> ```

It seems that `::<ipv4>` has been deprecated. Can we move to `::FFFF:<ipv4>` instead? Aaand i've just realised that `::127:0:0:0` is actually a false friend. In ipv6 each set of digits is 16bit, not 8bit like in ipv4. So, moving to the ipv4-mapped ipv6 address space, that's either `::ffff:7f00:0` or `::FFFF:127.0.0.0` (both represent the same value)

The prefix doesn't look right. I suggest `/104`, which matches the size of 127.0.0.0/8 (24bit) from IPv4.

---

*[python/integration/set_ipv6_addr.py, line 37 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8lx0xK3YClk7IircbD:-L8lx0xK3YClk7IircbE:bfzsrpn) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/integration/set_ipv6_addr.py#L37)):*
> ```Python
>                 if addr in IPv6Network('::127:0:0:0/112'):
>                     # Redirect stderr to NULL to avoid error in case the IP address is alredy set
>                     os.system('ip addr %s %s/128 dev lo 2>/dev/null' % (action, str(addr)))
> ```

I _think_ if you use `ip addr replace` instead of `ip addr add`, you won't get that error anymore, and hence can display stderr for ease of debuggability.

---

*[python/topology/generator.py, line 128 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m705rD_CRALRN0TMu:-L8m705rD_CRALRN0TMv:b8blp5k) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L128)):*
> ```Python
> DEFAULT_MININET_NETWORK = "100.64.0.0/10"
> 
> DEFAULT6_NETWORK = "::127:0:0:0/104"
> ```

As discussed elsewhere, this becomes `::ffff:127.0.0.0/104`.

---

*[python/topology/generator.py, line 129 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m-qj944YVtWyV2gIi:-L8m-qj944YVtWyV2gIj:b-n0nuzt) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L129)):*
> ```Python
> 
> DEFAULT6_NETWORK = "::127:0:0:0/104"
> DEFAULT6_PRIV_NETWORK = "fd00::192:168:0:0/112"
> ```

`fd00` is a slightly unusual prefix. It falls under the Unique-Local `fc00::/7` prefix, but that's not obvious without a calculator. If that is the network you've selected, i suggest using `fc00` instead, so it's easier for people to look it up.

---

*[python/topology/generator.py, line 201 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m2y1N7qxtXrrfhDnp:-L8m2y1O1er3GG3XPF9z:b-wth1nc) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L201)):*
> ```Python
>             if self.mininet and val['addr'] == "127.0.0.1":
>                 val['addr'] = "169.254.0.1"
>             self.default_zookeepers[key] = ZKTopo(
> ```

This var isn't actually referenced _anywhere_. It turns out to be something i forgot to remove in a rewrite a loong time ago. Can you remove it for me? Thanks :)

---

*[python/topology/generator.py, line 1027 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m46Yv2ypn-ZspfMWk:-L8m46Yv2ypn-ZspfMWl:b-z1ci65) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L1027)):*
> ```Python
>             addr = ip_address(zk.addr.ip)
>             if addr.version == 6:
>                 servers.append("server.%s=[%s]:%d:%d" %
> ```

This doesn't actually work - zookeeper will just say `does not have the form host:port or host:port:port  or host:port:port:type`.

Instead i think we should just drop this functionality altogether. The `manage:` attribute for zookeeper in the `.topo` files should be removed, the `ZKConfGenerator` class can be deleted.

---

*[python/topology/generator.py, line 1103 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m9mRN7sM6lS0tABNw:-L8m9mRN7sM6lS0tABNx:b-l3yqw6) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/python/topology/generator.py#L1103)):*
> ```Python
>     def __init__(self, network):
>         self._net = ip_network(network)
>         if "/" not in network:
> ```

This seems like a useful error check to keep.

---

*[python/topology/generator.py, line 1130 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m7Dn58eZnGvdo_V_y:-L8m7Dn58eZnGvdo_V_z:b-u47are) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L1130)):*
> ```Python
>         self._allocations = defaultdict(list)
>         # Initialise the allocations with the supplied network, making sure to
>         # exclude <net>/30 if it's contained in the network.
> ```

This stuff is only actually relevant for ipv4 (because there we are using the default loopback allocation), it doesn't apply to ipv6. I have no idea why `DEFAULT_PRIV_NETWORK` is involved at all in the previous code. I would simplify it to:

```
if self._net.version == 4:
  v4_lo = ip_network("127.0.0.0/30")
  if self._net.overlaps(v4_lo):
    self._exclude_net(self._net, v4_lo)
    return

self._allocations[self._net.prefixlen].append(self._net)
```

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[go/lib/overlay/conn/conn.go, line 100 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mLp1REkwTsUbzqCoH:-L95wW6u2PJHcEWCj_Ew:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/go/lib/overlay/conn/conn.go#L100)):*

Can be dropped.
</blockquote></details>

Done.

---

*[go/lib/overlay/conn/conn.go, line 174 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mK3VgE9hbZMA94tQp:-L95wZRMACAyNy1yqSjs:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/go/lib/overlay/conn/conn.go#L174)):*

Blank line.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
*[python/topology/generator.py, line 129 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m-qj944YVtWyV2gIi:-L9623N200figreoBB_I:b-7j9dbv) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L129)):*

`fd00` is a slightly unusual prefix. It falls under the Unique-Local `fc00::/7` prefix, but that's not obvious without a calculator. If that is the network you've selected, i suggest using `fc00` instead, so it's easier for people to look it up.
</blockquote></details>

Having another look at this, fd00: prefix is used when the random 40bits are generated following RFC recommendation, which I am not currently doing as all the bits are zero.
I see that we use this private addresses when enabling binding addresses, do we want this functionality for IPv6?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
*[python/topology/generator.py, line 1103 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m9mRN7sM6lS0tABNw:-L963rfIE_aMuyN87Zaf:b-rmm2cx) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/python/topology/generator.py#L1103)):*

This seems like a useful error check to keep.
</blockquote></details>

Not sure what you are implying, if we get the exception we quit. What do you mean by 'keep' ?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
*[python/topology/generator.py, line 1130 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m7Dn58eZnGvdo_V_y:-L965RRH0pXPyrgk6-fm:b2hjtig) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L1130)):*

This stuff is only actually relevant for ipv4 (because there we are using the default loopback allocation), it doesn't apply to ipv6. I have no idea why `DEFAULT_PRIV_NETWORK` is involved at all in the previous code. I would simplify it to:

```
if self._net.version == 4:
  v4_lo = ip_network("127.0.0.0/30")
  if self._net.overlaps(v4_lo):
    self._exclude_net(self._net, v4_lo)
    return

self._allocations[self._net.prefixlen].append(self._net)
```
</blockquote></details>

Did the change but there is an issue. I think we need to remote those for ipv6 too, otherwise the generator assigns those address .

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
*[python/integration/set_ipv6_addr.py, line 35 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8lw0SO0cxlu4TjlM93:-L96-M0J8rwMdPlQAquM:bb7tv54) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/integration/set_ipv6_addr.py#L35)):*

It seems that `::<ipv4>` has been deprecated. Can we move to `::FFFF:<ipv4>` instead? Aaand i've just realised that `::127:0:0:0` is actually a false friend. In ipv6 each set of digits is 16bit, not 8bit like in ipv4. So, moving to the ipv4-mapped ipv6 address space, that's either `::ffff:7f00:0` or `::FFFF:127.0.0.0` (both represent the same value)

The prefix doesn't look right. I suggest `/104`, which matches the size of 127.0.0.0/8 (24bit) from IPv4.
</blockquote></details>

Go would return ::FFFF:<ipv4> as IPv4 address, which won't match the overlay.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
*[scion.sh, line 47 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mGhESERmu7MzqbATs:-L96IRDk3sLtb6-k2Mp1:b5jz1ng) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/scion.sh#L47)):*

In general we try to only use sudo if required. Otherwise it gets very annoying for people who stay with the more secure default of being prompted for a password. In this case, the `sudo` should be part of the `os.system()` call in `set_ipv6_addr.py`. If someone is using a v4 topo, no addrs are being added/removed, and they shouldn't be prompted for a password.

Assuming a v6 topo, if a user is doing `./scion.sh stop; ./scion.sh start` fairly frequently, they will end up getting prompted for their sudo password a Lot. If the approach is instead changed to:
- if a v6 addr is missing during `./scion.sh start`, add it using sudo.
- don't remove v6 address during `./scion.sh stop` unless a `clean` keyword is passed. (`./scion.sh stop clean`)

Another complication is that if a user deletes a gen folder (e.g. by generating a new topology) without cleaning up the existing ipv6 addresses, they will have to manually remove all the extra IPs. So i suggest that `set_ipv6_addr` will remove _all_ ips in the `::FFFF:127.0.0.0/104` subnet on shutdown, ignoring the `gen/networks.yml` file completely.
</blockquote></details>

I found out that using 'ip addr flush dev lo to ::ffff:127.0.0.0/104' does exactly remove all IPs in loopback that match the subnet.
Only issue is that there is a bug in iptools and an message is shown, even though it works properly:
"Dump was interrupted and may be inconsistent."

Are we ok with that?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
*[integration/revocation_test.sh, line 1 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mEQDuAi4MvgLI5Tvt:-L99EoKG8fy5tzO9fsT3:bjinuhy) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/integration/revocation_test.sh#L1)):*

I'm not convinced by moving this into `integration_test.sh`. All the integration tests should be easily runnable independently. As an example, `./python/integration/end2end_test.py` fails with a v6 topology, because it binds to v4 addresses by default. A ~simple fix is:
- Change `python/integration/set_ipv6_addr.py` to always add `::FFFF:127.0.0.2` and `::FFFF:127.0.0.3` if the topo is v6.
- In `base_cli_srv:setup_main()` if no address has been specified, check if the topo is v6 and if so use the above default v6 addresses.
</blockquote></details>

Regarding the revocation_test, merging them removes one end2end_test run, from 3 to 2. I was talking to @shitz about it and he mentioned that initially was just to run the revocation_test independently and faster. I agree that it won't be an independent test anymore but it will improve ci time.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
*[integration/revocation_test.sh, line 1 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mEQDuAi4MvgLI5Tvt:-L99GeXQ3OUeDKAX1rbj:b-q09vpx) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/integration/revocation_test.sh#L1)):*

Regarding the revocation_test, merging them removes one end2end_test run, from 3 to 2. I was talking to @shitz about it and he mentioned that initially was just to run the revocation_test independently and faster. I agree that it won't be an independent test anymore but it will improve ci time.
</blockquote></details>

Regarding check if the topo is v6, afaik there is no quick way to check other than parsing topo files to check if addresses are v6, is there a quick way to check for it?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
*[scion.sh, line 47 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mGhESERmu7MzqbATs:-L99IIFODVE0SGEmZjj8:b8c63vs) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/scion.sh#L47)):*

I found out that using 'ip addr flush dev lo to ::ffff:127.0.0.0/104' does exactly remove all IPs in loopback that match the subnet.
Only issue is that there is a bug in iptools and an message is shown, even though it works properly:
"Dump was interrupted and may be inconsistent."

Are we ok with that?
</blockquote></details>

I reckon logic could be more straight forward if we have an easy way to check if topo is v6 or not.
Easy way would be to create a file for it in gen/, but then if the user removes the folder there won't be a way to know.
We could use env variable, but not reliable if working with multiple shells.
Is there a reliable way to do this?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

kormat:
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[scion.sh, line 47 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mGhESERmu7MzqbATs:-L99eeXj2AhgCPuDvMUs:big9u6z) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/scion.sh#L47)):*

I reckon logic could be more straight forward if we have an easy way to check if topo is v6 or not.
Easy way would be to create a file for it in gen/, but then if the user removes the folder there won't be a way to know.
We could use env variable, but not reliable if working with multiple shells.
Is there a reliable way to do this?
</blockquote></details>

There is unfortunately no reliable way to do this. Your `flush` approach looks promising, i think we can live with the error message if it works reliably.

---

*[integration/revocation_test.sh, line 1 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mEQDuAi4MvgLI5Tvt:-L99g1Xs2Sv_atWMFKfZ:bexszqq) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/integration/revocation_test.sh#L1)):*

I think the tests are easily run, you just need to set client/server flags. Initially I did not want to add .2 and .3 in the set_ip6_addr.py as they are not part of the infra, but only required for the tests. Now, I reckon the loopback setup is already mostly for testing so not a big issue adding those addresses too.
</blockquote></details>

The e2e run that's removed saves a total of 4s. I don't see that as being worth it, or having a significant impact on ci runs. Source (https://circleci.com/gh/scionproto/scion/6787):
```
========> (2018-03-29 13:31:04+00:00) Testing connectivity between all the hosts.
========> (2018-03-29 13:31:08+00:00) Stopping routers and waiting for 4s.
```
Being able to run revocation tests without needing to run all integration tests is a big quality-of-life improvement if you're trying to test revocations. Plus the scmp error test causes a lot of log "spam" which makes it harder to debug revocations.

It probably _does_ make sense to remove the first e2e run from `revocation_test.sh`, so it's easier to test when you have deliberately broken some extra connectivity.

I disagree that "you just need to set client/server flags" == "tests are easily run".

Re: quick way to test for v6 topo, i think the easiest is to add a new file, `gen/overlay`, and store the overlay in there. That would simplify things for anything which cares.

---

*[python/integration/set_ipv6_addr.py, line 35 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8lw0SO0cxlu4TjlM93:-L99hpwR8d7ODa9uwHWB:b-1xgmr1) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/integration/set_ipv6_addr.py#L35)):*

Go would return ::FFFF:<ipv4> as IPv4 address, which won't match the overlay.
</blockquote></details>

That's a good point, but probably the answer is that we need to fix our assumptions. It's legal for a site to use v4-mapped V6 addresses anyway, so we need to support that.

---

*[python/topology/generator.py, line 129 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m-qj944YVtWyV2gIi:-L9A-aQK1WNlp--VcCfD:b-33u2rl) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L129)):*

Having another look at this, fd00: prefix is used when the random 40bits are generated following RFC recommendation, which I am not currently doing as all the bits are zero.
I see that we use this private addresses when enabling binding addresses, do we want this functionality for IPv6?
</blockquote></details>

Good point re: `fd00::/8`, we should probably follow the RFC and generate a 40bit prefix to use statically for our setup.

Re: bind addresses, it's definitely _less_ likely that this is needed for v6, but unfortunately not impossible, so yeah it would be good to support it so we can test it etc.

---

*[python/topology/generator.py, line 1103 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m9mRN7sM6lS0tABNw:-L99ikff6sxMCRCKgpOu:b-am930e) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/python/topology/generator.py#L1103)):*

Not sure what you are implying, if we get the exception we quit. What do you mean by 'keep' ?
</blockquote></details>

The comment is on the `if "/" not in network:` check from the original code.

---

*[python/topology/generator.py, line 1130 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m7Dn58eZnGvdo_V_y:-L9A07pT2q_1WD1O3S5u:bde0hzp) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L1130)):*

Did the change but there is an issue. I think we need to remote those for ipv6 too, otherwise the generator assigns those address .
</blockquote></details>

Ah, hmm. Yeah, that's a fair point. I was thinking that `::ffff:127.0.0.1` _isn't_ treated specially by the kernel, but the client/server addresses should be excluded.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

kormat:
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[python/topology/generator.py, line 129 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m-qj944YVtWyV2gIi:-L9A0yDuA2T2uLonDscA:b-57hwer) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L129)):*

Good point re: `fd00::/8`, we should probably follow the RFC and generate a 40bit prefix to use statically for our setup.

Re: bind addresses, it's definitely _less_ likely that this is needed for v6, but unfortunately not impossible, so yeah it would be good to support it so we can test it etc.
</blockquote></details>

It would also be good to add a comment saying where this choice of network came from. Unfortunately there are no RFCs (that i can find) that document `fd00::/8` specifically, so probably the best we can do is link to https://en.wikipedia.org/wiki/Unique_local_address#Definition

(also based on that i agree with using `fd00::/8`, and not `fc00::/7`).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
Review status: 1 of 13 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[scion.sh, line 47 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mGhESERmu7MzqbATs:-L9Fsss66oBAo5djfDSs:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/scion.sh#L47)):*

There is unfortunately no reliable way to do this. Your `flush` approach looks promising, i think we can live with the error message if it works reliably.
</blockquote></details>

Done.

---

*[go/vendor/vendor.json, line 386 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8lu65s42LsYlIqalv9:-L99gTtgBspdLPoO2HWX:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/go/vendor/vendor.json#L386)):*

Ah, this is a trap. All deps from a single repo must use the same commit id. In this case, we have at least two existing deps for commit `c7086645de248775cbf2373cf5ca4d2fa664b8c1`, so we need to use that for the ipv6 dep as well.
</blockquote></details>

Done.

---

*[integration/revocation_test.sh, line 1 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mEQDuAi4MvgLI5Tvt:-L9FsHwS-SyKiaqpDq1v:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/integration/revocation_test.sh#L1)):*

The e2e run that's removed saves a total of 4s. I don't see that as being worth it, or having a significant impact on ci runs. Source (https://circleci.com/gh/scionproto/scion/6787):
```
========> (2018-03-29 13:31:04+00:00) Testing connectivity between all the hosts.
========> (2018-03-29 13:31:08+00:00) Stopping routers and waiting for 4s.
```
Being able to run revocation tests without needing to run all integration tests is a big quality-of-life improvement if you're trying to test revocations. Plus the scmp error test causes a lot of log "spam" which makes it harder to debug revocations.

It probably _does_ make sense to remove the first e2e run from `revocation_test.sh`, so it's easier to test when you have deliberately broken some extra connectivity.

I disagree that "you just need to set client/server flags" == "tests are easily run".

Re: quick way to test for v6 topo, i think the easiest is to add a new file, `gen/overlay`, and store the overlay in there. That would simplify things for anything which cares.
</blockquote></details>

Done.

---

*[python/integration/set_ipv6_addr.py, line 35 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8lw0SO0cxlu4TjlM93:-L9FsJJtEUAIVUiPZ80c:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/integration/set_ipv6_addr.py#L35)):*

That's a good point, but probably the answer is that we need to fix our assumptions. It's legal for a site to use v4-mapped V6 addresses anyway, so we need to support that.
</blockquote></details>

Done.

---

*[python/integration/set_ipv6_addr.py, line 37 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8lx0xK3YClk7IircbD:-L96-U5fBlYrXHod3xHn:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/integration/set_ipv6_addr.py#L37)):*

I _think_ if you use `ip addr replace` instead of `ip addr add`, you won't get that error anymore, and hence can display stderr for ease of debuggability.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 128 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m705rD_CRALRN0TMu:-L96-W-hDuYuCDyLLZJA:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L128)):*

As discussed elsewhere, this becomes `::ffff:127.0.0.0/104`.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 129 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m-qj944YVtWyV2gIi:-L9FsVTs5CnXMfvMaqXx:b9brxgx) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L129)):*

It would also be good to add a comment saying where this choice of network came from. Unfortunately there are no RFCs (that i can find) that document `fd00::/8` specifically, so probably the best we can do is link to https://en.wikipedia.org/wiki/Unique_local_address#Definition

(also based on that i agree with using `fd00::/8`, and not `fc00::/7`).
</blockquote></details>

https://tools.ietf.org/html/rfc4193#section-3.2
I guess you have a point that it should be fd00::/7 instead of fd00::/8 ?

---

*[python/topology/generator.py, line 201 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m2y1N7qxtXrrfhDnp:-L962fhTAJuZTM8TaAgn:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L201)):*

This var isn't actually referenced _anywhere_. It turns out to be something i forgot to remove in a rewrite a loong time ago. Can you remove it for me? Thanks :)
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 1027 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m46Yv2ypn-ZspfMWk:-L963WxFCim-GrPEjZER:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L1027)):*

This doesn't actually work - zookeeper will just say `does not have the form host:port or host:port:port  or host:port:port:type`.

Instead i think we should just drop this functionality altogether. The `manage:` attribute for zookeeper in the `.topo` files should be removed, the `ZKConfGenerator` class can be deleted.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 1103 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m9mRN7sM6lS0tABNw:-L9FsrJ56zBUyDFlo454:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/python/topology/generator.py#L1103)):*

The comment is on the `if "/" not in network:` check from the original code.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 1130 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m7Dn58eZnGvdo_V_y:-L9Fso1Q9p-hqmhC0bVc:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L1130)):*

Ah, hmm. Yeah, that's a fair point. I was thinking that `::ffff:127.0.0.1` _isn't_ treated specially by the kernel, but the client/server addresses should be excluded.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

kormat:
Reviewed 13 of 14 files at r2.

---

*[go/lib/overlay/conn/conn.go, line 174 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mK3VgE9hbZMA94tQp:-L9P1wml-oEb5E9H4E8K:baxypm9) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/go/lib/overlay/conn/conn.go#L174)):*

Done.
</blockquote></details>

Not visible in PR.

---

*[python/integration/base_cli_srv.py, line 387 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9PBnhY1Nb0RHmdZDYq:-L9PBnhY1Nb0RHmdZDYr:bg19nmi) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/integration/base_cli_srv.py#L387)):*
> ```Python
> 
>     overlay = get_overlay()
>     if overlay == 'UDP/IPv6' or overlay == 'IPv6':
> ```

Can be simplified to `if "IPv6" in overlay:`

---

*[python/integration/set_ipv6_addr.py, line 35 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8lw0SO0cxlu4TjlM93:-L9P3PnCAe9e1sHZcBbE:bmz96tm) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/integration/set_ipv6_addr.py#L35)):*

Done.
</blockquote></details>

(For posterity: it's not legal to use v4-mapped v6 like i was assuming. The linux kernel (at least) won't allow you to bind a v6 socket to a v4-mapped v6 addr)

---

*[python/integration/set_ipv6_addr.py, line 57 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9P4u3o2MQZjqXiPw7K:-L9P4u3o2MQZjqXiPw7L:b-e9xwr7) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/integration/set_ipv6_addr.py#L57)):*
> ```Python
> def main():
>     overlay = get_overlay()
>     if overlay == 'UDP/IPv4' or overlay == 'IPv4':
> ```

You can simplify this to: `if "IPv4" in overlay:`. 

---

*[python/integration/set_ipv6_addr.py, line 67 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9P5G-X14A-MVVU_HCL:-L9P5G-X14A-MVVU_HCM:b-3jgkvt) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/integration/set_ipv6_addr.py#L67)):*
> ```Python
>     args = parser.parse_args()
>     if args.add:
>         os.system('sudo ip addr replace %s dev lo' % (DEFAULT6_CLIENT + DEFAULT6_MASK))
> ```

It would be nicer to not require `sudo` if the addrs are already present. A suggestion:

```
def ip_missing(addr): # Can only safely detect the _absence_ of addr.
   return len(os.popen("ip addr show dev lo to %s 2>&1" % addr).read()) == 0

def ip_add(addr):
   if ip_missing(addr):
      os.system('sudo ip addr add %s%s dev lo' % (addr, DEFAULT6_MASK))

def ip_clear(self):
   if not ip_missing(DEFAULT6_NETWORK_ADDR + DEFAULT6_MASK):
       os.system('sudo ip addr flush dev lo to %s' % (DEFAULT6_NETWORK_ADDR + DEFAULT6_MASK))
   ... # repeat for DEFAULT6_PRIV_NETWORK + DEFAULT6_MASK.

...
  if args.add:
    ip_add(DEFAULT6_CLIENT)
    ip_add(DEFAULT6_SERVER)
```

Not tested, but it should be fairly close.

---

*[python/topology/generator.py, line 129 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m-qj944YVtWyV2gIi:-L9P2_7U6M7RwA2Vxf0l:b-pbmikz) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L129)):*

https://tools.ietf.org/html/rfc4193#section-3.2
I guess you have a point that it should be fd00::/7 instead of fd00::/8 ?
</blockquote></details>

I'm talking specifically about `fd00::/8`. The wikipedia article mentions that `fc00::/8` is not standardised yet, but that `fd00::/8` is, and i can't find an RFC to support that. I can however find a pdf from apnic that agrees with wikipedia, but it's not not a great reference.

TL;DR: i think just add the above wikipedia link in `defines.py`.

---

*[python/topology/generator.py, line 1103 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m9mRN7sM6lS0tABNw:-L9PACCSE8QipQ8gJ3wC:baxypm9) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/python/topology/generator.py#L1103)):*

Done.
</blockquote></details>

Not visible in PR.

---

*[python/topology/generator.py, line 1130 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m7Dn58eZnGvdo_V_y:-L9PAQRt3ixaIsXMA0yC:b-n80snl) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L1130)):*
Comment needs to be updated:

> making sure to exclude 127.0.0.0/30 (for v4) and DEFAULT6_NETWORK_ADDR/126 (for v6) if it's contained in the network.

---

*[python/topology/generator.py, line 216 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9P9Mmf-DtkA8s3jC4N:-L9P9Mmf-DtkA8s3jC4O:b5bft2h) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L216)):*
> ```Python
>         cert_files, trc_files, cust_files = self._generate_certs_trcs(ca_certs)
>         topo_dicts, zookeepers, networks, prv_networks = self._generate_topology()
>         self._generate_supervisor(topo_dicts, zookeepers)
> ```

This no longer needs to take `zookeepers`.

---

*[python/topology/generator.py, line 714 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9P9ds8-u_A8cb0PNjr:-L9P9ds8-u_A8cb0PNjs:b-19xjzq) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L714)):*
> ```Python
>     def _gen_zk_entries(self, topo_id, as_conf):
>         zk_conf = {}
>         if "zookeepers" in as_conf:
> ```

This `if` clause can be dropped.

---

*[python/topology/generator.py, line 947 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9PA0NO3Yw37hcO5mfr:-L9PA0NO3Yw37hcO5mfs:b-ecsjm) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L947)):*
> ```Python
>         if elem.startswith("br"):
>             prog['environment'] += ',GODEBUG="cgocheck=0"'
>         if elem.startswith("zk") and self.zk_config["Environment"]:
> ```

This can be dropped.

---

*[python/topology/generator.py, line 1029 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9PBOR900SRhlwYXbTJ:-L9PBOR900SRhlwYXbTK:b8b7sht) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L1029)):*
> ```Python
>         self.leaderPort = self._get_def("leaderPort")
>         self.electionPort = self._get_def("electionPort")
>         self.maxClientCnxns = self._get_def("maxClientCnxns")
> ```

These 3 attributes can be dropped.

---

*[python/topology/generator.py, line 1034 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9PBYtwE0ZQPsC8IT3I:-L9PBYtwE0ZQPsC8IT3J:b-ecsjm) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L1034)):*
> ```Python
>         return self.topo_config.get(key, self.zk_config["Default"][key])
> 
>     def zk_conf(self, data_dir, data_log_dir):
> ```

This can be dropped.

---

*[python/topology/generator.py, line 1045 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9PBeP24xa06-W8XfDW:-L9PBeP24xa06-W8XfDX:b-ecsjm) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L1045)):*
> ```Python
>         return "\n".join(c)
> 
>     def super_conf(self, topo_id, name, out_dir):
> ```

This can be dropped.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

sgmonroy:
Review status: 6 of 14 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[go/lib/overlay/conn/conn.go, line 174 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8mK3VgE9hbZMA94tQp:-L9Ptj7yDBbaf_lH26cO:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/go/lib/overlay/conn/conn.go#L174)):*

Not visible in PR.
</blockquote></details>

Done.

---

*[python/integration/base_cli_srv.py, line 387 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9PBnhY1Nb0RHmdZDYq:-L9PhnXK2Wngh4A5cRBo:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/integration/base_cli_srv.py#L387)):*

Can be simplified to `if "IPv6" in overlay:`
</blockquote></details>

Done.

---

*[python/integration/set_ipv6_addr.py, line 57 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9P4u3o2MQZjqXiPw7K:-L9Phtgk6VMdV3izBrZe:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/integration/set_ipv6_addr.py#L57)):*

You can simplify this to: `if "IPv4" in overlay:`. 
</blockquote></details>

Done.

---

*[python/integration/set_ipv6_addr.py, line 67 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9P5G-X14A-MVVU_HCL:-L9PsaWMC7qEl6f_Gg_9:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/integration/set_ipv6_addr.py#L67)):*

It would be nicer to not require `sudo` if the addrs are already present. A suggestion:

```
def ip_missing(addr): # Can only safely detect the _absence_ of addr.
   return len(os.popen("ip addr show dev lo to %s 2>&1" % addr).read()) == 0

def ip_add(addr):
   if ip_missing(addr):
      os.system('sudo ip addr add %s%s dev lo' % (addr, DEFAULT6_MASK))

def ip_clear(self):
   if not ip_missing(DEFAULT6_NETWORK_ADDR + DEFAULT6_MASK):
       os.system('sudo ip addr flush dev lo to %s' % (DEFAULT6_NETWORK_ADDR + DEFAULT6_MASK))
   ... # repeat for DEFAULT6_PRIV_NETWORK + DEFAULT6_MASK.

...
  if args.add:
    ip_add(DEFAULT6_CLIENT)
    ip_add(DEFAULT6_SERVER)
```

Not tested, but it should be fairly close.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 129 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m-qj944YVtWyV2gIi:-L9PsyNN6qKhwiG7hZbB:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L129)):*

I'm talking specifically about `fd00::/8`. The wikipedia article mentions that `fc00::/8` is not standardised yet, but that `fd00::/8` is, and i can't find an RFC to support that. I can however find a pdf from apnic that agrees with wikipedia, but it's not not a great reference.

TL;DR: i think just add the above wikipedia link in `defines.py`.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 1103 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m9mRN7sM6lS0tABNw:-L9Pyaz0BPXAArp__N7k:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/460ac3b88c2dcb6a8a63427d568f896e7a93f5f1/python/topology/generator.py#L1103)):*

Not visible in PR.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 1130 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L8m7Dn58eZnGvdo_V_y:-L9PtL8pBhR7zP-J8pI0:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fa21cbc9ef692324ba629cbc3e5fa2d1e6a23675/python/topology/generator.py#L1130)):*

Comment needs to be updated:

> making sure to exclude 127.0.0.0/30 (for v4) and DEFAULT6_NETWORK_ADDR/126 (for v6) if it's contained in the network.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 216 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9P9Mmf-DtkA8s3jC4N:-L9Pu70d93ABESZI93cg:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L216)):*

This no longer needs to take `zookeepers`.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 714 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9P9ds8-u_A8cb0PNjr:-L9PygiV3xMtUjo8tBRL:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L714)):*

This `if` clause can be dropped.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 947 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9PA0NO3Yw37hcO5mfr:-L9Px3y66PkFY8dSAirw:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L947)):*

This can be dropped.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 1029 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9PBOR900SRhlwYXbTJ:-L9Px4wy8w5WA2xfYsTO:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L1029)):*

These 3 attributes can be dropped.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 1034 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9PBYtwE0ZQPsC8IT3I:-L9Px5NJDGIwSeqklP6g:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L1034)):*

This can be dropped.
</blockquote></details>

Done.

---

*[python/topology/generator.py, line 1045 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1466#-L9PBeP24xa06-W8XfDW:-L9Px7-F56frweI66eOY:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/796f31529062fa052205e946af16f5c88ff1caa7/python/topology/generator.py#L1045)):*

This can be dropped.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1466)*

PR #1478, opened 2018-04-05
Check for unique ASes in generator, use non-production ASes.

kormat:
Now that we are using global AS numbers, the generator needs to catch a
`.topo` file that repeats AS numbers. Also, we need to change to using
example AS numbers where possible, and private-use AS numbers for the
default topo etc.

For private-use, i've gone with the following scheme for our test topos:
- The prefix is 4_295_00X for ISD X.
- Core ASes have a suffix of _0X0, where X is 1,2,3 for the
  first/second/third core AS in an ISD etc.
- Non-core ASes have the suffix of _0XY where X is their core _issuer_
  AS, and Y is incremented for each non-core AS under that issuer.

topology/Default.topo is annotated with the old AS IDs to help with the
transition.

N.B.: doc/fig/default-topo.pdf needs to be updated, i'll be looking for
a volunteer. :)

For the interfaces in go/lib/xtest/graph, i've gone with the last 4
digits of each AS number, which thankfully comfortably fits in the
64bit namespace.

Ref https://github.com/scionproto/scion/wiki/ISD-and-AS-numbering

Also:
- Remove an old ignore.goconvey
- Fix address regex in snet to cope with _'s in AS numbers.


scrye:
Reviewed 69 of 73 files at r1.

---

*[go/lib/pathmgr/pathmgr_test.go, line 235 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1478#-L9P6iYCAl7hFdt3uyh1:-L9P6iYCAl7hFdt3uyh2:b-fnmobk) ([raw file](https://github.com/scionproto/scion/blob/df6c2fb72555c922cfca50db39d6e172126fd828/go/lib/pathmgr/pathmgr_test.go#L235)):*
> ```Go
> func apsCheckPaths(desc string, aps AppPathSet, expValues ...string) {
> 	SoMsg(fmt.Sprintf("%s: len", desc), len(aps), ShouldEqual, len(expValues))
> 	//fmt.Println(getPathStrings(aps))
> ```

Leftover comment?

---

*[go/lib/xtest/graph/graph.go, line 294 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1478#-L9P7t9b2SBRaM2uvW1N:-L9P7t9b2SBRaM2uvW1O:bplfody) ([raw file](https://github.com/scionproto/scion/blob/df6c2fb72555c922cfca50db39d6e172126fd828/go/lib/xtest/graph/graph.go#L294)):*
> ```Go
> 		{"2-4_295_002_010", 20102011, "2-4_295_002_011", 20112010},
> 		{"2-4_295_002_020", 20202021, "2-4_295_002_021", 20212020},
> 		{"2-4_295_002_011", 20112021, "2-4_295_002_021", 20112021},
> ```

Just noticed that the original edge here was wrong, the right side should be `20212011`.

---

*[topology/Default.topo, line 39 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1478#-L9PB8Yn827jIoJcdLWf:-L9PB8Yn827jIoJcdLWg:b6u2hpa) ([raw file](https://github.com/scionproto/scion/blob/df6c2fb72555c922cfca50db39d6e172126fd828/topology/Default.topo#L39)):*
> ```
>   1-4_295_001_021: # old 1-15
>     cert_issuer: 1-4_295_001_020
>   1-4_295_001_022: # old 1-17
> ```

This should be 1-18.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1478)*

kormat:

---

*[go/lib/pathmgr/pathmgr_test.go, line 235 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1478#-L9P6iYCAl7hFdt3uyh1:-L9PD6xeDMuMLc_uxBi4:by83qqd) ([raw file](https://github.com/scionproto/scion/blob/df6c2fb72555c922cfca50db39d6e172126fd828/go/lib/pathmgr/pathmgr_test.go#L235)):*

Leftover comment?
</blockquote></details>

Yep, thanks! Fixed.

---

*[go/lib/xtest/graph/graph.go, line 294 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1478#-L9P7t9b2SBRaM2uvW1N:-L9PDJ9s2n1LkdoyQiMM:b-r1ij4e) ([raw file](https://github.com/scionproto/scion/blob/df6c2fb72555c922cfca50db39d6e172126fd828/go/lib/xtest/graph/graph.go#L294)):*

Just noticed that the original edge here was wrong, the right side should be `20212011`.
</blockquote></details>

Good catch :)

---

*[topology/Default.topo, line 39 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1478#-L9PB8Yn827jIoJcdLWf:-L9PE2ID93M7MEksLK1e:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/df6c2fb72555c922cfca50db39d6e172126fd828/topology/Default.topo#L39)):*

This should be 1-18.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1478)*

kormat:
There is one breaking change in this PR - it fixes the filenames of certs to be `_` separated. This will only impact you if your AS numbers are > 999.

PR #1480, opened 2018-04-06
WIP: Revocation Authentication Rework

worxli:
Model / type changes for the revocation authentication rework. 

RevInfo:
- Removed hash tree related fields
- Added timestamp and link type

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1480)


kormat:
Reviewed 3 of 3 files at r1.

---

*[proto/rev_info.capnp, line 7 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9PU_t42FfGeUQPvURR:-L9PU_t42FfGeUQPvURS:b-e9gouk) ([raw file](https://github.com/scionproto/scion/blob/e572815ad57c55b088ff7153884266c412f2089d/proto/rev_info.capnp#L7)):*
> ```
> 
> struct RevInfo {
> 	ifID @0 :UInt64;  # ID of the interface to be revoked
> ```

These should be spaces, not tabs. This seems to be the only file that was using tabs.

---

*[python/lib/packet/path_mgmt/rev_info.py, line 70 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9PWNIL51cWpvi_NkDO:-L9PWNIL51cWpvi_NkDP:b53b6ms) ([raw file](https://github.com/scionproto/scion/blob/e572815ad57c55b088ff7153884266c412f2089d/python/lib/packet/path_mgmt/rev_info.py#L70)):*
> ```Python
>     def validate(self):
>         if self.p.timestamp > datetime.now():
>             raise RevInfoValidationError("Invalid timestamp: %s" % self.p.timestamp)
> ```

I'd make this more descriptive: "Timestamp in the future". I'd also add 1 to "now" to allow a small bit of wiggle-room in clock synchronisation.
Aaalso, `datetime.now()` is probably not what you want. It returns a tuple. `int(time.time()) + 1` is probably the right answer.

---

*[python/lib/packet/path_mgmt/rev_info.py, line 95 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9PWZT341mMOrtP_0kb:-L9PWZT341mMOrtP_0kc:b-r8x68v) ([raw file](https://github.com/scionproto/scion/blob/e572815ad57c55b088ff7153884266c412f2089d/python/lib/packet/path_mgmt/rev_info.py#L95)):*
> ```Python
>     def short_desc(self):
>         return "RevInfo: %s IF: %d Link type: %s Timestamp: %s" % (
>             self.isd_as(), self.p.ifID, self.p.link_type, iso_timestamp(self.p.timestamp / 1000000))
> ```

Why the division? The capnp comment says it's in seconds, and `iso_timestamp` expects seconds. Ah, but the go code assumes it's in ųs. I don't know if that level of precision is desirable. If a machine is 100 ųs behind another, it could reject a revocation.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1480)*

worxli:
Review status: 0 of 13 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[proto/rev_info.capnp, line 7 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9PU_t42FfGeUQPvURR:-L9QDCY2BhlxLZ-pw0-X:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e572815ad57c55b088ff7153884266c412f2089d/proto/rev_info.capnp#L7)):*

These should be spaces, not tabs. This seems to be the only file that was using tabs.
</blockquote></details>

Done.

---

*[python/lib/packet/path_mgmt/rev_info.py, line 70 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9PWNIL51cWpvi_NkDO:-L9QD7kf8baEE2TnHOMT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e572815ad57c55b088ff7153884266c412f2089d/python/lib/packet/path_mgmt/rev_info.py#L70)):*

I'd make this more descriptive: "Timestamp in the future". I'd also add 1 to "now" to allow a small bit of wiggle-room in clock synchronisation.
Aaalso, `datetime.now()` is probably not what you want. It returns a tuple. `int(time.time()) + 1` is probably the right answer.
</blockquote></details>

Done.

---

*[python/lib/packet/path_mgmt/rev_info.py, line 95 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9PWZT341mMOrtP_0kb:-L9QD9Ci41sjY-vYxaHu:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e572815ad57c55b088ff7153884266c412f2089d/python/lib/packet/path_mgmt/rev_info.py#L95)):*

Why the division? The capnp comment says it's in seconds, and `iso_timestamp` expects seconds. Ah, but the go code assumes it's in ųs. I don't know if that level of precision is desirable. If a machine is 100 ųs behind another, it could reject a revocation.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1480)*

kormat:
Reviewed 12 of 13 files at r2.
Review status: 12 of 13 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[go/lib/ctrl/path_mgmt/rev_info.go, line 31 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9QIB2b9UycZ0eShmiG:-L9QIB2b9UycZ0eShmiH:b-8yrydh) ([raw file](https://github.com/scionproto/scion/blob/784f75c49b014e76d5355fad5ac50d5c4a7daea4/go/lib/ctrl/path_mgmt/rev_info.go#L31)):*
> ```Go
> )
> 
> const TTL = 10 * time.Second // Revocation TTL
> ```

As this is in the `path_mgmt` package, `path_mgmt.TTL` isn't very descriptive, so i'd change it to `MinRevTTL`.

---

*[go/lib/ctrl/path_mgmt/rev_info.go, line 55 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9QIkFW3q5Rfw-pE6nn:-L9QIkFW3q5Rfw-pE6no:bhl8bj2) ([raw file](https://github.com/scionproto/scion/blob/784f75c49b014e76d5355fad5ac50d5c4a7daea4/go/lib/ctrl/path_mgmt/rev_info.go#L55)):*
> ```Go
> 	now := uint32(time.Now().Second())
> 	// Revocation is not valid if its timestamp is not within the TTL
> 	if r.Timestamp > now || r.Timestamp < now-uint32(TTL.Seconds()) {
> ```

There's a difference between these two cases. In the first case the revinfo is _illegal_. In the second it's just expired. While in both cases the app shouldn't use it, there is a difference in logging etc. I think i would change the approach:
```
type RevExpired struct {} // Implement common.Timeout

func (r *RevInfo) Validate(now time.Time) error {
  ts := time.Unix(r.Timestamp, 0)
  if ts.After(now.Add(1 * time.Second)) {
    return common.NewBasicError("RevInfo timestamp in future", nil, "now", util.TimeFormat(now), "ts", util.TimeFormat(ts))
  }
  ttl := time.Duration(r.TTL) * time.Second
  if ts.After(now.Add(ttl)) {
    return common.NewBasicError("RevInfo expired", &RevExpired{}, "now", util.TimeFormat(now), "ts", util.TimeFormat(ts), "ttl", ttl)
  }
  return nil
}
```
If a caller gets `nil`, then the revocation is valid and active. If they get an error, they can call `common.IsTimeoutErr` on it to see if it's an expiry error or not.

---

*[python/lib/rev_cache.py, line 84 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9QFIvu0ox9w9cvDE6n:-L9QEayVCN2pTSjMiNQB:bf3bq0q) ([raw file](https://github.com/scionproto/scion/blob/784f75c49b014e76d5355fad5ac50d5c4a7daea4/python/lib/rev_cache.py#L84)):*
> ```Python
>             return default
> 
>     def get_by_link_type(self, type):
> ```

When you have names that clash with buildin types/keywords/etc, the convention is to add an underscore, so this becomes `type_`.

---

*[python/lib/packet/path_mgmt/rev_info.py, line 51 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9QG70779m8ceXm238o:-L9QG70779m8ceXm238p:bgke45l) ([raw file](https://github.com/scionproto/scion/blob/784f75c49b014e76d5355fad5ac50d5c4a7daea4/python/lib/packet/path_mgmt/rev_info.py#L51)):*
> ```Python
> 
>     @classmethod
>     def from_values(cls, isd_as, if_id, link_type, timestamp, ttl=10):
> ```

Don't use a "magic" value. Use the const from `defines.py`.

---

*[python/lib/packet/path_mgmt/rev_info.py, line 60 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9QGB1S15R38Ex5_7nH:-L9QGB1S15R38Ex5_7nI:bdoc7oy) ([raw file](https://github.com/scionproto/scion/blob/784f75c49b014e76d5355fad5ac50d5c4a7daea4/python/lib/packet/path_mgmt/rev_info.py#L60)):*
> ```Python
>         :param int timestamp: Revocation creation timestamp in seconds
>         :param int TTL: Revocation validity period in seconds
>         """
> ```

There should be an `assert` here to ensure that ttl >= 10.

---

*[python/lib/packet/path_mgmt/rev_info.py, line 80 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9QGOFW54KZz0r_6Plu:-L9QGOFW54KZz0r_6Plv:b-wk9y0l) ([raw file](https://github.com/scionproto/scion/blob/784f75c49b014e76d5355fad5ac50d5c4a7daea4/python/lib/packet/path_mgmt/rev_info.py#L80)):*
> ```Python
>         now = int(time.time())
>         # make sure the revocation timestamp is within the validity window
>         if self.p.timestamp > now + 1 or self.p.timestamp < now - self.p.ttl:
> ```

At this point validation should already have been run, so i would put an `assert` for the future-timestamp issue:

```
assert self.p.timestamp <= now + 1
return now < (self.p.timestamp + self.p.ttl)
```

---

*[python/lib/packet/path_mgmt/rev_info.py, line 102 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9QHfE81CHuIu9G36NX:-L9QHfE81CHuIu9G36NY:b-n4dw21) ([raw file](https://github.com/scionproto/scion/blob/784f75c49b014e76d5355fad5ac50d5c4a7daea4/python/lib/packet/path_mgmt/rev_info.py#L102)):*
> ```Python
> 
>     def short_desc(self):
>         return "RevInfo: %s IF: %d Link type: %s Timestamp: %s TTL: %d" % (
> ```

As python doesn't really have types, we usually use %s for all formatting so that it doesn't raise an exception at run-time if a type isn't what was expected. (FWIW we do the opposite in Go, because we have a compiler that understands that types are important ;).

---

*[python/sciond/sciond.py, line 89 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1480#-L9QLdlDCrdwN7saEBGr:-L9QLdlDCrdwN7saEBGs:b-656rgz) ([raw file](https://github.com/scionproto/scion/blob/784f75c49b014e76d5355fad5ac50d5c4a7daea4/python/sciond/sciond.py#L89)):*
> ```Python
> from scion_elem.scion_elem import SCIONElement
> 
> from python.lib.packet.path_mgmt.rev_info import ProtoLinkType
> ```

This indicates that this is the wrong place for the link type. I'd put it in `python/lib/types.py`

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1480)*

PR #1505, opened 2018-04-06
  Modify install dependencies script for Ubuntu 18.04

jinankjain:
Major changes:
- Install latest version of zlog than 1.2.12 as it is crashing on Ubuntu
18.04 because of some NULL pointer checking and use of legacy macros
like: #define _BSD_SOURCE. These problem are fixed on the master branch
but there is no new tagged version release from past 4 years

- Update pip dependencies: move to a newer version cffi, cryptography
and pyopenssl packages

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1505)


kormat:
Hi there. Thanks for your PR :) There are a couple of things about it i need to discuss however. 

The first is that we don't really have the resources to support more than one distro+release (i.e. ubuntu 16.04) at a time, for the moment. The second is that we _are_ going to move to ubuntu 18.04, probably in about 2 months. I basically want to wait a little while for the more obvious bugs that happen in any fresh release to be fixed.

What does this mean for your PR? I'm happy to keep it open until we do the switch, but i would rather not have the multi-os-support approach in it, as it implies a lot more than we can reasonably support.

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1505#-:-L9QT7Of2udRNnucnqvD:bo7d57s)*

scrye:
Closing due to age.

Issue #1493, opened 2018-04-06
Go CS needs to support config & topology reloading


kormat:
This is part of #983.

PR #1510, opened 2018-04-10
 CSGO: Support config reload

Oncilla:
Enable config reload on go certificate server (#1493).
To reload the configuration send a SIGHUP signal to the certificate server process.

Supported:
- Topology reload
- PublicAddr/BindAddr reload
- KeyConf reload
- TrustStore reload
- Customers reload

Not supported:
- Public IA change
- Graceful key rollover. I.e. CS does not check if the new keys are authenticated by any available certificate chain (or fetch a new chain to support the new keys).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1510)


Oncilla:

---

*[go/cert_srv/setup.go, line 46 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9t61kcDONGCYHg3YzW:-L9tAdXD0ka1zBcPtZq8:bb9pubv) ([raw file](https://github.com/scionproto/scion/blob/6a9976ef1dcff9f0d0d5c2633637b07aa7353bbd/go/cert_srv/setup.go#L46)):*

Doesn't this immediately break functionality?
</blockquote></details>

It only breaks functionality for the short-time that the config is reloaded.
I think this is necessary.

Scenario:
Key of AS A is leaked and needs to be removed/replace from the customer mapping.
During the reload AS B updates its verifying key (which is completely legal).
If we do not block customers for this short amount of time, the update might be written after the new customers mapping has been deleted.
=> AS B will no longer be able to update its verifying key, until the config is reloaded again.

We could hand-off the customer mapping from oldCtx to newCtx. To do this properly, we either need more granular signaling (i.e. one for general update, one for customer mapping update) or need to check what has changed in the customers dir.

I think blocking the customers mapping for a short time is a lot more sane. 

---

*[go/cert_srv/setup.go, line 85 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9t6REh8-patIBHfKod:-L9tDB7g--HhvItZyTSy:b-zfiewu) ([raw file](https://github.com/scionproto/scion/blob/6a9976ef1dcff9f0d0d5c2633637b07aa7353bbd/go/cert_srv/setup.go#L85)):*

What is the difference between `conf` and `csctx`? How do you decide what belongs to each? (and yeah, i just noticed we have the same structure in the BR :)
</blockquote></details>

I thought about leaving leaving `Store` and `TrustDB` in the config.
But then I decided to decouple them.

Thinking about it, there is no clear distinguishing element. I guess it makes sense to leave them inside config.
My reasoning was, that `Store` and `TrustDB` are constantly evolving, but `config` should be more static. 
(But by that logic, `Customers`, `singer` and `verifier` should be part of the context as well).

Also, overloading ctx seems bad.

I will change it back, if you agree.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1510)*

Oncilla:
Review status: 1 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[go/cert_srv/io.go, line 51 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9t7Wq994Sgtkpmrlqj:-L9tg64mAc9gSHMmPdRT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6a9976ef1dcff9f0d0d5c2633637b07aa7353bbd/go/cert_srv/io.go#L51)):*

As both of these channels will never be written to, they can be unbuffered.
</blockquote></details>

Done.

---

*[go/cert_srv/io.go, line 54 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9t7fWq0yD18peqzZQX:-L9tg6jvDsLzEJP6Bb-6:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6a9976ef1dcff9f0d0d5c2633637b07aa7353bbd/go/cert_srv/io.go#L54)):*

This is the zero value, so it doesn't need to be specified.
</blockquote></details>

Done.

---

*[go/cert_srv/setup.go, line 45 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9t5wmR2qPtBMxQsbuj:-L9tg8U6C7aOOgjcdlkk:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6a9976ef1dcff9f0d0d5c2633637b07aa7353bbd/go/cert_srv/setup.go#L45)):*

Why close the old one? The trustdb is not something that should change on a reload.
</blockquote></details>

Done.

---

*[go/cert_srv/setup.go, line 46 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9t61kcDONGCYHg3YzW:-L9txAEpC88TnoZkUIpf:bcn9qgh) ([raw file](https://github.com/scionproto/scion/blob/6a9976ef1dcff9f0d0d5c2633637b07aa7353bbd/go/cert_srv/setup.go#L46)):*

It only breaks functionality for the short-time that the config is reloaded.
I think this is necessary.

Scenario:
Key of AS A is leaked and needs to be removed/replace from the customer mapping.
During the reload AS B updates its verifying key (which is completely legal).
If we do not block customers for this short amount of time, the update might be written after the new customers mapping has been deleted.
=> AS B will no longer be able to update its verifying key, until the config is reloaded again.

We could hand-off the customer mapping from oldCtx to newCtx. To do this properly, we either need more granular signaling (i.e. one for general update, one for customer mapping update) or need to check what has changed in the customers dir.

I think blocking the customers mapping for a short time is a lot more sane. 




</blockquote></details>

During refactoring, the time period where customers is blocked is reduced heavily.

---

*[go/cert_srv/setup.go, line 85 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9t6REh8-patIBHfKod:-L9txPYH5et_96vmLyqu:b-ds1uba) ([raw file](https://github.com/scionproto/scion/blob/6a9976ef1dcff9f0d0d5c2633637b07aa7353bbd/go/cert_srv/setup.go#L85)):*

I thought about leaving leaving `Store` and `TrustDB` in the config.
But then I decided to decouple them.

Thinking about it, there is no clear distinguishing element. I guess it makes sense to leave them inside config.
My reasoning was, that `Store` and `TrustDB` are constantly evolving, but `config` should be more static. 
(But by that logic, `Customers`, `singer` and `verifier` should be part of the context as well).

Also, overloading ctx seems bad.

I will change it back, if you agree.




</blockquote></details>

I merged `conf` and `ctx`.

---

*[go/cert_srv/conf/customers.go, line 53 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9t890x5rt_NpGil-qh:-L9tv56OBcCmGV-6baBb:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6a9976ef1dcff9f0d0d5c2633637b07aa7353bbd/go/cert_srv/conf/customers.go#L53)):*

Line length.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1510)*

kormat:
Reviewed 7 of 8 files at r2.

---

*[go/cert_srv/setup.go, line 46 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9t61kcDONGCYHg3YzW:-L9u0z8TCrgsizXcavO_:b5uz1ny) ([raw file](https://github.com/scionproto/scion/blob/6a9976ef1dcff9f0d0d5c2633637b07aa7353bbd/go/cert_srv/setup.go#L46)):*

During refactoring, the time period where customers is blocked is reduced heavily.
</blockquote></details>

Suggestion - just hand over Customers instance from old to new configs.

---

*[go/cert_srv/conf/conf.go, line 61 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9tz3UH95vqWX8ybf9t:-L9tz3UH95vqWX8ybf9u:byiz5h9) ([raw file](https://github.com/scionproto/scion/blob/061b79dd79b92745a4252bbb63a096cfe3cf2058/go/cert_srv/conf/conf.go#L61)):*
> ```Go
> 	Customers *Customers
> 	// ID is the element ID.
> 	ID string
> ```

This should be beside `Topo`. Probably above.

---

*[go/cert_srv/conf/conf.go, line 78 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9tzbB16GkSujo9C5in:-L9tzbB16GkSujo9C5io:b-sq0sh4) ([raw file](https://github.com/scionproto/scion/blob/061b79dd79b92745a4252bbb63a096cfe3cf2058/go/cert_srv/conf/conf.go#L78)):*
> ```Go
> }
> 
> var conf atomic.Value
> ```

This shouldn't be between the type definition and its constructors/methods. I suggest moving it below.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1510)*

Oncilla:
Review status: 6 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[go/cert_srv/setup.go, line 46 at r1](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9t61kcDONGCYHg3YzW:-L9u46Wd6gulRwaki0dT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6a9976ef1dcff9f0d0d5c2633637b07aa7353bbd/go/cert_srv/setup.go#L46)):*

Suggestion - just hand over Customers instance from old to new configs.
</blockquote></details>

Done.

---

*[go/cert_srv/conf/conf.go, line 61 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9tz3UH95vqWX8ybf9t:-L9u45za9O-4HEmgfBc-:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/061b79dd79b92745a4252bbb63a096cfe3cf2058/go/cert_srv/conf/conf.go#L61)):*

This should be beside `Topo`. Probably above.
</blockquote></details>

Done.

---

*[go/cert_srv/conf/conf.go, line 78 at r2](https://beta.reviewable.io/reviews/scionproto/scion/1510#-L9tzbB16GkSujo9C5in:-L9u6ONV2iP1k9cEStoc:b-9lgqbl) ([raw file](https://github.com/scionproto/scion/blob/061b79dd79b92745a4252bbb63a096cfe3cf2058/go/cert_srv/conf/conf.go#L78)):*

This shouldn't be between the type definition and its constructors/methods. I suggest moving it below.
</blockquote></details>

Did not find suitable spot. To the bottom it goes.
Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1510)*

kormat:
:lgtm:

---

Reviewed 2 of 2 files at r3.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/scionproto/scion/1510#-:-L9u79Q71NMl8Or9ZZ-k:bnfp4nl)*

PR #1551, opened 2018-05-08
SIG: Add base events that modules can subscribe to

shitz:
Types of events that are being added:
* `NetworkChanged`: `NetworkChanged` is called when a remote network was added or removed from the configuration.
* `SigChanged`: `SigChanged` is called when a remote SIG is was added or removed from the configuration. Note, this event does not get generated yet.
* `RemoteHealthChanged`: `RemoteHealthChanged` is called when the reachability status of a remote AS changed.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1551)

### DIFF
--- a/go/sig/base/as.go
+++ b/go/sig/base/as.go
@@ -32,17 +32,21 @@ import (
 	"github.com/scionproto/scion/go/sig/siginfo"
 )
 
-const sigMgrTick = 10 * time.Second
+const (
+	sigMgrTick        = 10 * time.Second
+	healthMonitorTick = 5 * time.Second
+)
 
 // ASEntry contains all of the information required to interact with a remote AS.
 type ASEntry struct {
 	sync.RWMutex
-	Nets       map[string]*net.IPNet
-	Sigs       *siginfo.SigMap
-	IA         addr.IA
-	IAString   string
-	egressRing *ringbuf.Ring
-	sigMgrStop chan struct{}
+	Nets              map[string]*net.IPNet
+	Sigs              *siginfo.SigMap
+	IA                addr.IA
+	IAString          string
+	egressRing        *ringbuf.Ring
+	sigMgrStop        chan struct{}
+	healthMonitorStop chan struct{}
 	log.Logger
 
 	Session *egress.Session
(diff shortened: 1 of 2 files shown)


kormat:
Reviewed 2 of 2 files at r1.

---

*[go/sig/base/as.go, line 292 at r1](https://reviewable.io/reviews/scionproto/scion/1551#-LBzikP5DTy8J0pPfGwV:-LBzikP6BwF27bb0vHx6:b-pdpewm) ([raw file](https://github.com/scionproto/scion/blob/dfee8802249151999292a0c5ba3610e31871a44a/go/sig/base/as.go#L292)):*
> ```Go
> 			break Top
> 		case <-ticker.C:
> 			ae.RLock()
> ```

I'm a bit concerned about this. If this go routine somehow exits/panics while holding this lock, the SIG will break. It might be worth putting the contents of this `case` into a separate function, that can then use `Rlock; defer RUnlock`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1551)*

shitz:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[go/sig/base/as.go, line 292 at r1](https://reviewable.io/reviews/scionproto/scion/1551#-LBzikP5DTy8J0pPfGwV:-LBzn7Vs5N1W_al7XcX1:b-o7vuhc) ([raw file](https://github.com/scionproto/scion/blob/dfee8802249151999292a0c5ba3610e31871a44a/go/sig/base/as.go#L292)):*

I'm a bit concerned about this. If this go routine somehow exits/panics while holding this lock, the SIG will break. It might be worth putting the contents of this `case` into a separate function, that can then use `Rlock; defer RUnlock`.
</blockquote></details>

Good catch! Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1551)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1551#-:-LBznczE6KLpy9tch4Gn:bnfp4nl)*

PR #1556, opened 2018-05-09
Quit when Accept returns error.

juagargi:
As the quic implementation always returns the previous error when calling Accept, we have to avoid calling it after we see an error.

We can reproduce the issue by running pingpong as server, and then stopping SCION. It will output the same error message as if there was no tomorrow. The cause of the huge pingpong logs in SCIONlab is this one, although the reason `Accept` fails is probably other than stopping SCION.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1556)

### DIFF
--- a/go/examples/pingpong/pingpong.go
+++ b/go/examples/pingpong/pingpong.go
@@ -244,9 +244,8 @@ func Server() {
 		qsess, err := qsock.Accept()
 		if err != nil {
 			log.Error("Unable to accept quic session", "err", err)
-			// the quic implementation keeps the error present for the
-			// rest of the life of the socket. We quit
-			return
+			// Accept failing means the socket is unusable.
+			break
 		}
 		log.Debug("Quic session accepted", "src", qsess.RemoteAddr())
 		go handleClient(qsess)


kormat:
Reviewed 1 of 1 files at r1.

---

*[go/examples/pingpong/pingpong.go, line 247 at r1](https://reviewable.io/reviews/scionproto/scion/1556#-LC3CfmoEnygVvD2fCr4:-LC3CfmoEnygVvD2fCr5:b-b34dwv) ([raw file](https://github.com/scionproto/scion/blob/5d994f89ce1cb83544bee656311d5d04c89f9873/go/examples/pingpong/pingpong.go#L247)):*
> ```Go
> 		if err != nil {
> 			log.Error("Unable to accept quic session", "err", err)
> 			// the quic implementation keeps the error present for the
> ```

The comment isn't really identifying the actual issue. I suggest:
```
// Accept failing means the socket is unusable.
```

---

*[go/examples/pingpong/pingpong.go, line 249 at r1](https://reviewable.io/reviews/scionproto/scion/1556#-LC3DbkP9qi8haIuJ6Is:-LC3DbkP9qi8haIuJ6It:b-7rmk9e) ([raw file](https://github.com/scionproto/scion/blob/5d994f89ce1cb83544bee656311d5d04c89f9873/go/examples/pingpong/pingpong.go#L249)):*
> ```Go
> 			// the quic implementation keeps the error present for the
> 			// rest of the life of the socket. We quit
> 			return
> ```

I would use `break` rather than `return`, in case any code is added at the end of the function for shutdown.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1556)*

juagargi:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/examples/pingpong/pingpong.go, line 247 at r1](https://reviewable.io/reviews/scionproto/scion/1556#-LC3CfmoEnygVvD2fCr4:-LC3Isyq9iW4ShjsVRFT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5d994f89ce1cb83544bee656311d5d04c89f9873/go/examples/pingpong/pingpong.go#L247)):*

The comment isn't really identifying the actual issue. I suggest:
```
// Accept failing means the socket is unusable.
```
</blockquote></details>

Done.

---

*[go/examples/pingpong/pingpong.go, line 249 at r1](https://reviewable.io/reviews/scionproto/scion/1556#-LC3DbkP9qi8haIuJ6Is:-LC3IsXQAL2aYQ1xYcEN:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5d994f89ce1cb83544bee656311d5d04c89f9873/go/examples/pingpong/pingpong.go#L249)):*

I would use `break` rather than `return`, in case any code is added at the end of the function for shutdown.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1556)*

kormat:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1556#-:-LC3P2A_6GMFeypNHava:bnfp4nl)*

Issue #1183, opened 2017-07-26
AS Certificate Reissuance


Oncilla:
### AS Certificate expiration
Certificates are expected to expire in the order of days. Thus a mechanism for Certificate renewal shall be in place for seamless transition between validity periods.

### Certificate Reissuance Request
A Certificate Server proactively requests a renewed Certificate from the responsible Core AS (I.e. the issuer off the current Certificate). A CertReissueReq is signed with the private key of the current Certificate, to ensure the requester is still in possession of the private key. The request is handled by one of the Core AS' CS.

### Certificate Dissemination
Certificates are disseminated by the beacon server to the neighboring ASes as well as to the local Path Servers. An important observation is, that the new Certificate is only needed if ExpTime of new HopEntries is after the current Certificate's expiration time. 
The Beacon server (pre-)fetches the newest Certificate from the CS, such that it can sign new HopEntries. If for some reason the Certificate is not available, the BS has to adjust the HopEntry's expiration time to still fit into the current Certfificate's validity period.

shitz:
@Oncilla is the work here done?

Oncilla:
The issue can be closed. The go implementation now supports certificate reissuance.

Currently, the default topology uses the CS:PY and not CS:GO. Moving over will be done in a follow-up issue.

PR #1558, opened 2018-05-09
CSGO: Add Certificate Reissuance

Oncilla:
This PR adds certificate reissuance to the infrastructure. (#1183)

Core ASes periodically renew leaf AS and issuer AS certificates.
Non-Core ASes periodically request new certificate chains from their
responsible core AS.
Both these tasks are done on the Certificate Server

Request must be verifiable with the newest verifying key in the
customer mapping of the core AS.

The non-core Certificate Servers start to automatically request
reissued certificate chains one path segment validity time before
expiration to provide optimal coverage.

The reissued certificate chains are disseminated by the Beacon server.
They start polling the certificate server for new certificate chains
when the available certificate no longer covers the desired segment
validity period.

Missing:
Useful integration test

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1558)

### DIFF
--- a/go/cert_srv/cert.go
+++ b/go/cert_srv/cert.go
@@ -46,6 +46,9 @@ func (h *ChainHandler) HandleReq(a *snet.Addr, req *cert_mgmt.ChainReq, config *
 	var chain *cert.Chain
 	if req.Version == cert_mgmt.NewestVersion {
 		chain = config.Store.GetNewestChain(req.IA())
+		if chain != nil && chain.Leaf.VerifyTime(uint64(time.Now().Unix())) != nil {
+			chain = nil
+		}
 	} else {
 		chain = config.Store.GetChain(req.IA(), req.Version)
 	}
@@ -108,6 +111,7 @@ func (h *ChainHandler) HandleRep(a *snet.Addr, rep *cert_mgmt.Chain, config *con
 	chain, err := rep.Chain()
 	if err != nil {
 		log.Error("Unable to parse certificate reply", "err", err)
+		return
 	}
 	if err = config.Store.AddChain(chain, true); err != nil {
 		log.Error("Unable to store certificate chain", "key", chain.Key(), "err", err)
--- a/go/cert_srv/conf/conf.go
+++ b/go/cert_srv/conf/conf.go
@@ -18,9 +18,11 @@ import (
 	"path/filepath"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/scionproto/scion/go/lib/addr"
 	"github.com/scionproto/scion/go/lib/common"
+	"github.com/scionproto/scion/go/lib/crypto/cert"
 	"github.com/scionproto/scion/go/lib/ctrl"
 	"github.com/scionproto/scion/go/lib/infra/modules/trust/trustdb"
 	"github.com/scionproto/scion/go/lib/snet"
@@ -29,7 +31,17 @@ import (
 )
 
 const (
+	// LeafReissTime is the default value for Conf.LeafReissTime. It is the same
+	// as the default segment TTL in order to provide optimal coverage.
+	LeafReissTime = 6 * time.Hour
+	// IssuerReissTime is the default value for Conf.IssuerReissTime. It is the same
+	// as the leaf certificate validity period in order to provide optimal coverage.
+	IssuerReissTime = cert.DefaultLeafCertValidity * time.Second
+	// ReissReqRate is the default interval between two consecutive reissue requests.
+	ReissReqRate = 10 * time.Second
+
 	ErrorAddr      = "Unable to load addresses"
+	ErrorIssCert   = "Unable to load issuer certificate"
 	ErrorKeyConf   = "Unable to load KeyConf"
 	ErrorConfNil   = "Unable to reload conf from nil value"
 	ErrorStore     = "Unable to load TrustStore"
@@ -73,15 +85,24 @@ type Conf struct {
 	verifier ctrl.SigVerifier
 	// verifierLock guards verifier.
 	verifierLock sync.RWMutex
+	// LeafReissTime is the time between starting reissue requests and leaf cert expiration.
+	LeafReissTime time.Duration
+	// IssuerReissTime is the time between self issuing core cert and core cert expiration.
+	IssuerReissTime time.Duration
+	// ReissRate is the interval between two consecutive reissue requests.
+	ReissRate time.Duration
 }
 
 // Load initializes the configuration by loading it from confDir.
 func Load(id string, confDir string, cacheDir string, stateDir string) (*Conf, error) {
 	c := &Conf{
-		ID:       id,
-		ConfDir:  confDir,
-		CacheDir: cacheDir,
-		StateDir: stateDir,
+		ID:              id,
+		ConfDir:         confDir,
+		CacheDir:        cacheDir,
+		StateDir:        stateDir,
+		LeafReissTime:   LeafReissTime,
+		IssuerReissTime: IssuerReissTime,
+		ReissRate:       ReissReqRate,
 	}
 	if err := c.loadTopo(); err != nil {
 		return nil, err
@@ -100,6 +121,9 @@ func Load(id string, confDir string, cacheDir string, stateDir string) (*Conf, e
 		if c.Customers, err = c.LoadCustomers(); err != nil {
 			return nil, common.NewBasicError(ErrorCustomers, err)
 		}
+		if err = c.loadIssCert(); err != nil {
+			return nil, err
+		}
 	}
 	return c, nil
 }
@@ -112,12 +136,15 @@ func ReloadConf(oldConf *Conf) (*Conf, error) {
 	// FIXME(roosd): Changing keys for customers outside of the process on-disk
 	// requires a restart of the certificate server in order to be visible.
 	c := &Conf{
-		ID:        oldConf.ID,
-		TrustDB:   oldConf.TrustDB,
-		Customers: oldConf.Customers,
-		ConfDir:   oldConf.ConfDir,
-		CacheDir:  oldConf.CacheDir,
-		StateDir:  oldConf.StateDir,
+		ID:              oldConf.ID,
+		TrustDB:         oldConf.TrustDB,
+		Customers:       oldConf.Customers,
+		ConfDir:         oldConf.ConfDir,
+		CacheDir:        oldConf.CacheDir,
+		StateDir:        oldConf.StateDir,
+		LeafReissTime:   LeafReissTime,
+		IssuerReissTime: IssuerReissTime,
+		ReissRate:       ReissReqRate,
 	}
 	if err := c.loadTopo(); err != nil {
 		return nil, err
@@ -127,6 +154,11 @@ func ReloadConf(oldConf *Conf) (*Conf, error) {
 	}
 	if err := c.loadKeyConf(); err != nil {
 		return nil, err
+	}
+	if c.Topo.Core {
+		if err := c.loadIssCert(); err != nil {
+			return nil, err
+		}
 	}
 	return c, nil
 }
@@ -181,11 +213,30 @@ func (c *Conf) loadKeyConf() (err error) {
 	return nil
 }
 
+// loadKeyConf inserts the issuer certificate of the newest certificate chain into the trustdb.
+func (c *Conf) loadIssCert() error {
+	chain := c.Store.GetNewestChain(c.PublicAddr.IA)
+	if chain == nil {
+		return common.NewBasicError(ErrorIssCert, nil, "err", "No certificate chain present")
+	}
+	if _, err := c.TrustDB.InsertIssCert(chain.Issuer); err != nil {
+		return common.NewBasicError(ErrorIssCert, err)
+	}
+	return nil
+}
+
 // GetSigningKey returns the signing key of the current key configuration.
 func (c *Conf) GetSigningKey() common.RawBytes {
 	c.keyConfLock.RLock()
 	defer c.keyConfLock.RUnlock()
 	return c.keyConf.SignKey
+}
+
+// GetIssSigningKey returns the issuer signing key of the current key configuration.
+func (c *Conf) GetIssSigningKey() common.RawBytes {
+	c.keyConfLock.RLock()
+	defer c.keyConfLock.RUnlock()
+	return c.keyConf.IssSigKey
 }
 
 // GetDecryptKey returns the decryption key of the current key configuration.
--- a/go/cert_srv/conf/conf.go
+++ b/go/cert_srv/conf/conf.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/scionproto/scion/go/lib/addr"
+	"github.com/scionproto/scion/go/lib/as_conf"
 	"github.com/scionproto/scion/go/lib/common"
 	"github.com/scionproto/scion/go/lib/crypto/cert"
 	"github.com/scionproto/scion/go/lib/ctrl"
@@ -31,9 +32,6 @@ import (
 )
 
 const (
-	// LeafReissTime is the default value for Conf.LeafReissTime. It is the same
-	// as the default segment TTL in order to provide optimal coverage.
-	LeafReissTime = 6 * time.Hour
 	// IssuerReissTime is the default value for Conf.IssuerReissTime. It is the same
 	// as the leaf certificate validity period in order to provide optimal coverage.
 	IssuerReissTime = cert.DefaultLeafCertValidity * time.Second
@@ -100,9 +98,11 @@ func Load(id string, confDir string, cacheDir string, stateDir string) (*Conf, e
 		ConfDir:         confDir,
 		CacheDir:        cacheDir,
 		StateDir:        stateDir,
-		LeafReissTime:   LeafReissTime,
 		IssuerReissTime: IssuerReissTime,
 		ReissRate:       ReissReqRate,
+	}
+	if err := c.loadLeafReissTime(); err != nil {
+		return nil, err
 	}
 	if err := c.loadTopo(); err != nil {
 		return nil, err
@@ -142,9 +142,11 @@ func ReloadConf(oldConf *Conf) (*Conf, error) {
 		ConfDir:         oldConf.ConfDir,
 		CacheDir:        oldConf.CacheDir,
 		StateDir:        oldConf.StateDir,
-		LeafReissTime:   LeafReissTime,
 		IssuerReissTime: IssuerReissTime,
 		ReissRate:       ReissReqRate,
+	}
+	if err := c.loadLeafReissTime(); err != nil {
+		return nil, err
 	}
 	if err := c.loadTopo(); err != nil {
 		return nil, err
@@ -210,6 +212,16 @@ func (c *Conf) loadKeyConf() (err error) {
 	if err != nil {
 		return common.NewBasicError(ErrorKeyConf, err)
 	}
+	return nil
+}
+
+// loadLeafReissTime loads the as conf and sets the LeafReissTime to the PathSegmentTTL
+// to provide optimal coverage.
+func (c *Conf) loadLeafReissTime() error {
+	if err := as_conf.Load(filepath.Join(c.ConfDir, as_conf.CfgName)); err != nil {
+		return err
+	}
+	c.LeafReissTime = time.Duration(as_conf.CurrConf.PathSegmentTTL) * time.Second
 	return nil
 }
 
--- a/go/cert_srv/io.go
+++ b/go/cert_srv/io.go
@@ -80,7 +80,7 @@ func (d *Dispatcher) Run() {
 	}
 }
 
-// dispatch hands payload over tho the associated handlers.
+// dispatch hands payload over to the associated handlers.
 func (d *Dispatcher) dispatch(addr *snet.Addr, buf common.RawBytes, config *conf.Conf) error {
 	signed, err := ctrl.NewSignedPldFromRaw(buf)
 	if err != nil {
--- a/go/cert_srv/io.go
+++ b/go/cert_srv/io.go
@@ -22,6 +22,7 @@ import (
 	"github.com/scionproto/scion/go/lib/ctrl/cert_mgmt"
 	"github.com/scionproto/scion/go/lib/log"
 	"github.com/scionproto/scion/go/lib/snet"
+	"github.com/scionproto/scion/go/proto"
 )
 
 const MaxReadBufSize = 2 << 16
@@ -85,11 +86,8 @@ func (d *Dispatcher) dispatch(addr *snet.Addr, buf common.RawBytes, config *conf
 	if err != nil {
 		return common.NewBasicError("Unable to parse signed payload", err, "addr", addr)
 	}
-	if signed.Sign != nil {
-		verifier := config.GetVerifier()
-		if err := ctrl.VerifySig(signed, verifier); err != nil {
-			return common.NewBasicError("Unable to verify signed payload", err, "addr", addr)
-		}
+	if err := d.verifySign(signed, config.GetVerifier(), addr); err != nil {
+		return common.NewBasicError("Unable to verify sign", err, "addr", addr)
 	}
 	cpld, err := signed.Pld()
 	if err != nil {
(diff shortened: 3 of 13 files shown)


kormat:
Reviewed 7 of 8 files at r1.
Review status: 7 of 8 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[go/cert_srv/io.go, line 109 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD78t4I9ClvgrLYMnpb:-LD78t4I9ClvgrLYMnpc:b3rgn5h) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/io.go#L109)):*
> ```Go
> 			return common.NewBasicError("Unable to unpack cert_mgmt union", err)
> 		}
> 		switch pld.(type) {
> ```

This can be simplified a bit:
```
		switch pld := pld.(type) {
		case *cert_mgmt.ChainIssReq:
			d.reissHandler.HandleReq(addr, pld, signed, config)
		case *cert_mgmt.ChainIssRep:
			d.reissHandler.HandleRep(addr, pld, config)
		case *cert_mgmt.Chain:
			d.chainHandler.HandleRep(addr, pld, config)
		case *cert_mgmt.ChainReq:
			d.chainHandler.HandleReq(addr, pld, config)
		case *cert_mgmt.TRC:
			d.trcHandler.HandleRep(addr, pld, config)
		case *cert_mgmt.TRCReq:
			d.trcHandler.HandleReq(addr, pld, config)
```
It's a common idiom in Go.

---

*[go/cert_srv/reiss_handler.go, line 21 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD79cN210Aweauuf-j_:-LD79cN210Aweauuf-ja:bku9hi) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L21)):*
> ```Go
> 	"time"
> 
> 	log "github.com/inconshreveable/log15"
> ```

Use `github.com/scionproto/scion/go/lib/log` instead now (see https://github.com/scionproto/scion/pull/1546 for context).

---

*[go/cert_srv/reiss_handler.go, line 40 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7IgeL4xnyvFADRK-I:-LD7IgeL4xnyvFADRK-J:b-irmcpd) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L40)):*
> ```Go
> // automatically by the issuer ASes.
> type ReissHandler struct {
> 	conn *snet.Conn
> ```

As `config *conf.Conf` is a parameter for most methods in this type, what about just adding it to the ReissHandler struct?

---

*[go/cert_srv/reiss_handler.go, line 80 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7NFxUENdy6LZFWDaF:-LD7NFxUENdy6LZFWDaG:b-n2331f) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L80)):*
> ```Go
> 		}
> 		return
> 	}
> ```

There should be a check to ensure that `maxChain == crt.Version -1`, otherwise a customer can request any version they like.

---

*[go/cert_srv/reiss_handler.go, line 106 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7QFKi5O-0pYad_J9z:-LD7QFKi5O-0pYad_JA-:b-62hgzg) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L106)):*
The comment where this is called says:
> // Verify the request was correctly signed by the requester

Which is subtly different to this comment/implementation. Suggestion: rename it to `validateSign`, that way it's clearer that it's not doing signature verification (as you pointed out, that's already done by the dispatcher).

---

*[go/cert_srv/reiss_handler.go, line 116 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7D7NT-_hZLr1Pbggp:-LD7D7NT-_hZLr1Pbggq:b-y950u6) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L116)):*
> ```Go
> 	if err != nil {
> 		return nil, err
> 	}
> ```

There's a check missing: `src.IA.Eq(addr.IA)`. Actually, this should be done in `Dispatcher.dispatch()`.

---

*[go/cert_srv/reiss_handler.go, line 117 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7Vkno3zwX4ZHsKXCb:-LD7Vkno3zwX4ZHsKXCc:bn3p8az) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L117)):*
> ```Go
> 		return nil, err
> 	}
> 	verChain, err := config.GetVerifier().(*SigVerifier).getChainForSign(src)
> ```

This is a bit awkward. Skimming the code of `getChainForSign`, it looks like it doesn't access the parent type at all. That sounds like an opportunity to move it to a standalone function, which would also make this use cleaner.

---

*[go/cert_srv/reiss_handler.go, line 122 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7Ep926aIey5VadRMM:-LD7Ep926aIey5VadRMN:b-iotzov) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L122)):*
> ```Go
> 	}
> 	if signed.Sign.Type.String() != verChain.Leaf.SignAlgorithm {
> 		return nil, common.NewBasicError("Invalid sign type", nil, "type", signed.Sign.Type)
> ```

Always log what the expected value was, too. `"expected", verChain.Leaf.SignAlgorithm, "actual", signed.Sign.Type)`.

---

*[go/cert_srv/reiss_handler.go, line 134 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7NXkF6bqlodi27lVf:-LD7NXkF6bqlodi27lVg:b99iurc) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L134)):*
> ```Go
> // validateReq validates the requested certificate. Additionally, it validates that
> // the request was verified with the same verifying key as in the customer mapping.
> func (h *ReissHandler) validateReq(c *cert.Certificate, key common.RawBytes,
> ```

`vKey` would match the naming in `issueChain`, and make it a bit easier to follow. Maybe `vChain` too, for the same reason.

---

*[go/cert_srv/reiss_handler.go, line 172 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7H1J5438-45VsVRz5:-LD7H1J5438-45VsVRz6:b364xhz) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L172)):*
> ```Go
> 		chain.Leaf.ExpirationTime = chain.Issuer.ExpirationTime
> 	}
> 	if err = chain.Leaf.Sign(config.GetIssSigningKey(), crypto.Ed25519); err != nil {
> ```

`crypto.Ed25519` - can this be replaced with something like `chain.Issuer.SignAlgorithm`?

---

*[go/cert_srv/reiss_handler.go, line 197 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7WwCD0HfbMQJhmAPx:-LD7WwCD0HfbMQJhmAPy:b-h2jhj6) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L197)):*
> ```Go
> 	}
> 	log.Info("Send reissued certificate chain", "chain", chain, "addr", addr)
> 	return SendPayload(h.conn, cpld, addr)
> ```

Would it make sense to sign the response too?

---

*[go/cert_srv/reiss_handler.go, line 228 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7XEC49F__Rh0dGnA5:-LD7XEC49F__Rh0dGnA6:b-7l46ps) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L228)):*
> ```Go
> }
> 
> // validateRep validates that the received certificate chain can be added to the trust store.
> ```

Is the signature on the new cert verified anywhere?

---

*[go/cert_srv/conf/conf.go, line 35 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD773vE2AsSZvWC8IFB:-LD773vE2AsSZvWC8IFC:bm3sgbj) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/conf/conf.go#L35)):*
> ```Go
> const (
> 	// LeafReissTime is the default value for Conf.LeafReissTime. It is the same
> 	// as the default segment TTL in order to provide optimal coverage.
> ```

The segment TTL is a per-AS configuration choice. It's the `PathSegmentTTL` value in `as.yml`.

---

*[python/beacon_server/base.py, line 785 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD6qk6mCR08xhMO53MZ:-LD6qk6mCR08xhMO53M_:b-6s0x19) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/python/beacon_server/base.py#L785)):*
> ```Python
>             diff = exp - int(time.time())
>             if diff > self.config.segment_ttl:
>                 time.sleep(max(0, diff - self.config.segment_ttl))
> ```

I don't think the `max()` is necessary, the `if` condition already covers that.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1558)*

Oncilla:
Review status: 0 of 11 files reviewed at latest revision, 14 unresolved discussions.

---

*[go/cert_srv/io.go, line 109 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD78t4I9ClvgrLYMnpb:-LDB0Nye-r8rlcktkMw9:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/io.go#L109)):*

This can be simplified a bit:
```
		switch pld := pld.(type) {
		case *cert_mgmt.ChainIssReq:
			d.reissHandler.HandleReq(addr, pld, signed, config)
		case *cert_mgmt.ChainIssRep:
			d.reissHandler.HandleRep(addr, pld, config)
		case *cert_mgmt.Chain:
			d.chainHandler.HandleRep(addr, pld, config)
		case *cert_mgmt.ChainReq:
			d.chainHandler.HandleReq(addr, pld, config)
		case *cert_mgmt.TRC:
			d.trcHandler.HandleRep(addr, pld, config)
		case *cert_mgmt.TRCReq:
			d.trcHandler.HandleReq(addr, pld, config)
```
It's a common idiom in Go.
</blockquote></details>

Done.

---

*[go/cert_srv/reiss_handler.go, line 21 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD79cN210Aweauuf-j_:-LDB0rNr2ojs4W3xcAqc:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L21)):*

Use `github.com/scionproto/scion/go/lib/log` instead now (see https://github.com/scionproto/scion/pull/1546 for context).
</blockquote></details>

Done.

---

*[go/cert_srv/reiss_handler.go, line 40 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7IgeL4xnyvFADRK-I:-LDB1lBd3UKEubDETihl:b-wl7ffy) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L40)):*

As `config *conf.Conf` is a parameter for most methods in this type, what about just adding it to the ReissHandler struct?
</blockquote></details>

I wanted to keep it consistent with the other handlers.

I'm not sure what the correct behavior is during a config reload.
Currently, the same config is used when handling requests, even if a update happens.

---

*[go/cert_srv/reiss_handler.go, line 80 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7NFxUENdy6LZFWDaF:-LDB4_Ag227Jqbx_xW0U:bdyby2n) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L80)):*

There should be a check to ensure that `maxChain == crt.Version -1`, otherwise a customer can request any version they like.
</blockquote></details>

Added to validteReq.
Done.

---

*[go/cert_srv/reiss_handler.go, line 106 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7QFKi5O-0pYad_J9z:-LDB55fKD8hJpg9hKkdc:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L106)):*

The comment where this is called says:
> // Verify the request was correctly signed by the requester

Which is subtly different to this comment/implementation. Suggestion: rename it to `validateSign`, that way it's clearer that it's not doing signature verification (as you pointed out, that's already done by the dispatcher).
</blockquote></details>

Done.

---

*[go/cert_srv/reiss_handler.go, line 116 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7D7NT-_hZLr1Pbggp:-LDB9Z0I2N0ieiAgINdB:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L116)):*

There's a check missing: `src.IA.Eq(addr.IA)`. Actually, this should be done in `Dispatcher.dispatch()`.
</blockquote></details>

Done.

---

*[go/cert_srv/reiss_handler.go, line 117 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7Vkno3zwX4ZHsKXCb:-LDB9xkZ1OHittWsahFW:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L117)):*

This is a bit awkward. Skimming the code of `getChainForSign`, it looks like it doesn't access the parent type at all. That sounds like an opportunity to move it to a standalone function, which would also make this use cleaner.
</blockquote></details>

Done.

---

*[go/cert_srv/reiss_handler.go, line 122 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7Ep926aIey5VadRMM:-LDBAC-yEfXX4yp73oPU:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L122)):*

Always log what the expected value was, too. `"expected", verChain.Leaf.SignAlgorithm, "actual", signed.Sign.Type)`.
</blockquote></details>

Done.

---

*[go/cert_srv/reiss_handler.go, line 134 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7NXkF6bqlodi27lVf:-LDBAQ9y2wiLCULkzc7t:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L134)):*

`vKey` would match the naming in `issueChain`, and make it a bit easier to follow. Maybe `vChain` too, for the same reason.
</blockquote></details>

Done.

---

*[go/cert_srv/reiss_handler.go, line 172 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7H1J5438-45VsVRz5:-LDBAekj07ZPXq6A2JPE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L172)):*

`crypto.Ed25519` - can this be replaced with something like `chain.Issuer.SignAlgorithm`?
</blockquote></details>

Done.

---

*[go/cert_srv/reiss_handler.go, line 197 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7WwCD0HfbMQJhmAPx:-LDBB_szAr8GG4j74I2g:brc7c64) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L197)):*

Would it make sense to sign the response too?
</blockquote></details>

Done.

(Although, currently this signature is ignored to avoid deadlocks. The trust store needs to support recursive look ups)

---

*[go/cert_srv/reiss_handler.go, line 228 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7XEC49F__Rh0dGnA5:-LDBDpZq-kLzm24KrUNl:b-2io8gw) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L228)):*

Is the signature on the new cert verified anywhere?
</blockquote></details>

Done.

Issuer verifies signature
Requester checks chain is verifiable.

---

*[go/cert_srv/conf/conf.go, line 35 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD773vE2AsSZvWC8IFB:-LDBNQw13TK47bfcV8FI:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/conf/conf.go#L35)):*

The segment TTL is a per-AS configuration choice. It's the `PathSegmentTTL` value in `as.yml`.
</blockquote></details>

Done.

---

*[python/beacon_server/base.py, line 785 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD6qk6mCR08xhMO53MZ:-LDBNbuI9JbRxZApIR2s:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/python/beacon_server/base.py#L785)):*

I don't think the `max()` is necessary, the `if` condition already covers that.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1558)*

kormat:
Reviewed 13 of 13 files at r2.
Review status: all files reviewed, 4 unresolved discussions (waiting on @kormat and @Oncilla)

---

*[go/cert_srv/io.go, line 83 at r2](https://reviewable.io/reviews/scionproto/scion/1558#-LFqRE2OAmo2MvbKM5qr:-LFqRE2OAmo2MvbKM5qs:b-c67e15) ([raw file](https://github.com/scionproto/scion/blob/c3516d41fc16f8d91f7c31aca84f4a15941003af/go/cert_srv/io.go#L83)):*
> ```Go
> }
> 
> // dispatch hands payload over tho the associated handlers.
> ```

`s/tho/to/`

---

*[go/cert_srv/reiss_handler.go, line 40 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7IgeL4xnyvFADRK-I:-LFqReydEqJHbUXmdIhd:bk8850e) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L40)):*

I wanted to keep it consistent with the other handlers.

I'm not sure what the correct behavior is during a config reload.
Currently, the same config is used when handling requests, even if a update happens.




</blockquote></details>

How we handle this in the BR is that at the start of a request the handler grabs a pointer to the current config, and uses that for the current request. A config reload then just changes the shared pointer to point to a new version of the config, and all future requests will use that.

So, yeah, either you create a new ReissHandler per request, or you pass the config into methods like you do currently.

---

*[go/cert_srv/reiss_handler.go, line 106 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7QFKi5O-0pYad_J9z:-LFq_pwC4yRmSe4BllJO:b-il9icv) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L106)):*

Done.
</blockquote></details>

Old name still in docstring.

---

*[go/cert_srv/reiss_tasks.go, line 38 at r2](https://reviewable.io/reviews/scionproto/scion/1558#-LFqcAlACLPSjdaQAvMg:-LFqcAlACLPSjdaQAvMh:b87zote) ([raw file](https://github.com/scionproto/scion/blob/c3516d41fc16f8d91f7c31aca84f4a15941003af/go/cert_srv/reiss_tasks.go#L38)):*
> ```Go
> 
> // Run periodically issues certificate chains for the local AS.
> func (l *SelfIssuer) Run() {
> ```

How did `SelfIssuer` end up with a receiver var `l`? :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1558)*

Oncilla:
Review status: all files reviewed, 4 unresolved discussions (waiting on @Oncilla and @kormat)

---

*[go/cert_srv/io.go, line 83 at r2](https://reviewable.io/reviews/scionproto/scion/1558#-LFqRE2OAmo2MvbKM5qr:-LFvoaYM1yqZJV7VjOSQ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c3516d41fc16f8d91f7c31aca84f4a15941003af/go/cert_srv/io.go#L83)):*

`s/tho/to/`
</blockquote></details>

Done.

---

*[go/cert_srv/reiss_handler.go, line 40 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7IgeL4xnyvFADRK-I:-LFvpaLX9fKrG3WgubA4:brz9fsb) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L40)):*

How we handle this in the BR is that at the start of a request the handler grabs a pointer to the current config, and uses that for the current request. A config reload then just changes the shared pointer to point to a new version of the config, and all future requests will use that.

So, yeah, either you create a new ReissHandler per request, or you pass the config into methods like you do currently.
</blockquote></details>

Hmm, that's a bit inconclusive. What route should we take?

I prefer consistency, if we create a new handler for each request, I will do that for every other handler as well.
(Similar to what lib/infra is doing.)

This also begs the question, will we adapt the CS to use lib/infra at some point?

---

*[go/cert_srv/reiss_handler.go, line 106 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7QFKi5O-0pYad_J9z:-LFvrUpL5rdJJeo5a901:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L106)):*

Old name still in docstring.
</blockquote></details>

Done.

---

*[go/cert_srv/reiss_tasks.go, line 38 at r2](https://reviewable.io/reviews/scionproto/scion/1558#-LFqcAlACLPSjdaQAvMg:-LFvrWiY68010tQLUlEy:b-q7r5xj) ([raw file](https://github.com/scionproto/scion/blob/c3516d41fc16f8d91f7c31aca84f4a15941003af/go/cert_srv/reiss_tasks.go#L38)):*

How did `SelfIssuer` end up with a receiver var `l`? :)
</blockquote></details>

Struggling for a good name and constant renaming will do that ;)
l->s for `SelfIssuer`
l->r for `ReissueRequester`

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1558)*

kormat:
:lgtm:

---

Reviewed 4 of 4 files at r3.
Review status: all files reviewed, 1 unresolved discussion (waiting on @Oncilla)

---

*[go/cert_srv/reiss_handler.go, line 40 at r1](https://reviewable.io/reviews/scionproto/scion/1558#-LD7IgeL4xnyvFADRK-I:-LFvuTur13EoI1tuA4kW:b-hz876t) ([raw file](https://github.com/scionproto/scion/blob/380d1d71bef7db031f377c95fa25126f8bb1ecab/go/cert_srv/reiss_handler.go#L40)):*

Hmm, that's a bit inconclusive. What route should we take?

I prefer consistency, if we create a new handler for each request, I will do that for every other handler as well.
(Similar to what lib/infra is doing.)

This also begs the question, will we adapt the CS to use lib/infra at some point?

</blockquote></details>

Ok, let's keep the current approach, for now at least.

Re: lib/infra, yep, that is the plan when lib/infra is a bit more fleshed-out.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1558#-:-LFvumb98PVmXU-JsX8G:bnfp4nl)*

PR #1560, opened 2018-05-16
Check that ifids match

kormat:
The ifid from a packet must match the ifid of the interface it was
received on. This prevents a number of nasty attacks.

Thanks to Tobias Klenze for spotting this!

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1560)

### DIFF
--- a/go/border/rpkt/validate.go
+++ b/go/border/rpkt/validate.go
@@ -18,6 +18,7 @@ package rpkt
 
 import (
 	"github.com/scionproto/scion/go/lib/addr"
+	"github.com/scionproto/scion/go/lib/assert"
 	"github.com/scionproto/scion/go/lib/common"
 	"github.com/scionproto/scion/go/lib/scmp"
 )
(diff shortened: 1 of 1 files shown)


shitz:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1560#-:-LCcUdcj3ET636zfS2bA:bnfp4nl)*

tklenze:
Hi, thanks for this fix.
I am not sure what rp.Ingress.IfIDs[] actually contains. Why would there be more than one interface from which a packet is received? I would expect any received packet to be attributable to a single interface.

kormat:
@tklenze - this only applies to packets received from the local AS (i.e. _outgoing_ packets). When they are received on an internal socket, it is not automatically known which external interface they are associated with, so the slice contains all of the possible external interfaces associated with that internal socket. For incoming packet, there is only ever a single element in the slice.

Issue #1344, opened 2017-12-04
Path server replies too soon


kormat:
When a path server sends a query, it can reply to the client when it has only processed a single PCB in the response. This causes a severe lack of diversity in response to the first query for a destination.

kormat:
#1280 is the fix for a similar bug in sciond.

lukedirtwalker:
This should not be an issue in the PS:Go

PR #1573, opened 2018-05-28
path_server: handle pending request after PCB processing is complete

jinankjain:
Currently there is a race condition between Path Server main thread and
worker thread. When a sciond queries a local PS for a path from a non
core AS to a core AS. Local PS forwards that request to core PS for the
core segments and starts populating PathSegmentDB. While the main thread
is busy processing that PCB, worker thread handles pending request and
replies with whatever present inside PathSegmentDB. Usually it is just
one path and rest of the possible paths are neglected.

For example:

```
./bin/showpaths -dstIA 2-ff00:0:210 -srcIA 1-ff00:0:133
INFO[05-27|09:38:27] Started                                  id=df259bab goroutine=dispatcher_bck
Available paths to 2-ff00:0:210
[ 0] Hops: [1-ff00:0:133 86>96 1-ff00:0:132 39>73 1-ff00:0:131 91>95 1-ff00:0:130 59>76 1-ff00:0:120 77>55 1-ff00:0:110 57>19 2-ff00:0:210] Mtu: 1280
```

There are 4 possible paths from 133 to 210 but it only returns one. An
easy way to verify that there are more path just flush the cache of
sciond and query again for the same path and you will find that there
are more paths.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1573)

### DIFF
--- a/python/path_server/base.py
+++ b/python/path_server/base.py
@@ -401,6 +401,7 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
             seg_meta = PathSegMeta(pcb, self.continue_seg_processing, meta,
                                    type_, params)
             self._process_path_seg(seg_meta, req_id)
+        self._handle_pending_requests()
 
     def continue_seg_processing(self, seg_meta):
         """
@@ -415,7 +416,6 @@ class PathServer(SCIONElement, metaclass=ABCMeta):
         params = seg_meta.params
         self.handle_ext(pcb)
         self._dispatch_segment_record(type_, pcb, **params)
-        self._handle_pending_requests()
 
     def handle_ext(self, pcb):
         """


shitz:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[python/path_server/base.py, line 404 at r1](https://reviewable.io/reviews/scionproto/scion/1573#-LDbVdfA4gS5t-J64BVp:-LDbVdfA4gS5t-J64BVq:b6cwoc8) ([raw file](https://github.com/scionproto/scion/blob/625bd409d9f2408a12042c78cb0e1e0e2d947b5f/python/path_server/base.py#L404)):*
> ```Python
>                                    type_, params)
>             self._process_path_seg(seg_meta, req_id)
>         self._handle_pending_requests()
> ```

Unfortunately, this is a bit more involved than your proposal. I honestly don't see how your proposed fix improves the situation, since you still have the race condition. Take a look at https://github.com/scionproto/scion/blob/master/python/sciond/sciond.py#L188 to see how this is solved in sciond. Something similar would be needed for the path server, however, we decided that we will solve this issue with the rewrite of the path server in Go.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1573)*

shitz:

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1573)*

shitz:

---

*[python/path_server/base.py, line 404 at r1](https://reviewable.io/reviews/scionproto/scion/1573#-LDbVdfA4gS5t-J64BVp:-LDeffRH6WcVhS9QPdRj:b-7kl3ht) ([raw file](https://github.com/scionproto/scion/blob/625bd409d9f2408a12042c78cb0e1e0e2d947b5f/python/path_server/base.py#L404)):*

Unfortunately, this is a bit more involved than your proposal. I honestly don't see how your proposed fix improves the situation, since you still have the race condition. Take a look at https://github.com/scionproto/scion/blob/master/python/sciond/sciond.py#L188 to see how this is solved in sciond. Something similar would be needed for the path server, however, we decided that we will solve this issue with the rewrite of the path server in Go.
</blockquote></details>

#1344 is the related issue.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1573)*

shitz:
@jinankjain Do you intend to work more on this? Otherwise, I will close it since this PR does not fix the problem.

jinankjain:
@shitz For the time being I am not planning to work on this one. So you can close this one.

PR #1579, opened 2018-05-31
Fix file logging; support for unbuffered logging

scrye:
`log.SetupLogFile` now correctly initializes file logging from function arguments instead of flags.

Also changes the semantics behind a 0 value flush timer from "never flush" to a more intuitive "always flush immediately" (unbuffered logging). The old "never flush" behavior can now be selected by passing in a negative flush timer.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)

### DIFF
--- a/go/lib/log/log.go
+++ b/go/lib/log/log.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"os"
 	"runtime/debug"
+	"strings"
 	"time"
 
 	"github.com/inconshreveable/log15"
@@ -74,15 +75,34 @@ func SetupLogFile(name string, logDir string, logLevel string, logSize int, logA
 	if err != nil {
 		return common.NewBasicError("Unable to parse log.level flag:", err)
 	}
-	logBuf = newSyncBuf(mkLogfile(name))
+
+	// Strip .log extension s.t. config files can contain the exact filename
+	// while not breaking existing behavior for apps that don't contain the
+	// extension.
+	name = strings.TrimSuffix(name, ".log")
+	var fileLogger io.WriteCloser
+	fileLogger = &lumberjack.Logger{
+		Filename: fmt.Sprintf("%s/%s.log", logDir, name),
+		MaxSize:  logSize, // MiB
+		MaxAge:   logAge,  // days
+	}
+
+	if logFlush != 0 {
+		logBuf = newSyncBuf(fileLogger)
+		fileLogger = logBuf
+	}
+
 	logFileHandler = log15.LvlFilterHandler(logLvl,
-		log15.StreamHandler(logBuf, fmt15.Fmt15Format(nil)))
+		log15.StreamHandler(fileLogger, fmt15.Fmt15Format(nil)))
 	setHandlers()
-	go func() {
-		for range time.Tick(time.Duration(logFlush) * time.Second) {
-			Flush()
-		}
-	}()
+
+	if logFlush > 0 {
+		go func() {
+			for range time.Tick(time.Duration(logFlush) * time.Second) {
+				Flush()
+			}
+		}()
+	}
 	return nil
 }
 
@@ -124,14 +144,6 @@ func AddLogFileFlags() {
 	flag.IntVar(&logFlush, "log.flush", 5, "How frequently to flush to the log file, in seconds")
 }
 
-func mkLogfile(name string) io.WriteCloser {
-	return &lumberjack.Logger{
-		Filename: fmt.Sprintf("%s/%s.log", logDir, name),
-		MaxSize:  logSize, // MiB
-		MaxAge:   logAge,  // days
-	}
-}
-
 func LogPanicAndExit() {
 	if msg := recover(); msg != nil {
 		log15.Crit("Panic", "msg", msg, "stack", string(debug.Stack()))
@@ -141,7 +153,9 @@ func LogPanicAndExit() {
 }
 
 func Flush() {
-	logBuf.Flush()
+	if logBuf != nil {
+		logBuf.Flush()
+	}
 }
 
 func New(ctx ...interface{}) Logger {
--- a/go/lib/log/log.go
+++ b/go/lib/log/log.go
@@ -73,11 +73,10 @@ func SetupFromFlags(name string) error {
 // logLevel can be one of debug, info, warn, error, and crit and states the
 // minimum level of logging events that get written to the file. logSize is the
 // maximum size, in MiB, until the log rotates. logAge is the maximum number of
-// days to retain old log files. logFlush is the interval, in seconds, on which
-// log messages are flushed from memory to file; a logFlush of 0 means always
-// flush each message immediately, and a negative logFlush means never flush
-// automatically. In the latter case, the log can still be flushed by calling
-// Flush manually.
+// days to retain old log files. If logFlush > 0, logging output is buffered,
+// and flushed every logFlush seconds.  If logFlush < 0: logging output is
+// buffered, but must be manually flushed by calling Flush(). If logFlush = 0
+// logging output is unbuffered and Flush() is a no-op.
 func SetupLogFile(name string, logDir string, logLevel string, logSize int, logAge int,
 	logFlush int) error {
 
--- a/go/lib/log/log.go
+++ b/go/lib/log/log.go
@@ -68,6 +68,16 @@ func SetupFromFlags(name string) error {
 	return err
 }
 
+// SetupLogFile initializes a file for logging. The path is logDir/name.log if
+// name doesn't already contain the .log extension, or logDir/name otherwise.
+// logLevel can be one of debug, info, warn, error, and crit and states the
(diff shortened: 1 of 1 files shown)


sgmonroy:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579#-:-LDpUERj-huzHgte3XbT:bnfp4nl)*

sgmonroy:

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)*

kormat:
Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[go/lib/log/log.go, line 90 at r1](https://reviewable.io/reviews/scionproto/scion/1579#-LDub6-i0VeTZshcEwqV:-LDub6-i0VeTZshcEwqW:b-fiayc9) ([raw file](https://github.com/scionproto/scion/blob/8be616cd623a73882d2521cd92364927c5e73f01/go/lib/log/log.go#L90)):*
> ```Go
> 	}
> 
> 	if logFlush != 0 {
> ```

Might as well use `if logFlush > 0` here too, for consistency with the check below, and also so that negative values don't cause buffering that never gets flushed.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)*

scrye:
*[go/lib/log/log.go, line 90 at r1](https://reviewable.io/reviews/scionproto/scion/1579#-LDub6-i0VeTZshcEwqV:-LDubgmN16uwiyf3Bg46:bgn2ism) ([raw file](https://github.com/scionproto/scion/blob/8be616cd623a73882d2521cd92364927c5e73f01/go/lib/log/log.go#L90)):*

Might as well use `if logFlush > 0` here too, for consistency with the check below, and also so that negative values don't cause buffering that never gets flushed.
</blockquote></details>

The semantics behind negative values are "never flush automatically, but you are allowed to flush manually" (see PR description). Changing the condition to `>0` would mean `logBuf` never gets initialized, and `Flush()` does nothing, which is not intended behavior.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)*

kormat:

---

*[go/lib/log/log.go, line 90 at r1](https://reviewable.io/reviews/scionproto/scion/1579#-LDub6-i0VeTZshcEwqV:-LDucM0g4dmubiXol5O_:b1tpkes) ([raw file](https://github.com/scionproto/scion/blob/8be616cd623a73882d2521cd92364927c5e73f01/go/lib/log/log.go#L90)):*

The semantics behind negative values are "never flush automatically, but you are allowed to flush manually" (see PR description). Changing the condition to `>0` would mean `logBuf` never gets initialized, and `Flush()` does nothing, which is not intended behavior.
</blockquote></details>

Ahh, sorry, ok. In that case, this should be documented in the docstring for the function.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)*

kormat:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[go/lib/log/log.go, line 78 at r2](https://reviewable.io/reviews/scionproto/scion/1579#-LDuigud6A1tdaILRSCC:-LDuigud6A1tdaILRSCD:b-3nnkn4) ([raw file](https://github.com/scionproto/scion/blob/c1ec9c2e6673b90bbb9b37d0660bdc982a3c8e9c/go/lib/log/log.go#L78)):*
> a logFlush of 0 means always flush each message immediately

This isn't correct - the OS will block-buffer the output, which makes log files a lot less useful (it takes an indeterminate time before a given log line will appear on disk, and it will often flush partial lines).

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)*

scrye:
*[go/lib/log/log.go, line 78 at r2](https://reviewable.io/reviews/scionproto/scion/1579#-LDuigud6A1tdaILRSCC:-LDv1tu4CLqBiMZ2UieV:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c1ec9c2e6673b90bbb9b37d0660bdc982a3c8e9c/go/lib/log/log.go#L78)):*

> a logFlush of 0 means always flush each message immediately

This isn't correct - the OS will block-buffer the output, which makes log files a lot less useful (it takes an indeterminate time before a given log line will appear on disk, and it will often flush partial lines).


</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)*

kormat:

---

*[go/lib/log/log.go, line 78 at r2](https://reviewable.io/reviews/scionproto/scion/1579#-LDuigud6A1tdaILRSCC:-LDvfL3W4-7DkU_Aqq3T:b-z2sh8z) ([raw file](https://github.com/scionproto/scion/blob/7d6570d262997c567960bf4579b5933255e7b00d/go/lib/log/log.go#L78)):*
Looking at this again, it's a bit odd that logFlush:1 is buffered+autoflushed, logFlush:0 is unbuffered, and logFlush:-1 is buffered. How about swapping the last two. It would also be good to be more explicit about buffering. Suggested wording:

> logFlush controls how logs are written to disk:
>   \>0: logging output is buffered, and flushed every logFlush seconds.
>  0: logging output is buffered, but must be manually flushed by calling Flush()
> -1: logging output is unbuffered. Flush() is a no-op in this case.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)*

scrye:

---

*[go/lib/log/log.go, line 78 at r2](https://reviewable.io/reviews/scionproto/scion/1579#-LDuigud6A1tdaILRSCC:-LDvhguDDG9mLgadACL6:b-gc3p7f) ([raw file](https://github.com/scionproto/scion/blob/7d6570d262997c567960bf4579b5933255e7b00d/go/lib/log/log.go#L78)):*

Looking at this again, it's a bit odd that logFlush:1 is buffered+autoflushed, logFlush:0 is unbuffered, and logFlush:-1 is buffered. How about swapping the last two. It would also be good to be more explicit about buffering. Suggested wording:

> logFlush controls how logs are written to disk:
>   \>0: logging output is buffered, and flushed every logFlush seconds.
>  0: logging output is buffered, but must be manually flushed by calling Flush()
> -1: logging output is unbuffered. Flush() is a no-op in this case.

</blockquote></details>

I thought about having it like this, but I find it more odd that a flush interval of 0 has semantics of never. The most intuitive one, for me, is "if I put 0, then it means I'm constantly flushing your messages with a delay of 0".

I also find it useful that if someone leaves it as a default of 0, it defaults to a very intuitive the behavior of just flushing all the time. Expert users can set it to > 0 for performance, or < 0 for manual control, but the default behavior is the most intuitive.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)*

kormat:

---

*[go/lib/log/log.go, line 78 at r2](https://reviewable.io/reviews/scionproto/scion/1579#-LDuigud6A1tdaILRSCC:-LDvjgmGB-DxJ9yxld9C:bfezr52) ([raw file](https://github.com/scionproto/scion/blob/7d6570d262997c567960bf4579b5933255e7b00d/go/lib/log/log.go#L78)):*

I thought about having it like this, but I find it more odd that a flush interval of 0 has semantics of never. The most intuitive one, for me, is "if I put 0, then it means I'm constantly flushing your messages with a delay of 0".

I also find it useful that if someone leaves it as a default of 0, it defaults to a very intuitive the behavior of just flushing all the time. Expert users can set it to > 0 for performance, or < 0 for manual control, but the default behavior is the most intuitive.
</blockquote></details>

Yeah that's a good point. If someone uses a default `0`, they're going to be Very confused when they never see output (unless they call Flush()). Ok. How about (something like) my wording with the values as they are currently?

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)*

scrye:
*[go/lib/log/log.go, line 78 at r2](https://reviewable.io/reviews/scionproto/scion/1579#-LDuigud6A1tdaILRSCC:-LDw49HiC-VtC3XIMA5Z:bs560wh) ([raw file](https://github.com/scionproto/scion/blob/7d6570d262997c567960bf4579b5933255e7b00d/go/lib/log/log.go#L78)):*

Yeah that's a good point. If someone uses a default `0`, they're going to be Very confused when they never see output (unless they call Flush()). Ok. How about (something like) my wording with the values as they are currently?
</blockquote></details>

SGTM, updated.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579)*

kormat:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1579#-:-LE9-Lhv-PGEi3FL1hLq:bnfp4nl)*

PR #1580, opened 2018-05-31
Enable SCIOND-less operation for snet

scrye:
SCIOND-less operation is needed for applications that need to talk over the network without sciond access (e.g., the sciond server itself).

Also streamlines the API a bit and deletes some unused fields from `snet` types.

SCIOND-less operation can be selected by passing an empty sciond socket path, or by passing in a nil path resolver.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580)

### DIFF
--- a/go/lib/pathmgr/pathmgr.go
+++ b/go/lib/pathmgr/pathmgr.go
@@ -267,3 +267,7 @@ func (r *PR) revoke(b common.RawBytes) {
 		log.Warn("Found invalid revocation notification", "revInfo", revInfo)
 	}
 }
+
+func (r *PR) Sciond() sciond.Service {
+	return r.sciondService
+}
--- a/go/lib/snet/conn.go
+++ b/go/lib/snet/conn.go
@@ -76,7 +76,8 @@ type Conn struct {
 	recvBuffer common.RawBytes
 	sendBuffer common.RawBytes
 	// Pointer to slice of paths updated by continous lookups; these are
-	// used by default when creating a connection via Dial
+	// used by default when creating a connection via Dial on SCIOND-enabled
+	// networks. For SCIOND-less operation, this is set to nil.
 	sp *pathmgr.SyncPaths
 	// Reference to SCION networking context
 	scionNet *Network
@@ -220,8 +221,10 @@ func (c *Conn) handleSCMPRev(hdr *scmp.Hdr, pkt *spkt.ScnPkt) {
 			"type", common.TypeOf(scmpPayload.Info))
 	}
 	log.Info("Received SCMP revocation", "header", hdr.String(), "payload", scmpPayload.String())
-	// Extract RevInfo buffer and send it to path manager
-	c.scionNet.pathResolver.Revoke(info.RawSRev)
+	// If we have a path manager, extract RevInfo buffer and send it
+	if c.scionNet.pathResolver != nil {
+		c.scionNet.pathResolver.Revoke(info.RawSRev)
+	}
 }
 
 // WriteToSCION sends b to raddr.
@@ -270,6 +273,11 @@ func (c *Conn) write(b []byte, raddr *Addr) (int, error) {
 			nextHopHost = raddr.NextHopHost
 			nextHopPort = raddr.NextHopPort
 		} else {
+			if c.sp == nil {
+				return 0, common.NewBasicError("Trying to send traffic outside local AS, "+
+					"but no path is specified and SCIOND is disabled", nil)
+			}
+
 			pathEntry, err := c.selectPathEntry(raddr)
 			if err != nil {
 				return 0, err
--- a/go/lib/snet/conn.go
+++ b/go/lib/snet/conn.go
@@ -75,7 +75,7 @@ type Conn struct {
 	writeMutex sync.Mutex
 	recvBuffer common.RawBytes
 	sendBuffer common.RawBytes
-	// Pointer to slice of paths updated by continous lookups; these are
+	// Pointer to slice of paths updated by continuous lookups; these are
 	// used by default when creating a connection via Dial on SCIOND-enabled
 	// networks. For SCIOND-less operation, this is set to nil.
 	sp *pathmgr.SyncPaths
--- a/go/lib/snet/conn.go
+++ b/go/lib/snet/conn.go
@@ -341,7 +341,9 @@ func (c *Conn) write(b []byte, raddr *Addr) (int, error) {
 // running SCIOND-less.
 func (c *Conn) selectPathEntry(raddr *Addr) (*sciond.PathReplyEntry, error) {
 	var pathSet spathmeta.AppPathSet
-	assert.Must(c.scionNet.pathResolver != nil, "must run with SCIOND for path selection")
+	if assert.On {
+		assert.Must(c.scionNet.pathResolver != nil, "must run with SCIOND for path selection")
+	}
 	// If the remote address is fixed, register source and destination for
 	// continous path updates
 	if c.raddr == nil {
--- a/go/lib/snet/conn.go
+++ b/go/lib/snet/conn.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/scionproto/scion/go/lib/addr"
+	"github.com/scionproto/scion/go/lib/assert"
 	"github.com/scionproto/scion/go/lib/common"
 	"github.com/scionproto/scion/go/lib/hpkt"
 	"github.com/scionproto/scion/go/lib/l4"
@@ -222,6 +223,10 @@ func (c *Conn) handleSCMPRev(hdr *scmp.Hdr, pkt *spkt.ScnPkt) {
 	}
 	log.Info("Received SCMP revocation", "header", hdr.String(), "payload", scmpPayload.String())
 	// If we have a path manager, extract RevInfo buffer and send it
+	//
+	// FIXME(scrye): this completely hides the revocation from the application.
+	// This is problematic for applications that manage their own paths, as
+	// they need to be informed that they should try to use a different one.
 	if c.scionNet.pathResolver != nil {
 		c.scionNet.pathResolver.Revoke(info.RawSRev)
 	}
@@ -273,9 +278,8 @@ func (c *Conn) write(b []byte, raddr *Addr) (int, error) {
 			nextHopHost = raddr.NextHopHost
 			nextHopPort = raddr.NextHopPort
 		} else {
-			if c.sp == nil {
-				return 0, common.NewBasicError("Trying to send traffic outside local AS, "+
-					"but no path is specified and SCIOND is disabled", nil)
+			if c.scionNet.pathResolver == nil {
+				return 0, common.NewBasicError("Path required, but no path manager configured", nil)
 			}
 
 			pathEntry, err := c.selectPathEntry(raddr)
@@ -333,22 +337,16 @@ func (c *Conn) write(b []byte, raddr *Addr) (int, error) {
 	return pkt.Pld.Len(), nil
 }
 
+// selectPathEntry chooses a path to raddr. This must not be called if
+// running SCIOND-less.
 func (c *Conn) selectPathEntry(raddr *Addr) (*sciond.PathReplyEntry, error) {
-	var err error
 	var pathSet spathmeta.AppPathSet
+	assert.Must(c.scionNet.pathResolver != nil, "must run with SCIOND for path selection")
 	// If the remote address is fixed, register source and destination for
 	// continous path updates
 	if c.raddr == nil {
 		pathSet = c.scionNet.pathResolver.Query(c.laddr.IA, raddr.IA)
 	} else {
-		// Sanity check, as Dial already initializes this
-		if c.sp == nil {
-			c.sp, err = c.scionNet.pathResolver.Watch(c.laddr.IA, c.raddr.IA)
-			if err != nil {
-				return nil, common.NewBasicError("Unable to register src-dst IAs", err,
-					"src", c.laddr.IA, "dst", raddr.IA)
-			}
-		}
 		pathSet = c.sp.Load().APS
 	}
 
--- a/go/lib/snet/integration_test.go
+++ b/go/lib/snet/integration_test.go
@@ -154,11 +154,11 @@ func ClientServer(haveSciond bool, idx int, tc TestCase) {
 		cconn.SetDeadline(time.Now().Add(5 * time.Second))
 		SoMsg("Client deadline error", err, ShouldBeNil)
 
+		n, err := cconn.Write([]byte("Hello!"))
+		xtest.SoMsgError("Client write error", err, tc.expectWriteError)
 		// Only run the message exchange in cases where snet doesn't error out
 		// due to needing a path in SCIOND-less mode of operation.
 		if tc.expectWriteError == false {
-			n, err := cconn.Write([]byte("Hello!"))
-			xtest.SoMsgError("Client write error", err, false)
 			SoMsg("Client written bytes", n, ShouldEqual, len("Hello!"))
 
 			n, raddr, err := sconn.ReadFromSCION(b)
--- a/go/lib/snet/integration_test.go
+++ b/go/lib/snet/integration_test.go
@@ -18,7 +18,6 @@ package snet
 
 import (
 	"fmt"
-	"io/ioutil"
 	"math/rand"
 	"net"
 	"os"
@@ -31,6 +30,7 @@ import (
 	"github.com/scionproto/scion/go/lib/log"
 	"github.com/scionproto/scion/go/lib/sciond"
 	"github.com/scionproto/scion/go/lib/util"
+	"github.com/scionproto/scion/go/lib/xtest"
 )
 
 var _ = log.Root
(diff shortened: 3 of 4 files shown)


shitz:

---

*[go/lib/snet/conn.go, line 78 at r1](https://reviewable.io/reviews/scionproto/scion/1580#-LDvPbtnEvM0MWKDOX96:-LDvPbtnEvM0MWKDOX97:b-d913zj) ([raw file](https://github.com/scionproto/scion/blob/a393c7c22d26e72db192f7bd984dab1b69e55d5b/go/lib/snet/conn.go#L78)):*
> ```
> continous
> ```
drive-by grammar check: continuous

---

*[go/lib/snet/integration_test.go, line 127 at r1](https://reviewable.io/reviews/scionproto/scion/1580#-LDvRwCt31m_gv4DNQD8:-LDvRwCt31m_gv4DNQD9:bprj45m) ([raw file](https://github.com/scionproto/scion/blob/a393c7c22d26e72db192f7bd984dab1b69e55d5b/go/lib/snet/integration_test.go#L127)):*
> ```Go
> 		clientSciond := ""
> 		if haveSciond {
> 			clientSciond = fmt.Sprintf(SCIONDPath, tc.srcIA.FileFmt(false))
> ```

You can get rid of this and use `sciond.GetDefaultSCIONDPath(&tc.srcIA)` now (applies also in different places).

---

*[go/lib/snet/integration_test.go, line 161 at r1](https://reviewable.io/reviews/scionproto/scion/1580#-LDvSCgb06TIrXVdJdJW:-LDvSCgc3a_e-_Y6wrsr:b-pq1irl) ([raw file](https://github.com/scionproto/scion/blob/a393c7c22d26e72db192f7bd984dab1b69e55d5b/go/lib/snet/integration_test.go#L161)):*
> ```
> tc.expectWriteError
> ```
Here you could change this to `false` since you are in a `if tc.expectWriteError == false` block.

---

*[go/lib/snet/snet.go, line 44 at r1](https://reviewable.io/reviews/scionproto/scion/1580#-LDvU5B59xqPnlpY8zcw:-LDvU5B59xqPnlpY8zcx:bvrikhu) ([raw file](https://github.com/scionproto/scion/blob/a393c7c22d26e72db192f7bd984dab1b69e55d5b/go/lib/snet/snet.go#L44)):*
> ```Go
> //
> // Important: not draining SCMP errors via Read calls can cause the dispatcher
> // to shutdown the socket (see PR #1356). To prevent this on a Conn object with
> ```

include the entire URL to the PR

---

*[go/lib/snet/snet.go, line 119 at r1](https://reviewable.io/reviews/scionproto/scion/1580#-LDvUXjH2fcGqNIIecH2:-LDvUXjH2fcGqNIIecH3:bfkrlqe) ([raw file](https://github.com/scionproto/scion/blob/a393c7c22d26e72db192f7bd984dab1b69e55d5b/go/lib/snet/snet.go#L119)):*
> ```Go
> // mode of operation, the app is fully responsible with supplying paths for
> // sent traffic.
> func NewNetwork(ia addr.IA, sPath string, dPath string) (*Network, error) {
> ```

`sPath` and `dPath` are not very verbose. Change them to `sciondPath` and `dispatcherPath` (also in `NewNetworkWithPR`)

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580)*

scrye:
Review status: 1 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[go/lib/snet/conn.go, line 78 at r1](https://reviewable.io/reviews/scionproto/scion/1580#-LDvPbtnEvM0MWKDOX96:-LDvg-gz6-eaAJhedN-d:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/a393c7c22d26e72db192f7bd984dab1b69e55d5b/go/lib/snet/conn.go#L78)):*

> ```
> continous
> ```
drive-by grammar check: continuous
</blockquote></details>

Done.

---

*[go/lib/snet/integration_test.go, line 127 at r1](https://reviewable.io/reviews/scionproto/scion/1580#-LDvRwCt31m_gv4DNQD8:-LDvhSLm88bILVBiU6KU:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/a393c7c22d26e72db192f7bd984dab1b69e55d5b/go/lib/snet/integration_test.go#L127)):*

You can get rid of this and use `sciond.GetDefaultSCIONDPath(&tc.srcIA)` now (applies also in different places).
</blockquote></details>

Done.

---

*[go/lib/snet/integration_test.go, line 161 at r1](https://reviewable.io/reviews/scionproto/scion/1580#-LDvSCgb06TIrXVdJdJW:-LDvith_91VaJGHNXdzb:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/a393c7c22d26e72db192f7bd984dab1b69e55d5b/go/lib/snet/integration_test.go#L161)):*

> ```
> tc.expectWriteError
> ```
Here you could change this to `false` since you are in a `if tc.expectWriteError == false` block.
</blockquote></details>

Done.

---

*[go/lib/snet/snet.go, line 44 at r1](https://reviewable.io/reviews/scionproto/scion/1580#-LDvU5B59xqPnlpY8zcw:-LDw0T_CA4I2v7YJr06l:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/a393c7c22d26e72db192f7bd984dab1b69e55d5b/go/lib/snet/snet.go#L44)):*

include the entire URL to the PR
</blockquote></details>

Done.

---

*[go/lib/snet/snet.go, line 119 at r1](https://reviewable.io/reviews/scionproto/scion/1580#-LDvUXjH2fcGqNIIecH2:-LDw18Io7Sgx3v81ZS81:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/a393c7c22d26e72db192f7bd984dab1b69e55d5b/go/lib/snet/snet.go#L119)):*

`sPath` and `dPath` are not very verbose. Change them to `sciondPath` and `dispatcherPath` (also in `NewNetworkWithPR`)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580)*

shitz:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580#-:-LE904ZsEwZSFiRCvrIs:bnfp4nl)*

kormat:
Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[go/lib/snet/conn.go, line 224 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9N7JjETtbFYrkZqMs:-LE9N7JjETtbFYrkZqMt:bc1ga7i) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/conn.go#L224)):*
> ```Go
> 	}
> 	log.Info("Received SCMP revocation", "header", hdr.String(), "payload", scmpPayload.String())
> 	// If we have a path manager, extract RevInfo buffer and send it
> ```

If there is no path manager, revocations are effectively black-holed, is that correct? That is, the app never knows about the revocation, and sciond doesn't get informed.

---

*[go/lib/snet/conn.go, line 276 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9Oxg23B4nont1uLBr:-LE9Oxg23B4nont1uLBs:b-jlo7zp) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/conn.go#L276)):*
> ```Go
> 			nextHopPort = raddr.NextHopPort
> 		} else {
> 			if c.sp == nil {
> ```

`selectPathEntry` initializes `c.sp`, with a comment `// Sanity check, as Dial already initializes this`. Either this should change to `c.sciondNet.pathResolver == nil`, or the initialisation in `selectPathEntry` should be dropped. Probably the best would be to have an assert in selectPathEntry to ensure that it was already initialised, that way it's obvious what the expected state is.

---

*[go/lib/snet/conn.go, line 277 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9NQLG6gyc4cP5nOMq:-LE9NQLG6gyc4cP5nOMr:b-1yz2o1) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/conn.go#L277)):*
> ```Go
> 		} else {
> 			if c.sp == nil {
> 				return 0, common.NewBasicError("Trying to send traffic outside local AS, "+
> ```

Maybe simpler: "Path required (no path manager configured)".

---

*[go/lib/snet/integration_test.go, line 159 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9R5LUDq1_la-4S44g:-LE9R5LV4HZU19jppJKw:b-p72wxt) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/integration_test.go#L159)):*
> ```Go
> 		// Only run the message exchange in cases where snet doesn't error out
> 		// due to needing a path in SCIOND-less mode of operation.
> 		if tc.expectWriteError == false {
> ```

This means the expectWriteError test cases are just ignored? That seems to defeat the point.

---

*[go/lib/snet/snet.go, line 103 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9LA107sN2vxaGh78m:-LE9LA107sN2vxaGh78n:bh8887a) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/snet.go#L103)):*
> ```Go
> 
> // NewNetworkWithPR creates a new networking context with path resolver pr. A
> // nil path resolver means the Network will run without SCIOND.
> ```

I'm a little concerned that this is conflating "path resolution" with SCIOND. SCIOND serves a number of roles, not all of which are about path resolution. Is there an implicit assertion that snet only cares about SCIOND for path resolution/revocation verification?

---

*[go/lib/snet/snet.go, line 261 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9O3iHAjQSzurB6ZOg:-LE9O3iHAjQSzurB6ZOh:b-schsmn) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/snet.go#L261)):*
> ```Go
> // Sciond returns the sciond.Service that the network is using.
> func (n *Network) Sciond() sciond.Service {
> 	return n.pathResolver.Sciond()
> ```

This will panic if `pathResolver` is nil.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580)*

scrye:

---

*[go/lib/snet/conn.go, line 224 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9N7JjETtbFYrkZqMs:-LEJsMvN5WKU0zh17Bs6:b41gp8e) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/conn.go#L224)):*

If there is no path manager, revocations are effectively black-holed, is that correct? That is, the app never knows about the revocation, and sciond doesn't get informed.
</blockquote></details>

Yes, they are completely black-holed.

I have been thinking a while about adding an explicit SCMP mask to `snet` (so clients can say "I want to take care of these and these SCMP messages myself, so don't manage them and give them to me, but do what you have to do with the rest"). But I'd rather do it at a later point in a separate PR. I've added a FIXME s.t. we don't forget.

---

*[go/lib/snet/conn.go, line 276 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9Oxg23B4nont1uLBr:-LEJycRJ0ohf6HqKNdrA:b-3op1gg) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/conn.go#L276)):*

`selectPathEntry` initializes `c.sp`, with a comment `// Sanity check, as Dial already initializes this`. Either this should change to `c.sciondNet.pathResolver == nil`, or the initialisation in `selectPathEntry` should be dropped. Probably the best would be to have an assert in selectPathEntry to ensure that it was already initialised, that way it's obvious what the expected state is.
</blockquote></details>

Yes, now that "hot-swapping" a path resolver post-initialization is no longer possible, this can be simplified and we can treat the sanity check as an assertion. Done.

---

*[go/lib/snet/conn.go, line 277 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9NQLG6gyc4cP5nOMq:-LEJyzHrAQfqOhvz_KV0:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/conn.go#L277)):*

Maybe simpler: "Path required (no path manager configured)".
</blockquote></details>

Done.

---

*[go/lib/snet/integration_test.go, line 159 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9R5LUDq1_la-4S44g:-LEJz3oAE15Bz28g00yL:b371btz) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/integration_test.go#L159)):*

This means the expectWriteError test cases are just ignored? That seems to defeat the point.
</blockquote></details>

It still tests that `Listen` and `Dial` work correctly, and that the `Write` returns an error.

The `Write` is not placed correctly though, it should have been outside the `if`. Done.

---

*[go/lib/snet/snet.go, line 103 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9LA107sN2vxaGh78m:-LEJzcD24oyuNzjpJF2s:bon1au4) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/snet.go#L103)):*

I'm a little concerned that this is conflating "path resolution" with SCIOND. SCIOND serves a number of roles, not all of which are about path resolution. Is there an implicit assertion that snet only cares about SCIOND for path resolution/revocation verification?
</blockquote></details>

Ah, it's not about having access to just the path functionality bit of a path resolver. It's that a path resolver always comes with a SCIOND, so `snet` should use that one (if it ever needs it), so it doesn't need an additional reference of its own.

---

*[go/lib/snet/snet.go, line 261 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9O3iHAjQSzurB6ZOg:-LEK081k-o_90-wnxhTC:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/snet.go#L261)):*

This will panic if `pathResolver` is nil.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580)*

kormat:
Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/snet/conn.go, line 344 at r3](https://reviewable.io/reviews/scionproto/scion/1580#-LEK3MSF-7M227kVwAwi:-LEK3MSF-7M227kVwAwj:b-oi78ux) ([raw file](https://github.com/scionproto/scion/blob/5d2fcc964c773b4df830ad3ac451a0f85870d708/go/lib/snet/conn.go#L344)):*
> ```Go
> func (c *Conn) selectPathEntry(raddr *Addr) (*sciond.PathReplyEntry, error) {
> 	var pathSet spathmeta.AppPathSet
> 	assert.Must(c.scionNet.pathResolver != nil, "must run with SCIOND for path selection")
> ```

Asserts should be wrapped with `if assert.On{ ... }`

---

*[go/lib/snet/integration_test.go, line 159 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9R5LUDq1_la-4S44g:-LEK3Trp6zq_aPc5C3KX:b-8h8wjc) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/integration_test.go#L159)):*

It still tests that `Listen` and `Dial` work correctly, and that the `Write` returns an error.

The `Write` is not placed correctly though, it should have been outside the `if`. Done.
</blockquote></details>

Ahh ok, it was the placement of Write that confused me.

---

*[go/lib/snet/snet.go, line 103 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9LA107sN2vxaGh78m:-LEK4ItAED6hBXNfwa_g:bhksy9w) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/snet.go#L103)):*

Ah, it's not about having access to just the path functionality bit of a path resolver. It's that a path resolver always comes with a SCIOND, so `snet` should use that one (if it ever needs it), so it doesn't need an additional reference of its own.
</blockquote></details>

I'm not sure that covers my concern. If snet should have sciond access regardless of path manager or not, then it _should_ have its own sciond object.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580)*

scrye:
Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[go/lib/snet/conn.go, line 344 at r3](https://reviewable.io/reviews/scionproto/scion/1580#-LEK3MSF-7M227kVwAwi:-LEK9ObxE_vlIhzXFsJT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5d2fcc964c773b4df830ad3ac451a0f85870d708/go/lib/snet/conn.go#L344)):*

Asserts should be wrapped with `if assert.On{ ... }`
</blockquote></details>

Done.

---

*[go/lib/snet/snet.go, line 103 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9LA107sN2vxaGh78m:-LEK4l7d2qDKTUG127Zb:bo6mx5l) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/snet.go#L103)):*
I'm not following.

>  If snet should have sciond access regardless of path manager or not

snet never has sciond access if it doesn't have a path manager.

I'm suggesting the two be/are the same. In other words, "snet has a path manager" is equivalent to "snet has sciond access".

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580)*

kormat:

---

*[go/lib/snet/snet.go, line 103 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9LA107sN2vxaGh78m:-LEKA8rY2VV7fzWygr0F:b-k8id4x) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/snet.go#L103)):*

I'm not following.

>  If snet should have sciond access regardless of path manager or not

snet never has sciond access if it doesn't have a path manager.

I'm suggesting the two be/are the same. In other words, "snet has a path manager" is equivalent to "snet has sciond access".
</blockquote></details>

Ok, i think that answers my original question: you are asserting that snet doesn't care about sciond unless it has a path manager. In that case, i'm curious to see how DRKey integration will work. Doing it through snet would be the natural approach (for at least some uses of DRKey), which requires sciond, but is orthogonal to having a path manager.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580)*

scrye:
Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[go/lib/snet/snet.go, line 103 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9LA107sN2vxaGh78m:-LEKAuGg65DAcOAaOKyp:bjt735i) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/snet.go#L103)):*

Ok, i think that answers my original question: you are asserting that snet doesn't care about sciond unless it has a path manager. In that case, i'm curious to see how DRKey integration will work. Doing it through snet would be the natural approach (for at least some uses of DRKey), which requires sciond, but is orthogonal to having a path manager.
</blockquote></details>

I guess we'll see when we get there. It depends a lot on how we model the API. We'll need to look at multiple questions. Will snet bootstrap the DRKey ecosystem? Will it want the app to perform initial set-up, and then receive a reference?

Starting snet with a path to sciond already assumed (and still assumes) that a resolver will be bootstrapped internally. This PR did not change that behavior, it only made it more consistent across the different constructors.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580)*

kormat:
Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[go/lib/snet/snet.go, line 103 at r2](https://reviewable.io/reviews/scionproto/scion/1580#-LE9LA107sN2vxaGh78m:-LEKCFVq5HJ4imHP21R7:bhz2rkm) ([raw file](https://github.com/scionproto/scion/blob/242d1c2b608ac811de080c1c497e784f008243f5/go/lib/snet/snet.go#L103)):*

I guess we'll see when we get there. It depends a lot on how we model the API. We'll need to look at multiple questions. Will snet bootstrap the DRKey ecosystem? Will it want the app to perform initial set-up, and then receive a reference?

Starting snet with a path to sciond already assumed (and still assumes) that a resolver will be bootstrapped internally. This PR did not change that behavior, it only made it more consistent across the different constructors.
</blockquote></details>

Ok, let's leave it for now. I'm pretty sure it will need to change, soon, but we'll do that when we get there.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580)*

kormat:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1580#-:-LEKCMqp3XYzdnrbkx6D:bnfp4nl)*

PR #1602, opened 2018-06-10
snet: clean split between interface/basic types and implementation

### LINKED ISSUE #1504: snet: Clean split between interface/basic types and implementation
Currently, all the basic types, the API, and the implementation of `snet` is in one package. This can lead to hard to break import cycles, since the implementation has many dependencies, but often packages only need base types without the entire functionality. Proposed it to split up the current `snet` in a package that contains base types and the conn interface (this would be called `snet`), a package that provides the implementation of the interface, and `snet/snetutils` that provides helper methods for the most common instantiations of `snet`.

Along the same lines, `pathmgr` should be split into an interface and an implementation part.

jinankjain:
WIP: I would like to get some feedback about snet interface design
and how it's split into pkg internal and snetutils. @shitz

Fixes #1504 

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1602)


jinankjain:
@scrye I will fix the conflicts later once we finalize the design for snet and snetutils.

scrye:
Most new packages no longer use implementations directly and instead use interfaces, relying on main to initialize implementations. This eliminates all import cycles in a cleaner way. Closing this since it no longer applies.

(Also #1833, #2460, and #2462 introduced new interfaces that other packages can use instead of the concrete implementations)

PR #1607, opened 2018-06-13
BR: support for multiple external interfaces

sgmonroy:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1607)

### DIFF
--- a/go/border/router.go
+++ b/go/border/router.go
@@ -174,8 +174,7 @@ func (r *Router) processPacket(rp *rpkt.RtrPkt) {
 		r.handlePktError(rp, err, "Error processing packet")
 		return
 	}
-	// If the packet's destination is this router, there's no need to forward
-	// it.
+	// If the packet's destination is this router, there's no need to forward it.
 	if rp.DirTo != rcmn.DirSelf {
 		if err := rp.Route(); err != nil {
 			r.handlePktError(rp, err, "Error routing packet")
--- a/go/border/rpkt/create.go
+++ b/go/border/rpkt/create.go
@@ -22,6 +22,7 @@ import (
 	"github.com/scionproto/scion/go/border/rcmn"
 	"github.com/scionproto/scion/go/border/rctx"
 	"github.com/scionproto/scion/go/lib/addr"
+	"github.com/scionproto/scion/go/lib/assert"
 	"github.com/scionproto/scion/go/lib/common"
 	"github.com/scionproto/scion/go/lib/l4"
 	"github.com/scionproto/scion/go/lib/log"
@@ -172,7 +173,7 @@ func (rp *RtrPkt) CreateReplyScnPkt() (*spkt.ScnPkt, error) {
 		return nil, err
 	}
 	sp.SrcIA = rp.Ctx.Conf.IA
-	// Use the ingress address as the source host
+	// Use the local address as the source host
 	pubInt := rp.Ctx.Conf.Net.LocAddr.PublicAddrInfo(rp.Ctx.Conf.Topo.Overlay)
 	sp.SrcHost = addr.HostFromIP(pubInt.IP)
 	return sp, nil
@@ -231,7 +232,7 @@ func (rp *RtrPkt) CreateReply(sp *spkt.ScnPkt) (*RtrPkt, error) {
 func (rp *RtrPkt) replyEgress(dir rcmn.Dir, dst *topology.AddrInfo, ifid common.IFIDType) error {
 	// Destination is the local AS
 	if rp.dstIA.Eq(rp.Ctx.Conf.IA) {
-		// Write to internal interface
+		// Write to local socket
 		rp.Egress = append(rp.Egress, EgressPair{S: rp.Ctx.LocSockOut, Dst: dst})
 		return nil
 	}
@@ -239,6 +240,9 @@ func (rp *RtrPkt) replyEgress(dir rcmn.Dir, dst *topology.AddrInfo, ifid common.
 	if dir == rcmn.DirExternal {
 		rp.Egress = append(rp.Egress, EgressPair{S: rp.Ctx.ExtSockOut[ifid]})
 		return nil
+	}
+	if assert.On {
+		assert.Must(dir == rcmn.DirLocal, "Wrong direction value")
 	}
 	// DirFrom Local and destination is remote AS
 	// At this point we should have a valid path to route the packet
@@ -249,14 +253,14 @@ func (rp *RtrPkt) replyEgress(dir rcmn.Dir, dst *topology.AddrInfo, ifid common.
 		return err
 	}
 	if _, ok := rp.Ctx.Conf.Net.IFs[*rp.ifNext]; ok {
-		// Egress interface is local
+		// Egress interface is on this BR
 		// Re-inject to process the reply as an "Egress br"
-		// and make it look like it arrived in the internal interface
+		// and make it look like it arrived in the local socket
 		rp.hooks.Route = append(rp.hooks.Route, rp.reprocess)
 		return nil
 	}
-	// Egress interface is not local
-	// Write to internal interface
+	// Egress interface is not on this BR
+	// Write to local socket
 	rp.Egress = append(rp.Egress, EgressPair{S: rp.Ctx.LocSockOut, Dst: dst})
 	return nil
 }
--- a/go/border/rpkt/create.go
+++ b/go/border/rpkt/create.go
@@ -26,6 +26,7 @@ import (
 	"github.com/scionproto/scion/go/lib/l4"
 	"github.com/scionproto/scion/go/lib/log"
 	"github.com/scionproto/scion/go/lib/spkt"
+	"github.com/scionproto/scion/go/lib/topology"
 )
 
 // RtrPktFromScnPkt creates an RtrPkt from an spkt.ScnPkt.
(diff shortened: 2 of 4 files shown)


kormat:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed, 4 unresolved discussions (waiting on @sgmonroy)

---

*[go/border/rpkt/create.go, line 212 at r1](https://reviewable.io/reviews/scionproto/scion/1607#-LFwNpwq6VXPcJUuCDVd:-LFwNpwq6VXPcJUuCDVe:b-wdk5hd) ([raw file](https://github.com/scionproto/scion/blob/3a2ea7ebca3bdd7a308222e4e72853e7e51e30b0/go/border/rpkt/create.go#L212)):*
> ```Go
> 				}
> 			}
> 		} else if rp.Ingress.IfID != 0 {
> ```

Why the change from `rp.DirFrom` to checking the `Ingress.IfID`, here and below?

---

*[go/border/rpkt/route.go, line 155 at r1](https://reviewable.io/reviews/scionproto/scion/1607#-LFwO6sP38-e_7O4ZdjQ:-LFwO6sQ3ljc2xEAUQad:bftyao1) ([raw file](https://github.com/scionproto/scion/blob/3a2ea7ebca3bdd7a308222e4e72853e7e51e30b0/go/border/rpkt/route.go#L155)):*
> ```Go
> 			rp.CmnHdr.HdrLenBytes()
> 		if !onLastSeg {
> 			return HookError, common.NewBasicError("Destination ISD-AS not in the last segment", nil)
> ```

Line length.

---

*[go/border/rpkt/route.go, line 159 at r1](https://reviewable.io/reviews/scionproto/scion/1607#-LFwO5fp8xPsbfR2pReK:-LFwO5fp8xPsbfR2pReL:bftyao1) ([raw file](https://github.com/scionproto/scion/blob/3a2ea7ebca3bdd7a308222e4e72853e7e51e30b0/go/border/rpkt/route.go#L159)):*
> ```Go
> 		// Destination is a host in the local ISD-AS.
> 		if assert.On {
> 			assert.Mustf(!rp.hopF.ForwardOnly, rp.ErrStr, "Delivery forbidden for Forward-only HopF")
> ```

Line length.

---

*[go/border/rpkt/route.go, line 184 at r1](https://reviewable.io/reviews/scionproto/scion/1607#-LFwJgOjCpov2ZWdLgkC:-LFwJgOjCpov2ZWdLgkD:bz1bhgk) ([raw file](https://github.com/scionproto/scion/blob/3a2ea7ebca3bdd7a308222e4e72853e7e51e30b0/go/border/rpkt/route.go#L184)):*
> ```Go
> 		// Current Egress interface is local to BR
> 		// Re-Process the packet, there should only be HBH hooks
> 		if err := rp.Process(); err != nil {
> ```

I have a number of concerns, here. 

I think this will mean that hbh extensions will need to special-case the ingress+egress BR being the same process, which isn't pretty.

The BR supports a large number of hooks. This assumes that only the Process() hook needs to be re-run, but not, e.g. the Route hook. That's not something an hbh extension can really work around. This really cripples the whole hook system.

This is also the only time that `rp.ifCurr` is _changed_ in the BR.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1607)*

kormat:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed, 6 unresolved discussions (waiting on @kormat and @sgmonroy)

---

*[go/border/rpkt/create.go, line 175 at r2](https://reviewable.io/reviews/scionproto/scion/1607#-LG61Lsv9uXHmYq7nXf7:-LG61Lsv9uXHmYq7nXf8:b-cmp9cp) ([raw file](https://github.com/scionproto/scion/blob/c58cfb2e004285c13dd9453117ef6534b95c176d/go/border/rpkt/create.go#L175)):*
> ```Go
> 	}
> 	sp.SrcIA = rp.Ctx.Conf.IA
> 	// Use the ingress address as the source host
> ```

This comment isn't accurate, as it always uses the local address now.

---

*[go/border/rpkt/create.go, line 234 at r2](https://reviewable.io/reviews/scionproto/scion/1607#-LG63tYiBWkBgcxCPgYu:-LG63tYiBWkBgcxCPgYv:ba5x5t3) ([raw file](https://github.com/scionproto/scion/blob/c58cfb2e004285c13dd9453117ef6534b95c176d/go/border/rpkt/create.go#L234)):*
> ```Go
> 	// Destination is the local AS
> 	if rp.dstIA.Eq(rp.Ctx.Conf.IA) {
> 		// Write to internal interface
> ```

I'm a bit uncomfortable with the use of "interface" for this, as "interface" has a specific meaning in SCION, which doesn't apply here. That's why most/all of the code refers to it as the local socket, where local is defined as "facing the local AS". It's not _great_ naming, but it does at least tries to keep the two concepts separate.

---

*[go/border/rpkt/create.go, line 243 at r2](https://reviewable.io/reviews/scionproto/scion/1607#-LG63O7H60-TaDobAvoW:-LG63O7H60-TaDobAvoX:bpb0nws) ([raw file](https://github.com/scionproto/scion/blob/c58cfb2e004285c13dd9453117ef6534b95c176d/go/border/rpkt/create.go#L243)):*
> ```Go
> 		return nil
> 	}
> 	// DirFrom Local and destination is remote AS
> ```

Add an assertion for `dir == rcmn.DirLocal`, just in case it's either Unset or Self.

---

*[go/border/rpkt/create.go, line 252 at r2](https://reviewable.io/reviews/scionproto/scion/1607#-LG63gCAAQ5XBZuPsuu2:-LG63gCAAQ5XBZuPsuu3:b-ey8l1w) ([raw file](https://github.com/scionproto/scion/blob/c58cfb2e004285c13dd9453117ef6534b95c176d/go/border/rpkt/create.go#L252)):*
> ```Go
> 	}
> 	if _, ok := rp.Ctx.Conf.Net.IFs[*rp.ifNext]; ok {
> 		// Egress interface is local
> ```

"local" is very vague (it could be interpreted to mean "local AS"). "Egress interface is on this BR" would be clearer.

---

*[go/border/rpkt/route.go, line 159 at r1](https://reviewable.io/reviews/scionproto/scion/1607#-LFwO5fp8xPsbfR2pReK:-LG62EbB0jSRNYk7EFEz:b-n8h7mf) ([raw file](https://github.com/scionproto/scion/blob/3a2ea7ebca3bdd7a308222e4e72853e7e51e30b0/go/border/rpkt/route.go#L159)):*

Done.
</blockquote></details>

Still too long.

---

*[go/border/rpkt/route.go, line 283 at r2](https://reviewable.io/reviews/scionproto/scion/1607#-LG64nkV7hRAAHsMgc17:-LG64nkV7hRAAHsMgc18:b-n559tq) ([raw file](https://github.com/scionproto/scion/blob/c58cfb2e004285c13dd9453117ef6534b95c176d/go/border/rpkt/route.go#L283)):*
> ```Go
> 	ctx := rp.Ctx
> 	free := rp.Free
> 	timeIn := rp.TimeIn
> ```

I'm not sure about this one. Maybe add an `XXX` comment noting that this might need to be reconsidered  due to the effect it will have on metrics.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1607)*

sgmonroy:
Review status: all files reviewed, 6 unresolved discussions (waiting on @kormat)

---

*[go/border/rpkt/create.go, line 175 at r2](https://reviewable.io/reviews/scionproto/scion/1607#-LG61Lsv9uXHmYq7nXf7:-LGOt6cK98Av_nS2U4Gr:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c58cfb2e004285c13dd9453117ef6534b95c176d/go/border/rpkt/create.go#L175)):*

This comment isn't accurate, as it always uses the local address now.
</blockquote></details>

Done.

---

*[go/border/rpkt/create.go, line 234 at r2](https://reviewable.io/reviews/scionproto/scion/1607#-LG63tYiBWkBgcxCPgYu:-LGOtugA5wgHsT-TnZCv:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c58cfb2e004285c13dd9453117ef6534b95c176d/go/border/rpkt/create.go#L234)):*

I'm a bit uncomfortable with the use of "interface" for this, as "interface" has a specific meaning in SCION, which doesn't apply here. That's why most/all of the code refers to it as the local socket, where local is defined as "facing the local AS". It's not _great_ naming, but it does at least tries to keep the two concepts separate.
</blockquote></details>

Done.

---

*[go/border/rpkt/create.go, line 243 at r2](https://reviewable.io/reviews/scionproto/scion/1607#-LG63O7H60-TaDobAvoW:-LGOv57S2kaZ4rEJDYv1:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c58cfb2e004285c13dd9453117ef6534b95c176d/go/border/rpkt/create.go#L243)):*

Add an assertion for `dir == rcmn.DirLocal`, just in case it's either Unset or Self.
</blockquote></details>

Done.

---

*[go/border/rpkt/create.go, line 252 at r2](https://reviewable.io/reviews/scionproto/scion/1607#-LG63gCAAQ5XBZuPsuu2:-LGOuKSG4KX2LRzTj8hy:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c58cfb2e004285c13dd9453117ef6534b95c176d/go/border/rpkt/create.go#L252)):*

"local" is very vague (it could be interpreted to mean "local AS"). "Egress interface is on this BR" would be clearer.
</blockquote></details>

Done.

---

*[go/border/rpkt/route.go, line 283 at r2](https://reviewable.io/reviews/scionproto/scion/1607#-LG64nkV7hRAAHsMgc17:-LGOw1Vm-uMiv2RFE3Ma:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c58cfb2e004285c13dd9453117ef6534b95c176d/go/border/rpkt/route.go#L283)):*

I'm not sure about this one. Maybe add an `XXX` comment noting that this might need to be reconsidered  due to the effect it will have on metrics.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1607)*

PR #1618, opened 2018-06-20
Make protos consistent between go and python

### LINKED ISSUE #1524: SignedBlob Signature TS
In the signed blob proto object the timestamp was added at a later stage. Thus at the moment the timestamp is not taken into account for the signature. This should be changed. This will probably also affect the BR.

`b = [str(self.p.type).encode("utf-8"), self.p.src, self.p.timestamp.to_bytes(4, 'big')]`
 Whereas the ts should be added after the signature is added to `b`.

### LINKED ISSUE #1617: Make proto signatures consistent between Go and python
There are currently multiple issues with proto signing in the codebase:
- https://github.com/scionproto/scion/issues/1524 - the Sign timestamp is not included in the signature in python.
- `go/proto/sign.go` does not include any of the non-Signature fields in the signature.
- The ASEntry signature method is different between Go and python.

My proposal is to change the interface of `go/proto/sign.go` to automatically append the non-Signature fields to the message data for signining and verification, and update the go code that uses it. The python code currently _prepends_ the Sign data, so that'll be changed as well, and also include the timestamp.

worxli:
Make proto signature equal in go and python

- Add timestamp to sign in python
- Add non-signature fields into signature in go
- Make signature fields order consistent
- Fix ASEntry sign method in python

Fixes #1524. Fixes #1617.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1618)

### DIFF
--- a/go/lib/ctrl/seg/seg.go
+++ b/go/lib/ctrl/seg/seg.go
@@ -191,8 +191,7 @@ func (ps *PathSegment) VerifyASEntry(key common.RawBytes, idx int) error {
 }
 
 func (ps *PathSegment) sigPack(idx int) (common.RawBytes, error) {
-	err := ps.validateIdx(idx)
-	if err != nil {
+	if err := ps.validateIdx(idx); err != nil {
 		return nil, err
 	}
 	data := append(common.RawBytes(nil), ps.RawSData...)
--- a/go/proto/sign.go
+++ b/go/proto/sign.go
@@ -56,7 +56,7 @@ func (s *SignS) Sign(key, message common.RawBytes) (common.RawBytes, error) {
 	case SignType_none:
 		return nil, nil
 	case SignType_ed25519:
-		return crypto.Sign(s.sigInput(message), key, crypto.Ed25519)
+		return crypto.Sign(s.sigPack(message, false), key, crypto.Ed25519)
 	}
 	return nil, common.NewBasicError("SignS.Sign: Unsupported SignType", nil, "type", s.Type)
 }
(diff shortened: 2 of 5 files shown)


worxli:
Review status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @kormat)

---

*[go/proto/sign.go, line 100 at r1](https://reviewable.io/reviews/scionproto/scion/1618#-LFvqvAY3smvCAs-wkUe:-LFvvc1H2mk2YOmzm6nM:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3d36bf32d787a5a894d43dbe07958ef5fd354553/go/proto/sign.go#L100)):*

Rather than duplicating most of `Pack()`, instead do something like:
```
func (s *SignS) Pack() common.RawBytes {
  return s.sigPack(nil, true)
}

func (s *SignS) sigPack(msg common.RawBytes, inclSig bool) common.RawBytes {
	msg = append(msg, common.RawBytes(s.Type.String())...)
	msg = append(msg, s.Src...)
        if inclSig {
             msg = append(msg, s.Signature...)
        }
	t := make(common.RawBytes, 8)
	common.Order.PutUint64(t, s.Timestamp)
	return append(msg, t...)
}
```

You can then call `s.sigPack(message, false)` from `Verify()` and `Sign()`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1618)*

kormat:
:lgtm:

---

Reviewed 4 of 4 files at r2.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1618#-:-LFw-R-dEjEynNJ8y2Zv:bnfp4nl)*

PR #1632, opened 2018-06-27
Implement SCIOND API

scrye:
Also:
  - `segverifier` now exposes non-channel functions for revocation and segment verification (the former being needed by the `RevNotification` handler)
  - SCIOND API Handler initialization was moved to the main function, as it has access to all the needed data structures
  - `fetcher_test` functionality was migrated to cover the entire API in `sciond_test`
  - the revocation cache is now its own type, to simplify the sharing of the same cache between modules (specifically, the `fetcher` and the `RevNotification` handler)

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)


kormat:
Reviewed 10 of 11 files at r1.
Review status: 10 of 11 files reviewed, 24 unresolved discussions (waiting on @scrye)

---

*[go/sciond/main.go, line 109 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5EhkZ9YTuvCHcXN2s:-LG5EhkZ9YTuvCHcXN2t:bnr3bo3) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L109)):*
> ```Go
> 		return 1
> 	}
> 	trustStore, err := trust.NewStore(trustDB, config.General.Topology.ISD_AS, 1337, log.Root())
> ```

Using a hard-coded start ID rather defeats the purpose ;)

---

*[go/sciond/main.go, line 119 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5FA5pAbLSBBBOgoqr:-LG5FA5pAbLSBBBOgoqs:beoh9qq) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L119)):*
> ```Go
> 		return 1
> 	}
> 	conn, err := snet.ListenSCION("udp4", &config.SD.Public)
> ```

This ignores the Bind address.

---

*[go/sciond/main.go, line 125 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5FqdgB6ySN5neBsUu:-LG5FqdgB6ySN5neBsUv:b2x9bfa) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L125)):*
> ```Go
> 	}
> 	ctx, cancelF := context.WithTimeout(context.Background(), time.Second)
> 	localTRC, err := trustStore.GetValidTRC(
> ```

Hmm. How is an updated/new TRC loaded from the config dir? (Or, to put it another way, scion elements should load any TRC/Cert from the paths that config provides. The contents of the trustdb are secondary to these.)

Also, it looks like this will actually query the network if the local TRC isn't in the trustdb, which doesn't make sense. If you don't have the local TRC, you have zero information with which to verify any response.

---

*[go/sciond/main.go, line 153 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5XW4n73WdlwxvUJXE:-LG5XW4n73WdlwxvUJXF:btf5env) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L153)):*
> ```Go
> 				msger,
> 				pathDB,
> 				localTRC.CoreASList(),
> ```

This ignores TRC updates.

---

*[go/sciond/main.go, line 161 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5Xd5t9WULKSSnpDAG:-LG5Xd5t9WULKSSnpDAH:btf5env) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L161)):*
> ```Go
> 		proto.SCIONDMsg_Which_asInfoReq: &servers.ASInfoRequestHandler{
> 			TrustStore: trustStore,
> 			CoreASes:   localTRC.CoreASList(),
> ```

This ignores TRC updates.

---

*[go/sciond/main.go, line 163 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5annL34jAsRUPt_rA:-LG5annL34jAsRUPt_rB:b-o0bdl7) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L163)):*
> ```Go
> 			CoreASes:   localTRC.CoreASList(),
> 			Messenger:  msger,
> 			Topology:   config.General.Topology,
> ```

This (presumably) doesn't allow topo updates. Same for the other places.

---

*[go/sciond/main.go, line 254 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5aK462ygq_I291Kuf:-LG5aK462ygq_I291Kug:b-b3jbun) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L254)):*
> ```Go
> 	}
> 	dispatcher := disp.New(transport.NewPacketTransport(conn), messenger.DefaultAdapter, logger)
> 	// TODO: initialize actual trust store once it is available
> ```

Obsolete?

---

*[go/sciond/main.go, line 262 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5T_GP2pTGX4iY-5LU:-LG5T_GP2pTGX4iY-5LV:b-4gxcnb) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L262)):*
> ```Go
> 	logger log.Logger) (*servers.Server, func()) {
> 
> 	// FIXME(scrye): enable msger below
> ```

Is this still relevant?

---

*[go/sciond/internal/fetcher/revcache.go, line 47 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5Wzqp4HYtM7nFhexo:-LG5Wzqp4HYtM7nFhexp:bftyao1) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/fetcher/revcache.go#L47)):*
> ```Go
> }
> 
> func (c *RevCache) Add(ia addr.IA, ifid common.IFIDType, rev *path_mgmt.SignedRevInfo, ttl time.Duration) {
> ```

Line length.

---

*[go/sciond/internal/servers/api.go, line 34 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5_vRF8U0m_epoJVlk:-LG5_vRF8U0m_epoJVll:bn7ht39) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/api.go#L34)):*
> ```Go
> 
> const (
> 	DefaultRequestTimeout = 3 * time.Second
> ```

This isn't used.

---

*[go/sciond/internal/servers/handlers.go, line 37 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5ciy9Ed39jRWvKgt5:-LG5ciy9Ed39jRWvKgt6:b58dta3) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L37)):*
> ```Go
> const (
> 	DefaultHandlerLifetime = 10 * time.Second
> 	DefaultEarlyReply      = 2 * time.Second
> ```

This is far too long. For a start, python apps currently have a 2-3s timeout for the complete path query, which is pretty reasonable. That's why the existing sciond uses a 200ms early reply timer. I suggest doing the same here. We could, in the future, consider making it relative to the client-supplied timer, or just allowing the client to specify it if they want.

---

*[go/sciond/internal/servers/handlers.go, line 52 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5kQgqCSHEtXdB26L2:-LG5kQgqCSHEtXdB26L3:bxi1blw) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L52)):*
> ```Go
> 	defer cancelF()
> 	req := pld.PathReq
> 	getPathsReply, err := h.Fetcher.GetPaths(ctx, &req, DefaultEarlyReply)
> ```

`&pld.PathReq`

The request Id isn't being used anywhere - it would be good to log that with any relevant messages, otherwise debugging sciond will be a lot harder.

---

*[go/sciond/internal/servers/handlers.go, line 57 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5iuokB2wAIqVsWyN4:-LG5iuokB2wAIqVsWyN5:b8ovyl) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L57)):*
> ```Go
> 	}
> 	// Always reply, as the Fetcher will fill in the relevant error bits of the reply
> 	reply := &sciond.Pld{
> ```

Ah. I was getting confused by all the proto-struct boiler-plate, until i realised that `sciond.Pld` doesn't follow the design of union handling used in `go/lib/ctrl`. We should fix that. Filed https://github.com/scionproto/scion/issues/1637

---

*[go/sciond/internal/servers/handlers.go, line 91 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5cFmxDdGEFpney16y:-LG5cFmxDdGEFpney16z:b-8s65k2) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L91)):*
> ```Go
> 	asInfoRequest := pld.AsInfoReq
> 	asInfoReply := sciond.ASInfoReply{}
> 	if asInfoRequest.Isdas.IA().IsZero() || asInfoRequest.Isdas.IA().Eq(h.Topology.ISD_AS) {
> ```

Rather than calling `asInfoRequest.Isdas.IA()` 4 times, how about `reqIA := sInfoRequest.Isdas.IA()`, and reference that instead? (Bonus: it's a lot easier to read).

---

*[go/sciond/internal/servers/handlers.go, line 102 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5bQ2gBhfIGAeXXJEU:-LG5bQ2gBhfIGAeXXJEV:bwedzix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L102)):*
> ```Go
> 	} else {
> 		// Requested AS is not us
> 		trcObj, err := h.TrustStore.GetTRC(context.TODO(), asInfoRequest.Isdas.IA().I, 0)
> ```

`s/context.TODO()/ctx/`?

---

*[go/sciond/internal/servers/handlers.go, line 125 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5lWpc42V5M795_VwN:-LG5lWpc42V5M795_VwO:b-d9ojgp) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L125)):*
> ```Go
> 	b, err := proto.PackRoot(reply)
> 	if err != nil {
> 		log.Error("unable to serialize SCIONDMsg reply")
> ```

If this errors, i'm not sure it makes sense to send the empty `b`. Same elsewhere.

---

*[go/sciond/internal/servers/handlers.go, line 158 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5llVB8zMNcY514nQi:-LG5llVB8zMNcY514nQj:b1afndn) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L158)):*
> ```Go
> 			ifInfo, ok := h.Topology.IFInfoMap[ifid]
> 			if !ok {
> 				log.Info("Received IF Info Request, but IFID not found", "ifid", ifid)
> ```

The logging seems to be inconsistent - sometimes it uses the handler's logger, sometimes not.

---

*[go/sciond/internal/servers/handlers.go, line 203 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5m2L5C4dhAmWN4Yrr:-LG5m2L5C4dhAmWN4Yrs:b-h39wfh) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L203)):*
> ```Go
> 		case sciond.SvcCS:
> 			hostInfos = makeHostInfos(h.Topology.CS)
> 		case sciond.SvcBR:
> ```

I don't know how i missed it before, but, an SVC address for the BR doesn't make sense. `sciond.SvcBR` should be dropped.

---

*[go/sciond/internal/servers/handlers.go, line 210 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5mGBh-PEulv8fN3zG:-LG5mGBh-PEulv8fN3zH:b-waqle1) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L210)):*
> ```Go
> 		replyEntry := sciond.ServiceInfoReplyEntry{
> 			ServiceType: t,
> 			Ttl:         1337,
> ```

This shouldn't be hard-coded here. For now we can make it a `const` for, say,  5 minutes. Eventually it'll probably be dynamic.

---

*[go/sciond/internal/servers/handlers.go, line 251 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5noVTBWz_h3r41VyL:-LG5noVTBWz_h3r41VyM:b3p4wcp) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L251)):*
> ```Go
> 
> func topoAddrToHostInfo(topoAddr topology.TopoAddr) sciond.HostInfo {
> 	// FIXME(scrye): also add support for IPv6
> ```

```
var v4, v6 net.IP
ot := config.General.Topology.Overlay
if config.General.Topology.Overlay.IsIPv4() {
  v4 = topoAddr.PublicAddrInfo(ot.To4()).IP
}
if config.General.Topology.Overlay.IsIPv6() {
  v6 = topoAddr.PublicAddrInfo(ot.To6()).IP
}
port := topoAddr.PublicAddrInfo(ot).L4Port
```

---

*[go/sciond/internal/servers/handlers.go, line 266 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5ouzKBPGqDfxe1b3f:-LG5ouzKBPGqDfxe1b3g:b7n0lkf) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L266)):*
> ```Go
> // RevNotification announcements. The SCIOND API spawns a goroutine with method Handle
> // for each RevNotification it receives.
> type RevNotificationHandler struct {
> ```

How does sciond deal with revocations it receives over scmp? It currently does not send any packet outside the local AS (i believe), but that may change in the future, and in any case all well-mannered scion apps should be able to handle scmp revocations.

---

*[go/sciond/internal/servers/server.go, line 44 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5Ts1I4Yf9gve4cLcp:-LG5Ts1I4Yf9gve4cLcq:b-w6opaz) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/server.go#L44)):*
> ```Go
> 
> // NewServer initializes a new server at address on the specified network. The
> // server will use msger for network access. To start listening on the address,
> ```

`msger` isn't a parameter anymore.

---

*[go/sciond/internal/servers/server.go, line 62 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5Yo-D-HZXv50QJKrh:-LG5Yo-D-HZXv50QJKri:b-h70rsk) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/server.go#L62)):*
> ```Go
> // until it is closed by the client.
> func (srv *Server) ListenAndServe() error {
> 	srv.mu.Lock()
> ```

The locking here is racy. I think the locking is unnecessary complexity, but if it's going to be there it should at least be correct. You need to assign to `srv.listener` while the mutex is held.

---

*[go/sciond/internal/servers/server.go, line 82 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5_XiHAQ6CBXQ_Qu-v:-LG5_XiHAQ6CBXQ_Qu-w:b-t5uugg) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/server.go#L82)):*
> ```Go
> 			defer log.LogPanicAndExit()
> 			pconn := conn.(net.PacketConn)
> 			NewAPI(transport.NewPacketTransport(pconn), srv.handlers).Serve()
> ```

This discards the return value from `Server()`

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)*

scrye:
Review status: 4 of 14 files reviewed, 24 unresolved discussions (waiting on @kormat and @scrye)

---

*[go/sciond/main.go, line 109 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5EhkZ9YTuvCHcXN2s:-LG9KqJ50-E0sd0cpQAi:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L109)):*

Using a hard-coded start ID rather defeats the purpose ;)
</blockquote></details>

Done.

---

*[go/sciond/main.go, line 119 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5FA5pAbLSBBBOgoqr:-LG9Oxej1jZ2tw-gwAEy:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L119)):*

This ignores the Bind address.
</blockquote></details>

Done.

---

*[go/sciond/main.go, line 125 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5FqdgB6ySN5neBsUu:-LG9PBEsD0pZ3ejZPqYS:bg0c1s3) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L125)):*

Hmm. How is an updated/new TRC loaded from the config dir? (Or, to put it another way, scion elements should load any TRC/Cert from the paths that config provides. The contents of the trustdb are secondary to these.)

Also, it looks like this will actually query the network if the local TRC isn't in the trustdb, which doesn't make sense. If you don't have the local TRC, you have zero information with which to verify any response.
</blockquote></details>

The TrustDB allows us to have a single source of truth for all crypto material, which should include the local TRC. Otherwise, processes might get in weird states where they load a TRC from file, and the _same_ TRC from the database. This gets nastier if the two disagree due to some misconfiguration.

Always loading from the database seems like the simpler approach here.

The trust store doesn't go on the network when querying for something like `(ctx, isdX, isdX)`, because it tries to follow the trail of `isdX` (the exact trc we want), doesn't find it, and errors out immediately.

I've reworked a bit how the fetcher uses TRCs, s.t. it always loads fresh TRC state from the DB. This way, we don't need to add complexity for reloading.

---

*[go/sciond/main.go, line 153 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5XW4n73WdlwxvUJXE:-LG9VKrm0_J5k_Z9Esqp:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L153)):*

This ignores TRC updates.
</blockquote></details>

Done.

---

*[go/sciond/main.go, line 161 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5Xd5t9WULKSSnpDAG:-LG9VLSl0AJ58YaBkR0T:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L161)):*

This ignores TRC updates.
</blockquote></details>

Done.

---

*[go/sciond/main.go, line 163 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5annL34jAsRUPt_rA:-LG9WdFF7KtuSR7BErn7:b-u69ssf) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L163)):*

This (presumably) doesn't allow topo updates. Same for the other places.
</blockquote></details>

Yes, reloading is not supported yet. I've added a FIXME.

---

*[go/sciond/main.go, line 254 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5aK462ygq_I291Kuf:-LG9Z2KO40jQFYPLx4fS:b-vpdics) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L254)):*

Obsolete?
</blockquote></details>

Yes, the whole function. Done.

---

*[go/sciond/main.go, line 262 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5T_GP2pTGX4iY-5LU:-LG9Z8KCCm4yzUpYrWx4:bddp37h) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L262)):*

Is this still relevant?
</blockquote></details>

Nope, done.

---

*[go/sciond/internal/fetcher/revcache.go, line 47 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5Wzqp4HYtM7nFhexo:-LG9ZGms-B_SoA-ejKXZ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/fetcher/revcache.go#L47)):*

Line length.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/api.go, line 34 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5_vRF8U0m_epoJVlk:-LG9ZLuF7Icb9DKSEm1P:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/api.go#L34)):*

This isn't used.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/handlers.go, line 37 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5ciy9Ed39jRWvKgt5:-LG9ZSV4Ed3XiHcNlPtW:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L37)):*

This is far too long. For a start, python apps currently have a 2-3s timeout for the complete path query, which is pretty reasonable. That's why the existing sciond uses a 200ms early reply timer. I suggest doing the same here. We could, in the future, consider making it relative to the client-supplied timer, or just allowing the client to specify it if they want.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/handlers.go, line 52 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5kQgqCSHEtXdB26L2:-LG9bHQ0B5aybCBhjr32:bxtst01) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L52)):*

`&pld.PathReq`

The request Id isn't being used anywhere - it would be good to log that with any relevant messages, otherwise debugging sciond will be a lot harder.
</blockquote></details>

Done. The logging context now always comes populated with the id from the `ConnHandler`.

---

*[go/sciond/internal/servers/handlers.go, line 57 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5iuokB2wAIqVsWyN4:-LG9cUc-9ysCyRZgIfe4:bb6sh2i) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L57)):*

Ah. I was getting confused by all the proto-struct boiler-plate, until i realised that `sciond.Pld` doesn't follow the design of union handling used in `go/lib/ctrl`. We should fix that. Filed https://github.com/scionproto/scion/issues/1637
</blockquote></details>

Ack.

---

*[go/sciond/internal/servers/handlers.go, line 91 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5cFmxDdGEFpney16y:-LG9ds3K63V7yMRFTR9U:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L91)):*

Rather than calling `asInfoRequest.Isdas.IA()` 4 times, how about `reqIA := sInfoRequest.Isdas.IA()`, and reference that instead? (Bonus: it's a lot easier to read).
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/handlers.go, line 102 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5bQ2gBhfIGAeXXJEU:-LG9dxXd4U8ElQYs5ZVE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L102)):*

`s/context.TODO()/ctx/`?
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/handlers.go, line 125 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5lWpc42V5M795_VwN:-LG9e8vUAHKxSaYFlSc4:b-fgosu7) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L125)):*

If this errors, i'm not sure it makes sense to send the empty `b`. Same elsewhere.
</blockquote></details>

Ah, I think a panic here makes more sense, as this only fails on programming errors (bad schema probably), which is something an app cannot recover from. 

---

*[go/sciond/internal/servers/handlers.go, line 158 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5llVB8zMNcY514nQi:-LG9fuNp3bmFxKpsAXq8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L158)):*

The logging seems to be inconsistent - sometimes it uses the handler's logger, sometimes not.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/handlers.go, line 203 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5m2L5C4dhAmWN4Yrr:-LG9hBBQCImqqm4cqZ9K:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L203)):*

I don't know how i missed it before, but, an SVC address for the BR doesn't make sense. `sciond.SvcBR` should be dropped.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/handlers.go, line 210 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5mGBh-PEulv8fN3zG:-LG9huIL6HtHCUN0kvCU:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L210)):*

This shouldn't be hard-coded here. For now we can make it a `const` for, say,  5 minutes. Eventually it'll probably be dynamic.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/handlers.go, line 251 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5noVTBWz_h3r41VyL:-LG9icr-1Lx-_KzPZIK5:bxos2bf) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L251)):*

```
var v4, v6 net.IP
ot := config.General.Topology.Overlay
if config.General.Topology.Overlay.IsIPv4() {
  v4 = topoAddr.PublicAddrInfo(ot.To4()).IP
}
if config.General.Topology.Overlay.IsIPv6() {
  v6 = topoAddr.PublicAddrInfo(ot.To6()).IP
}
port := topoAddr.PublicAddrInfo(ot).L4Port
```
</blockquote></details>

That's beautiful, thanks!

---

*[go/sciond/internal/servers/handlers.go, line 266 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5ouzKBPGqDfxe1b3f:-LG9i41TD8CAgrf3sXEk:b-dtfcxk) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L266)):*

How does sciond deal with revocations it receives over scmp? It currently does not send any packet outside the local AS (i believe), but that may change in the future, and in any case all well-mannered scion apps should be able to handle scmp revocations.
</blockquote></details>

It will probably need a way to tell `snet` that it wants to handle SCMP revocation messages itself, and run code similar to the one in this handler if it sees one. We can revisit this when we get to that point, but it doesn't look hard to do.

---

*[go/sciond/internal/servers/server.go, line 44 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5Ts1I4Yf9gve4cLcp:-LG9kafW3lqyWbaY9nT2:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/server.go#L44)):*

`msger` isn't a parameter anymore.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/server.go, line 62 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5Yo-D-HZXv50QJKrh:-LG9kqR64UhS1Dd4wwZv:bddabg9) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/server.go#L62)):*

The locking here is racy. I think the locking is unnecessary complexity, but if it's going to be there it should at least be correct. You need to assign to `srv.listener` while the mutex is held.
</blockquote></details>

Ah, good point. Fixed.

---

*[go/sciond/internal/servers/server.go, line 82 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5_XiHAQ6CBXQ_Qu-v:-LG9mSpO8x7g5wqkR3aQ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/server.go#L82)):*

This discards the return value from `Server()`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)*

kormat:
Reviewed 1 of 11 files at r1, 9 of 9 files at r2.
Review status: all files reviewed, 3 unresolved discussions (waiting on @kormat and @scrye)

---

*[go/sciond/main.go, line 125 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5FqdgB6ySN5neBsUu:-LGAYwEK30DzsXC5pf0m:bo6pcok) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L125)):*

The TrustDB allows us to have a single source of truth for all crypto material, which should include the local TRC. Otherwise, processes might get in weird states where they load a TRC from file, and the _same_ TRC from the database. This gets nastier if the two disagree due to some misconfiguration.

Always loading from the database seems like the simpler approach here.

The trust store doesn't go on the network when querying for something like `(ctx, isdX, isdX)`, because it tries to follow the trail of `isdX` (the exact trc we want), doesn't find it, and errors out immediately.

I've reworked a bit how the fetcher uses TRCs, s.t. it always loads fresh TRC state from the DB. This way, we don't need to add complexity for reloading.
</blockquote></details>

As discussed offline, on startup/reload the on-disk TRC/cert should be loaded and inserted into the DB. If it conflicts with the db contents, that's a fatal error. The application will only ever query the DB.

---

*[go/sciond/internal/servers/handlers.go, line 251 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5noVTBWz_h3r41VyL:-LGAmNgd7phjY7gaEMNn:ba2tx55) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L251)):*

That's beautiful, thanks!
</blockquote></details>

You can also drop the FIXME now :)

---

*[go/sciond/internal/servers/handlers.go, line 61 at r2](https://reviewable.io/reviews/scionproto/scion/1632#-LGAnX6tCgr5_v_FXFRK:-LGAnX6tCgr5_v_FXFRL:b-j7n27l) ([raw file](https://github.com/scionproto/scion/blob/e19f4367ae45d8b7278d44a47b31d270db2d434b/go/sciond/internal/servers/handlers.go#L61)):*
> ```Go
> 	defer cancelF()
> 	pathReq := &pld.PathReq
> 	getPathsReply, err := h.Fetcher.GetPaths(ctx, pathReq, DefaultEarlyReply)
> ```

What's the purpose of the `pathReq` temp var, rather than just using `&pld.PathReq`?

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)*

scrye:
Review status: 10 of 15 files reviewed, 3 unresolved discussions (waiting on @kormat and @scrye)

---

*[go/sciond/main.go, line 125 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5FqdgB6ySN5neBsUu:-LGB5q311Sn5qKysga7E:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/main.go#L125)):*

As discussed offline, on startup/reload the on-disk TRC/cert should be loaded and inserted into the DB. If it conflicts with the db contents, that's a fatal error. The application will only ever query the DB.
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/handlers.go, line 251 at r1](https://reviewable.io/reviews/scionproto/scion/1632#-LG5noVTBWz_h3r41VyL:-LGApDq4-qe1AXEtHPHO:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/58c47305a4907562336d2c55e1dfca007c4dc6a5/go/sciond/internal/servers/handlers.go#L251)):*

You can also drop the FIXME now :)
</blockquote></details>

Done.

---

*[go/sciond/internal/servers/handlers.go, line 61 at r2](https://reviewable.io/reviews/scionproto/scion/1632#-LGAnX6tCgr5_v_FXFRK:-LGApO8N37InlBD78vFd:b-ayfdv1) ([raw file](https://github.com/scionproto/scion/blob/e19f4367ae45d8b7278d44a47b31d270db2d434b/go/sciond/internal/servers/handlers.go#L61)):*

What's the purpose of the `pathReq` temp var, rather than just using `&pld.PathReq`?
</blockquote></details>

Ah, it used to be used in multiple places so I got used to it, changed it now.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)*

kormat:
Reviewed 5 of 5 files at r3.
Review status: all files reviewed, 4 unresolved discussions (waiting on @scrye)

---

*[go/sciond/main.go, line 124 at r3](https://reviewable.io/reviews/scionproto/scion/1632#-LGB7i_z7DXe7TJgBz5c:-LGB7i_z7DXe7TJgBz5d:b-894hk6) ([raw file](https://github.com/scionproto/scion/blob/8cbf5829e3803c8b465e942f70003e1c7cc6997a/go/sciond/main.go#L124)):*
> ```Go
> 		return 1
> 	}
> 	log.Warn("public", config.SD.Public, "bind", config.SD.Bind)
> ```

Drop.

---

*[go/sciond/main.go, line 238 at r3](https://reviewable.io/reviews/scionproto/scion/1632#-LGB9hkCB6JcBVafo0mN:-LGB9hkCB6JcBVafo0mO:bwolel5) ([raw file](https://github.com/scionproto/scion/blob/8cbf5829e3803c8b465e942f70003e1c7cc6997a/go/sciond/main.go#L238)):*
> ```Go
> 		config.SD.Unix = "/run/shm/sciond/default-unix.sock"
> 	}
> 	if config.SD.Bind.IsZero() {
> ```

As discussed on slack, i'd prefer to take a different approach, and have `config.SD.Bind` be a pointer, as this current approach breaks a number of assumptions.

---

*[go/sciond/main.go, line 277 at r3](https://reviewable.io/reviews/scionproto/scion/1632#-LGBApzw2gZ0M2MRiP3q:-LGBApzw2gZ0M2MRiP3r:b-y41oor) ([raw file](https://github.com/scionproto/scion/blob/8cbf5829e3803c8b465e942f70003e1c7cc6997a/go/sciond/main.go#L277)):*
> ```Go
> 	case err != nil && fileTRC == nil:
> 		return common.NewBasicError("No TRC found", nil)
> 	case err != nil && fileTRC != nil:
> ```

I'd be happier if it could be determined that the precise error from the trust store indicates "not found", as opposed to any other error.

---

*[go/sciond/main.go, line 287 at r3](https://reviewable.io/reviews/scionproto/scion/1632#-LGBB67f6OGrGwg9GaCa:-LGBB67f6OGrGwg9GaCb:b1pbmk8) ([raw file](https://github.com/scionproto/scion/blob/8cbf5829e3803c8b465e942f70003e1c7cc6997a/go/sciond/main.go#L287)):*
> ```Go
> 	case err == nil && fileTRC != nil:
> 		// Check if the TRCs match
> 		if fileTRC.String() != dbTRC.String() {
> ```

This doesn't look right. It looks like this will generate two strings of the form "TRC XvY", it doesn't compare the _contents_ of the TRCs. Comparing signatures is probably better.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)*

scrye:
Review status: 12 of 16 files reviewed, 4 unresolved discussions (waiting on @kormat)

---

*[go/sciond/main.go, line 124 at r3](https://reviewable.io/reviews/scionproto/scion/1632#-LGB7i_z7DXe7TJgBz5c:-LGPYeLG7HK4-j1hJwCo:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/8cbf5829e3803c8b465e942f70003e1c7cc6997a/go/sciond/main.go#L124)):*

Drop.
</blockquote></details>

Done.

---

*[go/sciond/main.go, line 238 at r3](https://reviewable.io/reviews/scionproto/scion/1632#-LGB9hkCB6JcBVafo0mN:-LGPZUfGEiXvFmDY6smk:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/8cbf5829e3803c8b465e942f70003e1c7cc6997a/go/sciond/main.go#L238)):*

As discussed on slack, i'd prefer to take a different approach, and have `config.SD.Bind` be a pointer, as this current approach breaks a number of assumptions.
</blockquote></details>

Done.

---

*[go/sciond/main.go, line 277 at r3](https://reviewable.io/reviews/scionproto/scion/1632#-LGBApzw2gZ0M2MRiP3q:-LGPwe6Z1Dk7GOWf5YHZ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/8cbf5829e3803c8b465e942f70003e1c7cc6997a/go/sciond/main.go#L277)):*

I'd be happier if it could be determined that the precise error from the trust store indicates "not found", as opposed to any other error.
</blockquote></details>

Done.

---

*[go/sciond/main.go, line 287 at r3](https://reviewable.io/reviews/scionproto/scion/1632#-LGBB67f6OGrGwg9GaCa:-LGPweS26WNpVJpXnWnT:bsnbkq3) ([raw file](https://github.com/scionproto/scion/blob/8cbf5829e3803c8b465e942f70003e1c7cc6997a/go/sciond/main.go#L287)):*

This doesn't look right. It looks like this will generate two strings of the form "TRC XvY", it doesn't compare the _contents_ of the TRCs. Comparing signatures is probably better.
</blockquote></details>

Done, I went for full JSON comparison because it's a rare operation, and it will catch all the weird inconsistencies that might confuse users (even if it's related to some unsigned metadata).

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)*

kormat:
Reviewed 4 of 4 files at r4.
Review status: all files reviewed, 7 unresolved discussions (waiting on @kormat and @scrye)

---

*[go/lib/env/env.go, line 74 at r4](https://reviewable.io/reviews/scionproto/scion/1632#-LGU5D_I3SarJ-xMcWtQ:-LGU5D_I3SarJ-xMcWtR:b-clf01m) ([raw file](https://github.com/scionproto/scion/blob/da1545d9135c378daeb2bde8c36e6cf89c5a8ed4/go/lib/env/env.go#L74)):*
> ```Go
> 	Topology *topology.Topo `toml:"-"`
> 	// Seed for the PRNG. A value of 0 means use current UNIX time as seed.
> 	Seed int64
> ```

What is the purpose of including this in the config file? A static value is a bad idea. Is it here purely for testing?

---

*[go/lib/infra/modules/trust/trust.go, line 45 at r4](https://reviewable.io/reviews/scionproto/scion/1632#-LGU0aI1E_QhRJNlBt8e:-LGU0aI1E_QhRJNlBt8f:b-e3hfpn) ([raw file](https://github.com/scionproto/scion/blob/da1545d9135c378daeb2bde8c36e6cf89c5a8ed4/go/lib/infra/modules/trust/trust.go#L45)):*
> ```Go
> 
> var (
> 	ErrEndOfTrail = common.NewBasicError("reached end of trail, but no trusted TRC found", nil)
> ```

Our usual approach is to have a const string, and then to use `common.GetErrorMsg(err)`. Have a look at `addr.ErrorBadHostAddrType` for an example.

---

*[go/sciond/main.go, line 238 at r3](https://reviewable.io/reviews/scionproto/scion/1632#-LGB9hkCB6JcBVafo0mN:-LGU1B3r8MW2Xnylct_9:b-kkbjsk) ([raw file](https://github.com/scionproto/scion/blob/8cbf5829e3803c8b465e942f70003e1c7cc6997a/go/sciond/main.go#L238)):*

Done.
</blockquote></details>

Ah, not quite what i meant:
- `config.SD.Bind` should be a pointer (as you've done here)
- It should _not_ be initialized if it's `nil`. That way a `nil` gets passed into `snet`, etc.
- Optionally add a `config.SD.GetBindAddr()` method that does:
```
if Bind == nil {
  return Public
}
return Bind
```
Most likely an app won't care about the bind address, it just passes the `Bind` pointer to snet and snet takes care of the details.

This is how Public/Bind is handled in the rest of the codebase.

---

*[go/sciond/main.go, line 280 at r4](https://reviewable.io/reviews/scionproto/scion/1632#-LGU3cK204gcBk5QMXeA:-LGU3cK204gcBk5QMXeB:b6uszd1) ([raw file](https://github.com/scionproto/scion/blob/da1545d9135c378daeb2bde8c36e6cf89c5a8ed4/go/sciond/main.go#L280)):*
> ```Go
> 	case err == trust.ErrEndOfTrail && fileTRC == nil:
> 		return common.NewBasicError("No TRC found on disk or in trustdb", nil)
> 	case err == trust.ErrEndOfTrail && fileTRC != nil:
> ```

Hmm, wait. This logic doesn't cover the case of a TRC in the trust store and a newer version of the TRC on disk. There needs to be a version comparison.

---

*[go/sciond/main.go, line 303 at r4](https://reviewable.io/reviews/scionproto/scion/1632#-LGU3qQo6xxKVpTKcDHz:-LGU3qQo6xxKVpTKcDI-:b-q97egr) ([raw file](https://github.com/scionproto/scion/blob/da1545d9135c378daeb2bde8c36e6cf89c5a8ed4/go/sciond/main.go#L303)):*
> ```Go
> 
> // equals checks if two TRCs are the same based on their JSON serializations
> func equals(x, y *trc.TRC) (bool, error) {
> ```

This seems like a reasonable thing to add to `trc.TRC` type instead.

---

*[go/sciond/internal/servers/handlers.go, line 267 at r4](https://reviewable.io/reviews/scionproto/scion/1632#-LGU4MF73QJh6H3z_4sP:-LGU4MF73QJh6H3z_4sQ:bawalr9) ([raw file](https://github.com/scionproto/scion/blob/da1545d9135c378daeb2bde8c36e6cf89c5a8ed4/go/sciond/internal/servers/handlers.go#L267)):*
> ```Go
> }
> 
> func MakeBRHostInfos(ot overlay.Type, brMap map[string]topology.BRInfo,
> ```

This isn't used outside of test code, so maybe move it there?

---

*[go/sciond/internal/servers/handlers.go, line 273 at r4](https://reviewable.io/reviews/scionproto/scion/1632#-LGU4Ol8EaN4cCm_D3uk:-LGU4Ol8EaN4cCm_D3ul:bjeqkx7) ([raw file](https://github.com/scionproto/scion/blob/da1545d9135c378daeb2bde8c36e6cf89c5a8ed4/go/sciond/internal/servers/handlers.go#L273)):*
> ```Go
> 	for _, brInfo := range brMap {
> 		// One IFID is enough to find the unique internal address. Panic if no
> 		// IFIDs exist.
> ```

It's legal for a BR to have no interfaces. This will usually only happen while a link is being setup etc, but it's still allowed. If this is method is moved to test code, then this probably doesn't matter any more.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)*

kormat:
Reviewed 5 of 5 files at r5.
Review status: all files reviewed, 1 unresolved discussion (waiting on @scrye)

---

*[go/sciond/main.go, line 290 at r5](https://reviewable.io/reviews/scionproto/scion/1632#-LGUtdes64lp4_J-gz4u:-LGUtdes64lp4_J-gz4v:bfmvpof) ([raw file](https://github.com/scionproto/scion/blob/72f7aa17444ce8b06602598395b628ca24a0ec25/go/sciond/main.go#L290)):*
> ```Go
> 			return err
> 		}
> 		// Because it is an older version, check if the TRCs match
> ```

This isn't a good assumption. Say in the db you have: v1, v2, v3, and on disk you have v2. The comparison will fail, because it will be checking if v3 == v2.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)*

scrye:
Review status: 15 of 16 files reviewed, 1 unresolved discussion (waiting on @kormat)

---

*[go/sciond/main.go, line 290 at r5](https://reviewable.io/reviews/scionproto/scion/1632#-LGUtdes64lp4_J-gz4u:-LGUygpoDkYQ5uZ_UHKp:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/72f7aa17444ce8b06602598395b628ca24a0ec25/go/sciond/main.go#L290)):*

This isn't a good assumption. Say in the db you have: v1, v2, v3, and on disk you have v2. The comparison will fail, because it will be checking if v3 == v2.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)*

kormat:
Reviewed 1 of 1 files at r6.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632)*

kormat:
:lgtm:

---

Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1632#-:-LGV162330vudSmv1m_7:bnfp4nl)*

PR #1647, opened 2018-07-03
Verify network address in sig.json fixes #1596

### LINKED ISSUE #1596: SIG: networks in sig.json should be canonical
Currently the sig will accept a net of, e.g. `10.0.0.55/24`, which is not a valid network, it's a host address with mask. The SIG should be stricter, and reject host addresses in the network list.

lukedirtwalker:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1647)

### DIFF
--- a/go/sig/config/config.go
+++ b/go/sig/config/config.go
@@ -72,9 +72,13 @@ func (in *IPNet) UnmarshalJSON(b []byte) error {
 	if err := json.Unmarshal(b, &s); err != nil {
 		return common.NewBasicError("Unable to unmarshal IPnet from JSON", err, "raw", b)
 	}
-	_, ipnet, err := net.ParseCIDR(s)
+	ip, ipnet, err := net.ParseCIDR(s)
 	if err != nil {
 		return common.NewBasicError("Unable to parse IPnet string", err, "raw", s)
+	}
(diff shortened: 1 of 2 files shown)


kormat:
Reviewed 1 of 3 files at r1.
Review status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker and @kormat)

---

*[go/sig/config/config.go, line 86 at r1](https://reviewable.io/reviews/scionproto/scion/1647#-LGVEJKw3xPOOrVJx6xL:-LGVEJKw3xPOOrVJx6xM:b-eqd0go) ([raw file](https://github.com/scionproto/scion/blob/9b6785a3887db3df4999c439c88aaa0371a43d81/go/sig/config/config.go#L86)):*
> ```Go
> }
> 
> func verifyNetworkAddr(ip net.IP, mask net.IPMask, rawCIDR string) error {
> ```

I think this can be simplified. Something like:
```
nip := ip.Mask(mask)
if !nip.Equal(ip) {
   return common.NewBasicError(...)
}
return nil
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/scionproto/scion/1647)*

PR #1669, opened 2018-07-11
DRKey: specification

benrothen:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1669)


Oncilla:
___
*[doc/DRKeyInfra.md, line 112 at r2](https://reviewable.io/reviews/scionproto/scion/1669#-LK0nMge-9-YWweo82ac:-LLUQbo43wP-4VdY-BhX:b-51jg13) ([raw file](https://github.com/scionproto/scion/blob/68b55eed6a8d0cd3f9c9001e56985648cb21621e/doc/DRKeyInfra.md#L112)):*
<details><summary><i>Previously, benrothen wrote…</i></summary><blockquote>

The request contains a timestamp, which implicitly translates to the issuing epoch. You can ask your CS for the previous epoch's key using`valTime`.
</blockquote></details>

We do not need to include the issuing epoch. As @benrothen stated correctly, the host can specify the `valTime` during a request for the 2lvl DRKey.
 
There are two solutions to this issue:
- The authenticated packet contains a timestamp: The host can supply this timestamp during the request to the CS to receive the right key for the epoch.
- The authenticated packet does not contain a timestamp: There can be at most two keys which could be potentially valid -> try both.

Imo, we should always aim for the first case, since authenticated data should be accompanied by a timestamp. 
However, it is a on-wire size vs verification time trade-off.

PR #1675, opened 2018-07-12
Migrate CSGO to infra stack

scrye:
All features were migrated, with the exception of reloading which is currently disabled. Reloading will be implemented once the CS is migrated to use the new service config file format.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1675)


scrye:
___
*[go/lib/infra/modules/trust/trust.go, line 581 at r5](https://reviewable.io/reviews/scionproto/scion/1675#-LIAe7mvArrKuKW7a-Go:-LIQ-ix87wDSsa45B6bl:b-mgzo3u) ([raw file](https://github.com/scionproto/scion/blob/fea160f50a94f80e7795bc099239fb1cbed6327a/go/lib/infra/modules/trust/trust.go#L581)):*

@oncilla : ah, this is a terminology issue, then. For me, "invalidated" means, well, it's no longer valid, and can't be used for anything. So, i would phrase this differently: 

> An AS must get an updated cert any time the TRC changes. If the Grace time has expired then the cert renewal must be done via an out-of-band mechanism.
</blockquote></details>

From my understanding (anyone correct me if I'm wrong), only _certain_ TRC updates invalidate the Chain. And w.r.t. those TRC updates, the chain is indeed no longer valid, and a fresh one needs to be grabbed while the grace period is in effect. For other TRC updates the AS doesn't need to do anything.

Issue #1670, opened 2018-07-11
Infra: Add support for QUIC transport


lukedirtwalker:
Currently infra uses UDP for transport. 

We should also add support for QUIC transport.

scrye:
This was implemented by https://github.com/scionproto/scion/pull/2522.

Due to the current behavior of anycasts, https://github.com/scionproto/scion/issues/2554 should be implemented to allow for multiple server instances in remote ASes.

PR #1680, opened 2018-07-16
Blocked: Implement QUIC transport

lukedirtwalker:
Contributes to #1670

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1680)


lukedirtwalker:
depends on https://github.com/scionproto/scion/pull/1682

scrye:
Disregard ^, somehow this window had focus. I was wondering where my key presses went.

lukedirtwalker:
Currently on Hold:
This quic transport should actually use squic. But this would trigger an import cycle. We first need to clean up the code base (create interfaces), once we have that we can finally end the implementation of this.

scrye:
QUIC is now plugged directly in our RPC package as of https://github.com/scionproto/scion/pull/2664, and this is no longer needed.

PR #1724, opened 2018-07-26
Introduce new interfaces for Address types

sgmonroy:
This change introduces two new interfaces AppAddr and OverlayAddr.
Both have a similar interface but the work with different Address Types,
ie. SVC as an AppAddr.

Internally all current objects implementing the interfaces consist of
a HostAddr and a port.

Only allowing to create objects that implement these interfaces through
constructors, we can restrict which HostAddr as valid for each
interface.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1724)


sgmonroy:
___
*[go/lib/addr/addrport.go, line 30 at r1](https://reviewable.io/reviews/scionproto/scion/1724#-LIf6SGA6INH73XiUcJS:-LIjBbOu-BZoS5GIQkig:b-eyfmpd) ([raw file](https://github.com/scionproto/scion/blob/34b3dd8081e24fb8def73f4494aa39a5192b07e1/go/lib/addr/addrport.go#L30)):*

`Addr()` and `Port()` are never used outside this package, so they can be unexported.
</blockquote></details>

I think they need to be Public. addrport is an embedded struct, so these methods implement part of the AppAddr and OverlayAddr interface

sgmonroy:
___
*[go/lib/addr/addrport.go, line 30 at r1](https://reviewable.io/reviews/scionproto/scion/1724#-LIf6SGA6INH73XiUcJS:-LJDmlfU0oqC-c9QfRNx:b-7wfaz3) ([raw file](https://github.com/scionproto/scion/blob/34b3dd8081e24fb8def73f4494aa39a5192b07e1/go/lib/addr/addrport.go#L30)):*
<details><summary><i>Previously, sgmonroy wrote…</i></summary><blockquote>

I think they need to be Public. addrport is an embedded struct, so these methods implement part of the AppAddr and OverlayAddr interface
</blockquote></details>

But this structure has been removed.

sgmonroy:
___
*[go/lib/addr/l4info.go, line 38 at r2](https://reviewable.io/reviews/scionproto/scion/1724#-LJIfnKs2dg4lGwxS7TK:-LJNQ15jD4_WPtLAJQyL:bujfgko) ([raw file](https://github.com/scionproto/scion/blob/49684ab9624d5f612ef82c0a5ad60f421ed26cc0/go/lib/addr/l4info.go#L38)):*

This should error if a port is specified when it doesn't make sense. E.g. The proto type is SCMP.
</blockquote></details>

What about having constructors for each type?
NewL4UDPInfo(port)
NewL4SCMPInfo()
...

sgmonroy:
___
*[go/lib/addr/overlayaddr.go, line 87 at r2](https://reviewable.io/reviews/scionproto/scion/1724#-LJIhEIB--gG8PsbCujh:-LJNSHTJ2xYY0OmlaQqR:b-ak5pmc) ([raw file](https://github.com/scionproto/scion/blob/49684ab9624d5f612ef82c0a5ad60f421ed26cc0/go/lib/addr/overlayaddr.go#L87)):*

Hmm. I think we probably want different representations depending on whether there's an l4 or not. E.g.:

```
if a.l4 != nil { 
   return fmt.Sprintf("[%v]:%s", a.l3, a.l4)
}
return fmt.Sprintf("%v", a.l3)
```
</blockquote></details>

shouldn't that use %s for l3? it shoudl never be nil

PR #1728, opened 2018-07-30
Topology Generator enhancements

### LINKED ISSUE #830: More comprehensive topology
The existing topology isn't comprehensive enough
- all core ASes only have a single child
- only single links between any pair of ASes
- no peer links between leaf ASes

We need to make a list of all scenarios we want to test, and make a minimal topology to cover those.

sgmonroy:
Define a more comprehensive default topology:
- border routers with multiple interfaces
- multiple links between the same two ASes
- core AS with more than single child
- core AS without child
- peer link between leaf ASes

Changes to the topology generator required support for:
- border routers with multiple interfaces
- multiple links between the same two ASes

We can now specify an ID on the BR name in the links table:
 - {a: "1-ff00:0:110",   b: "1-ff00:0:120-1", linkAtoB: CORE}
 - {a: "1-ff00:0:120-1", b: "1-ff00:0:130", linkAtoB: CORE}

If no ID is specified, a BR with a dingle interface would be created
(old behaviour).
When an ID is specified, the link would be associated to the same BR.

In the example above, two links are defined resulting in:
 - BR 1-ff00:0:110 with single interface
 - BR 1-ff00:0:120 with multiple interfaces
 - BR 1-ff00:0:130 with single interface

Fixes #830

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1728)


sgmonroy:
___
*[topology/Default.topo, line 64 at r5](https://reviewable.io/reviews/scionproto/scion/1728#-LK1PflcB3Vesyf3-4SK:-LK1xhn2CvBqF4nIJsVE:b-6h5fzh) ([raw file](https://github.com/scionproto/scion/blob/e6eaa4f3be15a832303dea8f1d82aeb237bcf3f7/topology/Default.topo#L64)):*

Huh. You dropped the link from `110` to `111`. That breaks the numbering convention (X0 is core, X1 is the core's direct child, X2 is a grandchild, etc).

Another issue is that the only example of dual peer links has both interfaces on the same BR, on both sides.
</blockquote></details>

Right, so that As has to up links, which core as should follow? I left it as is because 110 is still its cert issuer
I can break up BR in 111 to have peer links to 211 on different BRs

sgmonroy:
___
*[topology/Default.topo, line 64 at r5](https://reviewable.io/reviews/scionproto/scion/1728#-LK1PflcB3Vesyf3-4SK:-LK23g5DCmCIw7Me6MGL:b-vl3igs) ([raw file](https://github.com/scionproto/scion/blob/e6eaa4f3be15a832303dea8f1d82aeb237bcf3f7/topology/Default.topo#L64)):*
>>Right, so that As has to up links, which core as should follow? I left it as is because 110 is still its cert issuer

>Ah, i see - having a core AS with no children is a useful case. The numbering should follow the issuer's range, and the issuer should be a parent of the child (so the issuer for 111 should be either 120 or 130, your choice).

So is this 'should' or 'must' ? Where does this requirement come from?

>>I can break up BR in 111 to have peer links to 211 on different BRs

>That sounds good, yeah, it'll just mean one more BR to stop during revocation testing i assume.

We could but we don't have to stop it, do we?

sgmonroy:
___
*[topology/Default.topo, line 64 at r5](https://reviewable.io/reviews/scionproto/scion/1728#-LK1PflcB3Vesyf3-4SK:-LK5grYq5exQyUXO_DGX:bfulmsm) ([raw file](https://github.com/scionproto/scion/blob/e6eaa4f3be15a832303dea8f1d82aeb237bcf3f7/topology/Default.topo#L64)):*

>>Right, so that As has to up links, which core as should follow? I left it as is because 110 is still its cert issuer

>Ah, i see - having a core AS with no children is a useful case. The numbering should follow the issuer's range, and the issuer should be a parent of the child (so the issuer for 111 should be either 120 or 130, your choice).

So is this 'should' or 'must' ? Where does this requirement come from?

>>I can break up BR in 111 to have peer links to 211 on different BRs

>That sounds good, yeah, it'll just mean one more BR to stop during revocation testing i assume.

We could but we don't have to stop it, do we?
</blockquote></details>

After chat in slack regarding this, I am actually against the asked changes.
If it is not a hard requirement for the cert issuer to be a parent core AS, then this is a valid use case and it has its place in a **test** topology.

PR #1757, opened 2018-08-08
PS: improve segments returned

### LINKED ISSUE #1735: PS: Python PathDB prunes too much
There can be situations, especially with core segments, where path resolution in the PS fails, due to the PathDB limiting the answer to at most `k` segments (currently `k == 5`). Due to limiting the number of returned path segments based purely on length of the segments it can happen that no valid combination from `src` to `dst` can be made even though there would be segments in the PathDB that together would form valid paths from `src` to `dst`. To fix that, three things need to be done:

1) The PathDB itself shouldn't limit the number of segments it returns. In Python this means adding a `full=True` parameter to the query.
2) When querying the PathDB not only the destination IA should be taken into account, but also only the possible starting IAs should be included in the query.
3) The final set of path segments must be pruned to an acceptable limit. This has to be done in a way that all segments included in the final path reply can be used to form valid end2end paths from `src` to `dst`.

sgmonroy:
Fixes #1735

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1757)

### DIFF
--- a/python/path_server/local.py
+++ b/python/path_server/local.py
@@ -106,16 +106,17 @@ class LocalPathServer(PathServer):
             self._request_paths_from_core(req, logger)
         return False
 
-    def _segs_from_buckets(self, buckets, total_segs):
+    def _get_segs_from_buckets(self, buckets, total_segs):
         """
-        Returns a set of reachable core ASes from all available segments.
+        Returns up to MAX_SEG_NO segments from all available segments in the buckets in
+        Round-Robin fashion.
         """
         segs = []
         ias = set()
         while total_segs > 0:
             for core_ia in buckets:
                 if len(segs) == self.MAX_SEG_NO:
-                    return segs
+                    return segs, ias
                 if len(buckets[core_ia]) > 0:
                     segs.append(buckets[core_ia].pop(0))
                     total_segs -= 1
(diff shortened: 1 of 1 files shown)


sgmonroy:
___
*[python/path_server/local.py, line 114 at r1](https://reviewable.io/reviews/scionproto/scion/1757#-LJPMzJx9rTQAZqJUMV_:-LJSWzVN4gh4DbVzBpEb:bu8jx46) ([raw file](https://github.com/scionproto/scion/blob/dd0da833b1c2e663d99e88623f1ca2b2310458aa/python/path_server/local.py#L114)):*

call this `ia_pairs`
</blockquote></details>

not sure ia_pairs is the right name for this ... it is just a set of IAs

sgmonroy:
___
*[python/path_server/local.py, line 116 at r1](https://reviewable.io/reviews/scionproto/scion/1757#-LJPMgHa1cj0uDCMBfIf:-LJSXIwg01d75rJS7C4t:b-clvevo) ([raw file](https://github.com/scionproto/scion/blob/dd0da833b1c2e663d99e88623f1ca2b2310458aa/python/path_server/local.py#L116)):*

`core_ia` is actually a tuple of IAs. It would make it much clearer if you wrote this as `for (first_ia, last_ia) in buckets`.
</blockquote></details>

That is the case when we take from buckets_core as its key is a tuple (first_ia, last_ia). But not the case when used with buckets_up or buckets_down.

sgmonroy:
___
*[python/path_server/local.py, line 118 at r2](https://reviewable.io/reviews/scionproto/scion/1757#-LJnDxoU2yflg8QTerW_:-LJwoDJQ9QKwoUqdAMSV:b-2idsa9) ([raw file](https://github.com/scionproto/scion/blob/0deab1014c3815ad91df8175c83cf365fef71bf0/python/path_server/local.py#L118)):*

Done.
</blockquote></details>

After testing the change, we need to leave the while condition as originally set, as we could potentially get an empty bucket with no segments and we would never get to check the inner condition.
I would still do the propose if change so we can leave asap from the function.

PR #1758, opened 2018-08-08
trustdb: Guard db operations with lock

### LINKED ISSUE #1750: trustdb: go-sqlite3 is not safe for multipler writers
As mentioned in the [go-sqlite3 FAQ](https://github.com/mattn/go-sqlite3#faq), it's not safe to have multiple writers on a go-sqlite3 database. As such, `go/lib/infra/modules/trust/trustdb` needs to have locking to prevent this from happening.

lukedirtwalker:
go-sqlite3 is not save for multiple writers. Fixes #1750

### DIFF
--- a/go/lib/infra/modules/trust/trustdb/db.go
+++ b/go/lib/infra/modules/trust/trustdb/db.go
@@ -1,4 +1,4 @@
-// Copyright 2018 ETH Zurich
+// Copyright 2018 ETH Zurich, Anapaya Systems
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@ package trustdb
 import (
 	"context"
 	"database/sql"
+	"sync"
 
 	_ "github.com/mattn/go-sqlite3"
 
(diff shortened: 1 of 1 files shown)


scrye:
___
*[go/lib/infra/modules/trust/trustdb/db.go, line 222 at r2](https://reviewable.io/reviews/scionproto/scion/1758#-LJrZ-pK-TmcRvhGSrm4:-LJsachv8CLASDPmh66V:bnvoazi) ([raw file](https://github.com/scionproto/scion/blob/fd6862fdeb8c3f64da9167b3c3fb59a2c6ff1ae5/go/lib/infra/modules/trust/trustdb/db.go#L222)):*

Done.
</blockquote></details>

I think this should stay outside the mutex for cleaner tear-down. According to documentation:
```
Close returns the connection to the connection pool. All operations after a Close will return with ErrConnDone. Close is safe to call concurrently with other operations and will block until all other operations finish. It may be useful to first cancel any used context and then call close directly after. 
```

We should strive for similar semantics in our database packages.

(Also, marking the db for closing internally ASAP---as opposed to waiting on a very coarse lock---prevents other writers from acquiring internal locks, thus making the database shutdown faster and cleaner).

PR #1760, opened 2018-08-10
testing: Use in memory DB

lukedirtwalker:
Instead of always creating a temp file, which has significant overhead.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1760)

### DIFF
--- a/go/lib/infra/modules/trust/trust_test.go
+++ b/go/lib/infra/modules/trust/trust_test.go
@@ -1,4 +1,4 @@
-// Copyright 2018 ETH Zurich
+// Copyright 2018 ETH Zurich, Anapaya Systems
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


scrye:
___
*[go/lib/infra/modules/trust/trust_test.go, line 769 at r1](https://reviewable.io/reviews/scionproto/scion/1760#-LJwY81vEs7QiJa9N9x4:-LJwY81w3BHUAuvpMu5F:b66nk0h) ([raw file](https://github.com/scionproto/scion/blob/95a3e53a8edd967535164c8853dd19f54de37dbb/go/lib/infra/modules/trust/trust_test.go#L769)):*
> ```Go
> 	}
> 	store, err := NewStore(db, ia, 0, options, log.Root())
> 	if err != nil {
> ```

Nit: maybe change to `xtest.FailOnErr` so it's similar to the above?

PR #1799, opened 2018-08-23
Fix overlay port handling for BR in topology.

kormat:
It is (currently) illegal to specify `OverlayPort` in a BR entry, as the
BR doesn't run on top of the dispatcher. Additionally, calling
`OverlayAddr` on the internal address of a BR should set the overlay
port appropriately (assuming a UDP-based overlay).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1799)

### DIFF
--- a/go/lib/topology/testdata/udpbr.json
+++ b/go/lib/topology/testdata/udpbr.json
@@ -1,0 +1,28 @@
+{
+    "ISD_AS": "6-ff00:0:362",
+    "Overlay": "UDP/IPv4+6",
+    "BorderRouters": {
+        "borderrouter6-ff00:0:362-1": {
+            "InternalAddr": {
+                  "Public": [
+                    {"Addr": "192.0.128.1", "L4Port": 30097},
+                    {"Addr": "2001:db8:a0b:12f0::1", "L4Port": 30098}
+                  ],
+                  "Bind": [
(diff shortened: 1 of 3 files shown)


sgmonroy:
___
*[go/lib/topology/topology.go, line 120 at r1](https://reviewable.io/reviews/scionproto/scion/1799#-LKfkMr7D2Ef_gj-Ge7m:-LKfr4nJDjhyrR8Ee6P_:besx2fn) ([raw file](https://github.com/scionproto/scion/blob/34488c0cef3d01c2e6d8bbf43978b10111580dc6/go/lib/topology/topology.go#L120)):*

Doing this check after ToTopoAddr wouldn't catch the case where someone has set the `OverlayPort` in the config. If they have set it to anything which is _not_ the `L4Port`, then that will get silently overwritten, which is bad. As it's not supported to set `OverlayPort` for the (current) BR, this block makes sure the config is legal, and then sets it up so that ToTopoAddr will do the right thing.

If we change it from RawAddrInfo, we can't use `ToTopoAddr` any more, and the logic would have to be rewritten as part of #1797 anyway. `localTopoAddr` is a method on `RawBRIntf`, so i'm not seeing how it's related.
</blockquote></details>

Ok, the logic is going to be rewritten by #1797 anyway.

PR #1806, opened 2018-08-28
snet: fix panic in Listen

### LINKED ISSUE #1803: SIG Panic
The SIG panics in [snet](https://github.com/scionproto/scion/blob/master/go/lib/snet/snet.go#L198) because `laddr.Host.L4` is `nil` in some cases. `ListenSCIONWithBindSVC` is called twice and only in the second case it panics.

sgmonroy:
Add more checks to avoid the panix and also provide a default value if
the address has no L4 information.
In such case, the default value would be UDP port 0, meaning that the
application wants a random port.

Fixes #1803

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1806)

### DIFF
--- a/go/lib/snet/snet.go
+++ b/go/lib/snet/snet.go
@@ -193,9 +193,6 @@ func (n *Network) ListenSCION(network string, laddr *Addr, timeout time.Duration
 // A negative timeout means infinite timeout.
 func (n *Network) ListenSCIONWithBindSVC(network string, laddr, baddr *Addr,
 	svc addr.HostSVC, timeout time.Duration) (*Conn, error) {
-	if network != "udp4" {
-		return nil, common.NewBasicError("Network not implemented", nil, "net", network)
-	}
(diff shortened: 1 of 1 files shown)


sgmonroy:
___
*[go/lib/snet/snet.go, line 206 at r1](https://reviewable.io/reviews/scionproto/scion/1806#-LKzSN7Y8kNEzRLVH3ib:-LKzYLZpB-aVZdZjSAcT:b-441xmb) ([raw file](https://github.com/scionproto/scion/blob/ad75d23d8138989343df5a0afdb0f0859a317bd4/go/lib/snet/snet.go#L206)):*

I'd prefer to have a switch statement here based on the `network` param, so we avoid further hard-coding of the "udp4" assumption.
</blockquote></details>

So basically, do the switch now, with default unsupported and add later case statements when we support other L4 protocols versus leave it as is then implement the switch when we support other L4 protocols.
Either way changes would be required, why not leave it like this?

PR #1829, opened 2018-09-03
Refactor TopoAddr from RAI and related tests

sgmonroy:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1829)


sgmonroy:
___
*[go/lib/topology/addr.go, line 78 at r1](https://reviewable.io/reviews/scionproto/scion/1829#-LLUdB1QC6drX0-FPRTT:-LLUkqZ807a4iKqqmhvi:brf4scl) ([raw file](https://github.com/scionproto/scion/blob/f49d8804aea47251bdc853dc726bc4cd4aa707b4/go/lib/topology/addr.go#L78)):*

As the "both are == `nil`" case is already taken care of above, this can be:
```
if t.IPv4 != nil && t.IPv6 != nil {
```
which is a bit easier to follow, for the context.

That said, i would move these checks back into a validation method. It's much easier to unit test that way, and none of the checks seem to need to be in between public & bind parsing.

There should also be a validation step to ensure that the addresses parsed are compatible with the defined overlay (at least i can't find such a check).
</blockquote></details>

I don't see why having those checks in a different method would change anything for the unit test. The unit test should not care about it.

They don't have to be between pub and bind, but might as well do them there as soon as possible. Why wouldn't you do them there?

So for IPv46, do you need both or just one them set?

sgmonroy:
___
*[go/lib/topology/addr_test.go, line 85 at r1](https://reviewable.io/reviews/scionproto/scion/1829#-LLUh2oV0OyKBra-rt9L:-LLUpoR7BZ4CRri8Lmes:bsqzmsv) ([raw file](https://github.com/scionproto/scion/blob/f49d8804aea47251bdc853dc726bc4cd4aa707b4/go/lib/topology/addr_test.go#L85)):*

Missing cases: `IPv46` v4Pub`, `IPv46 v4PubBind`, `IPv46 v6Pub`, `IPv46 v6PubBind`, and same again for IPv46+UDP.
</blockquote></details>

Why do we need those here in the basic test? Those look like it should be error tests once we have a check for proper overlay type and addresses.

sgmonroy:
___
*[go/lib/topology/testdata/basic.json, line 17 at r3](https://reviewable.io/reviews/scionproto/scion/1829#-LLibE5y5rUvqRjFaqTw:-LLidetz8qtd9TQh8-Ju:b-9oxehd) ([raw file](https://github.com/scionproto/scion/blob/9db43308775f59e396d53554ead29dcee305177e/go/lib/topology/testdata/basic.json#L17)):*

Hmm. I was expecting this to be changed to follow the pattern we're using for the other addresses.

E.g.:
```
"Addrs": {
    "IPv4": {
       "Public": {"Addr": 192.0.2.1, "L4Port": 44997},
       "Bind": {"Addr": "10.0.0.1", "L4Port": 30090},
    }
}
"RemoteAddr": {
      "IPv4": {"Addr": "192.0.2.2", "L4Port": 44998},
 }
```
There would then be a check to ensure there's only a single entry in each map (and that they match each other).
</blockquote></details>

Why ? These fields have not been modified because they were not RawAddrInfo.
The overlay in the interface is going to dictate the type of address they should be, so I do not see any benefit.

sgmonroy:
___
*[go/lib/topology/raw.go, line 93 at r3](https://reviewable.io/reviews/scionproto/scion/1829#-LLincRjBf911vmZW9OR:-LLj7ywq0VLOQQHJkv8c:b-cmtxq7) ([raw file](https://github.com/scionproto/scion/blob/9db43308775f59e396d53554ead29dcee305177e/go/lib/topology/raw.go#L93)):*

The key would usually be the host addr type, right? In that case what about using the overlay type string instead?
</blockquote></details>

So in json that would look like:
```
"UDP/IPv4": {"Public": ... }
```
And that would still be the key in an hypothetical overlay = IPv4 ?

sgmonroy:
___
*[go/lib/topology/topology.go, line 153 at r3](https://reviewable.io/reviews/scionproto/scion/1829#-LLivHZ8CUUBPBSrGWo6:-LLj8mRI3iJqAFZVHRJs:bxj7le1) ([raw file](https://github.com/scionproto/scion/blob/9db43308775f59e396d53554ead29dcee305177e/go/lib/topology/topology.go#L153)):*

Don't modify loop vars. As it happens in this case you are probably safe, but only because it's a map to pointers, and not to values, but if that ever changed in the future this code would silently break. That's why the existing code takes an explicit reference.
</blockquote></details>

mmm this is the whole reason why I changed to a map of pointers... 
IMO map of values should be avoid unless there is a good reason as they are harder to work with them if you need to update values.

sgmonroy:
___
*[go/lib/topology/addr_test.go, line 235 at r6](https://reviewable.io/reviews/scionproto/scion/1829#-LLo0ieC5UZl5uqLpztY:-LLo73BE24Pl8JrpyXM9:b-yoy1z4) ([raw file](https://github.com/scionproto/scion/blob/335fb8e83bc18e2ce3b6ab4c7ea652cf9343efcc/go/lib/topology/addr_test.go#L235)):*

If you're using `GetErrorMsg`, then you don't need to use `GetNestedError` - `GetErrorMsg` already does that for you. So, i think this simplifies to:
```
if common.GetErrorMsg(actual.(error)) == expected[0].(string) {
   return ""
}
return fmt.Sprintf("Expected error: %+q\nActual: %+q", expected, actual)
```
</blockquote></details>

I don't think so. GetErrorMsg will return the first ErrorMsger it finds.
I believe our code return nested ErrorMsger errors so I need to manually inspect them.

PR #1833, opened 2018-09-04
Implement reconnecting proxy on top of snet

scrye:
Implements two new types, `ProxyConn`, a wrapper around `snet.Conn` that supports transparent reconnects, and `ProxyNetwork`, a wrapper around `snet.Network` that provides some helper functions for initializing `ProxyConn`s (e.g., creating callbacks and checking that addresses stay the same). Callers can opt to use the helper `ProxyNetwork`, or manage reconnection logic themselves by using `ProxyConn` directly.

Known issues:
  - not integrated into `snet` yet, because some types need to be converted to interfaces first and that can go in a different PR;
  - callbacks for dispatcher up and dispatcher down events are not implemented yet, but they can be easily added to `ProxyConn`'s constructor once applications need them.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1833)


scrye:
___
*[go/lib/snet/snetproxy/conn.go, line 268 at r1](https://reviewable.io/reviews/scionproto/scion/1833#-LLZfoMF4ww7v_8_it4l:-LLdtpph1S4GrHTh9zDk:b-xw6ibi) ([raw file](https://github.com/scionproto/scion/blob/89c4b84b6c439ed46bfb7f7387aff6aacc1c99db/go/lib/snet/snetproxy/conn.go#L268)):*

It looks like if the dispatcher connection has died, calling `Close()` will return an error. From a test program:
```
close unixgram /tmp/tmp.sock: use of closed network connection
```
I'm not sure it makes sense to return this to the caller. Thoughts?
</blockquote></details>

I think returning it isn't exactly useful, but it probably doesn't hurt.

Callers won't be interested in recovering from a `Close()` error (there's not much to recover from), but the extra information feels a bit more useful for logging compared to just masking the error (it will suggest, for example, that a write that was blocked on reconnecting didn't go through; of course, the write will also return an error).

I'm pretty much fine either way though.

scrye:
___
*[go/lib/xtest/helpers.go, line 175 at r5](https://reviewable.io/reviews/scionproto/scion/1833#-LLj0BYd2r9FeM7UeunX:-LLj5y3s7W9oD_yrRfK8:bzgpdhx) ([raw file](https://github.com/scionproto/scion/blob/9839e836272ccec99e61445f20ee8154a44fffbe/go/lib/xtest/helpers.go#L175)):*

Hmm, what is causing the confusion? They are ordered chronologically. Renamed them to x and y because before and after were maybe causing the confusion.

Anyway, ASCII art time! It's something like:
```
o - denotes when t.Fatalf would get called on a read

|-------------|-----------|--------------
|             X           Y
|ooooooooooooo                                     AssertReadDoesNotReturnBefore (so, after X)
|                         ooooooooooooooo          AssertReadReturnsBefore (so, before Y)
|ooooooooooooo            ooooooooooooooo          AssertReadReturnsBetween (so, between X and Y)
```

Orthogonally, this was buggy, and probably needs some tests but I'll add them in a separate PR.
</blockquote></details>

The bug is referring to the `y-x` change, btw. Should be fixed now.

PR #1863, opened 2018-09-13
PS: Remove unnecessary db access

lukedirtwalker:
When we fetch segments and they are verified we can use the verified segments
directly no need to get them from the DB.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1863)


lukedirtwalker:
As discussed in the Review this PR doesn't make sense.

PR #1885, opened 2018-09-20
Try to ensure that addr.HostIPv4 is 4 bytes.

kormat:
- Add a new constructor, addr.HostFromIPStr, which takes a string, and
  return a addr.HostIPv4 or addr.HostIPv6 as appropriate.
- Convert unit tests to use/expect 4-byte IPv4 addresses.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1885)

### DIFF
--- a/go/lib/addr/host.go
+++ b/go/lib/addr/host.go
@@ -123,7 +123,7 @@ func (h HostIPv4) Type() HostAddrType {
 }
 
 func (h HostIPv4) Pack() common.RawBytes {
-	return common.RawBytes(net.IP(h).To4())
+	return common.RawBytes(h.IP())
 }
 
 func (h HostIPv4) IP() net.IP {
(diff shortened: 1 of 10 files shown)


scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/1885#-LMrXFCbArc_Kl9Plkwm:-LMrXFCbArc_Kl9Plkwn:bsk7xmu) (no related file):*
Looks good. If you haven't already, please make sure to run the python end to end integration test with Go SCIOND:

```
./scion.sh topology -sd=go
./scion.sh run
python/integration/end2end_test.py
```
because it's the only test we have right now were Python client code interacts with Go SCIOND code via the sciond capnp schema (and it was the only place that was exploding due to python not understanding the 16-byte representation).

---

PR #1906, opened 2018-09-26
Simplify some network errors

scrye:
The simplified errors are more similar to network errors returned by the standard library, and easier to act on in calling code.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1906)

### DIFF
--- a/go/lib/snet/snet.go
+++ b/go/lib/snet/snet.go
@@ -275,7 +275,7 @@ func (n *SCIONNetwork) ListenSCIONWithBindSVC(network string, laddr, baddr *Addr
 	rconn, port, err := reliable.RegisterTimeout(conn.scionNet.dispatcherPath,
 		conn.laddr.IA, conn.laddr.Host, bindAddr, svc, timeout)
 	if err != nil {
-		return nil, common.NewBasicError("Unable to register with dispatcher", err)
+		return nil, err
 	}
 	if port != conn.laddr.Host.L4.Port() {
 		// Update port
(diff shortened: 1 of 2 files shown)


scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/1906#-LNPg-VfEwmv82jPhk3s:-LNo1yJrC72ZtwTbVFI_:b-g3lub2) (no related file):*

Ok, that's exactly what i was concerned about. A client should be able to distinguish errors in communicating with the dispatcher from other types of error. The plan is to move the network stack into the dispatcher as much as possible, so a bare write error could indicate there was a problem with the connection to the dispatcher, _or_ there was a problem writing to the network. Right now snetproxy assumes that an EPIPE error is a dispatcher failure, but in the future it could be because the remote end has closed its connection, etc. The existing error-nesting approach allows a client to both know that it was a dispatcher-communication error, and what the underlying syscall error was. (Granted, it could do with being cleaned up a bit, and made more consistent.)
</blockquote></details>

I agree that distinguishing between errors is important, but due to the very large number of possible errors (and the massive redesign/changes coming in the near future), I think it's premature to classify errors at this point. The simple approach should suffice for now.

I mean, we _could_ decorate every syscall error in `lib/sock/reliable` with `Dispatcher error` and check for that, but it won't give us anything at the moment. Calling code still needs to check for `Dispatcher error` (instead of `syscall.Error` right now), and still needs to go in the embedded syscall error to see if it's a dispatcher error that warrants a reconnect.

That's why I recommend going with the current approach: a few helper functions like the ones in `go/lib/snet/snetproxy/errors.go` that can be fed errors from `go/lib/sock/reliable.go` and yield just one outcome, the answer to "is this an error where I should reconnect or not?". Every application that needs reconnects should go through those functions.

When we move the stack, we can add any errors we want, and we just need to update the few helper functions.

---

PR #1927, opened 2018-10-05
BR: Acceptance tests

sgmonroy:
Setup and interfaces currently doing with bash.
gopacket pcap for packet capture/injection.

Roughly, this is how it works:
- Create SCION headers for pkt to sent and for expected packets
- Create overlay for pkt to be sent
- Send pkt
- Loop receiving packets checking that they match expected packets

To check that the packet matches:
- Compare relevant information in L3/IP and L4/UDP
- byte compare the SCION headers

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1927)


sgmonroy:
___
*[go/border/border-acceptance/acceptance-check.go, line 7 at r2](https://reviewable.io/reviews/scionproto/scion/1927#-LQ96PwS892PtLZjklNj:-LQ9FiXyEsu28lxIFGL_:boow0de) ([raw file](https://github.com/scionproto/scion/blob/4de99a071db2ec6f55070be8924226362417e46a/go/border/border-acceptance/acceptance-check.go#L7)):*

I'm hoping that the use of reflect is tied to the design of `BRTest`/`pktInfo`, and that changing those will remove the use of reflect. Is that accurate? :)

We should avoid reflection where possible:
https://blog.golang.org/laws-of-reflection
> Once you understand these laws reflection in Go becomes much easier to use, although it remains subtle. It's a powerful tool that should be used with care and avoided unless strictly necessary.
</blockquote></details>

I don't think that is accurate :)
I am importing reflect to be able to use reflect.Select(), basically dynamically build select case, given that the number of interfaces to use is variable. 
Sure I could have implemented a goroutine approach with multiple channels, but I think this is better for this case.

sgmonroy:
___
*[acceptance/ctl, line 25 at r8](https://reviewable.io/reviews/scionproto/scion/1927#-LRR69Sp6bq5vLgETRIz:-LRRopwE1NJF9dq3yBWR:b7q8pvg) ([raw file](https://github.com/scionproto/scion/blob/d9ee177129858bbad457b1fbc91e8ce119bc5a31/acceptance/ctl#L25)):*

Put `shift` after `COMMAND="$1"` instead, and just use "$@" here.
</blockquote></details>

I see I made a mistake that I corrected in each individual test later. this should be ${@:3}.
I also think current notation is more clear when looking at this file to see the position of each parameter, instead of $1 ; shift; $1; shift; $@

PR #1932, opened 2018-10-08
BR: Packet to local AS through remote AS

### LINKED ISSUE #1928: BR: Packet to local AS through remote AS
When running traceroute on the default topo from 1-121 to 1-122 we get the following path `1-ff00:0:121 48>72 1-ff00:0:120 72>48 1-ff00:0:121 94>24 1-ff00:0:122`. Where `1-ff00:0:121` to `1-ff00:0:120` is an up segment and `1-ff00:0:120` to `1-ff00:0:121` to `1-ff00:0:122` is a down segment. In this case traceroute does not get an answer from the border router in `1-ff00:0:121` (connected to `1-ff00:0:120`) because the BR creates a reply to its own AS with a path through `1-ff00:0:120`. 

In theory such a path is valid, the Go path combinator filters these paths but the python version does not. But the border router should still know how to handle such a packet.

worxli:
Fixes #1928

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1932)

### DIFF
--- a/go/border/rpkt/create.go
+++ b/go/border/rpkt/create.go
@@ -221,6 +221,20 @@ func (rp *RtrPkt) CreateReply(sp *spkt.ScnPkt) (*RtrPkt, error) {
 				return nil, err
 			}
 		}
+	} else {
+		// Reply should go to local AS but the path might include a remote AS
+		srcHost, err := rp.SrcHost()
+		if err != nil {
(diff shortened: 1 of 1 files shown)


sgmonroy:
Final agreement is to drop packets when we can detect such paths, see #1937.

PR #1938, opened 2018-10-08
CS:GO: Move to lib/env

### LINKED ISSUE #1835: CS: Port CSGO to use package env for configuration
Currently CSGO uses custom code for its configuration.

Porting it to use package `env` will mean CSGO uses a similar configuration to PS and SD, thus improving code reuse.

### LINKED ISSUE #1898: CS should reconnect to the dispatcher
The Go CS should use the same reconnection logic that the Go SD and PS use.

### LINKED ISSUE #1910: Go: trust store config IsCS field is never set
https://github.com/scionproto/scion/blob/master/go/lib/infra/modules/trust/config.go#L27:
```
type Config struct {
        ...
	// IsCS is set to true on CSes and false on others.
	IsCS bool
}
```
This field is _read_ by the trust store in one location, but never set anywhere.

Oncilla:
Make CS use lib/env for config.
Prepare config for reloading.
Refactor CS to follow same internal structure as PS and sciond

fixes #1835
fixes #1898
fixes #1910

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1938)


scrye:
___
*[go/cert_srv/main.go, line 77 at r2](https://reviewable.io/reviews/scionproto/scion/1938#-LOTLmKdCkNuBQNjurbM:-LOTcnai6lH98r-hv72K:b-6avgkm) ([raw file](https://github.com/scionproto/scion/blob/438edee2a0037224568fd4cd64992b9d0a949e0c/go/cert_srv/main.go#L77)):*

Sure. Will do.
</blockquote></details>

Agree with @lukedirtwalker, this would be really nice to have in the other services.

Oncilla:
___
*[go/cert_srv/main.go, line 121 at r2](https://reviewable.io/reviews/scionproto/scion/1938#-LOTYANE212VzMDe5JTF:-LOWtjdf97HPkbYkOGLb:b-lfucoe) ([raw file](https://github.com/scionproto/scion/blob/438edee2a0037224568fd4cd64992b9d0a949e0c/go/cert_srv/main.go#L121)):*

I noticed the commit text mentions this fixes #1910, but the reconnection function is not actually initialized. This is done via `InitInfraEnvironment` (which basically consists only of initializing the reconnecter).

You probably want something similar to how SD and PS do it:
```
environment = infraenv.InitInfraEnvironment(config.General.TopologyPath)
```

</blockquote></details>

Wait. Issue #1910 is about `IsCS bool` not being set. 

Do you mean #1898?
But that is about reconnecting to dispatcher. 
`infraenv.InitInfraEnvironment` only sets the reload function, no?

Oncilla:
___
*[go/cert_srv/main.go, line 72 at r7](https://reviewable.io/reviews/scionproto/scion/1938#-LPfSPz-89ghtyd1o7un:-LPfSPz-89ghtyd1o7uo:b-vbl4zn) ([raw file](https://github.com/scionproto/scion/blob/a9223ac7972757346f5b3f41ffdd8fc2109c92eb/go/cert_srv/main.go#L72)):*
> ```Go
> 		return 1
> 	}
> 	defer log.LogPanicAndExit()
> ```

Use `env.CleanupLog`

PR #1972, opened 2018-10-11
Fix go sciond socket/unixpacket file permissions

lukedirtwalker:
By default the use the folders permissions.
But with the dockerized topology this doesn't play well.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/1972)


lukedirtwalker:
This is not needed after #1975

lukedirtwalker:
This is not needed after #1975

PR #2052, opened 2018-10-29
Process SCMP errors caused by SCMP packets

sgmonroy:
There are a few issues when processing SCMP errors caused by SCMP packes,
like SCMP echo, by the border router and the dispatcher:
The border router fails to include the SCMP Info field, containing the
ID which is required to deliver the packet to the application.
The dispatcher does not process SCMP errors of SCMP packets.

This change adds the required support to the border router and the
dispatcher as well as modifying the SCMP tool to be able to deal with
SCMP revocations. It also adds a refresh flag to the scmp tool, which
now talks to sciond instead of the path manager.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2052)

### DIFF
--- a/c/dispatcher/dispatcher.c
+++ b/c/dispatcher/dispatcher.c
@@ -1016,7 +1016,7 @@ void send_scmp_reply(uint8_t *buf, SCMPL4Header *scmp, HostAddr *from, uint16_t 
     reverse_packet(buf);
     scmp->type = htons(type);
     zlog_debug(zc, "send SCMP %s to %s:%d",
-            scmp_ct_to_str(strbuf, ntohs(scmp->class_), ntohs(scmp->type)),
+            scmp_ct_to_str(strbuf, scmp->class_, scmp->type),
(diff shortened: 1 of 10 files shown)


sgmonroy:
___
*[go/tools/scmp/main.go, line 60 at r2](https://reviewable.io/reviews/scionproto/scion/2052#-LQd32S80m55ohftzmLz:-LQdqj7N7u2lwXBofnnJ:b9l2uy7) ([raw file](https://github.com/scionproto/scion/blob/84d63dd1ce2a10b493a532b85c59126d3c603446/go/tools/scmp/main.go#L60)):*

This is orthogonal to having sciond update paths or not. It's about letting sciond know about the revocation. I'm not questioning the use of refresh.
</blockquote></details>

well if scmp tells sciond about a revocation, it will purge the path. By not doing it, we still have the path (invalid, I know) as a choice and we can do refresh to purge it.

PR #2079, opened 2018-11-01
PS: Push cert chain to local CS

### LINKED ISSUE #2056: Go PS does not push crypto to local CS
Go SD currently queries the local PS for crypto; however, for scalability reasons, this should be changed to query the local CS.

The change breaks connectivity on the default topology, however, because the CS cannot get crypto for local ISD non-core ASes. The loop is as follows:
  - to verify a segment to a local ISD non-core AS, SD requests some chains from the CS
  - the CS sees that the request is for a remote AS
  - the CS wants to contact the AS to grab crypto
  - the CS asks SD for the path segment

To fix this, Go PSes should push crypto to local CSes when receiving Segment Registrations and Segment Sync messages.

lukedirtwalker:
Fixes #2056

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2079)

### DIFF
--- a/go/cert_srv/setup.go
+++ b/go/cert_srv/setup.go
@@ -34,6 +34,7 @@ import (
 	"github.com/scionproto/scion/go/lib/infra/modules/trust/trustdb"
 	"github.com/scionproto/scion/go/lib/log"
 	"github.com/scionproto/scion/go/lib/snet"
+	"github.com/scionproto/scion/go/proto"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


scrye:
___
*[go/lib/infra/modules/trust/trust.go, line 356 at r1](https://reviewable.io/reviews/scionproto/scion/2079#-LQF6ylj4nmQxNuR7ZRP:-LQF6yllEPzZ_yNBU5o2:bxe8rwl) ([raw file](https://github.com/scionproto/scion/blob/a594eec4678739c1bbf2fbf14fb2e5c3f294d766/go/lib/infra/modules/trust/trust.go#L356)):*
> ```Go
> }
> 
> func (store *Store) newChainValidatorForwarding(validator *trc.TRC) ValidateChainFunc {
> ```

Please add the comment specifying the limitation of this solution that we discussed offline (or reference an issue stating it).

Issue #2096, opened 2018-11-07
Idea: Attach Logger to Context


lukedirtwalker:
So we can create a logger per request attach it to the context and libraries can use the logger.
With a convenience method like `log.LoggerFromContext(ctx)` libraries get the logger from the context or the default logger if none is attached. This way all log statements for a handler can be matched via a unique ID.

PR #2103, opened 2018-11-07
Propagate loggers via contexts

scrye:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2103)


scrye:
Fixes #2096

PR #2131, opened 2018-11-15
SIG dies immediately when if disconnects from dispatcher

sustrik:
This is a intermediate measure. Eventually, we want to have
proper reconnect functionality but until then it's better
to kill the process instead of keeping it alive and malfunctioning.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2131)

### DIFF
--- a/go/lib/snet/snetproxy/conn.go
+++ b/go/lib/snet/snetproxy/conn.go
@@ -23,6 +23,7 @@ import (
 	"github.com/scionproto/scion/go/lib/common"
 	"github.com/scionproto/scion/go/lib/log"
 	"github.com/scionproto/scion/go/lib/snet"
+	"github.com/scionproto/scion/go/lib/sock/reliable"
 )
 
 var _ snet.Conn = (*ProxyConn)(nil)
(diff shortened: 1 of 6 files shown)


scrye:
___
*[go/sig/ingress/dispatcher.go, line 95 at r1](https://reviewable.io/reviews/scionproto/scion/2131#-LRRpzuW2b2ktpZGmBlU:-LRRv5Gd2X0xlPydaFvC:bm96omg) ([raw file](https://github.com/scionproto/scion/blob/a6c1a80ffbbf276062e3f4a6810aefe07eff489f/go/sig/ingress/dispatcher.go#L95)):*

This doesn't run in a goroutine, it runs directly from `main()`. A simple change would be:
- make `read()` have an `error` return value, if a dispatcher error happens, have `read()` return that.
- change Dispatcher.Run to return the value of `read()`
- change the log message in `main()` to be slightly more generic, e.g. "Ingress dispatcher error".
</blockquote></details>

The main goroutine has the same treatment as all other goroutines; there is a `defer log.LogPanicAndExit()` at application start (`go/sig/main.go:75`).

It's maybe simpler to have the same logic in the reader and writer and not make it overly complex, especially since it's a temporary solution until we have proper reconnection (which should be _soon_).

PR #2133, opened 2018-11-16
SIG uses SvcSIG address to locate remote SIGs

sustrik:
This patch leaves the old manually-configured SIGs in place
for failover purposes.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2133)

### DIFF
--- a/go/lib/addr/host.go
+++ b/go/lib/addr/host.go
@@ -184,7 +184,7 @@ var _ HostAddr = (*HostSVC)(nil)
 type HostSVC uint16
 
 // HostSVCFromString returns the SVC address corresponding to str. For anycast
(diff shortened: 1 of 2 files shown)


scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/2133#-LRS-bVO3ISeeuBH1s-f:-LRS-bVO3ISeeuBH1s-g:ba2ps2i) (no related file):*
Looking at the code, I think adding a remote SIG with the `SvcSIG` means every message towards the remote AS gets anycast/multicast to a random remote SIG each time, as the exact remote address (L3 + L4) from the remote is never saved.

---

PR #2135, opened 2018-11-16
Fix issue with run_command and set -e in acceptance/lib.sh

kormat:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2135)


sustrik:
LGTM

sustrik:
:lgtm:

PR #2154, opened 2018-11-22
PS: Initial prom metrics

### LINKED ISSUE #1819: PS: Export Prometheus metrics in the Go-PS
Some of the metrics of the Python PS do not make sense in the context of the GoPS. So as a first step we should determine which metrics are relevant and then add them in the code.

lukedirtwalker:
Fixes #1819

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2154)


lukedirtwalker:
We have metrics in the PS already now. This is no longer needed.

Issue #1390, opened 2018-02-01
SIG needs to handle local BR being down


kormat:
In the current code, if a local BR is unreachable, the SIG will timeout, but never receive a revocation. This implies that sciond will not get a revocation, so it will keep serving the same paths from its cache. If all paths for the SIG's destination happen to use the same BR, this situation will take up to 6 minutes to resolve (sciond caches paths for 5mins, pathmgr caches for another minute).

Until #1389 is done, the SIG needs to handle by telling sciond to flush its paths.

scrye:
Gateway now only uses paths it detects as healthy, so a local router being down is no longer an issue if other healthy paths out of the AS exist.

PR #2155, opened 2018-11-22
Dynamic peer SIG discovery

sustrik:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2155)


sustrik:
The tests here will be somewhat lacking because of #1390, but that's out of scope of this issue.

PR #2198, opened 2018-12-05
Remove crypto handlers from non-core PSes

lukedirtwalker:
Non-core PSes should never receive TRC or Chain requests.
PSes always request crypto material from the sender:
Core PS asks BS or other core PSes.
Non-core PS asks BS or core PS.

All other processes should go through sciond or the CS.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2198)


lukedirtwalker:
IceBoxed, this doesn't play well together with python sciond.

Oncilla:
PS does not exist anymore

PR #2250, opened 2018-12-14
Refactor reliable socket implementation

scrye:
Decoupled streaming component of reliable socket from message parsing.
This allows the server side of the connection to parse the messages, and
simplifies the transition to SEQPACKET sockets.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2250)


sustrik:
___
*[go/lib/sock/reliable/frame.go, line 57 at r1](https://reviewable.io/reviews/scionproto/scion/2250#-LTvJuUS39ObLi7kqeL9:-LTvJuUS39ObLi7kqeLA:bxya83t) ([raw file](https://github.com/scionproto/scion/blob/080b971fcd8e0fe2f1ed394d51cb01ad3f829014/go/lib/sock/reliable/frame.go#L57)):*
> ```Go
> }
> 
> // Frame describes the wire format of the reliable socket framing protocol.
> ```

Fro a non-insider this entire thing is very confusing. It's "reliable" socket. But what's that? There's nothing about that in the SCION book. In what sense is it reliable? Is it in an overlay protocol? A core SCION protocol? There's "Frame" but that's a pretty generic name that doesn't give any hints. Same thing with "cookie", "address" etc. I would suggest adding a short README to the folder explaining what's going on here.

scrye:
___
*[go/lib/sock/reliable/reliable.go, line 241 at r1](https://reviewable.io/reviews/scionproto/scion/2250#-LTx1TmbBad45ehGMhAq:-LU5NLwl49-CU_JlsiOn:blghv0l) ([raw file](https://github.com/scionproto/scion/blob/080b971fcd8e0fe2f1ed394d51cb01ad3f829014/go/lib/sock/reliable/reliable.go#L241)):*

Just a note: it is becoming clear that OverlayAddr really is UDPAddr, and in the medium term future an IP only overlay with IPAddr.
It begs the question of OverlayAddr usability. It is very unlikely that we would ever support anything else for endhosts.
</blockquote></details>

I agree. I've opened https://github.com/scionproto/scion/issues/2259 to track this discussion.

Issue #2254, opened 2018-12-17
BR: Make topology reload transactional


Oncilla:
When reloading the topology, either all changes shall be successful or no changes must be done at all.

Currently, the BR closes sockets before creating new ones, possibly leading to an unwanted state of the BR when the socket creation fails later on.

PR #2252 adopts this specific issue, but we should design proper transactional behavior of topology reloading.

sgmonroy:
This approach does not support reusing the same ip+port on a different interface on a single step, given that the old sockets are kept open until the new configuration succeeds .

To avoid such restriction, socket should be open using SO_REUSEADDR, thus being able to have multiple sockets open for the same ip+port.

Oncilla:
According to https://stackoverflow.com/a/14388707 `SO_REUSEPORT` is what we want.

Oncilla:
Setting `SO_REUSEPORT` must be done before binding the socket.
Support to do so has only been recently added to go (https://github.com/golang/go/commit/3c4d3bdd3b454ef45ce00559d705fe5dc6f57cad)

PR #2256, opened 2018-12-17
BR: Acceptance tests for port changing

Oncilla:
Add acceptance tests for BR topology reloading with port changes.
The test checks that changing the local and remote port of an
interface works as expected.

To change the remote port, a two step process is necessary.
This will be removed when #2254 is fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2256)


Oncilla:
included in more extensive PR #2258

PR #2257, opened 2018-12-18
WIP: Create debian packages for x86/amd64/aarch64/arm32

martin31821:
This PR is a **work-in-progress** effort to create debian packages for different architectures.

**UPDATE**: The changes here are done against the latest scionlab release version and fairly minimal. I removed the setuptools integration for the python part and tried to keep changes minimal to make it easier to review.

## Rationale

- On small devices like RPi etc. we might not want to use docker for deployment, but also don't want to compile scion directly on these devices.
- Therefore it's required to be able to cross-compile scion without messing with the current build scripts
- As a byproduct, packages for amd64 and x86 are also created, for users who might not want to use the docker containers.

## Implementation

- I removed the requirement for root-access during build-time (root access is still required to install packages into `/usr`, though)
- I had to fix some issues with the makefiles, as well as the x86/x64 constants
- I couldn't use the existing build process, since the go tooling doesn't allow `go install` on foreign architecture binaries (which makes sense, but requires manual interaction for us)
- I decided to build the go-code within a temporary directory (using `mktemp -d`) to ensure a clean working directory

## Left to be done

- [x] Rebasing to current scion state 
- [ ] Correct semantic versioning

## Building

Building packages is fairly easy, just run `debuild -us -uc -b -a$ARCH -tc` (where $ARCH = i386, amd64, armhf, arm64).
To be able to do a cross-build, you need to have the correct toolchain setup on your system.

## Questions/Discussions

1. I'd like to get some opinions on having debian packages along the existing docker containers
2. Which architectures to support?
  - x86 and amd64 are fully supported
  - armv6, armv7, armv8 are working with the correct toolchain
  - Others? (I think of MIPS, for some cheap plastic routers?)
3. When and how do we want to build such packages (CI, automated publishing to package sources)

**NEW**
4. How do we want to manage the configuration once scion is installed using packages?
I think about the zookeeper configuration or the dispatcher config file, etc. etc.
I don't have a good idea on how we want to handle this.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2257)


martin31821:
@kormat @juagargi 

I updated this PR against a much more recent version of scion, getting rid of the whole python packaging infrastructure [here](https://github.com/martin31821/scion/tree/scionlab-next-pkg)

scrye:
Is this still actively being worked on? Is there some blocking issue/question we can help out with?

martin31821:
@mkowalski and I have worked on integrating these changes into the ETHZ infrastructrue and we're building packages from there, so I'll close this PR, since it's also implemented against a very old version of scion.

There is indeed a point where we need some help from your side, we have problems when crosscompiling the scion code with bazel from amd64 (host) to arm32/64 (device), which yields cgo errors.

PR #2280, opened 2018-12-28
CS: Use core CS for lookup

### LINKED ISSUE #2226: CS: Change lookup strategy
Currently if we need a chain for a remote AS we always query directly the remote AS's CS. But to query the remote CS we first need a path, by getting the path we already get the required chain in the PS.
A simpler lookup strategy only includes core CSes, to those the paths will most likely be cached already.
The simpler lookup strategy is as follows:
* If dest is core or isd-local: Ask a (or multiple) local core CS
* Else: Ask remote core CS

lukedirtwalker:
Instead of going to the dst IA's CS for lookup we only go to
- the local core CS for ISD-local queries
- the remote core CS for remote-ISD queres

This should have better caching:
- Connectivity should be cached (beaconing)
- Trust material should also be cached because it is needed for verification of the segments.

Fixes #2226


lukedirtwalker:
This contains the code of #2283  and #2284  as well, so we should merge those first.

PR #2283, opened 2018-12-28
sciond: Handle ISD-local AS-wildcard path requests

### LINKED ISSUE #2282: sciond: ISD-local AS-wildcard path query is not handled correctly in non-core Go-sciond

lukedirtwalker:
Fixes #2282

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2283)

### DIFF
--- a/go/sciond/internal/fetcher/fetcher.go
+++ b/go/sciond/internal/fetcher/fetcher.go
@@ -109,7 +109,7 @@ func (f *fetcherHandler) GetPaths(ctx context.Context, req *sciond.PathReq,
 		return f.buildSCIONDReply(nil, 0, sciond.ErrorBadSrcIA),
 			common.NewBasicError("Bad source AS", nil, "ia", req.Src.IA())
 	}
-	// Commit to a path server, and use it for path and crypto queries
+	// Commit to a path server, and use it for path queries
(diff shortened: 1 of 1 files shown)


scrye:
___
*[go/sciond/internal/fetcher/fetcher.go, line 523 at r2](https://reviewable.io/reviews/scionproto/scion/2283#-LVOB36y6DuUrp2YExE0:-LVOCwde6d2uVbjMmmTg:b-2bw7tj) ([raw file](https://github.com/scionproto/scion/blob/e308a84fee21d5a3a5b75b13fba6d7c6d6b2886e/go/sciond/internal/fetcher/fetcher.go#L523)):*

Now that the code below is reasonably complex, it might be useful to make `dsts` be initialized only once (right now it's initialized above for the unindented code path, and re-initialized below with `make` inside the `if`, which is a bit confusing).

Or maybe something like:
```
var dsts map[addr.IA]struct{}
if !isWildcardDestination {
    dsts = chooseDestinations(listOfCores, isLocalISD)
} else {
    dsts = reg.Dst.IA()
}
```

Even some micro-functions might help with readability a bit.
</blockquote></details>

The branches above should be the other way around, btw.

PR #2300, opened 2019-01-07
Topo: Add diff

Oncilla:
Add support for calculating the diff between two topologies.
 
The diff is necessary for infra elements to check whether a topology update is valid under the constraints.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2300)


Oncilla:
For the purpose of checking topo updates, a general diff between two topologies is not necessary.
If a general diff is necessary in the future, we can always go back to this.

PR #2304, opened 2019-01-08
CS: handle concurrent insertion of cust keys in startup

lukedirtwalker:
It could be that a cust key is inserted exactly at the time a server starts and therefore the insert of the cust key will fail. In this case we should just retry.

This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2304)


lukedirtwalker:
We discussed this and came to the conclusion that insertion should be done in a separate tool. This is done in #2314

PR #2355, opened 2019-01-17
Truststore: add metrics for get functions

lukedirtwalker:
Currently metrics are implemented just for Get.. functions. For the handlers we still have to add the counters.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2355)


scrye:
With TrustStore v2 coming soon, we can skip implementing metrics for v1.

PR #2375, opened 2019-01-22
itopo: Split topo in static and dynamic

Oncilla:
Split itopo topology into a static and dynamic version.

Currently, only the static topology can be set and fetched.
The code also contains an outline how the dynamic topology
will be used in a future PR.

The services currently only support setting the static
topology through a sighup reload. Querying the discovery
service will be added in a future PR.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2375)

### DIFF
--- a/.buildkite/scripts/build_scion_img
+++ b/.buildkite/scripts/build_scion_img
@@ -2,7 +2,7 @@
 
 set -e
 
-BASE_IMG=${BASE_IMG:-aa88b1da1b510af630544ff6a8ff4c09111c595764efb025e3565b95759f043b}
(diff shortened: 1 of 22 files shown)


scrye:
___
*[go/lib/infra/modules/itopo/validate.go, line 38 at r2](https://reviewable.io/reviews/scionproto/scion/2375#-LWv8_VI1TMKoq1qZvaZ:-LWvgy3j97eKPIqvbXnW:b-iv3q7f) ([raw file](https://github.com/scionproto/scion/blob/6660b24e62e9a7299493c89f3315e96aa2ad8516/go/lib/infra/modules/itopo/validate.go#L38)):*

Tried to solve your concerns with `FTSE`.
</blockquote></details>

FTSE solves everything yet again!

PR #2379, opened 2019-01-23
infra.Messenger: expose UpdateSigner/UpdateVerifier

### LINKED ISSUE #2030: lib: infra.Messenger does not expose UpdateSigner/UpdateVerifier
The `infra.Messenger` interface does not expose  `UpdateSigner` and `UpdateVerifier`.

Resolving this is a bit more involved, since exposing them will create  an import cycle between `ctrl` and `infra`

When this is resolved, `cert_srv/setup.go` should avoid the type assertion.

lukedirtwalker:
Moves signer and verifier implementation from ctrl to the trust package.
Use infra.Messenger in CS instead of *messenger.Messenger.

Fixes #2030

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2379)

### DIFF
--- a/go/border/braccept/brA.go
+++ b/go/border/braccept/brA.go
@@ -19,8 +19,8 @@ import (
 
 	"github.com/scionproto/scion/go/border/braccept/tpkt"
 	"github.com/scionproto/scion/go/lib/common"
-	"github.com/scionproto/scion/go/lib/ctrl"
 	"github.com/scionproto/scion/go/lib/ctrl/path_mgmt"
+	"github.com/scionproto/scion/go/lib/infra"
 	"github.com/scionproto/scion/go/lib/l4"
 	"github.com/scionproto/scion/go/lib/scmp"
 )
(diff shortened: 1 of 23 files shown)


Oncilla:
___
*[go/lib/infra/modules/trust/signhelper.go, line 36 at r1](https://reviewable.io/reviews/scionproto/scion/2379#-LWvMpgJ3Spv0Vj48GWD:-LWvQebp6w-0STUhuoxa:b-evhtia) ([raw file](https://github.com/scionproto/scion/blob/0f09e0633947b849666c9548dca0400f50d9e8da/go/lib/infra/modules/trust/signhelper.go#L36)):*

As mentioned before. I don't think this is the right place for `NullSigner` and `NullSigVerifier`.

Also, `go/border/braccept/{brA, brD, core_brA}.go`still use `ctrl.NullSigner` and `ctrl.NullSigVerifier`
</blockquote></details>

whoops, those are auto-generated.
Ignore that part of the comment.

PR #2419, opened 2019-02-04
Wireshark: pktlen undefined

jderuiter:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2419)


kormat:
@jderuiter : thanks for the patch! (and sorry it took so long; it dropped through the cracks somehow, i only happened to find it today)

PR #2422, opened 2019-02-05
Expand extension support for Go

scrye:
Also:
  - unknown extensions are now parseable;
  - unit test coverage for extension parsing;
  - integration with gopacket for low-level parsing;
  - move scattered small extensions to their own package;

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2422)


scrye:
___
*[go/lib/layers/extensions.go, line 204 at r1](https://reviewable.io/reviews/scionproto/scion/2422#-LXxUrnwDxsOY1PDCwvz:-LXxUrnwDxsOY1PDCww-:bo643ts) ([raw file](https://github.com/scionproto/scion/blob/21e6840e7477882271a6aabb874b25c962340402/go/lib/layers/extensions.go#L204)):*
> ```Go
> 	ClassField common.L4ProtocolType
> 	TypeField  uint8
> 	// Length, in bytes, including 3-byte extension header
> ```

Note to self: _not_ including 3-byte extension header.

PR #2445, opened 2019-02-12
Filter duplicate paths in path combinator

FR4NK-W:
Combine should not return dulicate paths,
sciond does not filter separately.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2445)

### DIFF
--- a/go/lib/infra/modules/combinator/combinator.go
+++ b/go/lib/infra/modules/combinator/combinator.go
@@ -51,7 +51,31 @@ func Combine(src, dst addr.IA, ups, cores, downs []*seg.PathSegment) []*Path {
 	for _, path := range paths {
 		pathSlice = append(pathSlice, path.GetFwdPathMetadata())
 	}
-	return FilterLongPaths(pathSlice)
+	shortPaths := FilterLongPaths(pathSlice)
+
+	// filter duplicates
+	var uniquePathSlice []*Path
+DUP:
+	for _, p1 := range shortPaths {
+		for _, p2 := range uniquePathSlice {
+			if pathsEqual(p1, p2) {
+				continue DUP
+			}
+		}
+		uniquePathSlice = append(uniquePathSlice, p1)
+	}
+	return uniquePathSlice
+}
+
+func pathsEqual(p1, p2 *Path) bool {
+	same_path := true
+	if len(p1.Interfaces) != len(p2.Interfaces) {
+		return false
+	}
+	for i := range p1.Interfaces {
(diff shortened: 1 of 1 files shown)


FR4NK-W:
@scrye 
This impacts the ability of the SIG to fail over to a working path after a link failure.
The SIG tries repeatedly to use a path that is identical at the interfaces level as the failed one.
The SIG does not get a new path that is different in the interfaces used.

E.g. There are 4 paths from A to B that are identical at the interface level as the one in use and 1 with different interfaces.
A link fails on the path currently in use. The SIG tries to get a new path with getNewPath, different from the current one and will use a path with the with different HFs but with the same interfaces.
The failure count on the paths is not sufficient to recover from this state.

In the case shown below, where the ASes involved in the cryptographic verification are the same, is there still a reason for keeping multiple ways of representing/verifying the same forwarding path?
Would the pathmgr be the right place to handle this ambiguity between forwarding path and cryptographic verification of the path (so that not each application has to be explicitly aware / implement its own handling)?
E.g.:
                                                          B -> A
          C                                              Path #0:
      12| |13                                            Fields:
          | |                                                  IF CS. ISD=1
      21| |31                                               HF .V InIF=0 OutIF=12
          B                                                    HF .. InIF=21 OutIF=14
       14|                                                    HF .. InIF=41 OutIF=0
           |                                                Interfaces:
       41|                                                  B#14
          A                                                  A#41


                                                         
                                                         Path #1:
                                                           Fields:
                                                             IF CS. ISD=1
                                                               HF .V InIF=0 OutIF=13
                                                               HF .. InIF=31 OutIF=14
                                                               HF .. InIF=41 OutIF=0
                                                           Interfaces:
                                                             B#14
                                                             A#41

scrye:
After discussing this internally for a bit, we agree that SCIOND should be able to delete duplicates, but this should be configurable and not the default. This means the combinator is not the place to do the filtering (the combinator should always build everything), and it should be done instead on the `Combine` call site (where callers can freely process the path set).

It might be the case that for Anapaya deployments we will also enable this in the near future.

To merge this, the following changes are needed:
  - duplicate paths removal logic moved out of package `combinator`, and into `SCIOND` code;
  - duplicate paths removal logic should be hidden behind a feature flag (see https://github.com/scionproto/scion/pull/3030 for feature flag support);
  - the default flag setting (`false`) should be to not filter duplicates (so the default should be the current logic on `master)`;
  - the duplicate removal function should have unit tests.

scrye:
With https://github.com/scionproto/scion/pull/3930 (and due to the new data-plane path format), duplicate paths are now filtered. Closing this.

PR #2464, opened 2019-02-15
Control-Plane PKI documentation

lchuat:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2464)


lchuat:
I just realized the term “principal” is often used in PKI literature to mean “client” or “verifier”, so it makes the expression “principal AS” potentially confusing, so I will rename these ASes “main ASes” (unless you have another suggestion).

scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/2464#-LZ-wgqS2S_tGHif3v8T:-LZd0av-09RXcoBDCUJ2:b-u9u136) (no related file):*
<details><summary><i>Previously, lchuat (Laurent Chuat) wrote…</i></summary><blockquote>

- Regional numbering authorities would be the main trust-anchor providers. But software vendors could be considered trust-anchor providers as well, if they pre-install TRCs on the devices they sell or populate/modify trust stores through software updates (similarly to how self-signed root CAs are pre-installed and updated by OS/browser vendors for HTTPS). These vendors should obtain their trust anchors from the regional authorities though.
- One possible model would be the following: numbering authorities have their own ISD, and they are directly connected to the other ISDs. Then certificate servers should have the TRC of that ISD pre-installed, from the outset. But I suppose other models are possible.
- A new ISD would need to contact the regional numbering authority (aka trust-anchor provider) first, in order to obtain a unique ISD identifier; then generate and sign the fresh TRC with the attributed identifier; then submit the version 1 TRC to the regional authority and wait for it to be published before using it.
- Some TRCs should be pre-installed on some machines (in particular, the TRCs of the ISD(s) where regional authorities are located should be pre-installed on certificates servers).
</blockquote></details>

The topology suggested in the second bullet, "and they are directly connected to the other ISDs" seems very difficult to achieve in practice; it requires physical connections from the authority to everywhere else.

Re: the final bullet, it seems like those machines also need to preload the TRCs of all transit ISDs needed to reach the authority. For redundancy, paths through different transit ISDs _should_ be available, which means even more preloads. This depends on the shape of the SCION world graph. Is the number of preloads estimated to stay low?

---

kormat:
Closing in favour of #2831

PR #2465, opened 2019-02-20
log: always log to stderr

cschutijser:
The SCION log library logs to stdout *unless* an application calls `SetupLogConsole()` in which case it will log to stderr.  The problem is that `SetupLogConsole()` is not always called.  For instance, if an application calls `SetupFromFlags()`, `SetupLogConsole()` is only called when the `-log.console` flag is specified.  As a result, if the user does not specify the `-log.console` flag, the application will log to stdout.

I believe this is an oversight; you probably want to log to stderr always (and actually, I think that was the intention of #1698).  This can be achieved by calling `SetupLogConsole()` from `init()`.  Maybe there are better ways.

Example of the new behavior:

Before:
```
$ ./pingpong >/dev/null
$
```

After:
```
$ ./pingpong >/dev/null
2019-02-20 11:12:33.213727+0100 [CRIT] Missing remote address
$
```

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2465)

### DIFF
--- a/go/lib/log/log.go
+++ b/go/lib/log/log.go
@@ -34,6 +34,7 @@ import (
 
 func init() {
 	fmt15.TimeFmt = common.TimeFmt
+	SetupLogConsole(DefaultConsoleLevel)
 }
 
 var logBuf *syncBuf


cschutijser:
> > _[Reviewable](https://reviewable.io/reviews/scionproto/scion/2465)_ status: all files reviewed, 1 unresolved discussion (waiting on @cschutijser)
> 
> _[a discussion](https://reviewable.io/reviews/scionproto/scion/2465#-L_HPk1n2xVIOCVZ0-jg:-L_HPk1oAu86Pb42UG4y:bej15me) (no related file):_
> Wait, this is not desirable. It means that all applications will log to console even if they don't want to.

Good point. I will try to come up with a better approach.

scrye:
Any update on this? Is it still being worked on?

cschutijser:
> Any update on this? Is it still being worked on?

Hi, I don't know if the question was directed at me but I'm not working on this right now.

scrye:
Logging has been completely reworked following the move to [zap](https://github.com/uber-go/zap). We now only log to console, and it's docker's job to pipe the logs to the relevant place. This should also fix the issue mentioned in this PR.

PR #2588, opened 2019-04-10
 BS: Add keepalive handler

Oncilla:
This PR adds the keepalive handler.

Upon receiving a IFID keepalive, the state of the interface is changed
to active. If the interface was not active before, the handler
pushes an ifid state update to all border routers to remove revocations,
starts beaconing on the interface immediately and deletes possible
revocations associated with the interfaces link (i.e. the local and
remote revocations).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2588)


scrye:
___
*[go/beacon_srv/internal/keepalive/handler.go, line 65 at r1](https://reviewable.io/reviews/scionproto/scion/2588#-LcVc6-VB8XgVQkvQG8n:-LcVw2B37Yuj6dN0WjKk:b-etw15i) ([raw file](https://github.com/scionproto/scion/blob/93a723758f3be4075d69f23389a6106afbdfba45/go/beacon_srv/internal/keepalive/handler.go#L65)):*

Clarify in the docs what happens on a `nil` task.
</blockquote></details>

So, a `nil` `Beaconer`, for example, or a `nil` `StateChangeTasks` altogether.

PR #2617, opened 2019-04-25
Remove HostNone implementation; it was never used

scrye:
If we ever need it we can roll back the change. The integer ID of
TypeNone addresses was kept because it is used by reliable socket
registrations.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2617)


scrye:
The data plane formats are undergoing a lot of changes, so this can wait.

PR #2764, opened 2019-06-17
Refactor reconnecting logic as a dispatcher service

scrye:
This simplifies how reconnecting logic is added to connections inside
snet.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2764)


scrye:
___
*[go/lib/snet/snetproxy/conn.go, line 169 at r2](https://reviewable.io/reviews/scionproto/scion/2764#-LheyKM830RJMkwMJRFX:-LheyKM830RJMkwMJRFY:b-ai7dw4) ([raw file](https://github.com/scionproto/scion/blob/cc478def03581a9af301bf6f57584e7f8973714a/go/lib/snet/snetproxy/conn.go#L169)):*
> ```Go
> 		return nil, 0, err
> 	}
> 	_, oldPort := conn.getConnAndPort()
> ```

Note to self: refactor this into just `getPort()`.

PR #2772, opened 2019-06-19
CS reconnects to the dispatcher in default topo

scrye:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2772)


scrye:
This partially implements https://github.com/scionproto/scion/issues/2753.

PR #2831, opened 2019-07-03
Control-Plane PKI documentation

Oncilla:
This PR is based on PR #2464.

(Note: will edit commit message before merging)

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2831)

### DIFF
--- a/doc/ControlPlanePKI.md
+++ b/doc/ControlPlanePKI.md
@@ -126,7 +126,7 @@ revocation) applies. Trust anchors often take the form of self-signed root certi
 roots of trust). In the context of SCION the trust anchors are TRCs.
 
 The two predominant trust models in today's Internet are monopolistic (single root of trust) and
-oligopolistic (multiple roots of trust). Typically, in both models, all or some certification
+oligopolistic (multiple roots of trust). Typically, in both models, some or all certification
 authorities are omnipotent. That is, if their key is compromised, then the security of the entire
 system collapses. Moreover, roots of trust are typically defined through independent self-signed
 certificates. The SCION trust model is different in mainly two ways. First, no entity is omnipotent;
@@ -176,16 +176,16 @@ introduced in a future version of the TRC format.
 
 This comprises all non-object values in the top level of the TRC.
 
-- __ISD__: 16-bit integer. Unique and immutable ISD identifier.
-- __TRCVersion__: 64-bit integer. TRC version, starts at 1. All TRC updates must increment this by
+- __ISD__: 16-bit unsigned integer. Unique and immutable ISD identifier.
+- __Version__: 64-bit unsigned integer. TRC version, starts at 1. All TRC updates must increment this by
     exactly 1 (i.e., no gaps, no repeats).
-- __BaseVersion__: 64-bit integer. Version of the last trust reset TRC.
+- __BaseVersion__: 64-bit unsigned integer. Version of the last trust reset TRC.
 - __Description__: String. Describes the ISD/TRC in human-readable form (possibly in multiple
     languages).
-- __VotingQuorum__: 8-bit integer. Defines how many voting ASes from this ISD need to agree to be
+- __VotingQuorum__: 8-bit unsigned integer. Defines how many voting ASes from this ISD need to agree to be
     able to modify the TRC.
-- __FormatVersion__: 8-bit integer. Version of the TRC/certificate format (currently 1).
-- __GracePeriod__: 32-bit integer. How long, in seconds, the previous unexpired version of the TRC
+- __FormatVersion__: 8-bit unsigned integer. Version of the TRC/certificate format (currently 1).
+- __GracePeriod__: 32-bit unsigned integer. How long, in seconds, the previous unexpired version of the TRC
     should still be considered *active*, i.e., `TRC(i)` is still active until the following time has
     passed (or `TRC(i+2)` has been announced):
 
@@ -253,9 +253,9 @@ The following are conditions that must hold true for every TRC:
 6. Each `Authoritative` AS is a `Core` AS.
 7. No non-`Voting` AS has an online or offline key.
 8. No non-`Issuing` AS has an issuing key.
-9. `(BaseVersion != TRCVersion) ==> (GracePeriod > 0)`
-10. `(BaseVersion == TRCVersion) <==> (GracePeriod == 0)` (Initial TRC or trust reset)
-11. `(BaseVersion == TRCVersion) ==> All keys attach proof of possession to TRC`
+9. `(BaseVersion != Version) ==> (GracePeriod > 0)`
+10. `(BaseVersion == Version) <==> (GracePeriod == 0)` (Initial TRC or trust reset)
+11. `(BaseVersion == Version) ==> All keys attach proof of possession to TRC`
 
 
 ### Example of a TRC Payload
@@ -263,7 +263,7 @@ The following are conditions that must hold true for every TRC:
 ````json
 {
     "ISD": 1,
-    "TRCVersion": 23,
+    "Version": 23,
     "BaseVersion": 1,
     "Description": "Example ISD",
     "VotingQuorum": 2,
@@ -395,7 +395,7 @@ update) TRCs.
 
 - __Subject__: String. ISD and AS identifiers of the entity that owns the certificate and the
     corresponding key pair.
-- __TRCVersion__: 64-bit integer. Version of the TRC the issuer used when signing the certificate.
+- __Version__: 64-bit integer. Version of the TRC the issuer used when signing the certificate.
     Note that a certificate can still be valid and verifiable, if the issuing AS has the same public
     key in any of the active TRC versions. Thus, TRC updates that do not change the issuing key do
     not affect the validity of a certificate.
@@ -625,7 +625,7 @@ For any kind of update, the following conditions must be met:
 - The [TRC Invariants](#trc-invariants) must hold.
 - The `ISD` identifier field is immutable.
 - The `TrustResetAllowed` field is immutable.
-- The `TRCVersion` field must be equal to the previous version + 1.
+- The `Version` field must be equal to the previous version + 1.
 - The `BaseVersion` must be equal to the base version of the pervious TRC.
 - The `GracePeriod` of the TRC must not be 0. (A TRC with a grace period of 0 indicates a trust
   reset)
--- a/doc/ControlPlanePKI.md
+++ b/doc/ControlPlanePKI.md
@@ -57,6 +57,15 @@ Other qualifiers for TRCs include the following:
 - __Expired:__ a TRC whose validity period has ended, or that has been replaced by an update whose
   grace period renders the previous TRC inactive.
 
+### Types
+
+Through out this document, the terms __string__, __integer__, __timestamp__ must be interpreted as
+follows (unless specified otherwise):
+
+- __string__: UTF-8 string.
+- __integer__: All integers are unsigned.
+- __timestamp__: 64-bit integer indicating seconds since the unix epoch.
+
 ## Primary ASes
 
 An ISD is made up of a number of ASes. There is a set of attributes that each AS may have:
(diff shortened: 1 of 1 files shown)


Oncilla:
___
*[doc/ControlPlanePKI.md, line 3 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisA2ecDEsonUsySNSY:-LisA2ecDEsonUsySNSZ:buhyxu3) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L3)):*
> ```Markdown
> # SCION Control-Plane PKI
> 
> The control-plane PKI (CP-PKI) allows each isolation domain (ISD) to define its own roots of trust
> ```

@kormat :

I think defining it was being “for routing-related decisions” is too narrow. The CP-PKI is primarily about identity; routing decisions are based made on knowing the identity of the entities on a path.

Oncilla:
___
*[doc/ControlPlanePKI.md, line 3 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisA2ecDEsonUsySNSY:-LisA8Ju-ExCNzF2BVPY:b20gk6s) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L3)):*

@kormat :

I think defining it was being “for routing-related decisions” is too narrow. The CP-PKI is primarily about identity; routing decisions are based made on knowing the identity of the entities on a path.
</blockquote></details>

@Oncilla: 
This depends on how you define “routing-related” decisions.

Identity is also a too broad term IMO. I.e. the identity of end hosts is not authenticated by CP-PKI and it probably shouldn’t. Also TLS or DNS identity is not covered by CP-PKI. Thus, stating that CP-PKI is about identity without further qualifications is not accurate enough.

Maybe “routing identity” is more fitting.

Oncilla:
___
*[doc/ControlPlanePKI.md, line 3 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisA2ecDEsonUsySNSY:-LisADuLCFheNpQKZeVw:bsjer6p) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L3)):*

@Oncilla: 
This depends on how you define “routing-related” decisions.

Identity is also a too broad term IMO. I.e. the identity of end hosts is not authenticated by CP-PKI and it probably shouldn’t. Also TLS or DNS identity is not covered by CP-PKI. Thus, stating that CP-PKI is about identity without further qualifications is not accurate enough.

Maybe “routing identity” is more fitting.

</blockquote></details>

@kormat: 
Ok, let me be more precise then: it’s about the identity of ASes (and ISDs).

Oncilla:
___
*[doc/ControlPlanePKI.md, line 38 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisARFQ8ABBXhqP_cMV:-LisARFQ8ABBXhqP_cMW:b581m8q) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L38)):*
> ```Markdown
>     invariants](#trc-invariants) and [TRC update validation](#trc-update), as well as a TRC chain
>     verification.
> 2. __Valid:__ a *verified* TRC whose "validity" period (defined in the TRC itself) has begun and has
> ```

@kormat The wording here is a bit ambiguous - does this mean a TRC that’s outside of the grace period of the latest TRC can still be considered Valid?

Oncilla:
___
*[doc/ControlPlanePKI.md, line 52 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisAZyd0Lhb8meg5XuR:-LisAZyd0Lhb8meg5XuS:b-6dsh0g) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L52)):*
> ```Markdown
> - __Expired:__ a TRC whose validity period has ended, or that has been replaced by an update whose
>     grace period renders the previous TRC inactive.
> - __Base:__ a TRC with the version equal to the base version. This indicates that the TRC chain
> ```

@scrye :
What is a base version?

Oncilla:
___
*[doc/ControlPlanePKI.md, line 52 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisAZyd0Lhb8meg5XuR:-LisAcgM6sdn4i2z3hJs:bmuoais) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L52)):*

@scrye :
What is a base version?
</blockquote></details>

@kormat :
The base version is the trust anchor TRC (i.e. the most recent TRC with a grace period of 0).

Oncilla:
___
*[doc/ControlPlanePKI.md, line 61 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisAk9T0tr5UROEm75v:-LisAk9T0tr5UROEm75w:bbs8qse) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L61)):*
> ```Markdown
> - __Core__: AS that has core links to other core ASes.
> - __Voting__: AS that participates in and signs TRC updates.
> - __Authoritative__: AS that is authoritative for TRCs and certificates for the local ISD. I.e. must
> ```

@scrye : 
How strict is this requirement, timing wise?

Can protocols rely on the assumption that if an authoritative AS states that it doesn’t have some material, then it doesn’t exist?

If it is strict, there is some overlap between Issuing and Authoritative ASes, as probably the Issuing ones need to either play the role of being Authoritative, or register material with Authoritative ones before they are allowed to release them in the SCION world (which might be awkward to do, from an operations perspective).

Oncilla:
___
*[doc/ControlPlanePKI.md, line 946 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisBtWU4iNKsj1dvRWj:-LisBtWU4iNKsj1dvRWk:b7ilc2k) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L946)):*
> ```Markdown
> Attestations are used by verifiers to validate fresh or reset TRCs for remote ISDs. They are stored
> in the trust store and are considered first order citizens. When a verifier encounters an unknown
> base TRC during a TRC request, it fetches the attestation from the same node. Thus, even
> ```

@shitz : 
How would that attestation look like? What would be the TAAName? Can ISDs that do not want to use the RAA make up their own TAA?

Oncilla:
___
*[doc/ControlPlanePKI.md, line 946 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisBtWU4iNKsj1dvRWj:-LisBySs0mA8g_Yqmiof:b-b6jli3) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L946)):*

@shitz : 
How would that attestation look like? What would be the TAAName? Can ISDs that do not want to use the RAA make up their own TAA?
</blockquote></details>

No, they need to fetch the attestation from their RAA.
But they already have to get the ISD number from their RAA, thus they have to put some trust in the RAA.

Dealing with ungoverned ISD number space is a whole different case.

In theory, they can make up their own TAA. They just need other entities to trust that TAA.

Oncilla:
___
*[doc/ControlPlanePKI.md, line 1130 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisC7DE16oL71TQd7CU:-LisC7DE16oL71TQd7CV:b28x5vj) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L1130)):*
> ```Markdown
> are first class citizens, we do not want to force ASes to use them for their internal decision
> making. The verifiers periodically check whether a new TAAC exists. This is done in a hierarchical
> manner. Core ASes query the RAA directly. Non-core query their core. End hosts query the local AS.
> ```

@shitz : 
Is this done via SCION? If yes, then all the TRCs needed to get to the RAA will need to be preloaded during bootstrap, right?

Oncilla:
___
*[doc/ControlPlanePKI.md, line 531 at r3](https://reviewable.io/reviews/scionproto/scion/2831#-LiwBcMt0N2VjvrflcYQ:-LiwJJqCEiQKiAPgTjFh:b-7wr4qr) ([raw file](https://github.com/scionproto/scion/blob/202e2f27dbe0cc356b3e89e80c4b4ec78c4e2462/doc/ControlPlanePKI.md#L531)):*

Specifying the key version here is a bit weird. If you don't have a TRC version containing that issuing key, you don't have a clear way to figure out what version of the TRC you _do_ need.
</blockquote></details>

One should parse the TRC before verification anyway, thus there is no way of not knowing which TRC is needed. 
Update -> TRC.Version-1
TrustReset -> TRC itself

Oncilla:
___
*[doc/ControlPlanePKI.md, line 531 at r3](https://reviewable.io/reviews/scionproto/scion/2831#-LiwBcMt0N2VjvrflcYQ:-LiwSXVk-WdPDeyCfNqz:binuljd) ([raw file](https://github.com/scionproto/scion/blob/202e2f27dbe0cc356b3e89e80c4b4ec78c4e2462/doc/ControlPlanePKI.md#L531)):*

In that case, what is the point of listing the key version? If it disagrees with what's in the TRC, what do you do?
</blockquote></details>

You need a way do identify the signature?

Oncilla:
___
*[doc/ControlPlanePKI.md, line 531 at r3](https://reviewable.io/reviews/scionproto/scion/2831#-LiwBcMt0N2VjvrflcYQ:-LjLD_GP-TvD5vyWdmlV:bv41vtj) ([raw file](https://github.com/scionproto/scion/blob/202e2f27dbe0cc356b3e89e80c4b4ec78c4e2462/doc/ControlPlanePKI.md#L531)):*

I'll try to list the steps that are involved:
- Receive issuing cert.
- Parse payload
- load/fetch the TRC version mentioned in the Issuer section
The TRC then contains the issuing key version.

Thus the combination of IA+trc version should identify the signature precisely, and KeyVersion here seems to be redundant.

If the key version listed in the JWS signature is different from the version in the TRC, that's a hard fail.

OTOH, being able to identify a signature by its medata is a nice property. It just requires that we specify the extra checks that are necessary.
</blockquote></details>

When verifying a signature with a certificate chain, we want to be able to immediately tell whether the chain is still verifiable with the latest TRC -> we need to know if the key is the same.
If we have the key version here, this is trivial to do.

But, I guess we could also keep that state in the trust store instead of the cert.

Oncilla:
___
*[doc/ControlPlanePKI.md, line 670 at r8](https://reviewable.io/reviews/scionproto/scion/2831#-LjLLJRf1pRIWEza1zVu:-LjLOKVzA899wmt09Jle:bva1du1) ([raw file](https://github.com/scionproto/scion/blob/a0802174f781e0387e73c604035766cc69d55e7e/doc/ControlPlanePKI.md#L670)):*

This seems a bit harsh. Any primary AS which updates it's online/issuing key requires everyone else to sign off using offline keys. That seems to make a perverse incentive, which will make ASes less likely to update those keys.
</blockquote></details>

What do you propose then? 

- Issuing keys can be changed in regular update?
- 3 types of updates: regular, OnlineKeyOnly, sensitive.

In an OnlineKeyOnly update, only the ASes that have a changed online key are required to cast the vote with the offline key.

Oncilla:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/2831#-LjL_x6E5aBLgdqYfWXc:-LjLbDVVAZuMN9q9KnXm:b-txq7pi) (no related file):*

The ordering of sections seems a bit arbitrary:
```
$ grep '^## ' doc/ControlPlanePKI.md 
## Glossary
## Primary ASes
## Design Goals
## Trust Model
## Beacon Format (AS Entry) Modification
## TRC Format
## Certificate Format
## Certificate Chain
## Overview of Keys and Certificates
## <a name="trc-update"></a> TRC Updates
## TRC Update Dissemination
## TRC Trust Reset
## Certificate Chain Dissemination
## Signature Validation based on Certificate Chains
## Trust Material Sources
## Certificate Revocation
## <a name="trc-bootstrapping"></a> TRC Bootstrapping
## Appendix
```
It goes overview -> trc -> certs -> keys+certs -> trc -> cert chains -> cert revocation -> trc bootstrapping.
</blockquote></details>

I agree.
Let's reorder after most comments have been addressed, otherwise the comments will no longer be attached to the section they refer to.

I propose:

````
## Glossary
## Design Goals
## Trust Model
## Beacon Format (AS Entry) Modification
## Primary ASes
## Overview of Keys and Certificates

## TRC Format
## <a name="trc-update"></a> TRC Updates
## TRC Update Dissemination
## TRC Trust Reset

## Certificate Format
## Certificate Chain Format
## Signature Validation based on Certificate Chains
## Certificate Chain Dissemination
## Certificate Revocation

## Trust Material Sources
## <a name="trc-bootstrapping"></a> TRC Bootstrapping
## Appendix
````

---

Oncilla:
___
*[doc/ControlPlanePKI.md, line 643 at r2](https://reviewable.io/reviews/scionproto/scion/2831#-LisHzRW1nrygie6UE1V:-LjQWmPL-z_7O9SxNm6b:bi6uvtb) ([raw file](https://github.com/scionproto/scion/blob/59fe41b114f29eec279eea44e1a2a4c61eda8249/doc/ControlPlanePKI.md#L643)):*

Sure, there is no way around that. But I don't think that should be in the spec.
</blockquote></details>

(In this spec. I think we should specify it somewhere.)

PR #2846, opened 2019-07-16
Add SegReplyHandler for seg verification and storage

lukedirtwalker:
This abstract partial logic of fetcher from sciond to its own lib.
The idea is that this lib can also be used in the PS.
The SegReplyHandler verifies and stores a segment reply.
It can be triggered to store a partial result earlier and
it will provide feedback how many segments have been stored early.
This way users of the lib can potentially reply to client before they have processed the full reply.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2846)


Oncilla:
___
*[go/lib/infra/modules/segfetcher/segreplyhandler.go, line 131 at r2](https://reviewable.io/reviews/scionproto/scion/2846#-LjzUJ_f2Y9NY13moAOW:-LjzUJ_g2SmreeSwqsDX:b4phlrm) ([raw file](https://github.com/scionproto/scion/blob/c1e765d687c418639846eadc0a35a38d0e15561e/go/lib/infra/modules/segfetcher/segreplyhandler.go#L131)):*
> ```Go
> }
> 
> // Err is set after the FullReplyProcessed contained the number of stored
> ```

`Err indicates the error that happend when storing the segments. This should only be accessed after FullReplyProcessed channel has been closed`

PR #2878, opened 2019-07-22
Keep generated go files in the git repo

matzf:
* Add generated go/proto files to git
* Add make target `make gogen` to generate go/proto files
  Generate *.capnp.go and structs.gen.go files in bazel and then copy to source directory in a make target.

  For the case of structs.gen.go which is generated with a simple shell script, this setup is somewhat overcomplicated. Generating the *.capnp.go files does use bazel infrastructure to get the capnp tooling.
  To keep things symmetric, both cases are handled the same.

*  Add CI check generated go/proto files
  Detect uncommitted changes to generated go/proto files by cleaning and
regenerating the files.

Fixes https://github.com/scionproto/scion/issues/2822

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2878)


matzf:
I had to setup a workaround for a bug in `misspell` that created false positives in one of the generated `capnp.go` files. I've created a PR to fix it (https://github.com/client9/misspell/pull/157), but unfortunately it appears that the misspell maintainer is no longer around.

PR #2951, opened 2019-08-05
Add method to filter segments with a path policy

### LINKED ISSUE #2820: Implement module to filter segments with a given path policy
We should have a module that filters a set of segments with a given path policy. (dependency for #2818)

lukedirtwalker:
This method pre-filters a list of segments with a given path policy.
Note that filtering with sequences is not easily possible since we do not know where a segment would fit into the sequence (up, core, down).
A possibility to add sequence filtering for segments would be to split the sequence into up, core, and down parts.

Fixes #2820

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/2951)

### DIFF
--- a/go/lib/pathmgr/pathmgr.go
+++ b/go/lib/pathmgr/pathmgr.go
@@ -274,8 +274,7 @@ type pathWrap struct {
 	*spathmeta.AppPath
 }
 
-func (p pathWrap) Key() string     { return p.AppPath.Key().String() }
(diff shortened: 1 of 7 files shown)


lukedirtwalker:
As pointed out in the review this approach will not work as intended. Until we have a better solution I will close this PR.

lukedirtwalker:
We still want to do the filtering for core segments.

PR #3040, opened 2019-08-23
BR: reconnect to dispatcher support

sgmonroy:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3040)


scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/3040#-Lnlfh8NBKAcTvB5KU_o:-Lnlfh8NBKAcTvB5KU_p:b-cu1jxd) (no related file):*
Also, note that this will try to reconnect forever. There are some use cases where this is awkward (e.g., if the dispatcher socket is incorrect, a BR that starts up will never error out completely). Is this intended?

You can avoid reconnecting forever at the start by passing in a timeout to the `Listen`.

---

PR #3051, opened 2019-08-27
Remove Convey from layers package

chaehni:
---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3051)

### DIFF
--- a/go/lib/layers/BUILD.bazel
+++ b/go/lib/layers/BUILD.bazel
@@ -25,9 +25,9 @@ go_test(
     ],
     embed = [":go_default_library"],
     deps = [
-        "//go/lib/xtest:go_default_library",
         "@com_github_google_gopacket//:go_default_library",
         "@com_github_google_gopacket//layers:go_default_library",
-        "@com_github_smartystreets_goconvey//convey:go_default_library",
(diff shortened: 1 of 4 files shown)


Oncilla:
___
*[go/lib/layers/extensions_layer_test.go, line 77 at r2](https://reviewable.io/reviews/scionproto/scion/3051#-LnIElFV0fDSKN7ysc2T:-LnIKQQe1ciVf2ng5qIh:b81zzde) ([raw file](https://github.com/scionproto/scion/blob/f3a3ad7a56bc3e35b44742548bbeffbac5389d71/go/lib/layers/extensions_layer_test.go#L77)):*

since we do this already:
change `ExpectedError bool` to: `ErrorAssertion assert.ErrorAssertionFunc` and `ExpectedError: true` to `ErrorAssertion: assert.Error`m `ExpectedError: false` to `ErrorAssertion: assert.NoError`.
Then the code here is simply:
```go
tc.ErrorAssertion(t, err)
assert.Equal(t, tc.ExpectedExtension, extn, "extension must match")
```
</blockquote></details>

I would use `require.ErrorAssertionFunc`, no point in continuing, if the error is not as expected.

PR #3060, opened 2019-08-28
Check buffer size before reading in raw functions

### LINKED ISSUE #2050: Parsing from raw functions should check that the buffer is big enough
As an example, MetaFromRaw should check buffer length before trying to parse anything from the buffer:
```
func MetaFromRaw(b []byte) (*Meta, error) {
        if len(b) < MetaLen {
                return nil, common.NewBasicError("Can't parse SCMP meta subheader, buffer is too short",
                        nil, "expected", MetaLen, "actual", len(b))
        }
        m := &Meta{}
        if err := restruct.Unpack(b, binary.BigEndian, m); err != nil {
                return nil, common.NewBasicError("Failed to unpack SCMP Metadata", err)
        }
        return m, nil
}
```

karampok:
Fixes #2050

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3060)


karampok:
___
*[go/lib/hpkt/read.go, line 182 at r2](https://reviewable.io/reviews/scionproto/scion/3060#-LnWKmTADFqCfYvEt1Fv:-LnWPGbtC5GYDjZyU4-K:b-9fefku) ([raw file](https://github.com/scionproto/scion/blob/9a5ae84247573236b20fc37c968183c88f39ea44/go/lib/hpkt/read.go#L182)):*

Todo for this PR, or a future one?
</blockquote></details>

Topic to discuss, I would not hardcoded but this makes is safe. What do you think?
(I vote to just remove the todo)

karampok:
___
*[go/lib/hpkt/read.go, line 176 at r2](https://reviewable.io/reviews/scionproto/scion/3060#-LnWO48O55iraE4P8rJB:-LnWP9-a2it_fhYxinWq:b-96wuy2) ([raw file](https://github.com/scionproto/scion/blob/9a5ae84247573236b20fc37c968183c88f39ea44/go/lib/hpkt/read.go#L176)):*

remove
</blockquote></details>

I think this error is needed because I have removed it (or I think) from the function we call the CmdHdrParser() . If you can check versus the master commit

karampok:
___
*[go/lib/hpkt/hpkt_test.go, line 173 at r4](https://reviewable.io/reviews/scionproto/scion/3060#-LnWqrfo0mO0SgLELvRt:-LnX36KE6_a6Vzv5tjxu:b-a4vq7w) ([raw file](https://github.com/scionproto/scion/blob/3b1aa58b6826876a4bea3e9c447ac5cfd5865111/go/lib/hpkt/hpkt_test.go#L173)):*
> require.NotContains(t, err.Error(), "panic")
the thing is I want to have a input that does panic, I make sure that nobody remove the defer/recover from the flow. Does it makes sense?

PR #3064, opened 2019-08-29
Avoid cgo in lib/overlay/conn to simplify cross-compilation

matzf:
cgo was only required to get a portable struct timespec.
The `syscall.Timespec` struct seems to provide this type just as well, without requiring cgo.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3064)


matzf:
@kormat asked for documentation that `unsafe.Sizeof(syscall.Timespec)` is *guaranteed* to be the same as what C considers.

I could not find any actual documentation (I find golang documentation to be rather lackluster). Still, this seems rather safe. Looking at the various `syscall/types_xxx.go` source files (see https://golang.org/src/syscall), Timespec is defined as `type Timespec C.struct_timespec` on most platforms. It's explicitly defined for windows, plan9 and nacl.

This is the complete list of `unsafe.Sizeof(syscall.Timespec)` for all supported GOOS/GOARCH combinations:
```
darwin/386: 8
darwin/amd64: 16
dragonfly/amd64: 16
freebsd/386: 8
freebsd/amd64: 16
freebsd/arm: 16
linux/386: 8
linux/amd64: 16
linux/arm: 8
linux/arm64: 16
linux/ppc64: 16
linux/ppc64le: 16
linux/mips: 8
linux/mipsle: 8
linux/mips64: 16
linux/mips64le: 16
linux/s390x: 16
nacl/386: 12
nacl/amd64p32: 16
nacl/arm: 12
netbsd/386: 12
netbsd/amd64: 16
netbsd/arm: 16
openbsd/386: 12
openbsd/amd64: 16
openbsd/arm: 16
plan9/386: 8
plan9/amd64: 8
plan9/arm: 8
solaris/amd64: 16
windows/386: 16
windows/amd64: 16
```
This corresponds to `sizeof(struct timespec)` in C for all platforms that I could check: see this compiler comparison on godbolt: https://godbolt.org/z/Lo3ekb

matzf:
It seems unlikely to me that this could be an issue as timespec is a particularly simple data type.
The issue you refer to with zero-length struct members really seems different; the cgo documentation (now) explicitly mentions that this isn't supported.

Regardless, as the in our case the sizeof is only used to allocate a big enough buffer, maybe a compromise could be to explicitly use a constant size that can be expected to be >= `sizeof(struct timespec)` on any platform? Like, e.g. 16?


Btw. looking at the documentation for `SO_RXQ_OVFL` (the other value that is requested into/read from the oob), the value is defined as a 32-bit integer but it's currently read as a golang `int`. Is that intentional or a typo?

Issue #3065, opened 2019-08-30
Env/deps does not respect non-interactive mode


mkowalski:
As an operator I want to have a possibility to perform a fully non-interactive installation of SCION and it's dependencies. Currently one of the steps is running `./env/deps` script. This one requires an admin interaction as in the session below

```
scionlab@scionlab-1107-einstein:~/go/src/github.com/scionproto/scion$ ./env/deps 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following additional packages will be installed:
  libsodium18 python3-cffi python3-cffi-backend python3-cryptography python3-idna python3-ply python3-pyasn1
  python3-pycparser
Suggested packages:
  python-cryptography-doc python3-cryptography-vectors python-nacl-doc python-openssl-doc python3-openssl-dbg python-ply-doc
  doc-base
The following NEW packages will be installed:
  libsodium18 python3-cffi python3-cffi-backend python3-cryptography python3-idna python3-nacl python3-openssl python3-ply
  python3-pyasn1 python3-pycparser
0 upgraded, 10 newly installed, 0 to remove and 109 not upgraded.
Need to get 736 kB of archives.
After this operation, 4,210 kB of additional disk space will be used.
Do you want to continue? [Y/n] 
```

As a result, installation or upgrade of SCION requires a manual interaction what is not a desired behaviour when performing the action in >1 machine.

kormat:
The recommend approach for this is to do `APTARGS=-y ./env/deps`, which is already supported.

PR #3066, opened 2019-08-30
Run apt-get without user interaction

mkowalski:
Currently apt-get will require a user confirmation if there are changes to be applied. This change makes apt-get assume "yes" as an answer to all the prompts.

Closes-bug: #3065

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3066)


kormat:
Making changes to a users system without explicit consent is something we want to avoid.

Issue #2817, opened 2019-06-28
PS API: add path policy to seg request


lukedirtwalker:
Add a path policy field to the segments request so that PS clients can specify a path policy.

lukedirtwalker:
This is not strictly required. Also as outlined in #2820 it is not clear how we would filter in the PS. So we close this issue for now, and create a new one if we really deem it necessary.

Issue #2816, opened 2019-06-28
sciond API: add path policy support to path requests


lukedirtwalker:
sciond path requests should have a path policy field that sciond considers.

scrye:
We should probably close this due to age/lack of context. This will come up again eventually, but we can create a new issue then.

@lukedirtwalker: what do you think?

lukedirtwalker:
Yeah this is old and not clearly specified. So let's close and create a new issue once we have the details specified.

PR #3082, opened 2019-09-03
pathpol: Add Policy to PathReq and SegReq capnp

lukedirtwalker:
Adds low and mid-level objects to represent path-policies.

Contributes #2816
Contributes #2817

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3082)


lukedirtwalker:
It's not yet clear we will need this in the near future, so we close it for now.

PR #3108, opened 2019-09-05
Use MinRevTTL instead of constant

juagargi:
In SCIONLab, if we change the value `path_mgmt.MinRevTTL` from 10 to e.g. 20, some UTs in `beacon_srv` won't pass. Using the constant `MinRevTTL` instead of hardcoded `10` fixes them.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3108)


juagargi:
Closing this in favor of https://github.com/scionproto/scion/pull/3111

PR #3151, opened 2019-09-16
BS: Add metrics for the keepalive messages

karampok:
This PR works the first bullet on https://github.com/scionproto/scion/issues/3104
```
# HELP bs_keepalive_receive_msgs_total Total number of received keepalive msgs.
# TYPE bs_keepalive_receive_msgs_total counter
bs_keepalive_receive_msgs_total{ifid="1",result="success"} 118
bs_keepalive_receive_msgs_total{ifid="2",result="success"} 118
# HELP bs_keepalive_transmit_msgs_total Total number of transmitted keepalive msgs.
# TYPE bs_keepalive_transmit_msgs_total counter
bs_keepalive_transmit_msgs_total{ifid="1",result="process_err"} 118
bs_keepalive_transmit_msgs_total{ifid="2",result="process_err"} 118
```

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3151)

### DIFF
--- a/go/beacon_srv/internal/keepalive/BUILD.bazel
+++ b/go/beacon_srv/internal/keepalive/BUILD.bazel
@@ -11,6 +11,7 @@ go_library(
     visibility = ["//go/beacon_srv:__subpackages__"],
     deps = [
         "//go/beacon_srv/internal/ifstate:go_default_library",
+        "//go/beacon_srv/internal/metrics:go_default_library",
         "//go/beacon_srv/internal/onehop:go_default_library",
         "//go/lib/addr:go_default_library",
         "//go/lib/common:go_default_library",
@@ -50,6 +51,7 @@ go_test(
         "//go/lib/topology:go_default_library",
         "//go/lib/xtest:go_default_library",
         "@com_github_golang_mock//gomock:go_default_library",
-        "@com_github_smartystreets_goconvey//convey:go_default_library",
+        "@com_github_stretchr_testify//assert:go_default_library",
+        "@com_github_stretchr_testify//require:go_default_library",
     ],
 )
--- a/go/beacon_srv/internal/keepalive/handler.go
+++ b/go/beacon_srv/internal/keepalive/handler.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/scionproto/scion/go/beacon_srv/internal/ifstate"
+	"github.com/scionproto/scion/go/beacon_srv/internal/metrics"
 	"github.com/scionproto/scion/go/lib/addr"
 	"github.com/scionproto/scion/go/lib/common"
 	"github.com/scionproto/scion/go/lib/ctrl/ifid"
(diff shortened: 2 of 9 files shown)


karampok:
___
*[go/beacon_srv/internal/keepalive/handler.go, line 102 at r1](https://reviewable.io/reviews/scionproto/scion/3151#-LoueWwP00RfIjMpZiMS:-LoxuUcG9Dg6RM_1J9eA:b-o7381o) ([raw file](https://github.com/scionproto/scion/blob/8c3e535e90cba495ff25fcbc15f6b279d56e546b/go/beacon_srv/internal/keepalive/handler.go#L102)):*

same, probably different err message. e.g. `err_processing`
</blockquote></details>

I have on purpose avoid having  label on what error  it is.
The idea is that we have an alert on a metric error that will show the errors   on bs on server X are increasing.  In that moment ops should go to logs to see details. What do you think?

karampok:
___
*[go/beacon_srv/internal/keepalive/metrics.go, line 37 at r1](https://reviewable.io/reviews/scionproto/scion/3151#-LouEto19amJTzVm5JQE:-Loxy92x7IVCPkyjCW3k:b-s10jhu) ([raw file](https://github.com/scionproto/scion/blob/8c3e535e90cba495ff25fcbc15f6b279d56e546b/go/beacon_srv/internal/keepalive/metrics.go#L37)):*

instead of having `outMsg` and `outErr`, have one counter vector with labels that distinguish the results.
This allows for easy sums in promql
e.g. labels: `ifid: {1,2,...} , `result`= { "success", "err_payload", "err_send", "err_db"}

`bs_keepalive_sent_total(ifid, result) count`
`bs_keepalive_received_total(ifid, result) count`



</blockquote></details>

Having labels with high cardinality will blow up eventually the prometheus server.
https://prometheus.io/docs/practices/naming/#labels (the yellow box).

And i followed the approach of node_exporter which is
node_network_{transmit,receive}_{bytes,error} and all metrics from component X should have the same labels (e.g. env=prod, os=linux)

karampok:
___
*[go/beacon_srv/internal/keepalive/metrics.go, line 25 at r1](https://reviewable.io/reviews/scionproto/scion/3151#-LouTS-PC0l1vKdQAHvo:-Loy5_be9rsAb-NrRnqE:byjbl2s) ([raw file](https://github.com/scionproto/scion/blob/8c3e535e90cba495ff25fcbc15f6b279d56e546b/go/beacon_srv/internal/keepalive/metrics.go#L25)):*

I would maybe add them to the variable declaration. So it's a bit closer and would maybe be seen when doing refactorings.
</blockquote></details>

I think this would add more confusion, so I remove them. In all case you can grep by `transmit_msgs_total`

karampok:
___
*[go/beacon_srv/internal/keepalive/metrics.go, line 44 at r1](https://reviewable.io/reviews/scionproto/scion/3151#-Loub0Cp1D7zTokvDIpy:-Loy5nBnE2PGyWs4Q_2W:bozrxep) ([raw file](https://github.com/scionproto/scion/blob/8c3e535e90cba495ff25fcbc15f6b279d56e546b/go/beacon_srv/internal/keepalive/metrics.go#L44)):*

the way this is done now, you could also simply use a `init` function.
there is no way of using the package without calling `InitMetrics` first anyway.
</blockquote></details>

I like the `init` function. Do you think we still need the `initOnce.Do`?
Or you prefer to do exactly as your `beaconing/metric` implementation with a struct?

karampok:
___
*[go/beacon_srv/internal/metrics/keepalive.go, line 35 at r2](https://reviewable.io/reviews/scionproto/scion/3151#-Lp2v1206mZ7n6Kibzcu:-Lp30h3g5L3XMG_np2xL:b2fed6p) ([raw file](https://github.com/scionproto/scion/blob/fbb7782e9c70ce16a7fd999d4e51177850eb74d0/go/beacon_srv/internal/metrics/keepalive.go#L35)):*

`Values`
</blockquote></details>

Why do we need to export that?

PR #3174, opened 2019-09-20
TrustStore: Define components

Oncilla:
This PR introduces the different components of the trust store.
In the end, the client package facing parts of the trust store are
constructed out of the `CryptoProvider` and the `Inspector`.

The other components are used internally to modify the behavior of the
trust store.

E.g. different `Recurser`s will be used between the CS and sciond.

The purpose of the components is the following:
- Inspector: Implement the infra.ASInspector
- CryptoProvider: Main logic when fetching trust material. It decides
  when to query the database and/or the network.
- Inserter: Verify and insert TRC/certificates into the database.
- Recurser: Decide whether recursion is allowed.
- Router: Get path based on the trust material subject.
- RPC: Abstraction of the messenger stack.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3174)


scrye:
___
*[go/lib/infra/modules/trust/v2/decode.go, line 25 at r1](https://reviewable.io/reviews/scionproto/scion/3174#-LpTij_IElCxMlMK0ocW:-LpcnYfcBV4mYWc0xJ9S:b3cqoc) ([raw file](https://github.com/scionproto/scion/blob/4c2a04894e4410036e647120e5439a76b13c1740/go/lib/infra/modules/trust/v2/decode.go#L25)):*

I don't think so.
Otherwise we need to always pass `Raw` and decode it everywhere?
</blockquote></details>

Now that it's just an internal convenience type it's fine; external packages no longer need to be worried about their objects becoming inconsistent.

PR #3180, opened 2019-09-24
bs: remove convey from beaconing handler

karampok:
As part of the general remove convey effort

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3180)


karampok:
The changed were as part of https://github.com/scionproto/scion/pull/3186

PR #3271, opened 2019-10-23
Add String() to snet.Path interface

sustrik:
Useful for logging & debugging purposes.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3271)


karampok:
Without being able to see how it is being used, I can not really have a review of the implementation.
I would suggest given the size is small that you insert the changes in the first time you actually make use of it.


Otherwise we can pair programming and you provide me a temporary branch where this commit is being used and show what I observe before and after.

PR #3332, opened 2019-11-06
Refactor overlay addresses to always use net.IP

scrye:
We currently do not support any other overlay L3. Also, the overlay
connection objects always expect IP addresses. This makes the
net.IP address requirement explicit in the overlay address constructor.
Also, the constructor can no longer error out.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3332)


scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/3332#-Lt_26EK1UJQD5w0Q3ez:-Lt_26EM3EihKhHxkU6h:b45n0yr) (no related file):*
Full acceptance run: https://buildkite.com/scionproto/acceptance/builds/615.

---

Issue #3016, opened 2019-08-20
test: Get rid of goconvey


oncilla:
Goconvey is slightly painful to work with. Also, upstream seems very inactive.

The vanilla golang sub-tests provide the functionality we need already.

matzf:
The following packages still use `goconvey`:
- `dispatcher/internal/registration` (~400 LoC)
- `pkg/sock/reliable/reconnect` (~550 LoC)
- `private/revcache` (~550 LoC)

PR #3358, opened 2019-11-12
godispatcher: remove convey

karampok:
Contributes #3016

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3358)

### DIFF
--- a/go/godispatcher/internal/registration/iatable_test.go
+++ b/go/godispatcher/internal/registration/iatable_test.go
@@ -18,159 +18,155 @@ import (
 	"net"
 	"testing"
 
-	. "github.com/smartystreets/goconvey/convey"
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


scrye:
___
*[go/godispatcher/internal/registration/table_test.go, line 199 at r1](https://reviewable.io/reviews/scionproto/scion/3358#-Lt_1A0nAZCDNdABt527:-Ltd_LIx0cFCwp06OdEP:be79e02) ([raw file](https://github.com/scionproto/scion/blob/2f1efd25cb779c9fc3717a1d206ffe8bf516656c/go/godispatcher/internal/registration/table_test.go#L199)):*

Is there an issue to remove everything around Binds?
I would suggest we let it until that story is implemented.
</blockquote></details>

Fair enough, this can wait.

PR #3366, opened 2019-11-12
Do not use snet global state

### LINKED ISSUE #3135: snet: Get rid of package global network
Package globals are bad.
If an application really wants an application global network, it should just create it.

karampok:
Do not use snet global state

* Changes the code of {pingpong, integration, sig} to not use the
global state of snet pkg

*Remove code around global state in snet pkg

Fixes #3135

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3366)

### DIFF
--- a/go/examples/pingpong/pingpong.go
+++ b/go/examples/pingpong/pingpong.go
@@ -56,24 +56,22 @@ const (
 )
 
 var (
-	local  snet.Addr
-	remote snet.Addr
-	file   = flag.String("file", "",
-		"File containing the data to send, optional to test larger data (only client)")
-	interactive = flag.Bool("i", false, "Interactive mode")
-	id          = flag.String("id", "pingpong", "Element ID")
-	mode        = flag.String("mode", ModeClient, "Run in "+ModeClient+" or "+ModeServer+" mode")
-	sciond      = flag.String("sciond", "", "Path to sciond socket")
-	dispatcher  = flag.String("dispatcher", "", "Path to dispatcher socket")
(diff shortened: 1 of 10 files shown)


scrye:
___
*[go/sig/sigcmn/common.go, line 84 at r3](https://reviewable.io/reviews/scionproto/scion/3366#-Ltd_YzF-duIvSJC3hF-:-Ltd_YzF-duIvSJC3hF0:bzhlis) ([raw file](https://github.com/scionproto/scion/blob/65dacc6581ec52b0a7247c2b5656a711e624459d/go/sig/sigcmn/common.go#L84)):*
> ```Go
> }
> 
> // initSNET initializes snet. Tries in second interval until sdCfg.InitialConnectPeriod is up.
> ```

I checked the code and it's not safe to remove this yet. In https://github.com/scionproto/scion/issues/3377 we can revisit this.

scrye:
___
*[go/sig/sigcmn/common.go, line 84 at r3](https://reviewable.io/reviews/scionproto/scion/3366#-Ltd_YzF-duIvSJC3hF-:-Ltd_lcE3cH7TJ9hl05m:b6rv81b) ([raw file](https://github.com/scionproto/scion/blob/65dacc6581ec52b0a7247c2b5656a711e624459d/go/sig/sigcmn/common.go#L84)):*

I checked the code and it's not safe to remove this yet. In https://github.com/scionproto/scion/issues/3377 we can revisit this.
</blockquote></details>

By safe to remove, I mean the initialization loop in the `for` below.

PR #3370, opened 2019-11-13
Reduce usage of hostinfo.Host

### LINKED ISSUE #3295: Remove hostinfo.Host from every package except sciond API
`hostinfo.Host` is a serialization format for addresses passed through the SCIOND client API. No other code should depend on this wire format.

`UDPAddr` should be used instead wherever possible.

Related issue: https://github.com/scionproto/scion/issues/1842 (although it mostly concerns capnp wire format changes, rather than the other uses of the type).

lukedirtwalker:
Instead use *net.UDPAddr

Fixes #3295

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3370)

### DIFF
--- a/go/integration/cert_req/BUILD.bazel
+++ b/go/integration/cert_req/BUILD.bazel
@@ -11,7 +11,6 @@ go_library(
         "//go/lib/addr:go_default_library",
         "//go/lib/common:go_default_library",
         "//go/lib/ctrl/cert_mgmt:go_default_library",
-        "//go/lib/hostinfo:go_default_library",
         "//go/lib/infra:go_default_library",
         "//go/lib/infra/disp:go_default_library",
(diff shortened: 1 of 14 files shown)


scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/3370#-Lt_4Urf9e54Tmak8KK-:-LtdCShE1gps1znZXLt8:btd5a1r) (no related file):*

Done. Added them pretty much everywhere.
</blockquote></details>

Looks good. Maybe add a `copyIP` function to `hostinfo` and `itopo`? It's very easy to make a mistake in one of the indices, and it's not caught by the compiler. Having every package that does extensive copying have a function is cleaner.

Don't make it a global package `util` or `common` function, I think that's overkill.

---

PR #3372, opened 2019-11-13
Install python3 version of supervisor, ditch pip2

matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3372)


matzf:
Note: somewhat disappointingly, the debian step still installs python2.

Issue #3344, opened 2019-11-08
Remove addr.AppAddr


scrye:
Instead, use `net.UDPAddr` whenever possible. For code needing to move SVC addresses around, consider using `net.Addr` instead and have the address cast to the type at the point of use.

Probably the easiest way to implement this is to create multiple PRs that remove `AppAddr` a few packages at a time.

karampok:
After the snet write/read refactor, there are the following references
```

15:46 $ ag addr.appaddr
go/lib/snet/svcaddr.go
40:             Host:    &addr.AppAddr{L3: a.SVC},

go/lib/snet/addr.go
37:     Host    *addr.AppAddr
184:    return &Addr{IA: ia, Host: &addr.AppAddr{L3: l3, L4: l4}}, nil

go/lib/snet/addr_test.go
30:     host4 := &addr.AppAddr{
34:     host6 := &addr.AppAddr{

go/lib/snet/udpaddr.go
43:             Host: &addr.AppAddr{

```
Which would be removed once the snet.Addr is removed (different stories)

PR #3375, opened 2019-11-13
BS, DISP: Use net.UDPAddr instead of addr.AppAddr

karampok:
Contributes #3344

This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3375)

### DIFF
--- a/go/beacon_srv/internal/beaconing/originator_test.go
+++ b/go/beacon_srv/internal/beaconing/originator_test.go
@@ -29,7 +29,6 @@ import (
 	"github.com/scionproto/scion/go/beacon_srv/internal/beacon"
 	"github.com/scionproto/scion/go/beacon_srv/internal/ifstate"
 	"github.com/scionproto/scion/go/beacon_srv/internal/onehop"
-	"github.com/scionproto/scion/go/lib/addr"
 	"github.com/scionproto/scion/go/lib/common"
 	"github.com/scionproto/scion/go/lib/ctrl"
 	"github.com/scionproto/scion/go/lib/ctrl/seg"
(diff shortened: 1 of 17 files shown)


karampok:
___
*[go/lib/discovery/discoveryinfo/info.go, line 61 at r1](https://reviewable.io/reviews/scionproto/scion/3375#-LtchG5j53UDknmCV9s3:-Ltd1hya9FEqYi2dhbIB:b-ecsq15) ([raw file](https://github.com/scionproto/scion/blob/f5037fa1625d9cd471ec341d485e7280605b12de/go/lib/discovery/discoveryinfo/info.go#L61)):*

Previously there was a copy, don't we need that anymore?
</blockquote></details>

To my understanding no, because inside the *net.UDPAddr we have no other pointer reference or interface. Therefore we are safe.

If not, we have to find a unit-test which proves otherwise.

scrye:
___
*[go/lib/discovery/discoveryinfo/info.go, line 61 at r1](https://reviewable.io/reviews/scionproto/scion/3375#-LtchG5j53UDknmCV9s3:-LteFA0N-DvXeQ1DyTOa:b-b33i5t) ([raw file](https://github.com/scionproto/scion/blob/f5037fa1625d9cd471ec341d485e7280605b12de/go/lib/discovery/discoveryinfo/info.go#L61)):*

To my understanding no, because inside the *net.UDPAddr we have no other pointer reference or interface. Therefore we are safe.

If not, we have to find a unit-test which proves otherwise.

</blockquote></details>

There is a reference though, the `net.IP` slice. If `a.IP` changes after `Update` is called (e.g., a.IP[2] = 42), it will change `h.addr.IP[2]` to 42.

PR #3376, opened 2019-11-14
snet: Make local address optional

matzf:
* ~~Make `IA` optional in `snet.Init`. 
   If it is unspecified, just ask the chosen sciond (using existing API).~~
* Allow unspecified local address in `snet.Dial/Listen`
  Register unspecified address (i.e. 0.0.0.0) in the dispatcher.
  The godispatcher already handles this correctly.

   When using an unspecified source address for sending a packet, look up the route to the next hop in the kernel routing table and use the corresponding source address as source address in the SCION packet.
   This uses `netlink` and so is specific to Linux.

* ~~Demo: make `-local` optional in `pingpong` and remove `-srcIA` in `showpath`~~

(Update: rebased only the essentials. The `snet.Init` part is gone because the API changed)

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3376)

### DIFF
--- a/go/lib/snet/BUILD.bazel
+++ b/go/lib/snet/BUILD.bazel
@@ -34,6 +34,7 @@ go_library(
         "//go/lib/spath:go_default_library",
         "//go/lib/spkt:go_default_library",
         "//go/lib/topology/overlay:go_default_library",
+        "@com_github_vishvananda_netlink//:go_default_library",
     ],
 )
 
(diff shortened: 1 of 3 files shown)


matzf:
In the meantime we've been working around this by using a simple default local IP instead of "proper" wildcard addresses, which is something like a ~90% solution for now.
I would still like to do this properly, eventually. Just a thought: perhaps, this could be combined with some changes to snet (or the dispatcher, whatever), to look up the next-hop address based on the first interface of the chosen path, only at the very moment of sending out the packet. This would free the "application layer" from passing around this next hop address in the `snet.Path`. This routing lookup would be the ideal place to also lookup (and cache) the correct source IP address for each next-hop interface address.

PR #3395, opened 2019-11-18
Avoid DialUDP in border to simplify startup/recovery

matzf:
This is a proposal to try to improve the error-recovery for the border router. In essence, the problem that we face is that when a network interface goes down, the border router will get an error for write operation once, and then shut down the writer for the corresponding border router interface, but it will never recover and will also not terminate the process.

Instead of `DialUDP`, simply `ListenUDP` and manually specify the remote address in each `Write`/`WriteBatch`.
Now this could be further simplified by removing `Remote` entirely from `overlay.Conn` and just letting the callers (border and dispatcher) always set the remote address directly.

The purpose of this change is to allow a border router to (properly) bind to the `0.0.0.0` overlay address. Binding to `0.0.0.0` could be very helpful for the situation where we don't have guarantees about whether the interface with the specified public IP is even up at the time the border router is started, or when the interface can go down intermittently. In SCIONLab, this situation occurs with running overlay links over VPN.
With dial-ing on startup, the socket is immediately bound to the effective source address, even if we specify `0.0.0.0` as the bind address. If the corresponding network interface ever goes down, the underlying socket breaks and the border router will never recover, even if the network interface comes back up.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3395)


mkowalski:
Hi guys,

A small comment regarding one of the propositions, i.e.

> If you want to be independent of a given interface existing, i would configure the ip on the loopback interface so that it's always available.

This solution is super-ugly for anyone operating something else than a local workstation (because it would mean later you get the same IP on a TUN interface and you either mangle with routing table or write lots of unnecessarily complicated machinery around). A better way to achieve it is is using `IP_FREEBIND` option which allows you to bind to an address which does not exist -- http://man7.org/linux/man-pages/man7/ip.7.html

matzf:
Fixed (as discussed above) in #3547

PR #3417, opened 2019-11-21
yang files for proposal

JordiSubira:
Yang modules:
I divided them into 3 modules:
scion-types.yang: Typedefs, types, common grouping, identities and so on for the topology.

scion-interfaces.yang: data structure with grouping for interface structure. I separated, just in case it could be reusable, besides it makes the scion-topology more readable.

scion-topology.yang: Data model exposing, somehow, a mapping for the topology.json file. With this setting, every device would maintain a copy from the topology for its ISD-AS (likewise now with the topology.json file). It'd be interesting considering whether the services should be exposed separately, i.e. creating a module for border-routers, beacon-servers, and so on, in case they don't need to maintain the entire configuration. 

Python script:
So far it parses the scion-topology data store and outputs one topology.json into every service folder for the ISD-AS (it takes the isd-as defined within the same datastore).

Some things to check and provide any remarks :
1. Servers and BR names pattern is correct
2. ISD-AS pattern is correct
3. Link enumeration is correct 
4. Improving descriptions if possible
5. Check solution for underlays and addresses in BR-interfaces
6. Must scion-topology.yang expose any state data.
7. Considering IPv6 within the parser (if the structure is kept in the topology.json the change should be trivial).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3417)

### DIFF
--- a/scion.sh
+++ b/scion.sh
@@ -276,7 +276,7 @@ cmd_lint() {
 py_lint() {
     lint_header "python"
     local ret=0
-    for i in acceptance python; do
+    for i in acceptance python yang; do
       [ -d "$i" ] || continue
       local cmd="flake8"
       lint_step "$cmd /$i"
--- a/yang/README.md
+++ b/yang/README.md
@@ -1,18 +1,29 @@
 # YANG modules and NETCONF
 
-The python script uses sysrepo+Netopeer2 as persistance and NETCONF Server solutions. For more information about how to install sysrepo+Netopeer2 see: https://github.com/sysrepo/sysrepo and https://github.com/CESNET/Netopeer2/tree/master/server.
+The python script uses sysrepo+Netopeer2 as persistance and NETCONF
+Server solutions. For more information about how to install sysrepo+Netopeer2
+see: <https://github.com/sysrepo/sysrepo>
+and <https://github.com/CESNET/Netopeer2/tree/master/server>.
 
 ## Installation YANG modules
+
 YANG module can be installed using CLI as follows:
+
 ```sh
-$ sysrepoctl --install --yang={MODULE_NAME} --owner={USER}:{GROUP} --permission={OCTAL-REP}
+sysrepoctl --install --yang={MODULE_NAME} --owner={USER}:{GROUP} --permission={OCTAL-REP}
 ```
 
-More information can be found here: http://www.sysrepo.org/.
+More information can be found here: <http://www.sysrepo.org/>.
+
 ## Usage python scripts
 
 The `write_topo.py` script can be used as follows:
+
 ```sh
-$ python write_topo.py [--test]
+python write_topo.py [--test]
 ```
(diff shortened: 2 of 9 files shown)


JordiSubira:
> > _[Reviewable](https://reviewable.io/reviews/scionproto/scion/3417)_ status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @JordiSubira)
> 
> _[yang/yang_modules/scion-types.yang, line 110 at r1](https://reviewable.io/reviews/scionproto/scion/3417#-LuCG9Mw9qkUn6EGK1qL:-LuCG9Mw9qkUn6EGK1qM:b-kv0guy) ([raw file](https://github.com/scionproto/scion/blob/a452426693288d02f9ab94a09a5f0be35c2cb25f/yang/yang_modules/scion-types.yang#L110)):_
> 
> > ```yang
> >     description "This is the type definition for the ISD-AS name";
> >     type string{
> >       pattern "[0-9]+-[0-9a-f]{4}:[0-9]:[0-9]+";
> > ```
> 
> This is quite wrong, see https://github.com/scionproto/scion/wiki/ISD-and-AS-numbering
> 
> Maybe better:
> 
> ```
> [0-9]+-[0-9a-fA-F]{1,4}:[0-9a-fA-F]{1,4}:[0-9a-fA-F]{1,4}
> ```

Ok, great. I wasn't aware about that documentation, thank you. Could you confirm this would be ok for the base-service:
`pattern "[a-z]{2}[0-9]+-[0-9a-fA-F]{1,4}_[0-9a-fA-F]{1,4}_[0-9a-fA-F]{1,4}+-[0-9]+"`

JordiSubira:
Thank y'all for feedback, I really appreciate. I might have done this a draft pull request since my first intention was to discuss the bullet list because I was not sure about certain requirements and I thought it was easier to share the code to look it up. I'll work on your comments, so that I can improve the solution. Feel free to suggest any other changes.

JordiSubira:
> > _[Reviewable](https://reviewable.io/reviews/scionproto/scion/3417)_ status: 0 of 6 files reviewed, 11 unresolved discussions (waiting on @JordiSubira, @scrye, and @worxli)
> 
> _[yang/yang_modules/scion-types.yang, line 110 at r1](https://reviewable.io/reviews/scionproto/scion/3417#-LuCG9Mw9qkUn6EGK1qL:-LuaUii5AXULTPRxP5JP:b-qbp8q6) ([raw file](https://github.com/scionproto/scion/blob/a452426693288d02f9ab94a09a5f0be35c2cb25f/yang/yang_modules/scion-types.yang#L110)):_
> 
> Previously, JordiSubira wrote…
> As far as I understand these service names are just the convention used by the current topology generator and is not at all mandatory, so I guess these could be just any string.

If you can confirm that this is as above, I'll remove the constraints for the services names. I'd appreciate if you or somebody else can double-check.

scrye:
I'm going to close this for now. The main issue is that the roadmap for the near/mid-term future is already pretty full, and there is just not enough room/necessity for NETCONF and YANG.

Additionally, we are looking at streamlining the topology file, and it might even come to the point where it is removed completely because it is not needed anymore.

We'll definitely revisit this contribution when NETCONF becomes a priority again.

PR #3419, opened 2019-11-21
Add go.mod file, use as master for go_deps.bzl

matzf:
Add `go.mod` file; semi-manually generated from bazel dependencies. Some tweaks:
* Explicitly specify version for dependencies vendored in quic-go (sidh, qtls, genny). This is required because go mod ignores indirect vendored dependencies.
* Update jaeger-client-go (to latest) to make go mod tidy work (cf. jaegertracing/jaeger-client-go#463)
* Add new dependency go.uber.org/atomic for jaeger-client-go
* Update go-colorable to fix warning about logrus importpath

Generate go_deps.bzl from `go.mod` by invoking `make godeps` -- the `godeps` target is just an alias for `make go_deps.bzl` to keep the usage somewhat consistent.

Remove `goenv`, `vendor` and make targets.
Remove  `bzlcompat` requirement.

To make the base build work, exclude various unused dependencies (these are indirect test dependencies) in fetch.sh.

Add CI step to check go_deps.bzl is up to date with go.mod, and that go.mod and go.sum are tidy.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3419)


matzf:
The `fetch.sh` seems even nastier now, it requires many trial-and-error steps to make it work. The only simple alternatives I can see are:
* run `bazel sync`; the generated cache is huge though (~3.5G, compressed)
* copy all the BUILD files into the build container and run something like `bazel fetch //:gazelle //:scion //:scion-ci //:lint //:build-tools //:nogo //docker/perapp:debug //docker/perapp:prod`; not tested, not sure if this would work

PR #3428, opened 2019-11-25
Remove unused functions from topology

scrye:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3428)


scrye:
Not needed anymore after https://github.com/scionproto/scion/pull/3440 was merged.

PR #3435, opened 2019-11-27
Add Path Transport Extension in go

### LINKED ISSUE #3405: Implement Path Transport Extension in Go
Implement the Path Transport Extension, as described in the book (section 15.1.4, p. 351)

> The path transport extension is an end-to-end extension that allows communicating end hosts to pass or update a path. It is designed to transport a forwarding path (i.e., a path in the data-plane format, see Section 15.1.3). The transported path can be accompanied with a SCION address, to indicate that the end host can be reached via another address and path to support host mobility.  [...]. 
> 
> The path transport extension that conveys a path in the data-plane format (set of info (INF) and hop (HF) fields) is depicted in Figure 15.8.  The extension contains the field SrcType, which indicates what type of SCION address the path is prepended with.  A source can inform a destination that the source is reachable through another address via the SrcAddr field (ISD and AS identifiers concatenated with a host address). If necessary, padding is added so that the full path is aligned to an 8-byte boundary (Section 15.1.3 describes details of the data-plane path format).  The Path field is not mandatory and an empty path can be sent, and through such a construction a source can indicate a new SCION address

matzf:
This implements the path transport extension, as described in the book. See the documentation in `path_trans_extn.go` for a description of the extension header format.

The extension header format is incompatible with the existing python implementation in python/lib/packet/ext/path_transport.py, in particular because the python version also allows to transport beaconing paths.

Resolves #3405 

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3435)


matzf:
Completely outdated by now (format for extensions has changed with the new header format, implementation has moved to gopacket layers).

PR #3444, opened 2019-12-02
sciond: do not use snet.Addr in config

karampok:
sciond config needs only the listening IP address.
The IA is being taken from topo `itopo.Get().IA()`

This commit introduces a breaking change in the sciond config.
The public now should be of `<ip>:<port>`, not `ia,<ip>:port`

This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3444)

### DIFF
--- a/go/sciond/internal/config/BUILD.bazel
+++ b/go/sciond/internal/config/BUILD.bazel
@@ -16,7 +16,6 @@ go_library(
         "//go/lib/pathstorage:go_default_library",
         "//go/lib/sciond:go_default_library",
         "//go/lib/serrors:go_default_library",
-        "//go/lib/snet:go_default_library",
         "//go/lib/truststorage:go_default_library",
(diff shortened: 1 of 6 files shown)


scrye:
___
*[go/sciond/internal/config/sample.go, line 36 at r1](https://reviewable.io/reviews/scionproto/scion/3444#-LvEgH5k3mJrKS3RQVrL:-LvFrfnX7gpaXLlY7cpx:b-pp1by6) ([raw file](https://github.com/scionproto/scion/blob/8e7802e33c4e3ac3ee944390c1e0933d6a0f9028/go/sciond/internal/config/sample.go#L36)):*

Yeah that's true. But I don't know what the correct wording should be. Maybe ask @scrye 
</blockquote></details>

It's probably the "Local" word that's misleading. Maybe reword to:
```
Listening address to register with the local dispatcher 
```
?

PR #3465, opened 2019-12-06
wip bazel ci

lukedirtwalker:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3465)


lukedirtwalker:
We already have the new pipeline since #3505

PR #3560, opened 2019-12-27
BS: Remove revocation storage & handling

lukedirtwalker:
Removes the revocation DB in the BS and all associated code.

Having revocations affecting the beaconing doesn't have a clear cut advantage,
but brings quite a large maintenance and complexity overhead.

It protects against propagating/registering a full set of revoked segments.
But on the other hand it could also be that it doesn't propagate/register segments
that could be useful, for example if an upstream two hops away has a revocation,
but we only want to communicate to an AS which is connected via the upstream one hop away,
we could use the revoked segment (see also https://github.com/scionproto/scion/issues/2534).
Furthermore respecting revocations only helps if they are there at the exact point in time when
the propagation/registration happens, if the revocations happen just a moment after
propagation/registration you could still have a full set of revoked segments.

If the full set of beacons is ever revoked it indicates that either 
1. the beacon selection policy is not diverse enough, e.g. there is a single point of failure because of bad selection
2. the candidate set or the propagation/registration set is not large enough (large diversity but still all revoked)
3. the actual network is not diverse enough / has too many failures.

This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3560)


Oncilla:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/3560#-Ly9VsAa9TX9jFJVh9y8:-Ly9VsAa9TX9jFJVh9y9:b-86ivj3) (no related file):*
Also remove revocation forwarding from the border router to the beacon server:
go/border/rpkt/process.go:L194

---

PR #3570, opened 2020-01-03
SPKI: Add sample configuration files.

Oncilla:
This change adds support for outputting sample configuration files.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3570)


Oncilla:
Obsolete since we pivot to a different cp pki approach

PR #3571, opened 2020-01-06
Remove path resolution step from snet writes

karampok:
Resolving paths in snet writes requires complex interactions between contexts and deadlines. However, most SCION applications never used this feature, instead opting to manage paths themselves (see CS, PS, SIG). Removing path resolution from writes makes snet simpler to reason about and maintain.

This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3571)


karampok:
@matzf  yes after this PR, user must always complete the path. Any objection use case for that?

PR #3588, opened 2020-01-08
SPKI: Update README

Oncilla:
Update the readme to cover the new scion-pki tool implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3588)


Oncilla:
Obsolete since we pivot to a different cp pki approach

PR #3592, opened 2020-01-08
Add support for log compression

scrye:
Fixes https://github.com/scionproto/scion/issues/1924.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3592)


scrye:
___
*[go/lib/log/log.go, line 57 at r2](https://reviewable.io/reviews/scionproto/scion/3592#-Ly8G5j99HeKqoGb9kqe:-Ly8jngT7hFhEC3K5Wap:b-u8iwn9) ([raw file](https://github.com/scionproto/scion/blob/ffbe28d8637ab19299006ecec8b95ce20766bdc3/go/lib/log/log.go#L57)):*

more than three arguments, should be a struct.  But not important.
Another golang stylish comment
`func SetupLogFile( name, dir, level string, size,age int)`
</blockquote></details>

This is something I was also thinking about; the refactor was unnecessarily painful due to this not taking a struct.

Added to https://github.com/scionproto/scion/issues/3594.

scrye:
___
*[go/lib/log/flags.go, line 31 at r2](https://reviewable.io/reviews/scionproto/scion/3592#-Ly8FminDz8vuko6VS8r:-Ly8jm6aCndvEy58vMIU:bxuvpp0) ([raw file](https://github.com/scionproto/scion/blob/ffbe28d8637ab19299006ecec8b95ce20766bdc3/go/lib/log/flags.go#L31)):*

having inside log package a variable starting with logX seems redundant. 
(log.logX)
</blockquote></details>

This is a good point, added to https://github.com/scionproto/scion/issues/3594.

PR #3611, opened 2020-01-15
Fix metric names

scrye:
This reverts the metric changes accidentally introduced by
https://github.com/scionproto/scion/pull/3590.

Fixes https://github.com/scionproto/scion/issues/3610.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3611)

### DIFF
--- a/go/beacon_srv/BUILD.bazel
+++ b/go/beacon_srv/BUILD.bazel
@@ -14,6 +14,7 @@ go_library(
         "//go/cs/config:go_default_library",
         "//go/cs/ifstate:go_default_library",
         "//go/cs/keepalive:go_default_library",
+        "//go/cs/metrics:go_default_library",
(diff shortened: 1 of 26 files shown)


scrye:
___
*[go/cs/main.go, line 112 at r2](https://reviewable.io/reviews/scionproto/scion/3611#-Lyd_cSeEEe-fCwBR3jf:-LydcLCU5WiQtEdZnewU:bnpc8o0) ([raw file](https://github.com/scionproto/scion/blob/f0ec0a10f700e10c2aba675d3f92b8b40ff9b9c3/go/cs/main.go#L112)):*

Maybe it would make sense to have a `InitMetrics` function
</blockquote></details>

I think it's a bit more transparent this way; I expect the code around application initialization to get simplified in the near future, so we can revisit this soon.

PR #3619, opened 2020-01-20
cert renewal: Implement renewal signed request

karampok:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3619)


karampok:
won't be fixed because we moved to v509

PR #3642, opened 2020-01-29
log: rename LogPanicAndExit to HandlePanic

karampok:
According to the linter
```
func name will be used as log.LogPanicAndExit by other packages, and that stutters; consider calling this PanicAndExit

```

Contributes #3592

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3642)

### DIFF
--- a/go/acceptance/sig_ping_acceptance/main.go
+++ b/go/acceptance/sig_ping_acceptance/main.go
@@ -45,7 +45,7 @@ func realMain() int {
 		fmt.Fprintf(os.Stderr, "Failed to init: %s\n", err)
 		return 1
 	}
-	defer log.LogPanicAndExit()
+	defer log.HandlePanic()
 	defer log.Flush()
 	if !*integration.Docker {
 		log.Crit(fmt.Sprintf("Can only run %s test with docker!", name))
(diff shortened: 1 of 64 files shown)


scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/3642#-LzpHt0i4_m3rjyhmSHr:-LzpHt0i4_m3rjyhmSHs:b-hkdz7d) (no related file):*
It feels like both the old and the new name are misleading.

Regarding the old one, at the call site:
```
defer log.LogPanicAndExit()
```
the reader would always think "Why am I logging a panic, I might not have one?", "Why am I exiting the application after finishing this goroutine?". Of course, this isn't what happens, but _this_ is what the naming suggested.

Regarding the new one, at the call site:
```
defer log.PanicAndExit()
```
similar questions pop up. "Why am I panicking when finishing this goroutine?", "Why am I exiting?".

I think it's better to not hint at what is happening. So something like:
```
defer log.HandlePanic()
```

It's more vague ("_How_ is it handled?"), but vague is better than misleading.

It will also free us from the name implying any contracts at the call site, so we can now plug handlers that do not kill a unit test when a goroutine in there panics (which is almost never what the programmer wants, since it black holes valuable test output).

---

scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/3642#-M0SaC8X0ee0oM2ul4AB:-M0SaC8ZBCGl3WvHylyv:bodvqfh) (no related file):*
When merging, please make sure the final commit message refers to `log.HandlePanic`, and not the older message.

---

PR #3645, opened 2020-01-29
BS: Drop convey from tests

Oncilla:
This PR drops all goconvey imports from go/cs/beacon*

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3645)


Oncilla:
Addressed by #3652

PR #3648, opened 2020-01-30
Fix snet.UDPAddress parsing & serialization

### LINKED ISSUE #3647: IPv4 addresses enclosed in brackets against RFC3986
Currently toml files use the following syntax for addresses of control services

```
Address = "[242.254.0.22]:31005"
```

According to RFC3986 this syntax is reserved for IPv6 and futures

```
A host identified by an Internet Protocol literal address, version 6
   [RFC3513] or later, is distinguished by enclosing the IP literal
   within square brackets ("[" and "]").  This is the only place where
   square bracket characters are allowed in the URI syntax.
```

The current behaviour is inconsistent as toml files also contain IPs without enclosing brackets, like `agent = "172.17.0.1:6831"`.

Probably the following regex (https://github.com/scionproto/scion/blob/master/go/lib/snet/udpaddr.go#L29-L30) should be updated to handle this more properly

```
var addrRegexp = regexp.MustCompile(
    `^(?P<ia>\d+-[\d:A-Fa-f]+),\[(?P<host>[^\]]+)\](?P<port>:\d+)?$`)
```

lukedirtwalker:
Allow <IPv4>:port without [].
Use net library functions instead of manual parsing.

Fixes #3647

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3648)

### DIFF
--- a/go/lib/snet/udpaddr.go
+++ b/go/lib/snet/udpaddr.go
@@ -26,8 +26,7 @@ import (
 	"github.com/scionproto/scion/go/lib/spath"
 )
 
-var addrRegexp = regexp.MustCompile(
-	`^(?P<ia>\d+-[\d:A-Fa-f]+),\[(?P<host>[^\]]+)\](?P<port>:\d+)?$`)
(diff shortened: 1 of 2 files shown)


lukedirtwalker:
Closing in favor of #3650

PR #3674, opened 2020-02-23
Fixed HBH ext parse panic when start >= end

AlexanderKunze:
The router could crash if a packet arrived with data resulting in the function `go/border/rpkt/extn_scmp.go/rSCMPExtFromRaw(...)` receiving a `start >= end`, which would panic on line 38.

The error is caused by `go/border/rpkt/parse.go/rp.parseHopExtns()` calling `rp.extnParseHBH(...)` on line 123, which then calls `rSCMPExtFromRaw(...)` with no checks for `start` or `end`.

I decided to add an error check in `rp.extnParseHBH(...)` instead of where the panic happens (`rSCMPExtFromRaw(...)`), because a header of invalid offset or length should fail, even when `start` or `end` are not read, such as in `rOneHopPathFromRaw`.

The problem might also be fixed earlier, in `rp.parseHopExtns()`, where the erroneous values are actually read in. However, I was unsure if this is better and so left the error catching in `rp.extnParseHBH(...)`

I also chose to use as an error check that the total length of the extension must be greater than `common.LineLen`, since an extension of any less size should not be valid anyway.

For reference: The packet that caused the panic was:

`00c10067070405000001ff00000001110001ff000000011080007f00000b0000857b3362e3400140003f0000014152c1000000000000000000000000000000001101010000000000792c000000271f920000001b10084002110582510801027f10065002020102`

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3674)

### DIFF
--- a/go/border/rpkt/extns.go
+++ b/go/border/rpkt/extns.go
@@ -18,6 +18,7 @@
 package rpkt
 
 import (
+	"github.com/scionproto/scion/go/lib/assert"
 	"github.com/scionproto/scion/go/lib/common"
 	"github.com/scionproto/scion/go/lib/scmp"
 	"github.com/scionproto/scion/go/lib/serrors"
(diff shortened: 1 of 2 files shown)


AlexanderKunze:
Thank you @sgmonroy and @matzf for those excellent and extensive comments. I have made changes and issued a new commit.

The changes are as follows:

- extnParseHBH now has an assert that `end-start >= common.LineLen`.
- Following the comments here, I added an additional check that returns an error if `*offset+2 >= len(rp.Raw)`, to ensure that `hdrLen` and `currExtn` cannot panic when accessing `rp.Raw[*offset+2]` and `rp.Raw[*offset+1]`. For this case, I copied the error message from below that the Extension header chain is too long.
- I moved the check for `start >= end` to `rp.parseHopExtns()`, as suggested, since I believe this possible panic is not caught by the `*offset+2 >= len(rp.Raw)` check. The function call to `rp.extnParseHBH` was adjusted and two variables were added so as to avoid double calculation of `start` and `end`
- An additional error message was created called ErrExtChainCorrupt for the case in which the `ExtHdrLen` field is corrupt, even if the extension header chain is long enough.

Please let me know any other feedback you have or if I've misunderstood something, made a mistake, or if there is something else I can do.

AlexanderKunze:
Hi and thanks for the comments once again. Sorry it took so long to make these changes. I have issued the commit. Let me know if there's something else I can do.

AlexanderKunze:
Thanks for the correction, adjusted it accordingly.

PR #3706, opened 2020-04-08
Remove redundant select expression

ekuiter:
This select expression was introduced in #3325, but is no longer necessary and actively harmful when trying to compile SCION for Android (the build fails due to this expression, so we had to [remove it](https://github.com/ekuiter/scion-android-binary/blob/60e97b837e73ec25fd3a3900e09f5488c0f4235d/build.sh#L71) manually). Semantically, it does not even make a difference, as in both selected cases the same dependencies are returned.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3706)


ekuiter:
But why is this file included with the SCION sources at all, if it is generated by gazelle after all?

lukedirtwalker:
We include all generated files, so that it's easy for users to just run everything without first requiring to generate.

PR #3754, opened 2020-05-07
ci: try yml anchors

karampok:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3754)


karampok:
For the time being, we defer using anchors for simplicity.

Issue #3731, opened 2020-04-28
New Path Registration and Path Lookup strategy for Down Segments


shitz:
Currently, non-core ASes register their down-segments with a single core AS only (not specified which). Core ASes run a syncing protocol to exchange registered down-segments amongst them. The property _intended_ to be achieved is: each core AS of an ISD has _all_ down-segments for that ISD. The advantage of this property is that during path lookup, any requester, ISD-local and ISD-remote, can ask _any_ core AS of an ISD about _any_ down-segment. However, this approach also has some short-comings:
1. The segment syncing protocol is highly complex. In essence, it needs to implement a distributed consistent DB for down-segments _across multiple organization_.
1. It's unclear how useful it is for a core AS to return down-segments that originate in a different core-AS, since it cannot be combined with core-segments that terminate in the replying core AS.

We thus propose a change to path registration (for down-segments) and path lookup:
1. During registration, a non-core AS only registers down-segments with the core AS they originate from.
1. There is no segment synchronization protocol between core ASes. Each core AS only has the down-segments for which it received an explicit registration.
1. During path lookup, a requester asks _each_ core AS for which it wants to include down-segments to a non-core AS. By default, a requester asks _all_ core ASes of an ISD.

The trade-offs are the following:
* We get rid of the highly complex and brittle down-segment synchronization (big win)
* The answer of each core AS is guaranteed to be useful, since its existence implies also the existence of a core-segment that can be used to connect with the returned down-segments.
* Path lookup gets slightly more complex, since it now needs to go to multiple core ASes.

We are convinced that the advantages far outweigh the disadvantages.

lukedirtwalker:
This is documented in #3747 and implemented in #3792

PR #3765, opened 2020-05-13
Remove obsolete limit on path segments in SegReply

matzf:
An arbitrary limit to 9 path segments in a SegReply was implemented some time ago as a hack to avoid exceeding MTU. This is no longer necessary as QUIC is now used as transport for these path segment requests.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3765)


matzf:
Note: I noticed this while working on the implementation of  #3731; in this context, the code modified here will be removed anyway, but it seemed to make sense to clean this up independently.

PR #3790, opened 2020-06-17
Clean debian dependencies, misc doc & bazel cleanup

### LINKED ISSUE #3788: Pop!_OS/Ubuntu 20.04 compatibility - `cgroup-bin` has no installation candidate
```
pascal@pop-os:~/scion$ ./env/deps 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package cgroup-bin is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  cgroup-tools

E: Package 'cgroup-bin' has no installation candidate
```

Potential Fix: change
`env/debian/pkgs.txt:3:cgroup-bin` to `cgroup-tools`

-----------------------------------------

New Problem: It wants to uninstall `systemd-timesyncd`
```
pascal@pop-os:~/scion$ ./env/deps 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following additional packages will be installed:
  libcapnp-0.7.0 libexpat1-dev libio-pty-perl libipc-run-perl libjq1 libonig5 libopts25 libpython3-dev libpython3.8-dev libtime-duration-perl python-pip-whl python3.8-dev
Suggested packages:
  ntp-doc python-nose-doc
Recommended packages:
  sntp
The following packages will be REMOVED
  systemd-timesyncd
The following NEW packages will be installed
  capnproto jq libcapnp-0.7.0 libcapnp-dev libexpat1-dev libio-pty-perl libipc-run-perl libjq1 libonig5 libopts25 libpython3-dev libpython3.8-dev libtime-duration-perl moreutils ntp python-pip-whl python3-dev python3-nose python3-pip python3-wheel
  python3.8-dev uthash-dev
0 to upgrade, 22 to newly install, 1 to remove and 0 not to upgrade.
E: Packages need to be removed but remove is disabled.
```
AFAIK, Gnome has dependencies on that package which means it can't just be uninstalled

OS Used:
```
Operating System: Pop!_OS 20.04 LTS (based on Ubuntu 20.04 LTS)
Kernel: Linux 5.4.0-7634-generic
Architecture: x86-64
```

matzf:
- Cleanup of debian/pkgs.txt (see below). 
  Resolves #3788
- Update docs to refer to installation of bazel version 3.2.0
- Update docs to clarify that docker is now required
  With the changes for x509 certificates (#3784), scionpki now requires docker to create the
  test topology.
- Actually use `pkg_tar` from `rules_pkg` loaded in `WORKSPACE`, instead of built-in.
  Seemed related to an issue I was debugging (it wasn't). Probably makes no difference, but it seems that this is intended.

Annotated changes of debian/pkgs.txt

```
  bridge-utils       # needed in docker (?)
  capnproto          # used during build for compiling capnproto files
- cgroup-bin         # needed in docker (?), replaced (was transitional package)
+ cgroup-tools
  curl               # used by install_docker/install_bazel
- dnsutils           # not needed (?)
- file               # no longer needed, was added for pycapnp
  g++                # used by bazel (to compile protobuf)
  gcc                # used by bazel's C toolchain 
  git
  jq                 # used in various tests
- libcapnp-dev       # no longer needed, was added for pycapnp
  make               # used to build
  moreutils          # used in various tests (sponge)
  net-tools
- netcat-openbsd     # no longer needed, was used for waiting for zookeper
- ntp                # was added for better time sync, not needed for development
- patch              # never used?
- python3-dev        # indirect, comes with python3-pip
  python3-nose       # used for python tests (could also be installed with pip)
  python3-pip        # used to installing python dependencies
+ python3-setuptools # needed for pip, previously missing (was broken!?)
  python3-wheel      #  needed for for pip
  software-properties-common  # used by install_docker
  sqlite3            # used in acceptance test
  sudo
  tzdata
- uthash-dev         # no longer needed, was once used in `SCIONWrapper.cpp`
- vim-common         # not needed, was just for convenience?
```

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3790)


pascalwacker:
Since Docker is a requirement, is it still possible to have the dev env itself run in a Docker container? If so, would one need [DinD](https://www.docker.com/blog/docker-can-now-run-within-docker/) or would it be enough to simply mount the docker socket as ex. [portainer.io](https://portainer.io) does?

PR #3799, opened 2020-06-23
Add git and python dependencies

pascalwacker:
Python 2 is needed during the run of `./scion.sh topology` (could potentially be moved to the deps file) and Git is needed to clone the repo

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3799)

### DIFF
--- a/doc/build/setup.rst
+++ b/doc/build/setup.rst
@@ -10,7 +10,7 @@ Setting up the development environment
 
    .. code-block:: bash
 
-      sudo apt-get install g++ unzip zip
+      sudo apt-get install g++ unzip zip git python
       wget https://github.com/bazelbuild/bazel/releases/download/3.2.0/bazel-3.2.0-installer-linux-x86_64.sh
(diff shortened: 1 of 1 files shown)


Oncilla:
Hi @pascalwacker 

Thanks for the contribution.

We only require python3 for the topology generator, did something fail without python2 installed?

As for git requirement, it is a basic assumption that it is installed, or how do you clone the project.
Can you maybe elaborate on the workflow that sparked this?

matzf:
The python thing comes up during the bazel build as part of the `pkg_tar` rules. The actual packaging script is python3 only. However, bazel executes it via a wrapper script which is python version agnostic but requires a `python` on the path.

If no python2 is installed, it is enough to setup a symlink `python -> python3`, e.g. by running `sudo update-alternatives --install /usr/bin/python python /usr/bin/python3 1`.

lukedirtwalker:
Since the PR is outdated and is missing activity for a long time I take the freedom to close it.
If you think it's still an issue please re-open or create an up to date PR. Thanks.

PR #3779, opened 2020-06-08
Extending Showpaths With Static Info

NE1997:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3779)

PR #3839, opened 2020-07-28
snet: move MTU/Expiry to separate snet.PathMetadata interface

matzf:
Add additional interface for (optional) path metadata on snet.Path.
This new snet.PathMetadata interface will be extended to add support for
the information in the StaticInfo beacon extension (Geo information,
Latency, Bandwidth, etc) as a next step.

Having a separate interface for the metadata clarifies that some
instances of snet.Path (e.g. the `partialPath` constructed from raw
forwarding information) do not have any of this metadata available.
Thus, moving the existing MTU and Expiry fields to this new interface
feels quite natural.

Additionally, remove `Fingerprint` from the `snet.Path` interface.
`Fingerprint` can be implemented in terms of snet.Path, so it does not
need to be part of the interface. Make this a simple helper function 
in `snet` instead.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3839)

### DIFF
--- a/acceptance/sig_failover/testdata/1-ff00_0_110/sig/fake_sciond.json
+++ b/acceptance/sig_failover/testdata/1-ff00_0_110/sig/fake_sciond.json
@@ -4,15 +4,31 @@
             "reply_start_timestamp": 0,
             "paths": [
                 {
-                    "fingerprint": "PathAAAA",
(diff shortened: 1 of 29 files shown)


matzf:
Related to #3779.

PR #3843, opened 2020-07-29
colibri: add store and segment admission

juagargi:
Introduces the COLIBRI store.
Adds the segment admission handling to the store.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)


Oncilla (changes requested):
> Reviewed 1 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 7 of 38 files reviewed, 5 unresolved discussions (waiting on @juagargi and @Oncilla)

___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 85 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MEwoIE-8Ya5Ougtmx3f:-MFUmnhG7o5YIEqMvz5j:bun6ke7) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L85)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

Note that the `free` variable contains the free bandwidth already discounting that of the request. The function `sumMaxBlockedBW` already excludes it. I have added comments to clarify.
</blockquote></details>

ah, okay. I missed that.
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 163 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MEwvaEG2QxpfnfUXH-x:-MFUnaA3Bg46hu7qjW6r:b6klz2j) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L163)):*
The intuition behind srcAlloc is the following:

> The source allocation computes how much bandwidth is actually reserved from the
ingress to egress interface for a given source.

If the source equals the source of the request, we have to add prevBW, because of:
>  Again, the reservation of the current request is treated, as if there is no previous index.

The formula should be read as:

`srcAlloc(s, in, eg, req) = (sum_all_rsv_except_rsv_id(s. in, eg, req)) + (req.src == s ? req.prevBW:0) `

I hope that makes sense
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 169 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MESZ4ar8F8kkGME7xHj:-MFUok3d38myvUMGphDe:bz1dnl2) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L169)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

I am very unsure of my understanding of `linkRatio`. The following clarifies what I believe `linkRatio` is, and how it's computed. 

The `if` condition prevents dividing by zero, among other things. I think it is better explained if you comment the whole `if` out, and run the tests. Among others, the test for `TestLinkRatio/empty` will fail, because we never saw any entry for the denominator.

This:
```go
// shortcut: no other appearances of the request's source in the DB. The ratio is 1
```
Makes me realize that I indeed didn't understand 3.23 correctly: it sums on all reservations with a given source, entering and exiting with given interfaces. And then it adds `prevBw` or zero, depending on the request's source.
I will adapt this and the above lines with this understanding. I leave both this and the previous messages for completeness, and because I still believe that the loop you propose in the previous comment is not doing what 3.23 says, so we can discuss this further.
</blockquote></details>

Right, I hope the following helps you get an intuition what link ratio is about:

When we compute the ideal bandwidth, we aggregate all requested bandwidths between interface pairs in so called "tubes".
I.e., all requests with the same ingress and egress pair are part of the same "tube."

All tubes with the same egress interface compete for the bandwidth on that interface. We attempt to calculate a fair split of the bandwidth on the egress interface, called the tubeRatio.
When we multiple all reservations of a given tube with that tubes tubeRatio -> we get a fair split among the tubes.

Now, if we simply split the tubes between the sources, a source AS can gain an advantage by entering the AS in multiple interfaces, and take a cut of multiple tubes.

To prevent this, we want to adjust the bandwidth a source AS can get on a specific egress link. The link ratio does this. The link ratio computes how much bandwidth of a given egress link the reservation should get, such that the split between the source ASes is fair according to their demand.

I hope this gives you a bit more intuition.
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 213 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MEwrn3YCVWeVEzIFYSt:-MFUsiLe3NlbHU23eTz7:b1dnpzs) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L213)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

I thought about that, but it requires that the request is embedded in the `rsv` collection that comes from the DB --and which not always contains the request--.
I have nevertheless altered a bit the loop to make clear that if there is a reservation with the ID of the request, it is ignored.
</blockquote></details>

oh, yeah. very good point. An initial request would not be part of the DB :thumbsup: 
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 277 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MEwmp460AGIbfj-KGdP:-MFUt4-15crgwHEPJ8yl:b65gtah) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L277)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

`reservation.SegmentID` is a structure; wouldn't passing a value incur in some sort of overhead, or is the compiler smart enough to optimize this? (I am unfamiliar with the optimizer).
FWIW I also agree with passing it per value.
</blockquote></details>

It depends.
For structs that are very small, it is cheaper to copy, than forcing the struct to escape to heap.
Could also be that this function call is in-lined, and there is no difference.

But without benchmarks + running heap escape analysis it is hard to guess which is faster.

We tend to pass structs since it is "safer", as you cannot modify the structs contents, and only worry about passing pointers for performance if it actually makes a difference.
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 41 at r2](https://reviewable.io/reviews/scionproto/scion/3843#-MFUknBU82SYLhXO1JsK:-MFUknBU82SYLhXO1JsL:b-28a7qq) ([raw file](https://github.com/scionproto/scion/blob/aaec7f6d0057ed95b54aa5be2528dfa8c1e16b9a/go/cs/reservation/segment/admission/impl/stateless.go#L41)):*
> ```Go
> 
> // AdmitRsv admits a segment reservation. The request will be modified with the allowed and
> // maximum bandwidths, also when returning error.
> ```

that means it needs to be added on every return path, no?
Currently that does not happen.

I'm not sure if the binary return is enough.
There are 3 results, afaict:
1. happy
2. computation succeeded, no bandwidth to spare.
3. computation failed.

Are cases 2. and 3, eventually, handled differently by calling code?
If so, I think it makes sense to have: `AdmissionResult, error` as the return from this method.
And let the caller create the beads if appropriate.

If cases 2 and 3 are not handled differently, binary result might suffice.

juagargi (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 7 of 38 files reviewed, 5 unresolved discussions (waiting on @Oncilla)

___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 163 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MEwvaEG2QxpfnfUXH-x:-MFVf52Z-DoZGEsdPJIL:bu2gm7m) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L163)):*

The intuition behind srcAlloc is the following:

> The source allocation computes how much bandwidth is actually reserved from the
ingress to egress interface for a given source.

If the source equals the source of the request, we have to add prevBW, because of:
>  Again, the reservation of the current request is treated, as if there is no previous index.

The formula should be read as:

`srcAlloc(s, in, eg, req) = (sum_all_rsv_except_rsv_id(s. in, eg, req)) + (req.src == s ? req.prevBW:0) `

I hope that makes sense


</blockquote></details>

Thanks @oncilla . I believe the code does what you describe: the variable `srcAllocPerSrc` in the code has all the values of `srcAlloc(s, in, eg, req)` for every possible value of `s`.  The addition of `prevBW` is done out of the `srcAllocPerSrc` computation and only when adding the values in the denominator.
Can you check if it is what you described above?
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 169 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MESZ4ar8F8kkGME7xHj:-MFVjsHfATxXMLuXFygj:box7phd) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L169)):*

Right, I hope the following helps you get an intuition what link ratio is about:

When we compute the ideal bandwidth, we aggregate all requested bandwidths between interface pairs in so called "tubes".
I.e., all requests with the same ingress and egress pair are part of the same "tube."

All tubes with the same egress interface compete for the bandwidth on that interface. We attempt to calculate a fair split of the bandwidth on the egress interface, called the tubeRatio.
When we multiple all reservations of a given tube with that tubes tubeRatio -> we get a fair split among the tubes.

Now, if we simply split the tubes between the sources, a source AS can gain an advantage by entering the AS in multiple interfaces, and take a cut of multiple tubes.

To prevent this, we want to adjust the bandwidth a source AS can get on a specific egress link. The link ratio does this. The link ratio computes how much bandwidth of a given egress link the reservation should get, such that the split between the source ASes is fair according to their demand.

I hope this gives you a bit more intuition.

















</blockquote></details>

Thanks @oncilla . It indeed clarifies things for me. I have added a couple of tests to the end of the `TestLinkRatio` with names `smaller prevBW` and `bigger prevBW`. Can you check for me if they are correct? 
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 213 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MEwrn3YCVWeVEzIFYSt:-MFVb0Ct3x_E5Ic4OlKi:b-ryvawc) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L213)):*

oh, yeah. very good point. An initial request would not be part of the DB :thumbsup: 
</blockquote></details>

Can you check this function and see if you are happy with it? I don't find it easy to simplify it without adding a new interface (essentially unifying the reservations from the DB and the request) and a new collection of objects with this interface from the DB + the request. But this will be much more code, and only used in this function.
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 277 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MEwmp460AGIbfj-KGdP:-MFVZOE66ElgWmPaLwxx:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L277)):*

It depends.
For structs that are very small, it is cheaper to copy, than forcing the struct to escape to heap.
Could also be that this function call is in-lined, and there is no difference.

But without benchmarks + running heap escape analysis it is hard to guess which is faster.

We tend to pass structs since it is "safer", as you cannot modify the structs contents, and only worry about passing pointers for performance if it actually makes a difference.


</blockquote></details>

Done.
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 41 at r2](https://reviewable.io/reviews/scionproto/scion/3843#-MFUknBU82SYLhXO1JsK:-MFVZcWeEIFijC0AUAda:b-nuvua0) ([raw file](https://github.com/scionproto/scion/blob/aaec7f6d0057ed95b54aa5be2528dfa8c1e16b9a/go/cs/reservation/segment/admission/impl/stateless.go#L41)):*

that means it needs to be added on every return path, no?
Currently that does not happen.

I'm not sure if the binary return is enough.
There are 3 results, afaict:
1. happy
2. computation succeeded, no bandwidth to spare.
3. computation failed.

Are cases 2. and 3, eventually, handled differently by calling code?
If so, I think it makes sense to have: `AdmissionResult, error` as the return from this method.
And let the caller create the beads if appropriate.

If cases 2 and 3 are not handled differently, binary result might suffice.




</blockquote></details>

Right now both cases 2. and 3. send a message backwards in the reservation chain. I.e. the "no bandwidth available" case is treated just as another error.

Oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 6 of 38 files reviewed, 2 unresolved discussions (waiting on @juagargi and @Oncilla)

___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 163 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MEwvaEG2QxpfnfUXH-x:-MFVz2kn9Meth7_3NudE:bmk3qdj) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L163)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

Thanks @oncilla . I believe the code does what you describe: the variable `srcAllocPerSrc` in the code has all the values of `srcAlloc(s, in, eg, req)` for every possible value of `s`.  The addition of `prevBW` is done out of the `srcAllocPerSrc` computation and only when adding the values in the denominator.
Can you check if it is what you described above?
</blockquote></details>

I think it is correct now. 
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 213 at r1](https://reviewable.io/reviews/scionproto/scion/3843#-MEwrn3YCVWeVEzIFYSt:-MFVv1gv07YDFq2rbQAD:b-w2x1bk) ([raw file](https://github.com/scionproto/scion/blob/0a857ad6e26ca59b0d0ffdb41c136f3e651cbc29/go/cs/reservation/segment/admission/impl/stateless.go#L213)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

Can you check this function and see if you are happy with it? I don't find it easy to simplify it without adding a new interface (essentially unifying the reservations from the DB and the request) and a new collection of objects with this interface from the DB + the request. But this will be much more code, and only used in this function.
</blockquote></details>

Yes, I think it is correct like it is now.
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 41 at r2](https://reviewable.io/reviews/scionproto/scion/3843#-MFUknBU82SYLhXO1JsK:-MFVyA7F3B_ojbufHjGr:b-db7n2o) ([raw file](https://github.com/scionproto/scion/blob/aaec7f6d0057ed95b54aa5be2528dfa8c1e16b9a/go/cs/reservation/segment/admission/impl/stateless.go#L41)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

Right now both cases 2. and 3. send a message backwards in the reservation chain. I.e. the "no bandwidth available" case is treated just as another error.
</blockquote></details>

okay, but in any case, this method does not add the beads in at least 2 execution paths.
___
*[go/cs/reservation/segment/admission/impl/stateless_test.go, line 384 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFVzrp9Bzbi90_bQYYi:-MFVzrp9Bzbi90_bQYYj:b-fgzesz) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/segment/admission/impl/stateless_test.go#L384)):*
> ```Go
> 			},
> 		},
> 		"smaller prevBW": {
> ```

That looks good now

Oncilla (changes requested):
> Reviewed 3 of 38 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 9 of 38 files reviewed, 4 unresolved discussions (waiting on @juagargi and @Oncilla)

___
*[go/cs/reservation/conf/capacities.go, line 29 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW-ZYe0HSfXHHyXhJl:-MFW-ZYe0HSfXHHyXhJm:bcel4do) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/capacities.go#L29)):*
> ```Go
> type capacities struct {
> 	// ingress capacities
> 	CapIn map[common.IFIDType]uint64 `json:"ingress_KBps"`
> ```

do we want to encode the unit in the name?
Also, KBps is a weird unit. In networking, we tend to use bps, so this would be kbps.
___
*[go/cs/reservation/conf/capacities.go, line 33 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW-t2F6HDU2FR30a1N:-MFW-t2F6HDU2FR30a1O:b-p5sbs1) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/capacities.go#L33)):*
> ```Go
> 	CapEg map[common.IFIDType]uint64 `json:"egress_KBps"`
> 	// configured allowed transit
> 	In2Eg map[common.IFIDType]map[common.IFIDType]uint64 `json:"in_to_eg_KBps"`
> ```

`ingress_to_egress` or `ingress_to_egress_kbps`
___
*[go/cs/reservation/conf/capacities.go, line 75 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW0BHYEn4SYLX1WOLm:-MFW0BHYEn4SYLX1WOLn:b4ux5ty) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/capacities.go#L75)):*
> ```Go
> 			totalEgress[egress] += cap
> 			if egress == ingress && cap != 0 {
> 				return serrors.New("capacity is inconsistent, ingress to itself not zero",
> ```

I like the error messages. :thumbsup:

Oncilla (changes requested):
> Reviewed 11 of 38 files at r1, 3 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 23 of 38 files reviewed, 7 unresolved discussions (waiting on @juagargi and @Oncilla)

___
*[go/cs/reservation/conf/capacities_test.go, line 115 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW0cV4BEnkUDCnrLGR:-MFW0cV4BEnkUDCnrLGS:bh4im5w) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/capacities_test.go#L115)):*
> ```Go
> 			},
> 		},
> 		"ingress to itself (reflexion not allowed)": {
> ```

I think that should be `reflection`
___
*[go/cs/reservation/conf/testdata/caps1.json, line 26 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW0izOCUQheGDwfKOe:-MFW0izOCUQheGDwfKOf:b-yigpbu) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/testdata/caps1.json#L26)):*
> ```JSON
>     }
>   }
> }
> ```

new line at the end
___
*[go/cs/reservation/conf/testdata/caps2.json, line 7 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW0lKD8Vhz89VNVJ-2:-MFW0lKD8Vhz89VNVJ-3:b-7ropln) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/testdata/caps2.json#L7)):*
> ```JSON
>   "egress_KBps": null,
>   "in_to_eg_KBps": null
> }
> ```

ditto

Oncilla (commented):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 24 of 38 files reviewed, 7 unresolved discussions (waiting on @juagargi and @Oncilla)

Oncilla (commented):
> Reviewed 3 of 38 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 27 of 38 files reviewed, 7 unresolved discussions (waiting on @juagargi and @Oncilla)

juagargi (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 23 of 38 files reviewed, 7 unresolved discussions (waiting on @juagargi and @Oncilla)

___
*[go/cs/reservation/conf/capacities_test.go, line 115 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW0cV4BEnkUDCnrLGR:-MFWr3qaCEHP9XAN8GiF:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/capacities_test.go#L115)):*

I think that should be `reflection`
</blockquote></details>

Done.
___
*[go/cs/reservation/conf/testdata/caps1.json, line 26 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW0izOCUQheGDwfKOe:-MFWv3VpCIdTSEmgpAWg:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/testdata/caps1.json#L26)):*

new line at the end
</blockquote></details>

Done.
___
*[go/cs/reservation/conf/testdata/caps2.json, line 7 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW0lKD8Vhz89VNVJ-2:-MFWv4KK9prqpLw5QezY:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/testdata/caps2.json#L7)):*

ditto
</blockquote></details>

Done.

juagargi (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 23 of 38 files reviewed, 7 unresolved discussions (waiting on @juagargi and @Oncilla)

___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 41 at r2](https://reviewable.io/reviews/scionproto/scion/3843#-MFUknBU82SYLhXO1JsK:-MFWzwPU6eADH9pbBm0X:b-38cxln) ([raw file](https://github.com/scionproto/scion/blob/aaec7f6d0057ed95b54aa5be2528dfa8c1e16b9a/go/cs/reservation/segment/admission/impl/stateless.go#L41)):*

okay, but in any case, this method does not add the beads in at least 2 execution paths.
</blockquote></details>

That's correct. Even though there is no explicit error code, the response message is sent without the beads, meaning "all the information the failing AS gives is that it failed".
We can flag this case differently or even add an explicit error code. Although in the latter case I don't know how useful it would be for the previous ASes and source to receive those codes stating e.g. "the failing AS could not retrieve all the reservations from the DB".

juagargi (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 23 of 38 files reviewed, 7 unresolved discussions (waiting on @Oncilla)

___
*[go/cs/reservation/conf/capacities.go, line 29 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW-ZYe0HSfXHHyXhJl:-MFZw_wJ1aHxq9AGk6Df:bjgwg6h) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/capacities.go#L29)):*

do we want to encode the unit in the name?
Also, KBps is a weird unit. In networking, we tend to use bps, so this would be kbps.
</blockquote></details>

Done. Also note that `go/lib/colibri/reservation/types.go` has changed, to have a consistent `BWCls.ToKbps()` name.
___
*[go/cs/reservation/conf/capacities.go, line 33 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW-t2F6HDU2FR30a1N:-MFZx-D8Bb9-qdsvpP1V:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/capacities.go#L33)):*

`ingress_to_egress` or `ingress_to_egress_kbps`
</blockquote></details>

Done.

Oncilla (changes requested):
> Reviewed 6 of 38 files at r1, 2 of 7 files at r2, 9 of 10 files at r4.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 37 of 38 files reviewed, 4 unresolved discussions (waiting on @juagargi and @Oncilla)

___
*[go/cs/reservation/conf/capacities.go, line 64 at r4](https://reviewable.io/reviews/scionproto/scion/3843#-MFin5wfBaI6REr3A88d:-MFin5wfBaI6REr3A88e:b-iovs3n) ([raw file](https://github.com/scionproto/scion/blob/fc937b11e513b631d9a89497ac026c725349b2f6/go/cs/reservation/conf/capacities.go#L64)):*
> ```Go
> // MarshalJSON serializes the internal json-friendly structure.
> func (c *Capacities) MarshalJSON() ([]byte, error) {
> 	b, err := json.MarshalIndent(c.c, "", "  ")
> ```

I am a bit confused by this. why do you do marshal indent anyway.
It should be up to the caller how to format this, no.

even if you call `json.Marshal` here, if the caller calls `json.MarshalIndent` it will be indented correctly
Actually, it will override whatever your specify here:

https://play.golang.org/p/q_AgE6HNbKz
___
*[go/cs/reservation/conf/testdata/caps1.json, line 26 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW0izOCUQheGDwfKOe:-MFip7HF8OXUU7TeogMh:b-d8cawy) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/testdata/caps1.json#L26)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Ha, so the MarshalIndent in capacities is responsible for this.

Where do you generate this from?
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 41 at r2](https://reviewable.io/reviews/scionproto/scion/3843#-MFUknBU82SYLhXO1JsK:-MFipa2XD0TotkEwaSgI:bsa6n5q) ([raw file](https://github.com/scionproto/scion/blob/aaec7f6d0057ed95b54aa5be2528dfa8c1e16b9a/go/cs/reservation/segment/admission/impl/stateless.go#L41)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

That's correct. Even though there is no explicit error code, the response message is sent without the beads, meaning "all the information the failing AS gives is that it failed".
We can flag this case differently or even add an explicit error code. Although in the latter case I don't know how useful it would be for the previous ASes and source to receive those codes stating e.g. "the failing AS could not retrieve all the reservations from the DB".
</blockquote></details>

This might be valid. I'm complaining about the comment being misleading.
___
*[go/lib/colibri/reservation/types.go, line 85 at r4](https://reviewable.io/reviews/scionproto/scion/3843#-MFiq6Oo2Qm3u9Kp5eKZ:-MFiq6Oo2Qm3u9Kp5eK_:b7vl8w6) ([raw file](https://github.com/scionproto/scion/blob/fc937b11e513b631d9a89497ac026c725349b2f6/go/lib/colibri/reservation/types.go#L85)):*
> ```Go
> 
> func (id *SegmentID) String() string {
> 	return id.ASID.String() + "-" + hex.EncodeToString(id.Suffix[:])
> ```

could be simplified to `fmt.Sprintf("%s-%x", id.ASID, id.Suffix)`

juagargi (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843)* status: 37 of 38 files reviewed, 4 unresolved discussions (waiting on @Oncilla)

___
*[go/cs/reservation/conf/capacities.go, line 64 at r4](https://reviewable.io/reviews/scionproto/scion/3843#-MFin5wfBaI6REr3A88d:-MFiymX48oTV1TiNJvaM:b-1suku2) ([raw file](https://github.com/scionproto/scion/blob/fc937b11e513b631d9a89497ac026c725349b2f6/go/cs/reservation/conf/capacities.go#L64)):*

I am a bit confused by this. why do you do marshal indent anyway.
It should be up to the caller how to format this, no.

even if you call `json.Marshal` here, if the caller calls `json.MarshalIndent` it will be indented correctly
Actually, it will override whatever your specify here:

https://play.golang.org/p/q_AgE6HNbKz



</blockquote></details>

Done. My main problem was having a pointer receiver.
___
*[go/cs/reservation/conf/testdata/caps1.json, line 26 at r3](https://reviewable.io/reviews/scionproto/scion/3843#-MFW0izOCUQheGDwfKOe:-MFiyvhn2ilv4pIOW8jO:bu2a3zy) ([raw file](https://github.com/scionproto/scion/blob/6b0eff79242e42b60fb8283aa75efb20c1273a35/go/cs/reservation/conf/testdata/caps1.json#L26)):*

Ha, so the MarshalIndent in capacities is responsible for this.

Where do you generate this from?
</blockquote></details>

It was generated from https://github.com/juagargi/scion/blob/colibri.store/go/cs/reservation/conf/capacities_test.go#L63
```go
			tc.cap.init()
			buf, err := tc.cap.MarshalJSON()
			require.NoError(t, err)
			expectedJSON := xtest.MustReadFromFile(t, tc.filename)
			require.Equal(t, expectedJSON, buf)
```
It should look better now.
___
*[go/cs/reservation/segment/admission/impl/stateless.go, line 41 at r2](https://reviewable.io/reviews/scionproto/scion/3843#-MFUknBU82SYLhXO1JsK:-MFisGWu7VRm6fYzdOPd:bptf3xd) ([raw file](https://github.com/scionproto/scion/blob/aaec7f6d0057ed95b54aa5be2528dfa8c1e16b9a/go/cs/reservation/segment/admission/impl/stateless.go#L41)):*

This might be valid. I'm complaining about the comment being misleading.
</blockquote></details>

Modified comment.
___
*[go/lib/colibri/reservation/types.go, line 85 at r4](https://reviewable.io/reviews/scionproto/scion/3843#-MFiq6Oo2Qm3u9Kp5eKZ:-MFisjSE9-VwyEazv7Ap:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/fc937b11e513b631d9a89497ac026c725349b2f6/go/lib/colibri/reservation/types.go#L85)):*

could be simplified to `fmt.Sprintf("%s-%x", id.ASID, id.Suffix)`
</blockquote></details>

Done.

Oncilla (approved):
:lgtm:

> Reviewed 5 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3843#-:-MFj08ge2dEnqbPZ-uCN:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[go/cs/reservation/conf/capacities.go, line 64 at r4](https://reviewable.io/reviews/scionproto/scion/3843#-MFin5wfBaI6REr3A88d:-MFj-oBw4wygAyrQi8Tc:bve4kd7) ([raw file](https://github.com/scionproto/scion/blob/fc937b11e513b631d9a89497ac026c725349b2f6/go/cs/reservation/conf/capacities.go#L64)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

Done. My main problem was having a pointer receiver.
</blockquote></details>

ha, yeah those are confusing at times.

PR #3851, opened 2020-08-10
 Return correct err on response error from DB query

andreatulimiero:
When looping over the responses from the DB the error of the response should be returned.
Now no error is returned since err will always be nil; otherwise it either stops before at line 68; or after at line 77.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3851)


Oncilla:
Nice catch :100:

PR #3852, opened 2020-08-11
build: update distroless and debian snapshots

lukedirtwalker:
We saw many issues during downloading debian snapshots,
hopefully this new version works better.

Unfortunately to run distroless we need to manually build the dpkg_parser,
see https://github.com/GoogleContainerTools/distroless/pull/477

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3852)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -109,23 +109,20 @@ _go_image_repos()
 # Distroless
 git_repository(
     name = "distroless",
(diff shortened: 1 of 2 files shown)


lukedirtwalker:
This didn't work as expected. We didn't have issues for a long time.
Eventually we want to build base images we the dependencies installed and then the build is based off on those. That will mean we only need to download deb packages when we change the base images.

PR #3854, opened 2020-08-12
Correctly handling PathCount in PathReq

andreatulimiero:
Currently the `PathCount` flag of a `PathReq` is ignored, thus the received value from `sciond` is always *0* -- which sould mean returning all the possible paths.
Furthemore, in the `sciond.capnp` protocol specification the flag is misplaced out the `flags` group.

P.S. `maxPaths` and `PathCount` (which is translated to `pathCount`) differ in serialization name. For maximum compatibility, `maxPaths` is used as name override for the `PathCount` struct field.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3854)


matzf:
It appears that `PathCount` field was removed altogether in #3862. The `maxpaths` limit is now simply implemented in the `showpaths` application. I think this PR can be closed.

lukedirtwalker:
closing, this no longer applies, as explained above.

PR #3874, opened 2020-09-15
seg: switch path segment representation from capnp to protobuf

Oncilla:
Switch the path segment representation from capnp to protobuf.

The protobuf representation makes a clean split between control plane and data
plane representation of a hop field. We do no longer encode the data plane wire
format in the control plane path segments.

`go/lib/ctrl/seg` is now only the application representation of the path
segments. The application representation can be encoded in the wire format using
`seg.PathSegmentToPB`. The applications are responsible for serialization and
deserialization themselves. This is no longer handled be the `seg` package. The
functionality is moved to the `pathdb` and `beacon` packages.

Signatures are now created more cleanly with the concept of associated data.

Not all extensions are ported. Some are destined to disappear, and some are not
yet ported. For example the static info extension will be ported at a later
stage. Thus, a lot of code regarding static info was disabled.

Even after this PR, code is still tainted by path_mgmt in a lot of places, but
it is now mostly used as a container for segments and revocations. When we drop
header v1 support, and we get rid of revocations, we can start removing
path_mgmt everywhere

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3874)


lukedirtwalker (changes requested):
> Reviewed 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3874)* status: 6 of 144 files reviewed, 1 unresolved discussion (waiting on @Oncilla)

___
*[go/cs/beaconing/main_test.go, line 21 at r2](https://reviewable.io/reviews/scionproto/scion/3874#-MHH4NcB6-3sBu7a-bcD:-MHH4NcB6-3sBu7a-bcE:bqebhii) ([raw file](https://github.com/scionproto/scion/blob/8c61f2d372314f8fa9b8b818bc4ba61cba58b745/go/cs/beaconing/main_test.go#L21)):*
> ```Go
> )
> 
> func init() {
> ```

why this change?

Oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3874)* status: 6 of 144 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

___
*[go/cs/beaconing/main_test.go, line 21 at r2](https://reviewable.io/reviews/scionproto/scion/3874#-MHH4NcB6-3sBu7a-bcD:-MHH6Yaf31m1S2g_4vpi:b-la3o23) ([raw file](https://github.com/scionproto/scion/blob/8c61f2d372314f8fa9b8b818bc4ba61cba58b745/go/cs/beaconing/main_test.go#L21)):*

why this change?
</blockquote></details>

Before, no unit tests were run.

I could also change this to os.Exit(m.Run) but having this in init seamed better

lukedirtwalker (approved):
> Reviewed 143 of 143 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3874)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3862, opened 2020-09-02
protocol: add gRPC messenger and header v2

Oncilla:
This PR introduces major breaking changes to both the control and
data plane of SCION we have been working on for the last couple
of months.

Control plane:

The messenger stack is switched from our homegrown messenger stack
to gRPC over QUIC/SCION and TCP/IP. This gives us many advantages, i.a.,
decreased maintenance efforts, battle tested implementation, designed
for extensibility, access to a large eco system of tools, proper
session management.

AS-Internal communication is switched from a mix of legacy messenger
over UDP/SCION and TCP/IP to just gRPC over TCP/IP. This is an AS
internal detail and is not discoverable by non-local ASes.

Host local communication between applications and SCION Daemon are
switched from custom capnp protocol to gRPC over TCP/IP. As a major
benefit, the SCION daemon can now do proper error signaling.

Communication between ASes is switched from mix of legacy messenger
over UDP/SCION and QUIC/SCION to gRPC over QUIC/SCION, and until v2
header is rolled out, to a minimal legacy UDP/SCION stack that does
the keepalives. Main benefit here will be that gRPC does session
monitoring, and will be able to route around black holes out of the box.

The messages are translated from capnp to protobuf. Some parts are not
yet completely ported, e.g., segements or certificate renewal, or not
yet ported at all. This will happen in follow up PRs.
Segments will also be translated to protobuf.

Legacy capnp definitions are still kept around, until everything is
ported or discarded.

Data plane:

This PR also bundles the header v2 efforts on the client side.
Via a feature flag, applications, such as `scion`, `cs`, and integration
tests can be switched to use the new header design.

The efforts for the router are currently not complete, and the code
will be added in a follow up PR.

Misc:

Because this is a major change and we worked on this internally for
quite some time, we have to bundle it together, and cannot provide a
clean git history with individual commits.

In this PR there are a lot of small changes that went in during this
time. For example, various doc improvements, minor fixes to `scion`
tool and testing pipeline.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3862)

PR #3878, opened 2020-09-16
mocks: update list, fix unexported interfaces in bfd

matzf:
During various refactors (e.g. #3862), some of the mocks were updated without updating the list of mock targets in `tools/gomocks`. Fixed by semi-manually syncing the mock target list with the actually existing mocks.

The interfaces in `go/pkg/router/bfd/` were not directly exported and gomock refused to generate the mock code.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3878)


Oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3878)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[go/pkg/router/bfd/export_test.go, line 48 at r1](https://reviewable.io/reviews/scionproto/scion/3878#-MHQ-P4U1FroJ10Bs6im:-MHQ-P4V9sqZdt9WSf7j:b-xiweex) ([raw file](https://github.com/scionproto/scion/blob/6a8fd3a94d1af8b6e8cdfe2469b9f5e65863c971/go/pkg/router/bfd/export_test.go#L48)):*
> ```Go
> 	State                    = state
> 	Event                    = event
> 	Source                   = source
> ```

Why are these changes needed?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3878)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @Oncilla)

___
*[go/pkg/router/bfd/export_test.go, line 48 at r1](https://reviewable.io/reviews/scionproto/scion/3878#-MHQ-P4U1FroJ10Bs6im:-MHQ0V6v5vvlCPQ2l9ll:b20d7ko) ([raw file](https://github.com/scionproto/scion/blob/6a8fd3a94d1af8b6e8cdfe2469b9f5e65863c971/go/pkg/router/bfd/export_test.go#L48)):*

Why are these changes needed?
</blockquote></details>

mockgen does not see the `Source` and `IntervalGenerator` types, as they are only exported in this `_test.go` file.
It also refuses to create mocks for the unexported interfaces. I don't think there is a way to make these interfaces visible to mockgen and at the same time keep them private -- at least not in the "reflect" mode of mockgen.
Not sure how creating the checked in mocks was accomplished, maybe just manually export interface, run mockgen and then change it back to private?

Oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3878#-:-MHVIVlO2ColRwh2ZFof:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[go/pkg/router/bfd/export_test.go, line 48 at r1](https://reviewable.io/reviews/scionproto/scion/3878#-MHQ-P4U1FroJ10Bs6im:-MHVHKmm0z3yXi9Sc1q6:b-w8x5n0) ([raw file](https://github.com/scionproto/scion/blob/6a8fd3a94d1af8b6e8cdfe2469b9f5e65863c971/go/pkg/router/bfd/export_test.go#L48)):*

mockgen does not see the `Source` and `IntervalGenerator` types, as they are only exported in this `_test.go` file.
It also refuses to create mocks for the unexported interfaces. I don't think there is a way to make these interfaces visible to mockgen and at the same time keep them private -- at least not in the "reflect" mode of mockgen.
Not sure how creating the checked in mocks was accomplished, maybe just manually export interface, run mockgen and then change it back to private?
</blockquote></details>

Hm, yeah this looks a bit awkward. It's a bit unfortunate that gomock does not see the `_test.go` files. But I guess it makes sense.

Conceptually, I think it is cleaner to hide these interfaces, since they solely exist for testing.

You could actually do it with a second definition of the same interface in `go/pkg/router/bfd/internal/`
and then apply mockgen on that. But IMO I don't think that is worth the effort.
___
*[go/pkg/router/bfd/jitter.go, line 67 at r1](https://reviewable.io/reviews/scionproto/scion/3878#-MHVHzpX-pjJGlxJL-17:-MHVHzpX-pjJGlxJL-18:b-cczma5) ([raw file](https://github.com/scionproto/scion/blob/244516d17a545cf27b8cdb77ecf7da288ed077b5/go/pkg/router/bfd/jitter.go#L67)):*
> ```Go
> }
> 
> // IntervalGenerator generates integers in [x, y). It panics if x < 0 or if y <= x.
> ```

Interesting, it looks like this was public once, and then hidden. :detective: 
Same for source.

PR #3883, opened 2020-09-18
docker: remove su-exec

lukedirtwalker:
Use user for running the images instead.

Using su-exec complicated the usage of our images.
Also it is not really clear what the advantages over using docker --user
is. A downside of using `docker --user` is that file permissions have to
be set before starting the image.

Also remove the root checking code in the binaries. The check was flawed
and required to bind mount an actual user in the docker containers.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3883)


lukedirtwalker (commented):
> Reviewed 19 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3883)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3890, opened 2020-10-07
header: the one before dropping legacy header

Oncilla:
The following pull request finalizes the new header format on the
wire. Moving forward, the data and control plane should be backwards
and forwards compatible.

Following is changed in the wire format:

- Paths are renumbered, empty path is now 0, SCION path 1...
- BFD uses empty path for AS internal sessions and one-hop paths for
  inter-AS sessions.
- Service resolution requests now use a protobuf in the request to ensure
  we can add additional data (if ever needed) in a backwards/forwards
  compatible way.

This pull request inverts the feature flag for applications such
as `scion`, `scion-pki`, and integration tests. Instead of specifying
`--features=header_v2`, now, header v2 is used by default and
`--features=header_legacy` needs to be specified to use legacy paths.

The config files do not change.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3890)


Oncilla (commented):
> Reviewed 148 of 148 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3890)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3893, opened 2020-10-08
unify path types in combinator, segutil, sciond

### LINKED ISSUE #3872: Unify path types
Working again on the beaconing static information metadata  (#3779) I observed the number of different types for paths, that are used throughout the system.
There are two main "chains" of type translations (annotated with roughly where the translation is implemented):
- without sciond (in the CS):
  ```
    path segments ---> combinator.Path ---> segutil.path (implements snet.Path)
              combinator           segutil.Router
  ```
- with sciond:
   ```
    path segments ---> combinator.Path ---> sciond.PathReplyEntry ---> sdpb.Paths (transmitted on the wire) ---> sciond.Path (implements snet.Path)
              combinator           sciond/fetcher          server/sciond/grpc                        lib/sciond/grpc
   ```
   In particular, `sciond.PathReplyEntry` is no longer used for anything other than as an intermediate step in this conversion chain.

The two chains result in different structs (`segutil.path` / `sciond.Path`) that almost identically implement `snet.Path`.

## Proposal

* Move `segutil.path` / `sciond.Path` to a share place, e.g. new library `go/lib/path`
* Move translation `combinator.Path -> path.Path` to this new library; alternatively, change the combinator to directly return `path.Path`.
* Change the `segutil.Pather` to return `path.Path`  instead of `combinator.Path` and drop the now redundant `segutil.Router`.
* Drop `sciond.PathReplyEntry`, translate directly from `path.Path` to `sdpd.Paths`

The two chains will then be:

```
    path segments ---> combinator.Path ---> path.Path (implements snet.Path)
              combinator             go/lib/path

    path segments ---> combinator.Path ---> path.Path ---> sdpb.Paths (transmitted on the wire) ---> path.Path (implements snet.Path)
              combinator             go/lib/path  server/sciond/grpc                       lib/sciond/grpc
```

matzf:
* Combine mostly equivalent implementations of `snet.Path`
segutil.path and sciond.Path, in a shared lib in `go/lib/snet/path`.

* Adapt combinator to return a Path that is already almost identical
with this path -- the only missing bit is the UnderlayNextHop (which is
outside of the scope of the combinator, not having access to the local
topology).
In particular, the combinator now directly builds a raw forwarding path
instead of returning a high level description of the same -- using the
structs of the slayers for this actually allows to get rid of most of the
intermediate helper types.
This makes the places that call the combinator fairly trivial.

* Side quest: strip down the public interface of the combinator to the
essentials (combinator.Combine and combinator.Path).

* Drop sciond.FwdPathMeta etc, use snet.Path directly
in the sciond grpc.Daemon server. This is nice because the "input" on the
server side and the output on the client side are now the same type.

* Drop various unused sciond helper types

* Remove sciond staticinfo "condenser" and seperate types. Instead of
this, the staticinfo metadata will be added to snet.PathMetadata.
This data will be prepared and returned in the identical format directly
by the path combinator -- this is the motivation for this refactoring!

* Remove unused sciond/fetcher/filter (instead of fixing the tests)

Planned next steps (future PRs):
* add the "staticinfo" information to the `snet.PathMetadata`, adapt the existing (currently unused) code in the combinator to fill this information, adapt the sciond grpc proto to carry this extended metadata.
* change `snet.PathMetadata` from an interface to a struct. Similar to the `PathInterface`, there simply seems no reason for this to be an interface.
Depending on how exactly the extended "staticinfo" data is represented, this might be useful to allow defining higher-level accessor functions to aggregate the information but still giving applications the full information (e.g. storing latency for each hop, but allowing a simple `TotalLatency()` getter) -- note that having to deal with missing information may require the applications to have more insight than only the pre-condensed data.   
Perhaps move the `Interfaces` list into this Metadata struct as well, as it seems to belong in here.
* perhaps, attempt to turn `snet.Path` into a struct as well -- it seems that the behaviour of all the different implementations (empty path, partial path, ...) can easily be combined.
* remove the now pointless `segutil.Router`

Closes #3872

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3893)


lukedirtwalker (changes requested):
> Reviewed 27 of 27 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3893)* status: all files reviewed, 4 unresolved discussions (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/3893#-MJahMFg4CCME6IKnf9t:-MJahMFg4CCME6IKnf9u:b-a331l0) (no related file):*
nice work :100: 

---
___
*[go/cs/segutil/router.go, line 26 at r1](https://reviewable.io/reviews/scionproto/scion/3893#-MJXVLjmDAmN7K-kwXwW:-MJXVLjmDAmN7K-kwXwX:bmi8ppe) ([raw file](https://github.com/scionproto/scion/blob/de0e91e1ba602fb747c935be79b677a676cd4fec/go/cs/segutil/router.go#L26)):*
> ```Go
> 
> // Router returns paths backed by the local path database.
> // XXX(matzf): this doesn't do meaningful work anymore, drop it.
> ```

:100: 
___
*[go/lib/snet/path/path.go, line 39 at r1](https://reviewable.io/reviews/scionproto/scion/3893#-MJXSVuy9I_qzcqce1b8:-MJXSVuz6-aa6lySrvOh:bmqe3uv) ([raw file](https://github.com/scionproto/scion/blob/de0e91e1ba602fb747c935be79b677a676cd4fec/go/lib/snet/path/path.go#L39)):*
> ```Go
> 	SPath   *spath.Path
> 	NextHop *net.UDPAddr
> 	IFaces  []snet.PathInterface
> ```

not super happy with this naming, but once we move it to PathMetadata we might be able to have a clean name, no method needed.
___
*[go/lib/snet/path/path.go, line 81 at r1](https://reviewable.io/reviews/scionproto/scion/3893#-MJXXwEc5LSVa5cKAXKC:-MJXXwEc5LSVa5cKAXKD:b-az4mzv) ([raw file](https://github.com/scionproto/scion/blob/de0e91e1ba602fb747c935be79b677a676cd4fec/go/lib/snet/path/path.go#L81)):*
> ```Go
> 		return p.Dst
> 	}
> 	return p.IFaces[len(p.IFaces)-1].IA
> ```

don't we always set Dst correctly?
___
*[go/pkg/sciond/fetcher/fetcher.go, line 114 at r1](https://reviewable.io/reviews/scionproto/scion/3893#-MJXUuUS8Axfft0w3r4s:-MJXUuUS8Axfft0w3r4t:beowled) ([raw file](https://github.com/scionproto/scion/blob/de0e91e1ba602fb747c935be79b677a676cd4fec/go/pkg/sciond/fetcher/fetcher.go#L114)):*
> ```
> Bad
> ```
I know this was as before, but it should be lowercase `bad` as error messages should start with lower case. (https://github.com/golang/go/wiki/CodeReviewComments#error-strings )

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3893)* status: all files reviewed, 3 unresolved discussions (waiting on @lukedirtwalker)

___
*[go/lib/snet/path/path.go, line 39 at r1](https://reviewable.io/reviews/scionproto/scion/3893#-MJXSVuy9I_qzcqce1b8:-MJazVBK23FcrzQWq4H8:bb6sh2i) ([raw file](https://github.com/scionproto/scion/blob/de0e91e1ba602fb747c935be79b677a676cd4fec/go/lib/snet/path/path.go#L39)):*

not super happy with this naming, but once we move it to PathMetadata we might be able to have a clean name, no method needed.
</blockquote></details>

Ack.
___
*[go/lib/snet/path/path.go, line 81 at r1](https://reviewable.io/reviews/scionproto/scion/3893#-MJXXwEc5LSVa5cKAXKC:-MJawYQk1a4v_M3S7pj7:b-te5fkt) ([raw file](https://github.com/scionproto/scion/blob/de0e91e1ba602fb747c935be79b677a676cd4fec/go/lib/snet/path/path.go#L81)):*

don't we always set Dst correctly?
</blockquote></details>

Indeed! Done.
___
*[go/pkg/sciond/fetcher/fetcher.go, line 114 at r1](https://reviewable.io/reviews/scionproto/scion/3893#-MJXUuUS8Axfft0w3r4s:-MJazOz2EJDr0aLOozKY:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/de0e91e1ba602fb747c935be79b677a676cd4fec/go/pkg/sciond/fetcher/fetcher.go#L114)):*

> ```
> Bad
> ```
I know this was as before, but it should be lowercase `bad` as error messages should start with lower case. (https://github.com/golang/go/wiki/CodeReviewComments#error-strings )
</blockquote></details>

Done.

lukedirtwalker (approved):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3893)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3899, opened 2020-10-14
port StaticInfoExtension to grpc, reimplement StaticInfo in combinator

matzf:
Add all the previously described metadata, latency, link type, geographical information, bandwidth, number of AS internal hops and generic notes.
Simplify & streamline the representation of the data by using maps instead of lists, in both the protocol and the related higher-level go structs.

Re-implement the extraction of the metadata in the path combinator. Instead doing the complicated walk over the `pathSolution` with explicitly handling cross over and peering etc, use the already created path to guide this. Converts the data directly into a form that will eventually be the representation used in `snet.PathMetadata`. Now explicitly handles conflicting information.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3899)


lukedirtwalker (changes requested):
> Reviewed 4 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3899)* status: 4 of 19 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @matzf)

___
*[proto/control_plane/v1/seg_extensions.proto, line 60 at r1](https://reviewable.io/reviews/scionproto/scion/3899#-MKFL0Sz4FWBIjVhJDxg:-MKFL0T0Bb2dy20hhNR2:b-7tnjv0) ([raw file](https://github.com/scionproto/scion/blob/2c16c2252d7248d31858b3359f6f25989cef364f/proto/control_plane/v1/seg_extensions.proto#L60)):*
> ```
> ·
> ```
trailing space
___
*[proto/control_plane/v1/seg_extensions.proto, line 85 at r1](https://reviewable.io/reviews/scionproto/scion/3899#-MKFLVCM3Ut31PXhZLG6:-MKFLVCM3Ut31PXhZLG7:b48adts) ([raw file](https://github.com/scionproto/scion/blob/2c16c2252d7248d31858b3359f6f25989cef364f/proto/control_plane/v1/seg_extensions.proto#L85)):*
> ```
> interface the relevant
> ```
`and` missing?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3899)* status: 4 of 19 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker)

___
*[proto/control_plane/v1/seg_extensions.proto, line 60 at r1](https://reviewable.io/reviews/scionproto/scion/3899#-MKFL0Sz4FWBIjVhJDxg:-MKFaZ49DPOEan29Y3Jt:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/2c16c2252d7248d31858b3359f6f25989cef364f/proto/control_plane/v1/seg_extensions.proto#L60)):*

> ```
> ·
> ```
trailing space
</blockquote></details>

Done.
___
*[proto/control_plane/v1/seg_extensions.proto, line 85 at r1](https://reviewable.io/reviews/scionproto/scion/3899#-MKFLVCM3Ut31PXhZLG6:-MKFaZo6Ayf6343MnSS_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/2c16c2252d7248d31858b3359f6f25989cef364f/proto/control_plane/v1/seg_extensions.proto#L85)):*

> ```
> interface the relevant
> ```
`and` missing?
</blockquote></details>

Done.

lukedirtwalker (changes requested):
> Reviewed 5 of 19 files at r1, 7 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3899)* status: 16 of 18 files reviewed, 27 unresolved discussions (waiting on @lukedirtwalker and @matzf)

___
*[go/cs/beaconing/staticinfo_config.go, line 116 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYb3Es9YCOxZq4UzOh:-MKYb3Es9YCOxZq4UzOi:b-yyxuyv) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L116)):*
> ```Go
> }
> 
> func (cfg StaticInfoCfg) gather(ifType map[common.IFIDType]topology.LinkType,
> ```

this could be directly inlined in the `Generate` method, or is there a specific reason for this to be separate?
___
*[go/cs/beaconing/staticinfo_config.go, line 130 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYaY1h0moBUYkRiI3y:-MKYaY1h0moBUYkRiI3z:btcdv7h) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L130)):*
> ```
> inserts 
> ```
the comments on those methods are not 100% correct. It doesn't insert it just creates. I wonder if we should rename the methods to `create*` or `generate*`
___
*[go/cs/beaconing/staticinfo_config.go, line 138 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYdqp7E3KQdiFL6lf-:-MKYdqp7E3KQdiFL6lf0:b-ptoqcn) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L138)):*
> ```Go
> 		Inter: make(map[common.IFIDType]time.Duration),
> 	}
> 	if egress != 0 {
> ```

isn't the `Intra` map just empty for the `0` ? 
___
*[go/cs/beaconing/staticinfo_config.go, line 163 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYco4gEOgFoaRBZaH3:-MKYco4gEOgFoaRBZaH4:b-7x6348) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L163)):*
> ```Go
> 		Inter: make(map[common.IFIDType]uint32),
> 	}
> 	if _, ok := cfg.Bandwidth[egress]; egress != 0 && ok {
> ```

hm why doesn't that use the exact same pattern as the method above?
___
*[go/cs/beaconing/staticinfo_config.go, line 164 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYd5VLAQwoqmJSzXn8:-MKYd5VLAQwoqmJSzXn9:b-nt2dwd) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L164)):*
> ```Go
> 	}
> 	if _, ok := cfg.Bandwidth[egress]; egress != 0 && ok {
> 		bw.Inter[egress] = cfg.Bandwidth[egress].Inter
> ```

isn't that already handled by the loop below?
___
*[go/cs/beaconing/staticinfo_config.go, line 201 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYdWEtAgB7gAsK8CAK:-MKYdWEtAgB7gAsK8CAL:bspq7rp) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L201)):*
> ```Go
> 
> 	ihi := make(seg.InternalHopsInfo)
> 	for ifid, v := range cfg.Hops[egress].Intra {
> ```

why do we not need the `egress != 0` check here but we need it above in the other methods?
___
*[go/cs/beaconing/staticinfo_config.go, line 255 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYbUM6-xQORN95TWP0:-MKYbUM6-xQORN95TWP1:b-3rb4m4) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L255)):*
> ```Go
> 
> func interfaceTypeTable(intfs *ifstate.Interfaces) map[common.IFIDType]topology.LinkType {
> 	ifTypes := make(map[common.IFIDType]topology.LinkType)
> ```

we should use the correct size for allocation:
```go
	ifMap := intfs.All()
	ifTypes := make(map[common.IFIDType]topology.LinkType, len(ifMap))
	for ifID, ifInfo := range ifMap {
```
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 15 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZ6ssH9pmT50IDtx1a:-MKZ6ssH9pmT50IDtx1b:b-b4pj4h) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L15)):*
> ```Go
> 
> package seg
> ```

hm I think we should move this to a separate package. For example having `LinkType` in the `seg` package might be confusing in the future since it might not be clear that it belongs to the StaticInfo extension only.

Either we create a `seg/extensions` package and rename the type so using it becomes `extensions.StaticInfo`
or we create a `seg/extensions/staticinfo` package and rename the type so using it becomes `staticinfo.Extension`

from reading I like the first one better, but I think it's cleaner to have separate packages per extension.

or maybe you have a better proposal?
  
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 25 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYi7J95sbEfJ_osW6I:-MKYi7J95sbEfJ_osW6J:bdos2d5) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L25)):*
> ```Go
> )
> 
> type StaticInfoExtension struct {
> ```

I think this and all other structs could have some documentation.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 26 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYfelB2kXcQy-BivdW:-MKYfelB2kXcQy-BivdX:bwpsbi) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L26)):*
> ```
> *LatencyInfo
> ```
I don't think, this and the entry below need to be pointer entries, since the structs are that simple.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 102 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYiQt19STTwsZ19XtQ:-MKYiQt28h_u7i65U-81:b-ojvy71) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L102)):*
> ```Go
> }
> 
> func latencyInfoFromPB(pb *cppb.LatencyInfo) (*LatencyInfo, error) {
> ```

never returns an error, remove it from the signature.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 106 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYjhin5fsVyIs8O3QI:-MKYjhioCHO4lrpNzJNV:b-gghx2h) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L106)):*
> ```Go
> 		return nil, nil
> 	}
> 	intra := make(map[common.IFIDType]time.Duration)
> ```

here and everywhere else us the `len` of the existing map for allocation.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 120 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYiV7D9gzK3V750epd:-MKYiV7D9gzK3V750epe:b-ojvy71) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L120)):*
> ```Go
> }
> 
> func bandwidthInfoFromPB(pb *cppb.BandwidthInfo) (*BandwidthInfo, error) {
> ```

never returns an error, remove it from the signature.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 138 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYiW7j-rfXElseep5T:-MKYiW7j-rfXElseep5U:b-ojvy71) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L138)):*
> ```Go
> }
> 
> func geoInfoFromPB(pb map[uint64]*cppb.GeoCoordinates) (GeoInfo, error) {
> ```

never returns an error, remove it from the signature.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 164 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYk2Hz4fQYuY4Woh9v:-MKYk2Hz4fQYuY4Woh9w:b-ta3u2r) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L164)):*
> ```
> option
> ```
I think this word is not needed.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 164 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYloKvC5P8FGGa961x:-MKYloKvC5P8FGGa961y:bz90t6n) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L164)):*
> ```Go
> 			v = LinkTypeOpennet
> 		default:
> 			return nil, serrors.New("invalid link type option", "link type", vpb)
> ```

I wonder what the correct behavior here is. Erroring out means that if we ever introduce a new value this will be a breaking change. If we'd instead just ignore it (and maybe log something) we would be compatible with new values in the future.

What could be the worst thing that happens if we ignore an unknown value instead of erroring?

(this is just a discussion I'm not sure what the correct behavior should be)
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 171 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYiXXh52P2UXTfJEQM:-MKYiXXh52P2UXTfJEQN:b-ojvy71) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L171)):*
> ```Go
> }
> 
> func internalHopsInfoFromPB(pb map[uint64]uint32) (InternalHopsInfo, error) {
> ```

never returns an error, remove it from the signature.
___
*[go/lib/ctrl/seg/staticinfo_extn_test.go, line 15 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZ6nJg2N3vwhwfFXkQ:-MKZ6nJg2N3vwhwfFXkR:b-h8okxj) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn_test.go#L15)):*
> ```Go
> 
> package seg
> ```

for tests we usually use the `_test` package.
___
*[go/lib/infra/modules/combinator/graph.go, line 559 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYfGTQ5M6Z1GxWInE9:-MKYfGTRCsGgdxhk-j3J:b24jt8m) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/graph.go#L559)):*
> ```
> extensions
> ```
why extensions? it's just the whole ASEntry isn't it?
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 24 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYnOwU2t4Fhy61HGO7:-MKYnOwV0F7B1jzQc_oY:b-5itovk) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L24)):*
> ```Go
> 	Latency []time.Duration
> 
> 	// Bandwidth lists the bandwidth between any two consecutive interfaces.
> ```

this comment should also specify the latency unit.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 102 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYt7hH9gulMc30lk6v:-MKYt7hIElHcC9ZcFeMh:b-g50dg6) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L102)):*
> ```
> Were
> ```
We're
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 113 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYtieW7S9Nd2n2weFd:-MKYtieW7S9Nd2n2weFe:b-y5zrdp) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L113)):*
> ```Go
> 	for _, asEntry := range p.ASEntries {
> 		staticInfo := asEntry.Extensions.StaticInfo
> 		if staticInfo != nil && staticInfo.Latency != nil {
> ```

we usually go for less indentation.
```go
if staticInfo == nil || staticInfo.Latency == nil {
   continue
}
```
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 211 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZAWE_EqEuQ1VYf2oU:-MKZAWE_EqEuQ1VYf2oV:bx28m7z) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L211)):*
> ```Go
> 	for _, asEntry := range p.ASEntries {
> 		staticInfo := asEntry.Extensions.StaticInfo
> 		if staticInfo != nil && staticInfo.Geo != nil {
> ```

use negation and `continue`
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 232 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZANnd4faLyqwDj5wI:-MKZANnd4faLyqwDj5wJ:bniyisa) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L232)):*
> ```Go
> 	for _, asEntry := range p.ASEntries {
> 		staticInfo := asEntry.Extensions.StaticInfo
> 		if staticInfo != nil {
> ```

us `if staticInfo == nil {continue}`
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 285 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZASct-BRhztG981EO:-MKZAScu2U8tBQirGUtG:bx28m7z) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L285)):*
> ```Go
> 			ID: common.IFIDType(asEntry.HopEntry.HopField.ConsEgress),
> 		}
> 		if staticInfo != nil && egIF.ID != 0 {
> ```

use negation and `continue`
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 310 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZ9dMSAuAYbKHU1R_3:-MKZ9dMSAuAYbKHU1R_4:b-mkk53j) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L310)):*
> ```Go
> func collectNotes(p pathInfo) []string {
> 	// can have multiple AS entries for the same AS (at segment cross over, or loop paths).
> 	// collect all notes first
> ```

can the notes from the same IA even be different? if not why not just use a `map[addr.IA]string` ?
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 337 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZ93jc1s7z0_gAgeWh:-MKZ93jc1s7z0_gAgeWi:b-rogl6o) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L337)):*
> ```Go
> }
> 
> func deduplicateStrings(elements []string) []string {
> ```

instead of using this why not simply using a `map[addr.IA]map[string]struct{}` above when collecting? then you only have to convert `map[string]struct{} -> string` in the end.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3899)* status: 7 of 20 files reviewed, 27 unresolved discussions (waiting on @lukedirtwalker and @matzf)

___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 15 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZ6ssH9pmT50IDtx1a:-MKd6mmG4q0lUceKu3e9:b-gxhr08) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L15)):*

hm I think we should move this to a separate package. For example having `LinkType` in the `seg` package might be confusing in the future since it might not be clear that it belongs to the StaticInfo extension only.

Either we create a `seg/extensions` package and rename the type so using it becomes `extensions.StaticInfo`
or we create a `seg/extensions/staticinfo` package and rename the type so using it becomes `staticinfo.Extension`

from reading I like the first one better, but I think it's cleaner to have separate packages per extension.

or maybe you have a better proposal?
  
</blockquote></details>

Done, moved to `seg/extensions/staticinfo`.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 25 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYi7J95sbEfJ_osW6I:-MKd8_7J6wNOqjyCf-C4:bdcor9s) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L25)):*

I think this and all other structs could have some documentation.
</blockquote></details>

Done. I've only added some dummy comments that refer to the more detailed documentation in the protobuf definitions. Otherwise, it would be identical comments, which seems nasty for maintenance.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3899)* status: 5 of 20 files reviewed, 26 unresolved discussions (waiting on @lukedirtwalker)

___
*[go/cs/beaconing/staticinfo_config.go, line 116 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYb3Es9YCOxZq4UzOh:-MKdSnjzDiTbVr9efy7I:bc48h9q) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L116)):*

this could be directly inlined in the `Generate` method, or is there a specific reason for this to be separate?
</blockquote></details>

It's used in the test, so we can directly pass in this `ifType` map.
___
*[go/cs/beaconing/staticinfo_config.go, line 130 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYaY1h0moBUYkRiI3y:-MKdXdBD5ctFAT7Z_L9p:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L130)):*

> ```
> inserts 
> ```
the comments on those methods are not 100% correct. It doesn't insert it just creates. I wonder if we should rename the methods to `create*` or `generate*`
</blockquote></details>

Done.
___
*[go/cs/beaconing/staticinfo_config.go, line 138 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYdqp7E3KQdiFL6lf-:-MKdVBlND3W2F5SjKOoy:b-5lrl3v) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L138)):*

isn't the `Intra` map just empty for the `0` ? 
</blockquote></details>

True. Done.
We should validate that some prankster doesn't use 0 interface IDs in the configuration file then :) I';; do this separately, there is another existing TODO regarding validation of this configuration file. And the configuration file syntax is a bit ... meh anyway.
___
*[go/cs/beaconing/staticinfo_config.go, line 163 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYco4gEOgFoaRBZaH3:-MKdUUpkEyTzbMaHr6hf:bt2vr1b) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L163)):*

hm why doesn't that use the exact same pattern as the method above?
</blockquote></details>

Fixed.
___
*[go/cs/beaconing/staticinfo_config.go, line 164 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYd5VLAQwoqmJSzXn8:-MKdUUZ-8N7UmsSeUNmV:b-tuomwa) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L164)):*

isn't that already handled by the loop below?
</blockquote></details>

Right! Done.
___
*[go/cs/beaconing/staticinfo_config.go, line 201 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYdWEtAgB7gAsK8CAK:-MKdW0q0Dok6SJJtaJIE:bhb1y4e) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L201)):*

why do we not need the `egress != 0` check here but we need it above in the other methods?
</blockquote></details>

Removed the checks above. Done.
___
*[go/cs/beaconing/staticinfo_config.go, line 255 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYbUM6-xQORN95TWP0:-MKdMqzRB_pIwHk88ees:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/cs/beaconing/staticinfo_config.go#L255)):*

we should use the correct size for allocation:
```go
	ifMap := intfs.All()
	ifTypes := make(map[common.IFIDType]topology.LinkType, len(ifMap))
	for ifID, ifInfo := range ifMap {
```
</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/graph.go, line 559 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYfGTQ5M6Z1GxWInE9:-MKdNTc47KMIV18Klf0Y:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/graph.go#L559)):*

> ```
> extensions
> ```
why extensions? it's just the whole ASEntry isn't it?
</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 24 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYnOwU2t4Fhy61HGO7:-MKdO-jE2Sqr0KrzGytp:bkx50vl) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L24)):*

this comment should also specify the latency unit.
</blockquote></details>

Done. Btw. also changed the type to 64bit; probably doesn't hurt and allows values > 4Tbit/s (which seems not entirely inconceivable).
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 102 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYt7hH9gulMc30lk6v:-MKdOzGZ7-nUczlr2ueS:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L102)):*

> ```
> Were
> ```
We're
</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 113 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYtieW7S9Nd2n2weFd:-MKdOvs4B45PI-o06MJy:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L113)):*

we usually go for less indentation.
```go
if staticInfo == nil || staticInfo.Latency == nil {
   continue
}
```
</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 211 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZAWE_EqEuQ1VYf2oU:-MKdPMxC1vKa9s2DO0xI:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L211)):*

use negation and `continue`
</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 232 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZANnd4faLyqwDj5wI:-MKdPO0dCxO-aTlb88CC:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L232)):*

us `if staticInfo == nil {continue}`
</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 285 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZASct-BRhztG981EO:-MKdPXIOD4SDKIxJHsPz:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L285)):*

use negation and `continue`
</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 310 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZ9dMSAuAYbKHU1R_3:-MKdPYWR07nEDVDN16Z-:be7zmvl) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L310)):*

can the notes from the same IA even be different? if not why not just use a `map[addr.IA]string` ?
</blockquote></details>

Yes, at least potentially. We could have segments created at a different point in time, so the note may have changed. Also, different implementations (or our implementation, in the future) could allow specifying different notes for, e.g. different egress interfaces.
I'm aware that just concatenating the different notes is not the greatest idea either, but I figured it's still "cleaner" than just pick one of the notes and pretend this cannot happen.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 337 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZ93jc1s7z0_gAgeWh:-MKdQDgh2qCRDxU9Cv9i:bc5t25k) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L337)):*

instead of using this why not simply using a `map[addr.IA]map[string]struct{}` above when collecting? then you only have to convert `map[string]struct{} -> string` in the end.
</blockquote></details>

Unfortunately this wouldn't give a stable order of the notes, so the concatenated notes could be randomly reordered.

Btw. I think a simpler implementation could be to collect the notes directly in a `map[addr.IA]string` and append to the note if a second entry is not identical to the existing. This would work "correctly" as long as there are only two AS entries for an IA on the path (which should usually be the case, right?), but would do something a bit weird with more than two entries (if some of the notes are different...).
Would seem like a good compromise?
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 26 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYfelB2kXcQy-BivdW:-MKdDXHkC7K2qVSWCDx5:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L26)):*

> ```
> *LatencyInfo
> ```
I don't think, this and the entry below need to be pointer entries, since the structs are that simple.
</blockquote></details>

Done.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 102 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYiQt19STTwsZ19XtQ:-MKdF1h1-HKZufO4fRDx:b-q6c1fv) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L102)):*

never returns an error, remove it from the signature.
</blockquote></details>

Done. Or should we check the range for the interface IDs etc? I think it's fine not to, we'll just ignore garbage values.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 106 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYjhin5fsVyIs8O3QI:-MKdFM1kDO7vPhJv9Wto:b-h8dsdc) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L106)):*

here and everywhere else us the `len` of the existing map for allocation.
</blockquote></details>

Done. TIL, make for maps takes a capacity. :)
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 120 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYiV7D9gzK3V750epd:-MKdL27F56tTUenkXX2a:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L120)):*

never returns an error, remove it from the signature.
</blockquote></details>

Done.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 138 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYiW7j-rfXElseep5T:-MKdL2i38KytYeolNa6g:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L138)):*

never returns an error, remove it from the signature.
</blockquote></details>

Done.
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 164 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYloKvC5P8FGGa961x:-MKdE6fnAmuleFwullk5:boor4sv) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L164)):*

I wonder what the correct behavior here is. Erroring out means that if we ever introduce a new value this will be a breaking change. If we'd instead just ignore it (and maybe log something) we would be compatible with new values in the future.

What could be the worst thing that happens if we ignore an unknown value instead of erroring?

(this is just a discussion I'm not sure what the correct behavior should be)
</blockquote></details>

Good point. It does seem safe to skip this. 
Generally, the risk with this extension seems to be that somebody could send us huge messages with garbage and we'd still accept, store, and forward it. But accepting unknown enum options here really doesn't seem to make this problem any worse at all, so...
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 164 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYk2Hz4fQYuY4Woh9v:-MKdEvYqENobhKZfwFtO:bovqxwe) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L164)):*

> ```
> option
> ```
I think this word is not needed.
</blockquote></details>

Done (error removed).
___
*[go/lib/ctrl/seg/staticinfo_extn.go, line 171 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKYiXXh52P2UXTfJEQM:-MKdMKbM87WDHUQfeY_A:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn.go#L171)):*

never returns an error, remove it from the signature.
</blockquote></details>

Done.
___
*[go/lib/ctrl/seg/staticinfo_extn_test.go, line 15 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZ6nJg2N3vwhwfFXkQ:-MKdMLj27_DC-H9MXeaK:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/ctrl/seg/staticinfo_extn_test.go#L15)):*

for tests we usually use the `_test` package.
</blockquote></details>

Done.

lukedirtwalker (commented):
> Reviewed 15 of 17 files at r3.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3899)* status: 18 of 20 files reviewed, all discussions resolved (waiting on @lukedirtwalker)

___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 337 at r2](https://reviewable.io/reviews/scionproto/scion/3899#-MKZ93jc1s7z0_gAgeWh:-MKi1RZH7c928utklu8W:b-binenc) ([raw file](https://github.com/scionproto/scion/blob/5e70e44566400d7561f22d71ab5d533ccd58e990/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L337)):*

Unfortunately this wouldn't give a stable order of the notes, so the concatenated notes could be randomly reordered.

Btw. I think a simpler implementation could be to collect the notes directly in a `map[addr.IA]string` and append to the note if a second entry is not identical to the existing. This would work "correctly" as long as there are only two AS entries for an IA on the path (which should usually be the case, right?), but would do something a bit weird with more than two entries (if some of the notes are different...).
Would seem like a good compromise?
</blockquote></details>

Ah ok, we can also leave it like this.

lukedirtwalker (changes requested):
> Reviewed 2 of 17 files at r3.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3899)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[go/lib/infra/modules/combinator/staticinfo_accumulator_test.go, line 134 at r3](https://reviewable.io/reviews/scionproto/scion/3899#-MKiWIoy5qdNoxEizR9w:-MKiWIoy5qdNoxEizR9x:bfsm94f) ([raw file](https://github.com/scionproto/scion/blob/625d38930cdb7aac0f4e4609ee397016f279f93a/go/lib/infra/modules/combinator/staticinfo_accumulator_test.go#L134)):*
> ```Go
> 
> 	if len(path) == 0 {
> 		assert.Equal(t, 0, len(latency))
> ```

`assert.Empty(t, latency)` here and everywhere else.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator_test.go, line 140 at r3](https://reviewable.io/reviews/scionproto/scion/3899#-MKiVnJf5xM69qsEiU3U:-MKiVnJf5xM69qsEiU3V:b-3uqjip) ([raw file](https://github.com/scionproto/scion/blob/625d38930cdb7aac0f4e4609ee397016f279f93a/go/lib/infra/modules/combinator/staticinfo_accumulator_test.go#L140)):*
> ```Go
> 	expected := []time.Duration{}
> 	for i := 0; i < len(path)-1; i++ {
> 		ifid_a := path[i].ID
> ```

`snake_case` is not really Go style. Please use camelCase instead. Maybe you can just directly use `path[i].ID``in the method call below?

here and in the rest of the file.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator_test.go, line 193 at r3](https://reviewable.io/reviews/scionproto/scion/3899#-MKiWaN494GyA_1_3NrL:-MKiWaN494GyA_1_3NrM:b-pkwmlb) ([raw file](https://github.com/scionproto/scion/blob/625d38930cdb7aac0f4e4609ee397016f279f93a/go/lib/infra/modules/combinator/staticinfo_accumulator_test.go#L193)):*
> ```
> })
> ```
I think it's slightly cleaner to move this to a new line.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3899)* status: all files reviewed, 3 unresolved discussions (waiting on @lukedirtwalker)

___
*[go/lib/infra/modules/combinator/staticinfo_accumulator_test.go, line 134 at r3](https://reviewable.io/reviews/scionproto/scion/3899#-MKiWIoy5qdNoxEizR9w:-MKjAOho9cv3JxKCRUv8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/625d38930cdb7aac0f4e4609ee397016f279f93a/go/lib/infra/modules/combinator/staticinfo_accumulator_test.go#L134)):*

`assert.Empty(t, latency)` here and everywhere else.
</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator_test.go, line 140 at r3](https://reviewable.io/reviews/scionproto/scion/3899#-MKiVnJf5xM69qsEiU3U:-MKjAnwW1w5LW7zPZjmE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/625d38930cdb7aac0f4e4609ee397016f279f93a/go/lib/infra/modules/combinator/staticinfo_accumulator_test.go#L140)):*

`snake_case` is not really Go style. Please use camelCase instead. Maybe you can just directly use `path[i].ID``in the method call below?

here and in the rest of the file.
</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator_test.go, line 193 at r3](https://reviewable.io/reviews/scionproto/scion/3899#-MKiWaN494GyA_1_3NrL:-MKjAv7K3bBbknWOZj4-:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/625d38930cdb7aac0f4e4609ee397016f279f93a/go/lib/infra/modules/combinator/staticinfo_accumulator_test.go#L193)):*

> ```
> })
> ```
I think it's slightly cleaner to move this to a new line.
</blockquote></details>

Done.

lukedirtwalker (approved):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3899)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3900, opened 2020-10-15
doc: hidden path design update

lukedirtwalker:
This adapts the original proposal to changes we made to the rest of the
SCION ecosystem:

- No longer speaks of PS/BS etc.
- Uses service discovery instead of a separate SVC address.
- Services are defined in gRPC.
- Configs use snake_case.

Apart from that small functional adaptions are proposed:

- The SCION daemon is configured with the hidden path group IDs it
  should query for, instead of having this in the path lookup request.
  This makes everything a bit simpler, doesn't need a special lookup for
  the hidden path config etc.

[doc]

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3900)

### DIFF
--- a/doc/hidden-paths.rst
+++ b/doc/hidden-paths.rst
@@ -2,242 +2,333 @@
 Hidden Paths
 ************
 
-This file documents the design for the Hidden Paths infrastructure.
-
-Naming
-======
-
-+------------------------------------------+-----------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


lukedirtwalker (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3900)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3907, opened 2020-10-20
testing: use bazel to generate mocks

### LINKED ISSUE #2960: Go: rework mock generation
The use of gomocks in our codebase currently has the following issues:
- Generating mocks is not hermetic - it uses whatever version of `mockgen` is in the user's path.
- There is no CI check to ensure that the generated mocks are consistent with the source.
- There is no way to automatically catch generated mock files that remain after their source has been removed.
- There is no documentation on what setup is required before `tools/gomocks` will actually work.

Proposed fix:
- Use bazel to build the `mockgen` binary for the version of `github.com/golang/mock` we're using, and then use _that_ in `tools/gomocks`.
- Move from `scion.sh mocks` to `make mocks`. That way it can depend on setting up the required env (including compiling the mockgen binary).
- CI needs to regenerate mocks and error out if they differ from the ones in git.
- CI needs to error out if there are mock files with no source.
  - This can be done by either:
    - have a whitelist of `mock_X` dirs that should exist, and throw an error if any other `mock_X` dirs exist.
    - nuke all `mock_X` dirs before regenerating mocks.

lukedirtwalker:
This makes `tools/gomocks` bazelingly fast since we can use the bazel
caching and parallelism.

Also add a step to CI that verifies that mocks are up to date.

Fixes #2960

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3907)


lukedirtwalker (commented):
> Reviewed 84 of 84 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3907)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3910, opened 2020-10-21
build/test: reduce make usage

lukedirtwalker:
Calling make directly does some extra work that is not needed.
For example it rebuilds all protobufs etc, we check in a separate
build step that those files are still up to date.

Also remove the nobuild option in the acceptance libs, it's always
true.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3910)

### DIFF
--- a/acceptance/common/scion.py
+++ b/acceptance/common/scion.py
@@ -42,10 +42,9 @@ class SCION(ABC):
         pass
 
     @LogExec(logger, 'running topology')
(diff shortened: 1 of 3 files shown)


lukedirtwalker (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3910)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3912, opened 2020-10-22
cmd/scion: more granular exit codes and sequence filtering

Oncilla:
Add more granular exit codes based on the type of error that is encountered.

`ping` now exits with code 1 if no reply packet was received. If at least one
packet was received, the exit code is 0. If any other error occurred, the exit
code is 2.

`traceroute` now exits with code 1 if any packet is dropped. If any other
error occurred, the exit code is 2.

`showpaths` now exits with code 1 if no path is alive and probing is not
disabled. If probing is disabled, and at least one path is found, the exit
code is 0. If any other error occurred, the exit code is 2.

`ping` and `traceroute` are now able to use the same sequence filter that
was previously introduced to `showpaths`. See the help for more information.

Additionally, fix a bug in `ping` that would report false packet
loss.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3912)

### DIFF
--- a/go/dispatcher/dispatcher/dispatcher.go
+++ b/go/dispatcher/dispatcher/dispatcher.go
@@ -56,7 +56,7 @@ func NewServer(address string, ipv4Conn, ipv6Conn net.PacketConn) (*Server, erro
 		}
 	}
 	return &Server{
-		routingTable: NewIATable(1024, 65535),
+		routingTable: NewIATable(32768, 65535),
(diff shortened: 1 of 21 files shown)


Oncilla (commented):
> Reviewed 21 of 21 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3912)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3915, opened 2020-10-23
cs: avoid HOL blocking in beacon origination

Oncilla:
Previously, the control server would originate on each interface
sequentially. If one beacon origination takes up the whole context,
all origination attempts on the subsequent interfaces would timeout.

This is especially bad, because we iterate the interfaces in a sorted
order. If the first interface that we want to originate on has a problem
and svc resolution fails consistently, this means we would not originate
any beacon.

With this change, beacon origination is executed in parallel
such that head-of-line blocking is no longer an issue.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3915)

### DIFF
--- a/go/cs/beaconing/originator.go
+++ b/go/cs/beaconing/originator.go
@@ -19,6 +19,7 @@ import (
 	"crypto/rand"
 	"math/big"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/scionproto/scion/go/cs/ifstate"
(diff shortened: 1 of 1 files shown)


Oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3915)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3919, opened 2020-10-26
topology: add gateway,interface ID mapping

Oncilla:
The DS will then attach the mapping to Gateway discovery replies. This
means all traffic that targets a remote gateway should flow only through
the interface IDs attached to it.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3919)


Oncilla (commented):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3919)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3920, opened 2020-10-26
cs: fix testing race

Oncilla:
Fix racy access to MAC object in the originator test.
This is hitting us only now, because the originator now does parallel
beacon origination to avoid HOL blocking.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3920)


Oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3920)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3879, opened 2020-09-17
snet: PathInterface struct

matzf:
Use a simple struct for `snet.PathInterface` directly instead of an interface.

Contrary to what was implied by code comments, this does not appear to
introduce any problematic dependencies.

Contributes to #3872 

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3879)

PR #3923, opened 2020-10-29
 snet: make PathMetadata a struct 

matzf:
Make `snet.PathMetadata` a struct instead of an interface. Analogous to how this was done for `snet.PathInterface` (#3879).
The indirection with an interface is simply unnecessary.
Making this a struct, will allow to extend the `PathMetadata` struct quite easily: in a next PR, we'll add the latency, bandwidth, geo coordinates etc. from the "static info" beaconing extension (currently, these fields are defined in a shadow copy of `PathMetadata` in `combinator/staticinfo_accumulator.go`).

Move `Interfaces` to `snet.PathMetadata`. The `Interfaces` list is available exactly if the other metadata items are available, so it makes sense to bundle these.
This requires/brings some changes to `pathpol.PathSet` -- now the pathpol is fed directly with the `snet.PathMetadata` struct. This seems quite natural, as it's conceivable that the path policy would access all the various metadata items in the future.
This change also requires some minor adjustments at the call sites for the path policies.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3923)


lukedirtwalker (changes requested):
> Reviewed 36 of 36 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3923)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/3923#-ML8dVWDB_opzGUW0Rh-:-ML8dVWDB_opzGUW0Rh0:b-5q87r5) (no related file):*
thanks for doing this. :100: 

---
___
*[go/cs/segutil/filter.go, line 48 at r1](https://reviewable.io/reviews/scionproto/scion/3923#-MLH-GXh4k7a-rlI4yBI:-MLH-GXh4k7a-rlI4yBJ:b-38thie) ([raw file](https://github.com/scionproto/scion/blob/20c62ae47cc886f44c08f988b2e078e6305a29cf/go/cs/segutil/filter.go#L48)):*
> ```Go
> // because of peering links.
> // NOTE: policy must not be nil.
> func Filter(segs seg.Segments, policy Policy, dir Direction) seg.Segments {
> ```

just drop this file and the corresponding test, this is never used.
___
*[go/lib/snet/path.go, line 51 at r1](https://reviewable.io/reviews/scionproto/scion/3923#-MLGyaP03W0p8LYbGCwJ:-MLGyaP03W0p8LYbGCwK:bgtagiz) ([raw file](https://github.com/scionproto/scion/blob/c3a80c1f2ec24e2af94265ca0a114cf8ed1e859f/go/lib/snet/path.go#L51)):*
> ```Go
> 	// Metadata returns supplementary information about this path.
> 	// Returns nil if the metadata is not available.
> 	Metadata() *PathMetadata
> ```

Why do we change to pointer here? In many places this is accessed without checking for nil now. for example `path.Metadata().Interfaces` I see in a lot of places. I think just returning by value is fine.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3923)* status: 35 of 40 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker)

___
*[go/cs/segutil/filter.go, line 48 at r1](https://reviewable.io/reviews/scionproto/scion/3923#-MLH-GXh4k7a-rlI4yBI:-MLIehbP4grPJUNNrQST:b555f0y) ([raw file](https://github.com/scionproto/scion/blob/20c62ae47cc886f44c08f988b2e078e6305a29cf/go/cs/segutil/filter.go#L48)):*

just drop this file and the corresponding test, this is never used.
</blockquote></details>

Dang. Done!
___
*[go/lib/snet/path.go, line 51 at r1](https://reviewable.io/reviews/scionproto/scion/3923#-MLGyaP03W0p8LYbGCwJ:-MLIfcpX3IS271zzY48c:bgupi80) ([raw file](https://github.com/scionproto/scion/blob/c3a80c1f2ec24e2af94265ca0a114cf8ed1e859f/go/lib/snet/path.go#L51)):*

Why do we change to pointer here? In many places this is accessed without checking for nil now. for example `path.Metadata().Interfaces` I see in a lot of places. I think just returning by value is fine.
</blockquote></details>

The idea was that it should be `nil`-able. With this, we have one single value for the unavailable metadata, instead of having a default value for each item.
The places that use the `Metadata().Interfaces` without checks are the places were we know the metadata is available, e.g. because the path comes directly from a path query. I quite like that it would panic if you do this mistakenly, instead of giving you a bad default value.

lukedirtwalker (commented):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3923)* status: all files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3923)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3928, opened 2020-11-11
Updating client bootstrapping

andreatulimiero:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3928)


lukedirtwalker:
During a discussion in Slack (https://scionproto.slack.com/archives/C8ADA9CEP/p1614159957002600) we agreed that we currently don't see the SCION repository as the correct place for this PR. We will accept the proposal (documentation), but the implementation doesn't have to live in this repo for now and can live for example in netsec/scion-apps for quicker iteration.

Thanks for your work.

PR #3932, opened 2020-11-20
posix-gateway: add new implementation

Oncilla:
Add a completely overhauled SCION Internet gateway. The inner workings
have completely changed.

The posix-gateway now has dynamic IP prefix discovery. To manage the IP prefix
discovery a routing policy file is used. (see: go/pkg/gateway/routing/doc.go)

Gateway to Gateway probing is now running on fixed port 30856 instead of 30256.
We will use 30256 for remote prefix discovery.

The posix-gateway now supports prefix pinning. To manage the prefix pinning the
`allow_interfaces` property in the gateway section of the topology file can be
used.

The TOML configs have to be changed in the following way:
  - The ip and ctrl_port fields are merged into a ctrl_addr field.
  - The encap_port field was replaced by data_addr field. This allows
    for different IP addresses between the control and data address.
  - The tunneling configuration is extracted in a separate [tunnel] section.
  - Renames:
    - [sig] is renamed to [gateway]
    - sig.sig_config is renamed to gateway.traffic_policy_file
    - sig.tun is renamed to tunnel.name
    - sig.tun_routing_table_id is renamed to tunnel.routing_table_id

The posix-gateway has completely revamped metrics (see: go/pkg/gateway/metrics.go)

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3932)


Oncilla (commented):
> Reviewed 224 of 224 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3932)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3937, opened 2020-11-24
cmds: usage string based on binary name

matzf:
Renaming binaries is sometimes preferrable or even necessary when
creating distribution packages. In order to reduce confusion for the
users, it's nice when the built in help text of the applications refers
to the actually used name of the binary.

By adapting the `Use` string, cobra also generates correct completion
scripts when using a changed name for the binary.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3937)


Oncilla (approved):
:lgtm: :100: 

> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3937#-:-MMzJ5R-3miZC_tJa2ok:bbie1ho)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3941, opened 2020-11-25
gateway: simplify defaults for policy file config

matzf:
The interpretation of the configured `ip_routing_policy_file`
configuration option was:
- if the value is identical to the default file path and the file does
  not exist, ignore it and return a default routing policy
- otherwise, read the file at the specified location (and fail if it
  doesn't exist)

Not only does this seem a bit convoluted, but it can also fail in
interesting ways; when the user does not want to specify an
`ip_routing_policy_file` and there is an error stat-ing the file at the
default location (e.g. no permission to read directory, error while
accessing network file system, etc.), the SIG would refuse to start.

Simplified to only handle the empty value as special default.
The consequence of this is that the configuration files now _must_
specify the `ip_routing_policy` file path if it should be loaded.

~~Also, removed the default for the `traffic_policy_file` option, making
this option mandatory in the configuration file.~~

Finally, removed the unused `dispatcher` configuration entry.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3941)

### DIFF
--- a/go/pkg/gateway/config/config.go
+++ b/go/pkg/gateway/config/config.go
@@ -20,7 +20,6 @@ import (
 	"strconv"
 
 	"github.com/scionproto/scion/go/lib/config"
-	"github.com/scionproto/scion/go/lib/serrors"
 )
 
 // Defaults.
@@ -55,7 +54,7 @@ func (cfg *Gateway) Validate() error {
 		cfg.ID = "gateway"
 	}
 	if cfg.TrafficPolicy == "" {
-		return serrors.New("no traffic policy file path specified")
+		cfg.TrafficPolicy = DefaultSessionPoliciesFile
 	}
 	cfg.CtrlAddr = DefaultAddress(cfg.CtrlAddr, defaultCtrlPort)
 	cfg.DataAddr = DefaultAddress(cfg.DataAddr, defaultDataPort)
--- a/go/pkg/gateway/config/config.go
+++ b/go/pkg/gateway/config/config.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 
 	"github.com/scionproto/scion/go/lib/config"
+	"github.com/scionproto/scion/go/lib/serrors"
 )
 
 // Defaults.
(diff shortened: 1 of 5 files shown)


scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/3941#-MNhEAkZ8Xe7Sxapg4hL:-MNhEAkZ8Xe7Sxapg4hM:bcop9m4) (no related file):*
> Also, removed the default for the traffic_policy_file option, making
> this option mandatory in the configuration file.

What are the benefits of this change? We have this pattern of a default for `""` for most of the `required` files in other applications (e.g., all the TOML configs and DBs have filename defaults). This leads to some inconsistency in how `(required)` files are handled.

---

scrye (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3941)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3941)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @scrye)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/3941#-MNhEAkZ8Xe7Sxapg4hL:-MNhjKNZBhHFxFUVZg3N:b-fsmz9w) (no related file):*

> Also, removed the default for the traffic_policy_file option, making
> this option mandatory in the configuration file.

What are the benefits of this change? We have this pattern of a default for `""` for most of the `required` files in other applications (e.g., all the TOML configs and DBs have filename defaults). This leads to some inconsistency in how `(required)` files are handled.
</blockquote></details>

Reverted it, I hadn't seen this was an established pattern. It seems fairly arbitrary.

---

scrye (approved):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3941)* status: :shipit: complete! all files reviewed, all discussions resolved

scrye:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/3941#-MNhEAkZ8Xe7Sxapg4hL:-MNhnF1_87Srf6LmeM0u:b2akdwa) (no related file):*

Reverted it, I hadn't seen this was an established pattern. It seems fairly arbitrary.
</blockquote></details>

Cool, thanks!

(We use this pattern because it allows us to skip generating most of the config file contents.)

---

PR #3943, opened 2020-11-27
doc: end host bootstrap design proposal

andreatulimiero:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)

### DIFF
--- a/doc/EndhostBootstrap.rst
+++ b/doc/EndhostBootstrap.rst
@@ -1,0 +1,162 @@
+******************
+End host bootstrap
+******************
+
+This file documents the design of the end host bootstrapping.
+
+Overview
+========
+
+Low entry barriers are crucial to push people to experiment with new
+technologies.
+The goal of the automated end host bootstrapping is to let anyone joining the
+SCION internet effortlessly.
+From within an AS, all an end user needs to do is installing a package.
+
+In particular, the package contains a bootstrapper daemon that retrieves
+hints from the local network using `zero conf` services, downloads the
+required SCION configuration files from a local web server, and starts
+the SCION Daemon (SD).
+
+..
+ image:: fig/hidden_paths/HiddenPath.png
+
+Design
+======
+
+Bootstrapping process
+---------------------
+
+An external orchestrator (e.g., *Systemd*) shall take care of: managing the
+bootstrapper daemon; and start the SD once the bootstrapper exits
+successfully.
+
+In the remainder of this section, we assume that the dispatcher, the
+bootstrapper, and the SD services are all enabled.
+
+The bootstrapping process consists of the following:
+
+1. The bootstrapper daemon probes the local network for hints about the
+   web server address using the available discovery mechanisms (i.e., DHCP , DNS and mDNS).
+2. Wait for hints from discoverers.
+3. Once a hint gets received, the bootstrapper tries to download the topology of
+   the AS and some TRCs from the hinted local web server.
+
+   a. On success, the bootstrapper prepares the SD's files and exits successfully, and the SD is automatically started by the orchestrator.
+   b. On failure, the bootstrapper tries to contact the web server specified by the next hint (go back to 2).
+
+If no hint is received after a certain span of time, the bootstrapper times out
+and exits with a non-zero value.
+
+NB: The TRCs retrieval is a temporary solution; in the future, they will be
+installed on a device via other means, ideally before it gets connected to
+a network at all.
+
+SD files creation
+^^^^^^^^^^^^^^^^^
+
+Once retrieved the topology JSON file and the TRCs archive, the bootstrapper
+moves the former to ``<scion_folder>/topology.json``, and the content of the
+latter to ``<scion_folder>/certs/``.
+If an SD alternative configuration file is specified, in the configuration of 
+the bootstrapper, the bootstrapper copies this file to ``<scion_folder>/sd.toml``.
+
+If not specified differently, ``<scion_folder>`` defaults to ``/etc/scion``.
+
+Discovery mechanisms
+--------------------
+
+In this section, we analyze the various discovery mechanisms supported
+by the bootstrapper.
+For clarity's sake, we suppose the following setting:
+
+- the end host is located in the ``ethz.ch`` domain, and
+- the IP address of the web server serving the bootstrapping files can
+  be reached at ``192.33.93.173``.
+
+DHCP
+^^^^
+
+This mechanisms consists of adding a Default World Wide Web server option
+(72) to the DHCP server of the network, whose content is ``192.33.93.173``.
+
+DNS
+^^^
+
+Various discovery mechanisms are based on DNS, each one with its pros and cons.
+
+*DNS SRV*
+
+As specified in [RFC2782]_, a DNS SRV record redirects to an ``A`` record pointing to the web server:
+
+- ``_sciondiscovery._tcp.ethz.ch IN SRV 8041 sciondiscovery.ethz.ch``
+- ``sciondiscovery.ethz.ch IN A 192.33.93.173``
+
+*DNS-SD*
+
+As specified in [RFC6763]_, a list of DNS PTR records point to SRV records,
+each of which define an instance of a SCION discovery service:
+
+- ``_sciondiscovery._tcp.ethz.ch IN PTR SCI-ED._sciondiscovery._tcp.ethz.ch``
+- ``SCI-ED._sciondiscovery._tcp.ethz.ch IN SRV 8041 scied-sciondiscovery.ethz.ch``
+- ``scied-sciondiscovery.ethz.ch IN A 192.33.93.173``
+
+*DNS-NAPTR*
+
+In this variant, a DNS NAPTR record redirects to an ``A`` record pointing to the
+web server:
+
+- ``ethz.ch IN NAPTR "A" "x-sciondiscovery:tcp" "" sciondiscovery.ethz.ch``
+- ``sciondiscovery.ethz.ch IN A 192.33.93.173``
+
+Like the DNS-SD option, multiple NAPTR records for different discovery services
+can be defined.
+
+mDNS
+^^^^
+
+mDNS, a decentralized DNS based on IP multicasting, is usually used
+in combination with the DNS-SD paradigm to achieve *zero conf* networks.
+It removes the need of a centralized DNS server, but it might not be 
+easy to correctly setup a network to work with multicast traffic.
+
+Web server
+----------
+
+The web server (e.g. *Nginx*) shall expose the following endpoints to erve the bootstrapping configuration files:
+
+- ``/scion/discovery/<version>/topology.json``: to retrieve the topology of
+  the AS, and
+- ``/scion/discovery/<version>/trcs.tar``: to retrieve the TRCs needed by the SD.
+
+NB: The endpoints are kept separate since in the future the latter will be removed.
+As previously pointed out, the TRCs will be installed on a device via different
+means.
+
+
+Security
+========
+
+**TBD**
+
+Request for Comments
+====================
+
+1. Unlike the DHCP option, the DNS SRV record can specify a port to reach the
+   service. Currently, if the port is not the canonical one, currently the 8041,
+   the hint is discarded.
+   Do we want this behavior?
+   In my opinion this should be changed.
+2. The name server the DNS discovery mechanisms uses is now retrieved via DHCP,
+   instead of looking it up locally (since most likely it has been already
+   retrieved with the DHCP exchange when the device obtained an IP address).
+   The was motivation for this was to be OS independent.
+   Do we want to keep this behavior?
+
+References
+==========
+
+.. [RFC2782] https://tools.ietf.org/html/rfc2782
+.. [RFC6763] https://tools.ietf.org/html/rfc6763
+
+
--- a/doc/EndhostBootstrap.rst
+++ b/doc/EndhostBootstrap.rst
@@ -1,6 +1,6 @@
-******************
-End host bootstrap
-******************
+********************************
+Automated end host bootstrapping
+********************************
 
 This file documents the design of the end host bootstrapping.
 
@@ -9,61 +9,63 @@ Overview
 
 Low entry barriers are crucial to push people to experiment with new
 technologies.
-The goal of the automated end host bootstrapping is to let anyone joining the
+The goal of automated end host bootstrapping is to let anyone joining the
 SCION internet effortlessly.
 From within an AS, all an end user needs to do is installing a package.
 
 In particular, the package contains a bootstrapper daemon that retrieves
 hints from the local network using `zero conf` services, downloads the
-required SCION configuration files from a local web server, and starts
+required SCION configuration files from a local discovery server, and starts
 the SCION Daemon (SD).
-
-..
- image:: fig/hidden_paths/HiddenPath.png
 
 Design
 ======
 
-Bootstrapping process
+Bootstrapping Process
 ---------------------
 
-An external orchestrator (e.g., *Systemd*) shall take care of: managing the
-bootstrapper daemon; and start the SD once the bootstrapper exits
-successfully.
+An external orchestrator (e.g., *systemd*) will manage the bootstrapper
+daemon and start the SD once the bootstrapper finishes successfully.
 
-In the remainder of this section, we assume that the dispatcher, the
-bootstrapper, and the SD services are all enabled.
+Bootstrapping Components
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+A **hint** is a piece of information returned by a `zero conf` service deployed
+in the local network.
+Depending on the discovery mechanism, a hint can be either sufficient to contact
+a discovery server (e.g., providing its IP address) or can be used to query the local
+network further (e.g., a DNS PTR response).
+
+A **discovery server** is a web server that exposes the endpoints required
+by the bootstrapper (more in the *Discovery Server* section).
+
+A **discoverer** is a client of a `zero conf` service. It, communicates with the service
+and provides hints to the bootstrapper.
+
+Bootstrapping Steps
+^^^^^^^^^^^^^^^^^^^
 
 The bootstrapping process consists of the following:
 
-1. The bootstrapper daemon probes the local network for hints about the
-   web server address using the available discovery mechanisms (i.e., DHCP , DNS and mDNS).
+1. The bootstrapper daemon probes the local network for hints about a
+   discovery server address using the available discovery mechanisms (i.e., DHCP, DNS, and mDNS).
 2. Wait for hints from discoverers.
 3. Once a hint gets received, the bootstrapper tries to download the topology of
-   the AS and some TRCs from the hinted local web server.
+   the AS and some TRCs from the discovery server. While there is no minimum amount of TRCs to
+   be served, the discovery server must provide at least the TRC of the ISD the AS is in.
 
    a. On success, the bootstrapper prepares the SD's files and exits successfully, and the SD is automatically started by the orchestrator.
-   b. On failure, the bootstrapper tries to contact the web server specified by the next hint (go back to 2).
+   b. On failure, go back to 2.
 
-If no hint is received after a certain span of time, the bootstrapper times out
+
+If no hint is received after a certain period, the bootstrapper times out
 and exits with a non-zero value.
 
 NB: The TRCs retrieval is a temporary solution; in the future, they will be
 installed on a device via other means, ideally before it gets connected to
-a network at all.
+a network at all (more in the *Security* section).
 
-SD files creation
-^^^^^^^^^^^^^^^^^
-
-Once retrieved the topology JSON file and the TRCs archive, the bootstrapper
-moves the former to ``<scion_folder>/topology.json``, and the content of the
-latter to ``<scion_folder>/certs/``.
-If an SD alternative configuration file is specified, in the configuration of 
-the bootstrapper, the bootstrapper copies this file to ``<scion_folder>/sd.toml``.
-
-If not specified differently, ``<scion_folder>`` defaults to ``/etc/scion``.
-
-Discovery mechanisms
+Discovery Mechanisms
 --------------------
 
 In this section, we analyze the various discovery mechanisms supported
@@ -71,59 +73,82 @@ by the bootstrapper.
 For clarity's sake, we suppose the following setting:
 
 - the end host is located in the ``ethz.ch`` domain, and
-- the IP address of the web server serving the bootstrapping files can
-  be reached at ``192.33.93.173``.
+- the IP address of the discovery server serving the bootstrapping files can
+  be reached at ``192.168.1.1``.
 
 DHCP
 ^^^^
 
-This mechanisms consists of adding a Default World Wide Web server option
-(72) to the DHCP server of the network, whose content is ``192.33.93.173``.
+The DHCP mechanism requires a DHCP server present in the network.
+Therefore, deploying the DHCP discovery mechanism is only worth the effort if a DHCP server
+is already running on the network.
+The DHCP server has to be configured to announce the addresses of the discovery services
+in the option field with ID 72 ”Default WWW server”, for example, whose content is ``192.168.1.1``.
+We chose this existing option to ease rapid development without going through a formal standardization
+at IETF and because we use the same application-level protocol as used in the WWW, namely HTTP.
+
+This mechanism is the obvious choice for small networks, covering scenarios such as household networks.
 
 DNS
 ^^^
 
-Various discovery mechanisms are based on DNS, each one with its pros and cons.
+The DNS-based mechanisms require public DNS to be set up to contain the necessary records.
+
+Some minimal setups are listed below:
 
 *DNS SRV*
 
-As specified in [RFC2782]_, a DNS SRV record redirects to an ``A`` record pointing to the web server:
+As specified in [RFC2782]_, a DNS SRV record redirects to an ``A`` record pointing to the discovery server:
 
 - ``_sciondiscovery._tcp.ethz.ch IN SRV 8041 sciondiscovery.ethz.ch``
-- ``sciondiscovery.ethz.ch IN A 192.33.93.173``
+- ``sciondiscovery.ethz.ch IN A 192.168.1.1``
 
 *DNS-SD*
 
-As specified in [RFC6763]_, a list of DNS PTR records point to SRV records,
-each of which define an instance of a SCION discovery service:
+As specified in [RFC6763]_, a list of DNS PTR records points to SRV records,
+each of which defines an instance of a SCION discovery service:
 
 - ``_sciondiscovery._tcp.ethz.ch IN PTR SCI-ED._sciondiscovery._tcp.ethz.ch``
 - ``SCI-ED._sciondiscovery._tcp.ethz.ch IN SRV 8041 scied-sciondiscovery.ethz.ch``
-- ``scied-sciondiscovery.ethz.ch IN A 192.33.93.173``
+- ``scied-sciondiscovery.ethz.ch IN A 192.168.1.1``
 
 *DNS-NAPTR*
 
 In this variant, a DNS NAPTR record redirects to an ``A`` record pointing to the
-web server:
+discovery server:
 
 - ``ethz.ch IN NAPTR "A" "x-sciondiscovery:tcp" "" sciondiscovery.ethz.ch``
-- ``sciondiscovery.ethz.ch IN A 192.33.93.173``
+- ``sciondiscovery.ethz.ch IN A 192.168.1.1``
 
 Like the DNS-SD option, multiple NAPTR records for different discovery services
 can be defined.
+
+This mechanism is well suited for large-scale networks having control over their DNS domain.
+Supporting both of them gives network operators more flexibility to choose depending on their DNS layout.
 
 mDNS
 ^^^^
 
 mDNS, a decentralized DNS based on IP multicasting, is usually used
-in combination with the DNS-SD paradigm to achieve *zero conf* networks.
-It removes the need of a centralized DNS server, but it might not be 
-easy to correctly setup a network to work with multicast traffic.
+in combination with DNS-SD to realize *zero conf* networks.
+It removes the need for a centralized DNS server, but it might not be 
+easy to correctly set up a network to work with multicast traffic.
+It is a lightweight solution that requires just one entity besides the client in the network.
+On the one hand, mDNS is a very lightweight and simple solution to deploy.
+However, when planning to deploy mDNS, one must keep in mind that it relies on IP multicast communication.
+If all participating hosts reside this should raise no issue; however, if the network is divided by a router,
+between the clients and the discovery server, the router has to be configured to propagate multicast traffic
+between the two subnets for the discovery to work.
 
-Web server
-----------
+This mechanism is quite flexible and can be used in almost every scenario.
+However, since devices need to be in the same subnet to discover each other it just does not work with a large,
+segmented enterprise network.
 
-The web server (e.g. *Nginx*) shall expose the following endpoints to erve the bootstrapping configuration files:
+Discovery Server
+----------------
+
+The discovery server (e.g. *Nginx*) exposes the following endpoints to
+serve the bootstrapping configuration files:
 
 - ``/scion/discovery/<version>/topology.json``: to retrieve the topology of
   the AS, and
@@ -137,20 +162,31 @@ means.
 Security
 ========
 
-**TBD**
+Guaranteeing the genuinity of the retrieved resources is crucial to ensure that
+a user's connection is safe. To certify a resource this can be signed
+so that after the download the bootstrapper can verify its authenticity.
+
+In the current internet, the root of trust is based on an oligopoly of CAs.
+In SCION, this root of trust is represented by ISD-wise TRCs.
+Nonetheless, like the current internet, a device joining a network for the first time
+needs to have some pre-shared knowledge to judge what it's authentic or not.
+
+While we can consider the TRCs as a temporary solution, the same is not true for the
+topology -- which is at the heart of the automatic bootstrapping.
+For this, a signing solution based on the asymmetric key of an AS can be implemented.
 
 Request for Comments
 ====================
 
 1. Unlike the DHCP option, the DNS SRV record can specify a port to reach the
-   service. Currently, if the port is not the canonical one, currently the 8041,
+   service. Currently, if the port is not the canonical one, the 8041,
    the hint is discarded.
    Do we want this behavior?
-   In my opinion this should be changed.
+   In my opinion, we should change this.
 2. The name server the DNS discovery mechanisms uses is now retrieved via DHCP,
    instead of looking it up locally (since most likely it has been already
    retrieved with the DHCP exchange when the device obtained an IP address).
-   The was motivation for this was to be OS independent.
+   The motivation for this was to be OS independent.
    Do we want to keep this behavior?
 
 References
--- a/doc/EndhostBootstrap.rst
+++ b/doc/EndhostBootstrap.rst
@@ -28,7 +28,7 @@ a discovery server (e.g., providing its IP address) or can be used to query the 
 network further (e.g., a DNS PTR response).
 
 A **discovery server** is a server that exposes the endpoints required
-by the bootstrapper (more in the *Discovery Server* section).
+by the bootstrapper (more in the :ref:`*Discovery Server*<*Discovery Server*>` section).
 
 A **discoverer** is a client of a `zero conf` service. It communicates with the service
 and provides hints to the bootstrapper.
@@ -63,7 +63,7 @@ and exits with a non-zero value.
 
 NB: The TRCs retrieval is a temporary solution; in the future, they will be
 installed on a device via other means, ideally before it gets connected to
-a network at all (more in the *Security* section).
+a network at all (more in the :ref:`*Security*<Security>` section).
 
 Discovery Mechanisms
 --------------------
@@ -165,9 +165,9 @@ a user's connection is not hijacked. To certify a resource, this resource can be
 so that after the download the bootstrapper can verify its authenticity.
 
 In the current Internet, the root of trust is based on an oligopoly of CAs.
-In SCION, this root of trust is represented by ISD-wise TRCs.
+In SCION, this root of trust is represented by one TRC per ISD.
 Nonetheless, like the current Internet, a device joining a network for the first time
-needs to have some pre-shared knowledge to judge what is authentic or not.
+needs to have some pre-shared knowledge to determine what is authentic or not.
 
 While we can consider the discovery of TRCs a temporary solution, the same is not true for the
 topology -- which is at the heart of the automatic bootstrapping.
--- a/doc/EndhostBootstrap.rst
+++ b/doc/EndhostBootstrap.rst
@@ -9,14 +9,29 @@ Overview
 
 Low entry barriers are crucial to push people to experiment with new
 technologies.
-The goal of automated end host bootstrapping is to let anyone joining the
+The goal of automated end host bootstrapping is to let anyone join the
 SCION internet effortlessly.
-From within an AS, all an end user needs to do is installing a package.
+From within an AS, all an end user needs to do is to install a package.
 
 In particular, the package contains a bootstrapper daemon that retrieves
 hints from the local network using `zero conf` services, downloads the
 required SCION configuration files from a local discovery server, and starts
 the SCION Daemon (SD).
+
+Terminology
+-----------
+
+A **hint** is a piece of information returned by a `zero conf` service deployed
+in the local network.
+Depending on the discovery mechanism, a hint can be either sufficient to contact
+a discovery server (e.g., providing its IP address) or can be used to query the local
+network further (e.g., a DNS PTR response).
+
+A **discovery server** is a server that exposes the endpoints required
+by the bootstrapper (more in the *Discovery Server* section).
+
+A **discoverer** is a client of a `zero conf` service. It communicates with the service
+and provides hints to the bootstrapper.
 
 Design
 ======
@@ -27,21 +42,6 @@ Bootstrapping Process
 An external orchestrator (e.g., *systemd*) will manage the bootstrapper
 daemon and start the SD once the bootstrapper finishes successfully.
 
-Bootstrapping Components
-^^^^^^^^^^^^^^^^^^^^^^^^
-
-A **hint** is a piece of information returned by a `zero conf` service deployed
-in the local network.
-Depending on the discovery mechanism, a hint can be either sufficient to contact
-a discovery server (e.g., providing its IP address) or can be used to query the local
-network further (e.g., a DNS PTR response).
-
-A **discovery server** is a web server that exposes the endpoints required
-by the bootstrapper (more in the *Discovery Server* section).
-
-A **discoverer** is a client of a `zero conf` service. It, communicates with the service
-and provides hints to the bootstrapper.
-
 Bootstrapping Steps
 ^^^^^^^^^^^^^^^^^^^
 
@@ -51,7 +51,7 @@ The bootstrapping process consists of the following:
    discovery server address using the available discovery mechanisms (i.e., DHCP, DNS, and mDNS).
 2. Wait for hints from discoverers.
 3. Once a hint gets received, the bootstrapper tries to download the topology of
-   the AS and some TRCs from the discovery server. While there is no minimum amount of TRCs to
+   the AS and some TRCs from the discovery server. While there is no maximum amount of TRCs to
    be served, the discovery server must provide at least the TRC of the ISD the AS is in.
 
    a. On success, the bootstrapper prepares the SD's files and exits successfully, and the SD is automatically started by the orchestrator.
@@ -80,14 +80,12 @@ DHCP
 ^^^^
 
 The DHCP mechanism requires a DHCP server present in the network.
-Therefore, deploying the DHCP discovery mechanism is only worth the effort if a DHCP server
-is already running on the network.
 The DHCP server has to be configured to announce the addresses of the discovery services
-in the option field with ID 72 ”Default WWW server”, for example, whose content is ``192.168.1.1``.
+in the option field with ID 72 ”Default WWW server”, in our example ``192.168.1.1``.
 We chose this existing option to ease rapid development without going through a formal standardization
 at IETF and because we use the same application-level protocol as used in the WWW, namely HTTP.
 
-This mechanism is the obvious choice for small networks, covering scenarios such as household networks.
+This mechanism is a simple one for small networks, covering scenarios such as household networks.
 
 DNS
 ^^^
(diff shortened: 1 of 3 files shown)


shitz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: all files reviewed, 23 unresolved discussions (waiting on @andreatulimiero)

___
*[doc/EndhostBootstrap.rst, line 2 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNP_9FZ9RgcBt9ppkEr:-MNP_9FZ9RgcBt9ppkEs:butfqz2) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L2)):*
> ```reStructuredText
> ******************
> End host bootstrap
> ```

Automated end host boostrapping
___
*[doc/EndhostBootstrap.rst, line 22 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPaMN5Eucx10aLITu0:-MNPaMN5Eucx10aLITu1:bnyu37a) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L22)):*
> ```reStructuredText
> 
> ..
>  image:: fig/hidden_paths/HiddenPath.png
> ```

This doesn't belong here
___
*[doc/EndhostBootstrap.rst, line 31 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPadOg8cQ3hScQ2HOK:-MNPadOg8cQ3hScQ2HOL:b-9ib38y) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L31)):*
> ```
> An external orchestrator (e.g., *Systemd*) shall take care of: managing the
> bootstrapper daemon; and start the SD once the bootstrapper exits
> successfully.
> ```
An external orchestrator (e.g., *systemd*) will manage the bootstrapper daemon and start the SD once the bootstrapper finishes successfully.
___
*[doc/EndhostBootstrap.rst, line 35 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPc1Zk2J1AaOKGwB7-:-MNPc1Zk2J1AaOKGwB70:bmjdnzl) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L35)):*
> ```
> are all enabled
> ```
"enabled" in the sense of systemd? I'm not sure this sentence makes sense without systemd and since systemd was only an example I'd drop it.
___
*[doc/EndhostBootstrap.rst, line 39 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPbIOY9o9-KxP48YIO:-MNPbIOY9o9-KxP48YIP:b22o729) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L39)):*
> ```
> hints
> ```
define "hint"
___
*[doc/EndhostBootstrap.rst, line 39 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPbfAKCvf-AkUJ7JuW:-MNPbfAKCvf-AkUJ7JuX:b-ucmm1j) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L39)):*
> ```
> the
> ```
a
___
*[doc/EndhostBootstrap.rst, line 41 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPbTtx79TpSuoB_sU2:-MNPbTtx79TpSuoB_sU3:bjwaoei) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L41)):*
> ```
> discoverers
>```
define "discoverers"
___
*[doc/EndhostBootstrap.rst, line 43 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPcT515yZuCxXSEaB9:-MNPcT515yZuCxXSEaBA:b-ngr30h) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L43)):*
> ```
> some TRCs
> ```
which TRCs?
___
*[doc/EndhostBootstrap.rst, line 43 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPcXiZALv0um6ww3km:-MNPcXiZALv0um6ww3kn:b6zlzpo) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L43)):*
> ```
> hinted
> ```
drop
___
*[doc/EndhostBootstrap.rst, line 45 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPcjVlBAQMEBsssFKe:-MNPcjVlBAQMEBsssFKf:b-pooixm) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L45)):*
> ```
> and the SD is automatically started by the orchestrator.
> ```
What's the reason for the orchestrator? Why couldn't the bootstrapper process directly start the SD?
___
*[doc/EndhostBootstrap.rst, line 53 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPcw6XCRqJv0waM5rW:-MNPcw6XCRqJv0waM5rX:b73a4oc) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L53)):*
> ```
> installed on a device via other means, ideally before it gets connected to
> a network at all.
> ```
Explain why
___
*[doc/EndhostBootstrap.rst, line 58 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPdBNfCU3svBNFnquz:-MNPdBNfCU3svBNFnqv-:b-p6jzgz) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L58)):*
> ```reStructuredText
> ^^^^^^^^^^^^^^^^^
> 
> Once retrieved the topology JSON file and the TRCs archive, the bootstrapper
> ```

Once IT retrieved
___
*[doc/EndhostBootstrap.rst, line 61 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPdcPbEVGLrJ0EnHIz:-MNPdcPbEVGLrJ0EnHJ-:b-59w5u6) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L61)):*
> ```reStructuredText
> moves the former to ``<scion_folder>/topology.json``, and the content of the
> latter to ``<scion_folder>/certs/``.
> If an SD alternative configuration file is specified, in the configuration of 
> ```

I don't understand this sentence. Does it mean that the bootstrapper will supply the SD with a configuration? Is that configuration downloaded as well from the web server or why is there a need for this?
___
*[doc/EndhostBootstrap.rst, line 75 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPe-453x7tp3lLGcKt:-MNPe-453x7tp3lLGcKu:b-4cnnxc) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L75)):*
> ```
> 192.33.93.173
> ```
This IP is very specific. Use a private IP in documentation.
___
*[doc/EndhostBootstrap.rst, line 77 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPeWtJ2iFPZ_TOk9U7:-MNPeWtJ2iFPZ_TOk9U8:blrbhx1) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L77)):*
> ```reStructuredText
>   be reached at ``192.33.93.173``.
> 
> DHCP
> ```

This section needs to be expanded. The thesis has more details that you can use.
___
*[doc/EndhostBootstrap.rst, line 86 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPf2vTCTBzA0vGVKkA:-MNPf2vTCTBzA0vGVKkB:bnm3xvs) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L86)):*
> ```
>  pros and cons
> ```
You list none of the pros and cons. In general, the sections below are very (too) brief.
___
*[doc/EndhostBootstrap.rst, line 118 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPfDj84vijHJzLA913:-MNPfDj84vijHJzLA914:b-oco2vh) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L118)):*
> ```reStructuredText
> ^^^^
> 
> mDNS, a decentralized DNS based on IP multicasting, is usually used
> ```

I'm not familiar with mDNS and I'm sure many readers won't be. You need to explain this approach. Btw, will you implement this? Otherwise, just drop it.
___
*[doc/EndhostBootstrap.rst, line 126 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPfSO77uOkm6fNRCbl:-MNPfSO77uOkm6fNRCbm:b919h50) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L126)):*
> ```
> shall expose
> ```
exposes
___
*[doc/EndhostBootstrap.rst, line 126 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPfV_3EBZoaqmwgvyS:-MNPfV_3EBZoaqmwgvyT:bep1eko) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L126)):*
> ```
> erve
> ```
serve
___
*[doc/EndhostBootstrap.rst, line 140 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgC7r7ai7T_WKZQp5:-MNPgC7r7ai7T_WKZQp6:bunvtox) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L140)):*
> ```reStructuredText
> ========
> 
> **TBD**
> ```

Please fill in
___
*[doc/EndhostBootstrap.rst, line 146 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgKP53BY8u1w29ka5:-MNPgKP53BY8u1w29ka6:b-gd5nsl) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L146)):*
> ```
> currently the 8041
> ```
Where is this port coming from?
___
*[doc/EndhostBootstrap.rst, line 147 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgQjn0ntjv2MDzDVL:-MNPgQjn0ntjv2MDzDVM:b87p2in) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L147)):*
> ```reStructuredText
> 1. Unlike the DHCP option, the DNS SRV record can specify a port to reach the
>    service. Currently, if the port is not the canonical one, currently the 8041,
>    the hint is discarded.
> ```

Why is that?
___
*[doc/EndhostBootstrap.rst, line 154 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgXdo8BggNO_rmiNt:-MNPgXdo8BggNO_rmiNu:b-fyfpa8) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L154)):*
> ```reStructuredText
>    retrieved with the DHCP exchange when the device obtained an IP address).
>    The was motivation for this was to be OS independent.
>    Do we want to keep this behavior?
> ```

Seems ok to me.  However, it also wouldn't be hard to support Linux Windows and MAC since they all support an `etc/hosts` file (only the location is OS specific)

andreatulimiero (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: all files reviewed, 23 unresolved discussions (waiting on @shitz)

___
*[doc/EndhostBootstrap.rst, line 2 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNP_9FZ9RgcBt9ppkEr:-MNXmyz9BSJ0nALeGcoE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L2)):*

Automated end host boostrapping
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 22 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPaMN5Eucx10aLITu0:-MNXmzZo0MDJkrMxnxuc:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L22)):*

This doesn't belong here
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 31 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPadOg8cQ3hScQ2HOK:-MNXn8TL3OsXo1vGmHXc:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L31)):*

> ```
> An external orchestrator (e.g., *Systemd*) shall take care of: managing the
> bootstrapper daemon; and start the SD once the bootstrapper exits
> successfully.
> ```
An external orchestrator (e.g., *systemd*) will manage the bootstrapper daemon and start the SD once the bootstrapper finishes successfully.
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 35 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPc1Zk2J1AaOKGwB7-:-MNXn9zi-ctLdzJ6gpms:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L35)):*

> ```
> are all enabled
> ```
"enabled" in the sense of systemd? I'm not sure this sentence makes sense without systemd and since systemd was only an example I'd drop it.
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 39 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPbIOY9o9-KxP48YIO:-MNXnAO18Tp1za4XkR2K:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L39)):*

> ```
> hints
> ```
define "hint"
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 39 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPbfAKCvf-AkUJ7JuW:-MNXnAs28OElLRJxewbV:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L39)):*

> ```
> the
> ```
a
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 41 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPbTtx79TpSuoB_sU2:-MNXnEpRCnKzGyYTouII:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L41)):*

> ```
> discoverers
>```
define "discoverers"
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 43 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPcT515yZuCxXSEaB9:-MNXnG2_9anpbeIU_ygm:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L43)):*

> ```
> some TRCs
> ```
which TRCs?
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 43 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPcXiZALv0um6ww3km:-MNXnGUnEBZQwKkYtRVR:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L43)):*

> ```
> hinted
> ```
drop
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 45 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPcjVlBAQMEBsssFKe:-MNXnLtv1ZsbI9WvPCrW:bh9n2hm) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L45)):*

> ```
> and the SD is automatically started by the orchestrator.
> ```
What's the reason for the orchestrator? Why couldn't the bootstrapper process directly start the SD?
</blockquote></details>

I believe it’s best practice for several reasons (e.g., standardized management or restart on fail) to manage a daemon process like the SD with an orcherstrator such as systemd. Since such orchestration tool also offers the chance to take into account the sequentiality requirement that we need, I don’t see why we should not employ one.
___
*[doc/EndhostBootstrap.rst, line 53 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPcw6XCRqJv0waM5rW:-MNXnQb7C42daPypLcum:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L53)):*

> ```
> installed on a device via other means, ideally before it gets connected to
> a network at all.
> ```
Explain why
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 58 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPdBNfCU3svBNFnquz:-MNXnSlJCxY95VnJo5iC:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L58)):*

Once IT retrieved
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 61 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPdcPbEVGLrJ0EnHIz:-MNXnTp58OotEymaj4SM:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L61)):*

I don't understand this sentence. Does it mean that the bootstrapper will supply the SD with a configuration? Is that configuration downloaded as well from the web server or why is there a need for this?
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 75 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPe-453x7tp3lLGcKt:-MNXnUhwAQe0gIrtWGwa:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L75)):*

> ```
> 192.33.93.173
> ```
This IP is very specific. Use a private IP in documentation.
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 77 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPeWtJ2iFPZ_TOk9U7:-MNXnV6ZE-zGpEVRBN6I:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L77)):*

This section needs to be expanded. The thesis has more details that you can use.
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 86 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPf2vTCTBzA0vGVKkA:-MNXnVS2-d4vbgekoNqX:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L86)):*

> ```
>  pros and cons
> ```
You list none of the pros and cons. In general, the sections below are very (too) brief.
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 118 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPfDj84vijHJzLA913:-MNXnW28C8PXWulmwoc0:b1rlfea) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L118)):*

I'm not familiar with mDNS and I'm sure many readers won't be. You need to explain this approach. Btw, will you implement this? Otherwise, just drop it.
</blockquote></details>

mDNS (basically a multicast version of DNS-SD) is already supported (there is as an mDNS discoverer), however I didn’t have the time to test it.
I won't keep on working on this, so it's up to you whether to keep it or not.
___
*[doc/EndhostBootstrap.rst, line 126 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPfSO77uOkm6fNRCbl:-MNXnZzI1YAzWKhaCQUu:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L126)):*

> ```
> shall expose
> ```
exposes
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 126 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPfV_3EBZoaqmwgvyS:-MNXn_NYEhEVRQTNfZHP:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L126)):*

> ```
> erve
> ```
serve
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 140 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgC7r7ai7T_WKZQp5:-MNXn_nv9HnanGD4mTU0:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L140)):*

Please fill in
</blockquote></details>

Done.
___
*[doc/EndhostBootstrap.rst, line 146 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgKP53BY8u1w29ka5:-MNXoP13ANXzqlUGJMig:b-up8k1l) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L146)):*

> ```
> currently the 8041
> ```
Where is this port coming from?
</blockquote></details>

 according to the thesis, the port `8041` was already used by the SCION discovery service before it was dismissed. Thus, it was suggested to keep using it.
___
*[doc/EndhostBootstrap.rst, line 147 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgQjn0ntjv2MDzDVL:-MNXoSROAMAkT_hNU2uo:bsx1ul2) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L147)):*

Why is that?
</blockquote></details>

What do you mean with this? Is this referred to the current behavior, or the fact that I believe this should be changed? In any case, it was already like this in the codebase and I believe this should be changed, since it makes little sense to ignore ports different from the default one, considering that this could be useful, for example, to avoid possible conflicts with existing services in the network.
___
*[doc/EndhostBootstrap.rst, line 154 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgXdo8BggNO_rmiNt:-MNXoW2616ScgrQ2PaQn:b-nc8fml) ([raw file](https://github.com/scionproto/scion/blob/252a59f83843c5798135f5baf60ca8b8b984a65c/doc/EndhostBootstrap.rst#L154)):*

Seems ok to me.  However, it also wouldn't be hard to support Linux Windows and MAC since they all support an `etc/hosts` file (only the location is OS specific)
</blockquote></details>

here I believe you meant `/etc/resolv.conf`.
Anyway, as far as I am concerned, looking up a file would be cleaner and would also make some things simpler — for instance, this would remove the dependency between DHCP and the DNS based discoverers.

shitz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: all files reviewed, 23 unresolved discussions (waiting on @andreatulimiero and @shitz)

___
*[doc/EndhostBootstrap.rst, line 2 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNP_9FZ9RgcBt9ppkEr:-MO8GiUn0cextc7q-_av:b-goslt2) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L2)):*
<details><summary><i>Previously, andreatulimiero (Andrea Tulimiero) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

I think you forgot to push. I don't see any changes here.

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: 0 of 1 files reviewed, 23 unresolved discussions (waiting on @shitz)

___
*[doc/EndhostBootstrap.rst, line 2 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNP_9FZ9RgcBt9ppkEr:-MUIMlZwD1uDK6dlB-Vo:bqr4skd) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L2)):*

I think you forgot to push. I don't see any changes here.
</blockquote></details>

Seems like the changes were pushed to https://github.com/andreatulimiero/scion/tree/bootstrap_update and had been there all along, only found those now.
___
*[doc/EndhostBootstrap.rst, line 39 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPbIOY9o9-KxP48YIO:-MUINCuE2VB7HuhlI26Y:bp39co0) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L39)):*
<details><summary><i>Previously, andreatulimiero (Andrea Tulimiero) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Defined in the subsection `Terminology`
___
*[doc/EndhostBootstrap.rst, line 41 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPbTtx79TpSuoB_sU2:-MUIN_iH4x9O7GLy96VR:bp39co0) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L41)):*
<details><summary><i>Previously, andreatulimiero (Andrea Tulimiero) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Defined in the subsection `Terminology`

shitz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: all files reviewed, 6 unresolved discussions (waiting on @andreatulimiero)

___
*[doc/EndhostBootstrap.rst, line 147 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgQjn0ntjv2MDzDVL:-MUKWgF41juDjQXrIMFq:b-i2jxl3) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L147)):*
<details><summary><i>Previously, andreatulimiero (Andrea Tulimiero) wrote…</i></summary><blockquote>

What do you mean with this? Is this referred to the current behavior, or the fact that I believe this should be changed? In any case, it was already like this in the codebase and I believe this should be changed, since it makes little sense to ignore ports different from the default one, considering that this could be useful, for example, to avoid possible conflicts with existing services in the network.
</blockquote></details>

I agree, I don't see a reason for keeping this behavior.
___
*[doc/EndhostBootstrap.rst, line 154 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgXdo8BggNO_rmiNt:-MUKYRpm7eJvrpg_pW0R:b-s4nc08) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L154)):*
<details><summary><i>Previously, andreatulimiero (Andrea Tulimiero) wrote…</i></summary><blockquote>

here I believe you meant `/etc/resolv.conf`.
Anyway, as far as I am concerned, looking up a file would be cleaner and would also make some things simpler — for instance, this would remove the dependency between DHCP and the DNS based discoverers.
</blockquote></details>

Ah sorry yeah I meant `/etc/resolv.conf`, however it seems that on Mac it's not used and on Windows there is no equivalent. For Linux hosts (which would be the majority at least for now) it would still make sense to use `/etc/resolv.conf` (which is often updated via dhcp anyway).
___
*[doc/EndhostBootstrap.rst, line 66 at r3](https://reviewable.io/reviews/scionproto/scion/3943#-MUK_HzcBXRQlaDaLqi4:-MUK_Hzd9WNVdJ_oQEkG:bsi4mr1) ([raw file](https://github.com/scionproto/scion/blob/41fd7730eb5d4ad78b791fef7f21995e49469574/doc/EndhostBootstrap.rst#L66)):*
> ```reStructuredText
> NB: The TRCs retrieval is a temporary solution; in the future, they will be
> installed on a device via other means, ideally before it gets connected to
> a network at all (more in the *Security* section).
> ```

This should be references (links)
___
*[doc/EndhostBootstrap.rst, line 168 at r3](https://reviewable.io/reviews/scionproto/scion/3943#-MUITbTm1xweyNuYlUnG:-MUITbTn7Qsw4IGuQJos:b-8kmk43) ([raw file](https://github.com/scionproto/scion/blob/41fd7730eb5d4ad78b791fef7f21995e49469574/doc/EndhostBootstrap.rst#L168)):*
> ```
> ISD-wise
> ```
ISD-wide?
___
*[doc/EndhostBootstrap.rst, line 170 at r3](https://reviewable.io/reviews/scionproto/scion/3943#-MUIUF90-3IqI8dRej_1:-MUIUF90-3IqI8dRej_2:b5tbuwx) ([raw file](https://github.com/scionproto/scion/blob/41fd7730eb5d4ad78b791fef7f21995e49469574/doc/EndhostBootstrap.rst#L170)):*
> ```
> judge
> ```
determine
___
*[doc/EndhostBootstrap.rst, line 174 at r3](https://reviewable.io/reviews/scionproto/scion/3943#-MUKZr247kt3vBeaNMSk:-MUKZr247kt3vBeaNMSl:bbjkgho) ([raw file](https://github.com/scionproto/scion/blob/41fd7730eb5d4ad78b791fef7f21995e49469574/doc/EndhostBootstrap.rst#L174)):*
> ```reStructuredText
> While we can consider the discovery of TRCs a temporary solution, the same is not true for the
> topology -- which is at the heart of the automatic bootstrapping.
> For this, a signing solution based on the cryptographic keys of an AS should be implemented.
> ```

If the TRC of the local ISD needs to be pre-installed we could use TLS based on the AS certificate (used as a server certificate).

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: 0 of 1 files reviewed, 6 unresolved discussions (waiting on @andreatulimiero and @shitz)

___
*[doc/EndhostBootstrap.rst, line 147 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgQjn0ntjv2MDzDVL:-MUN3Ut3AaNI1rIZZjpK:bs70uqh) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L147)):*

I agree, I don't see a reason for keeping this behavior.
</blockquote></details>

I think that approach was taken because of implementation simplicity and consistency.
Some mechanism like DHCP option www-server and DNS NAPTR do not provide a port information right now, so having one default port used by all five mechanisms simplified the implementation, and discarding unexpected information was the consequence.

Now for DNS NAPTR the port information could be easily added in the TXT record, but for the DHCP www-server option there is no such field to add information besides the IP. One option for the DHCP approach could be a vendor-encapsulated options (RFC3925), I've seen Anapaya registered a private enterprise number with IANA, PEN `55324` https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers.
___
*[doc/EndhostBootstrap.rst, line 154 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgXdo8BggNO_rmiNt:-MUNbGry4Xego1bg_fTT:b-qkon0l) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L154)):*
> it seems that on Mac it's not used

It is not used by the OS according to the notice, but it is generated there with the correct information, so we can use it to get the DNS server and DNS search domains information when we can't get it via DHCP or DHCP is too slow.

>For Linux hosts (which would be the majority at least for now) it would still make sense to use /etc/resolv.conf

I added that to the implementation, works also for most unix like OSes.

>on Windows there is no equivalent

There is no single file where the information can be easily looked up, but there is a Win32 API for it with GetAdaptersAddresses (https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getadaptersaddresses) that is exactly for that purpose to get DnsServer and DnsSuffix from the IP_ADAPTER_ADDRESSES structure.
I'll add it as well.

So we can be OS independent and use the locally available DNS configuration and DNS search domains information, without relying only on DHCP.
___
*[doc/EndhostBootstrap.rst, line 174 at r3](https://reviewable.io/reviews/scionproto/scion/3943#-MUKZr247kt3vBeaNMSk:-MUNcs2t84_DN9MPAt5X:behpybv) ([raw file](https://github.com/scionproto/scion/blob/41fd7730eb5d4ad78b791fef7f21995e49469574/doc/EndhostBootstrap.rst#L174)):*

If the TRC of the local ISD needs to be pre-installed we could use TLS based on the AS certificate (used as a server certificate). 
</blockquote></details>

If the TRC of the local ISD is already pre-installed we could use TLS based on the AS certificate, I agree.

Should the current implementation also already first fetch the TRC (un-authenticated, TOFU) and then try to validate the topology against that?
Should the bootstrapper assist in validating the TRC manually, have an interactive mode where the values in the TRC are presented to the user or where the user can provide a TRC they obtained in a trusted way out-of-band?

shitz (changes requested):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: all files reviewed, 6 unresolved discussions (waiting on @andreatulimiero and @FR4NK-W)

___
*[doc/EndhostBootstrap.rst, line 147 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgQjn0ntjv2MDzDVL:-MVH3oULBkFKfRKwjzfu:b-p48tq6) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L147)):*

I think that approach was taken because of implementation simplicity and consistency.
Some mechanism like DHCP option www-server and DNS NAPTR do not provide a port information right now, so having one default port used by all five mechanisms simplified the implementation, and discarding unexpected information was the consequence.

Now for DNS NAPTR the port information could be easily added in the TXT record, but for the DHCP www-server option there is no such field to add information besides the IP. One option for the DHCP approach could be a vendor-encapsulated options (RFC3925), I've seen Anapaya registered a private enterprise number with IANA, PEN `55324` https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers.
</blockquote></details>

How well are these vendor-encapsulated options supported by current dhcp servers? If they are well supported, then that would be a good option before something like this would be standardized. If not, we might need to rely on a canonical port.
___
*[doc/EndhostBootstrap.rst, line 154 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgXdo8BggNO_rmiNt:-MVH4tTY4S6grkXHJEK3:b-e91ykd) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L154)):*

> it seems that on Mac it's not used

It is not used by the OS according to the notice, but it is generated there with the correct information, so we can use it to get the DNS server and DNS search domains information when we can't get it via DHCP or DHCP is too slow.

>For Linux hosts (which would be the majority at least for now) it would still make sense to use /etc/resolv.conf

I added that to the implementation, works also for most unix like OSes.

>on Windows there is no equivalent

There is no single file where the information can be easily looked up, but there is a Win32 API for it with GetAdaptersAddresses (https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getadaptersaddresses) that is exactly for that purpose to get DnsServer and DnsSuffix from the IP_ADAPTER_ADDRESSES structure.
I'll add it as well.

So we can be OS independent and use the locally available DNS configuration and DNS search domains information, without relying only on DHCP.
</blockquote></details>

What do you mean by "added it"? 
___
*[doc/EndhostBootstrap.rst, line 174 at r3](https://reviewable.io/reviews/scionproto/scion/3943#-MUKZr247kt3vBeaNMSk:-MVH2kxKAVy9fCfFJUl5:b-89rqla) ([raw file](https://github.com/scionproto/scion/blob/41fd7730eb5d4ad78b791fef7f21995e49469574/doc/EndhostBootstrap.rst#L174)):*

If the TRC of the local ISD is already pre-installed we could use TLS based on the AS certificate, I agree.

Should the current implementation also already first fetch the TRC (un-authenticated, TOFU) and then try to validate the topology against that?
Should the bootstrapper assist in validating the TRC manually, have an interactive mode where the values in the TRC are presented to the user or where the user can provide a TRC they obtained in a trusted way out-of-band?
</blockquote></details>

I'd probably would make this configurable. The bootstrapper has a config option to allow it to download the TRC from the local AS infrastructure. Otherwise, a user needs to copy a TRC to a well-known location on the system.

lukedirtwalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: all files reviewed, 9 unresolved discussions (waiting on @andreatulimiero and @FR4NK-W)

___
*[doc/EndhostBootstrap.rst, line 1 at r4](https://reviewable.io/reviews/scionproto/scion/3943#-MVHMUqs8bkfugcEgp9s:-MVHMUqt8QZdE47BRz0Y:by492kn) ([raw file](https://github.com/scionproto/scion/blob/ba9618ed086a4c8abedc8742500a4a72b32d2ae5/doc/EndhostBootstrap.rst#L1)):*
> ```reStructuredText
> ********************************
> ```

please rename the file to `endhost-bootstrap.rst` we use `kebab-case` for documentation files.
___
*[doc/EndhostBootstrap.rst, line 1 at r4](https://reviewable.io/reviews/scionproto/scion/3943#-MVHNDa06bORagLWeaKv:-MVHNDa1EgdCaZVjdYfO:bxl8hu7) ([raw file](https://github.com/scionproto/scion/blob/ba9618ed086a4c8abedc8742500a4a72b32d2ae5/doc/EndhostBootstrap.rst#L1)):*
> ```reStructuredText
> ********************************
> ```

Do we want to include this file in the index, otherwise it will no be visible in the documentation portal. This should go under a new section called "Proposals".
___
*[doc/EndhostBootstrap.rst, line 4 at r4](https://reviewable.io/reviews/scionproto/scion/3943#-MVHWRJxCQo-BXyKSD8l:-MVHWRJxCQo-BXyKSD8m:b-2jbjkj) ([raw file](https://github.com/scionproto/scion/blob/ba9618ed086a4c8abedc8742500a4a72b32d2ae5/doc/EndhostBootstrap.rst#L4)):*
> ```reStructuredText
> Automated end host bootstrapping
> ********************************
> ```

Please add a proposal header something like this:
```
Author(s): ....

Issue: `xxxx <https://github.com/scionproto/scion/issues/xxxx>`__

Last Updated: 2021-02-09
```
___
*[doc/EndhostBootstrap.rst, line 31 at r4](https://reviewable.io/reviews/scionproto/scion/3943#-MVHNAH84OcHrQmBCXWM:-MVHNAH84OcHrQmBCXWN:b-o1pb5q) ([raw file](https://github.com/scionproto/scion/blob/ba9618ed086a4c8abedc8742500a4a72b32d2ae5/doc/EndhostBootstrap.rst#L31)):*
> ```
> :ref:`*Discovery Server*<*Discovery Server*>`
> ```
this reference doesn't seem to get resolved.
___
*[doc/EndhostBootstrap.rst, line 51 at r4](https://reviewable.io/reviews/scionproto/scion/3943#-MVHRPmf6jgCUWDTZ8xB:-MVHRPmf6jgCUWDTZ8xC:bxvyh65) ([raw file](https://github.com/scionproto/scion/blob/ba9618ed086a4c8abedc8742500a4a72b32d2ae5/doc/EndhostBootstrap.rst#L51)):*
> ```reStructuredText
> 
> 1. The bootstrapper daemon probes the local network for hints about a
>    discovery server address using the available discovery mechanisms (i.e., DHCP, DNS, and mDNS).
> ```

so do we just try everything in parallel?
___
*[doc/EndhostBootstrap.rst, line 66 at r4](https://reviewable.io/reviews/scionproto/scion/3943#-MVHP9IG0jCQ4FS4fJ5z:-MVHP9IHAgbQGHeaHWtD:b-snv4h1) ([raw file](https://github.com/scionproto/scion/blob/ba9618ed086a4c8abedc8742500a4a72b32d2ae5/doc/EndhostBootstrap.rst#L66)):*
> ```
> :ref:`*Security*<Security>`
> ```
the link doesn't seem to work
___
*[doc/EndhostBootstrap.rst, line 95 at r4](https://reviewable.io/reviews/scionproto/scion/3943#-MVHWfdB9gNuNJ1eWhA1:-MVHWfdB9gNuNJ1eWhA2:bkg8fg8) ([raw file](https://github.com/scionproto/scion/blob/ba9618ed086a4c8abedc8742500a4a72b32d2ae5/doc/EndhostBootstrap.rst#L95)):*
> ```reStructuredText
> The DNS-based mechanisms require public DNS to be set up to contain the necessary records.
> 
> Some minimal setups are listed below:
> ```

Do I need to setup everything? One of them? How do they interact with each other?

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: all files reviewed, 9 unresolved discussions (waiting on @andreatulimiero, @FR4NK-W, and @lukedirtwalker)

___
*[doc/EndhostBootstrap.rst, line 51 at r4](https://reviewable.io/reviews/scionproto/scion/3943#-MVHRPmf6jgCUWDTZ8xB:-MVHb84s5K3gWV3Wttts:bfn5xu5) ([raw file](https://github.com/scionproto/scion/blob/ba9618ed086a4c8abedc8742500a4a72b32d2ae5/doc/EndhostBootstrap.rst#L51)):*

so do we just try everything in parallel?
</blockquote></details>

Yes, all the discovery mechanisms that are configured in the config file are attempted in parallel.
___
*[doc/EndhostBootstrap.rst, line 95 at r4](https://reviewable.io/reviews/scionproto/scion/3943#-MVHWfdB9gNuNJ1eWhA1:-MVHbIYHE_Fd32wt_G_b:bskh4bm) ([raw file](https://github.com/scionproto/scion/blob/ba9618ed086a4c8abedc8742500a4a72b32d2ae5/doc/EndhostBootstrap.rst#L95)):*

Do I need to setup everything? One of them? How do they interact with each other?
</blockquote></details>

With the inclusion of support for `resolv.conf` for getting the local DNS configuration, there is no longer required interaction between the mechanisms.
Otherwise the DNS based mechanisms rely on DHCP to work to get the local DNS configuration, such as DNS server and local search domain.

Only a single mechanism is required for successful boostrapping.

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @andreatulimiero and @shitz)

___
*[doc/EndhostBootstrap.rst, line 154 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgXdo8BggNO_rmiNt:-MVWokGg1iuEnTEocNor:brsqmof) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L154)):*
>What do you mean by "added it"?

I had added code to support retrieving the OS provided DNS settings in an OS independent way in addition to fetching those values over DHCP.
Updated comment to show pros and cons of both approaches, and why both might be needed in different contexts.
___
*[doc/EndhostBootstrap.rst, line 174 at r3](https://reviewable.io/reviews/scionproto/scion/3943#-MUKZr247kt3vBeaNMSk:-MVWpX518yjb_hWJAVmU:b-kjbewu) ([raw file](https://github.com/scionproto/scion/blob/41fd7730eb5d4ad78b791fef7f21995e49469574/doc/EndhostBootstrap.rst#L174)):*

I'd probably would make this configurable. The bootstrapper has a config option to allow it to download the TRC from the local AS infrastructure. Otherwise, a user needs to copy a TRC to a well-known location on the system.
</blockquote></details>

Added your suggestion.

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @andreatulimiero and @shitz)

___
*[doc/EndhostBootstrap.rst, line 154 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgXdo8BggNO_rmiNt:-MVWqScQ5jMc4s3mXuh3:bsuuzpv) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L154)):*
> ```
> retrived
> ```
Note to self: typo

lukedirtwalker (approved):
> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: all files reviewed, 3 unresolved discussions (waiting on @andreatulimiero and @shitz)

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @lukedirtwalker and @shitz)

___
*[doc/EndhostBootstrap.rst, line 147 at r1](https://reviewable.io/reviews/scionproto/scion/3943#-MNPgQjn0ntjv2MDzDVL:-MX2HSOK-cP_98OLB3Yp:b5u1iyk) ([raw file](https://github.com/scionproto/scion/blob/46446663ad8d5264a397934b8f95f624d2d7f519/doc/EndhostBootstrap.rst#L147)):*
>How well are these vendor-encapsulated options supported by current dhcp servers?
They are supported for the dhcp servers we checked, cf. scionproto slack channel.

lukedirtwalker (approved):
> Reviewed 3 of 3 files at r7.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943)* status: all files reviewed, 3 unresolved discussions (waiting on @shitz)

shitz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3943#-:-MXfDHa72waq8gHoZSeB:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andreatulimiero)

PR #3944, opened 2020-11-30
epic-hp: control plane

mawyss:
This PR implements the EPIC control plane changes as described in [1].
- Support unauthenticated data in the beacons.
- Add the 10 bytes EPIC authenticators (for the hop entry as well as the peer entries) as unauthenticated data to the beacon.
- Add hash of those 10 bytes to the hop field, therefore this hash is included in the signature. This allows an AS to potentially remove the 10 bytes EPIC authenticators, and the beacon is still verifiable.
- Overhead: no new MAC needs to be calculated, only one additional hash per hop field.

[1] https://scion.docs.anapaya.net/en/latest/EPIC.html#control-plane

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3944)


Oncilla (changes requested):
> Reviewed 1 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3944)* status: 1 of 11 files reviewed, 1 unresolved discussion (waiting on @mawyss)

___
*[proto/control_plane/v1/seg.proto, line 184 at r1](https://reviewable.io/reviews/scionproto/scion/3944#-MNSHqcc8NV-xQhJkM6l:-MNSHqcc8NV-xQhJkM6m:b-2721gf) ([raw file](https://github.com/scionproto/scion/blob/7bb5e8ff606e9d7ce49c7f493d87b3e902a04e21/proto/control_plane/v1/seg.proto#L184)):*
> ```protobuf
>     bytes mac = 4;
>     // Hash of the remaining 10 bytes of the MAC
>     bytes hash_epic_mac = 5;
> ```

Conceptually this is not very clean way of doing.

EPIC is an extension to the core protocol. I don't see a good reason why it should taint the basic functionality here.
We have a pluggable way of adding signed extensions, but if I remember correctly that does not fit well, because you want the flexibility to remove EPIC extensions somewhere along the path.

Now tying the contents of an extension to the core HopField is the wrong direction and sets a dangerous precedent in my opinion.
If we use that pattern, we end up with extension details all over the place in the hop field and other parts of the AS entry.

I propose you go the other direction. Instead of tying the AS entry cryptographically to the EPIC extension, tie the EPIC extension cryptographically to the  AS entry.
This has the added benefit that you can transparently drop the EPIC extension from the segment.

so:

```
// seg.proto 

message ASEntry {
    // The signed part of the AS entry. The body of the SignedMessage is the
    // serialized ASEntrySignedBody. The signature input is defined as following:
    //
    //  input(ps, i) = signed.header_and_body || associated_data(ps, i)
    //
    //  associated_data(ps, i) = ps.segment_info ||
    //                           ps.as_entries[1].signed.header_and_body ||
    //                           ps.as_entries[1].signed.signature ||
    //                           ...
    //                           ps.as_entries[i-1].signed.header_and_body ||
    //                           ps.as_entries[i-1].signed.signature
    //
    proto.crypto.v1.SignedMessage signed = 1;
    // The unsigned part of the AS entry.
    proto.control_plane.v1.PathSegmentUnsignedExtensions unsigned =2;
}
```

```
// seg_extensions.proto

message PathSegmentUnsignedExtensions {
    // Optional EPIC extension.
    // roosd: Use the signature of the AS entry as associated_data here.
    proto.crypto.v1.SignedMessage epic = 1;
}

message EPICExtension {
    // roosd: not sure if that is needed, why not use bytes directly?
    message MAC {
        bytes mac = 1;
    }
    MAC hop_entry = 1;
    repeated MAC peer_entry = 2;
}
```

PR #3983, opened 2021-02-16
gateway: use tunnel.src_ipv4/6 for route src hint

### LINKED ISSUE #3980: gateway: disentangle IP route src hint from data address
On Linux, the SIG currently adds its IP routes to the routing table with the `src` hint field set to the IP of the SIG's `data_addr`.

**Example**: with the `data_addr = "172.16.12.1:30056"` in the configuration, we can see IP routes like this:
```
$ ip route show
...
172.16.11.0/24 dev sig src 172.16.12.1 metric 15
                       ---------------
```

**Problem**:
This `src` hint is incorrect when the `data_addr` is _not_ part of the public IP subnet.
This can occur even in sensible deployments of the SIG, e.g. when there is a separate interface for the connection to the border router (this will be the `data_addr`) and another interface for the  the connection to the hosts in the public IP network. Sketch:

    | Border router |*------- data_addr*[ SIG ]*public_IP------{               }
    |               |*-.                                     {   public subnet   }
                        '--------------------------------------{               }

Here, the correct `src` hint would be "public_IP", not "data_addr".

The equivalent situation occurs when running the SIG in a toy/demo setup with an entire SCION AS running on a single host, with only localhost AS-internal addresses. The SIG will then create routes with localhost as `src` hint, which is invalid (not allowed to send packets to non-localhost with localhost source address).

Having a wrong `src` hint for IP routes created by the SIG is rather inconvenient, as the ip routes are dynamic and so cannot be easily modified using the normal ip route tools. When the default selected source address is wrong because of a bad `src` hint, the correct source address has to be specified separately for each program invocation (e.g. `ping -I`, `nc -s`, `curl --interface`, ...) which is hard and messy.

Some further notes/observations:
* the `src` hint only affects traffic originated by the SIG host, it is not considered when forwarding IP traffic.
* the `src` hint is useful if the SIG host has multiple interfaces and the `data_addr` is part of the public IP subnet routed by this SIG.
* instead of the `src` hint, adding the relevant address on the on the `sig` tunnel interface would work too. This would be slightly more amenable to manual fixups if the wrong address is assigned.

**Suggestion**:
Instead of always using the `data_addr`, add a *another* configuration option for the SIG's public IP address (e.g. `public_addr` or `public_ip`).
This should default to the `data_addr`. 
If explicitly unset, the effect should be to not add a `src` hint (`xnet.AddRoute` already skips the `src` hint if it is empty).

matzf:
Use the existing but previously ignored `tunnel.src_ipv4` / `tunnel.src_ipv6` configuration options as the src IP hint in the IP routes. Explicitly pass different source address for IPv4/6 to `publishingRoutingTable` and use the appropriate address based on the IP version of the specific prefix.

Also, fix naming of "NextHop" member in routemgr.Route -- this is the source address hint for the route.

Finally, fix a small typo in the manual.

Fixes #3980

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3983)


scrye (approved):
:lgtm:

> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3983#-:-MUihila0sSO3UUuWNqY:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3984, opened 2021-02-16
snet: simplify reader code

matzf:
Simplify the somewhat verbose/confusing packet read code in
`snet.scionConnReader`.

* Remove the unnecessary check `if c.base.net == "udp"`. This is always
  true and would require major changes to snet to become false.
  Remove the unnecessary corresponding member in `snet.scionConnBase`.

* Remove redundant type assertions for UDPPayload

* Simplify payload copy

* Do not return a "buffer too small" error.
  `net.UDPConn` does not return an error in this case and it seems to
  make sense to behave in the same way.
  Note that even if we do want to return an error, we should do so
  _after_ copying the part of the payload that does fit into the buffer,
  otherwise the message is just lost entirely.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3984)

### DIFF
--- a/go/lib/snet/base.go
+++ b/go/lib/snet/base.go
@@ -31,9 +31,6 @@ type scionConnBase struct {
 
 	// Reference to SCION networking context
 	scionNet *SCIONNetwork
-
-	// Describes L4 protocol; currently only udp is implemented
-	net string
 }
 
 func (c *scionConnBase) LocalAddr() net.Addr {
--- a/go/lib/snet/reader.go
+++ b/go/lib/snet/reader.go
@@ -41,6 +41,8 @@ func newScionConnReader(base *scionConnBase, conn PacketConn) *scionConnReader {
 
 // ReadFrom reads data into b, returning the length of copied data and the
 // address of the sender.
+// If a message is too long to fit in the supplied buffer, excess bytes may be
+// discarded.
 func (c *scionConnReader) ReadFrom(b []byte) (int, net.Addr, error) {
 	n, a, err := c.read(b)
 	return n, a, err
@@ -48,6 +50,8 @@ func (c *scionConnReader) ReadFrom(b []byte) (int, net.Addr, error) {
 
 // Read reads data into b from a connection with a fixed remote address. If the
 // remote address for the connection is unknown, Read returns an error.
+// If a message is too long to fit in the supplied buffer, excess bytes may be
+// discarded.
 func (c *scionConnReader) Read(b []byte) (int, error) {
 	n, _, err := c.read(b)
 	return n, err
--- a/go/lib/snet/reader.go
+++ b/go/lib/snet/reader.go
@@ -81,17 +81,18 @@ func (c *scionConnReader) read(b []byte) (int, *UDPAddr, error) {
 	// Extract remote address.
 	// Copy the address data to prevent races. See
 	// https://github.com/scionproto/scion/issues/1659.
-	remote := &UDPAddr{}
-	remote.IA = pkt.Source.IA
-	remote.Host = CopyUDPAddr(&net.UDPAddr{
-		IP:   pkt.Source.Host.IP(),
-		Port: int(udp.SrcPort),
-	})
-	remote.Path = pkt.Path.Copy()
+	remote := &UDPAddr{
+		IA: pkt.Source.IA,
+		Host: CopyUDPAddr(&net.UDPAddr{
+			IP:   pkt.Source.Host.IP(),
+			Port: int(udp.SrcPort),
+		}),
+		Path:    pkt.Path.Copy(),
+		NextHop: CopyUDPAddr(&lastHop),
+	}
 	if err = remote.Path.Reverse(); err != nil {
 		return 0, nil, serrors.WrapStr("unable to reverse path on received packet", err)
 	}
-	remote.NextHop = CopyUDPAddr(&lastHop)
 	return n, remote, nil
 }
 
--- a/go/lib/snet/reader.go
+++ b/go/lib/snet/reader.go
@@ -72,48 +72,27 @@ func (c *scionConnReader) read(b []byte) (int, *UDPAddr, error) {
 		return 0, nil, err
 	}
 
-	var n int
 	udp, ok := pkt.Payload.(UDPPayload)
 	if !ok {
 		return 0, nil, serrors.New("unexpected payload", "type", common.TypeOf(pkt.Payload))
 	}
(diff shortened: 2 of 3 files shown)


Oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3984)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[go/lib/snet/reader.go, line 80 at r1](https://reviewable.io/reviews/scionproto/scion/3984#-MTj2n3aDMtLJxm9PWnH:-MTj2n3b9rXWH1agfBLU:b-cyd8d1) ([raw file](https://github.com/scionproto/scion/blob/8a0369cafc5c171bef0d2bdb2accb1efe7b93401/go/lib/snet/reader.go#L80)):*
> ```Go
> 		return 0, nil, serrors.New("unexpected payload", "type", common.TypeOf(pkt.Payload))
> 	}
> 	if len(udp.Payload) > len(b) {
> ```

I don't think you can get rid of this check. Otherwise applications will get truncated payloads without any error.
___
*[go/lib/snet/reader.go, line 84 at r1](https://reviewable.io/reviews/scionproto/scion/3984#-MTjA9XfAa56QtFxtG53:-MTjA9Xg9pqYaJnMvtpd:b-qmrwnb) ([raw file](https://github.com/scionproto/scion/blob/3e20dcf316ca05c06f683929bea3e265210a6d46/go/lib/snet/reader.go#L84)):*
> ```Go
> 	// Copy the address data to prevent races. See
> 	// https://github.com/scionproto/scion/issues/1659.
> 	remote := &UDPAddr{}
> ```

Hm, I don't know why the previous code did it so convoluted, this could simply be:

```go
	remote := &UDPAddr{
		IA: pkt.Source.IA,
		Host: CopyUDPAddr(&net.UDPAddr{
			IP:   pkt.Source.Host.IP(),
			Port: int(udp.SrcPort),
		}),
		Path: pkt.Path.Copy(),
		NextHop: CopyUDPAddr(&lastHop),
	}

	if err = remote.Path.Reverse(); err != nil {
		return 0, nil, serrors.WrapStr("unable to reverse path on received packet", err)
	}

```

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3984)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @Oncilla)

___
*[go/lib/snet/reader.go, line 80 at r1](https://reviewable.io/reviews/scionproto/scion/3984#-MTj2n3aDMtLJxm9PWnH:-MTjB-4J-fnOeAVzYypG:bmcipb4) ([raw file](https://github.com/scionproto/scion/blob/8a0369cafc5c171bef0d2bdb2accb1efe7b93401/go/lib/snet/reader.go#L80)):*

I don't think you can get rid of this check. Otherwise applications will get truncated payloads without any error.
</blockquote></details>

As mentioned in the PR description, this is copying the behaviour of `net.UDPConn`.  If there is a specific reason to still return an error, the error should be returned _after_ everything else, otherwise we're just losing the entire packet.
___
*[go/lib/snet/reader.go, line 84 at r1](https://reviewable.io/reviews/scionproto/scion/3984#-MTjA9XfAa56QtFxtG53:-MTjBwLJ1caAegFxITC-:b-zexlzh) ([raw file](https://github.com/scionproto/scion/blob/3e20dcf316ca05c06f683929bea3e265210a6d46/go/lib/snet/reader.go#L84)):*

Hm, I don't know why the previous code did it so convoluted, this could simply be:

```go
	remote := &UDPAddr{
		IA: pkt.Source.IA,
		Host: CopyUDPAddr(&net.UDPAddr{
			IP:   pkt.Source.Host.IP(),
			Port: int(udp.SrcPort),
		}),
		Path: pkt.Path.Copy(),
		NextHop: CopyUDPAddr(&lastHop),
	}

	if err = remote.Path.Reverse(); err != nil {
		return 0, nil, serrors.WrapStr("unable to reverse path on received packet", err)
	}

```

</blockquote></details>

Done, even better I guess.

Oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3984)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[go/lib/snet/reader.go, line 80 at r1](https://reviewable.io/reviews/scionproto/scion/3984#-MTj2n3aDMtLJxm9PWnH:-MTjFyG13ntD_EUwAOEc:b-a1cb6j) ([raw file](https://github.com/scionproto/scion/blob/8a0369cafc5c171bef0d2bdb2accb1efe7b93401/go/lib/snet/reader.go#L80)):*
Interesting, I did not know that. And I'm a bit surprised the godoc does not mention this behavior.

In that case, I'm fine with the behavior, but please add it to the doc comment of each read method, as others might be surprised as well.

something similar to the recfrom man maybe?
>  If a message is too long to fit in the supplied buffer, excess bytes may be discarded depending on the type of socket the message is received from.

https://linux.die.net/man/2/recvfrom

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3984)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @Oncilla)

___
*[go/lib/snet/reader.go, line 80 at r1](https://reviewable.io/reviews/scionproto/scion/3984#-MTj2n3aDMtLJxm9PWnH:-MTjVemKCMqxrJRmnn6e:bn6p3rd) ([raw file](https://github.com/scionproto/scion/blob/8a0369cafc5c171bef0d2bdb2accb1efe7b93401/go/lib/snet/reader.go#L80)):*

Interesting, I did not know that. And I'm a bit surprised the godoc does not mention this behavior.

In that case, I'm fine with the behavior, but please add it to the doc comment of each read method, as others might be surprised as well.

something similar to the recfrom man maybe?
>  If a message is too long to fit in the supplied buffer, excess bytes may be discarded depending on the type of socket the message is received from.

https://linux.die.net/man/2/recvfrom





</blockquote></details>

Good idea, done. :+1:

Oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3984)* status: :shipit: complete! all files reviewed, all discussions resolved

Oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3984#-:-MTjs3oG5VXfrNfOzFCk:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3993, opened 2021-03-12
cs: more robust retry mechanism for seg requests

matzf:
The current path management mechanism for segment requests is random
selection and retry. Previously, the full request deadline was divided
into three equal time slots and (at most) three attempts were made.
This has two problems:
- Each time slot is short and potentially simply too short to make the
  entire request (in particular for high latency connections). Thus,
  every try may fail even if it hit a working path and if the full
  request deadline would leave enough time to succeed.
- We give up after three failed attempts and do not use the
  remaining time until the request deadline to probe more paths.

These two main changes attempt to alleviate these problems and increase the
chances of success:
- Limit only the time for dialing a gRPC connection for each try; once
  it has dialed succesfully, we give the entire request deadline to each
  try.
- Allow an unlimited number of retries within the request deadline.

~~Additionally, the Dialer now uses `OpenStream` instead of
`OpenStreamSync`. This avoids an unnecessary roundtrip during dialing
which also contributes to increasing the chances of a successful request
within the request deadline.~~

**Note 1**: none of this is really relevant for segment requests from sciond
to CS as these go locally via TCP. This entire retry mechanism is in the
shared logic though and so it is still applied there. This should be
harmless.

**Note 2**: this is an improvement on a temporary solution. The longer term plan is
to replace this with tracking path health more explicitly by using longer lived
gRPC connections, as described in [doc/grpc.rst](https://scion.docs.anapaya.net/en/latest/grpc.html).

**Note 3**: in SCIONLab, this change increases the success chance of certain path
lookups (within the default 5s timeout of showpaths) from 0% to ~70-80%.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/3993)

### DIFF
--- a/go/cs/segreq/fetcher.go
+++ b/go/cs/segreq/fetcher.go
@@ -91,9 +91,8 @@ func NewFetcher(cfg FetcherConfig) *segfetcher.Fetcher {
 			},
 		},
 		Requester: &segfetcher.DefaultRequester{
-			RPC:           cfg.RPC,
-			DstProvider:   d,
-			TimeoutFactor: 0.33,
+			RPC:         cfg.RPC,
+			DstProvider: d,
 		},
 		Metrics: segfetcher.NewFetcherMetrics("control"),
 	}
--- a/go/lib/infra/modules/segfetcher/grpc/requester.go
+++ b/go/lib/infra/modules/segfetcher/grpc/requester.go
@@ -17,12 +17,25 @@ package grpc
 import (
 	"context"
 	"net"
+	"time"
 
 	"github.com/scionproto/scion/go/lib/ctrl/seg"
 	"github.com/scionproto/scion/go/lib/infra/modules/segfetcher"
 	"github.com/scionproto/scion/go/lib/serrors"
 	"github.com/scionproto/scion/go/pkg/grpc"
 	cppb "github.com/scionproto/scion/go/pkg/proto/control_plane"
+)
+
+const (
+	// DefaultRPCDialTimeout is the timeout used for dialing the gRPC ClientConn.
+	// This is shorter than the typical context deadline for the request.
+	// Having a separate, more aggressive timeout for dialing allows to abort
+	// quickly. This allows the surrounding infrastructure to retry quickly -- in
+	// the case where this request goes over SCION/QUIC, retries are used to
+	// route around broken paths.
+	// This timeout needs to be long enough to allow for service address
+	// resolution and the QUIC handshake to complete (two roundtrips).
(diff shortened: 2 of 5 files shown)


lukedirtwalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3993)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/3993#-MVtwHpp6sQ4f6GpoCa5:-MVtwHpp6sQ4f6GpoCa6:b5lfyam) (no related file):*
> Additionally, the Dialer now uses OpenStream instead of
> OpenStreamSync.
Would that still notice a broken connection?

---

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3993)* status: all files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/3993#-MVtwHpp6sQ4f6GpoCa5:-MVu9R1R0qKuGqO827FE:bcdngla) (no related file):*

> Additionally, the Dialer now uses OpenStream instead of
> OpenStreamSync.
Would that still notice a broken connection?
</blockquote></details>

While answering this I noticed that I misunderstood how `OpenStreamSync` differs from `OpenStream` -- from the documentation I had understood that `OpenStreamSync` will _always_ signal to the peer to establish a stream, and  that `OpenStream` somehow folds this signaling directly in the data frames.
Now looking closer, I found that each peer keeps track of a maximum stream ID, and this is already initialized during handshake (or restored for 0-RTT handshake) to the maximum allowed concurrent streams. This value comes from the peers `config.MaxIncomingStream` which defaults to 100. Streams below this limit can be created without any further signalling -- as we only open a single stream, this is the expected case for us.
Signalling is only be required to increase this bound of maximum allowed streams. `OpenStream` does not attempt to do so and just fails if there are no more streams available. `OpenStreamSync` will ask the peer to increase the limit and I think this should effectively block until either the maximum is increased or streams are closed and new IDs become available, but I couldn't quite figure out the details.

TL;DR: in our use case `OpenStream` and `OpenStreamSync` behave the same, but `OpenStreamSync` should probably be the default choice. Changed back.

---

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/3993)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4000, opened 2021-03-17
cs: limit retries for seg requests

matzf:
This fixes a regression introduced in #3993; certain types of errors, in
particular `ErrNotReachable` indicating that there is no path to the
server where the request should be made at, occur very quickly and will
not be fixed by retrying.
When each try fails very quickly, the retry mechanism would just busy
wait until the request context ends.

Abort immediately in case of `ErrNotReachable`.
Also limit the number of retries to catch any other cases of quickly
failing attempts; the number of allowed retries is currently 20, which
should be ample to keep exploring paths even when the request deadline
is generous, but low enough to make sure that run away retries are not
too much of a nuisance.

The retry limit applied for seg reqs from sciond to the CS is set to 1;
these retries are not useful here (TCP, no paths to explore).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4000)

### DIFF
--- a/go/cs/segreq/fetcher.go
+++ b/go/cs/segreq/fetcher.go
@@ -93,7 +93,7 @@ func NewFetcher(cfg FetcherConfig) *segfetcher.Fetcher {
 		Requester: &segfetcher.DefaultRequester{
 			RPC:         cfg.RPC,
 			DstProvider: d,
-			MaxTries:    20,
+			MaxRetries:  20,
 		},
 		Metrics: segfetcher.NewFetcherMetrics("control"),
 	}
--- a/go/cs/segreq/fetcher.go
+++ b/go/cs/segreq/fetcher.go
@@ -93,6 +93,7 @@ func NewFetcher(cfg FetcherConfig) *segfetcher.Fetcher {
 		Requester: &segfetcher.DefaultRequester{
 			RPC:         cfg.RPC,
 			DstProvider: d,
+			MaxTries:    20,
 		},
 		Metrics: segfetcher.NewFetcherMetrics("control"),
 	}
--- a/go/lib/infra/modules/segfetcher/requester.go
+++ b/go/lib/infra/modules/segfetcher/requester.go
@@ -57,7 +57,7 @@ type Requester interface {
 type DefaultRequester struct {
 	RPC         RPC
 	DstProvider DstProvider
-	MaxTries    int
+	MaxRetries  int
 }
 
 // Request all requests in the request set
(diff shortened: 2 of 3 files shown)


lukedirtwalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4000)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[go/lib/infra/modules/segfetcher/requester.go, line 121 at r1](https://reviewable.io/reviews/scionproto/scion/4000#-MW006ik-EPtrE7Yets0:-MW006il-ABxfC-ble_K:blvtruj) ([raw file](https://github.com/scionproto/scion/blob/3671e580a64fda1e5f77781824158c11a5d747bd/go/lib/infra/modules/segfetcher/requester.go#L121)):*
> ```
>  r.maxTries()
> ```
maybe we could simply use `MaxTries +1` to handle the default case.
___
*[go/lib/infra/modules/segfetcher/requester.go, line 126 at r1](https://reviewable.io/reviews/scionproto/scion/4000#-MW00Hg6Bj1wQVoJK3zQ:-MW00Hg6Bj1wQVoJK3zR:b-sd82kl) ([raw file](https://github.com/scionproto/scion/blob/3671e580a64fda1e5f77781824158c11a5d747bd/go/lib/infra/modules/segfetcher/requester.go#L126)):*
> ```Go
> 			logger.Debug("Segment lookup failed", "try", tryIndex+1, "peer", peer,
> 				"err", err)
> 			if errors.Is(err, ErrNotReachable) {
> ```

I think I would put this even before the `if err != nil` block.
___
*[go/lib/infra/modules/segfetcher/requester.go, line 129 at r1](https://reviewable.io/reviews/scionproto/scion/4000#-MW00V0YE-VH6XpP5P_p:-MW00V0Z-NulCYedKhEk:bmaqc0e) ([raw file](https://github.com/scionproto/scion/blob/3671e580a64fda1e5f77781824158c11a5d747bd/go/lib/infra/modules/segfetcher/requester.go#L129)):*
> ```
> else {
> ```
else is not needed.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4000)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @lukedirtwalker)

___
*[go/lib/infra/modules/segfetcher/requester.go, line 121 at r1](https://reviewable.io/reviews/scionproto/scion/4000#-MW006ik-EPtrE7Yets0:-MW03ee6E-tEccGR8gYp:b8fq4s3) ([raw file](https://github.com/scionproto/scion/blob/3671e580a64fda1e5f77781824158c11a5d747bd/go/lib/infra/modules/segfetcher/requester.go#L121)):*

> ```
>  r.maxTries()
> ```
maybe we could simply use `MaxTries +1` to handle the default case.
</blockquote></details>

Done. Renamed `MaxTries` to `Max*Re*tries`.
___
*[go/lib/infra/modules/segfetcher/requester.go, line 126 at r1](https://reviewable.io/reviews/scionproto/scion/4000#-MW00Hg6Bj1wQVoJK3zQ:-MW03m3B25nuKKhnTg3o:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3671e580a64fda1e5f77781824158c11a5d747bd/go/lib/infra/modules/segfetcher/requester.go#L126)):*

I think I would put this even before the `if err != nil` block.
</blockquote></details>

Done.
___
*[go/lib/infra/modules/segfetcher/requester.go, line 129 at r1](https://reviewable.io/reviews/scionproto/scion/4000#-MW00V0YE-VH6XpP5P_p:-MW03mVq65sXOKN8XIcN:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3671e580a64fda1e5f77781824158c11a5d747bd/go/lib/infra/modules/segfetcher/requester.go#L129)):*

> ```
> else {
> ```
else is not needed.
</blockquote></details>

Done.

lukedirtwalker (approved):
:lgtm: thanks

> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4000#-:-MW3Q_Bb1gWodLqGoNe9:b-v6ynew)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4005, opened 2021-03-26
daemon/snet/showpaths: allow zero for hop latency

matzf:
Allow to use zero for the latency metadata of each hop. Zero can be a sensible value for this; for example when multiple interfaces are part of the same router instance, then it makes sense to announce zero for the latency between these interfaces.

Use negative value (-1 nanosecond) to represent undefined values internally, instead.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4005)

### DIFF
--- a/go/lib/infra/modules/combinator/staticinfo_accumulator.go
+++ b/go/lib/infra/modules/combinator/staticinfo_accumulator.go
@@ -103,7 +103,12 @@ func collectLatency(p pathInfo) []time.Duration {
 	// 2)
 	latencies := make([]time.Duration, len(p.Interfaces)-1)
 	for i := 0; i+1 < len(p.Interfaces); i++ {
-		latencies[i] = hopLatencies[makeHopKey(p.Interfaces[i], p.Interfaces[i+1])]
+		l, ok := hopLatencies[makeHopKey(p.Interfaces[i], p.Interfaces[i+1])]
+		if ok {
+			latencies[i] = l
+		} else {
+			latencies[i] = snet.LatencyUnset
+		}
 	}
 
 	return latencies
@@ -114,9 +119,6 @@ func collectLatency(p pathInfo) []time.Duration {
 func addHopLatency(m map[hopKey]time.Duration, a, b snet.PathInterface, v time.Duration) {
 	// Skip incomplete entries; not strictly necessary, we'd just not look this up
 	if a.ID == 0 || b.ID == 0 {
-		return
-	}
-	if v == 0 {
 		return
 	}
 	k := makeHopKey(a, b)
@@ -165,9 +167,6 @@ func collectBandwidth(p pathInfo) []uint64 {
 func addHopBandwidth(m map[hopKey]uint64, a, b snet.PathInterface, v uint64) {
 	// Skip incomplete entries; not strictly necessary, we'd just not look this up
 	if a.ID == 0 || b.ID == 0 {
-		return
-	}
-	if v == 0 {
 		return
 	}
 	k := makeHopKey(a, b)
@@ -282,9 +281,6 @@ func collectInternalHops(p pathInfo) []uint32 {
 func addHopInternalHops(m map[hopKey]uint32, a, b snet.PathInterface, v uint32) {
 	// Skip incomplete entries; not strictly necessary, we'd just not look this up
 	if a.ID == 0 || b.ID == 0 {
-		return
-	}
-	if v == 0 {
 		return
 	}
 	k := makeHopKey(a, b)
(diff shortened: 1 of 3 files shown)


lukedirtwalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4005)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 110 at r1](https://reviewable.io/reviews/scionproto/scion/4005#-MXf_7Xb4DiDMXjt-8TH:-MXf_7Xb4DiDMXjt-8TI:bz6oj4m) ([raw file](https://github.com/scionproto/scion/blob/54e996a1b354ba1285637a9d77127115fd928163/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L110)):*
> ```Go
> 			latencies[i] = l
> 		} else {
> 			latencies[i] = snet.LatencyUnset
> ```

this could be set unconditionally before, then no else is needed.
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 119 at r1](https://reviewable.io/reviews/scionproto/scion/4005#-MXf_FWZ1TpEkA08wKrQ:-MXf_FWZ1TpEkA08wKrR:b-fb55if) ([raw file](https://github.com/scionproto/scion/blob/7bb534ff3d840d38c398a31672e98ecf951655a9/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L119)):*
> ```Go
> 		return
> 	}
> 	if v == 0 {
> ```

could the passed value be negative?
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 170 at r1](https://reviewable.io/reviews/scionproto/scion/4005#-MXfZgsY09nTitpdj6Ke:-MXfZgsY09nTitpdj6Kf:bukmp8w) ([raw file](https://github.com/scionproto/scion/blob/7bb534ff3d840d38c398a31672e98ecf951655a9/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L170)):*
> ```Go
> 		return
> 	}
> 	if v == 0 {
> ```

bandwidth 0 doesn't make sense to me, what is the use case here?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4005)* status: all files reviewed, 3 unresolved discussions (waiting on @lukedirtwalker)

___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 110 at r1](https://reviewable.io/reviews/scionproto/scion/4005#-MXf_7Xb4DiDMXjt-8TH:-MXgLFe96ypMJEwrXjSL:b1a7xcv) ([raw file](https://github.com/scionproto/scion/blob/54e996a1b354ba1285637a9d77127115fd928163/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L110)):*

this could be set unconditionally before, then no else is needed.
</blockquote></details>

Done, sort of. Better?
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 119 at r1](https://reviewable.io/reviews/scionproto/scion/4005#-MXf_FWZ1TpEkA08wKrQ:-MXg8WCT2NjnEQ7dhr96:bvrxyat) ([raw file](https://github.com/scionproto/scion/blob/7bb534ff3d840d38c398a31672e98ecf951655a9/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L119)):*

could the passed value be negative?
</blockquote></details>

Great point, yes I think it could be negative, as there is nothing up to this point that would filter out such invalid values .
___
*[go/lib/infra/modules/combinator/staticinfo_accumulator.go, line 170 at r1](https://reviewable.io/reviews/scionproto/scion/4005#-MXfZgsY09nTitpdj6Ke:-MXgFXnuELWJLYxjO6A3:b-pxaw9b) ([raw file](https://github.com/scionproto/scion/blob/7bb534ff3d840d38c398a31672e98ecf951655a9/go/lib/infra/modules/combinator/staticinfo_accumulator.go#L170)):*

bandwidth 0 doesn't make sense to me, what is the use case here?
</blockquote></details>

Great point, thanks!
I removed it because I thought it was redundant -- thinking it through again, I realized it's not entirely redundant as it will filter out an explicitly announced 0 value in the case that there is conflicting information.
Added the check back.

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4005)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4011, opened 2021-04-07
ping: attach error details if there are interfaces down on the path

juagargi:
When processing the response to an echo request, create specialized errors for replies that indicate that interfaces were down on the path.

The following error detail is added:
- SCMPExternalInterfaceDown: the offending ISD-AS and the interface identifier
- SCMPInternalConnectivityDown: the offending ISD-AS and the ingress/egress interface identifiers 

This helps pinpoint connectivity issues on a SCION path that is used for pinging.

An example output for the `scion` tool after the change looks like this:

    ERROR: external interface is down isd_as="1-ff00:0:110" interface="42"

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4011)

### DIFF
--- a/go/pkg/ping/ping.go
+++ b/go/pkg/ping/ping.go
@@ -313,7 +313,8 @@ func (h scmpHandler) handle(pkt *snet.Packet) (snet.SCMPEchoReply, error) {
 		return snet.SCMPEchoReply{}, serrors.New("internal connectivity is down",
 			"isd_as", s.IA, "ingress", s.Ingress, "egress", s.Egress)
 	default:
-		return snet.SCMPEchoReply{}, serrors.New("not SCMPEchoReply", "type", common.TypeOf(pkt.Payload))
+		return snet.SCMPEchoReply{}, serrors.New("not SCMPEchoReply",
+			"type", common.TypeOf(pkt.Payload))
 	}
 	r := pkt.Payload.(snet.SCMPEchoReply)
 	if r.Identifier != h.id {
--- a/go/pkg/ping/ping.go
+++ b/go/pkg/ping/ping.go
@@ -304,11 +304,20 @@ func (h scmpHandler) handle(pkt *snet.Packet) (snet.SCMPEchoReply, error) {
 	if pkt.Payload == nil {
 		return snet.SCMPEchoReply{}, serrors.New("no v2 payload found")
 	}
-	r, ok := pkt.Payload.(snet.SCMPEchoReply)
-	if !ok {
-		return snet.SCMPEchoReply{}, serrors.New("not SCMPEchoReply",
-			"type", common.TypeOf(pkt.Payload))
+	var errDetail []interface{}
+	switch s := pkt.Payload.(type) {
+	case snet.SCMPEchoReply:
+	case snet.SCMPExternalInterfaceDown:
+		errDetail = []interface{}{"IA", s.IA, "interface", s.Interface}
+	case snet.SCMPInternalConnectivityDown:
+		errDetail = []interface{}{"IA", s.IA, "ingress", s.Ingress, "egress", s.Egress}
+	default:
+		errDetail = []interface{}{"type", common.TypeOf(pkt.Payload)}
 	}
+	if len(errDetail) > 0 {
+		return snet.SCMPEchoReply{}, serrors.New("not SCMPEchoReply", errDetail...)
+	}
+	r := pkt.Payload.(snet.SCMPEchoReply)
 	if r.Identifier != h.id {
 		return snet.SCMPEchoReply{}, serrors.New("wrong SCMP ID",
 			"expected", h.id, "actual", r.Identifier)
--- a/go/pkg/ping/ping.go
+++ b/go/pkg/ping/ping.go
@@ -304,18 +304,16 @@ func (h scmpHandler) handle(pkt *snet.Packet) (snet.SCMPEchoReply, error) {
 	if pkt.Payload == nil {
 		return snet.SCMPEchoReply{}, serrors.New("no v2 payload found")
 	}
-	var errDetail []interface{}
 	switch s := pkt.Payload.(type) {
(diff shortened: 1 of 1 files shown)


Oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4011)* status: all files reviewed, 3 unresolved discussions (waiting on @juagargi)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4011#-MXgZQmJ43qLOoecbaoL:-MXgZQmJ43qLOoecbaoM:busbq8u) (no related file):*
Could you please add some sample output of the `scion-pki ping` command after this change?

---
___
*[go/pkg/ping/ping.go, line 311 at r1](https://reviewable.io/reviews/scionproto/scion/4011#-MXgZgQ27KfqRuF0DQW3:-MXgZgQ27KfqRuF0DQW4:b-es22is) ([raw file](https://github.com/scionproto/scion/blob/31038e8d4e3295952072d602ed560930ca9f7874/go/pkg/ping/ping.go#L311)):*
> ```Go
> 	case snet.SCMPEchoReply:
> 	case snet.SCMPExternalInterfaceDown:
> 		errDetail = []interface{}{"IA", s.IA, "interface", s.Interface}
> ```

I think this would read nicer if the error message is dependent on the reply, not just the details.

Something like:

```
case snet.SCMPExternalInterfaceDown:
    return snet.SCMPEchoReply{}, serrors.New("external interface down", "isd_as", s.IA, "interface", s.Interface)
case snet.SCMPInternalConnectivityDown:
    return snet.SCMPEchoReply{}, serrors.New("internal connectivity down", "isd_as", s.IA, ...)
```
___
*[go/pkg/ping/ping.go, line 313 at r1](https://reviewable.io/reviews/scionproto/scion/4011#-MXg_9B7AX3CSzkBflot:-MXg_9B7AX3CSzkBflou:biiurkr) ([raw file](https://github.com/scionproto/scion/blob/31038e8d4e3295952072d602ed560930ca9f7874/go/pkg/ping/ping.go#L313)):*
> ```Go
> 		errDetail = []interface{}{"IA", s.IA, "interface", s.Interface}
> 	case snet.SCMPInternalConnectivityDown:
> 		errDetail = []interface{}{"IA", s.IA, "ingress", s.Ingress, "egress", s.Egress}
> ```

we tend to use snake case for logging and error context keys.
Furthermore, we try to use ISD-AS (or isd_as) everywhere that is user facing because `IA` is never properly defined (outside of the codebase)


`IA` should be `isd_as`here.

Oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4011)* status: all files reviewed, 3 unresolved discussions (waiting on @juagargi)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4011#-MXgZQmJ43qLOoecbaoL:-MXgbTJxBcEMnk2_eMWr:b-vojsp3) (no related file):*

Could you please add some sample output of the `scion-pki ping` command after this change?
</blockquote></details>

uff, I mean `scion ping` of course :facepalm: 

---

juagargi (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4011)* status: all files reviewed, 3 unresolved discussions (waiting on @juagargi and @Oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4011#-MXgZQmJ43qLOoecbaoL:-MXgdG1h13IUz426sRko:b-vflv3) (no related file):*

uff, I mean `scion ping` of course :facepalm: 
</blockquote></details>

For instance, when pinging certain SCIONLab user AS sometimes we get this message:
```
ERROR: not SCMPEchoReply IA="17-ffaa:0:1107" interface="137"
```

---
___
*[go/pkg/ping/ping.go, line 311 at r1](https://reviewable.io/reviews/scionproto/scion/4011#-MXgZgQ27KfqRuF0DQW3:-MXgdhnSBqnNCc9r25ve:b-h811nx) ([raw file](https://github.com/scionproto/scion/blob/31038e8d4e3295952072d602ed560930ca9f7874/go/pkg/ping/ping.go#L311)):*

I think this would read nicer if the error message is dependent on the reply, not just the details.

Something like:

```
case snet.SCMPExternalInterfaceDown:
    return snet.SCMPEchoReply{}, serrors.New("external interface down", "isd_as", s.IA, "interface", s.Interface)
case snet.SCMPInternalConnectivityDown:
    return snet.SCMPEchoReply{}, serrors.New("internal connectivity down", "isd_as", s.IA, ...)
```



</blockquote></details>

I had it ~like~ similar to that beforehand, and it was definitely nicer to read. Changed it.

juagargi (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4011)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @Oncilla)

___
*[go/pkg/ping/ping.go, line 313 at r1](https://reviewable.io/reviews/scionproto/scion/4011#-MXg_9B7AX3CSzkBflot:-MXgfa6SCd54AqXC5WFh:bf3i65e) ([raw file](https://github.com/scionproto/scion/blob/31038e8d4e3295952072d602ed560930ca9f7874/go/pkg/ping/ping.go#L313)):*

we tend to use snake case for logging and error context keys.
Furthermore, we try to use ISD-AS (or isd_as) everywhere that is user facing because `IA` is never properly defined (outside of the codebase)


`IA` should be `isd_as`here.

</blockquote></details>

Yes, I saw that in your sample above (and `interface`, etc). It should be okay now.

Oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4011#-:-MXghNI_81Pkm0cNnik9:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @juagargi)

PR #4020, opened 2021-04-21
lib/underlay/conn: remove unused ReadMeta, cleanup

matzf:
The ReadMeta information (timestamp, number of packets dropped due to
receive buffer overflow) is no longer used by the router.
Remove the related sockopts and cmsg parsing logic. Remove allocation
and filling of the (unused) ReadMeta for ReadBatch.

The number of packets dropped had still been logged by the dispatcher.
As far as I can tell from the commit/issue history, this had been added
mainly because it was already available from the shared code with the router.
Note: the same counter information also available externally e.g. via
`/proc/net/udp` (column "drops") or aggregated over all sockets in
`netstat -su` as "receive buffer errors" or "RcvbufErrors".

Furthermore:
- simplify underlay/conn.Conn.Read to "standard" ReadFrom
- cleanup ReceiveBufferSize config:
  - make config non-optional and move constant to router
  - the `overlay.conn.sizeIgnore` flag allowed to handle a failure
    to set the requested receive buffer size as fatal instead of only
    logging it.
    It was no longer possible to set this flag on the command line.
    Removed the flag and only kept the default behaviour (log the
    failure).
  - fix minor bugs in checks (used constant instead of config value)
- remove unused NewWriteMessages

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4020)


lukedirtwalker (approved):
> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4020)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4022, opened 2021-04-21
launcher: add cpuprofile option

matzf:
Add an option for simple CPU profiling to the launcher library, and thus to all services.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4022)


matzf:
As @Oncilla pointed out to me, CPU profiling is already readily available and exposed through the debug API. See #4024

PR #4029, opened 2021-04-23
router: avoid per-packet allocations for MAC

matzf:
Avoid creating a separate CMAC hasher, each with a separate AES block
cipher object, for each MAC computation/verification.
This was responsible for a very large portion of allocations during
forwarding. Additionally, reusing the hasher amortizes the cost of the
internal initialization of both the AES block cipher (key expansion) and
CMAC (subkey derivation).
The CMAC hasher is not goroutine safe, so care must be taken to not
share it by accident.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4029)

PR #4030, opened 2021-04-27
router: reduce per-packet allocations

matzf:
Avoid allocating scionPacketProcessor; re-organized packet processing
functions. The scionPacketProcessor is now the main entry point for
packet processing and contains all the pre-allocated per-packet state.

Avoid allocation of message buffer for writing with WriteBatch.
We could just pre-allocate a buffer, but as we always write a batch of
size one, just use WriteTo instead.
For writing single packets, this is both more convenient and faster.
Additional benefit: Read/WriteBatch make rather large allocations
internally (see golang/go#26838) which is
avoided by using WriteTo.

The largest remaining allocations when processing SCION packets are:
- ReadBatch allocates temporary storage internally (as mentioned above).
  This is by far the biggest culprit, it accounts for almost 80% (after
  the changes in this patch).
- slayers/path.NewPath called during slayers.SCION.DecodeFromBytes:
  Could be avoided by reusing Path objects, but requires larger changes.
- slayers/path/scion.Raw.GetCurrentHopField and GetCurrentInfoField, called during parsePath:
  Both return pointer types. Should be possible to change this to value
  types instead, but requires larger changes.
- DataPlane.resolveLocalDst: allocates IP and UDPAddr. Can only really
  be fixed with better address types.
- slayers/path.FullMAC. Allocates both input and output. The function
  interface could be changed to allow reuse of the input buffer.
  The output buffer cannot be reused with this CMAC library.

Depends on (or actually includes) #4029.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4030)

### DIFF
--- a/go/pkg/router/BUILD.bazel
+++ b/go/pkg/router/BUILD.bazel
@@ -33,7 +33,6 @@ go_library(
         "@com_github_google_gopacket//layers:go_default_library",
         "@com_github_prometheus_client_golang//prometheus:go_default_library",
         "@com_github_prometheus_client_golang//prometheus/promauto:go_default_library",
-        "@org_golang_x_net//ipv4:go_default_library",
     ],
 )
 
--- a/go/pkg/router/dataplane.go
+++ b/go/pkg/router/dataplane.go
@@ -77,7 +77,6 @@ type bfdSession interface {
 type BatchConn interface {
 	ReadBatch(underlayconn.Messages) (int, error)
 	WriteTo([]byte, *net.UDPAddr) (int, error)
-	WriteBatch(underlayconn.Messages) (int, error)
 	Close() error
 }
 
--- a/go/pkg/router/dataplane.go
+++ b/go/pkg/router/dataplane.go
@@ -483,7 +483,10 @@ func (d *DataPlane) Run() error {
 				if result.OutConn == nil { // e.g. BFD case no message is forwarded
 					continue
 				}
-				outAddr, _ := result.OutAddr.(*net.UDPAddr)
+				var outAddr *net.UDPAddr
+				if result.OutAddr != nil {
+					outAddr = result.OutAddr.(*net.UDPAddr)
+				}
 				_, err = result.OutConn.WriteTo(result.OutPkt, outAddr)
 				if err != nil {
 					log.Debug("Error writing packet", "err", err)
--- a/go/pkg/router/dataplane.go
+++ b/go/pkg/router/dataplane.go
@@ -93,7 +93,7 @@ type DataPlane struct {
 	neighborIAs       map[uint16]addr.IA
 	internal          BatchConn
 	internalIP        net.IP
-	internalNextHops  map[uint16]net.Addr
+	internalNextHops  map[uint16]*net.UDPAddr
 	svc               *services
 	macFactory        func() hash.Hash
 	bfdSessions       map[uint16]bfdSession
@@ -362,7 +362,7 @@ func (d *DataPlane) DelSvc(svc addr.HostSVC, a *net.UDPAddr) error {
 // AddNextHop sets the next hop address for the given interface ID. If the
 // interface ID already has an address associated this operation fails. This can
 // only be called on a not yet running dataplane.
-func (d *DataPlane) AddNextHop(ifID uint16, a net.Addr) error {
+func (d *DataPlane) AddNextHop(ifID uint16, a *net.UDPAddr) error {
 	d.mtx.Lock()
 	defer d.mtx.Unlock()
 	if d.running {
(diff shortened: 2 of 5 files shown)


lukedirtwalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4030)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[go/pkg/router/dataplane.go, line 80 at r1](https://reviewable.io/reviews/scionproto/scion/4030#-MZroS3wCsSeAzrNUSDO:-MZroS3wCsSeAzrNUSDP:bvshln6) ([raw file](https://github.com/scionproto/scion/blob/84a97eda022090123efccab72202142eafe93fb6/go/pkg/router/dataplane.go#L80)):*
> ```Go
> 	ReadBatch(underlayconn.Messages) (int, error)
> 	WriteTo([]byte, *net.UDPAddr) (int, error)
> 	WriteBatch(underlayconn.Messages) (int, error)
> ```

this no longer seems to be used, please remove and update the mock.
___
*[go/pkg/router/dataplane.go, line 487 at r1](https://reviewable.io/reviews/scionproto/scion/4030#-MZrofc72JM1nb08IDNH:-MZrofc72JM1nb08IDNI:bhb1g65) ([raw file](https://github.com/scionproto/scion/blob/84a97eda022090123efccab72202142eafe93fb6/go/pkg/router/dataplane.go#L487)):*
> ```
> , _
> ```
Please check the cast succeeds and if it doesn't the packet should be discarded. I think this address comes from user input so we can not simply assume that the cast works.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4030)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker)

___
*[go/pkg/router/dataplane.go, line 80 at r1](https://reviewable.io/reviews/scionproto/scion/4030#-MZroS3wCsSeAzrNUSDO:-MZrtyk9AaFknx88mX2F:by1r7bn) ([raw file](https://github.com/scionproto/scion/blob/84a97eda022090123efccab72202142eafe93fb6/go/pkg/router/dataplane.go#L80)):*

this no longer seems to be used, please remove and update the mock.
</blockquote></details>

Done. I've not renamed it although the name doesn't seem very apt now. I cannot think of anything that's really better, both "Conn" or "UnderlayConn" would probably be net negative in clarity. The doc string is technically still correct: 'supports "batch reads" and "writes".' :wink: 
___
*[go/pkg/router/dataplane.go, line 487 at r1](https://reviewable.io/reviews/scionproto/scion/4030#-MZrofc72JM1nb08IDNH:-MZrtajQ89PGBDxtOFxp:bienmc8) ([raw file](https://github.com/scionproto/scion/blob/84a97eda022090123efccab72202142eafe93fb6/go/pkg/router/dataplane.go#L487)):*

> ```
> , _
> ```
Please check the cast succeeds and if it doesn't the packet should be discarded. I think this address comes from user input so we can not simply assume that the cast works.
</blockquote></details>

I've checked the different code paths setting this, the address is either nil (uninitialized) or UDPAddr. I've changed this to use a panic-y type assertion to make this expectation explicit.

lukedirtwalker (changes requested):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4030)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[go/pkg/router/dataplane.go, line 487 at r1](https://reviewable.io/reviews/scionproto/scion/4030#-MZrofc72JM1nb08IDNH:-MZv1zM8-KFPhi-JOFSA:b1qx25q) ([raw file](https://github.com/scionproto/scion/blob/84a97eda022090123efccab72202142eafe93fb6/go/pkg/router/dataplane.go#L487)):*

I've checked the different code paths setting this, the address is either nil (uninitialized) or UDPAddr. I've changed this to use a panic-y type assertion to make this expectation explicit.
</blockquote></details>

Hm then it would be cleaner to change the type of result.OutAddr, if that isn't too much work.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4030)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

___
*[go/pkg/router/dataplane.go, line 487 at r1](https://reviewable.io/reviews/scionproto/scion/4030#-MZrofc72JM1nb08IDNH:-MZwDQFW2lgx7d09YIS4:b-v3254i) ([raw file](https://github.com/scionproto/scion/blob/84a97eda022090123efccab72202142eafe93fb6/go/pkg/router/dataplane.go#L487)):*

Hm then it would be cleaner to change the type of result.OutAddr, if that isn't too much work.
</blockquote></details>

Done. This goes through quite a few places, now UDPAddr is now ubuquitous. I quite like that, though. I think it makes the previously hidden assumptions more clearly visible.

lukedirtwalker (approved):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4030)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4031, opened 2021-04-28
router: async read/write

matzf:
Decouple reading and processing input packets from writing messages back
out. This removes the blocking write from the reader loop. Output
messages are transferred from the reader to the writer asynchronously,
via a buffered channel.
If the writer is the slowest part, packets will now be dropped
internally -- previously, the blocking write would have caused the
router to stop reading and the packets to be eventually dropped by the
OS as it ran out of UDP rmem.

Depends on / includes #4030.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4031)

### DIFF
--- a/go/pkg/router/dataplane.go
+++ b/go/pkg/router/dataplane.go
@@ -45,7 +45,6 @@ import (
 	"github.com/scionproto/scion/go/lib/slayers/path/onehop"
 	"github.com/scionproto/scion/go/lib/slayers/path/scion"
 	"github.com/scionproto/scion/go/lib/topology"
-	"github.com/scionproto/scion/go/lib/underlay/conn"
 	underlayconn "github.com/scionproto/scion/go/lib/underlay/conn"
 	"github.com/scionproto/scion/go/lib/util"
 	"github.com/scionproto/scion/go/pkg/router/bfd"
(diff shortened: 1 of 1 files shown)


matzf:
In my test setup, this change does not improve the performance as much as I had hoped, in fact it's pretty much identical to the results with only #4030. I believe this approach with asynchronous writes could be beneficial in a setup where there is a large difference between the speeds of the different interfaces. 
There is obviously an overhead both in terms of complexity and the runtime cost of juggling the buffers. I'm putting this PR out here mainly to get some feedback from you all; do you think that it may be worthwhile to continue with this approach here? Or was the simpler approach a conscious decision to keep the implementation complexity down?

matzf:
Just a thought: the same overall "effect" of asynchronous writes could be achieved without any changes to the structure or logic of the router itself if we could use the semantics of `MSG_DONTWAIT` for our "send" calls. Unfortunately, this is not supported by the x/net library -- I think it would be (relatively) easy to add support for something like this, but I'm not sure whether this would be considered a good idea. I may give it a try and ask.

matzf:
Fixed by using `MSG_DONTWAIT` as described above in #4096.

PR #4014, opened 2021-04-09
doc: add Authenticator End-to-End option

matzf:
The Authenticator End-to-End option roughly follows what was previously
described in the SCION book as the "SCION Packet Security Extension". As
the name implies, however, this does not intend to be applicable for
packet encryption -- I think that encrypted packets can be more
appropriately represented by a separate next-header type analogous to IP
Encapsulating Security Payload.

This Authenticator option will be used for authenticated SCMP (to be described in a separate document, see #3969).

[doc]

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4014)

PR #4036, opened 2021-05-07
slayers: add Authenticator end-to-end option

matzf:
Add Authenticator end-to-end option, as described in doc/protocols/authenticator-option.rst (see #4014).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4036)

### DIFF
--- a/go/lib/slayers/extn.go
+++ b/go/lib/slayers/extn.go
@@ -360,10 +360,7 @@ func (e *EndToEndExtn) FindOption(typ OptionType) (*EndToEndOption, error) {
 type PacketAuthAlg uint8
 
 const (
-	PacketAuthCMAC       PacketAuthAlg = 0
-	PacketAuthGMAC       PacketAuthAlg = 1
-	PacketAuthHMACSHA256 PacketAuthAlg = 16
-	PacketAuthEd25519    PacketAuthAlg = 32
+	PacketAuthCMAC PacketAuthAlg = 0
 )
 
 type PacketAuthenticatorOption struct {
--- a/go/lib/slayers/extn.go
+++ b/go/lib/slayers/extn.go
@@ -360,10 +360,10 @@ func (e *EndToEndExtn) FindOption(typ OptionType) (*EndToEndOption, error) {
 type PacketAuthAlg uint8
 
 const (
-	E2EAuthCMAC       PacketAuthAlg = 0
-	E2EAuthGMAC       PacketAuthAlg = 1
-	E2EAuthHMACSHA256 PacketAuthAlg = 16
-	E2EAuthEd25519    PacketAuthAlg = 32
+	PacketAuthCMAC       PacketAuthAlg = 0
+	PacketAuthGMAC       PacketAuthAlg = 1
+	PacketAuthHMACSHA256 PacketAuthAlg = 16
+	PacketAuthEd25519    PacketAuthAlg = 32
 )
 
 type PacketAuthenticatorOption struct {
--- a/go/lib/slayers/extn.go
+++ b/go/lib/slayers/extn.go
@@ -31,6 +31,7 @@ type OptionType uint8
 const (
 	OptTypePad1 OptionType = iota
 	OptTypePadN
+	OptTypeAuthenticator
 )
 
 type tlvOption struct {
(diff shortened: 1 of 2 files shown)


matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4036)* status: all files reviewed, 5 unresolved discussions (waiting on @matzf and @shitz)

___
*[go/lib/slayers/extn.go, line 374 at r1](https://reviewable.io/reviews/scionproto/scion/4036#-M_R9s0o2YqthUIjns6E:-M_UsmxtAnHGlHexLARa:b-gkr2fd) ([raw file](https://github.com/scionproto/scion/blob/1e15a85cc658f772634dc213144ae23b34acba8d/go/lib/slayers/extn.go#L374)):*

Why make this part of the type? I'd prefer
`func NewPacketAuthenticatorOption(alg, data) *EndToEndOption`
</blockquote></details>

Sure, that would seem more natural. The idea was that this allows to reuse the `EndToEndOption` and so reuse the internal buffer. When returning a new `EndToEndOption` there is no way to reuse the option or avoid the allocation for the internal buffer.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4036)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @matzf and @shitz)

___
*[go/lib/slayers/extn.go, line 374 at r1](https://reviewable.io/reviews/scionproto/scion/4036#-M_R9s0o2YqthUIjns6E:-M_UwKmp7LUDUoKOiCCT:bbchqit) ([raw file](https://github.com/scionproto/scion/blob/1e15a85cc658f772634dc213144ae23b34acba8d/go/lib/slayers/extn.go#L374)):*

Sure, that would seem more natural. The idea was that this allows to reuse the `EndToEndOption` and so reuse the internal buffer. When returning a new `EndToEndOption` there is no way to reuse the option or avoid the allocation for the internal buffer.
</blockquote></details>

I had a simple idea; with a small wrapper type it seems we can get both a simple `NewPacketAuthenticatorOption` and also conveniently reuse the buffer later with a `Reset` method on this `PacketAuthenticatorOption` type -- even better, we can have a method that returns the slice for the Authenticator data so we could even write to this directly. It also makes the parsing nicer as it can now return the same wrapper type instead of the individual fields.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4036)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @shitz)

___
*[go/lib/slayers/extn.go, line 354 at r1](https://reviewable.io/reviews/scionproto/scion/4036#-M_R8W1m2R7LZEgc5P2q:-M_V0_0M8CiYRTU0dVlx:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1e15a85cc658f772634dc213144ae23b34acba8d/go/lib/slayers/extn.go#L354)):*

I suggest a slightly different API:

```
var ErrOptionNotFound = serrors.New("Option not found")

func (e *EndToEndExtn) FindOption(typ OptionType) (*EndToEndOption, error) {
	for _, o := range e.Options {
		if o.OptType == typ {
			return o, nil
		}
	}
	return nil, ErrOptionNotFound
}
```

This can then be used for any option. 
We could drop it entirely, since it doesn't prevent a whole lot of boilerplate.


</blockquote></details>

Done.
___
*[go/lib/slayers/extn.go, line 363 at r1](https://reviewable.io/reviews/scionproto/scion/4036#-M_RAny28tZ4_nBeCF1w:-M_V0b7Z9M2dL3yocmtq:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1e15a85cc658f772634dc213144ae23b34acba8d/go/lib/slayers/extn.go#L363)):*

`PacketAuthAlg` would be more concise
</blockquote></details>

Done.
___
*[go/lib/slayers/extn.go, line 394 at r1](https://reviewable.io/reviews/scionproto/scion/4036#-M_RAKgh70uDFR2gFTTR:-M_VNkNf9baAfSh8Uwhf:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1e15a85cc658f772634dc213144ae23b34acba8d/go/lib/slayers/extn.go#L394)):*

Same as above. Would make this a separate function (and public) that takes an `*EndToEndOption`
</blockquote></details>

Done.
___
*[go/lib/slayers/extn.go, line 396 at r1](https://reviewable.io/reviews/scionproto/scion/4036#-M_RAT6f0N8duGiQlGdN:-M_VNjYs9-Py6SRsfdr8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1e15a85cc658f772634dc213144ae23b34acba8d/go/lib/slayers/extn.go#L396)):*

Return an error instead of Panicking, especially if you make it public.
</blockquote></details>

Done.

shitz (changes requested):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4036)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[go/lib/slayers/extn.go, line 374 at r1](https://reviewable.io/reviews/scionproto/scion/4036#-M_R9s0o2YqthUIjns6E:-M_WBuyeEc1XB74qe0y_:bco11k8) ([raw file](https://github.com/scionproto/scion/blob/1e15a85cc658f772634dc213144ae23b34acba8d/go/lib/slayers/extn.go#L374)):*

I had a simple idea; with a small wrapper type it seems we can get both a simple `NewPacketAuthenticatorOption` and also conveniently reuse the buffer later with a `Reset` method on this `PacketAuthenticatorOption` type -- even better, we can have a method that returns the slice for the Authenticator data so we could even write to this directly. It also makes the parsing nicer as it can now return the same wrapper type instead of the individual fields. 
</blockquote></details>

Yup I like it. :+1: 
___
*[go/lib/slayers/extn.go, line 410 at r2](https://reviewable.io/reviews/scionproto/scion/4036#-M_WC5upE7uNRb0011Hr:-M_WC5upE7uNRb0011Hs:bkt97j3) ([raw file](https://github.com/scionproto/scion/blob/015784f9c85e91437ff99dfb7752088eac2cd635/go/lib/slayers/extn.go#L410)):*
> ```Go
> }
> 
> func (o PacketAuthenticatorOption) Authenticator() []byte {
> ```

Add a docstring explaining that this will return the underlying buffer and that changing it will be reflected on the wire once the extension gets serialized. I'm not really sure if there is a usecase for writing directly to this buffer (maybe when reverting a packet and updating the authenticator?). In any case, writing to the buffer should be done with care.

shitz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4036#-:-M_eLqfh48FkjTIwQ47t:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[go/lib/slayers/extn.go, line 377 at r3](https://reviewable.io/reviews/scionproto/scion/4036#-M_eIIFvAE0cdUpfoDNP:-M_eIIFw1VO7R2bCuhrL:b-v0e277) ([raw file](https://github.com/scionproto/scion/blob/7d44252e6f64caba1b2af2b07ad43a0dc3eb8c0e/go/lib/slayers/extn.go#L377)):*
> ```
> EndTOEndOption
> ```
nit - EndToEndOption

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4036)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @shitz)

___
*[go/lib/slayers/extn.go, line 410 at r2](https://reviewable.io/reviews/scionproto/scion/4036#-M_WC5upE7uNRb0011Hr:-M_dxW1u8nVcwZIirRRa:b-ece4xg) ([raw file](https://github.com/scionproto/scion/blob/015784f9c85e91437ff99dfb7752088eac2cd635/go/lib/slayers/extn.go#L410)):*

Add a docstring explaining that this will return the underlying buffer and that changing it will be reflected on the wire once the extension gets serialized. I'm not really sure if there is a usecase for writing directly to this buffer (maybe when reverting a packet and updating the authenticator?). In any case, writing to the buffer should be done with care.
</blockquote></details>

Done, added docs.

The use case I had in mind is for authenticated SCMP reply messages created in the router: we can create the PacketAuthenticatorOption once and write the MAC directly to this buffer. It's probably not a big win, but still nice.
___
*[go/lib/slayers/extn.go, line 377 at r3](https://reviewable.io/reviews/scionproto/scion/4036#-M_eIIFvAE0cdUpfoDNP:-M_esXaq9JznqJpxHopb:bo3vl6y) ([raw file](https://github.com/scionproto/scion/blob/7d44252e6f64caba1b2af2b07ad43a0dc3eb8c0e/go/lib/slayers/extn.go#L377)):*

> ```
> EndTOEndOption
> ```
nit - EndToEndOption
</blockquote></details>

Thanks, done.

PR #4041, opened 2021-05-11
cleanup: remove/fix dead code and unused definitions

matzf:
Remove numerous unused functions, types, variables and constants.

The vast majority is just trivial, left-over stuff that is no longer needed.
Exceptions:
- in go/pkg/gateway/control/watcher.go, prefixWatcher.runOnceCheck, locking
  the runMarkerLock appeared to be forgotten by accident.
- in go/cs/config/config_test.go, InitTestPSConfig was not added to
  InitTestConfig during some refactoring.
- in go/cs/beaconing/propagator_test.go, the list of inactive interfaces
  is no longer considered in the tests because this information is
  simply no longer available in the CS (BFD instead of keep alives
  between CSes). Only two distinct test cases are left.
- additionally, in go/pkg/ca/renewal/request_test.go the variable
  csrTmplGeneva was unused; there appears to be a copy-paste mistake in
  this will be fixed in a separate PR as it may require recreating the
  test data.

Found with `golangci-lint run -p unused -D unparam ./go/...`.
We should consider enabling a similar dead code linter in CI or in nogo.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4041)


lukedirtwalker (approved):
> Reviewed 43 of 43 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4041)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4041#-MaYauzQA_dFoorkg1X7:-MaYauzQA_dFoorkg1X8:b-o5yq45) (no related file):*
good stuff, thank you

---

PR #4049, opened 2021-05-19
slayers: check for repetition of extension headers

matzf:
The hop-by-hop and end-to-end extension headers can occur at most once
each in a scion packet, and they must be in the correct order, first
HBH, then E2E.
Both the serialization and decode logic had only checked the order
constraint, i.e. it checked that NextHdr of E2E is not HBH. There
was no check for repeated headers.

Note: this check is important not only for adherence to the protocol specifications, but also because gopacket's DecodingLayerParser cannot sensibly handle packets with repeated layer types. 

Fixed in two places:
- split up the scionNextLayerType helper function into three related but
  separate parts: scionNextLayerType (for use in the SCION base packet),
  scionNextLayerTypeAfterHBH and scionNextLayerTypeAfterE2E.
  These functions now don't allow the transition back to the same layer.

  golang trick: this allows to move the registration of the layer type
  to a variable instead of init, as it breaks the initialization cycle.
  We get a compile time guarantee that none of the layers is its own
  NextLayerType.

- check NextHdr for repetition in both serialize and decode

The first part is not strictly necessary, but I think that it helps to
clarify what's going on.

Also:
- fix extension protocol identifier in some places in the docs
- fix protocol identifier string representation for HopByHop extension

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4049)

### DIFF
--- a/doc/protocols/extension-header.rst
+++ b/doc/protocols/extension-header.rst
@@ -16,7 +16,7 @@ Hop-by-Hop Options Header
 
 The Hop-by-Hop Options header is used to carry optional information that may be
 examined and processed by every node along a packet's delivery path. The
-Hop-by-Hop Options header is identified by a Next Header value of ``0`` in the
+Hop-by-Hop Options header is identified by a Next Header value of ``200`` in the
 SCION common header and has the following format::
 
      0                   1                   2                   3
@@ -63,7 +63,7 @@ End-to-End Options Header
 
 The End-to-end  Options header is used to carry optional information that may be
 examined and processed by sender and/or receiver of the packet.  The End-to-end
-Options header is identified by a Next Header value of ``0xfd`` in the SCION
+Options header is identified by a Next Header value of ``201`` in the SCION
 common header and has the following format::
 
      0                   1                   2                   3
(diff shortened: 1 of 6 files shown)


matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4049)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker)

___
*[go/lib/slayers/extn_test.go, line 269 at r1](https://reviewable.io/reviews/scionproto/scion/4049#-Macz80TDz1gGge4GUCo:-Mad4WSNAj-hgmAh91jT:b4855pb) ([raw file](https://github.com/scionproto/scion/blob/883bf7ec83433cd986849696c6fe1da41065139f/go/lib/slayers/extn_test.go#L269)):*

> ```
> e2e, hbh}
> ```
this should be `e2e, e2e` to match the title
</blockquote></details>

Thanks! Done.
___
*[go/lib/slayers/extn_test.go, line 311 at r1](https://reviewable.io/reviews/scionproto/scion/4049#-Maczu8l5tRRDfcap6xR:-Mad4Yrj8DYBB-HT2Alw:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/883bf7ec83433cd986849696c6fe1da41065139f/go/lib/slayers/extn_test.go#L311)):*

> ```
> 			} else {
> 				if packet.ErrorLayer() != nil {
> ```
this could be combined in to an `else if`
</blockquote></details>

Done.

PR #4053, opened 2021-05-21
border router: avoid MAC allocations at runtime

mawyss:
This PR extends https://github.com/scionproto/scion/pull/4030/ with the suggested preallocation of the
SCION MAC input buffer. For this, FullMAC is changed to accept
the input buffer as argument.
Also the EPIC MAC buffers are now preallocated (both input and
output buffers, and the zero initialization vector). Only the AES-
cipher is still allocated for each packet. This is however
inherent to EPIC, because every packet can potentially have
a different authenticator (= key for the AES cipher).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)

### DIFF
--- a/go/cs/beaconing/extender.go
+++ b/go/cs/beaconing/extender.go
@@ -247,7 +247,8 @@ func (s *DefaultExtender) createHopF(ingress, egress uint16, ts time.Time,
 	beta uint16) (path.HopField, []byte) {
 
 	expTime := s.MaxExpTime()
-	input := path.MACInput(beta, util.TimeToSecs(ts), expTime, ingress, egress)
+	input := make([]byte, path.MACBufferSize)
+	path.MACInput(beta, util.TimeToSecs(ts), expTime, ingress, egress, input)
 
 	mac := s.MAC()
 	// Write must not return an error: https://godoc.org/hash#Hash
--- a/go/cs/beaconing/extender.go
+++ b/go/cs/beaconing/extender.go
@@ -247,8 +247,7 @@ func (s *DefaultExtender) createHopF(ingress, egress uint16, ts time.Time,
 	beta uint16) (path.HopField, []byte) {
 
 	expTime := s.MaxExpTime()
-	input := make([]byte, path.MACBufferSize)
-	path.MACInput(beta, util.TimeToSecs(ts), expTime, ingress, egress, input)
+	input := path.MACInput(beta, util.TimeToSecs(ts), expTime, ingress, egress, nil)
 
 	mac := s.MAC()
 	// Write must not return an error: https://godoc.org/hash#Hash
--- a/go/cs/beaconing/extender.go
+++ b/go/cs/beaconing/extender.go
@@ -247,7 +247,8 @@ func (s *DefaultExtender) createHopF(ingress, egress uint16, ts time.Time,
 	beta uint16) (path.HopField, []byte) {
 
 	expTime := s.MaxExpTime()
-	input := path.MACInput(beta, util.TimeToSecs(ts), expTime, ingress, egress, nil)
+	input := make([]byte, path.MACBufferSize)
+	path.MACInput(beta, util.TimeToSecs(ts), expTime, ingress, egress, input)
 
 	mac := s.MAC()
 	// Write must not return an error: https://godoc.org/hash#Hash
--- a/go/integration/braccept/cases/bfd.go
+++ b/go/integration/braccept/cases/bfd.go
@@ -90,7 +90,12 @@ func ExternalBFD(artifactsDir string, mac hash.Hash) runner.Case {
 			ConsEgress:  131,
 		},
 	}
-	ohp.FirstHop.Mac = path.MAC(mac, &ohp.Info, &ohp.FirstHop)
+	var err error
+	ohp.FirstHop.Mac, err = path.MAC(mac, &ohp.Info, &ohp.FirstHop,
+		make([]byte, path.MACBufferSize))
+	if err != nil {
+		panic(err)
+	}
 	scionL := &slayers.SCION{
 		Version:      0,
 		TrafficClass: 0xb8,
@@ -101,7 +106,7 @@ func ExternalBFD(artifactsDir string, mac hash.Hash) runner.Case {
 		DstIA:        localIA,
 		SrcIA:        remoteIA,
 	}
-	err := scionL.SetSrcAddr(&net.IPAddr{IP: net.IP{192, 168, 13, 3}})
+	err = scionL.SetSrcAddr(&net.IPAddr{IP: net.IP{192, 168, 13, 3}})
 	if err != nil {
 		panic(err)
 	}
--- a/go/integration/braccept/cases/bfd.go
+++ b/go/integration/braccept/cases/bfd.go
@@ -90,12 +90,7 @@ func ExternalBFD(artifactsDir string, mac hash.Hash) runner.Case {
 			ConsEgress:  131,
 		},
 	}
-	var err error
-	ohp.FirstHop.Mac, err = path.MAC(mac, &ohp.Info, &ohp.FirstHop,
-		make([]byte, path.MACBufferSize))
-	if err != nil {
-		panic(err)
-	}
+	ohp.FirstHop.Mac = path.MAC(mac, &ohp.Info, &ohp.FirstHop, nil)
 	scionL := &slayers.SCION{
 		Version:      0,
 		TrafficClass: 0xb8,
@@ -106,7 +101,7 @@ func ExternalBFD(artifactsDir string, mac hash.Hash) runner.Case {
 		DstIA:        localIA,
 		SrcIA:        remoteIA,
 	}
-	err = scionL.SetSrcAddr(&net.IPAddr{IP: net.IP{192, 168, 13, 3}})
+	err := scionL.SetSrcAddr(&net.IPAddr{IP: net.IP{192, 168, 13, 3}})
 	if err != nil {
 		panic(err)
 	}
--- a/go/integration/braccept/cases/child_to_child_xover.go
+++ b/go/integration/braccept/cases/child_to_child_xover.go
@@ -93,9 +93,18 @@ func ChildToChildXover(artifactsDir string, mac hash.Hash) runner.Case {
 			{ConsIngress: 411, ConsEgress: 0},
 		},
 	}
-	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1])
+	var err error
+	sp.HopFields[1].Mac, err = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1],
+		make([]byte, path.MACBufferSize))
+	if err != nil {
+		panic(err)
+	}
 	sp.InfoFields[0].UpdateSegID(sp.HopFields[1].Mac)
-	sp.HopFields[2].Mac = path.MAC(mac, sp.InfoFields[1], sp.HopFields[2])
+	sp.HopFields[2].Mac, err = path.MAC(mac, sp.InfoFields[1], sp.HopFields[2],
+		make([]byte, path.MACBufferSize))
+	if err != nil {
+		panic(err)
+	}
 
 	scionL := &slayers.SCION{
 		Version:      0,
--- a/go/integration/braccept/cases/child_to_child_xover.go
+++ b/go/integration/braccept/cases/child_to_child_xover.go
@@ -93,18 +93,9 @@ func ChildToChildXover(artifactsDir string, mac hash.Hash) runner.Case {
 			{ConsIngress: 411, ConsEgress: 0},
 		},
 	}
-	var err error
-	sp.HopFields[1].Mac, err = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1],
-		make([]byte, path.MACBufferSize))
-	if err != nil {
-		panic(err)
-	}
+	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1], nil)
 	sp.InfoFields[0].UpdateSegID(sp.HopFields[1].Mac)
-	sp.HopFields[2].Mac, err = path.MAC(mac, sp.InfoFields[1], sp.HopFields[2],
-		make([]byte, path.MACBufferSize))
-	if err != nil {
-		panic(err)
-	}
+	sp.HopFields[2].Mac = path.MAC(mac, sp.InfoFields[1], sp.HopFields[2], nil)
 
 	scionL := &slayers.SCION{
 		Version:      0,
--- a/go/integration/braccept/cases/child_to_internal.go
+++ b/go/integration/braccept/cases/child_to_internal.go
@@ -84,12 +84,7 @@ func ChildToInternalHost(artifactsDir string, mac hash.Hash) runner.Case {
 			{ConsIngress: 0, ConsEgress: 141},
 		},
 	}
-	var err error
-	sp.HopFields[1].Mac, err = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1],
-		make([]byte, path.MACBufferSize))
-	if err != nil {
-		panic(err)
-	}
+	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1], nil)
 	sp.InfoFields[0].UpdateSegID(sp.HopFields[1].Mac)
 
 	scionL := &slayers.SCION{
@@ -196,12 +191,7 @@ func ChildToInternalHostShortcut(artifactsDir string, mac hash.Hash) runner.Case
 			{ConsIngress: 191, ConsEgress: 141},
 		},
 	}
-	var err error
-	sp.HopFields[1].Mac, err = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1],
-		make([]byte, path.MACBufferSize))
-	if err != nil {
-		panic(err)
-	}
+	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1], nil)
 	sp.InfoFields[0].UpdateSegID(sp.HopFields[1].Mac)
 
 	scionL := &slayers.SCION{
@@ -319,12 +309,7 @@ func ChildToInternalParent(artifactsDir string, mac hash.Hash) runner.Case {
 			{ConsIngress: 0, ConsEgress: 911},
 		},
 	}
-	var err error
-	sp.HopFields[1].Mac, err = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1],
-		make([]byte, path.MACBufferSize))
-	if err != nil {
-		panic(err)
-	}
+	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1], nil)
 	sp.InfoFields[0].UpdateSegID(sp.HopFields[1].Mac)
 
 	scionL := &slayers.SCION{
--- a/go/integration/braccept/cases/child_to_internal.go
+++ b/go/integration/braccept/cases/child_to_internal.go
@@ -84,7 +84,12 @@ func ChildToInternalHost(artifactsDir string, mac hash.Hash) runner.Case {
 			{ConsIngress: 0, ConsEgress: 141},
 		},
 	}
-	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1])
+	var err error
+	sp.HopFields[1].Mac, err = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1],
+		make([]byte, path.MACBufferSize))
+	if err != nil {
+		panic(err)
+	}
 	sp.InfoFields[0].UpdateSegID(sp.HopFields[1].Mac)
 
 	scionL := &slayers.SCION{
@@ -191,7 +196,12 @@ func ChildToInternalHostShortcut(artifactsDir string, mac hash.Hash) runner.Case
 			{ConsIngress: 191, ConsEgress: 141},
 		},
 	}
-	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1])
+	var err error
+	sp.HopFields[1].Mac, err = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1],
+		make([]byte, path.MACBufferSize))
+	if err != nil {
+		panic(err)
+	}
 	sp.InfoFields[0].UpdateSegID(sp.HopFields[1].Mac)
 
 	scionL := &slayers.SCION{
@@ -309,7 +319,12 @@ func ChildToInternalParent(artifactsDir string, mac hash.Hash) runner.Case {
 			{ConsIngress: 0, ConsEgress: 911},
 		},
 	}
-	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1])
+	var err error
+	sp.HopFields[1].Mac, err = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1],
+		make([]byte, path.MACBufferSize))
+	if err != nil {
+		panic(err)
+	}
 	sp.InfoFields[0].UpdateSegID(sp.HopFields[1].Mac)
 
 	scionL := &slayers.SCION{
--- a/go/integration/braccept/cases/child_to_parent.go
+++ b/go/integration/braccept/cases/child_to_parent.go
@@ -92,12 +92,7 @@ func ChildToParent(artifactsDir string, mac hash.Hash) runner.Case {
 			{ConsIngress: 0, ConsEgress: 311},
 		},
 	}
-	var err error
-	sp.HopFields[1].Mac, err = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1],
-		make([]byte, path.MACBufferSize))
-	if err != nil {
-		panic(err)
-	}
+	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1], nil)
 	sp.InfoFields[0].UpdateSegID(sp.HopFields[1].Mac)
 
 	scionL := &slayers.SCION{
--- a/go/integration/braccept/cases/child_to_parent.go
+++ b/go/integration/braccept/cases/child_to_parent.go
@@ -92,7 +92,12 @@ func ChildToParent(artifactsDir string, mac hash.Hash) runner.Case {
 			{ConsIngress: 0, ConsEgress: 311},
 		},
 	}
-	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1])
+	var err error
+	sp.HopFields[1].Mac, err = path.MAC(mac, sp.InfoFields[0], sp.HopFields[1],
+		make([]byte, path.MACBufferSize))
+	if err != nil {
+		panic(err)
+	}
 	sp.InfoFields[0].UpdateSegID(sp.HopFields[1].Mac)
 
 	scionL := &slayers.SCION{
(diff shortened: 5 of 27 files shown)


matzf (commented):
I very much appreciate the effort to reduce these allocations, thank you. Unfortunately, this only removes one of the heap allocations done in the MAC computation, the other being the output buffer returned from the CMAC library. This cannot be fixed without changing to a different library, with a different interface. At the same time, this change makes the interface of these MAC / FullMAC functions a fair bit clunkier.

Maybe you could allow to pass `nil` for the buffer and simply fall back to internal allocation. And perhaps also panic internally instead of returning an if the buffer was given but is too small. This would make the invocation less awkward in the many test cases etc. that needed to be adapted here.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053#-:-MaF3jVt85Pd7vc3WZUB:bee5ii3)* status: 0 of 28 files reviewed, 2 unresolved discussions (waiting on @mawyss)

___
*[go/integration/braccept/cases/onehop.go, line 75 at r1](https://reviewable.io/reviews/scionproto/scion/4053#-MaF4tHs-VIzElNIE1J7:-MaF4tHs-VIzElNIE1J8:b-qj64z0) ([raw file](https://github.com/scionproto/scion/blob/e369af5bc7d56adf581355a50a6f549d0ac226d2/go/integration/braccept/cases/onehop.go#L75)):*
> ```Go
> 	if err != nil {
> 		panic(err)
> 	}
> ```

Maybe allow 
___
*[go/lib/slayers/path/mac.go, line 81 at r1](https://reviewable.io/reviews/scionproto/scion/4053#-MaF-WEGDLCBHWIdwVOR:-MaF-WEH6kBZxHwfPyHv:b85bg0g) ([raw file](https://github.com/scionproto/scion/blob/e369af5bc7d56adf581355a50a6f549d0ac226d2/go/lib/slayers/path/mac.go#L81)):*
> ```Go
> 	buffer[9] = expTime
> 	binary.BigEndian.PutUint16(buffer[10:12], consIngress)
> 	binary.BigEndian.PutUint16(buffer[12:14], consEgress)
> ```

This used to be zero initialized, now we must explicitly zero out these fields.

mawyss (commented):
Yes, the output buffer from the CMAC library is still allocated to the heap whenever MAC/FullMAC is called. The changes in this PR only remove the allocations due to the SCION MAC output buffer, as well as the input and output buffers in the EPIC MAC computation (possible due to the interface provided by the CBC-MAC).

I implemented your suggestions (however, I tried to avoid to `panic`), now the changes to the tests are less bulky.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053#-:-MaXe2DFCje5-TapkWNf:b-tyor3k)* status: 0 of 28 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[go/integration/braccept/cases/onehop.go, line 75 at r1](https://reviewable.io/reviews/scionproto/scion/4053#-MaF4tHs-VIzElNIE1J7:-MaXh0ib4-khfe0Jpxx_:b-mqb3ey) ([raw file](https://github.com/scionproto/scion/blob/e369af5bc7d56adf581355a50a6f549d0ac226d2/go/integration/braccept/cases/onehop.go#L75)):*

Maybe allow 
</blockquote></details>

Not sure what you meant here, but I allowed to pass `nil` as an input argument now, which will cause the input buffer to be allocated internally.
___
*[go/lib/slayers/path/mac.go, line 81 at r1](https://reviewable.io/reviews/scionproto/scion/4053#-MaF-WEGDLCBHWIdwVOR:-MaXhVKBDT820fsqQw9C:bm4saga) ([raw file](https://github.com/scionproto/scion/blob/e369af5bc7d56adf581355a50a6f549d0ac226d2/go/lib/slayers/path/mac.go#L81)):*

This used to be zero initialized, now we must explicitly zero out these fields.
</blockquote></details>

Done. I also adapted the EPIC MAC input accordingly.

matzf (commented):
> Reviewed 1 of 28 files at r1, 27 of 27 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: all files reviewed, 7 unresolved discussions (waiting on @mawyss)

___
*[go/integration/braccept/cases/onehop.go, line 75 at r1](https://reviewable.io/reviews/scionproto/scion/4053#-MaF4tHs-VIzElNIE1J7:-MaYBZBc7BO4YJHb-fsS:b8cr02p) ([raw file](https://github.com/scionproto/scion/blob/e369af5bc7d56adf581355a50a6f549d0ac226d2/go/integration/braccept/cases/onehop.go#L75)):*

Not sure what you meant here, but I allowed to pass `nil` as an input argument now, which will cause the input buffer to be allocated internally.
</blockquote></details>

Oops, sorry. The rest of the comment landed somewhere else.
___
*[go/lib/epic/epic.go, line 44 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabVmsNCVZs1-p_3eXW:-MabVmsO7RV5K_9xKI1N:b12hlaw) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L44)):*
> ```Go
> )
> 
> var zeroInitVector = make([]byte, 16)
> ```

Or `var zeroInitVector [16]byte`? 
___
*[go/lib/epic/epic.go, line 109 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabWGAN-kFvAHuxvQkH:-MabWGAN-kFvAHuxvQkI:b-1k19yz) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L109)):*
> ```Go
> 	// Calculate Epic MAC = first 4 bytes of the last CBC block
> 	output := outputBuffer[:len(input)]
> 	f.CryptBlocks(output, input)
> ```

Perhaps invoke this "in place", i.e. `f.CryptBlocks(input, input)` and then copy the relevant bytes to the output. That way you don't need the output buffer to be larger than the actual 4 byte output and you get a somewhat cleaner interface. In fact, then it could actually make sense to return the output as `[4]byte` array?
___
*[go/lib/epic/epic.go, line 185 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabXMPT8QiD-3odcUro:-MabXMPT8QiD-3odcUrp:b45xrl7) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L185)):*
> ```Go
> 			"srcAddr length", "srcAddrLen", srcAddrLen, "len(srcAddr)", len(srcAddr),
> 			"l", l)
> 	}
> ```

Not directly related to this PR; why even look at the `SrcAddrLen` here if we just want the length of `RawSrcAddr`? I would assume the correct relation is already enforced during parsing of the packet.
___
*[go/lib/epic/epic.go, line 192 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabagTnDW6V_v68Kgw5:-MabagTnDW6V_v68Kgw6:bo83x0g) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L192)):*
> ```Go
> 	if len(inputBuffer) < inputLength {
> 		inputBuffer = make([]byte, inputLength)
> 	}
> ```

As commented in the case for the other MAC (MACInput), perhaps it would be a bit more natural to move this up the caller and here just fill the buffer and return the size written.
___
*[go/lib/epic/epic.go, line 196 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabV6GV5M8kGnz8Lp9Q:-MabV6GV5M8kGnz8Lp9R:b-apz8gz) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L196)):*
> ```Go
> 	// Fill input
> 	offset := 0
> 	inputBuffer[0] = srcAddrLen
> ```

Not directly related to this PR but while looking at this I wondered, shouldn't this also include the type of the src address? Perhaps just add it in the same len/type format as it is represented in the SCION address header.
___
*[go/lib/slayers/path/mac.go, line 65 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabZ5ZVByxDtBU4wy8Z:-MabZ5ZVByxDtBU4wy8_:bnatlth) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/slayers/path/mac.go#L65)):*
> ```Go
> 	if len(buffer) < MACBufferSize {
> 		buffer = make([]byte, MACBufferSize)
> 	}
> ```

Maybe move this initialization up to FullMAC and the (few) other callers. As the size is fixed, this function can then just fill the provided buffer and there is no need to return it.
___
*[go/pkg/router/dataplane.go, line 753 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-Mab_ADAAuZibVkTvUuF:-Mab_ADAAuZibVkTvUuG:b-v5g4r6) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/pkg/router/dataplane.go#L753)):*
> ```Go
> 	cachedMac []byte
> 	// macBuffers avoid allocating memory during processing.
> 	macBuffers *macBuffers
> ```

Could keep the struct by value instead of a pointer indirection.

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: 24 of 28 files reviewed, 7 unresolved discussions (waiting on @matzf)

___
*[go/integration/braccept/cases/onehop.go, line 75 at r1](https://reviewable.io/reviews/scionproto/scion/4053#-MaF4tHs-VIzElNIE1J7:-Mabvg4fDD2PpeSQq-_2:b-o61icn) ([raw file](https://github.com/scionproto/scion/blob/e369af5bc7d56adf581355a50a6f549d0ac226d2/go/integration/braccept/cases/onehop.go#L75)):*

Oops, sorry. The rest of the comment landed somewhere else.
</blockquote></details>

Ok, no problem.
___
*[go/lib/epic/epic.go, line 44 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabVmsNCVZs1-p_3eXW:-Mac0m6J2BSkNYdRk7mp:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L44)):*

Or `var zeroInitVector [16]byte`? 
</blockquote></details>

Done.
___
*[go/lib/epic/epic.go, line 109 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabWGAN-kFvAHuxvQkH:-MacL5-PCV5iErP09YvK:b-umhjex) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L109)):*

Perhaps invoke this "in place", i.e. `f.CryptBlocks(input, input)` and then copy the relevant bytes to the output. That way you don't need the output buffer to be larger than the actual 4 byte output and you get a somewhat cleaner interface. In fact, then it could actually make sense to return the output as `[4]byte` array?
</blockquote></details>

Thank you for the hint, I did not know that `CryptBlocks` can operate in place. And with this, we do not need the output buffer at all, right? I now changed the code such that only one EPIC buffer is used for both in- and output.
___
*[go/lib/epic/epic.go, line 185 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabXMPT8QiD-3odcUro:-MacUDnl2WULeHlkQx6z:bkq7fhe) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L185)):*

Not directly related to this PR; why even look at the `SrcAddrLen` here if we just want the length of `RawSrcAddr`? I would assume the correct relation is already enforced during parsing of the packet.
</blockquote></details>

Thank you for pointing this out, it is indeed enforced already earlier. I adapted the code accordingly.
___
*[go/lib/epic/epic.go, line 192 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabagTnDW6V_v68Kgw5:-Mac-zlsDwV1LgDPJjsu:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L192)):*

As commented in the case for the other MAC (MACInput), perhaps it would be a bit more natural to move this up the caller and here just fill the buffer and return the size written.
</blockquote></details>

Done.
___
*[go/lib/epic/epic.go, line 196 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabV6GV5M8kGnz8Lp9Q:-MacNs-5D-MYChdHvfKR:blgjny8) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L196)):*

Not directly related to this PR but while looking at this I wondered, shouldn't this also include the type of the src address? Perhaps just add it in the same len/type format as it is represented in the SCION address header.
</blockquote></details>

The SL flag is included to make the MAC input size well-defined, so that this construction is actually secure. What would be the purpose of the source address type?
___
*[go/lib/slayers/path/mac.go, line 65 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabZ5ZVByxDtBU4wy8Z:-MabvK3yAjvjLORs6lW0:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/slayers/path/mac.go#L65)):*

Maybe move this initialization up to FullMAC and the (few) other callers. As the size is fixed, this function can then just fill the provided buffer and there is no need to return it.
</blockquote></details>

Done.
___
*[go/pkg/router/dataplane.go, line 753 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-Mab_ADAAuZibVkTvUuF:-MabvWJY3zKAUwywd0Ln:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/pkg/router/dataplane.go#L753)):*

Could keep the struct by value instead of a pointer indirection.
</blockquote></details>

Done.

matzf (commented):
> Reviewed 8 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: all files reviewed, 2 unresolved discussions (waiting on @mawyss)

___
*[go/lib/epic/epic.go, line 109 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabWGAN-kFvAHuxvQkH:-Mad6wqSCH0nrpGj_ugq:b5jc5lm) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L109)):*

Thank you for the hint, I did not know that `CryptBlocks` can operate in place. And with this, we do not need the output buffer at all, right? I now changed the code such that only one EPIC buffer is used for both in- and output.
</blockquote></details>

Good idea. Maybe add a comment that calling the function clobbers the output buffer from previous calls.
___
*[go/lib/epic/epic.go, line 196 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabV6GV5M8kGnz8Lp9Q:-Mad7NTFB5XJSklFFZFJ:bdmt3ky) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L196)):*

The SL flag is included to make the MAC input size well-defined, so that this construction is actually secure. What would be the purpose of the source address type?
</blockquote></details>

I'm not sure whether this is an actual concern in EPIC, but I believe there is a theoretical possibility that an attacker could obtain a MAC and replay it from the same raw source address but with a different source address type (e.g. SVC address `CS_M` is the same as IPv4 `128.2.0.0`) .
___
*[go/lib/epic/epic_test.go, line 388 at r3](https://reviewable.io/reviews/scionproto/scion/4053#-MadFLTGAvi1n9TuZvsh:-MadFLTGAvi1n9TuZvsi:b-fcvrv8) ([raw file](https://github.com/scionproto/scion/blob/689340a3d0719a588908f5a1602a983a93e9a4be/go/lib/epic/epic_test.go#L388)):*
> ```Go
> 	ip4Addr := &net.IPAddr{IP: net.ParseIP("10.0.0.100")}
> 	spkt.SetSrcAddr(ip4Addr)
> 	spkt.SrcAddrLen = 0
> ```

Nit: this is already set by `SetSrcAddr`.

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: 23 of 28 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[go/lib/epic/epic.go, line 109 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabWGAN-kFvAHuxvQkH:-Mb0dOd44HtxYvCi-2zT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L109)):*

Good idea. Maybe add a comment that calling the function clobbers the output buffer from previous calls.
</blockquote></details>

Done.
___
*[go/lib/epic/epic.go, line 196 at r2](https://reviewable.io/reviews/scionproto/scion/4053#-MabV6GV5M8kGnz8Lp9Q:-Mb0dSdk39iq1tp7_DYP:b-i2dkwv) ([raw file](https://github.com/scionproto/scion/blob/41a1506299f557d552dd6a6ceaea0e748223c760/go/lib/epic/epic.go#L196)):*

I'm not sure whether this is an actual concern in EPIC, but I believe there is a theoretical possibility that an attacker could obtain a MAC and replay it from the same raw source address but with a different source address type (e.g. SVC address `CS_M` is the same as IPv4 `128.2.0.0`) .
</blockquote></details>

Makes sense, thanks for pointing that out! I will handle this in a separate PR then.
___
*[go/lib/epic/epic_test.go, line 388 at r3](https://reviewable.io/reviews/scionproto/scion/4053#-MadFLTGAvi1n9TuZvsh:-Mb0fM8h7p7CHU_D_Nms:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/689340a3d0719a588908f5a1602a983a93e9a4be/go/lib/epic/epic_test.go#L388)):*

Nit: this is already set by `SetSrcAddr`.
</blockquote></details>

Done.

matzf (commented):
:lgtm: 

> Reviewed 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053#-:-Mbfms_V1oLacPUWL88y:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mawyss)

lukedirtwalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: all files reviewed, 1 unresolved discussion (waiting on @mawyss)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4053#-MdM4VH06NoeHEzZLnxI:-MdM4VH1BfJAv5yVnu4J:b-wisvs7) (no related file):*
Have you seen this blog post: https://blog.filippo.io/efficient-go-apis-with-the-inliner/ 
I wonder if that would be something we could leverage for the `FullMAC` API, what do you think?

---

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: all files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4053#-MdM4VH06NoeHEzZLnxI:-MdModwMESBS456sXfsd:b-lvrwja) (no related file):*

Have you seen this blog post: https://blog.filippo.io/efficient-go-apis-with-the-inliner/ 
I wonder if that would be something we could leverage for the `FullMAC` API, what do you think?
</blockquote></details>

The advice in post above appears to be bogus. The go build output in the example even says it explicitly "moved to heap: dst"; that's the only thing that matters, it has to be allocated on the heap. 
The main point appears to be the assumption that there is a difference between `return make([]byte, 32)` and
```
var dst [32]byte
return dst[:]
```

I'm not entirely sure, of course, but I can't quite think of any compiler optimisation that would be applicable for this longer form, but not for the simpler `make`.

Either way, I think that this approach of hoping for the escape analysis to do the right thing falls flat at the very latest once you try to pass the data as parameter in the call to an interface method. As the API of the `cmac` library is based on such an interface (`hash.Hash`), I believe this will not be applicable.

---

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: all files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4053#-MdM4VH06NoeHEzZLnxI:-MdMzRx70DXLo_74dT7r:b-uocmue) (no related file):*

The advice in post above appears to be bogus. The go build output in the example even says it explicitly "moved to heap: dst"; that's the only thing that matters, it has to be allocated on the heap. 
The main point appears to be the assumption that there is a difference between `return make([]byte, 32)` and
```
var dst [32]byte
return dst[:]
```

I'm not entirely sure, of course, but I can't quite think of any compiler optimisation that would be applicable for this longer form, but not for the simpler `make`.

Either way, I think that this approach of hoping for the escape analysis to do the right thing falls flat at the very latest once you try to pass the data as parameter in the call to an interface method. As the API of the `cmac` library is based on such an interface (`hash.Hash`), I believe this will not be applicable.
</blockquote></details>

I take it all back, it does seem to work after all :sweat_smile:  It's just that the escape analysis output is heckin confusing. It's also very easy to trip it up (as I accidentally did in my expirment :facepalm:) and accidentally make it escape after all.
And as mentioned in the blog post, devirtualisation _should_ sometimes make this work with interface methods.

---

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: all files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4053#-MdM4VH06NoeHEzZLnxI:-MdMzGCw2dH9YGXWCiBW:bz2tap7) (no related file):*
To quote from the blog post directly:

> What this is telling us is that dst escapes to the heap when allocated in X25519(), but when X25519() is inlined in main(), the dst instance that got inlined doesn't escape!

AFAIU, the whole point of the proposed solution is to play around with the "inliner budget". I.e., make the public function inlineable by making it very short.

It is a big "it depends" and we need to be very careful in order to not let stuff escape. But, if you follow what the blog post suggests, it is actually feasible to do.


On another note, the call to `h.Sum()` currently always allocates a new buffer on the heap, we could potentially get rid of it

---

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: 11 of 28 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker and @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4053#-MdM4VH06NoeHEzZLnxI:-Mda7aMb-ofOIMvbzKvw:b-1oxhyh) (no related file):*

To quote from the blog post directly:

> What this is telling us is that dst escapes to the heap when allocated in X25519(), but when X25519() is inlined in main(), the dst instance that got inlined doesn't escape!

AFAIU, the whole point of the proposed solution is to play around with the "inliner budget". I.e., make the public function inlineable by making it very short.

It is a big "it depends" and we need to be very careful in order to not let stuff escape. But, if you follow what the blog post suggests, it is actually feasible to do.


On another note, the call to `h.Sum()` currently always allocates a new buffer on the heap, we could potentially 










</blockquote></details>

This inlining sounds like a promising approach, but it does not look as if it was easily maintainable, as any change to the inlined code would need to make sure not to accidentally let the buffer escape again.
But please let me know if you think this is the right way to go here.

@oncilla Looks like a part of your comment got lost, could you further elaborate what you meant regarding `h.Sum()`?

---

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: 11 of 28 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker, @matzf, and @mawyss)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4053#-MdM4VH06NoeHEzZLnxI:-Mdapo_l5sL-owddKH8o:b-7q2xf6) (no related file):*

This inlining sounds like a promising approach, but it does not look as if it was easily maintainable, as any change to the inlined code would need to make sure not to accidentally let the buffer escape again.
But please let me know if you think this is the right way to go here.

@oncilla Looks like a part of your comment got lost, could you further elaborate what you meant regarding `h.Sum()`?
</blockquote></details>

Oh, I wonder how that went missing.
I added a comment in `mac.go` where we always force an allocation.

---
___
*[go/lib/slayers/path/mac.go, line 47 at r5](https://reviewable.io/reviews/scionproto/scion/4053#-Mdap_2Z5fxqav_eaj2G:-Mdap_2Z5fxqav_eaj2H:b-ue36p2) ([raw file](https://github.com/scionproto/scion/blob/5bc1849d84652e5ae1eedefe2f14bb6c81ef5b41/go/lib/slayers/path/mac.go#L47)):*
> ```Go
> 		panic(err)
> 	}
> 	return h.Sum(nil)[:16]
> ```

Here we always allocate a fresh buffer.
This could potentially be optimized to further reduce allocations.

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: 8 of 28 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker, @matzf, and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4053#-MdM4VH06NoeHEzZLnxI:-MduxxdyCw0i9ttXuuWj:b-txz1u6) (no related file):*

Oh, I wonder how that went missing.
I added a comment in `mac.go` where we always force an allocation.
</blockquote></details>

Ok, I also got rid of this allocation (by reusing the same buffer that stores the input to the MAC).

---
___
*[go/lib/slayers/path/mac.go, line 47 at r5](https://reviewable.io/reviews/scionproto/scion/4053#-Mdap_2Z5fxqav_eaj2G:-MduxpHTC52ystkQOivK:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5bc1849d84652e5ae1eedefe2f14bb6c81ef5b41/go/lib/slayers/path/mac.go#L47)):*

Here we always allocate a fresh buffer.
This could potentially be optimized to further reduce allocations.
</blockquote></details>

Done.

lukedirtwalker (approved):
> Reviewed 2 of 8 files at r3, 3 of 5 files at r4, 16 of 17 files at r5, 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: all files reviewed, 1 unresolved discussion (waiting on @oncilla)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: all files reviewed, 1 unresolved discussion (waiting on @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4053#-MdM4VH06NoeHEzZLnxI:-Me0QaJc1ydaLy_sMwSb:b-530n48) (no related file):*

Ok, I also got rid of this allocation (by reusing the same buffer that stores the input to the MAC).
</blockquote></details>

Oh, nice! I now see I had misunderstood the `hash.Hash` interface, thinking that the parameter to `Sum` was appended to the hash _input_ . It did seem a bit redundant, now it makes more sense :) Cool! 

---

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: all files reviewed, 1 unresolved discussion (waiting on @oncilla)

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4053)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mawyss)

PR #4057, opened 2021-05-26
router: simplify SCMP reply handling

matzf:
* Simplify the SCMP error reply quote
  Previously, an explicit copy of the input packet was made, so that it could be used in the quote for an SCMP error reply. But the path header of this quote was updated to the current processing state of the packet. This somewhat surprising step was done in order to have identical behaviour to anapayas VPP router.
   Now, the input packet is used for the quote directly, with the modifications from processing directly written back to it. The result is identical (with exception of router alert flag clearing mentioned below).
 
*  Parse/skip SCION extension headers from the beginning, so that the actual L4 protocol type is available when SCMP replies need to be sent.  Previously, the full packet was parsed again to determine whether the input packet was an SCMP error message.
  Also, handling traceroute packets now checks that the input packet was even an SCMP message.

* Simplify clearing router alert flags
  This step where the quoted packet was updated to the current processing state, mentioned above, was also responsible for clearing the router alert in traceroute replies.
  This is now done explicitly during processing, with the consequence that router alerts are always cleared, as was probably intended, also for non-traceroute packets.

  Some test cases fail because router flag is now reset during processing.
  Removed setting router alert in these test cases. Alternatively, we could check for the expected alert flag clearing in the test, but I think this is not really the scope of these test cases.

* Remove unnecessary updateSCIONLayer calls
  Unnecessary because SetInfoField etc directly write back to the Raw path and thus the input packet.

* braccept: dont panic on decode error

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4057)

### DIFF
--- a/go/integration/braccept/runner/run.go
+++ b/go/integration/braccept/runner/run.go
@@ -139,6 +139,11 @@ func (c *RunConfig) ExpectPacket(pkt ExpectedPacket, normalizeFn NormalizePacket
 				"pkt", i, "expected", pkt.DevName, "actual", c.deviceNames[idx], "packet", got))
 			continue
 		}
+		if err := got.ErrorLayer(); err != nil {
+			errors = append(errors, serrors.WrapStr("error decoding packet", err.Error(),
+				"pkt", i))
+			continue
+		}
 		if err := comparePkts(got, pkt.Pkt, normalizeFn); err != nil {
 			errors = append(errors, serrors.WrapStr("received mismatching packet", err,
 				"pkt", i))
--- a/go/pkg/router/dataplane.go
+++ b/go/pkg/router/dataplane.go
@@ -925,9 +925,6 @@ func (p *scionPacketProcessor) updateNonConsDirIngressSegID() error {
 		if err := p.path.SetInfoField(p.infoField, int(p.path.PathMeta.CurrINF)); err != nil {
 			return serrors.WrapStr("update info field", err)
 		}
-		if err := updateSCIONLayer(p.rawPkt, p.scionLayer, p.buffer); err != nil {
-			return err
-		}
 	}
 	return nil
 }
@@ -995,9 +992,6 @@ func (p *scionPacketProcessor) processEgress() error {
 		// TODO parameter problem invalid path
 		return serrors.WrapStr("incrementing path", err)
 	}
-	if err := updateSCIONLayer(p.rawPkt, p.scionLayer, p.buffer); err != nil {
-		return err
-	}
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


lukedirtwalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4057)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4057#-MdS8_aB7rKsV073BjKr:-MdS8_aC8DEx9dv-Y2QI:bbtlh3f) (no related file):*
great simplification :100: thanks.

---
___
*[go/pkg/router/dataplane.go, line 584 at r1](https://reviewable.io/reviews/scionproto/scion/4057#-MdS79jx1Qv9uNqZHfhK:-MdS79jy1ZWD99MXKPab:b-q6cnc1) ([raw file](https://github.com/scionproto/scion/blob/e47c04a2c38ace33a5dfb8ac811c9f3b62189e55/go/pkg/router/dataplane.go#L584)):*
> ```Go
> 		return processResult{}, err
> 	}
> 	var nextHdr common.L4ProtocolType
> ```

hm actually we only want to know if the next layer is BFD right? maybe we could just use `nextBFD := p.lastLayer.NextLayerType() == <bfd-layertype>`  what do you think?

If you disagree I think I would at least default initialize the nextHdr with `nextHdr := p.scionLayer.NextHdr` then the last `else` clause is not necessary.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4057)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

___
*[go/pkg/router/dataplane.go, line 584 at r1](https://reviewable.io/reviews/scionproto/scion/4057#-MdS79jx1Qv9uNqZHfhK:-MdSYL7e9m7xxFQINMnX:bvxvmbg) ([raw file](https://github.com/scionproto/scion/blob/e47c04a2c38ace33a5dfb8ac811c9f3b62189e55/go/pkg/router/dataplane.go#L584)):*

hm actually we only want to know if the next layer is BFD right? maybe we could just use `nextBFD := p.lastLayer.NextLayerType() == <bfd-layertype>`  what do you think?

If you disagree I think I would at least default initialize the nextHdr with `nextHdr := p.scionLayer.NextHdr` then the last `else` clause is not necessary.
</blockquote></details>

Indeed, you're right. The only justification for more verbosely using the `NextHdr` is that we want to use it for an error message below (in the case empty path and not BFD). 
Simplified the code to use `NextLayerType` as you suggested and extract the `NextHdr` for the error message with a short helper function.

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4057)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4065, opened 2021-06-18
slayers: fix padding after TLV options

matzf:
Padding after the last TLV option is responsible for bringing the total length of the extension header to a multiple of the `LineLen`, i.e. to a multiple of 4 bytes.
The calculation for this padding was slightly wrong. It added `length % LineLen` instead of `LineLen - length % LineLen` padding bytes.
As a consequence, the total size of the extension was not padded to a multiple of 4 when a TLV option with odd length was used, triggering the corresponding error later during serialization.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4065)

### DIFF
--- a/go/lib/slayers/extn.go
+++ b/go/lib/slayers/extn.go
@@ -144,8 +144,9 @@ func serializeTLVOptions(buf []byte, options []*tlvOption, fixLengths bool) int 
 		length += opt.length(fixLengths)
 	}
 	if fixLengths {
-		pad := length % LineLen
-		if pad != 0 {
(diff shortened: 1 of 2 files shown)


oncilla (approved):
:lgtm:

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4065#-:-MchfoRD6zYRT4eS-IGb:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4072, opened 2021-06-25
Remove SignedRevInfo and capnproto dependency

matzf:
The signed revocation mechanism has previously been removed from the SCION protocol and replaced by simpler (unsigned) Interface Down messages. Internally, the `SignedRevInfo` objects were still created to interface with various components that had not yet been updated during this change, in particular for the internal interface of the daemon, the `RevCache` and the segment verifier.

This change removes the `SignedRevInfo` type and replaces it by directly using the corresponding `RevInfo` where appropriate. In the segverifier, the (previously unused) support for handling revocations was removed altogether.

This allows to remove capnproto as a dependency; `SignedRevInfo` was the last active use for capnproto (to create a `SignedRevInfo` object, a `RevInfo` was `Pack`ed).  Some inactive colibri code was also still relying on capnproto, but this could easily be stripped out; work to replace this with gRPC is ongoing.
Dropping the capnproto dependency reduces the size of the resulting binaries by about 2MB (for the `end2end` binary, that's a bit under 10%).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4072)


lukedirtwalker (approved):
> Reviewed 90 of 90 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4072)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4072#-MdM1fSoCGWwdwNd7zPL:-MdM1fSoCGWwdwNd7zPM:bxpcdt7) (no related file):*
great stuff, thanks a lot

---

PR #4074, opened 2021-06-29
slayers: remove gopacket/layers dependency

matzf:
The `slayers` library, which is used by `snet` and therefore by most SCION applications, was importing the `gopacket/layers` library. This library contains a considerable amount of bloat as it supports parsing every protocol under the sun. This was used only for a small amount of basic functionality that can easily be replicated (like `BaseLayer` and the some parts of the SCION/UDP layer), as well as to support SCION/BFD. The BFD support is not needed by applications other than the router.
Fortunately, `gopacket`'s layer type system allows to define the BFD layer type as a constant, so that the `slayers.SCION` `NextLayerType` can support BFD without actually linking against `gopacket/layers`. (Note: this `NextLayerType` is not even used in the router as it handles BFD packets separately. It is however used in various test cases that parse BFD packets with `NewPacket`).

This change removes the `gopacket/layers` dependency from `slayers`. For applications that import `gopacket/layers` explicitly, e.g. because they need to parse SCION/BFD packets, the functionality remains unchanged. For applications that do not import `gopacket/layers`, SCION/BFD packets can no longer be parsed, and the binary size is reduced by by about 2MB.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4074)


lukedirtwalker (changes requested):
> Reviewed 37 of 37 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4074)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[go/lib/slayers/udp.go, line 56 at r1](https://reviewable.io/reviews/scionproto/scion/4074#-MdMBlbzEkOb-hREupUn:-MdMBlc-BDmo-yiGRAvo:b884r7b) ([raw file](https://github.com/scionproto/scion/blob/464f1721529ea4050350e328720fdba1aba6393e/go/lib/slayers/udp.go#L56)):*
> ```Go
> }
> 
> func (u *UDP) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
> ```

add a comment that this is copied from upstream layers.UDP

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4074)* status: 36 of 37 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

___
*[go/lib/slayers/udp.go, line 56 at r1](https://reviewable.io/reviews/scionproto/scion/4074#-MdMBlbzEkOb-hREupUn:-MdMKxqUAY_Z5N7lMoeZ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/464f1721529ea4050350e328720fdba1aba6393e/go/lib/slayers/udp.go#L56)):*

add a comment that this is copied from upstream layers.UDP
</blockquote></details>

Done.

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4074)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4079, opened 2021-07-01
epic: endhost logic

mawyss:
This PR implements the EPIC endhost logic.
Applications can now send packets using the EPIC path type header.

Major changes:
- The gRPC communication between the SCION daemon and the application is extended, so that the application gets the two authenticators necessary to calculate the PHVF and LHVF when fetching paths from the daemon.
- `snet` contains the major logic behind the creation of EPIC packets, where `snet/path` keeps the necessary state and uses it to create the PHVF and LHVF when the packet is serialized.
- There is an additional argument "--epic" for the ping, traceroute, showpaths, end2end_integration, and scion_integration tools. The argument enables those tools to send EPIC-HP path type packets (if paths with the necessary EPIC authenticators are available).

Potential next steps:
- [DONE] Wireshark dissector for EPIC packets (current status: EPIC path type header is recognized, but not parsed)
- Buildkite integration tests for EPIC
- `helloepic` application for `netsec-ethz/scion-apps`

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)


matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 0 of 26 files reviewed, 1 unresolved discussion (waiting on @mawyss)

___
*[go/lib/snet/packet.go, line 562 at r1](https://reviewable.io/reviews/scionproto/scion/4079#-Mda90bY9HgIg6V3c76D:-Mda90bY9HgIg6V3c76E:bi6behe) ([raw file](https://github.com/scionproto/scion/blob/ec592b34235995714184886079c3f3f5db886572/go/lib/snet/packet.go#L562)):*
> ```Go
> 	}
> 
> 	scionLayer.PayloadLen = uint16(p.Payload.Length())
> ```

We set `FixLengths: true` in the serialize options below. As far as I understand, the `PayloadLen` will be overwritten and setting it is not necessary.

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 0 of 26 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[go/lib/snet/packet.go, line 562 at r1](https://reviewable.io/reviews/scionproto/scion/4079#-Mda90bY9HgIg6V3c76D:-MdaTg2o3_AbbdzVmNxu:b-64u78y) ([raw file](https://github.com/scionproto/scion/blob/ec592b34235995714184886079c3f3f5db886572/go/lib/snet/packet.go#L562)):*

We set `FixLengths: true` in the serialize options below. As far as I understand, the `PayloadLen` will be overwritten and setting it is not necessary.
</blockquote></details>

Yes, this is true, it is redundant. However, we need the `PayloadLen` field to be filled already before the packet gets serialized: first the PHVF and LHVF are calculated (`AddEpicHVFs()`), which (if you further trace the call) depend on `prepareMacInput()` from `lib/epic/epic.go`. The `prepareMacInput()` function adds the `PayloadLen` field to the input of the MAC function.

oncilla (changes requested):
> Reviewed 15 of 26 files at r1, 8 of 9 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 23 of 26 files reviewed, 19 unresolved discussions (waiting on @matzf and @mawyss)

___
*[go/dispatcher/dispatcher/underlay.go, line 345 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgviidYClWn2VO7b-8A:-MgviidYClWn2VO7b-8B:bs2g81) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/dispatcher/dispatcher/underlay.go#L345)):*
> ```Go
> 		return serrors.WrapStr("setting destination address", err)
> 	}
> 	if pkt.SCION.PathType == epic.PathType {
> ```

This special logic shouldn't be necessary.
`pkt.SCION.Path.Reverse()` already constructs a SCION path if `pkt.SCION.Path` is an EPIC path.

What you need to do is after the if branch below, assign the new path type. (currently it is missing which is a bug.)
```go
	if pkt.SCION.Path, err = pkt.SCION.Path.Reverse(); err != nil {
		return serrors.WrapStr("reversing path", err)
	}
	pkt.SCION.PathType = pkt.SCION.Path.Type()
```
___
*[go/integration/end2end_integration/main.go, line 79 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvjYKW55WChjrHl2xY:-MgvjYKW55WChjrHl2xZ:bq63u00) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/integration/end2end_integration/main.go#L79)):*
> ```Go
> 		"-local", integration.SrcAddrPattern + ":0",
> 		"-remote", integration.DstAddrPattern + ":" + integration.ServerPortReplace,
> 		fmt.Sprintf("-epic=%s", strconv.FormatBool(epic)),
> ```

keep it consistent with the code above:
```go
"-epic", strconv.FormatBool(epic)
```

side-note: you write the above a bit simpler like this:
```
fmt.Sprintf("-epic=%t", epic)
```
___
*[go/integration/end2end_integration/main.go, line 279 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvjxsS-SDpIeRRaI3U:-MgvjxsS-SDpIeRRaI3V:b-9ul2ny) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/integration/end2end_integration/main.go#L279)):*
> ```Go
> 			"-local", integration.SrcAddrPattern + ":0",
> 			"-remote", integration.DstAddrPattern + ":" + integration.ServerPortReplace,
> 			fmt.Sprintf("-epic=%s", strconv.FormatBool(epic)),
> ```

keep it consistent with the other arguments.
___
*[go/integration/scion_integration/main.go, line 59 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvk14L5T5QwSK4Jfoa:-Mgvk14L5T5QwSK4Jfob:b71s6kd) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/integration/scion_integration/main.go#L59)):*
> ```Go
> 		"--sciond", integration.Daemon,
> 		"--log.level", "debug",
> 		fmt.Sprintf("--epic=%s", strconv.FormatBool(epic)),
> ```

keep it consistent.
___
*[go/lib/daemon/grpc.go, line 215 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvkV4p4GJ20rTPO-D7:-MgvkV4p4GJ20rTPO-D8:bw1cp3i) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/daemon/grpc.go#L215)):*
> ```Go
> 	}
> 
> 	var epicAuths *snet.EpicAuths = nil
> ```

`var epicAuths *snet.EpicAuths` is the idomatic way of declaring the variable. The `nil` can be dropped
___
*[go/lib/epic/epic.go, line 136 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvkqUL95cUxJd4TLoy:-MgvkqUL95cUxJd4TLoz:b-mtcpah) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/epic/epic.go#L136)):*
> ```Go
> 	if subtle.ConstantTimeCompare(hvf, mac) == 0 {
> 		return serrors.New("epic hop validation field verification failed",
> 			"hvf in packet", hvf, "calculated mac", mac, "auth", fmt.Sprintf("%x", auth))
> ```

do we need to format it as hex here? It is rather expensive operation for `VerifyHVF` which will be used in the data plane.
___
*[go/lib/infra/modules/combinator/graph.go, line 395 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvlEqV-PDyZglA6S7B:-MgvlEqV-PDyZglA6S7C:b-qpr3w9) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/infra/modules/combinator/graph.go#L395)):*
> ```Go
> 	staticInfo := collectMetadata(interfaces, asEntries)
> 
> 	var epicAuths *snet.EpicAuths = nil
> ```

drop the `=nil`
___
*[go/lib/snet/packet.go, line 566 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvu3dv-ROgVW-b5o13:-Mgvu3dv-ROgVW-b5o14:b-9c592w) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/snet/packet.go#L566)):*
> ```Go
> 		sp := scionLayer.Path.(*scion.Raw)
> 
> 		if p.Path.EpicEnabled() {
> ```

I don't think it is a good idea to special case every path type.

As far as I can tell, the path needs to keep state and it needs access to the scionLayer for an update.
I think we could build a plugin system that allows both of these requirements.

Essentially, what we want is to have a thing that gives us `slayers/path.Type` `slayers/path.Path` and that has a method to update it's state based on the `scionLayer`.
Maybe something like `(t thing) Compute(l *slayers.SCION)`.

I wonder if we can abuse the `ComputeChecksum` layers option for this :thinking: 

In any case, we need to come up with a better way of plugging this. Otherwise each path type adds additional spaghetti to this part of the code base.
___
*[go/lib/snet/path.go, line 120 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvnwpc-8sZKzEv52j_:-Mgvnwpc-8sZKzEv52ja:b8m42jt) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/snet/path.go#L120)):*
> ```Go
> 
> 	// EpicAuths denotes the EPIC authenticators (if added during beaconing).
> 	EpicAuths *EpicAuths
> ```

What is the intended usage of these fields? They are not used currently AFAICT.

My worry is that we pollute `PathMetadata` with all kinds of data.
If this is not strictly needed I would err on the side of not adding it.
___
*[go/lib/snet/path.go, line 260 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvoQq67UlPA0Vta7ky:-MgvoQq67UlPA0Vta7kz:b-sol9g2) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/snet/path.go#L260)):*
> ```Go
> func (p *partialPath) String() string {
> 	return fmt.Sprintf("{spath: %s, underlay: %s, dest: %s}",
> 		p.spath.String(), p.underlay, p.destination)
> ```

that is not needed, `%s` uses the `String` method if it is defined.
___
*[go/lib/spath/path.go, line 102 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvodVp9c9EZXTIa6nq:-MgvodVp9c9EZXTIa6nr:b-pknnz5) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/spath/path.go#L102)):*
> ```Go
> }
> 
> func (p Path) SupportsEpic() bool {
> ```

I don't think we should extend the API of this type with path type specific methods.
`spath.Path` was intended to be a dumb container for raw bytes.

If we add path type specific methods, we end up with a intractable API in the end.

I think we need to design a plugin system in snet that allows calculations based on the packet.
___
*[go/lib/spath/path.go, line 182 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvrKr2550_hI37c2W_:-MgvrKr2550_hI37c2Wa:b-oyugyz) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/spath/path.go#L182)):*
> ```Go
> 		e, ok := po.(*epic.Path)
> 		if !ok {
> 			return serrors.New("Path type and path data do not match")
> ```

errors should start with a lower case letter.
___
*[go/pkg/app/path/path.go, line 87 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvrkZ97QWEECf2inh8:-MgvrkZ97QWEECf2inh9:beje1bk) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/app/path/path.go#L87)):*
> ```Go
> 	if o.epic {
> 		// Only use paths that support EPIC and intra-AS (empty) paths.
> 		epicPaths := []snet.Path{}
> ```

the idomatic way is to do:
```go
var epicPaths []snet.Path
```

or 
```go
epicPaths = make([]snet.Path, 0, len(paths))
```
to avoid resizing the slice in the appends.
___
*[go/pkg/app/path/path.go, line 94 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvrUmYBeBx3gbuTO2B:-MgvrUmZCORruyBPH4-q:b-nx86w6) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/app/path/path.go#L94)):*
> ```Go
> 		}
> 		if len(epicPaths) == 0 {
> 			return nil, serrors.New("no EPIC paths available.")
> ```

drop the period at the end of the error message.
___
*[go/pkg/cs/tasks.go, line 213 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvs7uy8bWjlfd9i8XH:-Mgvs7uy8bWjlfd9i8XI:b-x1gd3x) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/cs/tasks.go#L213)):*
> ```Go
> 		StaticInfo: t.StaticInfo,
> 		Task:       task,
> 		EPIC:       true,
> ```

this should be a configuration option.
___
*[go/pkg/daemon/internal/servers/grpc.go, line 146 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvsCtSCgB006UVAXBp:-MgvsCtSCgB006UVAXBq:bn07hpz) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/daemon/internal/servers/grpc.go#L146)):*
> ```Go
> 	}
> 
> 	var epicAuths *sdpb.EpicAuths = nil
> ```

drop the `= nil`
___
*[go/pkg/pathprobe/paths.go, line 104 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvsNOd5NoohNNukxBX:-MgvsNOd5NoohNNukxBY:bub81w6) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/pathprobe/paths.go#L104)):*
> ```Go
> // non-empty paths.
> func (p Prober) GetStatuses(ctx context.Context, paths []snet.Path,
> 	epic bool) (map[string]Status, error) {
> ```

add options here to make the API more ergonomic and easier to change in the future.
___
*[go/scion/ping.go, line 131 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvtU4O5oj_ZXSDZicC:-MgvtU4O5oj_ZXSDZicD:b-4haria) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/scion/ping.go#L131)):*
> ```Go
> 			remote.Path = path.Path()
> 			if flags.epic && !remote.Path.IsEmpty() {
> 				if err = remote.Path.EnableEpic(); err != nil {
> ```

Can this not be part of `path.Choose`?

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 23 of 26 files reviewed, 19 unresolved discussions (waiting on @matzf and @mawyss)

___
*[go/lib/snet/packet.go, line 566 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvu3dv-ROgVW-b5o13:-Mgvwa-H7J1lBKHgzkZw:br76gxu) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/snet/packet.go#L566)):*

I don't think it is a good idea to special case every path type.

As far as I can tell, the path needs to keep state and it needs access to the scionLayer for an update.
I think we could build a plugin system that allows both of these requirements.

Essentially, what we want is to have a thing that gives us `slayers/path.Type` `slayers/path.Path` and that has a method to update it's state based on the `scionLayer`.
Maybe something like `(t thing) Compute(l *slayers.SCION)`.

I wonder if we can abuse the `ComputeChecksum` layers option for this :thinking: 

In any case, we need to come up with a better way of plugging this. Otherwise each path type adds additional spaghetti to this part of the code base.












</blockquote></details>

I need to think about a way to make this easier.

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 8 of 29 files reviewed, 19 unresolved discussions (waiting on @matzf and @oncilla)

___
*[go/dispatcher/dispatcher/underlay.go, line 345 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgviidYClWn2VO7b-8A:-MithZrm7e6gFK6woK6Q:b3oft2g) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/dispatcher/dispatcher/underlay.go#L345)):*

This special logic shouldn't be necessary.
`pkt.SCION.Path.Reverse()` already constructs a SCION path if `pkt.SCION.Path` is an EPIC path.

What you need to do is after the if branch below, assign the new path type. (currently it is missing which is a bug.)
```go
	if pkt.SCION.Path, err = pkt.SCION.Path.Reverse(); err != nil {
		return serrors.WrapStr("reversing path", err)
	}
	pkt.SCION.PathType = pkt.SCION.Path.Type()
```


</blockquote></details>

If the packet contains an EPIC path, then `pkt.SCION.Path.Reverse()` does not return a SCION path but an EPIC path (see `slayers/path/epic/epic.go`).
What I want to achieve here is to extract the SCION path from the EPIC path and assign it to the packet.

So the packet we receive has an EPIC path type, but the packet we send back should have a SCION path type.
___
*[go/integration/end2end_integration/main.go, line 79 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvjYKW55WChjrHl2xY:-Mitpzj70qaIcXC7u4X8:b-xztf0m) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/integration/end2end_integration/main.go#L79)):*

keep it consistent with the code above:
```go
"-epic", strconv.FormatBool(epic)
```

side-note: you write the above a bit simpler like this:
```
fmt.Sprintf("-epic=%t", epic)
```

</blockquote></details>

From the specification of package `flag` (`https://pkg.go.dev/flag`):

The following forms are permitted: 
`-flag`
`-flag=x`
`-flag x  // non-boolean flags only`

Therefore, passing false to the epic argument is only possible by removing the argument epic completely, or by using `-epic=false`.
I went for the second option to omit introducing an if-statement at this place.
___
*[go/integration/end2end_integration/main.go, line 279 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvjxsS-SDpIeRRaI3U:-MitraH68qf2IDNUitVW:b-q64pkt) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/integration/end2end_integration/main.go#L279)):*

keep it consistent with the other arguments.
</blockquote></details>

See above.
___
*[go/integration/scion_integration/main.go, line 59 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvk14L5T5QwSK4Jfoa:-MitrvKYBx8bsgeNSiLe:b-q64pkt) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/integration/scion_integration/main.go#L59)):*

keep it consistent.
</blockquote></details>

See above.
___
*[go/lib/daemon/grpc.go, line 215 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvkV4p4GJ20rTPO-D7:-Mits2Ko7qkOjl3vlTix:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/daemon/grpc.go#L215)):*

`var epicAuths *snet.EpicAuths` is the idomatic way of declaring the variable. The `nil` can be dropped
</blockquote></details>

Done.
___
*[go/lib/epic/epic.go, line 136 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvkqUL95cUxJd4TLoy:-Mits_-S5_bKthvbtJ75:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/epic/epic.go#L136)):*

do we need to format it as hex here? It is rather expensive operation for `VerifyHVF` which will be used in the data plane.
</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/graph.go, line 395 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvlEqV-PDyZglA6S7B:-MitsgYI1JCgxHN4OXqL:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/infra/modules/combinator/graph.go#L395)):*

drop the `=nil`
</blockquote></details>

Done.
___
*[go/lib/snet/packet.go, line 566 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvu3dv-ROgVW-b5o13:-MiyvNCV7NSleM0lACg5:b-ibgkiw) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/snet/packet.go#L566)):*

I need to think about a way to make this easier.
</blockquote></details>

Something like `func (s *SCION) UpdatePath(spath *spath.Path)`?
(Assuming the `EpicData` would still be stored in the spath.)
___
*[go/lib/snet/path.go, line 120 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvnwpc-8sZKzEv52j_:-MitydqLBEPaBwUdpoAz:blkbqi9) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/snet/path.go#L120)):*

What is the intended usage of these fields? They are not used currently AFAICT.

My worry is that we pollute `PathMetadata` with all kinds of data.
If this is not strictly needed I would err on the side of not adding it.


</blockquote></details>

Done.

Good observation. The `PathMetadata` was used to store the PHVF/LHVF, before transmitting them via grpc.
For consistency, the PHVF/LHVF are now added to `spath.Path.EpicData` also there.
___
*[go/lib/snet/path.go, line 260 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvoQq67UlPA0Vta7ky:-MiuS7FA6NmMxuCCbMmE:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/snet/path.go#L260)):*

that is not needed, `%s` uses the `String` method if it is defined.
</blockquote></details>

Done.
___
*[go/lib/spath/path.go, line 102 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvodVp9c9EZXTIa6nq:-Miz0eCGB4gAOEc7YNii:b-n7g4zb) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/spath/path.go#L102)):*

I don't think we should extend the API of this type with path type specific methods.
`spath.Path` was intended to be a dumb container for raw bytes.

If we add path type specific methods, we end up with a intractable API in the end.

I think we need to design a plugin system in snet that allows calculations based on the packet.





</blockquote></details>

If we would have `func (s *SCION) UpdatePath(spath *spath.Path)` as described above, then `spath.Path` would only need to keep the `EpicData` plus the methods `SupportsEpic/EnableEpic/EpicEnabled`, but not `AddEpicPktID` and `AddEpicHVFs`.
Those remaining methods could be generalized for other path types, e.g., `Supports(t path.Type)` etc.

In any case, please let me know if you have a more appropriate solution.
___
*[go/lib/spath/path.go, line 182 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvrKr2550_hI37c2W_:-MiuSaVe0tkFhbISDrP4:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/spath/path.go#L182)):*

errors should start with a lower case letter.
</blockquote></details>

Done.
___
*[go/pkg/app/path/path.go, line 87 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvrkZ97QWEECf2inh8:-MiuT5MKDLVVKJeDuqw8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/app/path/path.go#L87)):*

the idomatic way is to do:
```go
var epicPaths []snet.Path
```

or 
```go
epicPaths = make([]snet.Path, 0, len(paths))
```
to avoid resizing the slice in the appends.

</blockquote></details>

Done.
___
*[go/pkg/app/path/path.go, line 94 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvrUmYBeBx3gbuTO2B:-MiuT9Tk48ibkiyOTaD1:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/app/path/path.go#L94)):*

drop the period at the end of the error message.
</blockquote></details>

Done.
___
*[go/pkg/cs/tasks.go, line 213 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvs7uy8bWjlfd9i8XH:-MivS-JPErPx2w07yDvp:b-jnd55) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/cs/tasks.go#L213)):*

this should be a configuration option.
</blockquote></details>

Done. I added it to the `BSConfig` struct in the CS config, please let me know in case this is the wrong place.
___
*[go/pkg/daemon/internal/servers/grpc.go, line 146 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvsCtSCgB006UVAXBp:-MiuTjBI4Czv655JBCWm:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/daemon/internal/servers/grpc.go#L146)):*

drop the `= nil`
</blockquote></details>

Done.
___
*[go/pkg/pathprobe/paths.go, line 104 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvsNOd5NoohNNukxBX:-MivuY24BdNPV6yXbkGM:b-u97q19) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/pathprobe/paths.go#L104)):*

add options here to make the API more ergonomic and easier to change in the future.
</blockquote></details>

Done. Introduced a struct `Options` for this purpose.
___
*[go/scion/ping.go, line 131 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvtU4O5oj_ZXSDZicC:-MiunfK21hBc216IK1i6:b-ju8t4k) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/scion/ping.go#L131)):*

Can this not be part of `path.Choose`?
</blockquote></details>

Unfortunately not. The problem is that `snet.Path()` is implemented by returning a copy of the underlying `spath.Path`.
Therefore, modifying this copy by calling `EnableEpic()` has no effect.
For this reason we are first choosing the `snet.Path`, then copying the underlying `spath.Path`, and from then on working only on this copy (`remote.Path` in `ping.go`).

oncilla (changes requested):
> Reviewed 1 of 26 files at r1, 1 of 4 files at r3, 4 of 11 files at r4, 13 of 16 files at r5.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 21 of 29 files reviewed, 11 unresolved discussions (waiting on @matzf, @mawyss, and @oncilla)

___
*[go/cs/config/bs_sample.go, line 27 at r5](https://reviewable.io/reviews/scionproto/scion/4079#-MoECbUTAJdCHQmPEgQT:-MoECbUTAJdCHQmPEgQU:b-lu4mv3) ([raw file](https://github.com/scionproto/scion/blob/a2f12ba142365d3a81611cc5d8255ace2c44b78c/go/cs/config/bs_sample.go#L27)):*
> ```Go
> registration_interval = "5s"
> 
> # Add EPIC authenticators to the beacons. (default true)
> ```

Given this is experimental, the default should be false.
___
*[go/cs/config/config.go, line 157 at r5](https://reviewable.io/reviews/scionproto/scion/4079#-MoECfQX8OlGE77RD6p4:-MoECfQX8OlGE77RD6p5:b-i25khf) ([raw file](https://github.com/scionproto/scion/blob/a2f12ba142365d3a81611cc5d8255ace2c44b78c/go/cs/config/config.go#L157)):*
> ```Go
> 	Policies Policies `toml:"policies,omitempty"`
> 	// EPIC specifies whether the EPIC authenticators should be added to the beacons.
> 	EPIC bool `toml:"epic,omitempty" default:"true"`
> ```

Default should be false here.
___
*[go/dispatcher/dispatcher/underlay.go, line 345 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgviidYClWn2VO7b-8A:-MoEE0HuC9usunIBBuDu:bjxvny4) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/dispatcher/dispatcher/underlay.go#L345)):*

If the packet contains an EPIC path, then `pkt.SCION.Path.Reverse()` does not return a SCION path but an EPIC path (see `slayers/path/epic/epic.go`).
What I want to achieve here is to extract the SCION path from the EPIC path and assign it to the packet.

So the packet we receive has an EPIC path type, but the packet we send back should have a SCION path type.
</blockquote></details>

You right, the reversed path is still an EPIC path.

That got me thinking though. Does it ever make sense to reverse an EPIC path and *not* convert it to a SCION path? Is that a supported use-case by EPIC?
___
*[go/pkg/pathprobe/paths.go, line 104 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvsNOd5NoohNNukxBX:-MoEG4BiCrnpPAVMVmcK:b5fxogc) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/pathprobe/paths.go#L104)):*

Done. Introduced a struct `Options` for this purpose.
</blockquote></details>

We are using the functional options pattern to make our APIs ergonomic and extendable.

Please us the pattern that is used for example here: https://github.com/scionproto/scion/blob/b5954eafc987875b1cdea6a85f66fae15a54178e/go/pkg/app/path/path.go#L290-L298
___
*[go/pkg/router/dataplane.go, line 1413 at r5](https://reviewable.io/reviews/scionproto/scion/4079#-MoEGq0ECnNndmyCLOF3:-MoEGq0ECnNndmyCLOF4:burefje) ([raw file](https://github.com/scionproto/scion/blob/a2f12ba142365d3a81611cc5d8255ace2c44b78c/go/pkg/router/dataplane.go#L1413)):*
> ```Go
> 	// back to rawPkt (quote).
> 	var path *scion.Raw
> 	badPathErr := serrors.WithCtx(cannotRoute, "details", "unsupported path type",
> ```

This means that we always allocate the error, even if we do not run into a problem later on.
This puts unnecessary pressure on the garbage collector that we should avoid as much as possible.

Granted, the code in this package is not optimal regarding allocations, but we should still not add more to it.

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 7 of 29 files reviewed, 11 unresolved discussions (waiting on @matzf and @oncilla)

___
*[go/cs/config/bs_sample.go, line 27 at r5](https://reviewable.io/reviews/scionproto/scion/4079#-MoECbUTAJdCHQmPEgQT:-Mp6suYYDUQPLB56IS2N:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/a2f12ba142365d3a81611cc5d8255ace2c44b78c/go/cs/config/bs_sample.go#L27)):*

Given this is experimental, the default should be false.
</blockquote></details>

Done.
___
*[go/cs/config/config.go, line 157 at r5](https://reviewable.io/reviews/scionproto/scion/4079#-MoECfQX8OlGE77RD6p4:-Mp6tfzf0FvKxv41m1V-:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/a2f12ba142365d3a81611cc5d8255ace2c44b78c/go/cs/config/config.go#L157)):*

Default should be false here.
</blockquote></details>

Done.
___
*[go/dispatcher/dispatcher/underlay.go, line 345 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgviidYClWn2VO7b-8A:-Mp6tx8J6St6vrj6Jr0d:b-ryze1j) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/dispatcher/dispatcher/underlay.go#L345)):*

You right, the reversed path is still an EPIC path.

That got me thinking though. Does it ever make sense to reverse an EPIC path and *not* convert it to a SCION path? Is that a supported use-case by EPIC?




</blockquote></details>

If the source of the packet is behind a hidden path itself, the destination AS sends back EPIC-HP traffic (recommended if the source AS uses EPIC path type prioritization at the BRs, but necessary if the source AS disallows SCION path type traffic completely).
___
*[go/pkg/pathprobe/paths.go, line 104 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-MgvsNOd5NoohNNukxBX:-Mp6w2gC3wjRNVzEmBkI:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/pkg/pathprobe/paths.go#L104)):*

We are using the functional options pattern to make our APIs ergonomic and extendable.

Please us the pattern that is used for example here: https://github.com/scionproto/scion/blob/b5954eafc987875b1cdea6a85f66fae15a54178e/go/pkg/app/path/path.go#L290-L298
</blockquote></details>

Done.
___
*[go/pkg/router/dataplane.go, line 1413 at r5](https://reviewable.io/reviews/scionproto/scion/4079#-MoEGq0ECnNndmyCLOF3:-Mp6w5xF3JbED5CB8F68:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/a2f12ba142365d3a81611cc5d8255ace2c44b78c/go/pkg/router/dataplane.go#L1413)):*

This means that we always allocate the error, even if we do not run into a problem later on.
This puts unnecessary pressure on the garbage collector that we should avoid as much as possible.

Granted, the code in this package is not optimal regarding allocations, but we should still not add more to it.
</blockquote></details>

Done.

mawyss (commented):
Updated this PR to the recent changes in snet (https://github.com/scionproto/scion/commit/7663d2e9fa132afa0a5f8de2aabbb9e583b0275e).
I also tried to incorporate the feedback by @matzf from https://github.com/scionproto/scion/issues/4129.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079#-:-Mw6wiLW5scjCTDH3C9d:b-m0n82u)* status: 3 of 34 files reviewed, 11 unresolved discussions (waiting on @matzf and @oncilla)

oncilla (changes requested):
> Reviewed 2 of 15 files at r9, 15 of 23 files at r11.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 20 of 34 files reviewed, 13 unresolved discussions (waiting on @matzf, @mawyss, and @oncilla)

___
*[go/lib/snet/path.go, line 120 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvnwpc-8sZKzEv52j_:-Mwl3y3p8yx-0SfwbPm5:b-a9oz5u) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/snet/path.go#L120)):*

Done.

Good observation. The `PathMetadata` was used to store the PHVF/LHVF, before transmitting them via grpc.
For consistency, the PHVF/LHVF are now added to `spath.Path.EpicData` also there.
</blockquote></details>

I'm still worried about putting all of this data in the Metadata.
AFAICT, it is only used for path creation, but I think that can be solved in another way, that also makes it more convenient to use:


```go

// go/lib/snet/path/epic.go
package path

type EPIC struct {
	AuthPHVF []byte
	AuthLHVF []byte
	SCION    []byte

	mu      sync.Mutex
	active    bool 
	counter uint32
}


func (e *EPIC) SetPath(s *slayers.SCION) error {
  // Set SCION if !active
  // Set EPIC if active
}

func IsEPIC(path interface{}) bool {
  _, ok := path.(*EPIC)
  return ok
}

func ActivateEPIC(path interface{}) error {
  p, ok := path.(*EPIC)
  if !ok {
    return serrors.New("not an epic path", "type", fmt.Sprintf("%T", path))
  } 
  p.mtx.Lock()
  defer p.mtx.Unlock()
  p.active = true
  return nil
}


// go/integration/end2end/main.go
...
if epic {
    if err := snetpath.ActivateEPIC(path.Dataplane()); err != nil {
        return nil, serrors.New("activating epic path", err)
    }
    remote.Path = path.Dataplane()
```
___
*[go/lib/snet/path/epic.go, line 36 at r11](https://reviewable.io/reviews/scionproto/scion/4079#-Mwl9kVw3Hn89UqjK8MR:-Mwl9XGJ1wIYty1rdffw:b-upj58l) ([raw file](https://github.com/scionproto/scion/blob/306f5cbf58df551960ea10ce8dda6960f36f5a7e/go/lib/snet/path/epic.go#L36)):*
> ```Go
> }
> 
> func (e *EPIC) SetPath(s *slayers.SCION) error {
> ```

See my comments above.
___
*[go/lib/snet/path/scion.go, line 48 at r11](https://reviewable.io/reviews/scionproto/scion/4079#-Mwl9meo86wV7eCgeh_h:-Mwl9meo86wV7eCgeh_i:b-v3wr3q) ([raw file](https://github.com/scionproto/scion/blob/306f5cbf58df551960ea10ce8dda6960f36f5a7e/go/lib/snet/path/scion.go#L48)):*
> ```Go
> }
> 
> func (p SCION) NewEPICDataplanePath(auths snet.EpicAuths) (*EPIC, error) {
> ```

Drop this, see  me comments above.
___
*[go/pkg/app/path/path.go, line 362 at r11](https://reviewable.io/reviews/scionproto/scion/4079#-MwlANR919-ZTBT7352e:-MwlANRAEzIDW1zJBG-T:blp2ods) ([raw file](https://github.com/scionproto/scion/blob/306f5cbf58df551960ea10ce8dda6960f36f5a7e/go/pkg/app/path/path.go#L362)):*
> ```Go
> }
> 
> func WithEpic(epic bool) Option {
> ```

WithEPIC
___
*[go/pkg/pathprobe/paths.go, line 129 at r11](https://reviewable.io/reviews/scionproto/scion/4079#-MwlAQZbDzNlBiHA6gy-:-MwlAQZcDXRfwROn0tpq:blp2ods) ([raw file](https://github.com/scionproto/scion/blob/306f5cbf58df551960ea10ce8dda6960f36f5a7e/go/pkg/pathprobe/paths.go#L129)):*
> ```Go
> }
> 
> func WithEpic(epic bool) Option {
> ```

WithEPIC
___
*[go/pkg/pathprobe/paths.go, line 210 at r11](https://reviewable.io/reviews/scionproto/scion/4079#-MwlAbhl8u4D5e-KDvmq:-MwlAbhl8u4D5e-KDvmr:brwl30z) ([raw file](https://github.com/scionproto/scion/blob/306f5cbf58df551960ea10ce8dda6960f36f5a7e/go/pkg/pathprobe/paths.go#L210)):*
> ```Go
> 			// Send probe for each path.
> 			for _, path := range paths {
> 				originalPath, ok := path.Dataplane().(snetpath.SCION)
> ```

This needs to handle both snetpath.SCION and snetpath.EPIC with my suggestion above.

oncilla (commented):
> Reviewed 3 of 27 files at r8.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 23 of 34 files reviewed, 13 unresolved discussions (waiting on @matzf, @mawyss, and @oncilla)

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 23 of 34 files reviewed, 13 unresolved discussions (waiting on @matzf, @mawyss, and @oncilla)

___
*[go/lib/snet/path.go, line 120 at r2](https://reviewable.io/reviews/scionproto/scion/4079#-Mgvnwpc-8sZKzEv52j_:-MwlssS76SwivBdyLuH2:bvor0sr) ([raw file](https://github.com/scionproto/scion/blob/b22d88715116b3a492c6ad57a58db59e17be79e4/go/lib/snet/path.go#L120)):*

I'm still worried about putting all of this data in the Metadata.
AFAICT, it is only used for path creation, but I think that can be solved in another way, that also makes it more convenient to use:


```go

// go/lib/snet/path/epic.go
package path

type EPIC struct {
	AuthPHVF []byte
	AuthLHVF []byte
	SCION    []byte

	mu      sync.Mutex
	active    bool 
	counter uint32
}


func (e *EPIC) SetPath(s *slayers.SCION) error {
  // Set SCION if !active
  // Set EPIC if active
}

func IsEPIC(path interface{}) bool {
  _, ok := path.(*EPIC)
  return ok
}

func ActivateEPIC(path interface{}) error {
  p, ok := path.(*EPIC)
  if !ok {
    return serrors.New("not an epic path", "type", fmt.Sprintf("%T", path))
  } 
  p.mtx.Lock()
  defer p.mtx.Unlock()
  p.active = true
  return nil
}


// go/integration/end2end/main.go
...
if epic {
    if err := snetpath.ActivateEPIC(path.Dataplane()); err != nil {
        return nil, serrors.New("activating epic path", err)
    }
    remote.Path = path.Dataplane()
```








</blockquote></details>

Here, I implemented Matthias' suggestion from https://github.com/scionproto/scion/issues/4129:

"Creating this DataplanePath as EPIC path (in the sciond client library) whenever the authenticator info is available seems a bit at odds with the idea that this would generalize to other path types as well. My suggestion would be to always return the default DataplanePath as SCION and squirrel away the EPIC authenticator information somewhere in the PathMetadata struct. To enable EPIC, we could then replace the dataplane path with the result of some NewEPICDataplanePath function."

When considering additional path types such as EPIC-SAPV, it will become unclear which path type should be created during path construction.
Or how would that be solved then?

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 6 of 34 files reviewed, 13 unresolved discussions (waiting on @matzf, @mawyss, and @oncilla)

___
*[go/pkg/app/path/path.go, line 362 at r11](https://reviewable.io/reviews/scionproto/scion/4079#-MwlANR919-ZTBT7352e:-MxsrFJ147J-2S8Jo32s:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/306f5cbf58df551960ea10ce8dda6960f36f5a7e/go/pkg/app/path/path.go#L362)):*

WithEPIC
</blockquote></details>

Done.
___
*[go/pkg/pathprobe/paths.go, line 129 at r11](https://reviewable.io/reviews/scionproto/scion/4079#-MwlAQZbDzNlBiHA6gy-:-MxsrN8nDqAN0GlyLS8c:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/306f5cbf58df551960ea10ce8dda6960f36f5a7e/go/pkg/pathprobe/paths.go#L129)):*

WithEPIC
</blockquote></details>

Done.

oncilla (changes requested):
> Reviewed 1 of 26 files at r1, 1 of 27 files at r8, 3 of 15 files at r13, 23 of 23 files at r15.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: all files reviewed, 5 unresolved discussions (waiting on @matzf, @mawyss, and @oncilla)

___
*[go/lib/snet/path/scion.go, line 48 at r15](https://reviewable.io/reviews/scionproto/scion/4079#-Mxx22LM8gny21U8Z4BC:-Mxx22LM8gny21U8Z4BD:blotqiv) ([raw file](https://github.com/scionproto/scion/blob/3e88c8a55269798b2afaa9a8ce4fc975263b539d/go/lib/snet/path/scion.go#L48)):*
> ```Go
> }
> 
> func (p SCION) NewEPICDataplanePath(auths snet.EpicAuths) (*EPIC, error) {
> ```

Make this a function instead of a method and move it to the epic.go file
___
*[go/pkg/app/path/path.go, line 362 at r11](https://reviewable.io/reviews/scionproto/scion/4079#-MwlANR919-ZTBT7352e:-Mxx2EPMESGI8vMDkNq1:b3lx37y) ([raw file](https://github.com/scionproto/scion/blob/306f5cbf58df551960ea10ce8dda6960f36f5a7e/go/pkg/app/path/path.go#L362)):*

Done.
</blockquote></details>

You still need to capitalize EPIC.

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4079)* status: 26 of 34 files reviewed, 3 unresolved discussions (waiting on @matzf and @oncilla)

___
*[go/pkg/app/path/path.go, line 362 at r11](https://reviewable.io/reviews/scionproto/scion/4079#-MwlANR919-ZTBT7352e:-Mxy50oV50NAjCMwuyP2:bpgtsvc) ([raw file](https://github.com/scionproto/scion/blob/306f5cbf58df551960ea10ce8dda6960f36f5a7e/go/pkg/app/path/path.go#L362)):*

You still need to capitalize EPIC.
</blockquote></details>

Sorry, missed that somehow. Done now.
___
*[go/lib/snet/path/scion.go, line 48 at r15](https://reviewable.io/reviews/scionproto/scion/4079#-Mxx22LM8gny21U8Z4BC:-Mxy95ccBV9G2MA-v1NY:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3e88c8a55269798b2afaa9a8ce4fc975263b539d/go/lib/snet/path/scion.go#L48)):*

Make this a function instead of a method and move it to the epic.go file
</blockquote></details>

Done.

PR #4096, opened 2021-08-06
router: non-blocking write for packet forwarding

matzf:
The structure of the router's dataplane is currently very simple, with a
single processing goroutine per (internal and external) router
interface. This goroutine reads and processes packets arriving on this
interface and finally writes it out to a (different) interface. This
write was a blocking call, i.e. the call would block until the packet
could be copied into the socket's sendbuffer. While the processing
goroutine is blocked, it cannot read and process packets, and incoming
packets will eventually be dropped.
With this setup, a slow (for writing) router interface can block
processing packets for all router interfaces, affecting traffic that is
not even destined to the slow interface.
Worse yet, with the polling based I/O implementation in golang, the
write may be blocked for longer than what would immediately be needed to
write the packet; in linux, poll only considers a socket as writable
once the sendbuffer has been drained to half. This can lead to long
stalls during which no packets can be processed.

Using a non-blocking write call when forwarding packets fixes these
issues and allows to keep the simple structure of the router's
dataplane.

 Additionally, remove the redundant underlay/conn/mock_conn.MockConn and
 use the router/mock_router.MockBatchConn in the router tests instead.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4096)


oncilla (changes requested):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4096)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[go/pkg/router/dataplane.go, line 496 at r1](https://reviewable.io/reviews/scionproto/scion/4096#-Mgq-50K7Fe-48Yh1E3Z:-Mgq-50K7Fe-48Yh1E3_:b-6zam16) ([raw file](https://github.com/scionproto/scion/blob/47418cb072412838597c2576e891b21ebb3be024/go/pkg/router/dataplane.go#L496)):*
> ```Go
> 				// supports MSG_DONTWAIT.
> 				writeMsgs[0].Buffers[0] = result.OutPkt
> 				writeMsgs[0].Addr = nil
> ```

could you not do `writeMsgs[0].Addr = result.OutAddr` to get rid of the if branch?
___
*[go/pkg/router/dataplane.go, line 509 at r1](https://reviewable.io/reviews/scionproto/scion/4096#-Mgq-chM1bZw_GCijoIv:-Mgq-chM1bZw_GCijoIw:bvto2d2) ([raw file](https://github.com/scionproto/scion/blob/47418cb072412838597c2576e891b21ebb3be024/go/pkg/router/dataplane.go#L509)):*
> ```Go
> 				} else if err != nil {
> 					log.Debug("Error writing packet", "err", err)
> 					// error metric
> ```

I think we should also increment the dropped packet counter here.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4096)* status: 9 of 10 files reviewed, 2 unresolved discussions (waiting on @oncilla)

___
*[go/pkg/router/dataplane.go, line 496 at r1](https://reviewable.io/reviews/scionproto/scion/4096#-Mgq-50K7Fe-48Yh1E3Z:-MgyIdHGDku-HMpaotuq:bkvcxv6) ([raw file](https://github.com/scionproto/scion/blob/47418cb072412838597c2576e891b21ebb3be024/go/pkg/router/dataplane.go#L496)):*

could you not do `writeMsgs[0].Addr = result.OutAddr` to get rid of the if branch?
</blockquote></details>

Unfortunately not; `writeMsgs[0].Addr` is a `net.Addr` while `result.OutAddr` is `*net.UDPAddr`. If `result.OutAddr` is `nil`, `net.Addr(result.OutAddr) != nil`.  There is a check for `nil` on the address somewhere in the x/net library in `WriteBatch` and using a non-nil-interface-but-nil-pointer address will result in a panic.
___
*[go/pkg/router/dataplane.go, line 509 at r1](https://reviewable.io/reviews/scionproto/scion/4096#-Mgq-chM1bZw_GCijoIv:-MgyLsYw1B7j1TKrnUeN:b-nnphsn) ([raw file](https://github.com/scionproto/scion/blob/47418cb072412838597c2576e891b21ebb3be024/go/pkg/router/dataplane.go#L509)):*

I think we should also increment the dropped packet counter here.
</blockquote></details>

Done, combined the error branches.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4096)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[go/pkg/router/dataplane.go, line 496 at r1](https://reviewable.io/reviews/scionproto/scion/4096#-Mgq-50K7Fe-48Yh1E3Z:-MgyRGfJC7t1FuxMiVjW:bx4covr) ([raw file](https://github.com/scionproto/scion/blob/47418cb072412838597c2576e891b21ebb3be024/go/pkg/router/dataplane.go#L496)):*

Unfortunately not; `writeMsgs[0].Addr` is a `net.Addr` while `result.OutAddr` is `*net.UDPAddr`. If `result.OutAddr` is `nil`, `net.Addr(result.OutAddr) != nil`.  There is a check for `nil` on the address somewhere in the x/net library in `WriteBatch` and using a non-nil-interface-but-nil-pointer address will result in a panic.
</blockquote></details>

Uuuuh, fun. I haven't ran into a typed nil in a long time.

Please add the reasoning as a comment for the posterity.
Otherwise, people might try to change it in the future.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4096)* status: all files reviewed, 1 unresolved discussion (waiting on @oncilla)

___
*[go/pkg/router/dataplane.go, line 496 at r1](https://reviewable.io/reviews/scionproto/scion/4096#-Mgq-50K7Fe-48Yh1E3Z:-MgzLYGBE3x7lu4KKhqu:b-kgbc43) ([raw file](https://github.com/scionproto/scion/blob/47418cb072412838597c2576e891b21ebb3be024/go/pkg/router/dataplane.go#L496)):*

Uuuuh, fun. I haven't ran into a typed nil in a long time.

Please add the reasoning as a comment for the posterity.
Otherwise, people might try to change it in the future.


</blockquote></details>

Ok sure, done. I kept it terse, hoping this will the be just enough to ring a bell... A full, specific, explanation would just seem to be distracting.

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4096)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

___
*[go/pkg/router/dataplane.go, line 496 at r1](https://reviewable.io/reviews/scionproto/scion/4096#-Mgq-50K7Fe-48Yh1E3Z:-MgzPMqg15YAyTgKxUIQ:b-i85hio) ([raw file](https://github.com/scionproto/scion/blob/47418cb072412838597c2576e891b21ebb3be024/go/pkg/router/dataplane.go#L496)):*

Ok sure, done. I kept it terse, hoping this will the be just enough to ring a bell... A full, specific, explanation would just seem to be distracting.
</blockquote></details>

Should be good enough.

oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4096#-:-MgzPYFZ-o-T1kCT2fzR:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4109, opened 2021-09-09
doc: CASA

mawyss:
Introduce cryptographic agility for SCION ASes (CASA).
CASA allows ASes to have more flexibility in their choice
of cryptographic algorithms by extending the beacons
with information about certain algorithm categories.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4109)


marcfrei:
___
*[doc/CASA.rst, line 81 at r1](https://reviewable.io/reviews/scionproto/scion/4109#-Mj9kOzEBmOHXNCw9CeN:-Mj9kOzEBmOHXNCw9CeO:bijz2la) ([raw file](https://github.com/scionproto/scion/blob/9f55f250abf885334c2b4a36743d28b0d493805c/doc/CASA.rst#L81)):*
> ```
> Note the asymmetry in the CASA design: there is in general one side
> with higher and one with lower efficiency requirements.
> ```
Is efficiency the only criterium we are interested in here? What about, e.g., AS specific requirements for higher security standards or compliance with specific security policies?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4109)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @mawyss)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4109#-MjDctMfA16Xen7hjW1I:-MjDctMh7UYolKCYkUt8:b-oqojyw) (no related file):*
Some unstructured thoughts on this:

* For DRKey, this approach of announcing the algorithm choices with the beaconing appears to be hard to use in practice.   When the CS derives a key for a remote issuer AS, it would need to fetch the paths to that AS to determine the PRF to be used for the derivation.  When changing the PRF, the DRKey epochs would have to be synchronised with the path lifetime.

* An AS could, intentionally or accidentally, announce different CASA choices in beacons with overlapping validity periods -- how should this be resolved? For EPIC, the perhaps natural choice would be to use whatever was announced with the particular path we're adding the EPIC data to, but for DRKey there is no such natural choice.

* For DRKey, it would be easily possible to specify the PRF as part of the Level-1 key exchange. This would avoid the issues above and would also allow some more specific configuration (e.g. use different PRFs depending on the DRKey protocol type).

* So if we would handle DRKey separately, this leaves only EPIC. Are there other potential applications on the radar? Wouldn't it be easier to include this algorithm specification in the beacon extension used for EPIC, instead of defining a generic approach from only a single use case?

* Maybe this is a brainfart; how would this system look like if these CASA options would be included in the AS certificates instead of in the beacons?  

---

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4109)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @marcfrei and @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4109#-MjDctMfA16Xen7hjW1I:-MqxJtC74Pfq3hawdFhc:bf1fxc6) (no related file):*

Some unstructured thoughts on this:

* For DRKey, this approach of announcing the algorithm choices with the beaconing appears to be hard to use in practice.   When the CS derives a key for a remote issuer AS, it would need to fetch the paths to that AS to determine the PRF to be used for the derivation.  When changing the PRF, the DRKey epochs would have to be synchronised with the path lifetime.

* An AS could, intentionally or accidentally, announce different CASA choices in beacons with overlapping validity periods -- how should this be resolved? For EPIC, the perhaps natural choice would be to use whatever was announced with the particular path we're adding the EPIC data to, but for DRKey there is no such natural choice.

* For DRKey, it would be easily possible to specify the PRF as part of the Level-1 key exchange. This would avoid the issues above and would also allow some more specific configuration (e.g. use different PRFs depending on the DRKey protocol type).

* So if we would handle DRKey separately, this leaves only EPIC. Are there other potential applications on the radar? Wouldn't it be easier to include this algorithm specification in the beacon extension used for EPIC, instead of defining a generic approach from only a single use case?

* Maybe this is a brainfart; how would this system look like if these CASA options would be included in the AS certificates instead of in the beacons?  

</blockquote></details>

Thank you for your feedback, and sorry for the late response.
-  Regarding beacons with overlapping validity periods: this would actually be the common case. The idea is to run both algorithms during the transition period, i.e., a border router would calculate both the old and the new MAC. This reduces the security by one bit (as two MACs are valid now), but omits a hard transition and hence allows to omit coordination with DRKey epochs. So a PRF is valid to be used for as long as a segment is valid.
- Yes, with DRKey we would need to do some adoptions. A simple solution could be to set  the validity of the algorithms announced in the beacon to "End of segment validity + DRKey epoch duration + DRKey grace period + DRKey prefetching period". Idea behind this: to fetch a DRKey, the requesting AS already needs a path to the issuing AS, hence it knows the cryptographic algorithms. If the issuing AS now supports the old algorithms long enough, then the fetching AS can still derive valid keys for its end hosts.
- It might indeed be simpler to handle DRKey separately, and specify the PRFs etc. in the key exchange. With CASA I tried to find a solution to avoid every protocol having to define the PRF exchange on its own, but if this is too cumbersome for DRKey, we could make an exception?
- The systems that can potentially be supported by CASA are: DRKey, EPIC-HP, EPIC-SAPV, COLIBRI, (authenticated) SCMP/SPAO, and of course any future system.
Comment for SPAO: while the cryptographic algorithm is specified in the SPAO itself (which is important and should stay this way), it might still be desirable for an AS to indicate its preferred algorithm.
- For the AS certificate: this could be an option I guess, but I don't know about its details. One thing that is maybe undesirable is the rather long validity of the AS certificates?

---
___
*[doc/CASA.rst, line 81 at r1](https://reviewable.io/reviews/scionproto/scion/4109#-Mj9kOzEBmOHXNCw9CeN:-MqxJPhj-Hg92wNWCZ08:bw7b15c) ([raw file](https://github.com/scionproto/scion/blob/435f183a7bbd2edee70cf908709f3125f581f268/doc/CASA.rst#L81)):*

> ```
> Note the asymmetry in the CASA design: there is in general one side
> with higher and one with lower efficiency requirements.
> ```
Is efficiency the only criterium we are interested in here? What about, e.g., AS specific requirements for higher security standards or compliance with specific security policies? 
</blockquote></details>

Thank you for pointing that out: yes, there may be different motivations for cryptographic agility. I tried to make this more explicit and added your points to design document.

matzf:
Discussion in the proposal review meeting; as far as we are aware, there is currently no practical need for cryptographic agility for the SCION subsystems. Cryptographic agility is also perhaps no longer generally seen as a good thing, so without specific requirements it does not seem justified to dive deeper into this and build this feature now. As this proposes an extension mechanism anyway, there should not be great harm in deferring this.
Closing this as rejected for now.

PR #4110, opened 2021-09-09
doc: epic-sapv

mawyss:
Specifies EPIC-SAPV (Source Authentication and Path Validation).
EPIC-SAPV enables source authentication at border routers, packet
authentication and path validation at the destination host, and
optional path validation at the source host.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4110)


matzf:
Discussion in the proposal review meeting; as far as we are aware, there is currently no practical requirement to include EPIC-SAPV in this SCION implementation. As this is a major subsystem, adding this would incur both a significant maintenance burden and also increased surface area for bugs and attacks, which we don't think is justified without more concrete evidence for the practicality of this system.
Our recommendation is to build and develop this feature in a fork of scionproto/scion. Perhaps we can find a good place to link to such a fork, so that this can still find sufficient publicity.
Please note that this means that we reject including the implementation of this feature into scionproto/scion  for now. But discussions on the design and properties of such a system e.g. in issues in this repository is still welcome.

PR #4132, opened 2021-11-19
router/bfd: ensure no data race on received packet data

### LINKED ISSUE #4131: Possible race condition when BFD packets are processed
Hi all,

I think I have found a possible race condition that may appear if a router is under heavy load.

The `processPkt` method takes as input argument a buffer that is being reused.
https://github.com/scionproto/scion/blob/362c4622084037cd1d3ac593d2d78128a86638fd/go/pkg/router/dataplane.go#L473
Now, if a BFD packet is received, it is here decoded to a SCION packet.
Then if `processIntraBFD` is called, it receives the parameter `pld`. `pld` itself is a sub-slice of `p.rawPkt` (`p.lastLayer.LayerPayload()` returns a sub-slice of `p.rawPkt`) and `p.rawPkt` is set to `p.Buffers[0][:p.N]`
https://github.com/scionproto/scion/blob/362c4622084037cd1d3ac593d2d78128a86638fd/go/pkg/router/dataplane.go#L597-L616
Now, the `processIntraBFD` method deserializes `pld` to a BFD packet. However the `DecodeFromBytes` method embeds some sub-slices of `pld` in the BFD packet. The BFD packet is then sent to a Message channel. The thread that receives the packet from the channel then obtains a BFD packet which contains sub-slices of `p.Buffers[0][:p.N]`. Now If the router receives many messages, the buffer will be reused and will overwrite fields in the BFD packet.
https://github.com/scionproto/scion/blob/362c4622084037cd1d3ac593d2d78128a86638fd/go/pkg/router/dataplane.go#L655-L675

matzf:
The processing of the incoming BFD messages passed a BFD message as a
`*layers.BFD` to a channel, from where it would be asynchronously be
processed. The `layers.BFD` object contains slices that reference the
underlying packet data.
While we were not accessing any of the these slice field, this was a
lurking issue. Fixed by passing a separate struct not containing any of
the slice fields through the channel.
Messages are now validated synchronously in the router dataplane, and
only the processing depending on the BFD Session state happens
asynchronously.

Moving the validation (`shouldDiscard`) to be performed synchronously in
the dataplane reveiled that this logic was not faithfully following the
specifications in the RFC 5880, Section 6.8.6 (as referenced in the doc
comment) -- the RFC states:

  > If the Your Discriminator field is zero and the State field is not
  > Down or AdminDown, the packet MUST be discarded.

In `shouldDiscard`, the `localState` was used in the corresponding test
instead of the `State` field of the received message.

Also in this change:
- remove the unused bfd.Controller. Adapt TestSessionDebootstrap to
  not use this Controller.
- reuse layers.BFD in dataplane process{intra,inter}BFD

Fixes #4131

---
Note: first version of this PR included optimizations for the BFD message sender. Quite separate, so split off.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4132)


oncilla (approved):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4132)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4136, opened 2021-12-11
snet: drop spath

mawyss:
The `spath` package assumes dataplane paths to be static.
With path types such as EPIC and Colibri, certain fields have to be modifiable on a per-packet basis however.

This PR removes `spath` and allows for stateful and more dynamic dataplane paths.

Implements https://github.com/scionproto/scion/issues/4129.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4136)


oncilla (changes requested):
> Reviewed 81 of 89 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4136)* status: 81 of 89 files reviewed, 9 unresolved discussions (waiting on @mawyss and @oncilla)

___
*[go/cs/beaconing/writer_test.go, line 252 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWHRcx9FIy5jDY4vJa:-MrWHRcy0nsPLe8kNZEm:btwtgbs) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/cs/beaconing/writer_test.go#L252)):*
> ```Go
> 					var path scion.Decoded
> 					scionPath, ok := s.Addr.Path.(snetpath.SCION)
> 					assert.Equal(t, true, ok)
> ```

make this a `require.True(t, ok)`

It does not make sense to continue with the test, if we fail to type assert.
Note, that `assert` fails the test, but continues the execution `require` stops the execution immediately.
___
*[go/lib/infra/modules/combinator/combinator.go, line 76 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWUxLg46q0XSQlsouc:-MrWUxLh2d62I6pS05eI:bhpsas1) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/lib/infra/modules/combinator/combinator.go#L76)):*
> ```Go
> type Path struct {
> 	Dst       addr.IA
> 	ScionPath path.SCION
> ```

We follow the conventions in https://github.com/golang/go/wiki/CodeReviewComments#initialisms

Thus, this field would need to be `SCIONPath`
___
*[go/lib/snet/packet.go, line 33 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWVkCL117OT7_hDgtl:-MrWVkCL117OT7_hDgtm:b-oh2b7s) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/lib/snet/packet.go#L33)):*
> ```Go
> type Payload interface {
> 	toLayers(scn *slayers.SCION) []gopacket.SerializableLayer
> 	Length() int
> ```

there is no need for this to be exported yet.
Make this lower private to avoid extending the Payload type's API unnecessarily.
___
*[go/lib/snet/packet.go, line 337 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWWGG78yEHd9sPYcz2:-MrWWGG78yEHd9sPYcz3:b2z2i3w) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/lib/snet/packet.go#L337)):*
> ```Go
> // Consumers of SCIONPacketConn need to extract the path and handle it appropriately.
> func (r RawPath) SetPath(s *slayers.SCION) error {
> 	p, err := path.NewPath(r.PathType)
> ```

I would prefer if this just returned an error.
It forces the package client to do something reasonable with the path before it can be serialized again.

The implementation that is here does not do the right thing in 99% of the cases, because before you would want to use the path in the reply package, you need to reverse the path first.
___
*[go/lib/snet/path/onehop.go, line 46 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWY8G25l11m-oEGUyu:-MrWY8G25l11m-oEGUyv:b-m1tv35) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/lib/snet/path/onehop.go#L46)):*
> ```Go
> 
> // NewOneHop creates a onehop path that has the first hopfield initialized.
> func NewOneHop(egress uint16, timestamp time.Time, expiration uint8,
> ```

We use the following formatting nowadays to make arguments easier to read.

```go
func NewOneHop(
    egress uint16,
    timestamp time.Time,
    expiration uint8,
    mac hash.Hash,
) (OneHop, error) {

    segID, err := ...

```
___
*[go/lib/snet/path/path.go, line 42 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWZFPD7ayFjnv9dp60:-MrWZFPD7ayFjnv9dp61:buljhx3) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/lib/snet/path/path.go#L42)):*
> ```Go
> 	// MaxExpirationTime is the maximum absolute expiration time of SCION hop
> 	// fields.
> 	MaxExpirationTime = time.Unix(maxTimestamp, 0).Add(path.ExpTimeToDuration(math.MaxUint8))
> ```

Let's move this to the place where it is actually used in the combinator as a private constant for now.

I don't think it really fits here, and I also don't think much more than the combinator will ever care about this definition.
Thus, it makes sense to have it private there.
___
*[go/pkg/hiddenpath/beaconwriter_test.go, line 75 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWb92o6ylcC6p2UMuA:-MrWb92o6ylcC6p2UMuB:b-bz0kd2) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/pkg/hiddenpath/beaconwriter_test.go#L75)):*
> ```Go
> 		var path scion.Decoded
> 		scionPath, ok := a.Path.(snetpath.SCION)
> 		assert.Equal(t, true, ok)
> ```

`require.True(t, err)`
___
*[go/pkg/pathprobe/paths.go, line 90 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWbTSi3Y4fqUNLvu8f:-MrWbTSi3Y4fqUNLvu8g:b-v6ylwt) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/pkg/pathprobe/paths.go#L90)):*
> ```Go
> 		if !isEmpty {
> 			scionPath, isScion := path.Dataplane().(snetpath.SCION)
> 			if isScion && len(scionPath.Raw) > 0 {
> ```

I don't think this check is necessary.

The documentation of the function specifically mentions that only empty paths are filtered out.
Here, you now filter out any non-scion paths.


I think the loop body should just be:

```go
		
if _, isEmpty := path.Dataplane().(snetpath.Empty); !isEmpty {
  filtered = append(filtered, path)
}
```

In the existing code, raw == 0 could mean empty.
But with the new types we introduce in this PR, this should not be the case anymore. 
___
*[go/scion-pki/certs/renew.go, line 668 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWe42m9vZ0Rx7FJBaf:-MrWe42n5QH6ayJ2vfEK:b-grvi5h) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/scion-pki/certs/renew.go#L668)):*
> ```Go
> 		IA:   dstIA,
> 		SVC:  addr.SvcCS,
> 		Path: snetpath.Empty{},
> ```

I'm a bit surprised by that.

How does this work exactly?
If we need to go to a remote AS, we would need a path to get there, right?

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4136)* status: 69 of 89 files reviewed, 9 unresolved discussions (waiting on @oncilla)

___
*[go/cs/beaconing/writer_test.go, line 252 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWHRcx9FIy5jDY4vJa:-MsfBKoeAGGK9yfZeA9z:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/cs/beaconing/writer_test.go#L252)):*

make this a `require.True(t, ok)`

It does not make sense to continue with the test, if we fail to type assert.
Note, that `assert` fails the test, but continues the execution `require` stops the execution immediately.

</blockquote></details>

Done.
___
*[go/lib/infra/modules/combinator/combinator.go, line 76 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWUxLg46q0XSQlsouc:-MsfBqdY7wxJwn-DljQ8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/lib/infra/modules/combinator/combinator.go#L76)):*

We follow the conventions in https://github.com/golang/go/wiki/CodeReviewComments#initialisms

Thus, this field would need to be `SCIONPath`
</blockquote></details>

Done.
___
*[go/lib/snet/packet.go, line 33 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWVkCL117OT7_hDgtl:-MsfCiWm5pwo8tp3tG1p:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/lib/snet/packet.go#L33)):*

there is no need for this to be exported yet.
Make this lower private to avoid extending the Payload type's API unnecessarily.
</blockquote></details>

Done.
___
*[go/lib/snet/packet.go, line 337 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWWGG78yEHd9sPYcz2:-MskQY3BEBEIkuPB0Dk7:bnwqwv1) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/lib/snet/packet.go#L337)):*

I would prefer if this just returned an error.
It forces the package client to do something reasonable with the path before it can be serialized again.

The implementation that is here does not do the right thing in 99% of the cases, because before you would want to use the path in the reply package, you need to reverse the path first.
</blockquote></details>

Done. I adapted `packet_test.go` accordingly.
___
*[go/lib/snet/path/onehop.go, line 46 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWY8G25l11m-oEGUyu:-MsfEaLW848-zWujLbk4:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/lib/snet/path/onehop.go#L46)):*

We use the following formatting nowadays to make arguments easier to read.

```go
func NewOneHop(
    egress uint16,
    timestamp time.Time,
    expiration uint8,
    mac hash.Hash,
) (OneHop, error) {

    segID, err := ...

```
</blockquote></details>

Done.
___
*[go/lib/snet/path/path.go, line 42 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWZFPD7ayFjnv9dp60:-Msikfz8B00RJZ_jCVFu:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/lib/snet/path/path.go#L42)):*

Let's move this to the place where it is actually used in the combinator as a private constant for now.

I don't think it really fits here, and I also don't think much more than the combinator will ever care about this definition.
Thus, it makes sense to have it private there.


</blockquote></details>

Done.
___
*[go/pkg/hiddenpath/beaconwriter_test.go, line 75 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWb92o6ylcC6p2UMuA:-Msil5FYC_stmdBJKro8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/pkg/hiddenpath/beaconwriter_test.go#L75)):*

`require.True(t, err)`
</blockquote></details>

Done.
___
*[go/pkg/pathprobe/paths.go, line 90 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWbTSi3Y4fqUNLvu8f:-Msinj355SJZpXRdbkX6:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/pkg/pathprobe/paths.go#L90)):*

I don't think this check is necessary.

The documentation of the function specifically mentions that only empty paths are filtered out.
Here, you now filter out any non-scion paths.


I think the loop body should just be:

```go
		
if _, isEmpty := path.Dataplane().(snetpath.Empty); !isEmpty {
  filtered = append(filtered, path)
}
```

In the existing code, raw == 0 could mean empty.
But with the new types we introduce in this PR, this should not be the case anymore. 



</blockquote></details>

Done.
___
*[go/scion-pki/certs/renew.go, line 668 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWe42m9vZ0Rx7FJBaf:-Msjxb7tD_IlmP645tYv:b28qplg) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/scion-pki/certs/renew.go#L668)):*

I'm a bit surprised by that.

How does this work exactly?
If we need to go to a remote AS, we would need a path to get there, right?



</blockquote></details>

As far as I am aware, I did not modify anything regarding the certificate renewal logic:
Before, the `Path` field of `snet.SVCAddr` was of type `spath.Path`, and if uninitialized, this field defaults to an empty path.
The path is later on replaced when necessary in `buildFullAddress` of `lib/infra/messenger/addr.go`.

Now that the field is a `DataplanePath`, its value does not default to an empty path anymore (but to nil), and hence the type checking in `buildFullAddress` does not do the right thing in such a case. That is why I had to explicitly initialize it to the emtpy path.

I added a check for nil paths now in `buildFullAddress` instead of initializing the field explicitly.

oncilla (changes requested):
> Reviewed 2 of 89 files at r1, 14 of 16 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4136)* status: 85 of 89 files reviewed, 1 unresolved discussion (waiting on @mawyss and @oncilla)

___
*[go/lib/infra/messenger/addr.go, line 136 at r2](https://reviewable.io/reviews/scionproto/scion/4136#-MtMakJUB8IxYaeBe8JS:-MtMakJUB8IxYaeBe8JT:b-95v58b) ([raw file](https://github.com/scionproto/scion/blob/5ce382d801dad7040be443fa8d6ccb259216190f/go/lib/infra/messenger/addr.go#L136)):*
> ```Go
> 
> 	if s.Path == nil {
> 		s.Path = path.Empty{}
> ```

This sets the path and it is immediately changed to something else on line 159.

you can change the if condition below:

```go
if _, isEmpty := s.Path.(path.Empty); !isEmpty && s.Path != nil {
```
___
*[go/scion-pki/certs/renew.go, line 668 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWe42m9vZ0Rx7FJBaf:-MtM_vEfD4d6DRooRZh7:bu1lmfl) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/scion-pki/certs/renew.go#L668)):*

As far as I am aware, I did not modify anything regarding the certificate renewal logic:
Before, the `Path` field of `snet.SVCAddr` was of type `spath.Path`, and if uninitialized, this field defaults to an empty path.
The path is later on replaced when necessary in `buildFullAddress` of `lib/infra/messenger/addr.go`.

Now that the field is a `DataplanePath`, its value does not default to an empty path anymore (but to nil), and hence the type checking in `buildFullAddress` does not do the right thing in such a case. That is why I had to explicitly initialize it to the emtpy path.

I added a check for nil paths now in `buildFullAddress` instead of initializing the field explicitly.

</blockquote></details>

Right. I think that belongs to buildFullAddress anyway, because we do path resolution there if the path is missing.

oncilla (commented):
> Reviewed 1 of 16 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4136)* status: 86 of 89 files reviewed, 1 unresolved discussion (waiting on @mawyss and @oncilla)

oncilla (changes requested):
> Reviewed 1 of 89 files at r1, 1 of 16 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4136)* status: 88 of 89 files reviewed, 3 unresolved discussions (waiting on @mawyss and @oncilla)

___
*[go/pkg/pathprobe/paths.go, line 121 at r2](https://reviewable.io/reviews/scionproto/scion/4136#-MtMlvnC0ml8zYsDqQLU:-MtMlvnC0ml8zYsDqQLV:buigesm) ([raw file](https://github.com/scionproto/scion/blob/5ce382d801dad7040be443fa8d6ccb259216190f/go/pkg/pathprobe/paths.go#L121)):*
> ```Go
> 		return nil, serrors.New("deadline required on ctx")
> 	}
> ```

Let's add another check that iterates through all paths and makes sure that all of them are of type `snetpath.SCION`.
`PathKey` is not able to handle other paths.
___
*[go/pkg/pathprobe/paths.go, line 183 at r2](https://reviewable.io/reviews/scionproto/scion/4136#-MtMnvZU9H-Ucv62fS6f:-MtMnvZU9H-Ucv62fS6g:b-lff249) ([raw file](https://github.com/scionproto/scion/blob/5ce382d801dad7040be443fa8d6ccb259216190f/go/pkg/pathprobe/paths.go#L183)):*
> ```Go
> 					Host: addr.HostFromIP(localIP),
> 				}
> 				if err := p.sendProbe(conn, localAddr, path, uint16(seqNr)); err != nil {
> ```

I think this `sendProbe` makes code unclearer than it needs to be.

The initialization of the packet is spread across multiple methods like this.
I.e., `seqNr` and `localAddr` are initialized here, but the rest of the packet is initialized in sendProbe.
Personally, I think it would just be easier to inline sendProbe.

Furthermore, it is a bit dangerous to set the alert flag and unset it like this.
In the error case, the path is still modified.

From an API perspective, it is not obvious that it might be modified, because it is hidden behind an interface.
At least to me it would be surprising.

I think we need to copy the path:

```go

originalPath, ok := path.Dataplane().(snetpath.SCION)
if !ok {//handle}

// Copy path to avoid editing the underlying path.
alertPath := snetpath.SCION{Raw: append([]byte(nil), originalPath.Raw...)}
if err := setAlertFlag; err != nil {//handle}

pkt := &snet.Packet{
	PacketInfo: snet.PacketInfo{
		Destination: snet.SCIONAddress{
			IA:   p.DstIA,
			Host: addr.SvcNone,
		},
		Source: snet.SCIONAddress{
			IA:   p.LocalIA,
			Host: addr.HostFromIP(localIP),
		}.
		Path:   alertingPath,
		Payload: snet.SCMPTracerouteRequest{
			Identifier: p.ID,
			Sequence:  uint16(atomic.AddInt32(&seq, 1)),
		},
	},
}

if err := scionConn.WriteTo(pkt, path.UnderlayNextHop()); err != nil {
  return err
}
```

oncilla (changes requested):
> Reviewed 1 of 89 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4136)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @mawyss)

___
*[go/integration/end2end/main.go, line 336 at r2](https://reviewable.io/reviews/scionproto/scion/4136#-MtMtrU7BNLU8d6CPo21:-MtMtrU7BNLU8d6CPo22:biv4xs2) ([raw file](https://github.com/scionproto/scion/blob/7d06a3efe494f96b9a9f0ac7787fa0f2dabfbde0/go/integration/end2end/main.go#L336)):*
> ```Go
> 		},
> 	}
> 	log.Info("sending ping", "attempt", n, "path", path)
> ```

let's leave this log line. We can drop the path k/v pair if it leads to issues.

mawyss (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4136)* status: 86 of 89 files reviewed, 4 unresolved discussions (waiting on @oncilla)

___
*[go/integration/end2end/main.go, line 336 at r2](https://reviewable.io/reviews/scionproto/scion/4136#-MtMtrU7BNLU8d6CPo21:-MtNTkNA4oZaXFBtABbX:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/7d06a3efe494f96b9a9f0ac7787fa0f2dabfbde0/go/integration/end2end/main.go#L336)):*

let's leave this log line. We can drop the path k/v pair if it leads to issues.
</blockquote></details>

Done.
___
*[go/lib/infra/messenger/addr.go, line 136 at r2](https://reviewable.io/reviews/scionproto/scion/4136#-MtMakJUB8IxYaeBe8JS:-MtNUm904qqZzq2S3w1b:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5ce382d801dad7040be443fa8d6ccb259216190f/go/lib/infra/messenger/addr.go#L136)):*

This sets the path and it is immediately changed to something else on line 159.

you can change the if condition below:

```go
if _, isEmpty := s.Path.(path.Empty); !isEmpty && s.Path != nil {
```

</blockquote></details>

Done.
___
*[go/pkg/pathprobe/paths.go, line 121 at r2](https://reviewable.io/reviews/scionproto/scion/4136#-MtMlvnC0ml8zYsDqQLU:-MtNZjGu8nT74D00RyKG:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5ce382d801dad7040be443fa8d6ccb259216190f/go/pkg/pathprobe/paths.go#L121)):*

Let's add another check that iterates through all paths and makes sure that all of them are of type `snetpath.SCION`.
`PathKey` is not able to handle other paths.

</blockquote></details>

Done.
___
*[go/pkg/pathprobe/paths.go, line 183 at r2](https://reviewable.io/reviews/scionproto/scion/4136#-MtMnvZU9H-Ucv62fS6f:-MtO5tcZ-SnO5w3kzG0v:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5ce382d801dad7040be443fa8d6ccb259216190f/go/pkg/pathprobe/paths.go#L183)):*

I think this `sendProbe` makes code unclearer than it needs to be.

The initialization of the packet is spread across multiple methods like this.
I.e., `seqNr` and `localAddr` are initialized here, but the rest of the packet is initialized in sendProbe.
Personally, I think it would just be easier to inline sendProbe.

Furthermore, it is a bit dangerous to set the alert flag and unset it like this.
In the error case, the path is still modified.

From an API perspective, it is not obvious that it might be modified, because it is hidden behind an interface.
At least to me it would be surprising.

I think we need to copy the path:

```go

originalPath, ok := path.Dataplane().(snetpath.SCION)
if !ok {//handle}

// Copy path to avoid editing the underlying path.
alertPath := snetpath.SCION{Raw: append([]byte(nil), originalPath.Raw...)}
if err := setAlertFlag; err != nil {//handle}

pkt := &snet.Packet{
	PacketInfo: snet.PacketInfo{
		Destination: snet.SCIONAddress{
			IA:   p.DstIA,
			Host: addr.SvcNone,
		},
		Source: snet.SCIONAddress{
			IA:   p.LocalIA,
			Host: addr.HostFromIP(localIP),
		}.
		Path:   alertingPath,
		Payload: snet.SCMPTracerouteRequest{
			Identifier: p.ID,
			Sequence:  uint16(atomic.AddInt32(&seq, 1)),
		},
	},
}

if err := scionConn.WriteTo(pkt, path.UnderlayNextHop()); err != nil {
  return err
}
```






</blockquote></details>

Done.
___
*[go/scion-pki/certs/renew.go, line 668 at r1](https://reviewable.io/reviews/scionproto/scion/4136#-MrWe42m9vZ0Rx7FJBaf:-MtNZodh-Mkl9ayUaZAA:b84f451) ([raw file](https://github.com/scionproto/scion/blob/e9c1ef30dd500374228be2eea55786191db527d4/go/scion-pki/certs/renew.go#L668)):*

Right. I think that belongs to buildFullAddress anyway, because we do path resolution there if the path is missing.
</blockquote></details>

Ok, I leave it there then.

oncilla (approved):
:lgtm:

> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4136#-:-MtapuXHB5Fau4b7l58V:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mawyss)

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4136)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mawyss)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4136#-MtapwAmDU-M24ZQ5AxM:-MtapwAmDU-M24ZQ5AxN:b-qe05g2) (no related file):*
@mawyss Very nice work :thumbsup: 

We try to include this in our next open sourcing run.
However, this change also implies internal changes, thus, no guarantees on that.

Anyway, thank you very much for the contribution. I'm happy that this part of the code base is finally cleaned up a bit.

---

PR #4138, opened 2021-12-23
 Push-Based Path Quality Aware Beaconing (WIP)

Nearoo:
As my bachelor thesis, I intend to implement this algorithm into the SCION codebase. I've outline the reasoning, algorithm, and intended changes to the codebase in [this design document](https://github.com/Nearoo/scion/blob/pqa-beaconing/doc/pqa-beaconing.md). **I hereby ask for comment on it's rationale and implementation details, after which I plan to start implementation.**

## Outline

During core beaconing, core ASs continuously have to decide which beacons to forward to which interfaces in order to keep the message complexity sufficiently low. Traditional algorithms have focussed on selecting paths with low hop count[2] or high diversity w.r.t. links travelled [3]. The goal of either is to provide each AS with a high quality set of paths to every other AS, such that their target application can (hopefully) select one that suits its needs.

Since service providers know which optimality criteria is required for each type of services (application) they provide, and they have customers all around world in all ISDs (and therefor core ASes), it is reaonable to let origin ASes to specify the optimality criteria of the path originating from them. 

The changes outlined in the design document, and implemented in this PR, allow core-ASs to
* specify optimality criteria which should be marked in beacons originating from specific interfaces of the AS
* propagate beacons containing such criterias s.t. they construct paths that are approximately ideal w.r.t. these criterias

## Timetable:
* Intended Start of Implementation: 29th of December
* Intended Deadline for Implementation: 14th of January

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4138)


shitz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4138)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Nearoo and @shitz)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4138#-Mrbu0xbDrUgSITZn8gO:-Mrbu0xcC6nmeGK_eDxH:bcsy1vz) (no related file):*
Hi @nearoo 

Thank you for this submission.

Some high-level comments and general remarks:
* Is there a write-up of the algorithm, how it compares to the existing algorithm, analysis of runtime, optimization guarantees etc? What's included in your design document is certainly not enough to understand these aspects. You mention a few references in the outline, but I cannot the links to any of these.
* What's your evaluation strategy? As this is a fundamental piece of the SCION architecture this needs careful evaluation of correctness, performance, optimality etc.
* What's the deployment strategy? Since there are already active SCION deployments (productive SCION network and SCIONLab) a change like this needs to be incrementally deployable. We cannot require a "flag day" where every AS will be upgraded and the new beaconing algorithm will be activated.
* Related to the above: This is a highly invasive change. The implementation needs to take care of making this an "opt-in" feature until enough data has been gathered to make it the default.
* Given your timeline and the upcoming holiday break there is almost no chance to have an informed discussion around the algorithm as a whole or the implementation details thereof. If your time constraints are so tight, I suggest you implement this in a personal fork or maybe submit an experimental patch against the scion fork used in SCIONLab: https://github.com/netsec-ethz/scion. However, I expect similar questions come up from the maintainers

I'm happy to work with you on some of this questions in parallel, but given the timeline it's not realistic that this will make it into the main SCION repository.

---

SeyedaliTaba:
Hi @shitz 

This is (a part of) the algorithm Ahad (@noorizea) and I have been discussing for a while and we decided to define a (or multiple) student project(s) to implement it. Therefore, we have already discussed its run time and theoretical aspects with Ahad. It is not yet documented but we'll be writing the paper very soon. On my side, I'll implement the algorithm in the simulator as well. If you like we can discuss it in a joint meeting with you and Ahad.

The algorithm consists of a pull-based and a push-based algorithm. In this project, @Nearoo is only implementing the push-based one. The pull-based one will need even more invasive changes to the codebase.

It would be great if we discuss the evaluation and deployment strategy. At the moment, for the evaluation, I plan to simulate the algorithms on large scale in the simulator, emulate the SCION code the students develop on a single machine, then on the SCIONLab network. For the deployment, these algorithms are optional, meaning that they can be run in parallel with the baseline algorithm, and if an AS does not support the specified optimizations in the beacons, it just ignores it.

We do not expect this code to be merged by Jan 14th at all. @Nearoo needs to implement this part and evaluate it by running on a single machine by then so that he can defend his thesis in February. In the future, either he himself or other students will continue to work on this implementation.

Bests,

Ali

Nearoo (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4138)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SeyedaliTaba and @shitz)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4138#-Mrbu0xbDrUgSITZn8gO:-Ms-u8tF0-u4fDF7_naf:b-9oc9zw) (no related file):*

Hi @nearoo 

Thank you for this submission.

Some high-level comments and general remarks:
* Is there a write-up of the algorithm, how it compares to the existing algorithm, analysis of runtime, optimization guarantees etc? What's included in your design document is certainly not enough to understand these aspects. You mention a few references in the outline, but I cannot the links to any of these.
* What's your evaluation strategy? As this is a fundamental piece of the SCION architecture this needs careful evaluation of correctness, performance, optimality etc.
* What's the deployment strategy? Since there are already active SCION deployments (productive SCION network and SCIONLab) a change like this needs to be incrementally deployable. We cannot require a "flag day" where every AS will be upgraded and the new beaconing algorithm will be activated.
* Related to the above: This is a highly invasive change. The implementation needs to take care of making this an "opt-in" feature until enough data has been gathered to make it the default.
* Given your timeline and the upcoming holiday break there is almost no chance to have an informed discussion around the algorithm as a whole or the implementation details thereof. If your time constraints are so tight, I suggest you implement this in a personal fork or maybe submit an experimental patch against the scion fork used in SCIONLab: https://github.com/netsec-ethz/scion. However, I expect similar questions come up from the maintainers

I'm happy to work with you on some of this questions in parallel, but given the timeline it's not realistic that this will make it into the main SCION repository.
</blockquote></details>

Hi @shitz 

Thank you for the quick and detailed feedback!

As @seyedalitaba he said, I don't expect the changes to be merge-ready by January 14th. My primary goal is to get the algorithm implemented and evaluated to an extend that permits writing a bachelor thesis. However, as far as discussion here within that tight timeframe permits this, ideally the work that I do is on track for a potential merge after the thesis; I'm very open to continue working on it as a semester project until the PR is merged. And, I should also say: in case things turn south, the legal deadline of the thesis is in 1st of April; the January deadline is for finishing the thesis before the next semester.

* Concerning the writeup: Apart fromt the paper that Ali intends to write, a writeup of the algorithm that includes some theoretical justification, including goals, some runtime & message complexity analysis, path optimization guarantees etc. is planned as part of the bachelor thesis. I plan to finish a first draft of this section this week, while ideally (and perhaps wishfully), the review of the implementation strategy takes place here. As Ali pointed out, the idea of the algorithm itself should already be ratified. I was told I should focus on details strictly relevant for the implementation here.
* Concerning deployment: I'll extend the design document to address these aspects. Depending on how ASs handle received beacons with extensions they don't recognize, this might be fairly straight forward for propagation. For origination, the extension is intended as _supplementary_ to existing beacons; already in the design document presented, interfaces that aren't explicitly configured to originate the new beacons (i.e. don't "opt-in") fall back to regular beacons. (see [here](https://github.com/Nearoo/scion/blob/pqa-beaconing/doc/pqa-beaconing.md#config-file-specification)). I'm not sure that would address your 4th point as well?
* Concerning evaluation: That is also intended to part of the bachelor thesis. I'll add considerations relevant for the implementation to the design document asap and get back to you.

---

matzf (commented):
> Reviewed 1 of 19 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4138)* status: 1 of 19 files reviewed, 6 unresolved discussions (waiting on @matzf, @Nearoo, @SeyedaliTaba, and @shitz)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4138#-Ms_39K6-iBhnSakMmtS:-Ms_39K77XRBby2hKHPC:b-b2s5fv) (no related file):*
Thank you for the detailed write up of this proposal!

After thinking this through, I'm somewhat skeptical of this premise:

> Since service providers know which optimality criteria is required for each type of services (application) they provide, and they have customers all around world in all ISDs (and therefor core ASes), it is reaonable to let origin ASes to specify the optimality criteria of the path originating from them.

The core-ASes under consideration here provide transit, they will not generally be aware of or concerned with applications offered by their (direct and indirect) customer ASes. Furthermore, I'd expect that for most core ASes every one optimisation target would be desirable for _some_ customers in their "customer cone". 
Consequently, it seems likely that the outcome is that core ASes, if they enable this proposed system, would enable _all_ the supported optimization targets.

This suggests the following alternative approach; instead of this system that lets the beacon originator choose an optimisation target for each beacon, we describe quality aware beaconing as a recommended "best practice" beacon selection method for core-beaconing. The recommendation is simply to always include some beacons for each supported optimisation target (i.e. a few with lowest latency and a few with highest bandwidth) to each destination AS.
To me it seems that the result of such a recommendation would be effectively the same as the result of this more complex system proposed here.
The implementation of this would only require to implement these beacon selection functions, and no further beacon extensions would be necessary. It would be easier to integrate into the current system and it would fall back to the current beaconing scheme much more obviously and gracefully.

---
___
*[doc/pqa-beaconing.md, line 78 at r5](https://reviewable.io/reviews/scionproto/scion/4138#-MsZPCeGCoAd2mUfBrLE:-MsZPCeGCoAd2mUfBrLF:bxoglv5) ([raw file](https://github.com/scionproto/scion/blob/1654625484adaf286e2122b5db882cda17d5d986/doc/pqa-beaconing.md#L78)):*
> ```Markdown
> ### Direction of optimization
> 
> We allows to be optimized for packets travelling in the same direction as the beacon ("**forward**"), in opposite direction ("**backwards**"), and in both directions on the same path if the metrics are approximately equal ("**symmetric**"). 
> ```

As discussed in the [documentation for the StaticInfoExtension](https://scion.docs.anapaya.net/en/latest/beacon-metadata.html#symmetry), the extension in the current form cannot represent asymmetric information. Is there evidence suggesting that this would likely make any difference in practice? If this is purely for the sake of completeness, I would propose to reduce the scope of the project here and assume symmetry to remove this forward/backward distinction.
___
*[doc/pqa-beaconing.md, line 94 at r5](https://reviewable.io/reviews/scionproto/scion/4138#-Ms_MGfF0yRQeB2rqGiY:-Ms_MGfF0yRQeB2rqGiZ:b-5yjcth) ([raw file](https://github.com/scionproto/scion/blob/1654625484adaf286e2122b5db882cda17d5d986/doc/pqa-beaconing.md#L94)):*
> ```Markdown
> 
> Multiple interfaces can originate the same optimization target; a group of interfaces originating the same target consitutes an `optimization interface group` for that target. `optimization interface groups` are created implicitly by assigning multiple interfaces to the same `optimization target`. The term is nevertheless useful to decide which interfaces to assign to which optimization target. 
> ```


I really like the idea to enable optimized beacon selection on a finer granularity than just the destination AS. However, I would like to suggest an alternative to this explicit interface grouping.
What we want to achieve is to provide optimal options for combinations of this core path segment with any down path segments. Luckily, the StaticInfoExtension already includes the information for inter-AS hops (*). When selecting the best beacon candidates, we can consider the quality metrics to each down-interface in the destination AS and choose the beacons such that we have an optimal option for all (or at least for many) down interfaces.
With this approach, we can potentially reduce the number of required beacons, as some of the interface groups, although justified because e.g. the latencies are very different, are not useful in path segment combinations. Also, this would avoid this additional tweaking option.

(*) Currently, the information for the hop between down interface and core interface is only contained in the down-segments. This can easily be adapted so that this is included also when originating on a core interface. 
___
*[doc/pqa-beaconing.md, line 99 at r5](https://reviewable.io/reviews/scionproto/scion/4138#-Ms_PPXXAptW3FIichsR:-Ms_PPXYBKkodPrnTFr-:b-d8xl3g) ([raw file](https://github.com/scionproto/scion/blob/1654625484adaf286e2122b5db882cda17d5d986/doc/pqa-beaconing.md#L99)):*
> ```Markdown
> In non-origin (propagator) ASes, _interface_groups_ for each optimality criteria can be defined as the group of interfaces from which the intra-AS that criteria has almost the same value to any destination inside that AS (including other interfaces). For example, for latency grouping can be done based on proximity, but for bandwidth all interfaces with the same bandwidth can grouped with each other. Therefore, finding an optimal path from/to an ingress/egress interface in a group (for each optimality criteria) gives almost the same quality when we find an optimal path from/to all of the interfaces in that group. For traffic engineering puposes, ASes can violate this definition, which we do not conside now.
> 
> > In some cases neighbor ASes can put the interfaces connected to the same link between them in different groups. In this case thay can communicate their interface groupings, which we do not consider in this project.
> ```

This discussion of non-origin ASes is a bit confusing; first, the reason for interface grouping in the "origin ASes" is exactly the same as what is described only here, in the "non-origin ASes" case. Second, is there anything that we actually propose to _do_ for this case of "non-origin ASes"?
___
*[doc/pqa-beaconing.md, line 224 at r5](https://reviewable.io/reviews/scionproto/scion/4138#-Ms_-h_I1ZGxmiEC-0cp:-Ms_-h_I1ZGxmiEC-0cq:b-xvtbjd) ([raw file](https://github.com/scionproto/scion/blob/1654625484adaf286e2122b5db882cda17d5d986/doc/pqa-beaconing.md#L224)):*
> ```Markdown
>     * optimality: `min` or `max`
>     * `symmetry_tolerance`: tolerance to consider two metrics "the same" when optimizing for symmetric, as a fraction (`a == b <=> [a*(1-tol), a*(1+tol)] ∩ [b*(1-tol), b*(1+tol)] != ∅`)
>     * `proto_id`: The string representation of the field value of `OptimizationQuality` in the protobuf segment extension
> ```

There is no convincing reason to allow configuration the interpretation of these "qualities". For each of the data items included in the StaticInfoExtension, there is only one sensible interpretation for combination functions and optimization direction and it can be hard-coded. The set of qualities that make sense to be supported is quite limited and does not need to be extensible by configuration.

SeyedaliTaba:
*[a discussion](https://reviewable.io/reviews/scionproto/scion/4138#-Ms_39K6-iBhnSakMmtS:-Msa8oNJBb3Hv-VMggTS:bgr1xy6) (no related file):*

Thank you for the detailed write up of this proposal!

After thinking this through, I'm somewhat skeptical of this premise:

> Since service providers know which optimality criteria is required for each type of services (application) they provide, and they have customers all around world in all ISDs (and therefor core ASes), it is reaonable to let origin ASes to specify the optimality criteria of the path originating from them.

The core-ASes under consideration here provide transit, they will not generally be aware of or concerned with applications offered by their (direct and indirect) customer ASes. Furthermore, I'd expect that for most core ASes every one optimisation target would be desirable for _some_ customers in their "customer cone". 
Consequently, it seems likely that the outcome is that core ASes, if they enable this proposed system, would enable _all_ the supported optimization targets.

This suggests the following alternative approach; instead of this system that lets the beacon originator choose an optimisation target for each beacon, we describe quality aware beaconing as a recommended "best practice" beacon selection method for core-beaconing. The recommendation is simply to always include some beacons for each supported optimisation target (i.e. a few with lowest latency and a few with highest bandwidth) to each destination AS.
To me it seems that the result of such a recommendation would be effectively the same as the result of this more complex system proposed here.
The implementation of this would only require to implement these beacon selection functions, and no further beacon extensions would be necessary. It would be easier to integrate into the current system and it would fall back to the current beaconing scheme much more obviously and gracefully.
</blockquote></details>

Hi Matthias,

Thanks a lot for your feedback and suggestions. 

As it is a research project and Silas’s thesis would be used as a proof of concept, he will implement this proposal even if it does not get merged. 


Today, I had a discussion with Sam and Ahad from Anapaya about the next steps in implementing the new beaconing algorithms. We will of course discuss all aspects thoroughly before implementation.

But regarding your comments:

Of course, if core ASes only provide transit, this proposal does not help too much. However, considering the ongoing direction of SCION deployment (e.g. Sunrise lets their customers have their own ISD), the proposal helps a lot, specially if the number of desired path properties increase in the future. 

Of course each small company having its own ISD sounds crazy, but big companies and cloud providers constituting their own ISD is not unreasonable.

So, consider the case that for example Microsoft has its own core AS, and for different types of applications it provides, it needs different types of path. This specially helps when they need different types of paths at different locations. So, for example they can request low latency paths to some of their interfaces and high bandwidth paths to some other interfaces. Also, they can use this feature for traffic engineering.

---

matzf:
This has become stale by now. It also seems that further discussions on the idea would be necessary before continuing to finalize an implementation. 
Closing this for now. Feel free to re-open if you want to reactivate the work on this topic. Perhaps it would make sense to create a proposal as a github issue though, before continuing to work on the implementation.

PR #4142, opened 2022-01-19
router/bfd: reduce allocations, prometheus metrics

matzf:
Use prometheus metrics directly, bypassing the metrics library.
This avoids a large amount of allocations caused by usage patterns in
the metrics library. While this does not help making progress towards a
consistent way to handle metrics, it also does not appear to make
matters worse.

Avoid various other allocations in the BFD session and sender code,
by reusing the relevant objects and avoiding temporaries to escape to
the heap.
As the BFD messages are comparably infrequent, the allocations triggered
here were usually not critical. However, for setups with a large number
of interfaces, the wasteful allocations were indeed somewhat significant
compared to the allocations from the forwarding part of the router
dataplane (which are much harder to optimize away).

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4142)

### DIFF
--- a/go/pkg/router/BUILD.bazel
+++ b/go/pkg/router/BUILD.bazel
@@ -15,7 +15,6 @@ go_library(
         "//go/lib/common:go_default_library",
         "//go/lib/epic:go_default_library",
         "//go/lib/log:go_default_library",
-        "//go/lib/metrics:go_default_library",
         "//go/lib/scrypto:go_default_library",
         "//go/lib/serrors:go_default_library",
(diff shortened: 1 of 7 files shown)


oncilla (approved):
> Reviewed 3 of 7 files at r1, 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4142)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4146, opened 2022-01-25
showpaths: handle empty path

matzf:
Previously, showpaths would show confusing results when the destination
is the local AS; it did list an empty path, but also showed the
error "no path alive" (because the empty path cannot be probed).
With the `-e` flag, it would also show a spurious expiration time.

If the destination is the local AS, this is now specifically handled
with a message explaining that this results in an empty path.

For the json output, the `paths` field is now an empty list in this case
and so omitted.
Additionally, a `local_isd_as` (following the naming pattern from the
existing `local_ip` field) is added to the json output, to allow
consumers to detect and handle this situation.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4146)

### DIFF
--- a/go/pkg/showpaths/showpaths.go
+++ b/go/pkg/showpaths/showpaths.go
@@ -312,6 +312,9 @@ func Run(ctx context.Context, dst addr.IA, cfg Config) (*Result, error) {
 	if err != nil {
 		return nil, serrors.WrapStr("determining local ISD-AS", err)
 	}
+	if dst == localIA {
+		return nil, serrors.New("destination is the local AS", "dst", dst)
+	}
 
 	// TODO(lukedirtwalker): Replace this with snet.Router once we have the
 	// possibility to have the same functionality, i.e. refresh, fetch all paths.
--- a/go/pkg/showpaths/showpaths.go
+++ b/go/pkg/showpaths/showpaths.go
@@ -36,8 +36,9 @@ import (
 
 // Result contains all the discovered paths.
 type Result struct {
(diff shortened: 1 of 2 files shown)


lukedirtwalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4146)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker and @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4146#-MuPh0rNBVimorVCYrrV:-MuPh0rNBVimorVCYrrW:b-xwvj4s) (no related file):*
Instead of erroring out we could just return a fixed value, ignore all flags etc and also make sure that 

I don't think it is really an error if you showpaths to the local IA. I think there should just be a fixed output that informs the user about that this is an empty path towards the local IA.

---

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4146)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4146#-MuPh0rNBVimorVCYrrV:-MuV36DSBHFS6ks1Nqv2:b-jc0bsh) (no related file):*

Instead of erroring out we could just return a fixed value, ignore all flags etc and also make sure that 

I don't think it is really an error if you showpaths to the local IA. I think there should just be a fixed output that informs the user about that this is an empty path towards the local IA.
</blockquote></details>

All clear, done.

---

PR #4147, opened 2022-01-25
slayers: recycle path when decoding

matzf:
In `slayers.SCION.DecodeFromBytes`, we allocate a new path for the
correct path type on every invocation. This negates the advantages of
recycling the layer object.

Introduce a "recycling bin" for paths that can be reused in the SCION
layer's DecodeFromBytes.
This is only used when explicitly enabled, as it only makes sense if the
layer is really reused for parsing. This opt-in also avoids issues with
tests that deep-compare the SCION layer with an expected value where
this recycling bin is empty..

----

I'm not particularly fond of this approach, but I don't know a really good alternative either.

A somewhat simpler approach (implemented in the first commit in this PR) that does not need additional fields, would be to
only recycle the last path as long as the type matches. This would reduce allocations in typical situations, but has an obvious performance degradation if the path type switches back and forth (path type is attacker controlled, so possibly exploitable).

One variation on the current proposal could be to add a method like `slayers/SCION.SetPathsForDecoding(paths ...Path)` instead of the current `RecylePaths`. The caller would allocate an instance for each supported path type on startup. This would feel vaguely similar to passing layers to  the `gopacket.DecodingLayerParser`, except that it's passed to a layer.
I don't find this strictly better or worse than the current proposal, so just mentioning this as a possible alternative.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4147)

### DIFF
--- a/go/lib/slayers/scion.go
+++ b/go/lib/slayers/scion.go
@@ -146,6 +146,9 @@ type SCION struct {
 
 	// Path is the path contained in the SCION header. It depends on the PathType field.
 	Path path.Path
+
+	// pathPool is a recycling bin for paths created in DecodeFromBytes
+	pathPool map[path.Type]path.Path
 }
 
 func (s *SCION) LayerType() gopacket.LayerType {
@@ -246,11 +249,10 @@ func (s *SCION) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
 	}
 
 	if s.Path == nil || s.Path.Type() != s.PathType {
-		path, err := path.NewPath(s.PathType)
+		s.Path, err = s.getPath(s.PathType)
 		if err != nil {
 			return err
 		}
-		s.Path = path
 	}
 
 	err = s.Path.DecodeFromBytes(data[offset : offset+pathLen])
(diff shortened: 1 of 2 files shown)


oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4147)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker and @matzf)

___
*[go/lib/slayers/scion.go, line 251 at r2](https://reviewable.io/reviews/scionproto/scion/4147#-MxxFGP376zHr5l2X0IZ:-MxxFGP376zHr5l2X0I_:b-kcz73u) ([raw file](https://github.com/scionproto/scion/blob/2decf82969d1de15f167124e89dfa63c4bcb9d32/go/lib/slayers/scion.go#L251)):*
> ```Go
> 	}
> 
> 	if s.Path == nil || s.Path.Type() != s.PathType {
> ```

I think we want to do `s.getPaths` unconditionally.

The way it is now. We have a mix between recycling paths and not recycling paths.
Which is unexpected after reading the `RecyclePaths` doc comment.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4147)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[go/lib/slayers/scion.go, line 251 at r2](https://reviewable.io/reviews/scionproto/scion/4147#-MxxFGP376zHr5l2X0IZ:-My6Lyfx-iCt3FVHTriZ:b-jxutdf) ([raw file](https://github.com/scionproto/scion/blob/2decf82969d1de15f167124e89dfa63c4bcb9d32/go/lib/slayers/scion.go#L251)):*

I think we want to do `s.getPaths` unconditionally.

The way it is now. We have a mix between recycling paths and not recycling paths.
Which is unexpected after reading the `RecyclePaths` doc comment.


</blockquote></details>

True. I see that in the version that you merged, you've already changed this, as well as the representation of the pool. Thank you!

oncilla:
@matzf One drawback of the merged solution is that we do not handled dynamically registered path types.
But given that this is intended for the router only, I think we can live with it for now.

PR #4157, opened 2022-03-02
router: check src/dst IA, check transit source IP

matzf:
Add egress filtering on the source IA to mitigate accidental or
malicious source address spoofing by untrusted end-hosts.
Compring the source IA against the configured local IA for egress
packets is fast and reliable.

A malicious end host could bypass this source IA check relatively
easily; by prepending fake hop-fields, it can masquerade it's outgoing
packets as transit packets, for which the check is not applied.
To mitigate this form of spoofing attack, we add a check to for the
source IP address of transit traffic -- if this source IP doesn't match
the IP address of the corresponding sibling border router, the packet is
silently dropped.
Under reasonable the assumption that the underlying (IP-)network
infrastructure prevents untrusted end-hosts from spoofing IP addresses
of the SCION infrastructure hosts, this check prevents source IA
spoofing.

As a natural extension of the source IA checks for outgoing traffic,
we also add checks for source/destination IA for transit and incoming packets;
- for transit traffic:
  neither source nor destination IA must equal local IA
- for incoming packets:
  source IA must NOT equal local IA
  destination IA must equal local IA
- for incoming packets:
  source IA must equal local IA
  destination IA must NOT equal local IA

If any check fails, we reply with the appropriate SCMP Error
InvalidSourceAddress/InvalidDestinationAddress.
Observe that in case of InvalidSourceAddress, the correct destination IA
for the SCMP Error message is unknown. It's thus possible that the error
message will be dropped because of an invalid destination IA on the
way to the back to the originator.

We have all information needed to extend these source/destination IA
checks to packets that originate from or are destined to direct neighbors.
However, this would not provide additional tamper-resistance; if the
neighbor AS is benign, it will perform filtering on its own. In case the
neighbor AS is malicious, it could trivially bypass such a filter by
prepending fake hop fields.

Add router acceptance test cases for these new checks.
Multiple existing test cases needed to be adapted because the input
packet accidentally failed some of the new checks.


FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @FR4NK-W and @matzf)

___
*[router/dataplane.go, line 1265 at r1](https://reviewable.io/reviews/scionproto/scion/4157#-MyporXJAiIelGHDqHIQ:-MyporXJAiIelGHDqHIR:blcko6) ([raw file](https://github.com/scionproto/scion/blob/61437fed27ef50c646a263417229a6b5021277ff/router/dataplane.go#L1265)):*
> ```Go
> 
> 	if p.path.IsXover() {
> 		if r, err := p.doXover(); err != nil {
> ```

Should the `validateHopExpiry` and `verifyCurrentMAC` checks done after a Xover info and HF update also be moved here with the other validations?



_Code quote:_
```Go
p.doXover()
```

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @FR4NK-W)

___
*[router/dataplane.go, line 1265 at r1](https://reviewable.io/reviews/scionproto/scion/4157#-MyporXJAiIelGHDqHIQ:-N-8-XS9948b7KWOluTb:b2zly1q) ([raw file](https://github.com/scionproto/scion/blob/61437fed27ef50c646a263417229a6b5021277ff/router/dataplane.go#L1265)):*

Should the `validateHopExpiry` and `verifyCurrentMAC` checks done after a Xover info and HF update also be moved here with the other validations?
</blockquote></details>

Good point, thanks! Done.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: 0 of 10 files reviewed, 3 unresolved discussions (waiting on @lukedirtwalker and @matzf)

___
*[`router/dataplane.go` line 110 at r4](https://reviewable.io/reviews/scionproto/scion/4157#-NDgBrHdBtwmeko1c72s:-NDgBrHdBtwmeko1c72t:undefined) ([raw file](https://github.com/scionproto/scion/blob/79eaf2bc577e6bb4533e86789e3f926d2e41e6b3/router/dataplane.go#L110)):*
> ```Go
> var (
> 	alreadySet                    = serrors.New("already set")
> 	invalidSrcIA                  = serrors.New("invalid source IA")
> ```





_Suggestion:_
```Go
ISD-AS
```
___
*[`router/dataplane.go` line 111 at r4](https://reviewable.io/reviews/scionproto/scion/4157#-NDgBtKS5NL54EMrDFJ-:-NDgBtKS5NL54EMrDFJ0:undefined) ([raw file](https://github.com/scionproto/scion/blob/79eaf2bc577e6bb4533e86789e3f926d2e41e6b3/router/dataplane.go#L111)):*
> ```Go
> 	alreadySet                    = serrors.New("already set")
> 	invalidSrcIA                  = serrors.New("invalid source IA")
> 	invalidDstIA                  = serrors.New("invalid destination IA")
> ```





_Suggestion:_
```Go
ISD-AS
```
___
*[`router/dataplane.go` line 915 at r4](https://reviewable.io/reviews/scionproto/scion/4157#-NDgC-MgDfJJZRzh8thv:-NDgC-MgDfJJZRzh8thw:b7k0mbv) ([raw file](https://github.com/scionproto/scion/blob/79eaf2bc577e6bb4533e86789e3f926d2e41e6b3/router/dataplane.go#L915)):*
> ```Go
> 		pktIngressID := p.ingressInterface()
> 		expectedSrc, ok := p.d.internalNextHops[pktIngressID]
> 		if !ok || !expectedSrc.IP.Equal(p.srcAddr.IP) || expectedSrc.Port != p.srcAddr.Port {
> ```

I think you should only check the IP address, but not the port here.

Checking the port prohibits source port randomization that could be used for load balancing (e.g., in LAG or  ECMP)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: 0 of 10 files reviewed, 3 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`router/dataplane.go` line 110 at r4](https://reviewable.io/reviews/scionproto/scion/4157#-NDgBrHdBtwmeko1c72s:-NDlhDbMDBf3TGKNaZ_h:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/79eaf2bc577e6bb4533e86789e3f926d2e41e6b3/router/dataplane.go#L110)):*
> ```Go
> var (
> 	alreadySet                    = serrors.New("already set")
> 	invalidSrcIA                  = serrors.New("invalid source IA")
> ```

Done.
___
*[`router/dataplane.go` line 111 at r4](https://reviewable.io/reviews/scionproto/scion/4157#-NDgBtKS5NL54EMrDFJ-:-NDlhE-I7p5x57CSm5B_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/79eaf2bc577e6bb4533e86789e3f926d2e41e6b3/router/dataplane.go#L111)):*
> ```Go
> 	alreadySet                    = serrors.New("already set")
> 	invalidSrcIA                  = serrors.New("invalid source IA")
> 	invalidDstIA                  = serrors.New("invalid destination IA")
> ```

Done.
___
*[`router/dataplane.go` line 915 at r4](https://reviewable.io/reviews/scionproto/scion/4157#-NDgC-MgDfJJZRzh8thv:-NDlhEg94uwJZ52Pnz2g:b-88crd6) ([raw file](https://github.com/scionproto/scion/blob/79eaf2bc577e6bb4533e86789e3f926d2e41e6b3/router/dataplane.go#L915)):*

I think you should only check the IP address, but not the port here.

Checking the port prohibits source port randomization that could be used for load balancing (e.g., in LAG or  ECMP)


</blockquote></details>

Done, ok for me.

Just for the record, the behavior is different for the intra-AS links, where this router implementation uses a dialed UDP connection and consequently (implicitly) checks the source port as well.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4157#-NGXHwI95QVqkNv07lyd:-NGXHwIA83YicKvhc8Yq:b-eigam8) (no related file):*
Ping @oncilla, is this ok for you now?

___

lukedirtwalker (changes requested):
> Reviewed 2 of 10 files at r1, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: 3 of 10 files reviewed, 6 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`pkg/slayers/path/scion/base.go` line 99 at r6](https://reviewable.io/reviews/scionproto/scion/4157#-NHxZVnv6wzsyt4eeDbH:-NHxZVnw0jbn19R0v40j:b-47hyz0) ([raw file](https://github.com/scionproto/scion/blob/e110b35a13346465e8c4dbde36bb1cffff2eb4a5/pkg/slayers/path/scion/base.go#L99)):*
> ```Go
> 
> func (s *Base) infIndexForHF(hf uint8) uint8 {
> 	if s.NumINF >= 2 {
> ```

I think we could just make a switch out of this:
```go
switch {
case s.NumINF >= 1 && hf < s.PathMeta.SegLen[0]:
     return 0
case s.NumINF >= 2 && hf < s.PathMeta.SegLen[0]+s.PathMeta.SegLen[1]:
     return 1
default:
    return uint8(s.NumINF - 1)
}
```
___
*[`router/dataplane.go` line 912 at r6](https://reviewable.io/reviews/scionproto/scion/4157#-NHxaide-Id2nhL93Hb8:-NHxaide-Id2nhL93Hb9:bnxkgxb) ([raw file](https://github.com/scionproto/scion/blob/e110b35a13346465e8c4dbde36bb1cffff2eb4a5/router/dataplane.go#L912)):*
> ```Go
> // SrcIA checks by disguising packets as transit traffic.
> func (p *scionPacketProcessor) validateTransitUnderlaySrc() (processResult, error) {
> 	if p.ingressID == 0 && !p.path.IsFirstHop() {
> ```

you could use an early return here, for less indentation in the remaining code.
```go
if p.path.IsFirstHop() || p.ingressID != 0 {
    return processResult{},nil
}
```

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: 1 of 11 files reviewed, 6 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/slayers/path/scion/base.go` line 99 at r6](https://reviewable.io/reviews/scionproto/scion/4157#-NHxZVnv6wzsyt4eeDbH:-NHxhomnCt8MWmAzZLHI:bx1yty6) ([raw file](https://github.com/scionproto/scion/blob/e110b35a13346465e8c4dbde36bb1cffff2eb4a5/pkg/slayers/path/scion/base.go#L99)):*

I think we could just make a switch out of this:
```go
switch {
case s.NumINF >= 1 && hf < s.PathMeta.SegLen[0]:
     return 0
case s.NumINF >= 2 && hf < s.PathMeta.SegLen[0]+s.PathMeta.SegLen[1]:
     return 1
default:
    return uint8(s.NumINF - 1)
}
```
</blockquote></details>

Good idea, thanks! 
I opted to completely omit the `NumINF` checks and rely entirely on `SegLen`. This changes the behavior for out of range `hf` values (returns 2 instead of NumINF-1), so that the range check needs to be done at the caller (only for `IsXover`, other callers already check sufficiently). IMO this makes quite a bit easier to understand, also for `IsXover`.
___
*[`router/dataplane.go` line 912 at r6](https://reviewable.io/reviews/scionproto/scion/4157#-NHxaide-Id2nhL93Hb8:-NHxoZzOEyaQcFEAT31e:b-8tt3bs) ([raw file](https://github.com/scionproto/scion/blob/e110b35a13346465e8c4dbde36bb1cffff2eb4a5/router/dataplane.go#L912)):*

you could use an early return here, for less indentation in the remaining code.
```go
if p.path.IsFirstHop() || p.ingressID != 0 {
    return processResult{},nil
}
```

</blockquote></details>

Good idea, thanks. Done.

lukedirtwalker (changes requested):
> Reviewed 2 of 2 files at r7, 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: 3 of 11 files reviewed, 5 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`pkg/slayers/path/scion/base.go` line 99 at r6](https://reviewable.io/reviews/scionproto/scion/4157#-NHxZVnv6wzsyt4eeDbH:-NHxqCRQ2WlP4EoqTn7x:b-wnsfd1) ([raw file](https://github.com/scionproto/scion/blob/e110b35a13346465e8c4dbde36bb1cffff2eb4a5/pkg/slayers/path/scion/base.go#L99)):*

Good idea, thanks! 
I opted to completely omit the `NumINF` checks and rely entirely on `SegLen`. This changes the behavior for out of range `hf` values (returns 2 instead of NumINF-1), so that the range check needs to be done at the caller (only for `IsXover`, other callers already check sufficiently). IMO this makes quite a bit easier to understand, also for `IsXover`.
</blockquote></details>

couldn't you run into panics if the segments is invalid? i.e. accessing the SegLen out of bounds?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: 3 of 11 files reviewed, 5 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/slayers/path/scion/base.go` line 99 at r6](https://reviewable.io/reviews/scionproto/scion/4157#-NHxZVnv6wzsyt4eeDbH:-NHxsKjO0Dge9GPUmUgq:bljdz17) ([raw file](https://github.com/scionproto/scion/blob/e110b35a13346465e8c4dbde36bb1cffff2eb4a5/pkg/slayers/path/scion/base.go#L99)):*

couldn't you run into panics if the segments is invalid? i.e. accessing the SegLen out of bounds?
</blockquote></details>

No, I think not; `SegLen` is a fixed size array `[3]uint8`.

lukedirtwalker (commented):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: 4 of 11 files reviewed, 4 unresolved discussions (waiting on @oncilla)

oncilla (approved):
> Reviewed 6 of 10 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, 1 of 2 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: all files reviewed, 1 unresolved discussion (waiting on @FR4NK-W and @lukedirtwalker)

lukedirtwalker (approved):
> Reviewed 5 of 10 files at r1, 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4157)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

Issue #4149, opened 2022-01-26
Refactor addr.HostAddr types


matzf:
The `addr.HostAddr` type hierarchy is currently organized around this base interface:
```go
type HostAddr interface {
	Size() int
	Type() HostAddrType
	Pack() []byte
	IP() net.IP
	Copy() HostAddr
	Equal(HostAddr) bool
	fmt.Stringer
}
```
This interface is implemented by the types `HostNone`, `HostIPv4`, `HostIPv6` and `HostSVC`. 
The `Pack` and `Size` methods are not really used, as `slayers` uses separate serialization logic, based on separate types. The same is true for the `HostFromRaw` parsing function.

The `slayers` library does not generally use these `HostAddr` types. Instead, it relies on `net.Addr` with explicit type assertions to `net.IPAddr` and `addr.HostSVC`. The `net.Addr` interface here is merely a convention, none of the interface methods are actually used.

This situation is not ideal:
- creating `HostAddr` always allocates -- this cannot be fixed with the interface type.
- the `HostIPv4`/`HostIPv6` split is unnecessary as the `IP` is already aware of the type
- `HostAddr` is not a value type, cannot be used as map key etc.
- different "world" of address types in `slayers` and everywhere else
- arbitrarily pretends that `HostSVC` is a `net.Addr`

---

#### Proposal
Replace the `HostAddr` type hierarchy with a single value type `Host`.
This type is opaque, allowing to internally change or extend the representation.

```go
type Host struct {
	ip  netaddr.IP
	svc SVC         // could squeeze this into ip, is it worth it?
	t   HostAddrType
}

func (h Host) Type() HostAddrType {
	return h.t
}

func (h Host) IP() netaddr.IP {
	if h.t != HostTypeIP {
		panic("IP called on non-IP address")
	}
	return h.ip
}

func (h Host) SVC() SVC {
	if h.t != HostTypeSVC {
		panic("SVC called on non-SVC address")
	}
	return h.svc
}

func (h Host) String() string { ... }

// HostIP creates a Host address of type IP
func HostIP(ip netaddr.IP) Host {
	return Host{t: HostTypeIP, ip: ip}
}

// HostIPFromStd is a hopefully temporary conversion function to create a Host of type IP from net.IP.
// Panics if ip does not have length 4 or 16.
// Unmaps IPv6-mapped IPv4 addresses.
func HostIPFromStd(ip net.IP) Host {
	a, ok := netaddr.FromStdIP(ip)
	if !ok {
[...]

PR #4160, opened 2022-03-09
slayers: combine AddrType/Len

matzf:
AddrType/Len fields are always set and read together. The AddrType field
is meaningless without the Length.

This change combines these types/fields into one value that uniquely
identifies the address type, simplifying the processing of this address
length/type.
The length (in bytes) of the address type can be obtained with a method
call. There is (intentionally) no conversion functionality to extract
only the type (as this is meaningless on its own), or to construct a
AddrTypeLen from type/length parts.

This change prepares for possible further refactoring of the interface
for setting/getting src/dst addresses on slayers.SCION, as discussed
in #4149.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4160)


shitz (changes requested):
> Reviewed 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4160)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

___
*[`pkg/slayers/scion.go` line 55 at r2](https://reviewable.io/reviews/scionproto/scion/4160#-NFhAQ5FANQRCr4qc_7i:-NFhAQ5FANQRCr4qc_7j:b-qbgef9) ([raw file](https://github.com/scionproto/scion/blob/f7eb2cc5d82fb783edbcfb04a50e5ebf903f8a71/pkg/slayers/scion.go#L55)):*
> ```Go
> // The four possible lengths are 4B (0), 8B (1), 12B (2), or 16B (3) bytes.
> // There are four possible types per address length.
> type AddrTypeLen uint8
> ```

I would call this `AddrType` still, since it effectively defines a type of address which consists of a length and a _subtype_.
WDYT?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4160)* status: 8 of 11 files reviewed, 1 unresolved discussion (waiting on @oncilla and @shitz)

___
*[`pkg/slayers/scion.go` line 55 at r2](https://reviewable.io/reviews/scionproto/scion/4160#-NFhAQ5FANQRCr4qc_7i:-NFhR9Hn5J52_-uutP2P:b-jo48hh) ([raw file](https://github.com/scionproto/scion/blob/f7eb2cc5d82fb783edbcfb04a50e5ebf903f8a71/pkg/slayers/scion.go#L55)):*

I would call this `AddrType` still, since it effectively defines a type of address which consists of a length and a _subtype_.
WDYT?
</blockquote></details>

Ah yes, that's better indeed. Done.

shitz (approved):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4160)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

PR #4168, opened 2022-03-16
tools: await-connectivity to fix CI flakes

matzf:
It has previously been observed that the integration tests are flaky
in the supervisord-based setup unless we keep the specific startup
sequence that starts dispatcher and routers before all other services
(see commit b94c06378c9d5f5571bcbfebe7d992a20cccaf07, or also #4164).
If the integration tests start before the control services have
established segments for full connectivity, i.e. offer a least one path
to every destination, the tests will fail.
When control services start before routers, it takes longer to reach
full connectivity. The first beacon origination can happen up to 5
seconds later, as the first attempt may result in a failed service
resolution attempt that needs to time out.
The CI pipeline would sleep for a fixed duration of 10 seconds which
is just enough to reach connectivity with the ideal startup sequence,
but if the processes randomly start in a bad order, it can fail.

In the docker-based topology setup this issue does not appear, simply
because the process startup (in particular for the servers of the e2e
tests) is so much slower that there is ample time to reach connectivity.

This change replaces the fixed sleep duration with a script, based on
the control service's segments API, that explicitly waits for full
connectivity, with a timeout.
Remove the hand-optimized startup sequence for supervisor tasks from
scion.sh -- a randomly bad startup sequence can and should be tolerated.

Also, more cleanup for scion.sh: don't set up ipv6 local addresses when
running docker setup. Always perform the corresponding cleanup.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4168)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -85,8 +85,8 @@ steps:
       - make
       - echo "--- start topology"
       - ./scion.sh topology -c topology/default.topo
-      - ./scion.sh run && sleep 10
-      - echo "--- run tests"
+      - ./scion.sh run
+      - tools/await-connectivity
       - ./bin/scion_integration || ( echo "^^^ +++" && false )
       - ./bin/end2end_integration || ( echo "^^^ +++" && false )
     plugins:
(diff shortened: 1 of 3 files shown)


oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4168#-:-MyNa3pR96i9LcxJPmXc:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4168#-MyNa1pPE8GKPCwTxcrv:-MyNa1pQ1KAXwO4g4rYB:b-c0hj0l) (no related file):*
Cool stuff :100: 

---

PR #4176, opened 2022-03-21
scion-pki: add tests for cert validate command

FR4NK-W:
Add unit tests for the `scion-pki cert validate` command.

Furthermore, move the the command definition to its own file.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)


matzf (changes requested):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @FR4NK-W and @matzf)

___
*[scion-pki/certs/validate.go, line 59 at r1](https://reviewable.io/reviews/scionproto/scion/4176#-Mykln178erX9ftdpHDi:-Mykln18CUvR9jAb0lAy:bxsjhq0) ([raw file](https://github.com/scionproto/scion/blob/78fe876464b25212db9670cdb83abf2c76b19a6e/scion-pki/certs/validate.go#L59)):*
> ```Go
> 
> If a single certificate is validated, the subject identified by the certificate can be validated by
> specifying the --subject flag and the expected subject identifier (distinguished name IA).
> ```

Wouldn't it be much more natural to check the `subject` for the leaf certificate?
___
*[scion-pki/certs/validate.go, line 176 at r1](https://reviewable.io/reviews/scionproto/scion/4176#-MykoWCqBYZCzELC1Fq3:-MykoWCqBYZCzELC1Fq4:b45h) ([raw file](https://github.com/scionproto/scion/blob/78fe876464b25212db9670cdb83abf2c76b19a6e/scion-pki/certs/validate.go#L176)):*
> ```Go
> }
> 
> func validateSubject(certs []*x509.Certificate, subjectIdentifier string) error {
> ```





_Suggestion:_
```Go
expectedSubject
```
___
*[scion-pki/certs/validate.go, line 182 at r1](https://reviewable.io/reviews/scionproto/scion/4176#-Mykq_wT1WKOVoxUqpTG:-Mykq_wT1WKOVoxUqpTH:b-t52o2w) ([raw file](https://github.com/scionproto/scion/blob/78fe876464b25212db9670cdb83abf2c76b19a6e/scion-pki/certs/validate.go#L182)):*
> ```Go
> 	// Check certificate Distinguished Name
> 	certSubjectASID := ""
> 	for _, dn := range certs[0].Subject.Names {
> ```

Just `subject, err := cppki.ExtractIA(certs[0].Subject)`

This requires to also parse the user input subject as an IA, which is probably good to "canonicalize" the IA (gets rid of things like upper-case hex digits etc).

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: all files reviewed, 3 unresolved discussions (waiting on @FR4NK-W)

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @FR4NK-W and @matzf)

___
*[scion-pki/certs/validate.go, line 176 at r1](https://reviewable.io/reviews/scionproto/scion/4176#-MykoWCqBYZCzELC1Fq3:-MykwQFi6sssx5WldCIk:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/78fe876464b25212db9670cdb83abf2c76b19a6e/scion-pki/certs/validate.go#L176)):*
> ```Go
> }
> 
> func validateSubject(certs []*x509.Certificate, subjectIdentifier string) error {
> ```

Done.
___
*[scion-pki/certs/validate.go, line 182 at r1](https://reviewable.io/reviews/scionproto/scion/4176#-Mykq_wT1WKOVoxUqpTG:-MykyW-J9OoAIlxoXz1h:b-l99gsf) ([raw file](https://github.com/scionproto/scion/blob/78fe876464b25212db9670cdb83abf2c76b19a6e/scion-pki/certs/validate.go#L182)):*
> This requires to also parse the user input subject as an IA, which is probably good to "canonicalize" the IA (gets rid of things like upper-case hex digits etc).

I'm not sure we want that ("canonicalization", constraint on the subject value).
I would expect an exact match to be required when validating the subject. Otherwise I think we have to formally document which "canonicalizations" are applied, and if all forms are valid when creating the certificate, if different forms are treated as equivalent.

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[scion-pki/certs/validate.go, line 59 at r1](https://reviewable.io/reviews/scionproto/scion/4176#-Mykln178erX9ftdpHDi:-Myl2Iit6H9WOHfNFvM-:b-9taoqp) ([raw file](https://github.com/scionproto/scion/blob/78fe876464b25212db9670cdb83abf2c76b19a6e/scion-pki/certs/validate.go#L59)):*

Wouldn't it be much more natural to check the `subject` for the leaf certificate?
</blockquote></details>

Done.
You are right in a way it simplified things when not excluding that case. It just requires the flag description to be a bit longer and more explicit.

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[scion-pki/certs/validate.go, line 182 at r1](https://reviewable.io/reviews/scionproto/scion/4176#-Mykq_wT1WKOVoxUqpTG:-Myl7ofB9i2NxvgyeiXL:bo5ewig) ([raw file](https://github.com/scionproto/scion/blob/78fe876464b25212db9670cdb83abf2c76b19a6e/scion-pki/certs/validate.go#L182)):*

> This requires to also parse the user input subject as an IA, which is probably good to "canonicalize" the IA (gets rid of things like upper-case hex digits etc).

I'm not sure we want that ("canonicalization", constraint on the subject value).
I would expect an exact match to be required when validating the subject. Otherwise I think we have to formally document which "canonicalizations" are applied, and if all forms are valid when creating the certificate, if different forms are treated as equivalent.
</blockquote></details>

But it would indeed simplify the check:
```
@@ -106,7 +107,11 @@ specifying the --subject flag and the expected subject identifier (distinguished
                                        return err
                                }
                                if flags.subject != "" {
-                                       if err := validateSubject(certs, flags.subject); err != nil {
+                                       expectedSubject, err := addr.ParseIA(flags.subject)
+                                       if err != nil {
+                                               return err
+                                       }
+                                       if err := validateSubject(certs, expectedSubject); err != nil {
                                                return err
                                        }
                                }
@@ -173,22 +178,16 @@ func validateCert(
        return ct, nil
 }
 
-func validateSubject(certs []*x509.Certificate, subjectIdentifier string) error {
+func validateSubject(certs []*x509.Certificate, expectedSubject addr.IA) error {
        if len(certs) != 1 {
                return serrors.New("can only validate the subject of exactly one certificate")
        }
        // Check certificate Distinguished Name
-       certSubjectASID := ""
-       for _, dn := range certs[0].Subject.Names {
-               if dn.Type.Equal(cppki.OIDNameIA) {
-                       certSubjectASID = dn.Value.(string)
-                       break
-               }
-       }
-       if certSubjectASID != subjectIdentifier {
+       subject, err := cppki.ExtractIA(certs[0].Subject)
+       if err != nil || subject != expectedSubject {
                return serrors.New("wrong distinguished name in certificate subject",
-                       "expected", subjectIdentifier,
-                       "actual", certSubjectASID)
+                       "expected", expectedSubject,
+                       "actual", subject)
        }

```

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @FR4NK-W and @matzf)

___
*[scion-pki/certs/validate.go, line 182 at r1](https://reviewable.io/reviews/scionproto/scion/4176#-Mykq_wT1WKOVoxUqpTG:-Myl163U5hdOgLWDwiY_:broxewg) ([raw file](https://github.com/scionproto/scion/blob/78fe876464b25212db9670cdb83abf2c76b19a6e/scion-pki/certs/validate.go#L182)):*

But it would indeed simplify the check:
```
@@ -106,7 +107,11 @@ specifying the --subject flag and the expected subject identifier (distinguished
                                        return err
                                }
                                if flags.subject != "" {
-                                       if err := validateSubject(certs, flags.subject); err != nil {
+                                       expectedSubject, err := addr.ParseIA(flags.subject)
+                                       if err != nil {
+                                               return err
+                                       }
+                                       if err := validateSubject(certs, expectedSubject); err != nil {
                                                return err
                                        }
                                }
@@ -173,22 +178,16 @@ func validateCert(
        return ct, nil
 }
 
-func validateSubject(certs []*x509.Certificate, subjectIdentifier string) error {
+func validateSubject(certs []*x509.Certificate, expectedSubject addr.IA) error {
        if len(certs) != 1 {
                return serrors.New("can only validate the subject of exactly one certificate")
        }
        // Check certificate Distinguished Name
-       certSubjectASID := ""
-       for _, dn := range certs[0].Subject.Names {
-               if dn.Type.Equal(cppki.OIDNameIA) {
-                       certSubjectASID = dn.Value.(string)
-                       break
-               }
-       }
-       if certSubjectASID != subjectIdentifier {
+       subject, err := cppki.ExtractIA(certs[0].Subject)
+       if err != nil || subject != expectedSubject {
                return serrors.New("wrong distinguished name in certificate subject",
-                       "expected", subjectIdentifier,
-                       "actual", certSubjectASID)
+                       "expected", expectedSubject,
+                       "actual", subject)
        }

```
</blockquote></details>

I'd agree if this was an additional cleanup step, but here it's just parsing the value in the only applicable form. This is not a new constraint on the value of the subject as it must be consumed as an ISD-AS anyway. I see this as a courtesy to the user, showing an error because `1-ff00:0:1` does not match `1-FF00:0000:0001` just seems like a nuisance.
___
*[scion-pki/certs/validate.go, line 103 at r4](https://reviewable.io/reviews/scionproto/scion/4176#-MylECqI69q8GU9QLgHF:-MylECqJAoc72Wv9jRl-:b783ccb) ([raw file](https://github.com/scionproto/scion/blob/b2431dc7332bba45473be33ca7e614bb5f7d85c4/scion-pki/certs/validate.go#L103)):*
> ```Go
> 					return err
> 				}
> 				fmt.Printf("Valid %s certificate: %q\n", ct, filename)
> ```

So the subject check can now fail after saying "valid certificate". This could be weird. Perhaps move the check up, after the `checkTime` case.
___
*[scion-pki/certs/validate.go, line 106 at r4](https://reviewable.io/reviews/scionproto/scion/4176#-MylDBie3CxQKa0zg_v7:-MylDBie3CxQKa0zg_v8:b-i2iywl) ([raw file](https://github.com/scionproto/scion/blob/b2431dc7332bba45473be33ca7e614bb5f7d85c4/scion-pki/certs/validate.go#L106)):*
> ```Go
> 			}
> 			if flags.subject != "" {
> 				if err := validateSubject(certs[len(certs)-1], flags.subject); err != nil {
> ```

The leaf is first, `certs[0]`.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @FR4NK-W and @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4176#-MylKb4GD-nTdi8Vzx4w:-MylKb4HD0GC5wGzX3BV:b-xroso6) (no related file):*
I think validate is the wrong place to do this check. If you want to do this, it should be done in verify.

Validate simply checks that the format of a certificate is in line with the requirements for the CPPKI.
(Think of it more as a lint)

---

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @FR4NK-W and @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4176#-MylKb4GD-nTdi8Vzx4w:-MylL8sV0YDbncLFIbwC:bg1h1ov) (no related file):*

I think validate is the wrong place to do this check. If you want to do this, it should be done in verify.

Validate simply checks that the format of a certificate is in line with the requirements for the CPPKI.
(Think of it more as a lint)

</blockquote></details>

For reference, other tools do this in the verify step too:

- https://smallstep.com/docs/step-cli/reference/certificate/verify
  > --host=value Check whether the certificate is for the specified host.



- https://www.openssl.org/docs/man1.1.1/man1/openssl-verify.html
  > -verify_email email
  > Verify if the email matches the email address in Subject Alternative Name or the email in the subject Distinguished Name.

  > -verify_hostname hostname
  > Verify if the hostname matches DNS name in Subject Alternative Name or Common Name in the subject certificate.

  > -verify_ip ip
  > Verify if the ip matches the IP address in Subject Alternative Name of the subject certificate.

---

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @FR4NK-W and @matzf)

___
*[scion-pki/certs/validate.go, line 118 at r5](https://reviewable.io/reviews/scionproto/scion/4176#-MylMDZ_CJL-r2N8eURX:-MylMDZ_CJL-r2N8eURY:b-lvo9b2) ([raw file](https://github.com/scionproto/scion/blob/5bdf05cc44580b7058ce52be3d093a0c7a5ef160/scion-pki/certs/validate.go#L118)):*
> ```Go
> 		fmt.Sprintf("type of cert (%s) (required)", strings.Join(getTypes(), "|")),
> 	)
> 	cmd.Flags().StringVar(&flags.subject, "subject", "",
> ```

I don't think subject is the right term for what you want to express.

The subject of a certificate is expressed as a full DN (distinguished name).
However, you only check a part of that distinguished name.

Two certificates can have a completely different DN (-> different subjects) and still carry the same ISD-AS.
Of course the two subjects conceptually belong to the same SCION AS, but the terminology gets confusing this way.

Also it is inconsistent with existing subject flags:
- https://github.com/scionproto/scion/blob/8f6e2fa475d9e759f24b81dcc4e85fc5485e764a/scion-pki/certs/renew.go#L521-L523
- https://github.com/scionproto/scion/blob/8f6e2fa475d9e759f24b81dcc4e85fc5485e764a/scion-pki/certs/create.go#L153

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @FR4NK-W, @matzf, and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4176#-MylKb4GD-nTdi8Vzx4w:-Mylkws10YqrnHTlqYYD:b-z7vs6) (no related file):*
>I think validate is the wrong place to do this check. If you want to do this, it should be done in verify.
> (Think of it more as a lint)

Makes sense.

Do you want me to split out the test for the `validate` command (minus the ones about subject checks) into a separate PR?

---
___
*[scion-pki/certs/validate.go, line 118 at r5](https://reviewable.io/reviews/scionproto/scion/4176#-MylMDZ_CJL-r2N8eURX:-Mylk9ra3baL7EuT-YEx:bc0t2jx) ([raw file](https://github.com/scionproto/scion/blob/5bdf05cc44580b7058ce52be3d093a0c7a5ef160/scion-pki/certs/validate.go#L118)):*
> I don't think subject is the right term for what you want to express.

@oncilla 
Would `ISD-AS property of the subject` match more closely the current use, be more consistent with existing subject flags?

What's your view on:
>>This requires to also parse the user input subject as an IA, which is probably good to "canonicalize" the IA (gets rid of things like upper-case hex digits etc).

>I'm not sure we want that ("canonicalization", constraint on the subject value).
>I would expect an exact match to be required when validating the subject. Otherwise I think we have to formally document which "canonicalizations" are applied, and if all forms are valid when creating the certificate, if different forms are treated as equivalent.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @FR4NK-W, @matzf, and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4176#-MylKb4GD-nTdi8Vzx4w:-Mys-co-9T8-7qrADXTz:bh6vu5n) (no related file):*

>I think validate is the wrong place to do this check. If you want to do this, it should be done in verify.
> (Think of it more as a lint)

Makes sense.

Do you want me to split out the test for the `validate` command (minus the ones about subject checks) into a separate PR?
</blockquote></details>

I guess you could reframe this PR as "refactor the validate command and add tests" and open a new PR to add the subject ISD-AS to the verify command.

IMO, it makes sense to move the validate command to a new file.

---

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @FR4NK-W and @matzf)

___
*[scion-pki/certs/validate.go, line 182 at r1](https://reviewable.io/reviews/scionproto/scion/4176#-Mykq_wT1WKOVoxUqpTG:-Mys0N5z1RfUql8Ntpi6:b2kyqwn) ([raw file](https://github.com/scionproto/scion/blob/78fe876464b25212db9670cdb83abf2c76b19a6e/scion-pki/certs/validate.go#L182)):*

I'd agree if this was an additional cleanup step, but here it's just parsing the value in the only applicable form. This is not a new constraint on the value of the subject as it must be consumed as an ISD-AS anyway. I see this as a courtesy to the user, showing an error because `1-ff00:0:1` does not match `1-FF00:0000:0001` just seems like a nuisance.
</blockquote></details>

This is a tough one. I regret a bit that we encode this as a string instead of a integer in the DN of the certificate.

If we take the go implementation as the reference. We interpret 1-FF00:0:1, 1-ff00:0000:1 and 1-ff00:0:1 all as the same ISD-AS number during extraction.
On the other hand, a distinguished name with 1-FF00:0:1 is different from 1-ff00:0:1. This matters for example in signature verification.

So, I guess it depends on the intended use case of this flag.

I _think_ the command should not care about the formatting and should just use the canonical form.
But that needs to be explicitly mentioned. I'm also a bit worried that this could be used as a potential attack vector if you have to parser implementations and one tells you all good, and the other tells you the subject does not match.
___
*[scion-pki/certs/validate.go, line 118 at r5](https://reviewable.io/reviews/scionproto/scion/4176#-MylMDZ_CJL-r2N8eURX:-Mys003zDaSj9stgmwii:b-hel4kg) ([raw file](https://github.com/scionproto/scion/blob/5bdf05cc44580b7058ce52be3d093a0c7a5ef160/scion-pki/certs/validate.go#L118)):*
> Would ISD-AS property of the subject match more closely the current use, be more consistent with existing subject flags?

I think yes, as long as the flag is renamed to something other than `subject`

FR4NK-W:
The new flag is being added to the verify command, #4179, following the discussion above

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @FR4NK-W, @matzf, and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4176#-MylKb4GD-nTdi8Vzx4w:-MzxoNr1DY0HWQE7T8xw:b-896fix) (no related file):*

I guess you could reframe this PR as "refactor the validate command and add tests" and open a new PR to add the subject ISD-AS to the verify command.

IMO, it makes sense to move the validate command to a new file.
</blockquote></details>

Done.

---

matzf (approved):
:lgtm:

> Reviewed 2 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176#-:-Mzz51mq0GCNcigZdahL:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @FR4NK-W and @oncilla)

oncilla (approved):
:lgtm:

> Reviewed 1 of 4 files at r1, 2 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176#-:-N0FEQ740SZabnx5LnvS:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @oncilla)

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4176)* status: all files reviewed, 2 unresolved discussions (waiting on @FR4NK-W and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4176#-N0FER27EWilmq8HOrjD:-N0FER27EWilmq8HOrjE:bgnhtmv) (no related file):*
Thank you for the contribution. This will be included in our next open source run.

---

PR #4187, opened 2022-04-14
Add Gobra to the CI and specify and verify infofield.go and hopfield.go (experimental)

jcp19:
As discussed in a meeting with @shitz and @oncilla, this PR introduces experimental support for program verification in the CI of SCION, using [Gobra](https://github.com/viperproject/gobra). The idea is to evaluate how feasible it is to maintain a fraction of the SCION code-base that has been formally verified.

Initially, we focus on the files `infofield.go` and `hopfield.go` in the package `pkg/slayers/path`. These files implement fundamental data structures for the SCION implementation while being a small-enough target for our initial efforts to integrate Gobra in the CI of SCION.
Currently, we prove memory safety (incl. no accesses out-of-bounds) of the methods `SerializeTo` and `DecodeFromBytes` for both `InfoField` and `HopField`, and we prove that these methods always terminate.
There are also a couple of members marked with the annotation `//@ trusted`. These correspond to methods specified but not verified because they use some feature of Go that is still unsupported by Gobra. More concretely:
- ~~method `(*InfoField).UpdateSegID`: Gobra uses a stricter notion of Addressability at the moment and rejects the expression `hfMac[:2]`~~ (addressed)
- method `InfoField.String`: Gobra does not yet support the package `fmt`

To verify the annotated code-base, Gobra took around 3 minutes on a machine with a 2.6 GHz 6-Core Intel Core i7 and 16GB of RAM. Despite this, we expect the impact of these changes on contributions to the code base to be small given the modular nature of Gobra: Gobra verifies each method separately and our CI action caches the verification results per method.  As such, a re-verification of a specified method should only be triggered by changes to that method or to one of its dependencies.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4187)


matzf (changes requested):
Hi @jcp19, thank you for creating this PR, this is quite exciting!

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4187#-:-N00ywZa34UZI5CU1rs8:bd52no0)* status: 0 of 4 files reviewed, 7 unresolved discussions (waiting on @jcp19 and @oncilla)

___
*[.github/workflows/gobra.yml, line 14 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N018vZREMOUm6G8qWcd:-N018vZREMOUm6G8qWce:b-ibzs2n) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L14)):*
> ```YAML
>         uses: actions/checkout@v2
>       - name: Verify the specified files
>         uses: viperproject/gobra-action@main
> ```

Could we pin a specific version so that when there are breaking changes to the actions does not break our builds globally? And also pin  `imageVersion`.
This would allow to keep using the older version until we explicitly update the version and update all the annotations.
___
*[.github/workflows/gobra.yml, line 15 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01NKDbDMylWzJlDAjz:-N01NKDbDMylWzJlDAk-:b-j95da4) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L15)):*
> ```YAML
>       - name: Verify the specified files
>         uses: viperproject/gobra-action@main
>         with:
> ```

If a developer wants to run the verifier locally, either when adding new annotations or when changing annotated code, it would be helpful if there was a clear way to run the verifier with the same settings as this github action.
Is there an option to specify the settings in this block in a configuration file in the repository instead? Then we could e.g. add a `make verify` target to run the verifier locally, very similar to how we currently run `golangci-lint`.

Speaking of linters -- would it be at all reasonable to integrate this verifier into our build system by integrating it as a linter pass, e.g. in [nogo](https://github.com/bazelbuild/rules_go/blob/master/go/nogo.rst)? Not now, I mean, but for later.
___
*[.github/workflows/gobra.yml, line 25 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01CeA50fCN-2xCe0cs:-N01CeA50fCN-2xCe0ct:bx3x6i1) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L25)):*
> ```YAML
>           packages: 'path'
>           # Gobra only verifies files annotated with the header "// +gobra"
>           headerOnly: '1'
> ```

These settings appear redundant; if only `// +gobra` files are verified, why do we also need to specify `srcDirectory`, and `packages`?
FWIW, to me it would also work to just have an explicit list of files that are to be verified.
___
*[.github/workflows/gobra.yml, line 34 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N015wyz1tny4o8lJadi:-N015wyz1tny4o8lJadj:b4heeue) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L34)):*
> ```YAML
>         with:
>           name: cache
>           path: ${{ runner.workspace }}/.gobra/cache.json
> ```

I'm not very familiar with github actions; my superficial understanding is that this only stores the cache artifact, but it never appears to be "consumed". Should this use [actions/cache](https://github.com/actions/cache) instead? Or how does this work?
___
*[.github/workflows/gobra.yml, line 34 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01LH2O-qlYb6lPSjXP:-N01LH2O-qlYb6lPSjXQ:b-vup05w) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L34)):*
> ```YAML
>         with:
>           name: cache
>           path: ${{ runner.workspace }}/.gobra/cache.json
> ```

Should we also upload the report?
___
*[pkg/slayers/path/hopfield.go, line 95 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01J3zQ7jm7iAZBdAFJ:-N01J3zQ7jm7iAZBdAFK:b9842b6) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/pkg/slayers/path/hopfield.go#L95)):*
> ```Go
> 	//@ assert forall i int :: { h.Mac[:][i] } 0 <= i && i < len(h.Mac[:]) ==> acc(&h.Mac[:][i])
> 	//@ assert forall i int :: { raw[6:6+MacLen][i] } 0 <= i && i < len(raw[6:6+MacLen]) ==>
> 	//@     acc(&raw[6:6+MacLen][i], 1/2)
> ```

These `assert`s here and also in the simpler cases above just seem to describe the semantics of the slice syntax for every occurrence. Will gobra eventually be able to "understand" this on it's own? It looks entirely mechanical to me, and a comment block of this size is certainly distracting.
___
*[pkg/slayers/path/hopfield.go, line 96 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01JIm10NJU8Ekp7Df-:-N01JIm10NJU8Ekp7Df0:bo10gqp) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/pkg/slayers/path/hopfield.go#L96)):*
> ```Go
> 	//@ assert forall i int :: { raw[6:6+MacLen][i] } 0 <= i && i < len(raw[6:6+MacLen]) ==>
> 	//@     acc(&raw[6:6+MacLen][i], 1/2)
> 	copy(h.Mac[:], raw[6:6+MacLen] /*@, perm(1/4)@*/)
> ```

I don't understand this fractional bookkeeping. What is the difference between `1/2` and `1/4`? Why not just `read`?

jcp19 (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4187)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @jcp19, @matzf, and @oncilla)

___
*[.github/workflows/gobra.yml, line 34 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N015wyz1tny4o8lJadi:-N0HEf9QEVy7dYAaaKFO:b-4jwdme) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L34)):*

I'm not very familiar with github actions; my superficial understanding is that this only stores the cache artifact, but it never appears to be "consumed". Should this use [actions/cache](https://github.com/actions/cache) instead? Or how does this work?
</blockquote></details>

Yes, you are totally correct. I fixed it in the meanwhile
___
*[.github/workflows/gobra.yml, line 34 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01LH2O-qlYb6lPSjXP:-N0HF1N5ASklQUqYofTl:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L34)):*

Should we also upload the report?
</blockquote></details>

Done

jcp19 (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4187)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`.github/workflows/gobra.yml` line 14 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N018vZREMOUm6G8qWcd:-N30nRmAAcfbYvV9ki4s:b-jod0v8) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L14)):*

Could we pin a specific version so that when there are breaking changes to the actions does not break our builds globally? And also pin  `imageVersion`.
This would allow to keep using the older version until we explicitly update the version and update all the annotations.
</blockquote></details>

I have now fixed the version of the gobra-action used in the CI. Here, we don't need to pin the `imageVersion`, as the selected version for the gobra-action already uses a fixed  version of Gobra if none is specified.
___
*[`.github/workflows/gobra.yml` line 15 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01NKDbDMylWzJlDAjz:-N3130iU0XHIHewo5ZlT:bhx15wn) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L15)):*

If a developer wants to run the verifier locally, either when adding new annotations or when changing annotated code, it would be helpful if there was a clear way to run the verifier with the same settings as this github action.
Is there an option to specify the settings in this block in a configuration file in the repository instead? Then we could e.g. add a `make verify` target to run the verifier locally, very similar to how we currently run `golangci-lint`.

Speaking of linters -- would it be at all reasonable to integrate this verifier into our build system by integrating it as a linter pass, e.g. in [nogo](https://github.com/bazelbuild/rules_go/blob/master/go/nogo.rst)? Not now, I mean, but for later.
</blockquote></details>

Right now, there is no direct way to do this, although it is something that I would like to do in the near future. I really like your suggestion to try to integrate it as a linter pass in nogo. For the time being, I think that the easiest way to run gobra locally with the given configuration is to use something like [act](https://github.com/nektos/act) to run the Github action locally.
___
*[`.github/workflows/gobra.yml` line 25 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01CeA50fCN-2xCe0cs:-N30ny7S6qu0mGwntsKT:b-vjkps0) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L25)):*

These settings appear redundant; if only `// +gobra` files are verified, why do we also need to specify `srcDirectory`, and `packages`?
FWIW, to me it would also work to just have an explicit list of files that are to be verified.
</blockquote></details>

Yes, they are redundant but they were required in a previous version of the action. In the time since the review, we made extensive improvement to the action and gobra, making it possible to not provide these options.
___
*[`pkg/slayers/path/hopfield.go` line 95 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01J3zQ7jm7iAZBdAFJ:-N30o_5x0RLmOP3AsUb8:b-29fntb) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/pkg/slayers/path/hopfield.go#L95)):*

These `assert`s here and also in the simpler cases above just seem to describe the semantics of the slice syntax for every occurrence. Will gobra eventually be able to "understand" this on it's own? It looks entirely mechanical to me, and a comment block of this size is certainly distracting.
</blockquote></details>

You are right that these assertions are distracting and can be obtained mechanically. However, they are still required for Gobra to verify these files. Addressing this in Gobra requires extensive changes to our encoding of the slices, which is on our mid-term plans. I will update these files as soon as these annotations are no longer required.
___
*[`pkg/slayers/path/hopfield.go` line 96 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01JIm10NJU8Ekp7Df-:-N30z9IXBLZTTsfeW7vt:b95752v) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/pkg/slayers/path/hopfield.go#L96)):*

I don't understand this fractional bookkeeping. What is the difference between `1/2` and `1/4`? Why not just `read`?
</blockquote></details>

For practical purposes, the most crucial point is that both permission amounts (`1/2` and `1/4`) allow reading a value from a heap location, but not writing to it. For consistency, I now use the same fraction in all places where a "read" permission is required, when possible. Using `read` here though is not the same as using one of these fixed amounts. `read` stands for an existentially quantified positive value. Different occurrences of `read` in the spec of a method are thus not guaranteed to stand for the same permission value. Because of this, the following example fails to verify:
```go
package f

// `_` here stands for the `read` permission
//@ requires acc(x, _)
//@ ensures  acc(x, _)
func getValue(x *int) int {
	return *x
}

func main() {
	x := new(int)
	//@ assert acc(x)
	_ = getValue(x)
	// The following assertion fails:
	// Gobra cannot guarantee that the
	// permission amount transferred back
	// from `getValue` is the same as the
	// amount transferred to it.
	//@ assert acc(x)
}

```

The same example succeeds when `_` is replaced by a fixed permission amount:
```go
package f

//@ requires acc(x, 1/2)
//@ ensures  acc(x, 1/2)
func getValue(x *int) int {
	return *x
}

func main() {
	x := new(int)
	//@ assert acc(x)
	_ = getValue(x)
	// The following assertion
	// is verified successfully
	// now.
	//@ assert acc(x)
}
```

shitz (changes requested):
> Reviewed 1 of 4 files at r1, 1 of 2 files at r3, 2 of 3 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4187)* status: all files reviewed, 2 unresolved discussions (waiting on @jcp19 and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4187#-N4XSyzE7Q194OhMm3qt:-N4XSyzE7Q194OhMm3qu:b-aeb5ph) (no related file):*
Hi @jcp19 . Great to see this work. Would it be possible for you to add comments on the individual specifications that describe in human language what is going on. Most people looking at this won't be familiar with verification let alone Gobra. It would be helpful to still understand roughly what's going on.

___
___
*[`pkg/private/serrors/errors_spec.gobra` line 27 at r6](https://reviewable.io/reviews/scionproto/scion/4187#-N4XRZUD42kKO60XIYpJ:-N4XRZUEDXvke3iRtMlq:b-h0ksw5) ([raw file](https://github.com/scionproto/scion/blob/aaf73e7b8c8ee46e1ff19a72e666f8078ef4b697/pkg/private/serrors/errors_spec.gobra#L27)):*
> ```
> ensures err != nil
> decreases
> func New(msg string, errCtx ...interface{}) (err error)
> ```

nit - missing newline

jcp19 (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4187)* status: all files reviewed, 1 unresolved discussion (waiting on @jcp19 and @oncilla)

___
*[`pkg/slayers/path/hopfield.go` line 96 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01JIm10NJU8Ekp7Df-:-N7R36N61tft2H-t9wDu:bxyrmu8) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/pkg/slayers/path/hopfield.go#L96)):*

Thanks for the explanation. This example makes sense to me. What I still don't quite understand is why we care about different permission levels in the first place. Does this represent something like concurrent readers?

Anyway, I'm resolving the discussion -- this is probably not the best forum to ask beginner questions about gobra :smile: 
</blockquote></details>

This is a very good point. Unfortunately, these fractional permissions are an artifact of our verification approach. There are other approaches that do away with this fractional book-keeping (e.g., https://www.microsoft.com/en-us/research/wp-content/uploads/2016/12/krml225.pdf), essentially only requiring distinguishing between 'read' and 'write' permissions, but we do not support them in Gobra yet.

jcp19 (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4187)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @jcp19, @matzf, @oncilla, and @shitz)

___
*[`.github/workflows/gobra.yml` line 34 at r1](https://reviewable.io/reviews/scionproto/scion/4187#-N01LH2O-qlYb6lPSjXP:-N7tRiwU1Ae9X0uPzSa9:b-bcz690) ([raw file](https://github.com/scionproto/scion/blob/9e3c40e0839f7396fed748dfc0bec3ef6e17e912/.github/workflows/gobra.yml#L34)):*

The path does not seem to match the expected file:
```
No files were found with the provided path: /home/runner/work/scion/.gobra/stats.json. No artifacts will be uploaded.
```

Note: in the `gobra-action`'s docker invocation I see `-v "/home/runner/work/scion/scion":"/github/workspace"` (mismatch `scion/scion` vs `scion/.gobra`).

Nit: trailing whitespace

</blockquote></details>

This was actually caused by a bug in our Github action that was causing the verification statistics to not be produced. It is now fixed :)

jcp19 (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4187)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @matzf, @oncilla, and @shitz)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4187#-N4XSyzE7Q194OhMm3qt:-N7toOAU7uYLiIDnbtFQ:bb24egr) (no related file):*

Hi @jcp19 . Great to see this work. Would it be possible for you to add comments on the individual specifications that describe in human language what is going on. Most people looking at this won't be familiar with verification let alone Gobra. It would be helpful to still understand roughly what's going on.
</blockquote></details>

Hey @shitz. I've incorporated small descriptions of the contracts of every specified method which hopefully will make it easier to understand what the specs mean. At the same time, I avoided deep descriptions of the features of Gobra (like the permission model) in order to make these comments easy to grasp. Feel free to let me know if you think that the comments should be more detailed. Additionally, I tried to interleave specification clauses with sentences describing their meaning. Let me know if this style works for you.

___

shitz (changes requested):
> Reviewed 1 of 2 files at r7, 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4187)* status: all files reviewed, 1 unresolved discussion (waiting on @jcp19 and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4187#-N4XSyzE7Q194OhMm3qt:-N90BCj526gVtPpAJMeq:bxykfbm) (no related file):*
<details><summary><i>Previously, jcp19 (João Pereira) wrote…</i></summary><blockquote>

Hey @shitz. I've incorporated small descriptions of the contracts of every specified method which hopefully will make it easier to understand what the specs mean. At the same time, I avoided deep descriptions of the features of Gobra (like the permission model) in order to make these comments easy to grasp. Feel free to let me know if you think that the comments should be more detailed. Additionally, I tried to interleave specification clauses with sentences describing their meaning. Let me know if this style works for you.
</blockquote></details>

Very nice, thank you! 

___
___
*[`.gitignore` line 95 at r8](https://reviewable.io/reviews/scionproto/scion/4187#-N90B6Sd7LLI1mFchVIM:-N90B6Sd7LLI1mFchVIN:b-oioogp) ([raw file](https://github.com/scionproto/scion/blob/51436caa0dcd42f94ad64c087dc512ae08f0680e/.gitignore#L95)):*
> ```ignore
> *.vpr
> tmp/
> logger.log
> ```

Where does `tmp/` and `logger.log` come from?

shitz (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4187)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

___
*[`.gitignore` line 95 at r8](https://reviewable.io/reviews/scionproto/scion/4187#-N90B6Sd7LLI1mFchVIM:-N96Wn_aCvGNScSXA9py:b-p2uio8) ([raw file](https://github.com/scionproto/scion/blob/4430d88194dc8500aafe89ff8e12ea6e20f7d12f/.gitignore#L95)):*
<details><summary><i>Previously, jcp19 (João Pereira) wrote…</i></summary><blockquote>

These are just some additional files generated by Gobra when you verify the code. `tmp` basically contains some .smt2 which correspond to the proof obligations that Gobra ends up passing to an smt solver, which is sometimes (although very rarely) useful for debugging. I can remove these files from the .gitignore if you think they should not be here.
</blockquote></details>

If gobra creates them we should ignore them (note beside I think it should use the designated temp folder based on the OS it's running on, but that's not something you can influence here)

PR #4202, opened 2022-05-27
lint: detect captured loop var in go/defer statement

### LINKED ISSUE #2164: Change use of unsafe `for` construct
A common for-loop clause in Go looks like this:
```
for i, val := range vals {
    // do something with val
}
```
However, this has some hidden pitfalls if `val` is being capture by a closure that is run as a go-routine inside the for-loop, i.e.,
```
for i, val := range vals {
    go func() {
        // do something with val
    }
}
```
The problem is that by the time the go routine is run the value of `val` can have changed which leads to unexpected results. This has bitten us already multiple times in the past.
Also, even if a for-loop does not have a `go func() {}` statement the loop body can change over time and it's easy to miss this unintuitive interaction.

There are several safe variants of the for-loop clause, but for consistency reasons we decided to change all for-loops to the form
```
for i := range vals {
    val := vals[i]
    // do something with val
}
```

matzf:
Detect loop variable captured by function in go or defer statement.
The loop variable is reassigned when stepping in the loop, and so
by the time the go routine is run, the value may have changed.

Fix the single occurrence of the pattern.

Update semgrep, as the rules did not appear to work as expected with the
older version.
With this newer version, we now need an (empty) .semgrepignore file, as
by default semgrep ignores all _test.go files, which seems undesirable.
The docker command for semgrep is also changed to explicitly invoke
`semgrep`, as directed by a deprecation warning.

Closes #2164.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4202)

### DIFF
--- a/.semgrepignore
+++ b/.semgrepignore
@@ -1,0 +1,1 @@
+# Empty .semgrepignore; the default one skips all _test.go.
--- a/gateway/control/remotemonitor.go
+++ b/gateway/control/remotemonitor.go
@@ -119,7 +119,8 @@ func (rm *RemoteMonitor) process(ctx context.Context, ias []addr.IA) {
 	defer rm.stateMtx.Unlock()
 	logger := log.FromCtx(ctx)
 	newWatchers := make(map[addr.IA]watcherEntry)
-	for _, ia := range ias {
+	for i := range ias {
+		ia := ias[i]
 		we, ok := rm.currentWatchers[ia]
 		if ok {
 			// Watcher for the remote IA exists. Move it to the new map of
--- a/gateway/control/remotemonitor.go
+++ b/gateway/control/remotemonitor.go
@@ -119,8 +119,8 @@ func (rm *RemoteMonitor) process(ctx context.Context, ias []addr.IA) {
 	defer rm.stateMtx.Unlock()
 	logger := log.FromCtx(ctx)
 	newWatchers := make(map[addr.IA]watcherEntry)
-	for i := range ias {
-		ia := ias[i]
+	for _, ia := range ias {
+		ia := ia
 		we, ok := rm.currentWatchers[ia]
 		if ok {
 			// Watcher for the remote IA exists. Move it to the new map of
--- a/tools/lint.sh
+++ b/tools/lint.sh
@@ -89,8 +89,8 @@ md_lint() {
 semgrep_lint() {
     lint_header "semgrep"
     lint_step "custom rules"
-    docker run --rm -v "${PWD}:/src" returntocorp/semgrep@sha256:8b0735959a6eb737aa945f4d591b6db23b75344135d74c3021b7d427bd317a66 \
-        --config=/src/tools/lint/semgrep --error
+    docker run --rm -v "${PWD}:/src" returntocorp/semgrep@sha256:3bef9d533a44e6448c43ac38159d61fad89b4b57f63e565a8a55ca265273f5ba \
+       semgrep --config=/src/tools/lint/semgrep --error
 }
 
 openapi_lint() {
--- a/tools/lint/semgrep/loopvar_capture.yml
+++ b/tools/lint/semgrep/loopvar_capture.yml
@@ -1,24 +1,44 @@
 rules:
   - id: loopvar_capture
-    pattern-either:
-      - pattern: |
-          for $X := ...; ...; <... $X ...> { ...; go func(...){ ...; <... $X ...>; ... }(...); ... }
-      - pattern: |
-          for $X := ...; ...; <... $X ...> { ...; defer func(...){ ...; <... $X ...>; ... }(...); ... }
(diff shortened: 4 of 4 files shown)


oncilla (changes requested):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4202)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`tools/lint/semgrep/loopvar_capture.yml` line 3 at r1](https://reviewable.io/reviews/scionproto/scion/4202#-N3DeaFW1ErxKW-J717N:-N3DeaFW1ErxKW-J717O:btvz811) ([raw file](https://github.com/scionproto/scion/blob/cdca42c680803a73ab81298e5dbfb5fba19af5e1/tools/lint/semgrep/loopvar_capture.yml#L3)):*
> ```YAML
> rules:
>   - id: loopvar_capture
>     pattern-either:
> ```

Would be neat if we can also catch captured loop variables in errgroup.Go calls. See https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Go
___
*[`tools/lint/semgrep/loopvar_capture.yml` line 3 at r1](https://reviewable.io/reviews/scionproto/scion/4202#-N3DfKan3uGPJ4PU8jdR:-N3DfKan3uGPJ4PU8jdS:bhmfkc7) ([raw file](https://github.com/scionproto/scion/blob/cdca42c680803a73ab81298e5dbfb5fba19af5e1/tools/lint/semgrep/loopvar_capture.yml#L3)):*
> ```YAML
> rules:
>   - id: loopvar_capture
>     pattern-either:
> ```

These patterns lead to false positives for the following style of code:

```go
    for _, v := range values {
        v := v // create a new 'v'.
        go func() {
            fmt.Println(v)
            done <- true
        }()
    }
```

Both https://go.dev/doc/effective_go#channels and  https://go.dev/doc/faq#closures_and_goroutines  mention this style of code as valid and idomatic.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4202)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @oncilla)

___
*[`tools/lint/semgrep/loopvar_capture.yml` line 3 at r1](https://reviewable.io/reviews/scionproto/scion/4202#-N3DeaFW1ErxKW-J717N:-N3Im-CAAn_Z0ZB2fnLQ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cdca42c680803a73ab81298e5dbfb5fba19af5e1/tools/lint/semgrep/loopvar_capture.yml#L3)):*

Would be neat if we can also catch captured loop variables in errgroup.Go calls. See https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Go
</blockquote></details>

Done.
___
*[`tools/lint/semgrep/loopvar_capture.yml` line 3 at r1](https://reviewable.io/reviews/scionproto/scion/4202#-N3DfKan3uGPJ4PU8jdR:-N3Im0mH7v-7l79CLFm0:b-mejc6t) ([raw file](https://github.com/scionproto/scion/blob/cdca42c680803a73ab81298e5dbfb5fba19af5e1/tools/lint/semgrep/loopvar_capture.yml#L3)):*

These patterns lead to false positives for the following style of code:

```go
    for _, v := range values {
        v := v // create a new 'v'.
        go func() {
            fmt.Println(v)
            done <- true
        }()
    }
```

Both https://go.dev/doc/effective_go#channels and  https://go.dev/doc/faq#closures_and_goroutines  mention this style of code as valid and idomatic.

 
</blockquote></details>

At least with the version of semver that is now used in the repo, this case is handled correctly. I've updated the fix in `gateway/control/remotemonitor.go` using this style to illustrate this.


However, when adding the `errgroup.Go` match, I have encountered false-positive matches that use a double re-assignment of the loop variables;

```
for  i, x := range values {
   i, x := i, x
   go func() { fmt.Println(i, x) }()
}
```

Another false-positive that showed up is also related to double assignment:
```
for _, x := range values {
   go func() { v, _ := foo() } /// Captured loop variable `_` in go or defer statement
}
```

I've added specific rules to exclude both of these false positives. I might also file an issue with semgrep about this.

oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4202#-:-N3K4GUz2K6V28XVwt3U:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4202#-N3K49gcEErIyPc2USoU:-N3K49gcEErIyPc2USoV:biwvkbi) (no related file):*
Great stuff! :100: 

And the timing is great, we have also hit by this recently.

___

PR #4211, opened 2022-06-01
ci: speed up go-module-proxy

matzf:
Adapt configuration of go-module-proxy to improve performance, in
particular for `/list` queries, which occur for example during gazelle's
`update-repos` during `make go_deps.bzl`.

Upload go-module-proxy logs and metrics as artifacts.

Do not stop go-module-proxy or bazel-remote-cache on agent after exiting
between jobs to keep warmed up in-memory state. Any changes to the
docker configuration of these services between jobs are automatically
handled by docker-compose.

Aside; in tools/update_testdata.sh, put bazel's go_sdk first into the
path, so that the script picks up these binaries for running the
tests. This only makes a difference for local execution of the script,
the buildkite agents don't have go installed.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4211)

### DIFF
--- a/.buildkite/hooks/go-module-proxy.yml
+++ b/.buildkite/hooks/go-module-proxy.yml
@@ -4,7 +4,8 @@ services:
   go-module-proxy:
     container_name: go-module-proxy
     network_mode: host
-    image: $REGISTRY/gomods/athens:v0.11.0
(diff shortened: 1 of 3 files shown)


lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4211)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4213, opened 2022-06-03
deps: update dependencies to support go-1.18

matzf:
The previously used versions for inet.af/netaddr and lucas-clemente/quic-go
do not support building with go-1.18. Update to the respective latest versions.

Adaptation of the code is required due to a change in [quic-go v0.27.0](https://github.com/lucas-clemente/quic-go/releases/tag/v0.27.0):

> This release contains a big breaking change: We renamed the quic.Session
> to quic.Connection. The term "session" was used in gQUIC, but is not
> used in the QUIC RFC any more. Instead, it's called a "connection"
> there. We are aware that this API change will cause breakage for
> downstream users, but it had to be done sooner or later.


oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4213)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4217, opened 2022-06-10
drkey: public library and protobuf definitions

JordiSubira:
This PR is the first of a series of PRs which add the DRKey feature.

This PR contains:
- pkg/drkey: public drkey library.
- private/drkey/protocoltest: common utils for UT.
- proto/drkey: pb messages.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4217)


JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4217)* status: 8 of 14 files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`private/drkey/protocoltest/commons.go` line 28 at r1](https://reviewable.io/reviews/scionproto/scion/4217#-N4gk0cI4Ub4VTxI1iZ2:-N4mdca5BjKuEFEzuXVS:b-ran02i) ([raw file](https://github.com/scionproto/scion/blob/aec8932ba676d73ad5e8ba3877ba16f95be94809/private/drkey/protocoltest/commons.go#L28)):*

Encode the `srcIA` in the AS secret? This would ensure that keys derived by different source ASes are different and so catch any potential mixups during the tests.
</blockquote></details>

Good observation! I've just append it to the fixed bytes.
___
*[`proto/drkey/mgmt/v1/mgmt.proto` line 19 at r1](https://reviewable.io/reviews/scionproto/scion/4217#-N4c8a2K-1Mx_LkTvSxa:-N4mduZE1Hh1GR3RU5As:bh6wnkf) ([raw file](https://github.com/scionproto/scion/blob/aec8932ba676d73ad5e8ba3877ba16f95be94809/proto/drkey/mgmt/v1/mgmt.proto#L19)):*

I know we've discussed this before in some form, but now looking at this again I'm not sure that this is the right organization for these request definitions. In the full implementation, this package is shared for both the daemon service and the intra-AS drkey service on the control service. Looking at this now I realized that only two of these request/responses are actually available both on the daemon and the : `ASHost` and `HostAS`; the interfaces for `SV` and `Lvl1` are only available on the control service, and  the `HostHost` only needs to exist on the daemon (if at all; because it can be derived locally from ASHostKey, maybe this only needs to be supported in the library instead of via an RPC).

So with that background, I think it would make sense to only define the `enum Protocol` here in this shared package and define the request/responses separately in `control_plane/v1/drkey.proto` and `daemon/v1/daemon.proto`, and accept that this leads to duplication for the `ASHost` and `HostAS` request/response definitions.


Either way, could we ditch the `mgmt` subpackage, i.e. change `package` to `proto.drkey.v1`, and the file path to `proto/drkey/v1/drkey.proto`? 
</blockquote></details>

Yes you're right regarding the SV and the Lvl1Request, but as you said some other request/response messages are shared. Personally, I find it more organized to have the drkey messages in this package, especially, because we avoid this duplicity that you mentioned. However, we already discussed about PB best practices recommending not to use messages among several packages, so it's alright if we decide to go for this approach. Note that the IntraLvl1 message might be duplicated as well upon it's being use by other services (e.g. COLIBRI) (i.e. ASHost, HostAS and IntraLvl1).

When we agree on the previous, I'll remove the mgmt subpackage, in the same pass.

shitz (changes requested):
> Reviewed 4 of 14 files at r1, 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4217)* status: all files reviewed, 28 unresolved discussions (waiting on @JordiSubira and @matzf)

___
*[`pkg/drkey/db.go` line 26 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KZWGo4buxCvjc3S32:-N5KZWGo4buxCvjc3S33:undefined) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/db.go#L26)):*
> ```Go
> )
> 
> var ErrKeyNotFound = serrors.New("Key not found")
> ```





_Suggestion:_
```Go
k
```
___
*[`pkg/drkey/db.go` line 30 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KZd_VC4MeQzE9p7pb:-N5KZd_VC4MeQzE9p7pc:undefined) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/db.go#L30)):*
> ```Go
> // SecretValueDB is the database for Secret Values.
> type SecretValueDB interface {
> 	GetSV(ctx context.Context, meta SVMeta) (SV, error)
> ```





_Suggestion:_
```Go
GetValue
```
___
*[`pkg/drkey/db.go` line 31 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KZiPh8jfPu-3D9-hB:-N5KZiPh8jfPu-3D9-hC:undefined) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/db.go#L31)):*
> ```Go
> type SecretValueDB interface {
> 	GetSV(ctx context.Context, meta SVMeta) (SV, error)
> 	InsertSV(ctx context.Context, key SV) error
> ```





_Suggestion:_
```Go
InsertValue
```
___
*[`pkg/drkey/db.go` line 32 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KZm251-J33UdbVtKI:-N5KZm251-J33UdbVtKJ:undefined) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/db.go#L32)):*
> ```Go
> 	GetSV(ctx context.Context, meta SVMeta) (SV, error)
> 	InsertSV(ctx context.Context, key SV) error
> 	DeleteExpiredSV(ctx context.Context, cutoff time.Time) (int64, error)
> ```





_Suggestion:_
```Go
DeleteExpiredValues
```
___
*[`pkg/drkey/drkey.go` line 89 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KbRt6BVqmhkjpD9kL:-N5KbRt6BVqmhkjpD9kM:undefined) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/drkey.go#L89)):*
> ```Go
> 
> // SVMeta represents the information about a DRKey secret value.
> type SVMeta struct {
> ```





_Suggestion:_
```Go
SecretValueMeta
```
___
*[`pkg/drkey/drkey.go` line 95 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KbPM68htPCn8fMubl:-N5KbPM68htPCn8fMubm:undefined) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/drkey.go#L95)):*
> ```Go
> 
> // SV represents a DRKey secret value.
> type SV struct {
> ```





_Suggestion:_
```Go
SecretValue
```
___
*[`pkg/drkey/drkey.go` line 131 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KbxnR1CtGslXCkFa9:-N5KbxnR1CtGslXCkFaA:b-yscstj) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/drkey.go#L131)):*
> ```Go
> 
> // / Lvl1Meta contains metadata to obtain a lvl1 key.
> type Lvl1Meta struct {
> ```

Replace all occurrences of `Lvl1` with `Level1`



_Code quote:_
```Go
Lvl1
```
___
*[`pkg/drkey/generic.go` line 24 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KlD7_8QF2nV5r364j:-N5KlD7_8QF2nV5r364k:bnb3cly) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/generic.go#L24)):*
> ```Go
> 
> // GenericDeriver implements the level 2/3 generic drkey derivation.
> type GenericDeriver struct {
> ```

I'd put this in a separate package `drkey/generic`.
___
*[`pkg/drkey/generic.go` line 25 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KkG1P3yq8oTilnLcs:-N5KkG1P3yq8oTilnLct:b5l9ucx) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/generic.go#L25)):*
> ```Go
> // GenericDeriver implements the level 2/3 generic drkey derivation.
> type GenericDeriver struct {
> 	buf [32]byte
> ```

What is `buf`?
___
*[`pkg/drkey/generic.go` line 28 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KkU2b7bXF49JjAMh7:-N5KkU2b7bXF49JjAMh8:bpgqj9r) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/generic.go#L28)):*
> ```Go
> }
> 
> func (p *GenericDeriver) inputDeriveLvl2(input []byte, derType keyType,
> ```

Reformat
___
*[`pkg/drkey/generic.go` line 28 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KkeMe200oZVXqFU6p:-N5KkeMe200oZVXqFU6q:b-e07a4i) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/generic.go#L28)):*
> ```Go
> }
> 
> func (p *GenericDeriver) inputDeriveLvl2(input []byte, derType keyType,
> ```

What does this method do?

In general, it would be good to have pointers to the DRKey specification on https://docs.scion.org/en/latest/cryptography/drkey.html



_Suggestion:_
```Go
deriveLevel2Input
```
___
*[`pkg/drkey/protocol.go` line 33 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Km6Xz1ylsKT-ZdR1o:-N5Km6Xz1ylsKT-ZdR1p:b-mm7qjv) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L33)):*
> ```Go
> // Key types.
> const (
> 	asToAs keyType = iota
> ```

These are never used outside the `drkey` package?
___
*[`pkg/drkey/protocol.go` line 51 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KmD134JiLnNSD0KsE:-N5KmD134JiLnNSD0KsF:undefined) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L51)):*
> ```Go
> 
> // packtoHostAddr returns a HostAddr parsing a given address in string format.
> func packtoHostAddr(host string) (hostAddr, error) {
> ```





_Suggestion:_
```Go
hostAddrFromString
```
___
*[`pkg/drkey/protocol.go` line 81 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KmKZ66xEJCM_GgvgS:-N5KmKZ66xEJCM_GgvgT:undefined) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L81)):*
> ```Go
> 
> // AddrToString returns the string representation of the HostAddr.
> func (h *hostAddr) AddrToString() string {
> ```





_Suggestion:_
```Go
String()
```
___
*[`pkg/drkey/protocol.go` line 102 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KmNb049K57Mh-uQxr:-N5KmNb049K57Mh-uQxs:b-gtndqk) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L102)):*
> ```Go
> }
> 
> func inputDeriveHostToHost(input []byte, host hostAddr) int {
> ```

Reference the specification.



_Suggestion:_
```Go
deriveHostToHostInput
```
___
*[`pkg/drkey/protocol.go` line 122 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KmZGW86J35s2QQ3zT:-N5KmZGW86J35s2QQ3zU:b1qy7qy) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L122)):*
> ```Go
> // DeriveKey derives the following key given an input and a higher-level key.
> // The input buffer is overwritten.
> func deriveKey(input []byte, upKey Key) (Key, error) {
> ```

Reference the specification
___
*[`pkg/drkey/protocol.go` line 122 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KmkwUC4Ygd_5OBdok:-N5KmkwUC4Ygd_5OBdol:undefined) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L122)):*
> ```Go
> // DeriveKey derives the following key given an input and a higher-level key.
> // The input buffer is overwritten.
> func deriveKey(input []byte, upKey Key) (Key, error) {
> ```





_Suggestion:_
```Go
upperLevelKey
```
___
*[`pkg/drkey/protocol_test.go` line 28 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kn9gYBGpe5exEjXHo:-N5Kn9gYBGpe5exEjXHp:b-rwk1gf) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol_test.go#L28)):*
> ```Go
> )
> 
> func TestDeriveASHostGeneric(t *testing.T) {
> ```

The tests for the generic derivation should go to the `generic` package.
___
*[`pkg/drkey/protocol_test.go` line 60 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kmx620ftCKV3JKoeQ:-N5Kmx620ftCKV3JKoeR:btb92zw) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol_test.go#L60)):*
> ```Go
> 			},
> 			assertFormatErr: assert.NoError,
> 			expectedKey: drkey.Key{0x2b, 0xc8, 0xfb, 0xb9, 0x45, 0x55,
> ```

How did you get this?
Can you create a golden file tests for all of these such that these values could be easily updated?
___
*[`pkg/drkey/protocol_test.go` line 173 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KnGAKCBHY-O0NyPDh:-N5KnGAKCBHY-O0NyPDi:b-tf11nf) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol_test.go#L173)):*
> ```Go
> }
> 
> func TestDeriveASHostSpecific(t *testing.T) {
> ```

The tests for the specific derivation should go to the `specific` package.
___
*[`pkg/drkey/secret_value_test.go` line 28 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KoFL388hZM3xnD6iD:-N5KoFL388hZM3xnD6iE:b-k4df98) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/secret_value_test.go#L28)):*
> ```Go
> 
> 	asSecret := []byte{0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7}
> 	targetKey := drkey.Key{0xc3, 0xb4, 0x84, 0x3c, 0x99, 0x3,
> ```

Golden file test
___
*[`pkg/drkey/specific.go` line 26 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KoKRD11_mNVGkk8CC:-N5KoKRE-4TgRHoNe5p3:bf6we3o) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/specific.go#L26)):*
> ```Go
> 
> // SpecificDeriver implements the specific drkey derivation.
> type SpecificDeriver struct {
> ```

Similar comments apply to this file as for the `generic.go` file.
___
*[`private/drkey/protocoltest/commons.go` line 15 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Ko_vv2BniDRSdSyT7:-N5Ko_vwAGs893p6bkPp:b-jd5mbr) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/private/drkey/protocoltest/commons.go#L15)):*
> ```Go
> 
> package protocoltest
> ```

What are these used for?
___
*[`proto/drkey/mgmt/v1/mgmt.proto` line 19 at r1](https://reviewable.io/reviews/scionproto/scion/4217#-N4c8a2K-1Mx_LkTvSxa:-N5Kph5W0atbMzA9m5j4:b-7v9xf2) ([raw file](https://github.com/scionproto/scion/blob/aec8932ba676d73ad5e8ba3877ba16f95be94809/proto/drkey/mgmt/v1/mgmt.proto#L19)):*

Yes you're right regarding the SV and the Lvl1Request, but as you said some other request/response messages are shared. Personally, I find it more organized to have the drkey messages in this package, especially, because we avoid this duplicity that you mentioned. However, we already discussed about PB best practices, recommending not to use messages among several packages, so if we decide to go for this approach. Note that the IntraLvl1 message might be duplicated as well upon it's being use by other services (e.g. COLIBRI) (i.e. ASHost, HostAS and IntraLvl1).

When we agree on the previous, I'll remove the mgmt subpackage, in the same pass.
</blockquote></details>

I agree with @matzf , however, we also need to have a unit test that ensures that the two definitions of `ASHost` and `HostAS` protobufs are equivalent (and stay equivalent).
___
*[`proto/drkey/v1/mgmt.proto` line 1 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kq3Tt2X0vHAdxLqeK:-N5Kq3Tt2X0vHAdxLqeL:b-egj13w) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/proto/drkey/v1/mgmt.proto#L1)):*
> ```protobuf
> ```

Please name the file `drkey.proto`
___
*[`proto/drkey/v1/mgmt.proto` line 1 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kq9se5VsSUYevWZAc:-N5Kq9se5VsSUYevWZAd:b-r9py4d) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/proto/drkey/v1/mgmt.proto#L1)):*
> ```protobuf
> ```

ETH Zurich



_Code quote:_
```protobuf
Anapaya Systems
```
___
*[`proto/drkey/v1/mgmt.proto` line 154 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kp-wE5wugFlPaGYvU:-N5Kp-wE5wugFlPaGYvV:bz9wnla) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/proto/drkey/v1/mgmt.proto#L154)):*
> ```protobuf
>     PROTOCOL_SCMP = 1;
>     // DNS protocol
>     PROTOCOL_DNS = 2;
> ```

Let us only include `Generic` and `SCMP` for now.

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4217)* status: 1 of 35 files reviewed, 28 unresolved discussions (waiting on @matzf and @shitz)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4217#-N5KtjzpBfPAzLQJu-V7:-N5dPq_J2-FyN6gIajyr:b-896fix) (no related file):*

Please make the PR title more descriptive. Something like
"drkey: add main interfaces and protobuf definitions"

Also 1/n is not really useful if we don't know what n is.
</blockquote></details>

Done.

___
___
*[`pkg/drkey/db.go` line 26 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KZWGo4buxCvjc3S32:-N5dPuKT23g9Sp5joF-r:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/db.go#L26)):*
> ```Go
> )
> 
> var ErrKeyNotFound = serrors.New("Key not found")
> ```

Done.
___
*[`pkg/drkey/db.go` line 30 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KZd_VC4MeQzE9p7pb:-N5dPvwx0pimJXkN_4v3:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/db.go#L30)):*
> ```Go
> // SecretValueDB is the database for Secret Values.
> type SecretValueDB interface {
> 	GetSV(ctx context.Context, meta SVMeta) (SV, error)
> ```

Done.
___
*[`pkg/drkey/db.go` line 31 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KZiPh8jfPu-3D9-hB:-N5dPwPr90mhDbPQtedg:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/db.go#L31)):*
> ```Go
> type SecretValueDB interface {
> 	GetSV(ctx context.Context, meta SVMeta) (SV, error)
> 	InsertSV(ctx context.Context, key SV) error
> ```

Done.
___
*[`pkg/drkey/db.go` line 32 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KZm251-J33UdbVtKI:-N5dPwpJ6TmjFNWPa-4m:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/db.go#L32)):*
> ```Go
> 	GetSV(ctx context.Context, meta SVMeta) (SV, error)
> 	InsertSV(ctx context.Context, key SV) error
> 	DeleteExpiredSV(ctx context.Context, cutoff time.Time) (int64, error)
> ```

Done.
___
*[`pkg/drkey/drkey.go` line 89 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KbRt6BVqmhkjpD9kL:-N5dPyRX6N4eHP2wBT5z:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/drkey.go#L89)):*
> ```Go
> 
> // SVMeta represents the information about a DRKey secret value.
> type SVMeta struct {
> ```

Done.
___
*[`pkg/drkey/drkey.go` line 95 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KbPM68htPCn8fMubl:-N5dPytTBsdJsqPJCZcl:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/drkey.go#L95)):*
> ```Go
> 
> // SV represents a DRKey secret value.
> type SV struct {
> ```

Done.
___
*[`pkg/drkey/drkey.go` line 131 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KbxnR1CtGslXCkFa9:-N5dPzl42dIZ7oR6X3Am:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/drkey.go#L131)):*

Replace all occurrences of `Lvl1` with `Level1`
</blockquote></details>

Done.
___
*[`pkg/drkey/protocol.go` line 33 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Km6Xz1ylsKT-ZdR1o:-N5dQ1P1BnhVMVBrZskF:b8xjrll) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L33)):*

These are never used outside the `drkey` package?
</blockquote></details>

Now they are used by the `generic` and `specific` packages.
___
*[`pkg/drkey/protocol.go` line 51 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KmD134JiLnNSD0KsE:-N5dQ6xF0MwqmUePpesR:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L51)):*
> ```Go
> 
> // packtoHostAddr returns a HostAddr parsing a given address in string format.
> func packtoHostAddr(host string) (hostAddr, error) {
> ```

Done.
___
*[`pkg/drkey/protocol.go` line 81 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KmKZ66xEJCM_GgvgS:-N5dQ7_YAWn5pHzEQDV4:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L81)):*
> ```Go
> 
> // AddrToString returns the string representation of the HostAddr.
> func (h *hostAddr) AddrToString() string {
> ```

Done.
___
*[`pkg/drkey/protocol.go` line 102 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KmNb049K57Mh-uQxr:-N5dQ8yz5yVqGwzesIab:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L102)):*

Reference the specification.
</blockquote></details>

Done.
___
*[`pkg/drkey/protocol.go` line 122 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KmZGW86J35s2QQ3zT:-N5dQ9h0CvlOS3sjTZTz:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L122)):*

Reference the specification
</blockquote></details>

Done.
___
*[`pkg/drkey/protocol.go` line 122 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KmkwUC4Ygd_5OBdok:-N5dQA6a2WPXSi0JOaaT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol.go#L122)):*
> ```Go
> // DeriveKey derives the following key given an input and a higher-level key.
> // The input buffer is overwritten.
> func deriveKey(input []byte, upKey Key) (Key, error) {
> ```

Done.
___
*[`pkg/drkey/secret_value_test.go` line 28 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KoFL388hZM3xnD6iD:-N5dQCVV96GEwfPCIB9i:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/secret_value_test.go#L28)):*

Golden file test
</blockquote></details>

Done.
___
*[`private/drkey/protocoltest/commons.go` line 15 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Ko_vv2BniDRSdSyT7:-N5dR9g--j_NKprG87fN:b-nbivcf) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/private/drkey/protocoltest/commons.go#L15)):*

What are these used for?
</blockquote></details>

These are common test helpers used in the `pkg/drkey` and are also used in the `dbtest` package that will be added in future PRs.
___
*[`pkg/drkey/protocol_test.go` line 28 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kn9gYBGpe5exEjXHo:-N5dTrGz3QdzVaGU2Lew:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol_test.go#L28)):*

The tests for the generic derivation should go to the `generic` package.
</blockquote></details>

Done.
___
*[`pkg/drkey/protocol_test.go` line 60 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kmx620ftCKV3JKoeQ:-N5dTqq3AvCFW2G2tk-N:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol_test.go#L60)):*

How did you get this?
Can you create a golden file tests for all of these such that these values could be easily updated?
</blockquote></details>

Done.
___
*[`pkg/drkey/protocol_test.go` line 173 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KnGAKCBHY-O0NyPDh:-N5dTrds1haoZaCj0zGm:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/protocol_test.go#L173)):*

The tests for the specific derivation should go to the `specific` package.
</blockquote></details>

Done.
___
*[`pkg/drkey/generic.go` line 24 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KlD7_8QF2nV5r364j:-N5dQEihEwe32YVK3K-f:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/generic.go#L24)):*

I'd put this in a separate package `drkey/generic`.
</blockquote></details>

Done.
___
*[`pkg/drkey/generic.go` line 25 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KkG1P3yq8oTilnLcs:-N5dQFCD2Nw0OGidjwKj:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/generic.go#L25)):*

What is `buf`?
</blockquote></details>

Done.
___
*[`pkg/drkey/generic.go` line 28 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KkU2b7bXF49JjAMh7:-N5dQGKv4zUnkNFBVi6I:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/generic.go#L28)):*

Reformat
</blockquote></details>

Done.
___
*[`pkg/drkey/generic.go` line 28 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KkeMe200oZVXqFU6p:-N5dQHuT260HyKm18zK3:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/generic.go#L28)):*

What does this method do?

In general, it would be good to have pointers to the DRKey specification on https://docs.scion.org/en/latest/cryptography/drkey.html
</blockquote></details>

Done.
___
*[`pkg/drkey/specific.go` line 26 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KoKRD11_mNVGkk8CC:-N5dQLyM-kPYCY_pBI5H:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/specific.go#L26)):*

Similar comments apply to this file as for the `generic.go` file.
</blockquote></details>

Done.
___
*[`proto/drkey/mgmt/v1/mgmt.proto` line 19 at r1](https://reviewable.io/reviews/scionproto/scion/4217#-N4c8a2K-1Mx_LkTvSxa:-N5dRY57-h2JLjH2UfOo:baxsyxx) ([raw file](https://github.com/scionproto/scion/blob/aec8932ba676d73ad5e8ba3877ba16f95be94809/proto/drkey/mgmt/v1/mgmt.proto#L19)):*

I agree with @matzf , however, we also need to have a unit test that ensures that the two definitions of `ASHost` and `HostAS` protobufs are equivalent (and stay equivalent).
</blockquote></details>

I added some simple UTs in `private/proto/drkey`  to compare the messages. I don't have much expertise with protobuf, if you think it is a more appropriate way to check that, please speak up :) 
___
*[`proto/drkey/v1/mgmt.proto` line 1 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kq9se5VsSUYevWZAc:-N5dRWKf-WrkDpCqCs2O:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/proto/drkey/v1/mgmt.proto#L1)):*

ETH Zurich
</blockquote></details>

Done.
___
*[`proto/drkey/v1/mgmt.proto` line 1 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kq3Tt2X0vHAdxLqeK:-N5dRWh40bqQ9PXgvnEd:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/proto/drkey/v1/mgmt.proto#L1)):*

Please name the file `drkey.proto`
</blockquote></details>

Done.
___
*[`proto/drkey/v1/mgmt.proto` line 154 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kp-wE5wugFlPaGYvU:-N5dSH8bE7g_qxisBwJy:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/proto/drkey/v1/mgmt.proto#L154)):*

Let us only include `Generic` and `SCMP` for now.
</blockquote></details>

Done.

shitz (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4217)* status: all files reviewed, 17 unresolved discussions (waiting on @JordiSubira)

___
*[`proto/drkey/mgmt/v1/mgmt.proto` line 19 at r1](https://reviewable.io/reviews/scionproto/scion/4217#-N4c8a2K-1Mx_LkTvSxa:-N5eIuHZ0G_miKzoO1z6:b-5l1pej) ([raw file](https://github.com/scionproto/scion/blob/aec8932ba676d73ad5e8ba3877ba16f95be94809/proto/drkey/mgmt/v1/mgmt.proto#L19)):*

I added some simple UTs in `private/proto/drkey`  to compare the messages. I don't have much expertise with protobuf, if you think it is a more appropriate way to check that, please speak up :) 
</blockquote></details>

I checked briefly, but didn't really find something that I could make work using https://pkg.go.dev/github.com/google/go-cmp/cmp. Maybe you want to look into that, because you might make it work by spending a bit more time. With that being said, while the current test doesn't guarantee equivalence, I think it's good enough for our purposes.

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4217)* status: 10 of 35 files reviewed, 18 unresolved discussions (waiting on @shitz)

___
*[`pkg/drkey/db.go` line 49 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e-EfnBZaaeA4FtzlW:-N5jrwQZ3vseOZ4OY9Xy:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/db.go#L49)):*
> ```Go
> 
> // Lvl2DB is the drkey database interface for end-host keys.
> type Lvl2DB interface {
> ```

Done.
___
*[`pkg/drkey/protocol.go` line 102 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e-YWiC64oqw_ug24V:-N5jsunn9kf3QZkn7S1j:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/protocol.go#L102)):*
> ```Go
> }
> 
> // SerializeHostToHostInput serializes the input for deriving a HosToHost key,
> ```

Done.
___
*[`pkg/drkey/protocol.go` line 124 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e-d6M9ly0Yd16ke7R:-N5js0ga2mHUu8A0Xs2z:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/protocol.go#L124)):*

line length
</blockquote></details>

Done.
___
*[`pkg/drkey/secret_value_test.go` line 41 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e2hji7QcjdTGhv7Td:-N5js1QV17yLtDPFjENm:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/secret_value_test.go#L41)):*

Use `hex.EncodeToString` and `hex.DecodeString`
</blockquote></details>

Done.
___
*[`pkg/drkey/generic/generic_test.go` line 79 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e1cAo6A4r3GVP3oc7:-N5jt-t4-0XV8ykho4pK:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/generic/generic_test.go#L79)):*

Use `hex.EncodeToString` and `hex.DecodeString`
</blockquote></details>

Done.
___
*[`pkg/drkey/specific/specific_test.go` line 99 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e2EuBALAQiFWi4Enr:-N5jt65o5aS-01j0BKhU:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/specific/specific_test.go#L99)):*

`expectedKey` is not needed anymore.
</blockquote></details>

Done.
___
*[`pkg/drkey/specific/specific_test.go` line 129 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e2NcvAlujw1lkzEkV:-N5jt77PBeynBzrDNo_R:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/specific/specific_test.go#L129)):*

Use `hex.EncodeToString` and `hex.DecodeString`
</blockquote></details>

Done.
___
*[`private/proto/drkey/eq_definiton_test.go` line 13 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e3QqL6Ue_mMjPjzCW:-N5jtD0H9EyVP0xd23vB:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/private/proto/drkey/eq_definiton_test.go#L13)):*

add blank line
</blockquote></details>

Done.
___
*[`private/proto/drkey/eq_definiton_test.go` line 52 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5eIXkn1BZTEErWQMWi:-N5jtDiB28WCIZNrKK0a:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/private/proto/drkey/eq_definiton_test.go#L52)):*

I think you also want to compare `TextName()`.
</blockquote></details>

Done.
___
*[`private/proto/drkey/eq_definiton_test.go` line 53 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e47Sr11L4fOzxdzrP:-N5jtEcPDKqBVFQmMsRV:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/private/proto/drkey/eq_definiton_test.go#L53)):*

You want to replace all `require` with `assert` here in the loop. Require will make the test fail immediately, while assert will continue and output all errors. Note, you need the require when testing the lengths, because it doesn't make sense to continue if the lengths are not equal.
</blockquote></details>

Done.
___
*[`pkg/drkey/specific.go` line 26 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5KoKRD11_mNVGkk8CC:-N5jt2wo6AG4EjXr5sv_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/pkg/drkey/specific.go#L26)):*

Please also update the order of methods/types etc according to the comment on `generic.go`.
</blockquote></details>

Done.
___
*[`pkg/drkey/generic/generic.go` line 26 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e00kx0zzcBuD1kKWI:-N5js3R5AgJH-EX9LVOa:bkujrmm) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/generic/generic.go#L26)):*

Seems to me that this could now be simply a package variable. With that being sad, it smells like premature optimization and thus I'm not the biggest fan of having this at all. I'd simply remove it.
</blockquote></details>

I guess that by making it a package variable, we should treat it in some way that we do not incur in race conditions if more than one Deriver is using it. 

Indeed it is an optimization to avoid unnecessary memory allocations. Although I think it is not adding too much complexity, if'd rather remove it, I can do it in the next pass.
___
*[`pkg/drkey/generic/generic.go` line 33 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e0u7w5CS1CjolI5hd:-N5jrfDr9NF1Sn1OX4pB:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/generic/generic.go#L33)):*

line length
</blockquote></details>

Done.
___
*[`pkg/drkey/generic/generic.go` line 34 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e0HT81-v5z_6ZBAyZ:-N5jrfYqCkcl-cm6LPRb:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/generic/generic.go#L34)):*

Put private functions after the public ones. In general, try to order the different entities in a file according to their importance:
* package constants and variables
* exported interface definitions
* exported functions
* for each exported type: type definition, public methods, private methods
* private functions and types
</blockquote></details>

Done.
___
*[`pkg/drkey/specific/specific.go` line 40 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e1yrhAroTV41YjpvO:-N5jt4CP8z3TP1kyJqt4:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/specific/specific.go#L40)):*

line length
</blockquote></details>

Done.
___
*[`private/drkey/protocoltest/commons.go` line 15 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Ko_vv2BniDRSdSyT7:-N5jtC0g2cELBrxnbAuu:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/private/drkey/protocoltest/commons.go#L15)):*

Would call the package `drkey/drkeytest` in that case (much like `http` and `http/httptest` in the standard Go library)
</blockquote></details>

Done.
___
*[`proto/control_plane/v1/drkey.proto` line 24 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5jjbSnAcJNthq6KKWU:-N5jtrO93d9fAC3w4rpX:beqkbz8) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/proto/control_plane/v1/drkey.proto#L24)):*

Actually, I think have a `DRKey` prefix for all these messages would improve overall clarity. `cp_pb.SecretValueRequest` is not very informative while `cp_pb.DRKeySecretValueRequest` immediately tells everyone that this is a DRKey protocol message.

Same for the protos in `daemon`.
</blockquote></details>

You're right, those should be renamed now. Done!
___
*[`proto/drkey/mgmt/v1/mgmt.proto` line 19 at r1](https://reviewable.io/reviews/scionproto/scion/4217#-N4c8a2K-1Mx_LkTvSxa:-N5jtGOY2ipUyFB8sftS:bohk2ri) ([raw file](https://github.com/scionproto/scion/blob/aec8932ba676d73ad5e8ba3877ba16f95be94809/proto/drkey/mgmt/v1/mgmt.proto#L19)):*

I checked briefly, but didn't really find something that I could make work using https://pkg.go.dev/github.com/google/go-cmp/cmp. Maybe you want to look into that, because you might make it work by spending a bit more time. With that being said, while the current test doesn't guarantee equivalence, I think it's good enough for our purposes.
</blockquote></details>

As discussed offline, the current implementation is likely to identify divergences between both protobuf definitions. In the future, we might investigate whether we get something leveraging https://pkg.go.dev/github.com/google/go-cmp/cmp and https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp
___
*[`proto/drkey/v1/mgmt.proto` line 1 at r2](https://reviewable.io/reviews/scionproto/scion/4217#-N5Kq9se5VsSUYevWZAc:-N5jtFY1AWvaVRwb5aq-:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d9c0af77bff293554574a287bfe7438d32142ee9/proto/drkey/v1/mgmt.proto#L1)):*

We use ETH ZUrich (not ZÜrich)
</blockquote></details>

Done.

shitz (changes requested):
> Reviewed 25 of 25 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4217)* status: all files reviewed, 2 unresolved discussions (waiting on @JordiSubira)

___
*[`pkg/drkey/generic/generic.go` line 26 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e00kx0zzcBuD1kKWI:-N5sHhH13dZvhM47sei-:b-k3hxq1) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/generic/generic.go#L26)):*
> I guess that by making it a package variable, we should treat it in some way that we do not incur in race conditions if more than one Deriver is using it.

Yes that's definitely true. I somehow assumed wrongly that the deriver would be a singleton.

I'd prefer removing it. It could lead to subtle bugs if there is some stale data in the buffer that gets accidently used. Also, there is almost never a reason to use a fixed-size array over a slice.
___
*[`proto/control_plane/v1/drkey.proto` line 123 at r4](https://reviewable.io/reviews/scionproto/scion/4217#-N5sMfU7Ct2Wrl_J0RE9:-N5sMfU8EwpCZTLbnwrY:b-1m9ux7) ([raw file](https://github.com/scionproto/scion/blob/32e8fc98b0b201ae19ff8fcad8ef16863bf7ffd1/proto/control_plane/v1/drkey.proto#L123)):*
> ```protobuf
>     // Level2 key.
>     bytes key = 3;
> }
> ```

missing newline

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4217)* status: 32 of 35 files reviewed, 2 unresolved discussions (waiting on @shitz)

___
*[`pkg/drkey/generic/generic.go` line 26 at r3](https://reviewable.io/reviews/scionproto/scion/4217#-N5e00kx0zzcBuD1kKWI:-N5tq4A76DmZSuk10VRi:b-73fo7z) ([raw file](https://github.com/scionproto/scion/blob/c8fe2b4c2dcc2b6ed7419ba0a19cd367844a3cca/pkg/drkey/generic/generic.go#L26)):*

> I guess that by making it a package variable, we should treat it in some way that we do not incur in race conditions if more than one Deriver is using it.

Yes that's definitely true. I somehow assumed wrongly that the deriver would be a singleton.

I'd prefer removing it. It could lead to subtle bugs if there is some stale data in the buffer that gets accidently used. Also, there is almost never a reason to use a fixed-size array over a slice.
</blockquote></details>

I removed it and add byte slices within the derivation functions. We can always revisit it and include this internal buffer in the future.
___
*[`proto/control_plane/v1/drkey.proto` line 123 at r4](https://reviewable.io/reviews/scionproto/scion/4217#-N5sMfU7Ct2Wrl_J0RE9:-N5tqVg-1JtCiaCzjVNM:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/32e8fc98b0b201ae19ff8fcad8ef16863bf7ffd1/proto/control_plane/v1/drkey.proto#L123)):*

missing newline
</blockquote></details>

Done.

shitz (approved):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4217)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JordiSubira)

shitz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4217#-:-N5uDKvP6z0M98Jihj-z:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JordiSubira)

PR #4218, opened 2022-06-15
close reliable conn before and after stopping reconnecer

uniquefine:
When a reliable `PacketConn` is closed it first stops the `reconnector` and then closes the current connection. 
To make sure all connections are closed, close the current connection before AND after closing the `reconnector`.

This fixes an issue where `ReadFrom`/`WriteTo` calls on the connection do not unblock when the connection is closed.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4218)

### DIFF
--- a/pkg/sock/reliable/reconnect/conn.go
+++ b/pkg/sock/reliable/reconnect/conn.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/scionproto/scion/pkg/log"
+	"github.com/scionproto/scion/pkg/private/serrors"
 	"github.com/scionproto/scion/pkg/sock/reliable"
 )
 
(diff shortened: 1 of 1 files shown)


matzf (commented):
Thank you.

I'm trying to understand the situation that triggers this issue. My understanding is this; the symptom that we're fixing is that an ongoing reconnect attempt sets a new connection, leaving the old connection un-closed. This reconnect attempt was triggered by some erroring IO operation, but a different IO operation is left blocked with the old connection.
Is this roughly the right thing here? Is this specific to `Close`-ing the connection, or could this also occur during any other error condition leading to a reconnect? I'm wondering whether it would make sense to always explicitly close the old dispatcher conn before any reconnect attempt ( `conn.getConn().Close()` at the beginning of `asyncReconnectWrapper`). 

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4218#-:-N4b4yGi7gyioQh2BRNj:b-sfpm97)* status: 0 of 1 files reviewed, all discussions resolved

lukedirtwalker:
Actually we realized that this should be a non-issue. Somehow close didn't work in the (our) Gateway as a termination signal for probe connections. This was kind of the only place I could see which could be the issue, but it turned out it was something completely unrelated. Just looking at the code I can't really see how this would block, therefore closing this for now.

PR #4219, opened 2022-07-01
scion-pki: print verification error

oncilla:
Print the verification error when certificate verification fails.
Before, there was a bug that would always print the nil error that was previously defined.


shitz (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4219)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

PR #4228, opened 2022-07-20
dispatcher: set large send buffer size on underlay socket

oncilla:
Set a large send buffer size on the underlay socket in the dispatcher.
This is done to tolerate unreachable next hop routers that do not respond
to ARP requests.

Without this change, the send buffer is the same size as the unresolved
queue (packets with destinations that have not been resolved via ARP yet).
As a result, the send buffer could be filled by these packets, and legitimate
packets would get dropped.

---
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4228)


sustrik (commented):
This will only cause the problem to hit later on, no? Is the idea to delay the occurrence of the problem long enough so that the pending packets expire before it hits?

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4228#-:-N7PzXUl0IA6fCIe-bCo:b-af0k8y)* status: 0 of 2 files reviewed, all discussions resolved

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4228)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

PR #4236, opened 2022-07-28
grpc: remove roundrobin dependency from dialer.go

kmavromati:
This PR removes the dependency to the `grpc/balancer/roundrobin` package.

The use of the `roundrobin` package is deprecated and is removed in a future version of the `grpc` package.


kmavromati:
PR replaced by [4237](https://github.com/scionproto/scion/pull/4237)

Issue #4241, opened 2022-08-03
Memory safety issue when reversing an empty `scion.Raw`


jcp19:
While verifying SCION with Gobra, I ran into what might be a bug in the method `Raw.Reverse` in the package `github.com/scionproto/scion/pkg/slayers/path/scion`. Basically, the implementation of `Reverse` starts by converting `s` to an element of type `Decoded`, and then calls `Reverse` on the var `decoded`.
```go
func (s *Raw) Reverse() (p path.Path, err error) {
	// XXX(shitz): The current implementation is not the most performant, since it parses the entire
	// path first. If this becomes a performance bottleneck, the implementation should be changed to
	// work directly on the raw representation.

	decoded, err := s.ToDecoded()
	if err != nil {
		return nil, err
	}
	reversed, err := decoded.Reverse()
	if err != nil {
		return nil, err
	}
        if err := reversed.SerializeTo(s.Raw); err != nil {
		return nil, err
	}
        ...
}
```
There is a problem when `s` is empty (i.e., `s.NumINF == 0`) though - converting it to `Decoded` and calling `Reverse` on it always returns a `nil` of the interface type `path.Path`:
```go
func (s *Decoded) Reverse() (p path.Path, r error) {
	if s.NumINF == 0 {
		// Empty path doesn't need reversal.
		return nil, nil
	}
        ...
}
```
In this case, the call to `decoded.Reverse()` in `Raw.Reverse` will succeed and `reversed` and `err` will both be `nil`. Finally, the call to `reversed.SerializeTo` will panic, because the receiver is `nil` of the interface type, and thus, it cannot successfully dispatch the method call.

I wrote a test case in the file `github.com/scionproto/scion/pkg/slayers/path/scion/raw_test.go` which triggers this problem:
```go
var emptyRawTestPath = &scion.Raw{
	Base: scion.Base{
		PathMeta: scion.MetaHdr{
			CurrINF: 0,
			CurrHF:  0,
			SegLen:  [3]uint8{0, 0, 0},
		},
		NumINF:  0,
		NumHops: 0,
	},
	Raw: make([]byte, scion.MetaLen),
}

func TestReverseEmpty(t *testing.T) {
	_, err := emptyRawTestPath.Reverse()
	assert.NoError(t, err)
}
```
I believe that the easiest way to address this bug is by changing `Decoded.Reverse` to return `empty.Path{}, nil` when the receiver is empty. I can open a PR if you agree :)

PR #4242, opened 2022-08-03
Return an error when reversing an empty SCION path

jcp19:
Fixes issue #4241.

I introduced a separate test `TestEmptyRawReverse` instead of adding an extra case to `TestRawReverse` because the test cases of the latter are shared with the test for reversing `Decoded` paths, which expects that reversing a `Decoded` yields a `Decoded`. Putting everything in the same test can be done with a bigger change-set though.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4242)


shitz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4242)* status: all files reviewed, 1 unresolved discussion (waiting on @jcp19)

___
*[`pkg/slayers/path/scion/decoded.go` line 98 at r1](https://reviewable.io/reviews/scionproto/scion/4242#-N96dpRb7Ym1sn21KS8G:-N96dpRb7Ym1sn21KS8H:beoy2zi) ([raw file](https://github.com/scionproto/scion/blob/c9061d18780856160493ba881c013bfb8b59fec5/pkg/slayers/path/scion/decoded.go#L98)):*
> ```Go
> 	if s.NumINF == 0 {
> 		// Empty path doesn't need reversal.
> 		return empty.Path{}, nil
> ```

This is not the proper fix. A decoded path that is "empty" is not valid and reversing it should return an error. For the empty path we explicitly have the empty path type.

jcp19 (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4242)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @shitz)

___
*[`pkg/slayers/path/scion/decoded.go` line 98 at r1](https://reviewable.io/reviews/scionproto/scion/4242#-N96dpRb7Ym1sn21KS8G:-N96z8i63yw2aJMH-h4R:b-thfyhr) ([raw file](https://github.com/scionproto/scion/blob/3957386b60c7ca901da1005e19a08405af9a79c9/pkg/slayers/path/scion/decoded.go#L98)):*

This is not the proper fix. A decoded path that is "empty" is not valid and reversing it should return an error. For the empty path we explicitly have the empty path type.
</blockquote></details>

I changed this to return an error message

shitz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4242)* status: all files reviewed, 1 unresolved discussion (waiting on @jcp19)

___
*[`pkg/slayers/path/scion/raw_test.go` line 91 at r1](https://reviewable.io/reviews/scionproto/scion/4242#-N97IrhZ9A9prZ0hamSN:-N97IrhZ9A9prZ0hamSO:b-7ejxlw) ([raw file](https://github.com/scionproto/scion/blob/3957386b60c7ca901da1005e19a08405af9a79c9/pkg/slayers/path/scion/raw_test.go#L91)):*
> ```Go
> }
> 
> func TestEmptyRawReverse(t *testing.T) {
> ```

The same test is also needed for the decoded path (they internally use the same reverse logic, however, that's an implementation detail that might change at some point).

jcp19 (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4242)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @shitz)

___
*[`pkg/slayers/path/scion/raw_test.go` line 91 at r1](https://reviewable.io/reviews/scionproto/scion/4242#-N97IrhZ9A9prZ0hamSN:-N97S36i-cIhATCsN-20:b-896fja) ([raw file](https://github.com/scionproto/scion/blob/3957386b60c7ca901da1005e19a08405af9a79c9/pkg/slayers/path/scion/raw_test.go#L91)):*

The same test is also needed for the decoded path (they internally use the same reverse logic, however, that's an implementation detail that might change at some point).
</blockquote></details>

Done!

shitz (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4242)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jcp19)

PR #4243, opened 2022-08-04
Pass parameter by reference in `Path.Reverse` in package `onehop`

jcp19:
A call to `Reverse` cannot modify the receiver `o` (not directly inside the body of `Reverse` or indirectly via a call to `ToSCIONDecoded`, given that the later copies all fields that it requires from `o` to a new struct). As such, it seems like a good idea to pass this by reference, reducing the amount of copying needed.

Incidentally, making this change also allows us to reduce the verification time of this method in the context of the VerifiedSCION project.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4243)


shitz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4243#-:-N96hv3o0y0nMrkKJtVN:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jcp19)

PR #4250, opened 2022-09-08
cmd/scion: set payload size appropriately if --max-mtu is set

bunert:
Correct the behavior of the `--max-mtu` flag for `scion ping`.

Previously, there was a bug where the calculated payload size was ignored and the value from `--payload-size` was taken.
In most cases, this would be the default.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4250)


oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4250)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @bunert)

PR #4254, opened 2022-09-13
pkg/snet: add URI scion address encoding

bunert:
pkg/snet: add URI scion address encoding

When converting an address string to a SCION address, the URI scion address encoding is now tried first, while the old parser is still supported for backward compatibility.

### DIFF
--- a/pkg/snet/udpaddr.go
+++ b/pkg/snet/udpaddr.go
@@ -26,8 +26,8 @@ import (
 	"github.com/scionproto/scion/pkg/private/serrors"
 )
 
-var addrRegexp = regexp.MustCompile(`^(?P<ia>\d+-[\d:A-Fa-f]+),(?P<host>.+)$`)
-var rfcAddrRegexp = regexp.MustCompile(`^[\[](?P<ia>\d+-[\d:A-Fa-f]+),(?P<host>.+)[\]]:(?P<port>.+)$`)
+var addrRegexpLegacy = regexp.MustCompile(`^(?P<ia>\d+-[\d:A-Fa-f]+),(?P<host>.+)$`)
+var addrRegexp = regexp.MustCompile(`^[\[](?P<ia>\d+-[\d:A-Fa-f]+),(?P<host>.+)[\]]:(?P<port>.+)$`)
 
 // UDPAddr to be used when UDP host.
 type UDPAddr struct {
@@ -165,7 +165,7 @@ func CopyUDPAddr(a *net.UDPAddr) *net.UDPAddr {
 }
 
 func parseAddr(s string) (string, string, error) {
-	match := addrRegexp.FindStringSubmatch(s)
+	match := addrRegexpLegacy.FindStringSubmatch(s)
 	if len(match) != 3 {
 		return "", "", serrors.New("invalid address: regex match failed", "addr", s)
 	}
@@ -180,7 +180,7 @@ func parseAddr(s string) (string, string, error) {
 }
 
 func parseScionAddr(s string) (string, string, error) {
-	match := rfcAddrRegexp.FindStringSubmatch(s)
+	match := addrRegexp.FindStringSubmatch(s)
 	if len(match) != 4 {
 		return "", "", serrors.New("invalid address: regex match failed", "addr", s)
 	}
--- a/pkg/snet/udpaddr.go
+++ b/pkg/snet/udpaddr.go
@@ -27,6 +27,7 @@ import (
 )
 
 var addrRegexp = regexp.MustCompile(`^(?P<ia>\d+-[\d:A-Fa-f]+),(?P<host>.+)$`)
+var rfcAddrRegexp = regexp.MustCompile(`^[\[](?P<ia>\d+-[\d:A-Fa-f]+),(?P<host>.+)[\]]:(?P<port>.+)$`)
 
 // UDPAddr to be used when UDP host.
 type UDPAddr struct {
@@ -37,7 +38,43 @@ type UDPAddr struct {
 }
 
 // ParseUDPAddr converts an address string to a SCION address.
-// The supported formats are:
+func ParseUDPAddr(s string) (*UDPAddr, error) {
+	addr, err := parseUDPAddr(s)
+	if err != nil {
+		addr, err = parseUDPAddrLegacy(s)
+	}
+	return addr, err
+
+}
+
+// The supported formats are based on the extensions of RFC 3986:
+// https://scion.docs.anapaya.net/en/latest/uri.html#scion-udp
+//
+// Examples:
+//  - [isd-as,ipv4]:port       (e.g., [1-ff00:0:110,192.0.2.1]:80)
+//  - [isd-as,ipv6%zone]:port  (e.g., [1-ff00:0:110,2001:DB8::1]:80)
+func parseUDPAddr(s string) (*UDPAddr, error) {
+	rawIA, rawHost, err := parseScionAddr(s)
+	if err != nil {
+		return nil, err
+	}
+
+	ia, err := addr.ParseIA(rawIA)
+	if err != nil {
+		return nil, serrors.WrapStr("invalid address: IA not parsable", err, "ia", ia)
+	}
+
+	udp, err := net.ResolveUDPAddr("udp", rawHost)
+	if err != nil {
+		return nil, serrors.WrapStr("invalid address: host not parsable", err, "host", rawHost)
+	}
+	if udp.IP == nil {
+		return nil, serrors.WrapStr("invalid address: ip not specified", err, "host", rawHost)
+	}
+	return &UDPAddr{IA: ia, Host: udp}, nil
+}
+
+// The supported legacy formats are:
 //
 // Recommended:
 //  - isd-as,ipv4:port        (e.g., 1-ff00:0:300,192.168.1.1:8080)
@@ -56,7 +93,7 @@ type UDPAddr struct {
 // Not supported:
 //  - isd-as,ipv6:port    (caveat if ipv6:port builds a valid ipv6 address,
 //                         it will successfully parse as ipv6 without error)
-func ParseUDPAddr(s string) (*UDPAddr, error) {
+func parseUDPAddrLegacy(s string) (*UDPAddr, error) {
 	rawIA, rawHost, err := parseAddr(s)
 	if err != nil {
 		return nil, err
@@ -140,6 +177,20 @@ func parseAddr(s string) (string, string, error) {
 		return "", "", serrors.New("invalid address: trailing ':'", "addr", s)
 	}
 	return match[1], match[2], nil
+}
+
+func parseScionAddr(s string) (string, string, error) {
+	match := rfcAddrRegexp.FindStringSubmatch(s)
+	if len(match) != 4 {
+		return "", "", serrors.New("invalid address: regex match failed", "addr", s)
+	}
+	left, right := strings.Count(s, "["), strings.Count(s, "]")
+	if left != right {
+		return "", "", serrors.New("invalid address: bracket count mismatch", "addr", s)
+	}
+	var rawHost string = "[" + match[2] + "]:" + match[3]
+
+	return match[1], rawHost, nil
 }
 
 func ipOnly(s string) bool {
--- a/pkg/snet/udpaddr_test.go
+++ b/pkg/snet/udpaddr_test.go
@@ -157,6 +157,9 @@ func TestParseUDPAddr(t *testing.T) {
 		{address: "1-ff00:0:300,[1.2.3.4]:70000", isError: true},
 		{address: "1-ff00:0:300,[1.2.3.4]]", isError: true},
 		{address: "1-ff00:0:300,::1:60000", isError: true},
+		{address: "[1-ff00:0:110,192.0.2.1]:70000", isError: true},
+		{address: "[1-ff00:0:110,192.0.2.290]:80", isError: true},
+		{address: "[1-,127.0.0.1]:80", isError: true},
 		{address: "", isError: true},
 		{address: "1-ff00:0:300,[1.2.3.4]:80",
 			ia:   "1-ff00:0:300",
@@ -235,6 +238,11 @@ func TestParseUDPAddr(t *testing.T) {
 			ia:   "1-64496",
 			host: "2001:DB8::1",
 			port: 80,
+		},
+		{address: "[1-64496,2001:DB8::1]:60000",
+			ia:   "1-64496",
+			host: "2001:DB8::1",
+			port: 60000,
 		},
 	}
 	for _, test := range tests {
--- a/pkg/snet/udpaddr_test.go
+++ b/pkg/snet/udpaddr_test.go
@@ -221,6 +221,21 @@ func TestParseUDPAddr(t *testing.T) {
 			port: 0,
 			zone: "some-zone",
 		},
+		{address: "[1-ff00:0:110,192.0.2.1]:80",
+			ia:   "1-ff00:0:110",
+			host: "192.0.2.1",
+			port: 80,
+		},
+		{address: "[1-ff00:0:110,2001:DB8::1]:80",
+			ia:   "1-ff00:0:110",
+			host: "2001:DB8::1",
+			port: 80,
+		},
+		{address: "[1-64496,2001:DB8::1]:80",
+			ia:   "1-64496",
+			host: "2001:DB8::1",
+			port: 80,
+		},
 	}
 	for _, test := range tests {
 		t.Log(fmt.Sprintf("given address %q", test.address))


matzf (changes requested):
Thanks @bunert for this contribution.

Is there a specific reason to change this now? Yes, I would also have preferred this format, but it's _also_ not a proper URI and it's rejected by any URI parser except golang's (which seems to be somewhat more lenient than others). If we're just adding this as new variant because it's nicer, to me that feels like it's _only_ making things more complicated. 
FWIW, in cases where round-tripping a SCION address through golang's web stack is required, we've "just" been mangling the address into some internal form that doesn't 

If we decide to change the address format, we should perhaps discuss whether there are further changes to the address format that would simplify processing. In particular, should also get rid of the comma (e.g. replace it with a second dash)?

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254#-:-NBqxbSwBjZ8l8YSubHv:bxz8fah)* status: all files reviewed, 4 unresolved discussions (waiting on @bunert)

___
*[`pkg/snet/udpaddr.go` line 30 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBqzHIGEfB86oFmiEQ4:-NBqzHIGEfB86oFmiEQ5:b-qekhwc) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L30)):*
> ```Go
> 
> var addrRegexpLegacy = regexp.MustCompile(`^(?P<ia>\d+-[\d:A-Fa-f]+),(?P<host>.+)$`)
> var addrRegexp = regexp.MustCompile(`^[\[](?P<ia>\d+-[\d:A-Fa-f]+),(?P<host>.+)[\]]:(?P<port>.+)$`)
> ```

Simplify single-character-character-lists (`[\[]` === `\[`)



_Suggestion:_
```Go
^\[(?P<ia>\d+-[\d:A-Fa-f]+),(?P<host>.+)\]:(?P<port>.+)$
```
___
*[`pkg/snet/udpaddr.go` line 67 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBr-Rv50ncyw6aviBZy:-NBr-Rv50ncyw6aviBZz:bn2y7zi) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L67)):*
> ```Go
> 	}
> 
> 	udp, err := net.ResolveUDPAddr("udp", rawHost)
> ```

Use `netip.ParseAddr` to parse IP addresses or `netip.ParseAddrPort` to parse host:port. 
___
*[`pkg/snet/udpaddr.go` line 188 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBr-CCA8ykdVoBt3B07:-NBr-CCA8ykdVoBt3B08:bqe1u62) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L188)):*
> ```Go
> 	}
> 	left, right := strings.Count(s, "["), strings.Count(s, "]")
> 	if left != right {
> ```

Both should be equal to 1, the brackets are no longer optional.
___
*[`pkg/snet/udpaddr.go` line 191 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBr-M8S1FG32udO0xbr:-NBr-M8S1FG32udO0xbs:bbi595q) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L191)):*
> ```Go
> 		return "", "", serrors.New("invalid address: bracket count mismatch", "addr", s)
> 	}
> 	var rawHost string = "[" + match[2] + "]:" + match[3]
> ```

Just return the port string separately (as string or int)

bunert (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @bunert and @matzf)

___
*[`pkg/snet/udpaddr.go` line 30 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBqzHIGEfB86oFmiEQ4:-NBrROMw152YpKxUhuNj:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L30)):*

Simplify single-character-character-lists (`[\[]` === `\[`)
</blockquote></details>

Done.
___
*[`pkg/snet/udpaddr.go` line 67 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBr-Rv50ncyw6aviBZy:-NBrSE_o80UD0L2nofXs:b-9qkkbj) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L67)):*

Use `netip.ParseAddr` to parse IP addresses or `netip.ParseAddrPort` to parse host:port. 
</blockquote></details>

May I ask why? `snet.UDPAddr` contains a `*net.UDPAddr`, so it seemed more convenient to use the `net` pkg for parsing...
___
*[`pkg/snet/udpaddr.go` line 188 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBr-CCA8ykdVoBt3B07:-NBrRRhU3Q4zYp2srKNb:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L188)):*

Both should be equal to 1, the brackets are no longer optional.
</blockquote></details>

Done.

oncilla (commented):
@matzf 

> Is there a specific reason to change this now?

This was an open "good first issue" for a long time, and we finally wanted to address it. There is no particular reason to do it right now, other than preparing for cleaning up the parsing mess that is snet.UDPAddr.

> Yes, I would also have preferred this format, but it's also not a proper URI and it's rejected by any URI parser except golang's (which seems to be somewhat more lenient than others). 

While this is true, the parsers we have checked when writing https://scion.docs.anapaya.net/en/latest/uri.html (python/go) were both happy with it. The two alternatives to make it URI compatible are also listed in that document with the reasoning why they were not chosen. Of course we can reopen the discussion there.

> If we're just adding this as new variant because it's nicer, to me that feels like it's only making things more complicated.

To me the end goal is to get rid of all the variants that are not inline with https://scion.docs.anapaya.net/en/latest/uri.html.
The current code is a mess, I agree. If we change to the proposed encoding, there is exactly one way to specify the UDP address.

From
```go
//  - isd-as,ipv4:port        (e.g., 1-ff00:0:300,192.168.1.1:8080)
//  - isd-as,[ipv6]:port      (e.g., 1-ff00:0:300,[f00d::1337]:8080)
//  - isd-as,[ipv6%zone]:port (e.g., 1-ff00:0:300,[f00d::1337%zone]:8080)
//  - isd-as,[ipv4]:port (e.g., 1-ff00:0:300,[192.168.1.1]:8080)
//  - isd-as,[ipv4]      (e.g., 1-ff00:0:300,[192.168.1.1])
//  - isd-as,[ipv6]      (e.g., 1-ff00:0:300,[f00d::1337])
//  - isd-as,[ipv6%zone] (e.g., 1-ff00:0:300,[f00d::1337%zone])
//  - isd-as,ipv4        (e.g., 1-ff00:0:300,192.168.1.1)
//  - isd-as,ipv6        (e.g., 1-ff00:0:300,f00d::1337)
//  - isd-as,ipv6%zone   (e.g., 1-ff00:0:300,f00d::1337%zone)
```

To
```go
//  - [isd-as,ipv4]:port        (e.g., [1-ff00:0:300,192.168.1.1]:8080)
//  - [isd-as,ipv6]:port      (e.g., [1-ff00:0:300,f00d::1337]:8080)
//  - [isd-as,ipv6%zone]:port (e.g., [1-ff00:0:300,f00d::1337%zone]:8080)
```

For me, that is a big win. Especially when using tools like scion ping/traceroute.

> If we decide to change the address format, we should perhaps discuss whether there are further changes to the address format that would simplify processing. In particular, should also get rid of the comma (e.g. replace it with a second dash)?

Sure, we can discuss. But let's do that outside of this PR. I guess until this is decided we will need to freeze this PR.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254#-:-NBudjtP5EXg0P3qs0SM:b-z4zxv6)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @bunert and @matzf)

matzf (changes requested):
> [ ... ] preparing for cleaning up the parsing mess that is snet.UDPAddr.

Ok if that's the plan than I'm more than happy with this! :+1:  

> While this is true, the parsers we have checked when writing https://scion.docs.anapaya.net/en/latest/uri.html (python/go) were both happy with it

That's good, I was not aware that this works in python :thumbsup:

> For me, that is a big win. Especially when using tools like scion ping/traceroute.

These tools require host addresses only, not host:port UDP addresses. For this case, not including the brackets should still be possible, as described in the proposal document (the definition of `SCIONAddress` does not include the brackets).
To support the different use cases properly, I'd propose to add parsing functionality for `snet.SCIONAddress`, accepting the `SCIONAddress` format (i.e. no brackets, no port) (*).  This functionality is adopted eg. for ping/traceroute. Parsing a UDP address in the URI format then becomes `net.SplitHostPort`(**), followed by parsing the host part as SCIONAddress and creating a `snet.UDPAddr` by something like `host.WithPort(port)`.

If you think this deviates too much from what you had in mind, we should perhaps discuss the details in an issue.

(*) This is a functionality that I've been missing from `snet` anyway. Our `pan` library re-implements address parsing for contexts where the host:port format is illegal, https://github.com/netsec-ethz/scion-apps/blob/ef972274f4190de5ac702cc7eff579d936c6650d/pkg/pan/addr.go#L141-L189
(**) Thankfully, `net.SplitHostPort` works with the address format proposed here :tada: 

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254#-:-NBumKXUAJvM4IB8wUAP:b-okvchl)* status: all files reviewed, 2 unresolved discussions (waiting on @bunert)

___
*[`pkg/snet/udpaddr.go` line 67 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBr-Rv50ncyw6aviBZy:-NBuSLmn9VhZ6jPNMEbo:b-9fleue) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L67)):*
<details><summary><i>Previously, bunert wrote…</i></summary><blockquote>

May I ask why? `snet.UDPAddr` contains a `*net.UDPAddr`, so it seemed more convenient to use the `net` pkg for parsing...
</blockquote></details>

Ah sure, yes sorry for not giving any justification. In my view, `netip.ParseAddrPort` is preferable here because it only _parses_ addresses and never resolves names. Resolving names is not necessary (or even desirable) here and introduces additional failure modes.

In the longer run, we'll want to change our `snet.UDPAddr` to internally use `netip.Addr`, not only for performance reasons but also because the usability of the API is just far better. And btw., `netip`'s "predecessor" package, `inet.af/netaddr` is already used in this file.

oncilla (commented):
> These tools require host addresses only, not host:port UDP addresses. For this case, not including the brackets should still be possible, as described in the proposal document (the definition of SCIONAddress does not include the brackets).

Yes, you are absolutely right. `scion ping` was the wrong example to take here. I meant cases where you actually need a SCION UDP address.

> To support the different use cases properly, I'd propose to add parsing functionality for snet.SCIONAddress, accepting the SCIONAddress format (i.e. no brackets, no port) (*). This functionality is adopted eg. for ping/traceroute. Parsing a UDP address in the URI format then becomes net.SplitHostPort(**), followed by parsing the host part as SCIONAddress and creating a snet.UDPAddr by something like host.WithPort(port).

This is exactly what I had in mind. Eventually, snet.UDPAddr should no longer be able to parse addresses without a port.
(It is quite odd that we had this in the first place)

I want a strict separation between SCION UDP, SCION SVC and SCION Address parsing.

> If you think this deviates too much from what you had in mind, we should perhaps discuss the details in an issue.

I think we agree on the end state of what should be done. I wanted to go a non-breaking route where we still are able to parse the "legacy" way until other tools have been adapted. However, if you feel we can do a breaking change here, we can already clean this up now.

> (**) Thankfully, net.SplitHostPort works with the address format proposed here 🎉

It was written with net.SplitHostPort and net.JoinHostPort in mind. Glad that it still works after all this time :) 

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254#-:-NBvnII6B89eKVb9ZgRA:b6aauuq)* status: all files reviewed, 2 unresolved discussions (waiting on @bunert)

matzf (commented):
Excellent, that sounds like a good plan. :thumbsup: 

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254#-:-NBvsyfb6FM5GoBmdtoa:bf1m796)* status: all files reviewed, 3 unresolved discussions (waiting on @bunert)

___
*[`pkg/snet/udpaddr.go` line 77 at r2](https://reviewable.io/reviews/scionproto/scion/4254#-NBvtS5T9byp3o4KO-A7:-NBvtS5T9byp3o4KO-A8:b-lhlrjq) ([raw file](https://github.com/scionproto/scion/blob/4c83959b6cc66983a9330b4f3c2280aa6ad7d131/pkg/snet/udpaddr.go#L77)):*
> ```Go
> }
> 
> // The supported legacy formats are:
> ```

Nit, perhaps elaborate a bit what "legacy" means here and what the plan is for phasing out these formats.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254)* status: all files reviewed, 5 unresolved discussions (waiting on @bunert and @matzf)

___
*[`pkg/snet/udpaddr.go` line 67 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBr-Rv50ncyw6aviBZy:-NCOzRBg7BXhOQ7TwAEQ:b50of15) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L67)):*

Ah sure, yes sorry for not giving any justification. In my view, `netip.ParseAddrPort` is preferable here because it only _parses_ addresses and never resolves names. Resolving names is not necessary (or even desirable) here and introduces additional failure modes.

In the longer run, we'll want to change our `snet.UDPAddr` to internally use `netip.Addr`, not only for performance reasons but also because the usability of the API is just far better. And btw., `netip`'s "predecessor" package, `inet.af/netaddr` is already used in this file.
</blockquote></details>

+1 for using netip when necessary. 

However, I think parsing is a bit convoluted. Parsing the address can be boiled down to these steps:

1. net.SplitHostPort
2. parse port
3. split host at comma
4. parse IA
5. parse IP address

The beautiful thing is that this does not even require a regex.

outline https://go.dev/play/p/KCGrcLzUo2M
___
*[`pkg/snet/udpaddr.go` line 42 at r2](https://reviewable.io/reviews/scionproto/scion/4254#-NCOz8Mv7o_C3ZQUTt6X:-NCOz8Mv7o_C3ZQUTt6Y:b-53z4b1) ([raw file](https://github.com/scionproto/scion/blob/4c83959b6cc66983a9330b4f3c2280aa6ad7d131/pkg/snet/udpaddr.go#L42)):*
> ```Go
> // ParseUDPAddr converts an address string to a SCION address.
> func ParseUDPAddr(s string) (*UDPAddr, error) {
> 	addr, err := parseUDPAddr(s)
> ```

To reduce cognitive load when reading this code, you can make the exit conditions more unambiguous.



_Suggestion:_
```Go
	addr, err := parseUDPAddr(s)
	if err != nil {
		return parseUDPAddrLegacy(s)
	}
	return addr, nil
```
___
*[`pkg/snet/udpaddr.go` line 55 at r2](https://reviewable.io/reviews/scionproto/scion/4254#-NCOyuFN0YbC__rvXzJ8:-NCOyuFO8QZkWN7_sImN:bhmsgsh) ([raw file](https://github.com/scionproto/scion/blob/4c83959b6cc66983a9330b4f3c2280aa6ad7d131/pkg/snet/udpaddr.go#L55)):*
> ```Go
> // Examples:
> //  - [isd-as,ipv4]:port       (e.g., [1-ff00:0:110,192.0.2.1]:80)
> //  - [isd-as,ipv6%zone]:port  (e.g., [1-ff00:0:110,2001:DB8::1]:80)
> ```

The example does not match. No zone vs zone.



_Code quote:_
```Go
//  - [isd-as,ipv6%zone]:port  (e.g., [1-ff00:0:110,2001:DB8::1]:80)
```

bunert (commented):
I adjusted the requested changes. Should the parsing functionality for the `snet.SCIONAddress` be part of this PR or was it just a general discussion?

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254#-:-NDT4j-u-FHTGwNN5w-U:b-t2w747)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`pkg/snet/udpaddr.go` line 67 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBr-Rv50ncyw6aviBZy:-NDStP5VBSZOrVNocaz-:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L67)):*

+1 for using netip when necessary. 

However, I think parsing is a bit convoluted. Parsing the address can be boiled down to these steps:

1. net.SplitHostPort
2. parse port
3. split host at comma
4. parse IA
5. parse IP address

The beautiful thing is that this does not even require a regex.

outline https://go.dev/play/p/KCGrcLzUo2M





</blockquote></details>

Done.
___
*[`pkg/snet/udpaddr.go` line 191 at r1](https://reviewable.io/reviews/scionproto/scion/4254#-NBr-M8S1FG32udO0xbr:-NDStVhT-6P_fKI8iDdl:b-jwbfmq) ([raw file](https://github.com/scionproto/scion/blob/177fd0152b989ca100679026e9c44a70a3e1b5a8/pkg/snet/udpaddr.go#L191)):*

Just return the port string separately (as string or int)
</blockquote></details>

removed this part.
___
*[`pkg/snet/udpaddr.go` line 42 at r2](https://reviewable.io/reviews/scionproto/scion/4254#-NCOz8Mv7o_C3ZQUTt6X:-NDSo5Aj3MKLn-fAkboW:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/4c83959b6cc66983a9330b4f3c2280aa6ad7d131/pkg/snet/udpaddr.go#L42)):*

To reduce cognitive load when reading this code, you can make the exit conditions more unambiguous.
</blockquote></details>

Done.
___
*[`pkg/snet/udpaddr.go` line 55 at r2](https://reviewable.io/reviews/scionproto/scion/4254#-NCOyuFN0YbC__rvXzJ8:-NDSoJbn6OlSQ69O5nFU:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/4c83959b6cc66983a9330b4f3c2280aa6ad7d131/pkg/snet/udpaddr.go#L55)):*

The example does not match. No zone vs zone.
</blockquote></details>

Done.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254)* status: all files reviewed, 6 unresolved discussions (waiting on @bunert and @matzf)

___
*[`pkg/snet/udpaddr.go` line 42 at r2](https://reviewable.io/reviews/scionproto/scion/4254#-NCOz8Mv7o_C3ZQUTt6X:-NDTTuPI57gXK1JvWlqs:b-l9hbrj) ([raw file](https://github.com/scionproto/scion/blob/4c83959b6cc66983a9330b4f3c2280aa6ad7d131/pkg/snet/udpaddr.go#L42)):*
<details><summary><i>Previously, bunert wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Only half. Always explicitly return nil if there is no error.
___
*[`pkg/snet/udpaddr.go` line 63 at r3](https://reviewable.io/reviews/scionproto/scion/4254#-NDTUGGz0khucldnhiZ9:-NDTUGGz0khucldnhiZA:b5ij18d) ([raw file](https://github.com/scionproto/scion/blob/40b29028d75d1f6af2f7e94db9c78ec73b9435f5/pkg/snet/udpaddr.go#L63)):*
> ```Go
> 	}
> 	parts := strings.Split(host, ",")
> 	if len(parts) != 2 {
> ```

This is a new error. The wrapped error is nil here.



_Code quote:_
```Go
	if len(parts) != 2 {
		return nil, serrors.WrapStr("invalid address: host parts invalid", err, "host", host)
	}
```
___
*[`pkg/snet/udpaddr.go` line 71 at r3](https://reviewable.io/reviews/scionproto/scion/4254#-NDTUPsB9yom0TFoJqW3:-NDTUPsB9yom0TFoJqW4:undefined) ([raw file](https://github.com/scionproto/scion/blob/40b29028d75d1f6af2f7e94db9c78ec73b9435f5/pkg/snet/udpaddr.go#L71)):*
> ```Go
> 	}
> 	ip, err := netip.ParseAddr(parts[1])
> 	if len(parts) != 2 {
> ```





_Suggestion:_
```Go
if err != nil {
```
___
*[`pkg/snet/udpaddr_test.go` line 260 at r3](https://reviewable.io/reviews/scionproto/scion/4254#-NDTUjyD8ODy8wy3zrUD:-NDTUjyD8ODy8wy3zrUE:bmjzlbl) ([raw file](https://github.com/scionproto/scion/blob/40b29028d75d1f6af2f7e94db9c78ec73b9435f5/pkg/snet/udpaddr_test.go#L260)):*
> ```Go
> 			assert.NoError(t, err)
> 			assert.Equal(t, test.ia, a.IA.String())
> 			ipLegacy := net.ParseIP(test.host)
> ```

I think you will appreciate https://pkg.go.dev/net#IP.Equal

Also, try to figure out why it failed in the first place here.
Hint: IPv6 vs IPv4 form https://pkg.go.dev/net#IP



_Code quote:_
```Go
			ipLegacy := net.ParseIP(test.host)
			ip, err := netip.ParseAddr(test.host)
			assert.NoError(t, err)
			assert.Contains(t, []net.IP{ipLegacy, ip.AsSlice()}, a.Host.IP)
```

bunert (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254)* status: all files reviewed, 6 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`pkg/snet/udpaddr.go` line 42 at r2](https://reviewable.io/reviews/scionproto/scion/4254#-NCOz8Mv7o_C3ZQUTt6X:-NDWmG3QBHe9usiZErQd:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/4c83959b6cc66983a9330b4f3c2280aa6ad7d131/pkg/snet/udpaddr.go#L42)):*

Only half. Always explicitly return nil if there is no error.
</blockquote></details>

Done.
___
*[`pkg/snet/udpaddr.go` line 63 at r3](https://reviewable.io/reviews/scionproto/scion/4254#-NDTUGGz0khucldnhiZ9:-NDWmng68wZ02FaXFjz8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/40b29028d75d1f6af2f7e94db9c78ec73b9435f5/pkg/snet/udpaddr.go#L63)):*

This is a new error. The wrapped error is nil here.
</blockquote></details>

Done.
___
*[`pkg/snet/udpaddr.go` line 71 at r3](https://reviewable.io/reviews/scionproto/scion/4254#-NDTUPsB9yom0TFoJqW3:-NDWn6oh6Y5T9r5rz7V5:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/40b29028d75d1f6af2f7e94db9c78ec73b9435f5/pkg/snet/udpaddr.go#L71)):*
> ```Go
> 	}
> 	ip, err := netip.ParseAddr(parts[1])
> 	if len(parts) != 2 {
> ```

Done.
___
*[`pkg/snet/udpaddr_test.go` line 260 at r3](https://reviewable.io/reviews/scionproto/scion/4254#-NDTUjyD8ODy8wy3zrUD:-NDWvZS2-5zv5S0ur_IQ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/40b29028d75d1f6af2f7e94db9c78ec73b9435f5/pkg/snet/udpaddr_test.go#L260)):*

I think you will appreciate https://pkg.go.dev/net#IP.Equal

Also, try to figure out why it failed in the first place here.
Hint: IPv6 vs IPv4 form https://pkg.go.dev/net#IP
</blockquote></details>

Done.

oncilla (approved):
This should be part of a separate PR. Here we focus on the UDPAddr only.

> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254#-:-NDXKddz4IRIT4ZmQpAR:b-ebrhrd)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf)

matzf (approved):
:lgtm:

> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254#-:-NDY4O9j56ma98HXfA0K:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bunert)

matzf:
The linter is failing here because we use a relatively old version of golangci-lint, which does not support go-1.18. Without investigating this thoroughly, I suspect that this now fails because we're using `net/netip` for the first time. I'll create a separate pull request to update golangci-lint, then this should work.

matzf:
The lint issue is fixed by #4269.

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4254)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bunert)

PR #4259, opened 2022-09-27
doc, slayers: correct maximum for HdrLen

matzf:
The maximum for the representable value for HdrLen was incorrectly documented as 1024 bytes, but it is 255*4 == 1020 bytes.

Add sanity checks for the header length (which in essence are checks for the path length) when serializing the SCION header, to avoid silently creating malformed packets.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4259)

### DIFF
--- a/doc/protocols/scion-header.rst
+++ b/doc/protocols/scion-header.rst
@@ -51,7 +51,7 @@ HdrLen
     common header, the address header, and the path header). All SCION header
     fields are aligned to a multiple of 4 bytes. The SCION header length is
     computed as ``HdrLen * 4 bytes``. The 8 bits of the ``HdrLen`` field limit
-    the SCION header to a maximum of 1024 bytes.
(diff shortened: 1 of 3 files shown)


shitz (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4259)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4259#-NCy8d2f1hXbCHNDszIu:-NCy8d2f1hXbCHNDszIv:b-fwijcy) (no related file):*
Uff good catch. I'd prefer if we'd define the SCION header length as being `(HdrLen+1) * 4B`, however, that would probably require changes in the router (have not checked our router, but it seems you encountered this with the open-source router) and that's simply not worth the effort.

___

PR #4268, opened 2022-10-03
control: use AS certs in QUIC TLS, for DRKey auth

matzf:
The DRKey Level1 requests use TLS with certificates from the control plane PKI for authentication.
In the current implementation, we run two separate grpc "servers" on the same QUIC socket -- one with authentication, and one where authentication is disabled. Note that in either case, we use the TLS from QUIC, without authentication (`InsecureSkipVerify: true`), and so for the grpc server with authentication we run two nested TLS sessions.
This does not work properly. The two different grpc servers compete for accepted connections on the QUIC socket and chance determines whether a request is processed by the appropriate server. Automatic request retry has hidden the problem to some extent in early testing.

Fix this by using AS certificates in the QUIC connection directly, and never using grpc's TLS.
This approach has been previously considered impractical, because it was not clear how to access the QUIC connection's TLS state from the context of grpc.

Note that:
- we cannot verify for all requests; this would be a circular dependency e.g. for fetching certificate chains and trcs
- the client certificate is now validated directly from the DRKey grpc server logic. At this point, the TLS session has already been established and any errors result in a grpc error, not a TLS handshake error. This is unavoidable, as we can only know the method invoked after accepting the TLS session to know whether the certificates should be validated (see point above).

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4268)


JordiSubira (commented):
Just for completeness, there exists at the moment the alternative to implement two separate stacks (QUIC sockets) that will be resolved as different transport on the SVC resolution (QUIC vs TLS/QUIC). Similar to what was implemented on scionlab https://github.com/netsec-ethz/scion/pull/106.

An additional comment with this solution: In the future we might want to authenticate connections for intra-AS requests (e.g., `DRKeySecretValue`, etc.). For this it seems that the way to go is using gRPC with TLS enabled. Hence, we will probably have to adapt this alternative solution (i.e.,the "two-stacks" one).

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268#-:-NDYOnfj7q2viu2SKCKP:b4n7pa3)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`private/trust/tls_handshake.go` line 89 at r1](https://reviewable.io/reviews/scionproto/scion/4268#-NDYYDyw4ZVhhZweGUek:-NDYYDyw4ZVhhZweGUel:b-obpl9n) ([raw file](https://github.com/scionproto/scion/blob/32d5d703d7185a281b543091316ee0993b661242/private/trust/tls_handshake.go#L89)):*
> ```Go
> // VerifyClientCertificate verifies the certificate presented by the client
> // using the CP-PKI.
> func (m *TLSCryptoManager) VerifyClientCertificate(
> ```

This callback is not used anymore in real code. This would require to adapt/remove test cases for `tls_handshake_test.go`

matzf (commented):
Thanks for starting the discussion.

To me it seems quite clear that this approach here should be preferred. The only downside that I can see is that the implementation is a bit messy, requiring a little hack to pass the required information around. However, this is entirely an artifact of limitations of the grpc-go library -- I've filed [an issue](https://github.com/grpc/grpc-go/issues/5689) about exposing the required functionality directly.
From a higher level view, I cannot see any reason why using two ports (and the somewhat odd TLS-in-TLS) would be preferable.

Regarding the TLS for intra-AS requests, note that the situation is slightly different here, because we have the option to omit TLS entirely (using TCP). For the QUIC based inter-AS requests, we always have TLS, it's just that we don't always need to care about checking the certificates.
So, yes, if we want to support both TLS and non-TLS for intra-AS requests, using two ports would be an option. There are, however, also ways to serve these on the same port if we'd want to. One option for this is the [cmux](https://github.com/soheilhy/cmux) library. And btw, this would _also_ be an option for the inter-AS requests to serve the both the QUIC/TLS/HTTP2 and QUIC/HTTP2 requests on the same port.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268#-:-NDb3rcuEG1YqPATRwQt:b-abk56g)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

JordiSubira (commented):
Thank you for filing the issue. As you said both solutions seem a bit of a hack to me and I do not see any strong reason for discarding the one proposed in this PR.

Sure, I meant in the case we enable TLS to secure connections, coexisting with insecure gRPC ones. In any case, you're right and firstly, we should decide whether the two must coexist. Not sure how cmux would handle the lookahead inspection to distinguish plain gRPC and gRPC/TLS requests.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268#-:-NDhYN9t9VkQ97YvkDb9:b-b2b99x)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

matzf (commented):
In [the grpc issue mentioned above](https://github.com/grpc/grpc-go/issues/5689) a cleaner approach was suggested. I've replaced the hacky `ContextSmuggleAddr`/`PeerAuthInterceptor` with a custom implementation of the `grpc/credentials.TransportCredentials` interface which can achieve the same thing much more cleanly. Works like a charm :tada: 

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268#-:-NEAG9Us-VuHxiiMEqWl:bvry9tf)* status: 9 of 15 files reviewed, 1 unresolved discussion (waiting on @JordiSubira and @oncilla)

matzf (commented):
"Works like a charm"  ... :man_facepalming: 

It's flaky, the cert renewal test fails about half the time, and there is a very suspicious tls handshake error. I have no idea yet what is going on. The previous, hackier implementation works fine -- not sure what is different.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268#-:-NEB_fTr9asCj9Ik0NRj:blthjdb)* status: 9 of 15 files reviewed, 1 unresolved discussion (waiting on @JordiSubira and @oncilla)

matzf (commented):
Fixed. The problem was that when loading the key-pair for TLS, only the key was loaded from disk but the certificate chain was loaded only from the trust DB. During certificate renewal, in particular when overwriting the key/cert files with the `--force` flag, this results in no valid key pair being found.

I've pushed this fix to this branch for now, but I'll split this off into a separate PR.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268#-:-NEeI2me8zNm3JLo-2Xn:b-adszeb)* status: 8 of 18 files reviewed, 1 unresolved discussion (waiting on @JordiSubira and @oncilla)

JordiSubira (commented):
> Reviewed 3 of 6 files at r2, 2 of 2 files at r3, 4 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268)* status: 17 of 18 files reviewed, 2 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`private/trust/x509_provider.go` line 144 at r4](https://reviewable.io/reviews/scionproto/scion/4268#-NF9GkmI8_1CSKE8fHiE:-NF9GkmI8_1CSKE8fHiF:b51bibu) ([raw file](https://github.com/scionproto/scion/blob/9cebfa389b6208e7e6a8e7a1abff947e499f8acd/private/trust/x509_provider.go#L144)):*
> ```Go
> 			continue
> 		}
> 		if err := cppki.VerifyChain(chain, opts); err != nil {
> ```

Are we dropping this verification to simplify the code? If I didn't miss any piece of code, up to this point, it might still happen that a given certificate chain is loaded but not valid against any active TRC. Couldn't raising an error here help to catch the error before trying an unsuccessful handshake?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268)* status: 14 of 18 files reviewed, 2 unresolved discussions (waiting on @JordiSubira and @oncilla)

___
*[`private/trust/x509_provider.go` line 144 at r4](https://reviewable.io/reviews/scionproto/scion/4268#-NF9GkmI8_1CSKE8fHiE:-NFEo1u-4MZPS7kxbn-f:b-uny9t9) ([raw file](https://github.com/scionproto/scion/blob/9cebfa389b6208e7e6a8e7a1abff947e499f8acd/private/trust/x509_provider.go#L144)):*

Are we dropping this verification to simplify the code? If I didn't miss any piece of code, up to this point, it might still happen that a given certificate chain is loaded but not valid against any active TRC. Couldn't raising an error here help to catch the error before trying an unsuccessful handshake?
</blockquote></details>

This will be solved differently in #4281. I've removed this change from this branch here.

marcfrei (approved):
:lgtm:

> Reviewed 8 of 13 files at r1, 2 of 6 files at r2, 2 of 2 files at r3, 1 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268#-:-NMT1ngpEZUj29MIr4FZ:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

oncilla (commented):
> Reviewed 6 of 13 files at r1, 3 of 6 files at r2, 1 of 2 files at r3, 1 of 5 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

oncilla (approved):
> Reviewed 2 of 13 files at r1, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4268)* status: all files reviewed, 1 unresolved discussion (waiting on @JordiSubira)

___
*[`private/trust/tls_handshake.go` line 89 at r1](https://reviewable.io/reviews/scionproto/scion/4268#-NDYYDyw4ZVhhZweGUek:-NMdKocX6ca4hnGZMCGG:btxhbpo) ([raw file](https://github.com/scionproto/scion/blob/32d5d703d7185a281b543091316ee0993b661242/private/trust/tls_handshake.go#L89)):*

This callback is not used anymore in real code. This would require to adapt/remove test cases for `tls_handshake_test.go`
</blockquote></details>

I usually prefer not to keep unused code around too. In this case I'd prefer to keep it for the sake of completeness. As it's short one-liner and the logic is shared anyways, it doesn't hurt much to keep it.

PR #4269, opened 2022-10-05
lint: bump golangci-lint version for go-1.18, enable caching

matzf:
* bump golangci-lint version for go-1.18
   Update golangci-lint version to a version that support go-1.18, in
   particular the new additions to the standard libraries.
   Use last version before go-1.19 support was added, i.e. one version
   before golangci-lint v1.48.0. The reason is that the quic-go library
   version that we currently use cannot be built with go-1.19 and so
   linting fails.

* use docker volumes for golangci-lint cache
   Use docker volumes for the module and build cache in the golangci-lint
   container. Avoids having to very slowly refetch and build the all
   dependencies. Ideally this would somehow reuse the caches of the bazel
   golang toolchain, but I don't know if that's possible.
   This adds a third, redundant cache of the entire build (one on the host,
   one in bazel, and this third one for the golangci-lint cache).

   This cache will typically not be effective in buildkite CI (the docker
   volume goes away whenever a runner is shut down).

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4269)


oncilla (approved):
FWIW: I plan on updating quic-go to v0.29.1 and go version to 1.19.2 in the next couple of days.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4269#-:-NDe0Os74hpqlxhIpnHp:b-svmbk7)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

Issue #3461, opened 2019-12-06
CI: Setup fuzzing


oncilla:
Setup a fuzzing CI that fuzzes all our targets
- https://fuzzbuzz.io/
- https://fuzzit.dev/

PR #4270, opened 2022-10-05
slayers: check CurrINF/HF in decoding SCION path

### LINKED ISSUE #4133: Possibly missing check in SCION path DecodeFromBytes
Hi all,

There might be some checks missing in this function:
This decoder first decodes the MetaHdr, then it counts the number of Info fields and hops. The MetaHdr decoder first decodes the current info field and hop number. The *Base decoder never checks if CurrINF and CurrHF are valid indices.
Is this behavior intended?
https://github.com/scionproto/scion/blob/362c4622084037cd1d3ac593d2d78128a86638fd/go/lib/slayers/path/scion/base.go#L51-L70
https://github.com/scionproto/scion/blob/362c4622084037cd1d3ac593d2d78128a86638fd/go/lib/slayers/path/scion/base.go#L125-L137

matzf:
Check that the CurrINF and CurrHF fields are in the ranges defined by the number and lengths of segments.

Fixes #4133

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4270)

### DIFF
--- a/pkg/slayers/path/scion/base.go
+++ b/pkg/slayers/path/scion/base.go
@@ -66,6 +66,13 @@ func (s *Base) DecodeFromBytes(data []byte) error {
 		}
 		s.NumHops += int(s.PathMeta.SegLen[i])
 	}
+	if int(s.PathMeta.CurrINF) >= s.NumINF {
+		return serrors.New("CurrINF out of range", "value", s.PathMeta.CurrINF, "numInfo", s.NumINF)
+	}
+	if int(s.PathMeta.CurrHF) >= s.NumHops {
+		return serrors.New("CurrHF out of range", "value", s.PathMeta.CurrHF, "numHops", s.NumHops)
+	}
+
 	return nil
 }
 
--- a/pkg/slayers/path/scion/base_test.go
+++ b/pkg/slayers/path/scion/base_test.go
@@ -23,6 +23,147 @@ import (
 	"github.com/scionproto/scion/pkg/slayers/path/scion"
 )
 
+func TestPathMetaDecode(t *testing.T) {
+	testCases := map[string]struct {
+		raw       []byte
+		assertErr assert.ErrorAssertionFunc
+		expected  scion.MetaHdr
+	}{
+		"nil": {
+			raw:       nil,
+			assertErr: assert.Error,
+		},
+		"short": {
+			raw:       []byte{1, 2, 3},
+			assertErr: assert.Error,
+		},
+		"ok": {
+			raw: []byte{0x42, 0x0, 0x31, 0x05},
+			expected: scion.MetaHdr{
+				CurrINF: 1,
+				CurrHF:  2,
+				SegLen:  [3]uint8{3, 4, 5},
+			},
+			assertErr: assert.NoError,
+		},
+		"limits": {
+			raw: []byte{0xff, 0x03, 0xff, 0xff},
+			expected: scion.MetaHdr{
+				CurrINF: 3, // nonsensical but can be represented
+				CurrHF:  63,
+				SegLen:  [3]uint8{63, 63, 63},
+			},
+			assertErr: assert.NoError,
+		},
+	}
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			var actual scion.MetaHdr
+			err := actual.DecodeFromBytes(tc.raw)
+			tc.assertErr(t, err)
+			assert.Equal(t, tc.expected, actual)
+		})
+	}
+}
+
+func TestPathBaseDecode(t *testing.T) {
+	testCases := map[string]struct {
+		raw       []byte
+		assertErr assert.ErrorAssertionFunc
+		expected  scion.Base
+	}{
+		"nil": {
+			raw:       nil,
+			assertErr: assert.Error,
+		},
+		"short": {
+			raw:       []byte{1, 2, 3},
+			assertErr: assert.Error,
+		},
+		"three segments": {
+			raw: []byte{0x42, 0x0, 0x31, 0x05},
+			expected: scion.Base{
+				PathMeta: scion.MetaHdr{
+					CurrINF: 1,
+					CurrHF:  2,
+					SegLen:  [3]uint8{3, 4, 5},
+				},
+				NumINF:  3,
+				NumHops: 3 + 4 + 5,
+			},
+			assertErr: assert.NoError,
+		},
+		"two segments": {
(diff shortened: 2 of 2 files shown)


oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4270)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4270#-NDdqW0Z3nwhrWZfsRee:-NDdqW0Z3nwhrWZfsRef:bx9h1w9) (no related file):*
I would try to avoid doing these types of validation in the slayers package.

Slayers is a parsing layer. If we do validation in this package, we cannot use slayers to parse invalid packets.
IMO this is the responsibility of the upper layer that uses slayers, not slayers itself. 
E.g., for the raw path, this is done here:

https://github.com/scionproto/scion/blob/8fc16481027ca3547cfa519e7f714ac7149b68b7/pkg/slayers/path/scion/raw.go#L119-L121

and 

https://github.com/scionproto/scion/blob/8fc16481027ca3547cfa519e7f714ac7149b68b7/pkg/slayers/path/scion/raw.go#L146-L150

Note that before header v2 rework, we used to have tight coupling between validation and parsing and we consciously did not want to re-introduce this in slayers.

I wonder if #4133 was opened in response to something that is misbehaving and is missing the check.
If yes, we should fix that.

In any case, I think we should better document the intention behind slayers and packet validation in the package.

___

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4270)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4270#-NDdqW0Z3nwhrWZfsRee:-NDlipfIDEnezwPE-B-H:brnfs1w) (no related file):*

I would try to avoid doing these types of validation in the slayers package.

Slayers is a parsing layer. If we do validation in this package, we cannot use slayers to parse invalid packets.
IMO this is the responsibility of the upper layer that uses slayers, not slayers itself. 
E.g., for the raw path, this is done here:

https://github.com/scionproto/scion/blob/8fc16481027ca3547cfa519e7f714ac7149b68b7/pkg/slayers/path/scion/raw.go#L119-L121

and 

https://github.com/scionproto/scion/blob/8fc16481027ca3547cfa519e7f714ac7149b68b7/pkg/slayers/path/scion/raw.go#L146-L150

Note that before header v2 rework, we used to have tight coupling between validation and parsing and we consciously did not want to re-introduce this in slayers.

I wonder if #4133 was opened in response to something that is misbehaving and is missing the check.
If yes, we should fix that.

In any case, I think we should better document the intention behind slayers and packet validation in the package.





</blockquote></details>

Fair point.

The getter functions that you point out are indeed perfectly good for the practical use cases in the router etc.
However, for the `Decoded` path type, we don't have such a thing  and we probably don't want to rely on everybody checking the range before accessing `p.HopFields[p.CurrHF]` etc. This seems like a footgun waiting to cause problems. For example, the path `Reverse` does not seem to catch this and, althought it doesn't cause a panic straight away, will create malformed paths.

While I agree that slayers should be able to parse invalid packets, and that in places like the router we only want to validate the immediately relevant bits, there would ideally still be some way to perform a full validation. For example, paths can currently include trailing garbage and nothing will catch this. Even though I wouldn't know directly how this can be exploited, I think it's generally preferable not to be too lenient with things like this.
Where could we implement such an optional validation step?

___

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4270)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4270#-NDdqW0Z3nwhrWZfsRee:-NDmKXEL8D6vy2L6Lyxs:bpagygy) (no related file):*
> and we probably don't want to rely on everybody checking the range before accessing p.HopFields[p.CurrHF] etc.

IMO nobody other than library authors should be using in the first place. We could also expose a "safe" layer that is built on top of slayers. However, I think slayers needs to stay dumb. Otherwise, parsing and creating malformed packets will be impossible/annoyingly hard. 

> For example, the path Reverse does not seem to catch this and, althought it doesn't cause a panic straight away, will create malformed paths.

That is a bug we need to fix then. And probably we should start fuzzing slayers again (and maybe even continuously in CI)

> For example, paths can currently include trailing garbage and nothing will catch this. Even though I wouldn't know directly how this can be exploited, I think it's generally preferable not to be too lenient with things like this.

Agreed, we shouldn't be lenient about these things. However, I don't think the parsing layer is the right place to enforce it.

> Where could we implement such an optional validation step?

We can have a "safe" layer on top of slayers, if it is required. I wouldn't combine it with the parsing, because the validation can get arbitrarily complex, and I would really want the parsing logic to be as small as possible.

___

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4270)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4270#-NDdqW0Z3nwhrWZfsRee:-NE-mDd-CnS114fOcMHv:b-69qer3) (no related file):*

> and we probably don't want to rely on everybody checking the range before accessing p.HopFields[p.CurrHF] etc.

IMO nobody other than library authors should be using in the first place. We could also expose a "safe" layer that is built on top of slayers. However, I think slayers needs to stay dumb. Otherwise, parsing and creating malformed packets will be impossible/annoyingly hard. 

> For example, the path Reverse does not seem to catch this and, althought it doesn't cause a panic straight away, will create malformed paths.

That is a bug we need to fix then. And probably we should start fuzzing slayers again (and maybe even continuously in CI)

> For example, paths can currently include trailing garbage and nothing will catch this. Even though I wouldn't know directly how this can be exploited, I think it's generally preferable not to be too lenient with things like this.

Agreed, we shouldn't be lenient about these things. However, I don't think the parsing layer is the right place to enforce it.

> Where could we implement such an optional validation step?

We can have a "safe" layer on top of slayers, if it is required. I wouldn't combine it with the parsing, because the validation can get arbitrarily complex, and I would really want the parsing logic to be as small as possible.



</blockquote></details>

I've created #4277. For the fuzzing, we have #3461.

I like this idea that the decoder should only parse and not validate. We should make this clearer and more consistent. I've created #4278 and I'll close this PR.

___

Issue #3961, opened 2021-01-05
dispatcher: Replace with SCMP Daemon


scrye:
The dispatcher is the cause of many issues:
- It is a single point of failure for the SCION network stack of a host. If it goes down, a lot of processes (control service, gateway, tooling) go haywire.
- It lives in userspace, moving every forwarded packet from kernelspace to userspace and back to kernelspace, meaning a performance hit.
- It requires a UNIX DOMAIN stream socket, meaning awkward volume mounts in dockerized environments (and awkward resource cleanup/access management).
- It imposes stateful connections that need to be reestablished by client applications if the dispatcher goes down and then up. This has been the cause of many bugs in the past, and the performance impact of the reconnection logic is currently unknown.
- The UNIX DOMAIN socket is a stream, while the data being sent/received are packets. This requires packetization logic and slows forwarding even further.
- It has cost us a lot of time in performance tuning, because we never knew if it is the root cause of performance problems in apps like the POSIX gateway.
- It is an additional application to configure/deploy/operate on which, crucially, other applications depend on. This has caused headaches in testing where apps need to retry to connect to the dispatcher if it is not up yet, and it led to dependencies in `docker-compose`.
- It masks what applications are actually talking on the network, so a simple packet capture is often not sufficient to debug.
- It breaks native network tooling; `netstat` is ~ useless due to it.

The dispatcher has the following advantages:
- All SCION traffic is tunneled through a single port; this simplifies connectivity matrices and port firewall rules.
- The last SCION router on a path does not need to look up the L4 header to detect the underlay port, because it can assume 30041 there.

Also, the dispatcher as a stand-alone application (and not a forwarder) implements the logic to reply to certain SCMP messages like SCMP ping.

### The proposal

We should remove the dispatcher and replace it with an SCMP Daemon that only handles SCMP traffic.

SCION applications should open a UDP underlay port directly on the local host, on the same port as the UDP/SCION port. This will make both the port and the IP address match on both underlay and SCION.

Note that this is not a SCION protocol breaking change, because this is AS specific.

The roadmap would look like the following:
- Add a feature flag to the posix router, changing its behavior to set the UDP destination underlay port to the value of the UDP destination SCION port _only when they are the last router on the path_. Crucially, _this does not affect the performance of the high-speed core routers that need to forward huge volumes of data_.
- Have SCION applications use native `net.PacketConn` instead of `reliable.Conn`. This only needs changes in `snet`, and because the interfaces are compatible it is trivial to do.
[...]

PR #4280, opened 2022-10-11
doc: design doc for end host without dispatcher

matzf:
Design document for proposal #3961.

[doc]

---

In a separate PR, once this is merged, I will attempt to improve the organisation of these design documents and add a template.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)

### DIFF
--- a/doc/dev/design/index.rst
+++ b/doc/dev/design/index.rst
@@ -41,6 +41,7 @@ Active
    grpc
    BorderRouter
    router-perf-model
+   router-port-dispatch
 
 .. _design-docs-completed:
 
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -1,0 +1,180 @@
+End hosts without dispatcher
+============================
+
+- Author: Matthias Frei. 
+  Originally proposed by Sergiu Costea.
+- Last updated: 2022-10-11
+- Discussion at: https://github.com/scionproto/scion/issues/3961
+
+Abstract
+--------
+
+Remove the dispatcher, because it causes a variety of problems (some merely implementation problems, others inherent to the design).
+SCION applications open a UDP underlay socket directly, and use the corresponding port as the UDP/SCION port.
+The router in the destination AS inspects the Layer 4 UDP header and delivers the packet to this port.
+
+Background
+----------
+
+The SCION router in the destination AS delivers a packet to the end host at the specified destination address.
+The packet is delivered to the end host over UDP/IP, using a fixed end-host data port (30041).
+On the end host the packet is then delivered to the appropriate application socket determined by the protocol and port number; only UDP/SCION (and SCMP) are currently supported.
+We do not have operating system support for SCION on any platform, so this last step happens in user space, in the "dispatcher".
+
+Applications create UDP/SCION sockets by registering with the dispatcher, establishing a connection to the dispatcher's unix domain socket.
+The application then receives and transmits SCION packets (exclusively) over this unix domain socket connection.
+
+The dispatcher also implements other end host stack functionality; in particular, it replies to certain SCMP messages like SCMP Echo (ping).
+
+Problems
+^^^^^^^^
+This dispatcher is the cause of many issues:
+
+- It is a single point of failure for the SCION network stack of a host. If it goes down, a lot of processes (control service, gateway, tooling) go haywire.
+- It lives in userspace, moving every forwarded packet from kernelspace to userspace and back to kernelspace, meaning a performance hit.
+- It requires a UNIX DOMAIN stream socket, meaning awkward volume mounts in dockerized environments (and awkward resource cleanup/access management).
+- It imposes stateful connections that need to be reestablished by client applications if the dispatcher goes down and then up. This has been the cause of many bugs in the past, and the performance impact of the reconnection logic is currently unknown.
+- The UNIX DOMAIN socket is a stream, while the data being sent/received are packets. This requires packetization logic and slows forwarding even further.
+- It has cost us a lot of time in performance tuning, because we never knew if it is the root cause of performance problems in apps like the SCION IP gateway.
+- It is an additional application to configure/deploy/operate on which, crucially, other applications depend on. This has caused headaches in testing where apps need to retry to connect to the dispatcher if it is not up yet, and it led to dependencies in docker-compose.
+
+  This is also a big blocker for deploying multiple separate SCION-enabled applications to a larger variety of platforms.
+  For example, on mobile platforms (iOS, Android) there is simply no good way to install and start the dispatcher as a shared dependency of multiple applications.
+  Attempts to do so seem to require additional user interactions (in the form of: "please install and enable this other application"), which effectively blocks adoption of SCION use in mobile applications for most average users.
+
+- It masks what applications are actually talking on the network, so traditional firewalls cannot filter based on ports and a simple packet capture is often not sufficient to debug.
+- It breaks native network tooling; netstat is ~ useless due to it.
+
+The dispatcher has the following advantages:
+
+- It simulates an end host with full operating system support for SCION, i.e. it's conceptually "correct" and in line with our long-term vision.
+- All SCION traffic is tunneled through a single port; this simplifies connectivity matrices and port firewall rules.
+
+
+Long term vision
+^^^^^^^^^^^^^^^^
+
+In the longer term future, all of the functionality of the dispatcher may be
+included in the operating system's network stack. Applications open a UDP/SCION
+socket directly and the packet dispatching happens in the kernel.
+In this scenario, neither the user-space dispatcher nor the proposed workaround is needed.
+
+Proposal
+--------
+
+For UDP/SCION, SCION applications open a UDP/IP underlay socket directly, and use the corresponding port as the local UDP/SCION port.
+The last SCION router on a path inspects the Layer 4 UDP/SCION destination port to determine the underlay UDP/IP destination port.
+For traffic in the local AS, the source end host determines the underlay UDP/IP destination port analogously.
+
+SCMP
+^^^^
+
+The same procedure applies for SCMP messages, wherever possible.
+For SCMP error messages in response to UDP/SCION packets, the router uses the source port from the quoted offending packet as the underlay destination port.
+For SCMP error messages in response to a packet sent from a local end host, the underlay source address is used directly for the error message.
+
+For SCMP :ref:`Echo Requests <echo-request>` and :ref:`Traceroute Requests <traceroute-request>`, SCION applications open a UDP underlay port on the local host and use this port as the Identifier for the request messages.
+The requests are sent to the destination end host to the default end-host port 30041.
+For SCMP :ref:`Echo Replies <echo-reply>` and :ref:`Traceroute Replies <traceroute-reply>`, the router uses the Identifier field as the destination UDP port.
+
+SCMP Daemon
+^^^^^^^^^^^
+
+The remaining functionality of the dispatcher, namely responding to SCMP echo requests, is implemented to a new, very simple "SCMP daemon".
+This daemon opens UDP/IP port 30041, where it receives and replies to SCMP Echo requests.
+On this port, it will also receive any packet where an appropriate destination port could not be determined (e.g. SCMP error messages for malformed packages).
+These events are only logged and counted, but otherwise no appropriate action is possible.
+
+The SCMP daemon is an optional component for end hosts.
+If it's not running, the host simply doesn't respond to pings.
+
+Port Unreachable
+^^^^^^^^^^^^^^^^
+
+No SCMP Error messages for Port unreachable are sent. On the end host, there is simply no component that could trigger this. Instead, an ICMP port closed message for the UDP/IP port may be triggered.
+Given that the dispatcher currently doesn't even send out these SCMP messages, it does not seem to be worth the effort to translate the ICMP message to an SCMP in the router.
+
+Processing rule
+^^^^^^^^^^^^^^^
+
+1. The underlay UDP/IP destination port for packets towards the destination end host is chosen as follows:
+
+  - UDP/SCION: UDP/SCION destination port
+  - SCMP:
+
+    - :ref:`Echo Reply <echo-reply>`, :ref:`Traceroute Reply <traceroute-reply>`: Identifier field
+    - SCMP error messages:
+
+      - If quoted message is UDP/SCION: UDP/SCION source port
+      - Error message originating from this router to local end host: underlay source address of offending packet
+
+    - any other, in particular :ref:`Echo Request <echo-request>` and :ref:`Traceroute Request <traceroute-request>`: default end-host port 30041
+
+
+.. Hint:: This only applies to the ingress-router in the destination AS at the end of the path.
+   This does not affect the performance of the high-speed core routers that need to forward huge volumes of data.
+
+
+Long term vision compatibility
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If our long term vision materializes and we'd have SCION support directly built-in to the operating system's network stack, then this workaround becomes obsolete.
+In an optimistic scenario, where there are millions of end hosts running SCION-enabled applications, we can not expect that all devices and applications will updated to the same level of SCION support within a useful time frame.
+Therefore, it will be necessary to be able to gradually phase out the use of this workaround, keeping it around for all the future legacy applications.
+For this we propose to *conditionally* use the underlay UDP/IP destination port determined with the rules above only for specific *port ranges*.
+These port ranges are AS specific and are included in the topology configuration that end hosts and routers receive.
+
+Applications pick an appropriate local port when opening a socket.
+For servers listening on "well known ports", a (AS-local) flag day is still required at the point in time where these ports are removed from the special port range.
+
+
+The processing rule above is extended:
+
+2. If the underlay UDP/IP destination port determined above is within the port range specified in the topology configuration,
+   the packet is sent to that destination port.
+
+   Otherwise, the packet is sent to the default end-host port 30041.
+
+   .. Note:: In the future, we'd perhaps use a different port, or no longer use UDP/IP but directly IP as the underlay.
+
+.. Note:: This should not be considered a promise to never break compatiblity for end hosts again.
+
+Rationale
+---------
+
+Alternative without router support
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Instead of having the router inspect the Layer 4 data to determine the destination underlay port, we could keep this logic on the end host.
+Applications open IP/UDP sockets for transmitting packets, as above. To receive packets, a (much simplified) dispatcher process
+listens on port 30041 and forwards the packets to localhost:<UDP/SCION destination port>.
+This could also be implemented as a port rewrite step for example in an XDP program.
+
+This has some advantages:
+
+- it retains the "correct" end-host model, packets on the wire don't change
+- it's an entirely end-host local change
+- it also fixes many of the problems listed above related to unix domain sockets, reconnection, etc.
+
+However, this still requires a shared component on the end-host if multiple applications want to use SCION concurrently.
+For deployment to mobile platforms, in particular, this is still as much of a blocker as the current dispatcher. 
+
+Alternative SCMP handling
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The original proposal did not require the router to inspect SCMP messages. 
+All SCMP messages would be forwarded to the default end-host port and dispatched from there to the correct application with the SCMP daemon.
+
+Same issue as above; this still requires a shared component on the end-host if applications should be able to receive SCMP messages.
+SCMP error messages are crucial for efficient fail-over in SCION. Simply omitting these would not be a good option.
+
+Implementation
+--------------
+
+The roadmap would look like the following:
+
+- Add a feature flag to the router, changing its behavior to set the UDP destination underlay port as described above.
+- Have SCION applications use native ``net.PacketConn`` instead of ``reliable.Conn``. 
+  This only needs changes in ``snet``, and because the interfaces are compatible it is trivial to do.
+- Create new SCMP Daemon (scmpd) that replies to SCMP traffic that targets a host (scion ping)
+- Remove dispatcher, package ``reliable`` and ``ReconnectToDispatcher`` from everywhere.
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -157,29 +157,19 @@ The port range is configured in the ``topology.json`` file in the following form
 
   .. code-block:: yaml
 
-     underlay: {
-       dispatched_ports: {
-         "default": {
-            "min": <port>,
-            "max": <port>
-         },
-         "<ip-subnet>": { # any number of subnet-specific port-range configurations (in CIDR-notation)
-            "min": <port>,
-            "max": <port>
-         },
+     "underlay": {
+       "dispatched_ports": {
+         "min": <port>,
+         "max": <port>
+       }
      }
-
-For a given destination IP address, the configuration entry with longest matching prefix is
-considered, or if no match is found the entry "default" is used.
 
 The ``min``, ``max`` fields define the range of ports ``[min, max]`` (inclusive).
 The configured ``max`` value may be lower than ``min`` to specify an empty port range.
 
 If nothing is configured, the port range initially defaults to an empty range.
-The recommended port range is:
-- ``min``: ``31000``
-- ``max``: ``32767``
 
+The recommended port range is ``31000-32767``.
 This range is just below the port range that is assigned by the old dispatcher (32768-65535), ensuring that UDP traffic from legacy end hosts will be unaffected by the port dispatching in the router.
 On legacy hosts, SCMP echo and error requests currently use random IDs, and thus have a low chance (~2.5%) to pick an ID in the range that is port dispatched by the router. As a preparatory change, the range of IDs can reduced, so that there is no intersection.
 
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -220,10 +220,9 @@ The roadmap would look like the following:
 - Prepare:
    - Reduce the range of IDs used for SCMP echo and traceroute requests, so that it matches the range of ports assigned by the legacy dispatcher (32768-65535).
 
-   - Add support for dispatched/forwarded port ranges to the topology.json configuration.
-     This first implementation step only ensures that this entry can be parsed from the configuration.
+- Add support for dispatched/forwarded port ranges to the topology.json configuration.
 
-     Update: this can be skipped, as the topology.json parsing is lenient about unkown keys by default.
+  As the topology.json parsing is lenient about unkown keys by default, tue updated topology.json file can still be consumed by "legacy" applications.
 
 - Change SCION applications to use native ``net.UDPConn`` instead of ``reliable.Conn``.
 
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -1,9 +1,9 @@
 End hosts without dispatcher
 ============================
 
-- Author: Matthias Frei. 
+- Author: Matthias Frei.
   Originally proposed by Sergiu Costea.
-- Last updated: 2022-10-11
+- Last updated: 2022-11-15
 - Discussion at: https://github.com/scionproto/scion/issues/3961
 
 Abstract
@@ -88,6 +88,15 @@ These events are only logged and counted, but otherwise no appropriate action is
 The SCMP daemon is an optional component for end hosts.
 If it's not running, the host simply doesn't respond to pings.
 
+Service Addresses
+^^^^^^^^^^^^^^^^^
+
+Service destination address resolve to a configured underlay UDP/IP
+address, that is, to an IP *and* a (default) port number.
+
+The common use case for service addresses uses a UDP/SCION destination port of 0. In this case, the default underlay port is used.
+In case any other destination port is set, it overrides the default. The processing here is analogous to UDP/SCION packets with IP destination address type.
+
 Port Unreachable
 ^^^^^^^^^^^^^^^^
 
@@ -99,7 +108,11 @@ Processing rule
 
 1. The underlay UDP/IP destination port for packets towards the destination end host is chosen as follows:
 
-  - UDP/SCION: UDP/SCION destination port
+  - UDP/SCION:
+
+      - SVC destination address type and UDP/SCION destination port is 0: default port for resolved service address
+      - Else: UDP/SCION destination port
+
   - SCMP:
 
     - :ref:`Echo Reply <echo-reply>`, :ref:`Traceroute Reply <traceroute-reply>`: Identifier field
@@ -157,12 +170,12 @@ This has some advantages:
 - it also fixes many of the problems listed above related to unix domain sockets, reconnection, etc.
 
 However, this still requires a shared component on the end-host if multiple applications want to use SCION concurrently.
-For deployment to mobile platforms, in particular, this is still as much of a blocker as the current dispatcher. 
+For deployment to mobile platforms, in particular, this is still as much of a blocker as the current dispatcher.
 
 Alternative SCMP handling
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The original proposal did not require the router to inspect SCMP messages. 
+The original proposal did not require the router to inspect SCMP messages.
 All SCMP messages would be forwarded to the default end-host port and dispatched from there to the correct application with the SCMP daemon.
 
 Same issue as above; this still requires a shared component on the end-host if applications should be able to receive SCMP messages.
@@ -174,7 +187,7 @@ Implementation
 The roadmap would look like the following:
 
 - Add a feature flag to the router, changing its behavior to set the UDP destination underlay port as described above.
-- Have SCION applications use native ``net.PacketConn`` instead of ``reliable.Conn``. 
+- Have SCION applications use native ``net.PacketConn`` instead of ``reliable.Conn``.
   This only needs changes in ``snet``, and because the interfaces are compatible it is trivial to do.
 - Create new SCMP Daemon (scmpd) that replies to SCMP traffic that targets a host (scion ping)
 - Remove dispatcher, package ``reliable`` and ``ReconnectToDispatcher`` from everywhere.
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -128,50 +128,96 @@ Processing rule
    This does not affect the performance of the high-speed core routers that need to forward huge volumes of data.
 
 
+Compatibility
+^^^^^^^^^^^^^
+
+This change "only" affects the intra-AS forwarding, that is, there is no requirement to coordinate this update between different domains.
+
+Within each AS, we still need to be able to migrate to this new underlay without disrupting the network and without a synchronized update of all the hosts and routers of the AS.
+For this, we add two mechanisms:
+
+- the "shim dispatcher"; a simple, stateless UDP packet forwarder running on the updated end hosts, listening on UDP port 30041.
+  It inspects the L4 header and forwards all SCION packets to corresponding underlay port on the local host, following the processing rules for the router outlined above.
+
+  The shim dispatcher allows to update individual hosts before updating all of the routers.
+  The applications can receive the packets on individual UDP underlay ports and don't need to be aware of whether a packet was forwarded with the local dispatcher or was received directly (from the ingress router or an AS-local source host).
+
+- *conditionally* use the underlay UDP/IP destination port determined with the rules above only for specific *port ranges*.
+  These port ranges are AS specific and are included in the topology configuration that end hosts and routers receive.
+
+  As long as there there are no legacy devices/applications using ports in this range, we can update routers without disrupting any old hosts.
+  In this port range, we can operate *new* devices/applications *without* support from the shim dispatcher.
+
+  The processing rule above is extended:
+
+  2. If the underlay UDP/IP destination port determined above is within the port range specified in the topology configuration,
+     the packet is sent to that destination port.
+
+     Otherwise, the packet is sent to the default end-host port 30041.
+
+  The port range is configured in the ``topology.json`` file in the following format:
+
+    .. code-block:: yaml
+
+       "underlay": {
+         "dispatched_ports": "<min>-<max>"
+       }
+
+  The ``min``, ``max`` values define the range of ports ``[min, max]`` (inclusive).
+  The value ``"-"`` explicitly represents an empty range.
+  The value ``"all"`` represents the full range (``1-65535``).
+  If nothing is configured, the port range defaults to an empty range.
+
+  Applications pick ephemeral ports from this range when opening a socket.
+
+
+Update procedure
+""""""""""""""""
+
+With these mechanisms, the update procedure for an individual AS is:
+
+1. Pick a range for ``dispatched_ports``, and ensure that *no* existing applications are using ports in this range.
+
+   The recommended initial port range for the transition is ``31000-32767``.
+   This range is just below the range of ephemeral ports that is assigned by the old dispatcher (32768-65535), ensuring that UDP traffic from legacy end hosts will be unaffected by the port dispatching in the router.
+   On legacy hosts, SCMP echo and error requests currently use random IDs, and thus have a low chance (~2.5%) to pick an ID in the range that is port dispatched by the router. As a preparatory change, the range of IDs can reduced, so that there is no intersection.
+
+2. Update devices, **in any order**, without requiring synchronisation:
+
+   a. Update routers individually and enable ``dispatched_ports`` in their ``topology.json``.
+
+   b. Update hosts individually; replace the dispatcher with the shim-dispatcher, rebuild all
+      applications based on the updated libraries, and enable ``dispatched_ports`` in their ``topology.json``.
+
+3. Once all (or at least a significant portion of the) routers have been updated, new applications/devices can use ports in the range ``dispatched_ports`` without the shim dispatcher.
+
+   Depending on the types and number of end hosts in the AS and the time it takes for updates to be picked up, this state is may occur for anything from a few minutes to months, until ...
+
+4. Once all hosts have been updated, the ``dispatched_ports`` range can be extended to the entire port range.
+   The shim dispatchers can be disabled.
+
+.. Note:: If an AS operator controls all devices in the AS and/or does not plan to allow operating
+   new applications/devices without the shim dispatcher, they can pick the empty range in step 1.,
+   and state 3. is skipped.
+
+
 Long term vision compatibility
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-If our long term vision materializes and we'd have SCION support directly built-in to the operating system's network stack, then this workaround becomes obsolete.
+Conversely, if our long term vision materializes and we'd have SCION support directly built-in to the operating system's network stack, then this workaround becomes obsolete.
 In an optimistic scenario, where there are millions of end hosts running SCION-enabled applications, we can not expect that all devices and applications will updated to the same level of SCION support within a useful time frame.
 Therefore, it will be necessary to be able to gradually phase out the use of this workaround, keeping it around for all the future legacy applications.
-For this we propose to *conditionally* use the underlay UDP/IP destination port determined with the rules above only for specific *port ranges*.
-These port ranges are AS specific and are included in the topology configuration that end hosts and routers receive.
 
-Applications pick an appropriate local port when opening a socket.
-For servers listening on "well known ports", a (AS-local) flag day is still required at the point in time where these ports are removed from the special port range.
+.. Note:: In the future, we'd perhaps use a different port, or no longer use UDP/IP but directly IP as the underlay.
 
+The compatibility mechanisms introduced for the update can be reused for this "reverse" transition:
 
-The processing rule above is extended:
-
-2. If the underlay UDP/IP destination port determined above is within the port range specified in the topology configuration,
-   the packet is sent to that destination port.
-
-   Otherwise, the packet is sent to the default end-host port 30041.
-
-   .. Note:: In the future, we'd perhaps use a different port, or no longer use UDP/IP but directly IP as the underlay.
+- enable shim dispatcher for services outside of intended ``dispatched_port`` range
+- shrink ``dispatched_ports`` range and configure this on routers and hosts
+- on individual hosts: enable OS network stack support and update applications, disable shim dispatcher
 
 .. Note:: This should not be considered a promise to never break compatibility for end hosts again.
 
-
-The port range is configured in the ``topology.json`` file in the following format:
-
-  .. code-block:: yaml
-
-     "underlay": {
-       "dispatched_ports": {
-         "min": <port>,
-         "max": <port>
-       }
-     }
-
-The ``min``, ``max`` fields define the range of ports ``[min, max]`` (inclusive).
-The configured ``max`` value may be lower than ``min`` to specify an empty port range.
-
-If nothing is configured, the port range initially defaults to an empty range.
-
-The recommended port range is ``31000-32767``.
-This range is just below the port range that is assigned by the old dispatcher (32768-65535), ensuring that UDP traffic from legacy end hosts will be unaffected by the port dispatching in the router.
-On legacy hosts, SCMP echo and error requests currently use random IDs, and thus have a low chance (~2.5%) to pick an ID in the range that is port dispatched by the router. As a preparatory change, the range of IDs can reduced, so that there is no intersection.
 
 Rationale
 ---------
@@ -231,3 +277,10 @@ The roadmap would look like the following:
      Remove the packet dispatching/forwarding functionality from "dispatcher".
      Only SCMP echo responder remains in dispatcher. Rename to "SCMP Daemon" (scmpd).
    - set suitable default for port range in ``dispatched_ports`` topology configuration
+
+
+Compatibility and Update Procedure
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For operators,
+
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -91,7 +91,7 @@ If it's not running, the host simply doesn't respond to pings.
 Port Unreachable
 ^^^^^^^^^^^^^^^^
 
-No SCMP Error messages for Port unreachable are sent. On the end host, there is simply no component that could trigger this. Instead, an ICMP port closed message for the UDP/IP port may be triggered.
+No SCMP error messages for Port unreachable are sent. On the end host, there is simply no component that could trigger this. Instead, an ICMP port closed message for the UDP/IP port may be triggered.
 Given that the dispatcher currently doesn't even send out these SCMP messages, it does not seem to be worth the effort to translate the ICMP message to an SCMP in the router.
 
 Processing rule
@@ -137,7 +137,7 @@ The processing rule above is extended:
 
    .. Note:: In the future, we'd perhaps use a different port, or no longer use UDP/IP but directly IP as the underlay.
 
-.. Note:: This should not be considered a promise to never break compatiblity for end hosts again.
+.. Note:: This should not be considered a promise to never break compatibility for end hosts again.
 
 Rationale
 ---------
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -258,7 +258,7 @@ The roadmap would look like the following:
 
 - Add support for dispatched/forwarded port ranges to the topology.json configuration.
 
-  As the topology.json parsing is lenient about unkown keys by default, the updated topology.json file can still be consumed by "legacy" applications.
+  As the topology.json parsing is lenient about unknown keys by default, the updated topology.json file can still be consumed by "legacy" applications.
 
 - Change SCION applications to use native ``net.UDPConn`` instead of ``reliable.Conn``.
 
@@ -277,10 +277,3 @@ The roadmap would look like the following:
      Remove the packet dispatching/forwarding functionality from "dispatcher".
      Only SCMP echo responder remains in dispatcher. Rename to "SCMP Daemon" (scmpd).
    - set suitable default for port range in ``dispatched_ports`` topology configuration
-
-
-Compatibility and Update Procedure
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-For operators,
-
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -3,7 +3,7 @@ End hosts without dispatcher
 
 - Author: Matthias Frei.
   Originally proposed by Sergiu Costea.
-- Last updated: 2022-11-15
+- Last updated: 2024-03-26
 - Discussion at: https://github.com/scionproto/scion/issues/3961
 
 Abstract
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -80,7 +80,7 @@ For SCMP :ref:`Echo Replies <echo-reply>` and :ref:`Traceroute Replies <tracerou
 SCMP Daemon
 ^^^^^^^^^^^
 
-The remaining functionality of the dispatcher, namely responding to SCMP echo requests, is implemented to a new, very simple "SCMP daemon".
+The remaining functionality of the dispatcher, namely responding to SCMP echo requests, is implemented in a new, very simple "SCMP daemon".
 This daemon opens UDP/IP port 30041, where it receives and replies to SCMP Echo requests.
 On this port, it will also receive any packet where an appropriate destination port could not be determined (e.g. SCMP error messages for malformed packages).
 These events are only logged and counted, but otherwise no appropriate action is possible.
@@ -178,9 +178,10 @@ The configured ``max`` value may be lower than ``min`` to specify an empty port 
 If nothing is configured, the port range initially defaults to an empty range.
 The recommended port range is:
 - ``min``: ``31000``
-- ``max``: ``39999``
+- ``max``: ``32767``
 
-As later step, this range will be adopted as the default.
+This range is just below the port range that is assigned by the old dispatcher (32768-65535), ensuring that UDP traffic from legacy end hosts will be unaffected by the port dispatching in the router.
+On legacy hosts, SCMP echo and error requests currently use random IDs, and thus have a low chance (~2.5%) to pick an ID in the range that is port dispatched by the router. As a preparatory change, the range of IDs can reduced, so that there is no intersection.
 
 Rationale
 ---------
@@ -216,8 +217,13 @@ Implementation
 
 The roadmap would look like the following:
 
-- Add support for dispatched/forwarded port ranges to the topology.json configuration.
-  This first implementation step only ensures that this entry can be parsed from the configuration.
+- Prepare:
+   - Reduce the range of IDs used for SCMP echo and traceroute requests, so that it matches the range of ports assigned by the legacy dispatcher (32768-65535).
+
+   - Add support for dispatched/forwarded port ranges to the topology.json configuration.
+     This first implementation step only ensures that this entry can be parsed from the configuration.
+
+     Update: this can be skipped, as the topology.json parsing is lenient about unkown keys by default.
 
 - Change SCION applications to use native ``net.UDPConn`` instead of ``reliable.Conn``.
 
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -222,7 +222,7 @@ The roadmap would look like the following:
 
 - Add support for dispatched/forwarded port ranges to the topology.json configuration.
 
-  As the topology.json parsing is lenient about unkown keys by default, tue updated topology.json file can still be consumed by "legacy" applications.
+  As the topology.json parsing is lenient about unkown keys by default, the updated topology.json file can still be consumed by "legacy" applications.
 
 - Change SCION applications to use native ``net.UDPConn`` instead of ``reliable.Conn``.
 
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -150,7 +150,7 @@ For this, we add two mechanisms:
 
   The processing rule above is extended:
 
-  2. If the underlay UDP/IP destination port determined above is within the port range specified in the topology configuration,
+  2. If the underlay UDP/IP destination port determined above, i.e. in processing rule 1, is within the port range specified in the topology configuration,
      the packet is sent to that destination port.
 
      Otherwise, the packet is sent to the default end-host port 30041.
@@ -205,17 +205,19 @@ Long term vision compatibility
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Conversely, if our long term vision materializes and we'd have SCION support directly built-in to the operating system's network stack, then this workaround becomes obsolete.
-In an optimistic scenario, where there are millions of end hosts running SCION-enabled applications, we can not expect that all devices and applications will updated to the same level of SCION support within a useful time frame.
+In an optimistic scenario, where there are millions of end hosts running SCION-enabled applications, we can not expect that all devices and applications will be updated to the same level of SCION support within a useful time frame.
 Therefore, it will be necessary to be able to gradually phase out the use of this workaround, keeping it around for all the future legacy applications.
 
 .. Note:: In the future, we'd perhaps use a different port, or no longer use UDP/IP but directly IP as the underlay.
 
 The compatibility mechanisms introduced for the update can be reused for this "reverse" transition:
 
-- enable shim dispatcher for services outside of intended ``dispatched_port`` range
+- enable shim dispatcher for services outside of intended ``dispatched_port`` range.
 - shrink ``dispatched_ports`` range and configure this on routers and hosts
 - on individual hosts: enable OS network stack support and update applications, disable shim dispatcher
 
+.. Note:: It's not quite clear how we can practically coordinate that all relevant hosts have (re-)enabled the shim dispatcher.
+   This idea will need more work.
 .. Note:: This should not be considered a promise to never break compatibility for end hosts again.
 
 
@@ -276,4 +278,4 @@ The roadmap would look like the following:
    - remove support for old UDP underlay with default port 30041.
      Remove the packet dispatching/forwarding functionality from "dispatcher".
      Only SCMP echo responder remains in dispatcher. Rename to "SCMP Daemon" (scmpd).
-   - set suitable default for port range in ``dispatched_ports`` topology configuration
+   - set suitable default for port range in ``dispatched_ports`` topology configuration.
--- a/doc/dev/design/router-port-dispatch.rst
+++ b/doc/dev/design/router-port-dispatch.rst
@@ -152,6 +152,36 @@ The processing rule above is extended:
 
 .. Note:: This should not be considered a promise to never break compatibility for end hosts again.
 
+
+The port range is configured in the ``topology.json`` file in the following format:
+
+  .. code-block:: yaml
+
+     underlay: {
+       dispatched_ports: {
+         "default": {
+            "min": <port>,
+            "max": <port>
+         },
+         "<ip-subnet>": { # any number of subnet-specific port-range configurations (in CIDR-notation)
+            "min": <port>,
+            "max": <port>
+         },
+     }
+
+For a given destination IP address, the configuration entry with longest matching prefix is
+considered, or if no match is found the entry "default" is used.
+
+The ``min``, ``max`` fields define the range of ports ``[min, max]`` (inclusive).
+The configured ``max`` value may be lower than ``min`` to specify an empty port range.
+
+If nothing is configured, the port range initially defaults to an empty range.
+The recommended port range is:
+- ``min``: ``31000``
+- ``max``: ``39999``
+
+As later step, this range will be adopted as the default.
+
 Rationale
 ---------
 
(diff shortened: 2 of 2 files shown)


matzf:
For the continued discussion: please continue discussing the overall idea in #3961, and discuss the specifics of the proposed design document in this PR here.

doc/design/router-port-dispatch.rst
```diff
@@ -0,0 +1,180 @@
...
+On this port, it will also receive any packet where an appropriate destination port could not be determined (e.g. SCMP error messages for malformed packages).
+These events are only logged and counted, but otherwise no appropriate action is possible.
+
+The SCMP daemon is an optional component for end hosts.
+If it's not running, the host simply doesn't respond to pings.
+
+Port Unreachable
+^^^^^^^^^^^^^^^^
+
+No SCMP Error messages for Port unreachable are sent. On the end host, there is simply no component that could trigger this. Instead, an ICMP port closed message for the UDP/IP port may be triggered.
```
marcfrei: `SCMP Error` > `SCMP error`, like above

doc/design/router-port-dispatch.rst
```diff
@@ -0,0 +1,180 @@
...
+The processing rule above is extended:
+
+2. If the underlay UDP/IP destination port determined above is within the port range specified in the topology configuration,
+   the packet is sent to that destination port.
+
+   Otherwise, the packet is sent to the default end-host port 30041.
+
+   .. Note:: In the future, we'd perhaps use a different port, or no longer use UDP/IP but directly IP as the underlay.
+
+.. Note:: This should not be considered a promise to never break compatiblity for end hosts again.
```
marcfrei: compatibility

marcfrei (approved):
Thanks a lot, Matthias! I very much support this proposal, especially to have a convincing deployment story for mobile and small embedded platforms.

Another niche use-case that would benefit from this design are applications that rely on kernel or hardware based send and receive timestamps. In our ongoing time synchronization project we first tried to implement timestamping with a somewhat messy extension of the dispatcher and the ReliableSocket protocol over the UNIX domain socket before we finally decided not to have the default dispatcher on time server hosts and instead to listen directly on the end-host port from our application. This works for dedicated time servers but it wouldn't really be practical for end hosts where, for example, besides a time service client multiple other SCION based apps are deployed.

doc/design/router-port-dispatch.rst
matzf: Thanks, done. Turns out my spell-checker ignores these admonition blocks...

doc/design/router-port-dispatch.rst
matzf: Done.

marcfrei (approved):

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4280#-NGVwaD6ErVD7CYilfVf:-NGVyV_j7k5WNBN0diAJ:bx91rsn) (no related file):*

One detail that has not been discussed yet is the SVC redirection.

Currently, the dispatcher delivers SVC packets to a random process that registered for this service.
Currently, we just have control service and discovery service that need to reply to SVC resolution.

As I see it, there are two ways of solve this problem:

- The router is currently configured with IP addresses where it dispatches SVC packets to. We could modify this list to be UDP addresses. The services then open 2 UDP ports. One for SVC redirects and one for the actual QUIC connections.

- The "SCMP daemon" handles the redirection directly. I.e., the Service registers the port where it will accept QUIC connections with the "SCMP daemon". The services then only have to open 1 UDP port.

I'm slightly preferring the first option because it seems more reliable, because it resolves less processes. E.g., if the SCMP daemon unexpectedly stops, all SVC redirects fail. Also, after it starts again, all the services would need to re-register.






</blockquote></details>

(fwiw, I played around with the second approach in a hackathon because it required no changes to the router. It is feasible, but requires quite some state management to handle edge cases.)

___

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @oncilla)

___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-NG_gyJo8rBaFw4dn3vo:b-tzgj69) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*
That's a very fair concern.

> Checking whether the shared application is installed is fairly simple and and the user can be instructed to install the right application, which is a one time operation.
From the complexity to a user it feels ~same as granting camera access for an application.
And as soon as this would make it to the kernel, the additional app is also no longer needed. (To be fair, that's gonna take some time)

I don't share this view. While such an approach may be appropriate for niche cases like SCION demo applications, this will not work for more "consumer" oriented applications that just want to make use of SCION whenever it is available. For example, consider a browser (let's say Brave) and/or some video chat tool (e.g. Zoom) add SCION support to their apps at some point. If these now suddenly prompt the user with "hey please install this third party application `SCION SCMP dispatcher` to make things even better", this would confuse and annoy a lot of people and only a tiny fraction would follow through.

Also, somebody would have to own the shared dependency application. That would probably need to be us -- maintaining a user-facing package for various different platforms does not seem fun.

> Furthermore, while SCMP messages are good for efficient fail-over, they are solely informational.
Everything should work without SCMP messages (if not, we need to address this).
I would not classify them as crucial (and nor should they be)

That's a good point. So we could just say, tough luck, no SCMPs on this platform. Or, alternatively, we could offload the SCMP quote parsing to the router as proposed, but strictly rate limit this -- as they are optional, dropping any fraction is fair game.

---

Alternative idea:

Set the SCION FlowID based on the source port for all UDP packets. The flow ID is a 20 bit field, so we have room for the port plus 4 free bits. SCMP errors (already now) use the flow ID of the offending packet.
When delivering the SCMP message to the end host, the underlay destination port is determined by extracting the corresponding bits from the flow ID of the SCMP message. The quote does not have to be inspected.
This approach could also be used for SCMP Echo and Traceroute replies.

Setting the FlowID in this particular way would be required only for hosts that want packets to be port-dispatched by the router. Once we don't need this anymore (because e.g. SCION support is available from the operating system), the Flow ID field can again be chosen as an arbitrary high-entropy identifier.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4280#-NGVwaD6ErVD7CYilfVf:-NG_r8Qf23ewHXj8-XT2:b-6jr5lx) (no related file):*

(fwiw, I played around with the second approach in a hackathon because it required no changes to the router. It is feasible, but requires quite some state management to handle edge cases.)
</blockquote></details>

Thanks for pointing this out!

The first approach seems perfectly appropriate to me. I'll include this in the document.
Just for the record, the 2 ports are only necessary due to "implementation details" in snet/appnet. The service could also listen on the same UDP/IP port for both purposes, the application could internally dispatch packets for the two cases (based on the SCION destination address type).

___

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4280#-NGVwaD6ErVD7CYilfVf:-NGasQUj0vAIimo6FciE:b-xonxvu) (no related file):*
> The service could also listen on the same UDP/IP port for both purposes, the application could internally dispatch packets for the two cases (based on the SCION destination address type).

True. But I think it is easier to let the kernel take care of doing the dispatching instead of us needing to implement additional dispatching logic.

___

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed (commit messages unreviewed), 5 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`doc/design/router-port-dispatch.rst` line 27 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGq8iE0-wO6KAhIbMxl:-NGq8iE0-wO6KAhIbMxm:b-m1g4w2) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L27)):*
> ```reStructuredText
> The application then receives and transmits SCION packets (exclusively) over this unix domain socket connection.
> 
> The dispatcher also implements other end host stack functionality; in particular, it replies to certain SCMP messages like SCMP Echo (ping).
> ```

Another



_Code quote:_
```reStructuredText
 other
```
___
*[`doc/design/router-port-dispatch.rst` line 34 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGq9ox26lT-CE5EgjGe:-NGq9ox26lT-CE5EgjGf:b62o1np) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L34)):*
> ```reStructuredText
> 
> - It is a single point of failure for the SCION network stack of a host. If it goes down, a lot of processes (control service, gateway, tooling) go haywire.
> - It lives in userspace, moving every forwarded packet from kernelspace to userspace and back to kernelspace, meaning a performance hit.
> ```

user space; or at least try to make it consistent, e.g., above we have user space



_Code quote:_
```reStructuredText
userspace
```
___
*[`doc/design/router-port-dispatch.rst` line 159 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGqF0F7DZS866C3dyaN:-NGqF0F85Jm_Hg1pIg3S:b-79nrxq) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L159)):*
> ```reStructuredText
> - it also fixes many of the problems listed above related to unix domain sockets, reconnection, etc.
> 
> However, this still requires a shared component on the end-host if multiple applications want to use SCION concurrently.
> ```

end host



_Code quote:_
```reStructuredText
end-host
```
___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-NGqGWl_-_XgYy0Nxqm2:bc08rgo) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*

That's a very fair concern.

> Checking whether the shared application is installed is fairly simple and and the user can be instructed to install the right application, which is a one time operation.
From the complexity to a user it feels ~same as granting camera access for an application.
And as soon as this would make it to the kernel, the additional app is also no longer needed. (To be fair, that's gonna take some time)

I don't share this view. While such an approach may be appropriate for niche cases like SCION demo applications, this will not work for more "consumer" oriented applications that just want to make use of SCION whenever it is available. For example, consider a browser (let's say Brave) and/or some video chat tool (e.g. Zoom) add SCION support to their apps at some point. If these now suddenly prompt the user with "hey please install this third party application `SCION SCMP dispatcher` to make things even better", this would confuse and annoy a lot of people and only a tiny fraction would follow through.

Also, somebody would have to own the shared dependency application. That would probably need to be us -- maintaining a user-facing package for various different platforms does not seem fun.

> Furthermore, while SCMP messages are good for efficient fail-over, they are solely informational.
Everything should work without SCMP messages (if not, we need to address this).
I would not classify them as crucial (and nor should they be)

That's a good point. So we could just say, tough luck, no SCMPs on this platform. Or, alternatively, we could offload the SCMP quote parsing to the router as proposed, but strictly rate limit this -- as they are optional, dropping any fraction is fair game.

---

Alternative idea:

Set the SCION FlowID based on the source port for all UDP packets. The flow ID is a 20 bit field, so we have room for the port plus 4 free bits. SCMP errors (already now) use the flow ID of the offending packet.
When delivering the SCMP message to the end host, the underlay destination port is determined by extracting the corresponding bits from the flow ID of the SCMP message. The quote does not have to be inspected.
This approach could also be used for SCMP Echo and Traceroute replies.

Setting the FlowID in this particular way would be required only for hosts that want packets to be port-dispatched by the router. Once we don't need this anymore (because e.g. SCION support is available from the operating system), the Flow ID field can again be chosen as an arbitrary high-entropy identifier. 
</blockquote></details>

I find this alternative idea just fine if we want to spare the quoted packet inspection.  How are SCMP errors using the flowID now? In a similar way to decide the upper-layer process that must process the packet?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @JordiSubira, @marcfrei, and @oncilla)

___
*[`doc/design/router-port-dispatch.rst` line 27 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGq8iE0-wO6KAhIbMxl:-NGuYBJb5llMrnkcOLlV:bnoxinu) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L27)):*

Another
</blockquote></details>

Hmm, "functionality" is in plural here (uncountable) . To me, "other" seems better, or at least also acceptable.
___
*[`doc/design/router-port-dispatch.rst` line 34 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGq9ox26lT-CE5EgjGe:-NGuZXTm8fXWfmQJByad:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L34)):*

user space; or at least try to make it consistent, e.g., above we have user space
</blockquote></details>

Done.
___
*[`doc/design/router-port-dispatch.rst` line 159 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGqF0F7DZS866C3dyaN:-NGuZYGnA4XxsjbaLAqN:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L159)):*

end host
</blockquote></details>

Done.
___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-NGuZlDb9I_iXh65TOt-:b5k3iao) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*

I find this alternative idea just fine if we want to spare the quoted packet inspection.  How are SCMP errors using the flowID now? In a similar way to decide the upper-layer process that must process the packet?
</blockquote></details>

The FlowID is meant to be consumed by network devices, in particular the routers, to allow safely processing packets in parallel; only packets with the same FlowID are required to be kept in order.

Currently, this is not currently used in any more specific way for SCMPs. The router sets the FlowID of any SCMP error/traceroute response packet to the FlowID of the offending/request packet. That's simply the most sensible choice, but currently nothing depends on this behavior; with the proposed alternative idea, this would become mandatory.

The application process to receive an SCMP error message is determined by inspecting the source UDP port from the quoted offending message. SCMP traceroute or echo reply packets are delivered to applications based on the identifier field of these messages. Currently this happens in the dispatcher, in the long term future, this would happen in the OS network stack.

JordiSubira (commented):
:lgtm:

> Reviewed 1 of 2 files at r1, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280#-:-NGufdJq2gsyDBtZkXKW:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`doc/design/router-port-dispatch.rst` line 27 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGq8iE0-wO6KAhIbMxl:-NGueDIE8q8MNVEoxBhL:b9cxri5) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L27)):*

Hmm, "functionality" is in plural here (uncountable) . To me, "other" seems better, or at least also acceptable.
</blockquote></details>

No big deal, it can behave as both according to the dict.
___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-NGuetL8DyUKeCBiFvtK:b-stz2te) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*

The FlowID is meant to be consumed by network devices, in particular the routers, to allow safely processing packets in parallel; only packets with the same FlowID are required to be kept in order.

Currently, this is not currently used in any more specific way for SCMPs. The router sets the FlowID of any SCMP error/traceroute response packet to the FlowID of the offending/request packet. That's simply the most sensible choice, but currently nothing depends on this behavior; with the proposed alternative idea, this would become mandatory.

The application process to receive an SCMP error message is determined by inspecting the source UDP port from the quoted offending message. SCMP traceroute or echo reply packets are delivered to applications based on the identifier field of these messages. Currently this happens in the dispatcher, in the long term future, this would happen in the OS network stack.
</blockquote></details>

Thanks for the explanation! If we decide we adopt this alternative, I'd summarize and add this explanation.

marcfrei (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`doc/design/router-port-dispatch.rst` line 83 at r5](https://reviewable.io/reviews/scionproto/scion/4280#-NOnYGtAC5eMXND7Q7vl:-NOnYGtAC5eMXND7Q7vm:b-97dvvn) ([raw file](https://github.com/scionproto/scion/blob/4a1f243cf8afe511dd682aad5e67fb4ea066dc30/doc/design/router-port-dispatch.rst#L83)):*
> ```reStructuredText
> ^^^^^^^^^^^
> 
> The remaining functionality of the dispatcher, namely responding to SCMP echo requests, is implemented to a new, very simple "SCMP daemon".
> ```

"implemented in"?

marcfrei (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-NOySDya09bbBeaFDiMf:bpmmvtu) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*

Any more opinions on this point?

Summary of the proposed options:
- SCMP daemon (leading to no SCMPs on systems where this is not running/available)
- SCMP quote parsing in the router (as currently described in the design document), possibly with strict rate limits
- (ab-)using the FlowID to encode an underlay port for SCMP error message replies.

My current personal favorite is the FlowID hack.
I understood that @oncilla, you preferred the SCMP daemon variant over the parsing in the router, but what is your opinion about the FlowID hack?
</blockquote></details>

Two questions to better understand the three proposed alternatives:
- How likely is it that some future SCMP message types will not be "solely informational"? I.e., for example, that a particular SCMP error message type will become important for applications on the end host to work properly? I think this question is important because the new design really should not (or cannot) assume the presence of a shared SCMP daemon on the end host.
- Setting the FlowID based on the underlay port only for SCMP traffic wouldn't work because the assumption here is that peeking into the L4 payload could be too expensive also at the source of an SCMP message, right?

marcfrei (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-NOyeHUw5Ewab0PUqAfW:b-xck6qn) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*

> How likely is it that some future SCMP message types will not be "solely informational"? I.e., for example, that a particular SCMP error message type will become important for applications on the end host to work properly? I think this question is important because the new design really should not (or cannot) assume the presence of a shared SCMP daemon on the end host.

The "interface down" error messages already fall into this category, to some extent. While applications need to be aware that they cannot solely rely on these SCMP error messages to determine path health (as the SCMPs are rate limited and e.g. are not available if the first router on the path is dead), having the information about where the path failure occurs can help _a lot_ in handling this. Otherwise applications are just tapping in the dark; in many cases it's hard to even tell whether the path is down or whether the remote host or application failed. So applications either need to give up after some retries, pessimistically avoid a large section of the internet ("the failure might be in _any_ of the ASes on this path, so avoid them all"), or explore large number of paths to determine the broken link.


> Setting the FlowID based on the underlay port only for SCMP traffic wouldn't work because the assumption here is that peeking into the L4 payload could be too expensive also at the source of an SCMP message, right?

The idea was to set this "specially crafted" FlowID already when sending out the UDP packets. For SCMP error messages, we just use the FlowID of the offending packet (i.e. the value in the first line of the packet). The router already does this, and this seems like a sensible choice anyway.
The good thing is that only the sending host needs to be aware of this scheme. The bad bit is that all applications running on this host need to be aware of this scheme.
</blockquote></details>

Thanks! By using the underlay source port as the FlowID,  applications would loose the ability to implement recommendations like, e.g, the use of flow labels in QUIC [1]. This wouldn't be too big of a disadvantage?

[1] https://www.rfc-editor.org/rfc/rfc9000.html#name-use-of-ipv6-flow-label-and-

marcfrei (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-NOzTjo86mLE4B3x4yF2:bxnvc4) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*

Correct, although there are still 4 bits left to play with :wink: 
In the particular case, it seems unproblematic; the recommendation is to avoid reusing the same flow ID when migrating QUIC connection to a new address. When a new socket opened for this new address to migrate to, it will be assigned a different (random) port and thus, with the proposed approach, automatically use a different flow ID.

I don't know how big of a problem this is in general. Reading [RFC6437 ](https://www.rfc-editor.org/rfc/rfc6437.html), it doesn't feel like this would be super terrible though.
</blockquote></details>

So using the FlowID seems to have no obvious disadvantages so far, ignoring the fact that the source port would appear three times in an outgoing UDP/SCION packet...

JordiSubira:
For the record, I've started with the implementation of the part among which there's consensus (i.e., I'll leave the SCMP daemon for later). Any idea if how long and what range should we use to have compatibility.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 2 unresolved discussions (waiting on @marcfrei and @matzf)

___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-NTE7HVZ9Njy65ZKED16:bbzjdbc) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*

So using the FlowID seems to have no obvious disadvantages so far, ignoring the fact that the source port would appear three times in an outgoing UDP/SCION packet...
</blockquote></details>

Sorry for the late response.

This is quite an important topic, and the decision has a large impact on current and future deployments.
Because these changes are so fundamental, I have also brought up this topic internally at Anapaya, and we have discussed them extensively.

There are three main concerns that we have when changing such fundamental parts of the networking stack.

1. Can the change be rolled out incrementally?
2. Do we fragment the ecosystem?
3. Do we obstruct future progress?

Concern 1 is very important to us (Anapaya), because we already have a large productive deployment which must be updated incrementally.
We cannot have a single update day where the behavior of a full AS changes atomically. We must provide an interrupt free upgrade path on the granularity of one host/router.

Concern 2 is important if we want to keep the ecosystem coherent. It is also required if we want to support ASes that contain components from different vendors.

Concern 3 is important to us long-term. We do not want to back ourselves into a corner that we cannot get out of anymore. While this discussion is about an implementation detail that will not be put in the standard, we think it will still serve as a reference implementation and will be considered as the de-facto standard.

## Thoughts

Given these three concerns, here are our thoughts on the proposed solutions:

### Doing all the processing in the endhost (replace dispatcher ebpf solution)

This would be the cleanest solution because it does not introduce yet another layer violation.
Conceptually, the underlay port should not be tied to the UDP/SCION port. This change is compatible with the current deployment,
and does not fragment the ecosystem. It can also be incrementally rolled out.

However, it obstructs future progress in the mobile platforms. Running ebpf on an android host requires either root
privileges or a system code signature. Whether it is supported on iOS, is unknown to me.

### UDP/SCION Port == Underlay Port && SCMP Daemon does SCMP error dispatching

This solution introduces a layer violation by tying the underlay to the UDP/SCION layer.
It is incrementally deployable with the proposed reserved port range for packets where the rewrite should happen.
We run the risk of fragmenting the ecosystem, but it is rather low. The current dispatcher allocates ports 
sequentially in the range [32768-65535](https://github.com/scionproto/scion/blob/8b41109bcfdaa8c5dc2fd49a4e3a3274aa40de16/dispatcher/dispatcher.go#L66)
If the reserved range is chosen carefully, it is very unlikely that there is a clash.

The drawback of this approach is the SCMP daemon. According to @marcfrei it is hard to implement it on mobile platforms.
Thus, we might obstruct future progress with this solution.

###  UDP/SCION Port == Underlay Port && SCMP Quote parsing

This solution is essentially the same as the solution above, except that the SCMP Quote is read by the router and the underlay port is set to the original source port.
Same as above, this change should be incrementally deployable, and we run low risk of fragmenting the ecosystem.

The additional benefit is that the end hosts do not need the SCMP daemon to receive SCMP error messages. This will simplify deployments on mobile platforms a lot. However, this has a severe performance impact. Limiting the performance of routers (that are under heavy load) for a hack because the OS of a mobile platform can not provide the necessary tooling feels off.

### Flow ID

This solution introduces two layer violations. It ties both the underlay and the flow id to UDP/SCION.
This change can be incrementally rolled out, if we assume that the flow ID has not been used so far.
However, it is important to note that all other proposals above are AS-local changes. This change on the other hand requires all ASes to behave in an expected manner around the flow ID (re-using flow id for scmp error packet)

Tying the flow ID to the source port limits the number of different flows to 16. This might not be an issue for client applications that can open as many ports as they want, but on the server side, this can be quite restrictive.

According to [RFC6437](https://www.rfc-editor.org/rfc/rfc6437) "flow label values should be chosen such that their bits exhibit a high degree of variability".
This would be violated because even though ports can be chosen from the 16-bit space, usually they have very low variability.
Of course, the flow ID is not the single input, so it might be fine. But I don't really like going against an IPv6 RFC that was the inspiration for our flow ID.

While the flow ID is currently not used by scionproto, we (Anapaya) are actively working on RSS that includes the flow ID. We were planning to use in the order of hundreds of flow IDs per connection. Limiting the number of different flows to 16 is concerning to us as this heavily restricts the possible values.


## Conclusion

While replacing the dispatcher with an ebpf program that deals with the rewriting would be our favorite solution, we understand that it is not feasible at the moment for many platforms. In the interest of bringing native SCION applications to the end hosts, I'm in favor of going forward with one of the other proposals.

However the choice is not obvious to me yet. I would push against sticking it in the flow ID, for the reasons outlined above.  It is one more layer violation and limits the usefulness of the flow ID to much for its actual purpose .


Parsing the quote on the router is quite expensive. Essentially, we at least double the parsing effort. It would be best to avoid this if possible. Thus, I'm torn between parsing the SCMP packets in the router, and having an SCMP daemon on the end host. If we were to implement the parsing option, we will definitely start with SCMP rate limiting.

@marcfrei could you elaborate on the challenges of implementing an SCMP daemon on mobile platforms?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 2 unresolved discussions (waiting on @oncilla)

___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-NTTWauQDLqb4_y7jUZ2:b-8vs893) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*

Sorry for the late response.

This is quite an important topic, and the decision has a large impact on current and future deployments.
Because these changes are so fundamental, I have also brought up this topic internally at Anapaya, and we have discussed them extensively.

There are three main concerns that we have when changing such fundamental parts of the networking stack.

1. Can the change be rolled out incrementally?
2. Do we fragment the ecosystem?
3. Do we obstruct future progress?

Concern 1 is very important to us (Anapaya), because we already have a large productive deployment which must be updated incrementally.
We cannot have a single update day where the behavior of a full AS changes atomically. We must provide an interrupt free upgrade path on the granularity of one host/router.

Concern 2 is important if we want to keep the ecosystem coherent. It is also required if we want to support ASes that contain components from different vendors.

Concern 3 is important to us long-term. We do not want to back ourselves into a corner that we cannot get out of anymore. While this discussion is about an implementation detail that will not be put in the standard, we think it will still serve as a reference implementation and will be considered as the de-facto standard.

## Thoughts

Given these three concerns, here are our thoughts on the proposed solutions:

### Doing all the processing in the endhost (replace dispatcher ebpf solution)

This would be the cleanest solution because it does not introduce yet another layer violation.
Conceptually, the underlay port should not be tied to the UDP/SCION port. This change is compatible with the current deployment,
and does not fragment the ecosystem. It can also be incrementally rolled out.

However, it obstructs future progress in the mobile platforms. Running ebpf on an android host requires either root
privileges or a system code signature. Whether it is supported on iOS, is unknown to me.

### UDP/SCION Port == Underlay Port && SCMP Daemon does SCMP error dispatching

This solution introduces a layer violation by tying the underlay to the UDP/SCION layer.
It is incrementally deployable with the proposed reserved port range for packets where the rewrite should happen.
We run the risk of fragmenting the ecosystem, but it is rather low. The current dispatcher allocates ports 
sequentially in the range [32768-65535](https://github.com/scionproto/scion/blob/8b41109bcfdaa8c5dc2fd49a4e3a3274aa40de16/dispatcher/dispatcher.go#L66)
If the reserved range is chosen carefully, it is very unlikely that there is a clash.

The drawback of this approach is the SCMP daemon. According to @marcfrei it is hard to implement it on mobile platforms.
Thus, we might obstruct future progress with this solution.

###  UDP/SCION Port == Underlay Port && SCMP Quote parsing

This solution is essentially the same as the solution above, except that the SCMP Quote is read by the router and the underlay port is set to the original source port.
Same as above, this change should be incrementally deployable, and we run low risk of fragmenting the ecosystem.

The additional benefit is that the end hosts do not need the SCMP daemon to receive SCMP error messages. This will simplify deployments on mobile platforms a lot. However, this has a severe performance impact. Limiting the performance of routers (that are under heavy load) for a hack because the OS of a mobile platform can not provide the necessary tooling feels off.

### Flow ID

This solution introduces two layer violations. It ties both the underlay and the flow id to UDP/SCION.
This change can be incrementally rolled out, if we assume that the flow ID has not been used so far.
However, it is important to note that all other proposals above are AS-local changes. This change on the other hand requires all ASes to behave in an expected manner around the flow ID (re-using flow id for scmp error packet)

Tying the flow ID to the source port limits the number of different flows to 16. This might not be an issue for client applications that can open as many ports as they want, but on the server side, this can be quite restrictive.

According to [RFC6437](https://www.rfc-editor.org/rfc/rfc6437) "flow label values should be chosen such that their bits exhibit a high degree of variability".
This would be violated because even though ports can be chosen from the 16-bit space, usually they have very low variability.
Of course, the flow ID is not the single input, so it might be fine. But I don't really like going against an IPv6 RFC that was the inspiration for our flow ID.

While the flow ID is currently not used by scionproto, we (Anapaya) are actively working on RSS that includes the flow ID. We were planning to use in the order of hundreds of flow IDs per connection. Limiting the number of different flows to 16 is concerning to us as this heavily restricts the possible values.


## Conclusion

While replacing the dispatcher with an ebpf program that deals with the rewriting would be our favorite solution, we understand that it is not feasible at the moment for many platforms. In the interest of bringing native SCION applications to the end hosts, I'm in favor of going forward with one of the other proposals.

However the choice is not obvious to me yet. I would push against sticking it in the flow ID, for the reasons outlined above.  It is one more layer violation and limits the usefulness of the flow ID to much for its actual purpose .


Parsing the quote on the router is quite expensive. Essentially, we at least double the parsing effort. It would be best to avoid this if possible. Thus, I'm torn between parsing the SCMP packets in the router, and having an SCMP daemon on the end host. If we were to implement the parsing option, we will definitely start with SCMP rate limiting.

@marcfrei could you elaborate on the challenges of implementing an SCMP daemon on mobile platforms?

</blockquote></details>

Thanks for the very structured write up, that's great. For what it's worth, I believe we've implicitly considered exactly these three concerns when we discussed this design, but it's great to spell this out explicitly.

As a general point regarding the "layer violation" objection. The way I see the UDP underlay is that this is already a stop-gap approach. There is a mismatch between the address type in the SCION header (IP v4/v6) and the underlay (UDP, i.e. IP + port). Now, actually having the UDP port to the SCION address would seem very odd. Instead, the "right" approach should be to have IP directly as the underlay.
My guess as to why we don't have IP underlay directly (yet) is that it's just much more practical to use UDP compared to defining a new protocol number which would certainly be dropped by all kinds of middleboxes. So, the UDP underlay is just a pragmatic, convenient choice to aid deployability. 
So, the UDP underlay is already "wrong", so milking this UDP underlay a bit more to aid deployability _more_ seems fair game.
I admit this is a bit of a foot-in-the-door argument.

What we ideally want to achieve is to make it possible to deploy multiple SCION-native applications independently and fully self-contained. There should be no requirement for any non-standard system components to be installed or running, and no coordination between different applications.
The ebpf-dispatcher approach does not achieve this (generally, not only on mobile platforms); something will need to install this ebpf program (and even needs special privileges for this).
Requiring an SCMP daemon fails this ideal goal in the same way (needs to be installed and started _somehow_).

I can understand the concerns against ab-using the flow ID.

----

Here are two variants on the approaches suggested before:

#### UDP/SCION Port == Underlay Port && shared responsibility for SCMP error dispatching

This solution is essentially the same as the SCMP daemon approach, except that instead of having a specific SCMP daemon process, all the SCION applications listen on the UDP port 30041 for SCMP error messages with SO_REUSEPORT (or some analogous platform specific option). The applications parse the SCMP packets and forward to the correct application (send via IP to localhost:port).
This approach does not achieve the ideal of "no coordination between applications". The coordination effort is modest, though, and there is no need to explicitly register with or startup/shutdown central components. The mechanism appears to be available in some form on almost any platform. There is, however, information leaking between the non-privileged applications (applications can see the quoted message causing the SCMP error).
On Linux, the SO_REUSEPORT mechanism is limited to processes of the same user ID; this is a bit awkward for our use case, but not a huge problem. Applications need to ignore bind errors when opening this socket, and, in case of a bind error periodically re-try to open the socket to cover the case all applications previously "serving" the SCMP error port have gone away.

#### UDP/SCION Port == Underlay Port && _limited_ SCMP Quote parsing by abusing the reserved bits in the common header

This is a frankeinstein monster mash-up of the SCMP Quote parsing and the Flow ID ideas.
To avoid having to parse the full quoted packet header down to the L4, we squirrel away the port information to a fixed location early in the packet header.
We could again use the Flow ID for this idea, but actually there are 16 reserved bits in the common header just waiting to be (ab-)used.
In contrast to the Flow ID approach, this does not require any behavioral change of the routers in other ASes, as long as they faithfully copy the offending packet to the quote.

marcfrei:
> ... could you elaborate on the challenges of implementing an SCMP daemon on mobile platforms?

I don't know whether and how this could work on Android but at least on iOS/iPadOS there is as far as I can see simply no concept of a user-installable background daemon. There is however the possibility to install and enable a separate app that provides a so-called Network Extension (https://developer.apple.com/documentation/networkextension) which could come close to what is needed here.

We would need to investigate whether it is feasible in principle to implement SCMP daemon functionality based on what the network extension model allows. In addition, there's then also the question how such a network extension would work together with other network extensions. VPN apps (like, e.g. Tailscale) have to be implemented as network extensions and at least VPN extensions are mutually exclusive among each other. Independent of the technical feasibility it also seems to me that such a solution would be rather finicky to manage for typical end-users. Installing a network extension requires entry of the device passcode which already feels a little scary to begin with. It's then very easy to imagine support situations where a SCION-based application doesn't work as expected. The provider of such an app would need to ask the user to check whether the separate SCION SCMP daemon app that provides the network extension is already installed and enabled correctly. The approach thus wouldn't really satisfy Matthias's requirement:

> What we ideally want to achieve is to make it possible to deploy multiple SCION-native applications independently and fully self-contained. There should be no requirement for any non-standard system components to be installed or running, and no coordination between different applications.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @JordiSubira, @marcfrei, and @oncilla)

___
*[`doc/dev/design/router-port-dispatch.rst` line 227 at r8](https://reviewable.io/reviews/scionproto/scion/4280#-NX9bZ5zDWhB55kH40VJ:-NX9bZ5zDWhB55kH40VK:bs7neqv) ([raw file](https://github.com/scionproto/scion/blob/1df97866af9b591d4c35c0ab49069416277742c3/doc/dev/design/router-port-dispatch.rst#L227)):*
> ```reStructuredText
>   packet forwarder. This forwarder listens on UDP port 30041. It inspects the L4 header and forwards
>   all SCION packets to corresponding underlay port on the local host, following the processing
>   rules for the router outlined above.
> ```

As mentioned in https://github.com/scionproto/scion/pull/4344#pullrequestreview-1443688080 and discussed in the [TC Implementation meeting on 1st June](https://github.com/scionproto/scion/issues/4327#issuecomment-1573359749), this design document was missing a compatibility mechanism to ensure that applications/services can function even if the port-dispatching feature is only partially available in the routers of the AS.

JordiSubira:
While I was implementing this, one question strikes me as a relevant one. Perhaps, you have already discussed this on some TC meeting or elsewhere, however I don't remember discussing it with anyone.

As it is implemented right now in #4344 , the shim has become a UDP datagram forwarder.  In practice, we are not simulating the SCION underlay on the endhost stack as before. Previously, if one application doesn't register explicitly for SCION/UDP address,the dispatcher wouldn't send packets to this (IP:port). In contrast, right now, the shim would send packets to this IP:Port. Perhaps, explicitly documenting this, for helping operator on setting mechanism as network namespaces (or similar) is desirable.

On top of that, if the shim listens on address any, the shim MAY send packets intended to "internal" services. Of course, this will be subject on the NW traffic rules, whether they allow the BR to send packets to this "internal" address and whether by configuration the operator configures the shim with the ANY address.

Finally, I think it's weird having legitimate use case, in which the IP address for UDP/IP is different from the IP for SCION/UDP at the endhost. For this, we can add a check ( as suggested here https://github.com/scionproto/scion/pull/4344#pullrequestreview-1692508277).

matzf:
Ah, good points, Jordi.

So the "shim" unintentionally allows bypassing the firewalls. That's certainly a concern. As we forward the packet to the address listed in the SCION address header, there is also the risk that the shim is abused as a reflector to forward packets to different hosts.
I think the suggested additional check to validate that the underlay destination IP corresponds to the SCION destination address would be a good mitigation. It ensures that we only send packets to addresses on this host and also that at least the IP destination is routable from the sender (some strange routing rules could still invalidate this, but still...). 
The only thing that we'd then allow to bypass is filtering of specific UDP ports based on the source address. If we make this clear in the documentation, it should be relatively straight forward to extend the rules to also block packets from the shim.

To implement this check, we'll need to obtain the underlay destination IP address of the UDP packet. I believe we'll have to use sockopt `IP_PKTINFO`. I don't know in detail how portable this is, but quite certainly we'll have to have some platform specific code for this.

JordiSubira:
Another point to discuss, which was raised in a discussion started by @marcfrei: How should we handle the server side running with a shim component? Currently the reader uses the udp address of the lastHop to send the reply back (https://github.com/JordiSubira/scion/blob/dispatcher_off/pkg/snet/reader.go#L105C3-L105C34). For traffic going from a legacy BR -> shim dispatcher -> endhost app, the endhost app listening on the UDP/SCION socket will see the shim address as the _lastHop_.

In this discussion 3 alternatives came up, although each one of them with its own pros/cons:
1. The shim adds the udpAddress of the BR in one especial extension before sending the packet to the endhost application. The reader will check whether the incoming packet includes this extension and, if so, it will use the address encoded as the nextHop for sending the reply back. This would add a new extension header (it would feel weird to use H2H or E2E for this) and the packet will grow some extra bytes between the shim and the endhost app. We could also abuse the H2H extension and include there the information.
2. Since the path contains the interface ID (ifID) of the BR, the endhost application (via the library) could poll periodically the Interface information from the Daemon (https://github.com/JordiSubira/scion/blob/dispatcher_off/daemon/internal/servers/grpc.go#L259) and then resolve this interface address for setting the NextHop address in the reply packets. The issue is how often this poll is done. Even though, the topology information is not expected to change often, when it does so and until the next poll, the server will be effectually down. Thus we might want that this poll is in the order of some seconds. 
3. Instead of polling the daemon, the endhost application would consume this information from the topology file. The topology loader allows to reload the information based on signaling and thus could reload the information if this ever changes on the FS. The downside is that we make the end applications having to deal also with the topology file, which doesn't seem convinient at all.
4. Create some subscrption-like RPC on the daemon, i.e., the end application subscribes to it and receives topology changes whenever the daemon is aware of those (if ever). Then it uses the locally stored information to reply back.

Let us know what do you think is the best approach or suggest yours if it's not listed before :)

JordiSubira:
> > Ah, good points, Jordi.
> > So the "shim" unintentionally allows bypassing the firewalls. That's certainly a concern. As we forward the packet to the address listed in the SCION address header, there is also the risk that the shim is abused as a reflector to forward packets to different hosts. I think the suggested additional check to validate that the underlay destination IP corresponds to the SCION destination address would be a good mitigation. It ensures that we only send packets to addresses on this host and also that at least the IP destination is routable from the sender (some strange routing rules could still invalidate this, but still...). The only thing that we'd then allow to bypass is filtering of specific UDP ports based on the source address. If we make this clear in the documentation, it should be relatively straight forward to extend the rules to also block packets from the shim.
> > To implement this check, we'll need to obtain the underlay destination IP address of the UDP packet. I believe we'll have to use sockopt `IP_PKTINFO`. I don't know in detail how portable this is, but quite certainly we'll have to have some platform specific code for this.
> 
> I took a look at this and I would like to follow the discussion in here. After looking more deeply into this, I reached the conclusion that Windows support in Golang for this is far from being great. There's no proper support for doing this using x/net/ipv[4,6] and this issue [https://github.com/[golang/go/issues/7175](https://github.com/golang/go/issues/7175)] is still open since 2014... so it's a bit unclear. Similarly, there is no support from syscall/windows for auxiliary classes which could help parsing this ancillary information, .e.g., https://pkg.go.dev/syscall#Cmsghdr.
> 
> I think we could get our hands dirty by dealing with lower level code, doing something like this: https://gist.github.com/JordiSubira/775693d03e6fcfe71f9c67779013e136. However, I do not know whether this is a path we want to go through.
> 
> Thus, there are two alternatives (at least):
> 
>     1. Fiddling with `syscall/windows` and low-level code, hoping that we don't mess with anything.
> 
>     2. Changing the approach in Windows (or by extension in all platforms) Hooking the binding to "0.0.0.0"  and binding on all reasonable addresses in parallel in different go routines and do the proper check.
> 
> 
> Please, share your thoughts on that @martenwallewein since you were also playing with this PR your contribution would be useful as well :)

After offline discussion, we finally opted for discontinuing the shim component for windows. Therefore, in network environments with legacy BRs, scion-enabled Windows applications with the new endhost stack (as described in this documentation) will not be able to communicate with those legacy BRs. Take into account, that so far the scion endhost stack (more concretely, the dispatcher) could not be compiled for Windows.

martenwallewein:
> > Ah, good points, Jordi.
> > So the "shim" unintentionally allows bypassing the firewalls. That's certainly a concern. As we forward the packet to the address listed in the SCION address header, there is also the risk that the shim is abused as a reflector to forward packets to different hosts. I think the suggested additional check to validate that the underlay destination IP corresponds to the SCION destination address would be a good mitigation. It ensures that we only send packets to addresses on this host and also that at least the IP destination is routable from the sender (some strange routing rules could still invalidate this, but still...). The only thing that we'd then allow to bypass is filtering of specific UDP ports based on the source address. If we make this clear in the documentation, it should be relatively straight forward to extend the rules to also block packets from the shim.
> > To implement this check, we'll need to obtain the underlay destination IP address of the UDP packet. I believe we'll have to use sockopt `IP_PKTINFO`. I don't know in detail how portable this is, but quite certainly we'll have to have some platform specific code for this.
> 
> I took a look at this and I would like to follow the discussion in here. After looking more deeply into this, I reached the conclusion that Windows support in Golang for this is far from being great. There's no proper support for doing this using x/net/ipv[4,6] and this issue [https://github.com/[golang/go/issues/7175](https://github.com/golang/go/issues/7175)] is still open since 2014... so it's a bit unclear. Similarly, there is no support from syscall/windows for auxiliary classes which could help parsing this ancillary information, .e.g., https://pkg.go.dev/syscall#Cmsghdr.
> 
> I think we could get our hands dirty by dealing with lower level code, doing something like this: https://gist.github.com/JordiSubira/775693d03e6fcfe71f9c67779013e136. However, I do not know whether this is a path we want to go through.
> 
> Thus, there are two alternatives (at least):
> 
> 1. Fiddling with `syscall/windows` and low-level code, hoping that we don't mess with anything.
> 2. Changing the approach in Windows (or by extension in all platforms) Hooking the binding to "0.0.0.0"  and binding on all reasonable addresses in parallel in different go routines and do the proper check.
> 
> Please, share your thoughts on that @martenwallewein since you were also playing with this PR your contribution would be useful as well :)

Hey @JordiSubira thanks for summarizing this. I've thought some time about this and here is my conclusion:

For me it feels like the windows low-level code is more a shot in our own knee, that is probably resulting in some effort to maintain this and to keep it up-to-date with other features.

The approach of binding to each reasonable interface manually with an own go routine is o.k. for me (I mean if I remember correctly it was my idea). If the core processing routine is the same for all interfaces which just checks different addresses then I think the overhead of a bit of concurrency is fine.

What's the state of the idea of dropping shim support for windows? I mean we still have the firewall issue, and running the newest routers seems like a good idea anyway. Hope this helps, thanks all for your work on this!

JordiSubira:
Hello @martenwallewein, thank you for jumping in. Of course, this is a possibility, having different code for each platform. However, we should either specify somewhere that in windows the user of the library shouldn't bind to 0.0.0.0 or do it under their own risk of bypassing firewall rules, as you said. Or we limit it futher with more platform specific condition, in which we don't allow to bind to 0.0.0.0 in Windows. If we agree on that, I can try to have this for future iterations.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @JordiSubira, @marcfrei, and @oncilla)

___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-Nttr5YH3XkrfCy-WAO5:bklw6hl) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*

> My guess as to why we don't have IP underlay directly (yet) is that it's just much more practical to use UDP compared to defining a new protocol number which would certainly be dropped by all kinds of middleboxes. So, the UDP underlay is just a pragmatic, convenient choice to aid deployability.

Yes, this is exactly the reason. This is also why QUIC builds on top of UDP/IP. Protocol numbers unfortunately have ossified, and are hard to introduce.

> So, the UDP underlay is already "wrong", so milking this UDP underlay a bit more to aid deployability _more_ seems fair game.  

I don't agree with this. The closer you keep everything to the target state, the less do you back yourself in the corner.
IMO, every layer violation counts, because every layer violation gets you further away from the target. Just because we have one violation, doesn't mean we should introduce another one. This is a slippery slope, and we should try our best not to slip.

> I admit this is a bit of a foot-in-the-door argument.

good :)

> What we ideally want to achieve is to make it possible to deploy multiple SCION-native applications independently and fully self-contained. There should be no requirement for any non-standard system components to be installed or running, and no coordination between different applications.  
The ebpf-dispatcher approach does not achieve this (generally, not only on mobile platforms); something will need to install this ebpf program (and even needs special privileges for this).  
Requiring an SCMP daemon fails this ideal goal in the same way (needs to be installed and started _somehow_).

I think we all agree that having kernel support would be beautiful.

(quoting @marcfrei response from above)

> I don't know whether and how this could work on Android but at least on iOS/iPadOS there is as far as I can see simply no concept of a user-installable background daemon. There is however the possibility to install and enable a separate app that provides a so-called Network Extension ([https://developer.apple.com/documentation/networkextension](https://developer.apple.com/documentation/networkextension)) which could come close to what is needed here.
>
> We would need to investigate whether it is feasible in principle to implement SCMP daemon functionality based on what the network extension model allows. In addition, there's then also the question how such a network extension would work together with other network extensions. VPN apps (like, e.g. Tailscale) have to be implemented as network extensions and at least VPN extensions are mutually exclusive among each other. Independent of the technical feasibility it also seems to me that such a solution would be rather finicky to manage for typical end-users. Installing a network extension requires entry of the device passcode which already feels a little scary to begin with. It's then very easy to imagine support situations where a SCION-based application doesn't work as expected. The provider of such an app would need to ask the user to check whether the separate SCION SCMP daemon app that provides the network extension is already installed and enabled correctly. The approach thus wouldn't really satisfy Matthias's requirement:
>
> What we ideally want to achieve is to make it possible to deploy multiple SCION-native applications independently and fully self-contained. There should be no requirement for any non-standard system components to be installed or running, and no coordination between different applications.

Okay, that sounds painful. I think quote parsing and an assigned range is probably the most feasible path that does not back us into a corner.
Putting information in the flow ID or reserved bits of the header might work right now, but transitioning away from these hacks will be nasty IMO, and we should try to avoid them.






</blockquote></details>

I realized that we never concluded this discussion. From your last statement, @oncilla, I understand that you agree with moving ahead with the approach outlined in the design document (which is the variant "UDP/SCION Port == Underlay Port && SCMP Quote parsing").
This is what @JordiSubira has implemented in his epic #4344. There, the SCMP Quote parsing exists in two places, in the router and in the "shim" dispatcher which incidentally makes it quite easy to experiment with the different approaches.

If you agree, I'd like to conclude this discussion for now and merge this design document, so Jordi has something more solid to refer to.

Small detail changed based on your first comment: updated the proposed port range in the document so there is no intersection with the range that the dispatcher currently uses. "Legacy" end hosts should thus be mostly unaffected by the port dispatching in the router.

lukedirtwalker (changes requested):
> Reviewed 1 of 3 files at r6, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`doc/dev/design/router-port-dispatch.rst` line 225 at r9](https://reviewable.io/reviews/scionproto/scion/4280#-Nu-6js53f-6QgNwitMF:-Nu-6js53f-6QgNwitMG:undefined) ([raw file](https://github.com/scionproto/scion/blob/6e52ec1f91712bff2b19e8a3bf7f3bdebc354d06/doc/dev/design/router-port-dispatch.rst#L225)):*
> ```reStructuredText
> - Add support for dispatched/forwarded port ranges to the topology.json configuration.
> 
>   As the topology.json parsing is lenient about unkown keys by default, tue updated topology.json file can still be consumed by "legacy" applications.
> ```





_Suggestion:_
```reStructuredText
the
```

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`doc/dev/design/router-port-dispatch.rst` line 227 at r8](https://reviewable.io/reviews/scionproto/scion/4280#-NX9bZ5zDWhB55kH40VJ:-NhRSCse4gI8qBjyqpsZ:bz1pz23) ([raw file](https://github.com/scionproto/scion/blob/1df97866af9b591d4c35c0ab49069416277742c3/doc/dev/design/router-port-dispatch.rst#L227)):*

As mentioned in https://github.com/scionproto/scion/pull/4344#pullrequestreview-1443688080 and discussed in the [TC Implementation meeting on 1st June](https://github.com/scionproto/scion/issues/4327#issuecomment-1573359749), this design document was missing a compatibility mechanism to ensure that applications/services can function even if the port-dispatching feature is only partially available in the routers of the AS.
</blockquote></details>

While I was implementing this. In fact, I already did it, I realized that this shim has to be used "carefully". Since, essentially, it has become a stateless UDP datagram forwarder, no registration from the application is expected to happen. In certain scenarios, one attacker could send packets which reach services that were not intended to be accessible from"remote. Let's say host X with  address X.X.X.X, has two services: SCION service listening on port S and non-scion service listening on port Y. Assuming, network rules allow traffic to X.X.X.X 
___
*[`doc/dev/design/router-port-dispatch.rst` line 166 at r10](https://reviewable.io/reviews/scionproto/scion/4280#-Nu8nKGWAJoDgFJ-Gjfv:-Nu8nKGWAJoDgFJ-Gjfw:b-ps3rg8) ([raw file](https://github.com/scionproto/scion/blob/ac167a959d0cae0cb70fd5819ca1ed6c582e9e14/doc/dev/design/router-port-dispatch.rst#L166)):*
> ```reStructuredText
>             "max": <port>
>          },
>          "<ip-subnet>": { # any number of subnet-specific port-range configurations (in CIDR-notation)
> ```

@lukedirtwalker brought this up in the implementation PR and I realized I never stated my doubts about this idea. Is it really worth it to allow different port ranges per prefix? It is clear that it brings flexibility for the AS to have a clear segmentation between deployments, i.e., you put different deployments in different subnets. However, this brings complexity. I guess what concerns me the most is complexity on the BR. I know this only affects to the last-mile BR, but now the BR also has to do longest matching prefix. Besides, from the point of view of the application, it is more counterintuitive binding to 0.0.0.0:port (it is true now we don't even allow this due to finding out the address of the next hop BR, but we could). This would force us to have some wrapper implementation under the hood that decides on potentially different ports depending on the address of the interfaces if the belong to different subnets.

Note: at the moment the current implementation PR only implements the common port range. I can adapt it if this is the right way to go.

lukedirtwalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 2 unresolved discussions (waiting on @JordiSubira and @oncilla)

___
*[`doc/dev/design/router-port-dispatch.rst` line 166 at r10](https://reviewable.io/reviews/scionproto/scion/4280#-Nu8nKGWAJoDgFJ-Gjfv:-NuSLXhY2ssMPqR5SX4-:b-ctyubi) ([raw file](https://github.com/scionproto/scion/blob/ac167a959d0cae0cb70fd5819ca1ed6c582e9e14/doc/dev/design/router-port-dispatch.rst#L166)):*

@lukedirtwalker brought this up in the implementation PR and I realized I never stated my doubts about this idea. Is it really worth it to allow different port ranges per prefix? It is clear that it brings flexibility for the AS to have a clear segmentation between deployments, i.e., you put different deployments in different subnets. However, this brings complexity. I guess what concerns me the most is complexity on the BR. I know this only affects to the last-mile BR, but now the BR also has to do longest matching prefix. Besides, from the point of view of the application, it is more counterintuitive binding to 0.0.0.0:port (it is true now we don't even allow this due to finding out the address of the next hop BR, but we could). This would force us to have some wrapper implementation under the hood that decides on potentially different ports depending on the address of the interfaces if the belong to different subnets.

Note: at the moment the current implementation PR only implements the common port range. I can adapt it if this is the right way to go.
</blockquote></details>

Do we remember why we brought this up initially? IMO we don't necessarily need it. But I would like to know the reasoning behind why it was brought up.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 2 unresolved discussions (waiting on @JordiSubira, @lukedirtwalker, and @oncilla)

___
*[`doc/dev/design/router-port-dispatch.rst` line 166 at r10](https://reviewable.io/reviews/scionproto/scion/4280#-Nu8nKGWAJoDgFJ-Gjfv:-NuScTDvEg5LEK_n0g2h:b-l1aq13) ([raw file](https://github.com/scionproto/scion/blob/ac167a959d0cae0cb70fd5819ca1ed6c582e9e14/doc/dev/design/router-port-dispatch.rst#L166)):*

Do we remember why we brought this up initially? IMO we don't necessarily need it. But I would like to know the reasoning behind why it was brought up.
</blockquote></details>

This was my suggestion. The idea was mostly to demonstrate that this approach can be tuned flexibly by the AS operator, and that per-host or per-subnet overrides could be applied if port ranges clash or if other issues occur e.g. with future modifications to the end host stack.

Just to clarify regarding applications binding on 0.0.0.0: the subnets-overrides would (only) be configuration for the router; the thought was that end hosts would get "tailored" topology files with only port range relevant to them, so they don't need to explicitly manage this.

I agree that this feature is not immediately useful. We have port ranges that don't overlap with the legacy-dispatcher's ports, which should make this per-host/per-subnet override mostly unnecessary for now. Also, this can still be added in a router implementation later on, so we don't block anything by omitting this in a first pass.
I'll revise the document to remove it.

lukedirtwalker (approved):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 2 unresolved discussions (waiting on @JordiSubira and @oncilla)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 2 unresolved discussions (waiting on @JordiSubira and @oncilla)

___
*[`doc/dev/design/router-port-dispatch.rst` line 166 at r10](https://reviewable.io/reviews/scionproto/scion/4280#-Nu8nKGWAJoDgFJ-Gjfv:-NuSt5DN-QKB-aqlp9AO:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/ac167a959d0cae0cb70fd5819ca1ed6c582e9e14/doc/dev/design/router-port-dispatch.rst#L166)):*

This was my suggestion. The idea was mostly to demonstrate that this approach can be tuned flexibly by the AS operator, and that per-host or per-subnet overrides could be applied if port ranges clash or if other issues occur e.g. with future modifications to the end host stack.

Just to clarify regarding applications binding on 0.0.0.0: the subnets-overrides would (only) be configuration for the router; the thought was that end hosts would get "tailored" topology files with only port range relevant to them, so they don't need to explicitly manage this.

I agree that this feature is not immediately useful. We have port ranges that don't overlap with the legacy-dispatcher's ports, which should make this per-host/per-subnet override mostly unnecessary for now. Also, this can still be added in a router implementation later on, so we don't block anything by omitting this in a first pass.
I'll revise the document to remove it.
</blockquote></details>

Done.

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 1 unresolved discussion (waiting on @JordiSubira)

___
*[`doc/design/router-port-dispatch.rst` line 168 at r3](https://reviewable.io/reviews/scionproto/scion/4280#-NGRYQhJCbZjAdgagtCm:-NuThKLo69Gu94lKUSYq:b-alch2x) ([raw file](https://github.com/scionproto/scion/blob/d0277dce4ba4d58041b46030c871a19a6aa8cdd3/doc/design/router-port-dispatch.rst#L168)):*

I realized that we never concluded this discussion. From your last statement, @oncilla, I understand that you agree with moving ahead with the approach outlined in the design document (which is the variant "UDP/SCION Port == Underlay Port && SCMP Quote parsing").
This is what @JordiSubira has implemented in his epic #4344. There, the SCMP Quote parsing exists in two places, in the router and in the "shim" dispatcher which incidentally makes it quite easy to experiment with the different approaches.

If you agree, I'd like to conclude this discussion for now and merge this design document, so Jordi has something more solid to refer to.

Small detail changed based on your first comment: updated the proposed port range in the document so there is no intersection with the range that the dispatcher currently uses. "Legacy" end hosts should thus be mostly unaffected by the port dispatching in the router.
</blockquote></details>

ack. Resolved from my side.

marcfrei:
*[`doc/dev/design/router-port-dispatch.rst` line 172 at r11](https://reviewable.io/reviews/scionproto/scion/4280#-NuYnxUa-xi3GC-T72BQ:-Nv1jaus017nFeIjCNT5:bz4tfzy) ([raw file](https://github.com/scionproto/scion/blob/5eeee19007b260c0f2479dbdc17c31824529e0b2/doc/dev/design/router-port-dispatch.rst#L172)):*

For well known listener ports this seems sufficient, but I worry a bit about the number of available ephemeral client ports. In one project, we discussed sending out UDP requests to ca. 2k - 10k servers in parallel. With this default allocation of ports, we would need to introduce a somewhat complicated sharing scheme for source ports (net.UDPConns), right?
</blockquote></details>

Resolved in an offline discussion: the port range is intended to get removed or made wider again after an AS is updated completely.

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 2 unresolved discussions (waiting on @JordiSubira and @matzf)

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`doc/dev/design/router-port-dispatch.rst` line 164 at r11](https://reviewable.io/reviews/scionproto/scion/4280#-Nv6bV7D56FVqKe5T7h8:-Nv6bV7D56FVqKe5T7h9:ba231pr) ([raw file](https://github.com/scionproto/scion/blob/5eeee19007b260c0f2479dbdc17c31824529e0b2/doc/dev/design/router-port-dispatch.rst#L164)):*
> ```reStructuredText
>          "min": <port>,
>          "max": <port>
>        }
> ```

Right now I am using the following format based on a Marc's suggestion on the implementation PR (https://reviewable.io/reviews/scionproto/scion/4344#-NqRlyUyA8rSxIAJFYWy):
```
"endhost_port_range": "1024-65535",
```
It also allows the syntax "" and "-" to indicate an empty port range. Where should we change it?



_Code quote:_
```reStructuredText
       "dispatched_ports": {
         "min": <port>,
         "max": <port>
       }
```

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @JordiSubira, @lukedirtwalker, and @marcfrei)

___
*[`doc/dev/design/router-port-dispatch.rst` line 164 at r11](https://reviewable.io/reviews/scionproto/scion/4280#-Nv6bV7D56FVqKe5T7h8:-NvCRDlZ8jEEdwEKyEcS:b-9wl963) ([raw file](https://github.com/scionproto/scion/blob/5eeee19007b260c0f2479dbdc17c31824529e0b2/doc/dev/design/router-port-dispatch.rst#L164)):*

Right now I am using the following format based on a Marc's suggestion on the implementation PR (https://reviewable.io/reviews/scionproto/scion/4344#-NqRlyUyA8rSxIAJFYWy):
```
"endhost_port_range": "1024-65535",
```
It also allows the syntax "" and "-" to indicate an empty port range. Where should we change it?
</blockquote></details>

Done, i've changed it here. I've also added an `"all"` value that is a shorthand for `1-65535`. When documenting the intended procedure for the update, I realized that this will be the value used for most of the time once the update is complete, so it makes sense to have a short value.
___
*[`doc/dev/design/router-port-dispatch.rst` line 172 at r11](https://reviewable.io/reviews/scionproto/scion/4280#-NuYnxUa-xi3GC-T72BQ:-NvCRl80-w_Z25dIfHwU:b-epyxgd) ([raw file](https://github.com/scionproto/scion/blob/5eeee19007b260c0f2479dbdc17c31824529e0b2/doc/dev/design/router-port-dispatch.rst#L172)):*

Resolved in an offline discussion: the port range is intended to get removed or made wider again after an AS is updated completely.
</blockquote></details>

I've added more details on the compatibility mechanisms and the intended update procedure now, as discussed in the chat and in tuesday's call.
The significance of this port range has been somewhat reduced since the first version of this document (which had not yet included the "shim dispatcher"). Now it is really only the range of (mostly ephemeral) ports that can be used by "new" devices/applications without the shim dispatcher, during the transition phase before the update of the AS is complete. I hope this is clearer now  -- admittedly, it had not been entirely clear to me before thinking it through again to write it down.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @JordiSubira, @lukedirtwalker, and @marcfrei)

___
*[`doc/dev/design/router-port-dispatch.rst` line 215 at r11](https://reviewable.io/reviews/scionproto/scion/4280#-NuZ0nVuDwyxvzOPC0kD:-NvCTM5Q8l428EIIuo0z:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/5eeee19007b260c0f2479dbdc17c31824529e0b2/doc/dev/design/router-port-dispatch.rst#L215)):*

unknown
</blockquote></details>

Done.

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 3 unresolved discussions (waiting on @JordiSubira and @matzf)

___
*[`doc/dev/design/router-port-dispatch.rst` line 208 at r13](https://reviewable.io/reviews/scionproto/scion/4280#-NvC_KMd5VC51OXSz0i4:-NvC_YTd6JFAZSwxlULu:b-7ilvfz) ([raw file](https://github.com/scionproto/scion/blob/75c395bf76471f7d8511096c1f356bf28a2f177d/doc/dev/design/router-port-dispatch.rst#L208)):*
> ```reStructuredText
> 
> Conversely, if our long term vision materializes and we'd have SCION support directly built-in to the operating system's network stack, then this workaround becomes obsolete.
> In an optimistic scenario, where there are millions of end hosts running SCION-enabled applications, we can not expect that all devices and applications will updated to the same level of SCION support within a useful time frame.
> ```

typo:

marcfrei (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280#-:-NxE7Ce77Vmzyy_jgEdX:bnfp4nl)* status: all files reviewed, 7 unresolved discussions (waiting on @JordiSubira and @matzf)

___
*[`doc/dev/design/router-port-dispatch.rst` line 153 at r13](https://reviewable.io/reviews/scionproto/scion/4280#-NxDVvTiDc7g9wg7E8z7:-NxDVvTjCoYEmI5rFVei:b-vfefa1) ([raw file](https://github.com/scionproto/scion/blob/75c395bf76471f7d8511096c1f356bf28a2f177d/doc/dev/design/router-port-dispatch.rst#L153)):*
> ```reStructuredText
>   The processing rule above is extended:
> 
>   2. If the underlay UDP/IP destination port determined above is within the port range specified in the topology configuration,
> ```

above, i.e. in processing rule 1,



_Code quote:_
```reStructuredText
above
```
___
*[`doc/dev/design/router-port-dispatch.rst` line 202 at r13](https://reviewable.io/reviews/scionproto/scion/4280#-NxDc33_931smKvLINV0:-NxDc33_931smKvLINV1:b-x62ska) ([raw file](https://github.com/scionproto/scion/blob/75c395bf76471f7d8511096c1f356bf28a2f177d/doc/dev/design/router-port-dispatch.rst#L202)):*
> ```reStructuredText
>    new applications/devices without the shim dispatcher, they can pick the empty range in step 1.,
>    and state 3. is skipped.
> ```

Should we add an additional note that explicitly explains how the transition works for server applications listening on well-known ports? I guess they wouldn't start listening in the dispatched_ports range but just use the shim dispatcher until dispatched_ports gets extended to also cover the well known ports?
___
*[`doc/dev/design/router-port-dispatch.rst` line 216 at r13](https://reviewable.io/reviews/scionproto/scion/4280#-NxDfsfc6iMiDt3Wq7l8:-NxDfsfc6iMiDt3Wq7l9:btpy57d) ([raw file](https://github.com/scionproto/scion/blob/75c395bf76471f7d8511096c1f356bf28a2f177d/doc/dev/design/router-port-dispatch.rst#L216)):*
> ```reStructuredText
> 
> - enable shim dispatcher for services outside of intended ``dispatched_port`` range
> - shrink ``dispatched_ports`` range and configure this on routers and hosts
> ```

When can this step safely be initiated, i.e., when do we know that all hosts in a given AS have enabled the shim dispatchers again?
___
*[`doc/dev/design/router-port-dispatch.rst` line 279 at r13](https://reviewable.io/reviews/scionproto/scion/4280#-NxE6M454idEu658j26F:-NxE6M46BmMfO5beFkAE:b-l0tgoz) ([raw file](https://github.com/scionproto/scion/blob/75c395bf76471f7d8511096c1f356bf28a2f177d/doc/dev/design/router-port-dispatch.rst#L279)):*
> ```reStructuredText
>      Remove the packet dispatching/forwarding functionality from "dispatcher".
>      Only SCMP echo responder remains in dispatcher. Rename to "SCMP Daemon" (scmpd).
>    - set suitable default for port range in ``dispatched_ports`` topology configuration
> ```

configuration.



_Code quote:_
```reStructuredText
configuration
```

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on @JordiSubira, @lukedirtwalker, and @marcfrei)

___
*[`doc/dev/design/router-port-dispatch.rst` line 202 at r13](https://reviewable.io/reviews/scionproto/scion/4280#-NxDc33_931smKvLINV0:-NyP_VRg3YC3EUyNzYav:b-fznogt) ([raw file](https://github.com/scionproto/scion/blob/75c395bf76471f7d8511096c1f356bf28a2f177d/doc/dev/design/router-port-dispatch.rst#L202)):*

Should we add an additional note that explicitly explains how the transition works for server applications listening on well-known ports? I guess they wouldn't start listening in the dispatched_ports range but just use the shim dispatcher until dispatched_ports gets extended to also cover the well known ports?
</blockquote></details>

Good idea, done.
___
*[`doc/dev/design/router-port-dispatch.rst` line 216 at r13](https://reviewable.io/reviews/scionproto/scion/4280#-NxDfsfc6iMiDt3Wq7l8:-Ny4mlbQ18C5dYG214YW:b-e5w7mo) ([raw file](https://github.com/scionproto/scion/blob/75c395bf76471f7d8511096c1f356bf28a2f177d/doc/dev/design/router-port-dispatch.rst#L216)):*

When can this step safely be initiated, i.e., when do we know that all hosts in a given AS have enabled the shim dispatchers again?
</blockquote></details>

Good question, I don't have a good answer. I've added a note to point out this flaw.

marcfrei (approved):
> Reviewed 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: all files reviewed, 2 unresolved discussions (waiting on @JordiSubira)

matzf (commented):
> Dismissed @JordiSubira from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4280)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4281, opened 2022-10-18
control: cached loading of keys/certs from disk for TLS

matzf:
Previously only the keys were loaded from disk, the chains were fetched from the DB. This becomes a problem with certificate renewal (especially when overwriting the key/cert files) because the new key is loaded from the disk but no corresponding certificate is loaded into the DB. This issue was noticed in #4268. Additionally, loading certs from disk for each individual TLS session establishment is not ideal, so we should have a cache.

Removed the separate `X509KeyPairLoader` and replaced it with the almost identical `trust.SignerGen` (extended to allow filtering certificates for key usage). For this `SignerGen`, a `CachingSignerGen` implementation already exists.

Note: The filtering of the certificates for the ext key usage now is implemented in the `SignerGen`, based on a member field `KeyUsage`. We use a separate `SignerGen` and `CachingSignerGen` for each of the three key usages that we need to filter for (any, server auth, and client auth).
Alternatively, the filtering could have been implemented in the DB/Loader level, as part of a `ChainQuery`. This approach was not taken because it would seemingly lead to ambigous behaviour in the `CryptoLoader`, where it would not have been clear whether only matching certificates are inserted to the DB.
Yet another option would have been to expose the `KeyUsage` option to the `SignerGen` interface (something like `signerGen.Generate(ctx, WithExtKeyUsage(ExtKeyUsageServerAuth))`, but this would make the `CachingSignerGen` much more complicated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)


JordiSubira (commented):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`-- commits` line 22 at r1](https://reviewable.io/reviews/scionproto/scion/4281#-NEjxQLg5_I_wnD8H5qH):*
Sorry, I do not completely understand the reason to discard the of filtering as part of the `chainQuery`. The `chainQuery` is only used to query chains and similarly the CryptoLoader only exposes the query method, where would the ambiguity come from.  
___
*[`control/trust.go` line 76 at r1](https://reviewable.io/reviews/scionproto/scion/4281#-NEfpNzE8xt-cSk7L1h2:-NEfpNzE8xt-cSk7L1h3:b5x84db) ([raw file](https://github.com/scionproto/scion/blob/4cba79e6846edee2057b884a69e858e0f1388b89/control/trust.go#L76)):*
> ```Go
> 	db trust.DB,
> 	cfgDir string,
> ) trust.SignerGenerator {
> ```

I would change what is returned by the function or the naming. Either returning a trust.SignerGen or changing the name to `NewSignerCachingSignerGen`. By the way, could we move it to the respective package instead of having it here.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)* status: 18 of 20 files reviewed, 2 unresolved discussions (waiting on @JordiSubira, @matzf, and @oncilla)

___
*[`-- commits` line 22 at r1](https://reviewable.io/reviews/scionproto/scion/4281#-NEjxQLg5_I_wnD8H5qH):*

Sorry, I do not completely understand the reason to discard the of filtering as part of the `chainQuery`. The `chainQuery` is only used to query chains and similarly the CryptoLoader only exposes the query method, where would the ambiguity come from.  
</blockquote></details>

What I meant is this: `CryptoLoader.Chains(ctx, query ChainQuery)` first loads chains from disk, inserts them into the DB, and then return the chains. To me it seems somewhat unclear whether non-matching chains will be inserted into the DB.

Looking at this again, I realize that this is already the case now and adding another filter to `ChainQuery` would not make any difference here.
So, I guess I can move the filter implementation to the DB. I'll update the PR.
___
*[`control/trust.go` line 76 at r1](https://reviewable.io/reviews/scionproto/scion/4281#-NEfpNzE8xt-cSk7L1h2:-NEkBE4-1yBy0iCe2B9T:b-qco7g4) ([raw file](https://github.com/scionproto/scion/blob/4cba79e6846edee2057b884a69e858e0f1388b89/control/trust.go#L76)):*

I would change what is returned by the function or the naming. Either returning a trust.SignerGen or changing the name to `NewSignerCachingSignerGen`. By the way, could we move it to the respective package instead of having it here.
</blockquote></details>

Done.

It's not completely clear to me what should go into `control/trust.go` vs the `control/trust` package. It looks like code that specifies concrete directory names goes to  `control/trust.go`. I'll leave this for now.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)* status: 18 of 20 files reviewed, 2 unresolved discussions (waiting on @JordiSubira, @matzf, and @oncilla)

___
*[`-- commits` line 22 at r1](https://reviewable.io/reviews/scionproto/scion/4281#-NEjxQLg5_I_wnD8H5qH):*

What I meant is this: `CryptoLoader.Chains(ctx, query ChainQuery)` first loads chains from disk, inserts them into the DB, and then return the chains. To me it seems somewhat unclear whether non-matching chains will be inserted into the DB.

Looking at this again, I realize that this is already the case now and adding another filter to `ChainQuery` would not make any difference here.
So, I guess I can move the filter implementation to the DB. I'll update the PR.
</blockquote></details>

Hmm, no, `ChainQuery` and the `DB` interface does not seem like a good fit to filter by key usage. It looks like `ChainQuery` should map directly to a database query and the key usages are not represented in the DB schema.

Another option would be the `trust.Provider` interface with a new `trust.Option` to filter by key usage. This would require even more refactoring though (would need to make `CryptoLoader` a `Provider` or something like that) .

@oncilla can you advise, what do you think is the best place to apply this filter?

JordiSubira (commented):
:lgtm:

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281#-:-NEkfHDIBCLfz3p0sk_k:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

___
*[`-- commits` line 22 at r1](https://reviewable.io/reviews/scionproto/scion/4281#-NEjxQLg5_I_wnD8H5qH):*

Hmm, no, `ChainQuery` and the `DB` interface does not seem like a good fit to filter by key usage. It looks like `ChainQuery` should map directly to a database query and the key usages are not represented in the DB schema.

Another option would be the `trust.Provider` interface with a new `trust.Option` to filter by key usage. This would require even more refactoring though (would need to make `CryptoLoader` a `Provider` or something like that) .

@oncilla can you advise, what do you think is the best place to apply this filter?
</blockquote></details>

Okay, I see, if that's the case, I guess we can have the filter at the `SignerGen` level.

oncilla (changes requested):
> Reviewed 2 of 20 files at r1, 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)* status: all files reviewed, 4 unresolved discussions (waiting on @matzf)

___
*[`-- commits` line 22 at r1](https://reviewable.io/reviews/scionproto/scion/4281#-NEjxQLg5_I_wnD8H5qH):*

Okay, I see, if that's the case, I guess we can have the filter at the `SignerGen` level. 
</blockquote></details>

I think filtering in the database and filtering on the go side is fine.
But given that you will have just one certificate per key in most cases anyway, I don't think it is worth the effort to move filtering into the database.
You won't gain much, but it requires touching a lot of code.

So, filtering in the SignerGen is the better option IMO.
___
*[`private/trust/signer_gen.go` line 35 at r2](https://reviewable.io/reviews/scionproto/scion/4281#-NEo3oy20UUgkY0ua2yG:-NEo3oy20UUgkY0ua2yH:b-s5y2ho) ([raw file](https://github.com/scionproto/scion/blob/acbee26f66d4688a076fef8d5289c6f4672eae0e/private/trust/signer_gen.go#L35)):*
> ```Go
> }
> 
> // SignerGenerator generates signers.
> ```

I would prefer not moving this interface in this package.
We try to use client-side interfaces.

So, if we were to move it in this package, it would go to the tls_handshake.go file where it would actually be referenced. (See note on that below.)
___
*[`private/trust/signer_gen.go` line 46 at r2](https://reviewable.io/reviews/scionproto/scion/4281#-NEo4MGk7N2Jz2JbZSIF:-NEo4MGk7N2Jz2JbZSIG:b-fktuph) ([raw file](https://github.com/scionproto/scion/blob/acbee26f66d4688a076fef8d5289c6f4672eae0e/private/trust/signer_gen.go#L46)):*
> ```Go
> 	DB      DB // FIXME(roosd): Eventually this should use a crypto provider
> 
> 	KeyUsage x509.ExtKeyUsage // If not ExtKeyUsageAny, return Signer with matching certificate
> ```

Add a proper doc comment on this field.
___
*[`private/trust/tls_handshake.go` line 31 at r2](https://reviewable.io/reviews/scionproto/scion/4281#-NEo47SQ7r03kPTsaqWh:-NEo47SQ7r03kPTsaqWi:b-euwywu) ([raw file](https://github.com/scionproto/scion/blob/9cebfa389b6208e7e6a8e7a1abff947e499f8acd/private/trust/tls_handshake.go#L31)):*
> ```Go
> const defaultTimeout = 5 * time.Second
> 
> type X509KeyPairLoader interface {
> ```

I think this interface is the right abstraction.

The TLSCrypto manager does not need to know how the tls.Certificate is loaded.
All it cares about is getting a tls.Certificate.

By changing this interface, you make it harder to use this type generically because you have an additional assumption how certificates are provided.

I would keep this file as is, and let a control service specific type deal with the conversion.
(Because how the tls.Certificates are generated is control service specific behavior.)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)* status: all files reviewed, 3 unresolved discussions (waiting on @oncilla)

___
*[`-- commits` line 22 at r1](https://reviewable.io/reviews/scionproto/scion/4281#-NEjxQLg5_I_wnD8H5qH):*

I think filtering in the database and filtering on the go side is fine.
But given that you will have just one certificate per key in most cases anyway, I don't think it is worth the effort to move filtering into the database.
You won't gain much, but it requires touching a lot of code.

So, filtering in the SignerGen is the better option IMO.

</blockquote></details>

Ok, thanks.
___
*[`private/trust/tls_handshake.go` line 31 at r2](https://reviewable.io/reviews/scionproto/scion/4281#-NEo47SQ7r03kPTsaqWh:-NFDwW3wCBvOms0JjISs:b-6feqw5) ([raw file](https://github.com/scionproto/scion/blob/9cebfa389b6208e7e6a8e7a1abff947e499f8acd/private/trust/tls_handshake.go#L31)):*

I think this interface is the right abstraction.

The TLSCrypto manager does not need to know how the tls.Certificate is loaded.
All it cares about is getting a tls.Certificate.

By changing this interface, you make it harder to use this type generically because you have an additional assumption how certificates are provided.

I would keep this file as is, and let a control service specific type deal with the conversion.
(Because how the tls.Certificates are generated is control service specific behavior.)


</blockquote></details>

Ok, I see what you mean. In this case though, having this interface with `LoadClient/ServerKeyPair`, then the `TLSCryptoManager` simply doesn't provide any useful functionality for `GetCertificate/ClientCertificate`. Using this `TLSCryptoManager` with a custom `X509KeyPairLoader` becomes a fairly convoluted way to specify the certificate-closure for `tls.Config`.
Perhaps a simpler solution is to remove the `GetCertificate` and `GetClientCertificate` functions from `TLSCryptoManager`, making it  focused on only _validating_ the certificates. Providing the certificates we can do straight forwardly in the control service specific code, without any interface definitions here.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)* status: 4 of 21 files reviewed, 3 unresolved discussions (waiting on @JordiSubira and @oncilla)

___
*[`private/trust/tls_handshake.go` line 31 at r2](https://reviewable.io/reviews/scionproto/scion/4281#-NEo47SQ7r03kPTsaqWh:-NFEBKp27Yu8hWd-MQ6U:b2lsc76) ([raw file](https://github.com/scionproto/scion/blob/9cebfa389b6208e7e6a8e7a1abff947e499f8acd/private/trust/tls_handshake.go#L31)):*

Ok, I see what you mean. In this case though, having this interface with `LoadClient/ServerKeyPair`, then the `TLSCryptoManager` simply doesn't provide any useful functionality for `GetCertificate/ClientCertificate`. Using this `TLSCryptoManager` with a custom `X509KeyPairLoader` becomes a fairly convoluted way to specify the certificate-closure for `tls.Config`.
Perhaps a simpler solution is to remove the `GetCertificate` and `GetClientCertificate` functions from `TLSCryptoManager`, making it  focused on only _validating_ the certificates. Providing the certificates we can do straight forwardly in the control service specific code, without any interface definitions here.
</blockquote></details>

I've changed this as described above (i.e. moved the `GetCertificate/ClientCertificate` functions out of `TLSCryptoManager`). Maybe the name of the remaining `TLSCryptoManager` should be changed...

Let me know what you think.

JordiSubira (commented):
> Reviewed 17 of 17 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)* status: all files reviewed, 5 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`private/trust/signer_gen.go` line 44 at r3](https://reviewable.io/reviews/scionproto/scion/4281#-NFErplIDPYWgjkTn1XH:-NFErplIDPYWgjkTn1XI:b-h64ibm) ([raw file](https://github.com/scionproto/scion/blob/6b73ef5f5416425a71587d8df6d4fae6a9528bf6/private/trust/signer_gen.go#L44)):*
> ```Go
> 	// If KeyUsage is not ExtKeyUsageAny, Generate will return a Signer with
> 	// a certificate supporting this usage type (if one exists).
> 	KeyUsage x509.ExtKeyUsage
> ```

Perhaps call it `ExtKeyUsage`
___
*[`private/trust/signer_gen.go` line 117 at r3](https://reviewable.io/reviews/scionproto/scion/4281#-NFEr6Og-TSefCtWt59t:-NFEr6Og-TSefCtWt59u:b6d0l9f) ([raw file](https://github.com/scionproto/scion/blob/6b73ef5f5416425a71587d8df6d4fae6a9528bf6/private/trust/signer_gen.go#L117)):*
> ```Go
> 		return nil, err
> 	}
> 	if s.KeyUsage != x509.ExtKeyUsageAny {
> ```

Just an observation, if we use this comparison we will not be able to filter certificates with "extKeyUsageAny" from certificates which do not contain the extension. Probably this is alright for now, since this class is packed in the private package and we are not using/accepting this "extKeyUsage" but could we check if `s.KeyUsage != nil`
___
*[`private/trust/tls_handshake.go` line 31 at r2](https://reviewable.io/reviews/scionproto/scion/4281#-NEo47SQ7r03kPTsaqWh:-NFEoOTXB2sqOdC8gGeP:bfn4b37) ([raw file](https://github.com/scionproto/scion/blob/9cebfa389b6208e7e6a8e7a1abff947e499f8acd/private/trust/tls_handshake.go#L31)):*

I've changed this as described above (i.e. moved the `GetCertificate/ClientCertificate` functions out of `TLSCryptoManager`). Maybe the name of the remaining `TLSCryptoManager` should be changed...

Let me know what you think.
</blockquote></details>

Since now this class basically implements the verification part for the TLS handshake, I'd call it `TLSVerifier` and I would rename the file name as well.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)* status: 14 of 21 files reviewed, 5 unresolved discussions (waiting on @JordiSubira and @oncilla)

___
*[`private/trust/signer_gen.go` line 44 at r3](https://reviewable.io/reviews/scionproto/scion/4281#-NFErplIDPYWgjkTn1XH:-NFI84AR5KzpP87eMWKp:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6b73ef5f5416425a71587d8df6d4fae6a9528bf6/private/trust/signer_gen.go#L44)):*

Perhaps call it `ExtKeyUsage`
</blockquote></details>

Done.
___
*[`private/trust/signer_gen.go` line 117 at r3](https://reviewable.io/reviews/scionproto/scion/4281#-NFEr6Og-TSefCtWt59t:-NFI8J_4DLy69SVDurVC:b-oypw7q) ([raw file](https://github.com/scionproto/scion/blob/6b73ef5f5416425a71587d8df6d4fae6a9528bf6/private/trust/signer_gen.go#L117)):*

Just an observation, if we use this comparison we will not be able to filter certificates with "extKeyUsageAny" from certificates which do not contain the extension. Probably this is alright for now, since this class is packed in the private package and we are not using/accepting this "extKeyUsage" but could we check if `s.KeyUsage != nil`
</blockquote></details>

Fair point. Unfortunately the `ExtKeyUsage` "enum" does not provide a default value other than `ExtKeyUsageAny` -- using `nil` would only work if we'd make this a pointer. I think this restriction is fine for the moment. 
___
*[`private/trust/tls_handshake.go` line 31 at r2](https://reviewable.io/reviews/scionproto/scion/4281#-NEo47SQ7r03kPTsaqWh:-NGL2II614s9096u4vj0:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9cebfa389b6208e7e6a8e7a1abff947e499f8acd/private/trust/tls_handshake.go#L31)):*

Since now this class basically implements the verification part for the TLS handshake, I'd call it `TLSVerifier` and I would rename the file name as well.
</blockquote></details>

Done.

JordiSubira (commented):
> Reviewed 2 of 7 files at r4, 1 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)* status: 17 of 22 files reviewed, 3 unresolved discussions (waiting on @oncilla)

oncilla (approved):
> Reviewed 2 of 20 files at r1, 11 of 17 files at r3, 4 of 7 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4281)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #1264, opened 2017-10-01
SIG ingress rework

shitz:
This PR adds an `IngressDispatcher` that takes a SIG frame, classifies it based on remote IA -> remote host -> session id and hands it off the the corresponding `IngressWorker` for processing.

Also:
* Moved ingress logic out of `base` and created a new `ingress` package
* Replaced `sync.Pool` and channels with `ringbuf`

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion/1264)

PR #4291, opened 2022-11-08
gateway: fix panic on data port

### LINKED ISSUE #4196: gateway: IngressServer panics when receiving probe packet on data port
When SIG A sends a probe packet to the data port of SIG B, SIG B panics with the error message `frame too short`.

This happens when the AS of SIG B is misconfigured and uses the same address for the SIG's `data_addr` and `probe_addr` fields.

topology file:
```
...
  "sigs": {
    "sig1-ff00_0_111": {
      "ctrl_addr": "127.0.0.18:30256",
      "data_addr": "127.0.0.18:30056",
      "probe_addr": "127.0.0.18:30056"
    }
  }
...
```

fstreun:
When receiving a malformed packet, the gateway should not panic. Rather, the packet should be discarded.

Fixes #4196

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4291)

### DIFF
--- a/gateway/dataplane/ingressserver.go
+++ b/gateway/dataplane/ingressserver.go
@@ -85,6 +85,8 @@ func (d *IngressServer) read(ctx context.Context) error {
 		n := newFrameBufs(frames)
 		for i := 0; i < n; i++ {
 			frame := frames[i].(*frameBuf)
+			// Clear FrameBuf reference
+			frames[i] = nil
 			read, src, err := d.Conn.ReadFrom(frame.raw)
 			if err != nil {
 				logger.Error("IngressServer: Unable to read from external ingress", "err", err)
@@ -93,7 +95,6 @@ func (d *IngressServer) read(ctx context.Context) error {
 				}
 				increaseCounterMetric(d.Metrics.ReceiveExternalError, 1)
 				frame.Release()
-				frames[i] = nil
 				continue
 			}
 			v, ok := src.(*snet.UDPAddr)
@@ -106,7 +107,6 @@ func (d *IngressServer) read(ctx context.Context) error {
 				logger.Info("IngressServer: Frame to short ",
 					"expected", sigHdrSize, "actual", read)
 				frame.Release()
-				frames[i] = nil
 				continue
 			}
 			if frame.raw[0] != 0 {
@@ -115,7 +115,6 @@ func (d *IngressServer) read(ctx context.Context) error {
 				logger.Info("IngressServer: Unsupported SIG protocol version",
 					"supported", 0, "actual", frame.raw[0])
 				frame.Release()
-				frames[i] = nil
 				continue
 			}
 
@@ -126,7 +125,6 @@ func (d *IngressServer) read(ctx context.Context) error {
 			metrics.CounterAdd(metrics.CounterWith(d.Metrics.FrameBytesRecv,
 				"remote_isd_as", v.IA.String()), float64(read))
 			d.dispatch(ctx, frame, v)
-			frames[i] = nil
 		}
 		if time.Since(lastCleanup) >= workerCleanupInterval {
 			d.cleanup()
--- a/gateway/dataplane/ingressserver.go
+++ b/gateway/dataplane/ingressserver.go
@@ -93,38 +93,39 @@ func (d *IngressServer) read(ctx context.Context) error {
 				}
 				increaseCounterMetric(d.Metrics.ReceiveExternalError, 1)
 				frame.Release()
-			} else {
-				switch v := src.(type) {
-				case *snet.UDPAddr:
-					if read < sigHdrSize {
-						metrics.CounterInc(metrics.CounterWith(d.Metrics.FramesDiscarded,
-							"remote_isd_as", v.IA.String(), "reason", "invalid"))
-						logger.Info("IngressServer: Frame to short ",
-							"expected", sigHdrSize, "actual", read)
-						frame.Release()
-					} else if frame.raw[0] != 0 {
-						metrics.CounterInc(metrics.CounterWith(d.Metrics.FramesDiscarded,
-							"remote_isd_as", v.IA.String(), "reason", "invalid"))
-						logger.Info("IngressServer: Unsupported SIG protocol version",
-							"supported", 0, "actual", frame.raw[0])
-						frame.Release()
-					} else {
-						frame.frameLen = read
-						frame.sessId = frame.raw[1]
-						metrics.CounterInc(metrics.CounterWith(d.Metrics.FramesRecv,
-							"remote_isd_as", v.IA.String()))
-						metrics.CounterAdd(metrics.CounterWith(d.Metrics.FrameBytesRecv,
-							"remote_isd_as", v.IA.String()), float64(read))
-						d.dispatch(ctx, frame, v)
-					}
-				default:
-					metrics.CounterInc(metrics.CounterWith(d.Metrics.FramesDiscarded,
-						"reason", "invalid"))
-					logger.Info("IngressServer: Not a valid snet address", "address", src)
-					frame.Release()
-				}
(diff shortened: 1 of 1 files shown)


matzf (changes requested):
Thank you, @fstreun! Some nitpicky remarks added.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4291#-:-NGMFcgb3WIlqMxkDRSI:b-6779xi)* status: all files reviewed, 2 unresolved discussions (waiting on @fstreun)

___
*[`gateway/dataplane/ingressserver.go` line 105 at r1](https://reviewable.io/reviews/scionproto/scion/4291#-NGMEebWAziYwEOgPN3e:-NGMEebWAziYwEOgPN3f:b2gtuuj) ([raw file](https://github.com/scionproto/scion/blob/109f5ba3f6af171634843ac3ab35b47e358c15d0/gateway/dataplane/ingressserver.go#L105)):*
> ```Go
> 							"expected", sigHdrSize, "actual", read)
> 						frame.Release()
> 					} else if frame.raw[0] != 0 {
> ```

Perhaps using `continue` for all the error cases, instead of the (nested) `else`,  would help to improve the readability here. The `frame.Release(); continue;` statements are then also easier to visually check for consistency.
___
*[`gateway/dataplane/ingressserver.go` line 124 at r1](https://reviewable.io/reviews/scionproto/scion/4291#-NGMAwH3A0RAE-_kIzjD:-NGMAwH3A0RAE-_kIzjE:bc18n45) ([raw file](https://github.com/scionproto/scion/blob/109f5ba3f6af171634843ac3ab35b47e358c15d0/gateway/dataplane/ingressserver.go#L124)):*
> ```Go
> 						"reason", "invalid"))
> 					logger.Info("IngressServer: Not a valid snet address", "address", src)
> 					frame.Release()
> ```

Wrong address type is a logic error, panic makes sense here.
In fact, I think it would be fair game to just type assert `v := src.(*snet.UDPAddr)` without the type switch, making this whole thing here a bit easier to parse.

fstreun (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4291)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`gateway/dataplane/ingressserver.go` line 105 at r1](https://reviewable.io/reviews/scionproto/scion/4291#-NGMEebWAziYwEOgPN3e:-NGMSPSAB0ieNX5pExBt:b-lkp5ih) ([raw file](https://github.com/scionproto/scion/blob/109f5ba3f6af171634843ac3ab35b47e358c15d0/gateway/dataplane/ingressserver.go#L105)):*

Perhaps using `continue` for all the error cases, instead of the (nested) `else`,  would help to improve the readability here. The `frame.Release(); continue;` statements are then also easier to visually check for consistency.
</blockquote></details>

I agree. As far as I understand the code also the statement `frames[i] = nil` is required.
___
*[`gateway/dataplane/ingressserver.go` line 124 at r1](https://reviewable.io/reviews/scionproto/scion/4291#-NGMAwH3A0RAE-_kIzjD:-NGMQNPiAcsAWH3HtJxm:b-k9c31r) ([raw file](https://github.com/scionproto/scion/blob/109f5ba3f6af171634843ac3ab35b47e358c15d0/gateway/dataplane/ingressserver.go#L124)):*

Wrong address type is a logic error, panic makes sense here.
In fact, I think it would be fair game to just type assert `v := src.(*snet.UDPAddr)` without the type switch, making this whole thing here a bit easier to parse.
</blockquote></details>

If it is impossible that a packet with another address type can be received here, then I agree.
Otherwise, this would provide an easy DoS vector.

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4291)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @fstreun and @shitz)

___
*[`gateway/dataplane/ingressserver.go` line 105 at r1](https://reviewable.io/reviews/scionproto/scion/4291#-NGMEebWAziYwEOgPN3e:-NGMWlbj8BkLeRpi8OHt:b-riru91) ([raw file](https://github.com/scionproto/scion/blob/109f5ba3f6af171634843ac3ab35b47e358c15d0/gateway/dataplane/ingressserver.go#L105)):*
<details><summary><i>Previously, fstreun wrote…</i></summary><blockquote>

I agree. As far as I understand the code also the statement `frames[i] = nil` is required.
</blockquote></details>

Aha, good point, I missed that. Of course we can add this to all the cases, but that seems a bit awkward. We could move that to the loop iteration clause (`for i := 0; i < n; func(){frames[i]=nil; i++}() {`), but that may be a bit nasty too.

I have to admit I don't actually understand the purpose of this statement. Certainly the frame will not be garbage collected anyway (because it is either still being processed or it was just put back into the ring buffer). If this is just a safe-guard to avoid accidental double-processing and it hampers readability, I'd remove it. Note: this was added by @shitz in #1264. 
___
*[`gateway/dataplane/ingressserver.go` line 124 at r1](https://reviewable.io/reviews/scionproto/scion/4291#-NGMAwH3A0RAE-_kIzjD:-NGMWCqW57FVLSSbtWVn:bmewcg1) ([raw file](https://github.com/scionproto/scion/blob/109f5ba3f6af171634843ac3ab35b47e358c15d0/gateway/dataplane/ingressserver.go#L124)):*
<details><summary><i>Previously, fstreun wrote…</i></summary><blockquote>

If it is impossible that a packet with another address type can be received here, then I agree.
Otherwise, this would provide an easy DoS vector.
</blockquote></details>

Yes, it is my understanding that `(*snet.Conn).ReadFrom` always returns `*snet.UDPAddr`.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4291)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @fstreun)

___
*[`gateway/dataplane/ingressserver.go` line 105 at r1](https://reviewable.io/reviews/scionproto/scion/4291#-NGMEebWAziYwEOgPN3e:-NGMfw1w-6iRCumSNtKx:b1l20r6) ([raw file](https://github.com/scionproto/scion/blob/109f5ba3f6af171634843ac3ab35b47e358c15d0/gateway/dataplane/ingressserver.go#L105)):*
<details><summary><i>Previously, fstreun wrote…</i></summary><blockquote>

Done.

Added it to each case. If this is not satisfactory, please let me know.
</blockquote></details>

Ok. Or maybe just move it to the top of the loop body? Once we have the `frame` variable, we can clear `frames[i]`. :shrug:

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4291)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @fstreun)

fstreun (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4291)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`gateway/dataplane/ingressserver.go` line 105 at r1](https://reviewable.io/reviews/scionproto/scion/4291#-NGMEebWAziYwEOgPN3e:-NGMiOXY-J-mT6sG0oks:b-4izkyc) ([raw file](https://github.com/scionproto/scion/blob/109f5ba3f6af171634843ac3ab35b47e358c15d0/gateway/dataplane/ingressserver.go#L105)):*

Ok. Or maybe just move it to the top of the loop body? Once we have the `frame` variable, we can clear `frames[i]`. :shrug: 
</blockquote></details>

Done.

Thank you.

matzf (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4291)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @fstreun)

PR #4294, opened 2022-11-11
router: add checks for intra-segment forwarding

### LINKED ISSUE #4148: Adding checks to intra-segment forwarding to achieve stronger security
Hi all, and happy SCION Day. Sofia Giampietro and I (Tobias Klenze) from the Information Security group at ETH are formally proving the security of the SCION forwarding protocol. We believe that we have found a way for SCION to achieve stronger guarantees than it currently does given a powerful adversary.

### Already implemented check for segment switching
If a router needs to switch between segments (IsXover), then it checks that the ingress and egress interfaces are combined in a valid way (core-to-child, child-to-core, or child-to-child). In particular, L900-L907:
https://github.com/scionproto/scion/blob/bddecd65d2723ac60727300d47cedb1c3ec17562/go/pkg/router/dataplane.go#L877-L920

### Proposed new check for intra-segment forwarding
We propose a similar check for the case in which the router forwards a packet along a single segment. This check would only allow **child-to-parent**, **parent-to-child**, and **core-to-core** combinations, as well as packets coming from an **internal** interface. Such an explicit check would entail valley freedom. At the end of this post is a draft of how it could look like.

### Security analysis
What is this check useful for? If we assume a benign setting in which all paths being used are authorized, then the check is not necessary, since authorized paths are valley-free by construction.
In an adversarial setting, there are some attacker model in which this check can lead to stronger properties. We will focus here on an attacker model in which there is a malicious end host, who has compromised the AS keys of some ASes. The ASes themselves are assumed to act in an honest way, in particular the routers. The attacker can obtain authorized paths from the path server, and, using the keys of compromised ASes, the attacker can modify them, yielding forwarding paths with valid MACs that are not authorized in the control plane.

Given such a powerful attacker, it is clear that we cannot achieve much: for instance, path authorization goes down the drain if the attacker knows all relevant AS keys. However, we believe that the proposed additional check is sufficient to guarantee two stronger properties than the ones we can currently guarantee:

(i) We currently prove valley-free forwarding, assuming that an attacker does not know any keys of on-path ASes. (We could prove that the AS key of the "valley AS" has to be known, which is slightly stronger, but so far we have not). With the additional check, we think we can show valley-free forwarding even if **all AS keys** are known to the attacker.

(ii) We currently formally prove "loop freedom, unless there exists an on-loop AS whose key is known to the attacker". We think that the additional check would allow us to show that loops can only exist if **all on-loop AS keys** are known to the attacker. (NB: We define a loop as "the same link is traversed twice in the same direction by a packet". Infinite loops are ruled out in SCION due to the limitation to three segments and the maximum header size)

While the attacker model may sound unreasonably quite strong, I don't think it is. Imagine there is a bug in a widely used implementation of the router that would allow an attacker to obtain the AS keys. This attacker could instantly use the obtained keys to create arbitrary, unauthorized forwarding paths. By adding the proposed check, we ensure a baseline of forwarding security (e.g., ruling out valleys) is always achieved, even if keys are compromised.

### Proposed implementation
Rough sketch of how the check could be implemented in the "validateEgressID" function. Disclaimer: IANAGD -- I am not a Go developer
```go
        ingress, egress := p.d.linkTypes[p.ingressID], p.d.linkTypes[pktEgressID]
        if !p.segmentChange {
                // Check that the interface pair is valid within a single segment.
                // No check required if the packet is received from an internal interface.
                switch {
                case ingress == topology.Unset: // Tobias: Is this the correct check to see if an interface is internal?
                        return processResult{}, nil
                case ingress == topology.Core && egress == topology.Core:
                        return processResult{}, nil
                case ingress == topology.Child && egress == topology.Parent:
                        return processResult{}, nil
                case ingress == topology.Parent && egress == topology.Child:
                        return processResult{}, nil
                default: // error handling...
                }
        } else {
                // Check that the interface pair is valid on a segment switch.
                // Having a segment change received from the internal interface is never valid.
                switch {
                case ingress == topology.Core && egress == topology.Child:
                        return processResult{}, nil
                case ingress == topology.Child && egress == topology.Core:
                        return processResult{}, nil
                case ingress == topology.Child && egress == topology.Child:
                        return processResult{}, nil
                default: // error handling...
                }
        }
```

matzf:
Similar to the existing checks for segment switching, add checks for the
case in which the router forwards a packet along a single segment. This
check only allows child-to-parent, parent-to-child, and core-to-core
combinations, as well as packets coming from an internal interface
(peering links are not yet considered).
This explicit check ensures valley-free forwarding, even if an attacker
knows all AS hop keys.
Invalid hops result in an SCMP Error message with type `Parameter
Problem` and code `Invalid Path`.

Add new braccept cases for the invalid hops child-to-child and
parent-to-parent. Additional cases would be child-to-core and
core-to-child, but these cannot be checked in the current braccept
framework.
Adapt existing braccept case SCMPTracerouteEgress, where a child-to-peer
hop was used, which is currently not considered valid (peering links are
not supported). It appears that this test case did not specifically
intend to use a peering link (as the path only consists of a single
segment).

Aside: restructure parameters for packSCMP in router to make calls more
compact and increase readability.

~~Based on #4157~~ 
Resolves #4148.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4294)

### DIFF
--- a/router/dataplane.go
+++ b/router/dataplane.go
@@ -778,8 +778,12 @@ type macBuffers struct {
 	epicInput  []byte
 }
 
-func (p *scionPacketProcessor) packSCMP(scmpH *slayers.SCMP, scmpP gopacket.SerializableLayer,
-	cause error) (processResult, error) {
+func (p *scionPacketProcessor) packSCMP(
+	typ slayers.SCMPType,
+	code slayers.SCMPCode,
+	scmpP gopacket.SerializableLayer,
+	cause error,
+) (processResult, error) {
 
 	// check invoking packet was an SCMP error:
 	if p.lastLayer.NextLayerType() == slayers.LayerTypeSCMP {
@@ -793,11 +797,7 @@ func (p *scionPacketProcessor) packSCMP(scmpH *slayers.SCMP, scmpP gopacket.Seri
 		}
 	}
 
-	rawSCMP, err := p.prepareSCMP(
-		scmpH,
-		scmpP,
-		cause,
-	)
+	rawSCMP, err := p.prepareSCMP(typ, code, scmpP, cause)
 	return processResult{OutPkt: rawSCMP}, err
 }
 
@@ -824,9 +824,8 @@ func (p *scionPacketProcessor) validateHopExpiry() (processResult, error) {
 		return processResult{}, nil
 	}
 	return p.packSCMP(
-		&slayers.SCMP{TypeCode: slayers.CreateSCMPTypeCode(slayers.SCMPTypeParameterProblem,
-			slayers.SCMPCodePathExpired),
-		},
+		slayers.SCMPTypeParameterProblem,
+		slayers.SCMPCodePathExpired,
 		&slayers.SCMPParameterProblem{Pointer: p.currentHopPointer()},
 		serrors.New("expired hop", "cons_dir", p.infoField.ConsDir, "if_id", p.ingressID,
 			"curr_inf", p.path.PathMeta.CurrINF, "curr_hf", p.path.PathMeta.CurrHF),
@@ -842,9 +841,8 @@ func (p *scionPacketProcessor) validateIngressID() (processResult, error) {
 	}
 	if p.ingressID != 0 && p.ingressID != pktIngressID {
 		return p.packSCMP(
-			&slayers.SCMP{
-				TypeCode: slayers.CreateSCMPTypeCode(slayers.SCMPTypeParameterProblem, errCode),
-			},
+			slayers.SCMPTypeParameterProblem,
+			errCode,
 			&slayers.SCMPParameterProblem{Pointer: p.currentHopPointer()},
 			serrors.New("ingress interface invalid",
 				"pkt_ingress", pktIngressID, "router_ingress", p.ingressID),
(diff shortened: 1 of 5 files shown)


JordiSubira (commented):
:lgtm:

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4294#-:-NIbsxrMCgeADDkxPOLK:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

FR4NK-W (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4294#-:-NIkuzjA4AfSaj-Z4hQu:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

___
*[`router/dataplane.go` line 957 at r1](https://reviewable.io/reviews/scionproto/scion/4294#-NIkuLi05GCD_GSMl_W5:-NIkuLi05GCD_GSMl_W6:bv8by5u) ([raw file](https://github.com/scionproto/scion/blob/9fbf132c3cb6823f7500e1acf3dbf7945199116f/router/dataplane.go#L957)):*
> ```Go
> 					"egress_id", pktEgressID, "egress_type", egress))
> 		}
> 	} else {
> ```

The additional nesting is not necessary since all branches in the `!p.segmentChange` block do and have to return.



_Code quote:_
```Go
else {
```

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4294)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @JordiSubira)

___
*[`router/dataplane.go` line 957 at r1](https://reviewable.io/reviews/scionproto/scion/4294#-NIkuLi05GCD_GSMl_W5:-NJkUPE10LjHwk7snzEN:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/9fbf132c3cb6823f7500e1acf3dbf7945199116f/router/dataplane.go#L957)):*

The additional nesting is not necessary since all branches in the `!p.segmentChange` block do and have to return.
</blockquote></details>

Done.

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4294)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4303, opened 2023-01-04
lint: cleanup lint output, consolidate lint steps

matzf:
* Only print verbose lint output if errors occur, in particular for the "spectral" OpenAPI linter.
  Tweak the `tools/quiet` script (equivalent to the `run_silently` function in the `tools/lint.sh` script) in the lint script to retain colored output when printing to a TTY.
* Integrate `golangci-lint` into the the normal `make lint` target
* Use `gofmt`, `lll`, `missspell`, `goheader` from golangci-lint to replace previously used individual lint tools
  Replace running individual lint tools by a run of golangci-lint. This simplifies the scripts and gives us a more consistent way to configure which files are ignored from the linter etc.

  The `//:lint` binary tar is no longer needed. The only remaining linter binary (buf) can be invoked more easily by `bazel run`.

  Fix a few small errors that are newly reported by this setup.
  Also, reformat golangci-lint configuration for errorcheck exceptions, making it more compact and sorted.
  Also, add a `--config=quiet` profile to `.bazelrc` to keep down bazel chatter for various `bazel run` invocations; errors are still reported. Arguably this should just be a commandline option of bazel.
* Move the lint step execution to the Makefile.
  In contrast to the removed `tools/lint.sh` script, this offers a simple way to run only a subset of the lint steps:

  - `make lint` runs all lint steps
  - `make lint-<group>` runs a group of lint steps, for the groups `go, bazel, protobuf, openapi`.
    E.g. `make lint-go`
  - `make lint-<group>-<step>` runs an individual lint step.
    E.g. `make lint-go-golangci`

  Note that in many shells, we get auto-completion of make targets for free.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4303)


oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4303)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`.golangcilint.yml` line 71 at r1](https://reviewable.io/reviews/scionproto/scion/4303#-NKw9lN4EOnsWVNONLyv:-NKw9lN4EOnsWVNONLyw:bo6ap9z) ([raw file](https://github.com/scionproto/scion/blob/23de649e14728d3ef26e56b4f5e8814669c7ce9c/.golangcilint.yml#L71)):*
> ```YAML
> 
>     # list of exceptions to the errcheck check
>     # TODO(matzf): most of these should rather be fixed! It may be ok to be a
> ```

Indeed!
___
*[`acceptance/cmd/sig_ping_acceptance/main.go` line 17 at r1](https://reviewable.io/reviews/scionproto/scion/4303#-NKwAdAu9819cXgUZBc7:-NKwAdAu9819cXgUZBc8:bgxnqdb) ([raw file](https://github.com/scionproto/scion/blob/23de649e14728d3ef26e56b4f5e8814669c7ce9c/acceptance/cmd/sig_ping_acceptance/main.go#L17)):*
From https://tip.golang.org/doc/comment

> A package comment for a command is similar, but it describes the behavior of the program rather than the Go symbols in the package. The first sentence conventionally begins with the name of the program itself, capitalized because it is at the start of a sentence. For example, here is an abridged version of the package comment for gofmt:

Granted it will look a bit awkward, maybe `sig_ping_acceptance` would be better



_Suggestion:_
```Go
// Sig_ping_acceptance runs checks for basic connectivity between AS through the SIG, using
// standard IP ping.
package main
```
___
*[`pkg/scrypto/cms/protocol/protocol.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4303#-NKwBUu06XF7bvTt7KKW:-NKwBUu06XF7bvTt7KKX:b-jvyl65) ([raw file](https://github.com/scionproto/scion/blob/23de649e14728d3ef26e56b4f5e8814669c7ce9c/pkg/scrypto/cms/protocol/protocol.go#L1)):*
> ```Go
> // Copyright 2022 SCION Association
> ```

This code was copied from https://github.com/github/ietf-cms

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4303)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4303#-NKwCGxc5hZLwAkBRAFm:-NKwCGxc5hZLwAkBRAFn:b-yqc5y6) (no related file):*
Nice improvements :100: 

___

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4303)* status: 10 of 12 files reviewed, 2 unresolved discussions (waiting on @oncilla)

___
*[`acceptance/cmd/sig_ping_acceptance/main.go` line 17 at r1](https://reviewable.io/reviews/scionproto/scion/4303#-NKwAdAu9819cXgUZBc7:-NKwZdzf6ROVu0U4APUx:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/23de649e14728d3ef26e56b4f5e8814669c7ce9c/acceptance/cmd/sig_ping_acceptance/main.go#L17)):*

From https://tip.golang.org/doc/comment

> A package comment for a command is similar, but it describes the behavior of the program rather than the Go symbols in the package. The first sentence conventionally begins with the name of the program itself, capitalized because it is at the start of a sentence. For example, here is an abridged version of the package comment for gofmt:

Granted it will look a bit awkward, maybe `sig_ping_acceptance` would be better
</blockquote></details>

Done.
___
*[`pkg/scrypto/cms/protocol/protocol.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4303#-NKwBUu06XF7bvTt7KKW:-NKwXKp87IS2iXU8u-4N:boxm6wk) ([raw file](https://github.com/scionproto/scion/blob/23de649e14728d3ef26e56b4f5e8814669c7ce9c/pkg/scrypto/cms/protocol/protocol.go#L1)):*

This code was copied from https://github.com/github/ietf-cms
</blockquote></details>

Oooh, ooops! I don't remember why I added this, it's already on the exception list anyway. Great catch, thanks! :sweat_smile:

PR #4304, opened 2023-01-09
pkg/log: improve performance of log.Debug()

rohrerj:
I saw that even if the log level is info a call to log.Debug(...) requires on my system 135ns because convertCtx() is executed before checking the log level. Therefore I added a check to the log.Debug(...) function. With the help of the following benchmark tests I was able to reduce the execution time from 135ns to 30ns.

func BenchmarkLog(b *testing.B) {
	err := log.Setup(log.Config{Console: log.ConsoleConfig{Level: "info"}})
	assert.NoError(b, err)
	for i := 0; i < b.N; i++ {
		log.Debug("hello world", "i", i)
	}
}

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4304)

### DIFF
--- a/pkg/log/wrappers.go
+++ b/pkg/log/wrappers.go
@@ -22,7 +22,9 @@ import (
 
 // Debug logs at debug level.
 func Debug(msg string, ctx ...interface{}) {
-	zap.L().Debug(msg, convertCtx(ctx)...)
+	if zap.L().Core().Enabled(zapcore.DebugLevel) {
+		zap.L().Debug(msg, convertCtx(ctx)...)
+	}
 }
 
 // Info logs at info level.
--- a/pkg/log/wrappers.go
+++ b/pkg/log/wrappers.go
@@ -22,26 +22,26 @@ import (
 
 // Debug logs at debug level.
 func Debug(msg string, ctx ...interface{}) {
-	if Enabled(DebugLevel) {
+	if enabled(DebugLevel) {
 		zap.L().Debug(msg, convertCtx(ctx)...)
 	}
 }
 
 // Info logs at info level.
 func Info(msg string, ctx ...interface{}) {
-	if Enabled(InfoLevel) {
+	if enabled(InfoLevel) {
 		zap.L().Info(msg, convertCtx(ctx)...)
 	}
 }
 
 // Error logs at error level.
 func Error(msg string, ctx ...interface{}) {
-	if Enabled(ErrorLevel) {
+	if enabled(ErrorLevel) {
 		zap.L().Error(msg, convertCtx(ctx)...)
 	}
 }
 
-func Enabled(lvl Level) bool {
+func enabled(lvl Level) bool {
 	return zap.L().Core().Enabled(zapcore.Level(lvl))
 }
 
--- a/pkg/log/wrappers.go
+++ b/pkg/log/wrappers.go
@@ -22,19 +22,27 @@ import (
 
 // Debug logs at debug level.
 func Debug(msg string, ctx ...interface{}) {
-	if zap.L().Core().Enabled(zapcore.DebugLevel) {
+	if Enabled(DebugLevel) {
 		zap.L().Debug(msg, convertCtx(ctx)...)
 	}
 }
 
 // Info logs at info level.
 func Info(msg string, ctx ...interface{}) {
-	zap.L().Info(msg, convertCtx(ctx)...)
+	if Enabled(InfoLevel) {
+		zap.L().Info(msg, convertCtx(ctx)...)
+	}
 }
 
 // Error logs at error level.
 func Error(msg string, ctx ...interface{}) {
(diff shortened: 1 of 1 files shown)


rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4304)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`pkg/log/wrappers.go` line 27 at r1](https://reviewable.io/reviews/scionproto/scion/4304#-NLM5gbt8BnYnGSsyjkN:-NLN9uC07hpbsttJUyaV:b-fgbdb6) ([raw file](https://github.com/scionproto/scion/blob/f814667f91563f5db78d7035787d48e1cfc2b198/pkg/log/wrappers.go#L27)):*

Once you've added the same check in `(*logger).Debug` etc. (see comment below), we _could_ replace this implementation by `Root().Debug(msg, ctx...)` to avoid the code duplication.

This might lead to excessive allocation of the wrapper objects, showing up in your benchmark. It would make sense to also change `logger` to be passed around by value instead of by pointer (in particular, change all of `logger`'s methods to take a value-receiver instead of a pointer-receiver) to (hopefully) avoid these allocations. Feel free to do this or just leave it like it is now, either way seems fine for me.
</blockquote></details>

I added the check now also in the (*logger).Debug function. I tried to replace the implementation as you suggested but it lead to really poor performance. I think some additional changes would be required to solve that. Therefore I just left it as it is.
___
*[`pkg/log/wrappers.go` line 32 at r1](https://reviewable.io/reviews/scionproto/scion/4304#-NLM3z6U1Ky7MB2TS5Ob:-NLNAJyp61vcIYBjEpwU:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/f814667f91563f5db78d7035787d48e1cfc2b198/pkg/log/wrappers.go#L32)):*

Same log-level check as in `Debug` should be applied here and in `Error` as well.
</blockquote></details>

Done.
___
*[`pkg/log/wrappers.go` line 77 at r1](https://reviewable.io/reviews/scionproto/scion/4304#-NLM383O7vK30i7MvG-V:-NLNASvMCV-XB6kFpf--:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/f814667f91563f5db78d7035787d48e1cfc2b198/pkg/log/wrappers.go#L77)):*

Let's add the level check here as well, and also in the Info and Error methods of this `logger` type. We can use the existing `Enabled` wrapper function instead of using the `l.logger.Core().Enabled(...)`.
```go
if l.Enabled(DebugLevel) {
   ...
}
```
</blockquote></details>

Done.

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4304)* status: all files reviewed, 1 unresolved discussion (waiting on @rohrerj)

___
*[`pkg/log/wrappers.go` line 44 at r2](https://reviewable.io/reviews/scionproto/scion/4304#-NLPbJRNCbf21-cfzO62:-NLPbJRNCbf21-cfzO63:b-60f0sq) ([raw file](https://github.com/scionproto/scion/blob/7a2717d149622d565f4c6af0b3a253d1b76d73ee/pkg/log/wrappers.go#L44)):*
> ```Go
> }
> 
> func Enabled(lvl Level) bool {
> ```

This makes sense to avoid the repetition, but I'd rather not add this function to the API; the package level log functions are mainly here for convenience, the functionality is identical to using the `Root()` logger. If a user of this package wants to check if a log level is enabled, they can use `Root().Enabled(...)`.

So, keep the function, but make it un-exported (rename it to `enabled`).

rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4304)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`pkg/log/wrappers.go` line 44 at r2](https://reviewable.io/reviews/scionproto/scion/4304#-NLPbJRNCbf21-cfzO62:-NLPi1orDUdjExaY7bhp:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/7a2717d149622d565f4c6af0b3a253d1b76d73ee/pkg/log/wrappers.go#L44)):*

This makes sense to avoid the repetition, but I'd rather not add this function to the API; the package level log functions are mainly here for convenience, the functionality is identical to using the `Root()` logger. If a user of this package wants to check if a log level is enabled, they can use `Root().Enabled(...)`.

So, keep the function, but make it un-exported (rename it to `enabled`).
</blockquote></details>

Done.

rohrerj (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4304#-:-NLPm00Q6ZSSEjCaA8td:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rohrerj)

PR #4307, opened 2023-01-17
db: check tx.Rollback error / serrors: add Join

matzf:
Check returned error of tx.Rollback where this had previously just been a `defer tx.Rollback`. The errcheck linter (rightly) complains about this. The Rollback was previously also be called after a successful commit, in which case it would have returned the error `ErrTxDone`. Now, Rollback is only called in error cases.

Not fixed the `defer tx.Rollback` pattern in colibri code, as this appears to be pretty much dead code by now.

Add a `serrors.Join` helper function to make it easier to return multiple potential errors. This API is modeled closely after `errors.Join`, which will be available in the go standard library in the next release (go 1.20).

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4307)

### DIFF
--- a/.golangcilint.yml
+++ b/.golangcilint.yml
@@ -138,9 +138,7 @@ issues:
              ^private/mgmtapi/segments/api/api.go$|\
              ^private/path/combinator/combinator.go$|\
              ^private/revcache/revcachetest/revcachetest.go$|\
-             ^private/segment/seghandler/storage.go$|\
              ^private/service/statuspages.go$|\
-             ^private/storage/db/sqler.go$|\
              ^private/storage/trust/fspersister/db_test.go$|\
              ^private/svc/internal/ctxconn/ctxconn.go$|\
              ^private/trust/db_inspector.go$|\
--- a/pkg/private/serrors/errors.go
+++ b/pkg/private/serrors/errors.go
@@ -263,6 +263,28 @@ func (e List) MarshalLogArray(ae zapcore.ArrayEncoder) error {
 	return nil
 }
 
+// Join returns an error that wraps the given errors in a List error.
+// Any nil error values are discarded.
+// Join returns nil if errs contains no non-nil values.
+func Join(errs ...error) error {
+	n := 0
+	for _, err := range errs {
+		if err != nil {
+			n++
+		}
+	}
+	if n == 0 {
+		return nil
+	}
+	l := make(List, 0, n)
+	for _, err := range errs {
+		if err != nil {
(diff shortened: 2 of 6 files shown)


oncilla (approved):
:lgtm: (with one non-blocking nit)

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4307#-:-NLz1G369Gmn4F12DBpz:b-k4gw0)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

___
*[`private/segment/seghandler/storage.go` line 80 at r1](https://reviewable.io/reviews/scionproto/scion/4307#-NLz14TKB0wRjCqq-VjU:-NLz14TKB0wRjCqq-VjV:bftczxl) ([raw file](https://github.com/scionproto/scion/blob/08f372fe8374ed7895d8028f524180e7b9fcfa9e/private/segment/seghandler/storage.go#L80)):*
> ```Go
> 		stats, err := tx.Insert(ctx, seg)
> 		if err != nil {
> 			errRollback := tx.Rollback()
> ```

nit: I don't think the new variable is required here.

(non-blocking as it is matter of taste.)

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4307)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4307)* status: 4 of 6 files reviewed, all discussions resolved (waiting on @lukedirtwalker and @oncilla)

___
*[`private/segment/seghandler/storage.go` line 80 at r1](https://reviewable.io/reviews/scionproto/scion/4307#-NLz14TKB0wRjCqq-VjU:-NLz31B9D1HeP5ug3K6w:b-e7pk75) ([raw file](https://github.com/scionproto/scion/blob/08f372fe8374ed7895d8028f524180e7b9fcfa9e/private/segment/seghandler/storage.go#L80)):*

nit: I don't think the new variable is required here.

(non-blocking as it is matter of taste.)
</blockquote></details>

Ah sure, so invoke `tx.Rollback` directly in the `Join` call. Is that what you meant?

oncilla (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4307)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4308, opened 2023-01-17
Handling of extension headers

jcp19:
I believe I may have come across two minor issues related to how extension headers are handled:
- the first one is demonstrated by the tests `TestHopByHopExtnDecodeReuse`. Basically, if we try to decode an HopByHopExtn into a receiver `e`, then we never erase the field `e.Options` and this slice will accumulate all the Options parsed in previous calls to `DecodeFromBytes`. The same seems to happen with `EndToEndExtn` and is demonstrated in the test `TestEndToEndExtnDecodeReuse`.
- in the function `nextHdr` in the router, we match the type of `layer` with both `EndToEndExtn` and `HopByHopExtn`, but judging from the only place from where this function is called, these comparisons can never succeed. Instead, it seems to me that we should compare them to `EndToEndExtnSkipper` and `HopByHopExtnSkipper` (both of which could succeed). These correspond to the types of  `p.e2eLayer` and `p.hbhLayer` in the method `processPkt`, respectively.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4308)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,26 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+# Order is important. The last matching pattern has the most precedence.
+
+# More details are here: https://help.github.com/articles/about-codeowners/
+
+# Human summary:
+# - The entire core team should be involved for changes that affect the wire
+#   format(s) and the main public libraries (snet).
+# - Dominik (oncilla) should be involed in changes that affect cppki.
+# - Either Dominik (oncilla) or Lukas (lukedirtwalker) should be involved for
+#   changes that could impact the compatibility between this implementation and
+#   their proprietary system.
+# - Matthias (matzf) would like to be involved for changes to the router
+antlr/** @oncilla @lukedirtwalker
+control/trust/** @oncilla
+gateway/dataplane/encoder.go @oncilla @lukedirtwalker
+pkg/scrypto/** @oncilla
+pkg/segment/** @oncilla @lukedirtwalker
+pkg/slayers/** @scionproto/scion-core-team
+pkg/snet/** @scionproto/scion-core-team
+private/trust/** @oncilla
+proto/** @oncilla @lukedirtwalker
+router/dataplane.go @matzf
+scion-pki/** @oncilla
+tools/braccept/cases/** @scionproto/scion-core-team
--- a/.golangcilint.yml
+++ b/.golangcilint.yml
@@ -138,9 +138,7 @@ issues:
              ^private/mgmtapi/segments/api/api.go$|\
              ^private/path/combinator/combinator.go$|\
              ^private/revcache/revcachetest/revcachetest.go$|\
-             ^private/segment/seghandler/storage.go$|\
              ^private/service/statuspages.go$|\
-             ^private/storage/db/sqler.go$|\
              ^private/storage/trust/fspersister/db_test.go$|\
              ^private/svc/internal/ctxconn/ctxconn.go$|\
              ^private/trust/db_inspector.go$|\
--- a/pkg/private/serrors/errors.go
+++ b/pkg/private/serrors/errors.go
@@ -263,6 +263,28 @@ func (e List) MarshalLogArray(ae zapcore.ArrayEncoder) error {
 	return nil
 }
 
+// Join returns an error that wraps the given errors in a List error.
+// Any nil error values are discarded.
+// Join returns nil if errs contains no non-nil values.
+func Join(errs ...error) error {
+	n := 0
+	for _, err := range errs {
+		if err != nil {
+			n++
+		}
+	}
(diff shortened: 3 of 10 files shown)


matzf (commented):
These are great findings, thanks for this contribution!

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4308#-:-NM0NZpK8Ffnrh364mAa:bczv5mm)* status: all files reviewed, 1 unresolved discussion (waiting on @jcp19)

___
*[`pkg/slayers/extn.go` line 256 at r1](https://reviewable.io/reviews/scionproto/scion/4308#-NM0IP3n0RZ55vcX9OE5:-NM0IP3n0RZ55vcX9OE6:b-4jrpdn) ([raw file](https://github.com/scionproto/scion/blob/b281645cc36488a58fcfcc4caf59bb67a80b581b/pkg/slayers/extn.go#L256)):*
> ```Go
> 	// Ignore Options that may have been parsed in previous calls to
> 	// DecodeFromBytes (potentially with different values of `data`).
> 	h.Options = nil
> ```

I think the statement to reset the `Options` value, even thought it has previously been missing by mistake, is fairly self-explanatory in this context. I would omit the comment in the interest of brevity, both here and below in EndToEndExtn.
___
*[`router/dataplane.go` line 1669 at r1](https://reviewable.io/reviews/scionproto/scion/4308#-NM0JBur5zJlRPw3VBL1:-NM0JBur5zJlRPw3VBL2:bjrnn5c) ([raw file](https://github.com/scionproto/scion/blob/b281645cc36488a58fcfcc4caf59bb67a80b581b/router/dataplane.go#L1669)):*
> ```Go
> 	case *slayers.EndToEndExtnSkipper:
> 		return v.NextHdr
> 	case *slayers.HopByHopExtnSkipper:
> ```

Great catch! Luckily this only used for diagnostics. 
This functions seems rather ill-conceived now, the type assertions are clearly very brittle. Perhaps we should add a common interface `NextHdr` field instead.

jcp19 (commented):
happy to help :)

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4308#-:-NM0dHH07o8hrfaccXPh:b-gxqcqa)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`pkg/slayers/extn.go` line 256 at r1](https://reviewable.io/reviews/scionproto/scion/4308#-NM0IP3n0RZ55vcX9OE5:-NM0dMUP0UV_qTdMsJVN:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/b281645cc36488a58fcfcc4caf59bb67a80b581b/pkg/slayers/extn.go#L256)):*

I think the statement to reset the `Options` value, even thought it has previously been missing by mistake, is fairly self-explanatory in this context. I would omit the comment in the interest of brevity, both here and below in EndToEndExtn.
</blockquote></details>

Done.

jcp19 (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4308)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`router/dataplane.go` line 1669 at r1](https://reviewable.io/reviews/scionproto/scion/4308#-NM0JBur5zJlRPw3VBL1:-NM0dfPQ9dzX9htlbHt5:bj6yaoh) ([raw file](https://github.com/scionproto/scion/blob/b281645cc36488a58fcfcc4caf59bb67a80b581b/router/dataplane.go#L1669)):*

Great catch! Luckily this only used for diagnostics. 
This functions seems rather ill-conceived now, the type assertions are clearly very brittle. Perhaps we should add a common interface `NextHdr` field instead.
</blockquote></details>

I do agree that this is very brittle. I would however do that in a separate PR (I am a bit in a time-crunch for the VerifiedSCION project right now)

matzf (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4308#-:-NM2tz0w8I94RsqjYoU4:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jcp19)

___
*[`router/dataplane.go` line 1669 at r1](https://reviewable.io/reviews/scionproto/scion/4308#-NM0JBur5zJlRPw3VBL1:-NM2t_4H-f_2LaQwOdUM:b-8nl7j3) ([raw file](https://github.com/scionproto/scion/blob/b281645cc36488a58fcfcc4caf59bb67a80b581b/router/dataplane.go#L1669)):*
<details><summary><i>Previously, jcp19 (João Pereira) wrote…</i></summary><blockquote>

I do agree that this is very brittle. I would however do that in a separate PR (I am a bit in a time-crunch for the VerifiedSCION project right now)
</blockquote></details>

Yes, that's perfectly fine. :thumbsup:  That was also what I had in mind; sorry, I should have phrased this more clearly.

PR #4318, opened 2023-02-16
pkt/spao: Fix computation of header length metadata item

marcfrei:
The length of the SCION header is part of the metadata flowing into the computation of the MAC value for SPAO [0]. The corresponding common header field value is stored in multiples of 4-bytes [1].

This PR makes sure that indeed multiples of 4-bytes are used instead of the actual length in bytes of the SCION header.

The problem did not appear in testing up till now, because all endpoints are based on the same implementation so far.

[0] https://docs.scion.org/en/latest/protocols/authenticator-option.html#authenticated-data

[1] https://docs.scion.org/en/latest/protocols/scion-header.html#common-header

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4318)

### DIFF
--- a/pkg/spao/mac.go
+++ b/pkg/spao/mac.go
@@ -113,7 +113,17 @@ func serializeAuthenticatedData(
 
 	_ = buf[MACBufferSize-1]
 
-	buf[0] = byte(slayers.CmnHdrLen + s.AddrHdrLen() + s.Path.Len())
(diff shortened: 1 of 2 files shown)


matzf (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4318#-:-NOPugfG-BR-f7zSI1GS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marcfrei)

PR #4330, opened 2023-03-14
lint: enable black formatter

oncilla:
Enable the [black formatter](https://github.com/psf/black) for python files.

This will give us consistent python formatting similar to gofmt.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4330)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,9 @@ __pycache__
 
 # VSCode folder
 #########################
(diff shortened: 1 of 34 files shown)


oncilla:
Hm, I was a bit to trigger happy with this one.
Flake8 and black do not like eachother/

PR #4325, opened 2023-03-07
doc: reorganize and extend documentation

matzf:
The documentation has been a somewhat unsorted bag of documents. It was mostly focused on contributors to this project, but some sections did target users of the software, without a clear distinction.

This PR attempts to bring some more structure into this. There are now three separate sections:
   - Technology:
     Explanations and Specifications. Explain SCION on a conceptual level.
     Should introduce concepts that a user or dev needs to make sense of the manuals etc.

     Target audience: anyone (users, developers, outsiders)
   - Reference manuals:
     Target audience: users of this SCION implementation (operator of SCION infrastructure or hosts), users of any of the SCION APIs.
   - Developer section:
     Target audience: contributors to this SCION implementation
     
In particular, the developer section now "hides" the design documents away a bit and the intended workflow for these documents is clarified.

There is still a lot to do, in particular for overview and introductory material and the manuals. I've started to work on the manual for the control service, but I've decided to defer this for a separate PR so that we can finalize the organizational changes.
Note that the content of some design documents, in particular the one on path policies is now relegated to the less findable design document section. The relevant information will be included in the manuals again.

----

A rendered "preview" for this PR can be found here: https://scion--4325.org.readthedocs.build/en/4325/

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4325)

PR #4331, opened 2023-03-14
doc: improve generated scion and scion-pki documentation

oncilla:
Unify the documentation generation command for scion and scion-pki. Instead of rst, produce markdown which is easier to process.

Properly handled references between the individual commands. This way, we even get proper indexing in the toc tree on the side.
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4331)


matzf:
Can we defer this until #4325?
If we add a customized gendoc tool, I'd prefer it to spit out something similar to manually written RST using all the nice sphinx features for layout and cross-referencing of command line options (e.g. [`.. option::`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-option)).

matzf:
I just came across [sphinx-contrib/datatemplates](https://github.com/sphinx-contrib/datatemplates). This allows to write a template `.rst` file and render it using data from yaml/json files. 
Perhaps this could be a way to allow generating the content of commandline docs as we currently do, but allow keeping the markup styling as part of the documentation tree.

PR #4332, opened 2023-03-15
drkey: add generic-protocol derivation to demo and integration test

marcfrei:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)


marcfrei (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @JordiSubira)

___
*[`demo/drkey/main.go` line 56 at r1](https://reviewable.io/reviews/scionproto/scion/4332#-NQeLM-117Ki1hPKNDzM:-NQeTKcR-DLyjOVF60sf:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6ceef8281dc5aa5055bf50dc1381fd35ce455fe5/demo/drkey/main.go#L56)):*

Use 1 as default value.
</blockquote></details>

Done.
___
*[`demo/drkey/test.py` line 126 at r1](https://reviewable.io/reviews/scionproto/scion/4332#-NQeL4UB55kb7zi676Al:-NQeWg_f3eiqXniHc00R:bgzqo45) ([raw file](https://github.com/scionproto/scion/blob/6ceef8281dc5aa5055bf50dc1381fd35ce455fe5/demo/drkey/test.py#L126)):*

If we merge the two functions in one, we have to also change the "server_ip" in L73, otherwise I guess those may differ:
```
# Enable delegation for demo "server", i.e. allow server to
 # access the base secret value from which keys can be derived locally.
 server_ip = self._container_ip("scion_disp_tester_%s" % self.server_isd_as.file_fmt())
```

Alternatively we can use `_server_ip()` for test fetching SV abd `_endhost_ip()`for those fetching HostAS keys
</blockquote></details>

On this line we allow the server to directly access the CS, which is used to fetch the SV. In the other cases, we have to use the IP of the daemon (now called `_endhost_ip`) because the CS checks that the actual source host of a host-AS request matches with the peer sending the request, which is the daemon running on a separate IP in our test topologies. So the two IPs actually have to be different because access happens in one case directly and in the other case via the daemon. Or am I missing something?

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @marcfrei)

___
*[`demo/drkey/test.py` line 126 at r1](https://reviewable.io/reviews/scionproto/scion/4332#-NQeL4UB55kb7zi676Al:-NQeiKQH9fo4PL9ym57t:bd5p8ni) ([raw file](https://github.com/scionproto/scion/blob/6ceef8281dc5aa5055bf50dc1381fd35ce455fe5/demo/drkey/test.py#L126)):*

On this line we allow the server to directly access the CS, which is used to fetch the SV. In the other cases, we have to use the IP of the daemon (now called `_endhost_ip`) because the CS checks that the actual source host of a host-AS request matches with the peer sending the request, which is the daemon running on a separate IP in our test topologies. So the two IPs actually have to be different because access happens in one case directly and in the other case via the daemon. Or am I missing something?
</blockquote></details>

You are right I got confuse by the naming, merging that way would not work because. I suggested changing the description above to avoid "some" confusion. In any case, there is still some confusion due to the  fact that tester host derive secrets for the scion_sd addresses which are different in our test topology.

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @marcfrei)

___
*[`demo/drkey/test.py` line 126 at r1](https://reviewable.io/reviews/scionproto/scion/4332#-NQeL4UB55kb7zi676Al:-NQfN5eR3ykiVMRlpkqE:b1nksbl) ([raw file](https://github.com/scionproto/scion/blob/6ceef8281dc5aa5055bf50dc1381fd35ce455fe5/demo/drkey/test.py#L126)):*

I see, it's confusing indeed. In the latest version, I now  tried to explain the configuration of the two variables `server_ip` and `client_ip` a little better: lines 93..97. What do you think? 
</blockquote></details>

Yes, this is more clear to me. I added before some suggestions but I think they didn't show up. You can also add that explanation of leave it with the last change which reads more easily IMO :)
___
*[`demo/drkey/test.py` line 72 at r4](https://reviewable.io/reviews/scionproto/scion/4332#-NQeknf11TZMcNJ-9Wtv:-NQeknf11TZMcNJ-9Wtw:b-i1g53h) ([raw file](https://github.com/scionproto/scion/blob/82f837ef3f8659ed1ef962b86c3874e0a7a25cc0/demo/drkey/test.py#L72)):*
> ```Python
> 
>         # Enable delegation for demo "server", i.e. allow server to
>         # access the base secret value from which keys can be derived locally.
> ```

# Enable delegation for the demo tester host on the fast side (server side), i.e., it allows
# the tester host to directly request the secret value from which keys can be derived locally for any host.



_Code quote:_
```Python
        # Enable delegation for demo "server", i.e. allow server to
        # access the base secret value from which keys can be derived locally.
```
___
*[`demo/drkey/test.py` line 73 at r4](https://reviewable.io/reviews/scionproto/scion/4332#-NQelE_2B4o7uXHKtDUy:-NQelE_2B4o7uXHKtDUz:b-xxmj0e) ([raw file](https://github.com/scionproto/scion/blob/82f837ef3f8659ed1ef962b86c3874e0a7a25cc0/demo/drkey/test.py#L73)):*
> ```Python
>         # Enable delegation for demo "server", i.e. allow server to
>         # access the base secret value from which keys can be derived locally.
>         server_ip = self._container_ip("scion_disp_tester_%s" % self.server_isd_as.file_fmt())
> ```

tester_ip



_Code quote:_
```Python
server_ip
```

marcfrei (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @JordiSubira)

___
*[`demo/drkey/test.py` line 72 at r4](https://reviewable.io/reviews/scionproto/scion/4332#-NQeknf11TZMcNJ-9Wtv:-NQii81Y5hqXoA_7jJt2:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/82f837ef3f8659ed1ef962b86c3874e0a7a25cc0/demo/drkey/test.py#L72)):*

# Enable delegation for the demo tester host on the fast side (server side), i.e., it allows
# the tester host to directly request the secret value from which keys can be derived locally for any host.
</blockquote></details>

Done.
___
*[`demo/drkey/test.py` line 73 at r4](https://reviewable.io/reviews/scionproto/scion/4332#-NQelE_2B4o7uXHKtDUy:-NQiiA6XBO5Uqr1sAhjD:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/82f837ef3f8659ed1ef962b86c3874e0a7a25cc0/demo/drkey/test.py#L73)):*

tester_ip
</blockquote></details>

Done.

matzf (changes requested):
> Reviewed 1 of 5 files at r1, 1 of 1 files at r3, 2 of 2 files at r4, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: all files reviewed, 3 unresolved discussions (waiting on @JordiSubira and @marcfrei)

___
*[`demo/drkey/main.go` line 107 at r6](https://reviewable.io/reviews/scionproto/scion/4332#-NQyYF1D-wbuTvP5uqEG:-NQyYF1D-wbuTvP5uqEH:b-divc9x) ([raw file](https://github.com/scionproto/scion/blob/3998e87c683c633368b1614b07d1538b8a171639/demo/drkey/main.go#L107)):*
> ```Go
> 		var serverKey drkey.HostHostKey
> 		var durationServer time.Duration
> 		if meta.ProtoId.IsPredefined() {
> ```

Whether the server can fetch the SV does technically not depend on whether this is a predefined protocol, but whether this service is configured to have access to it (which can only be the case for "predefined" protocols). Perhaps the meaning of this would be clearer if we trigger this based on a config or command line flag.
___
*[`demo/drkey/main.go` line 127 at r6](https://reviewable.io/reviews/scionproto/scion/4332#-NQyWt-z05QSjH2ghaDS:-NQyWt-z05QSjH2ghaDT:bdhziro) ([raw file](https://github.com/scionproto/scion/blob/3998e87c683c633368b1614b07d1538b8a171639/demo/drkey/main.go#L127)):*
> ```Go
> 		} else {
> 			// Fetch host-AS key (Level 2) in a real application, this is only done at
> 			// startup or on demand and refreshed for each epoch.
> ```

This doesn't sound right. 
The destination ASes cannot generally be enumerated in advance. Maybe just note that this key can be cached and can be used to derive keys for all hosts in that destination AS.
I don't think we should pretend that this is the same thing as having access to the protocol SV, and that we should include this request in the timing. (I was rather confused when I saw the timing output, as this seemed to imply that we have access to the "GENERIC" SV (which we don't)).

JordiSubira (commented):
> Reviewed 2 of 2 files at r4, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: all files reviewed, 2 unresolved discussions (waiting on @marcfrei and @matzf)

___
*[`demo/drkey/main.go` line 127 at r6](https://reviewable.io/reviews/scionproto/scion/4332#-NQyWt-z05QSjH2ghaDS:-NR3E1RX0lwQ8VaAQGQG:b-5vths1) ([raw file](https://github.com/scionproto/scion/blob/3998e87c683c633368b1614b07d1538b8a171639/demo/drkey/main.go#L127)):*

This doesn't sound right. 
The destination ASes cannot generally be enumerated in advance. Maybe just note that this key can be cached and can be used to derive keys for all hosts in that destination AS.
I don't think we should pretend that this is the same thing as having access to the protocol SV, and that we should include this request in the timing. (I was rather confused when I saw the timing output, as this seemed to imply that we have access to the "GENERIC" SV (which we don't)).
</blockquote></details>

I agree that we cannot generally enumerate remote ASes in advance, but in a similar vain it is also not completely clear what's the benefit of fetching the Host-AS key only based on the timing output. An intermediate solution might be to have the following output:
```
Server host key: xxxx.yyyy.zzzz.wwww
duration w/o cached Host-AS key (localIA:host - remoteIA:remoteHost): t0
duration with cached Host-AS key (localIA:host - remoteIA:remoteHost): t1
```

marcfrei (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @JordiSubira and @matzf)

___
*[`demo/drkey/main.go` line 127 at r6](https://reviewable.io/reviews/scionproto/scion/4332#-NQyWt-z05QSjH2ghaDS:-NR2mOpx27zKQLjufICd:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3998e87c683c633368b1614b07d1538b8a171639/demo/drkey/main.go#L127)):*

This doesn't sound right. 
The destination ASes cannot generally be enumerated in advance. Maybe just note that this key can be cached and can be used to derive keys for all hosts in that destination AS.
I don't think we should pretend that this is the same thing as having access to the protocol SV, and that we should include this request in the timing. (I was rather confused when I saw the timing output, as this seemed to imply that we have access to the "GENERIC" SV (which we don't)).
</blockquote></details>

Done.

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @JordiSubira and @marcfrei)

___
*[`demo/drkey/main.go` line 107 at r6](https://reviewable.io/reviews/scionproto/scion/4332#-NQyYF1D-wbuTvP5uqEG:-NR7R8zj78CMmI4Mhg8n:b-1ue88r) ([raw file](https://github.com/scionproto/scion/blob/3998e87c683c633368b1614b07d1538b8a171639/demo/drkey/main.go#L107)):*

Tried and defined the server-side derivation type based on command-line option. What do you think?
</blockquote></details>

The logic makes sense, but to me the terminology seems not ideal. The difference here is not whether we apply the generic or the specific key  derivation procedure. The difference is whether or not the server uses the "privileged" access to the SV to perform the entire key derivation locally.
To avoid confusing these concepts, a clearer name for the flag could be something like "--fetch-sv" or "--privileged" or something along these lines.
The CS configuration file currently uses the term "delegation" here, maybe this could also be the right terminology for the flag..

Note that the last step of the derivation, deriving the HostHost from the HostAS key works the same for the specific and the generic derivation scheme, and to perform this last step, an application explicitly does not need to know which category the protocol falls under. The organisation of the drkey API does unfortunately not help to clarify this.
The demo could attempt to clarify this by executing the SCMP test case twice, once with the "--fetch-sv" flag and once without it, showing that we'll get the same key!

marcfrei (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @JordiSubira and @matzf)

___
*[`demo/drkey/main.go` line 107 at r6](https://reviewable.io/reviews/scionproto/scion/4332#-NQyYF1D-wbuTvP5uqEG:-NR7vZCGAX6t9aILIg2A:b2181rr) ([raw file](https://github.com/scionproto/scion/blob/3998e87c683c633368b1614b07d1538b8a171639/demo/drkey/main.go#L107)):*

The logic makes sense, but to me the terminology seems not ideal. The difference here is not whether we apply the generic or the specific key  derivation procedure. The difference is whether or not the server uses the "privileged" access to the SV to perform the entire key derivation locally.
To avoid confusing these concepts, a clearer name for the flag could be something like "--fetch-sv" or "--privileged" or something along these lines.
The CS configuration file currently uses the term "delegation" here, maybe this could also be the right terminology for the flag..

Note that the last step of the derivation, deriving the HostHost from the HostAS key works the same for the specific and the generic derivation scheme, and to perform this last step, an application explicitly does not need to know which category the protocol falls under. The organisation of the drkey API does unfortunately not help to clarify this.
The demo could attempt to clarify this by executing the SCMP test case twice, once with the "--fetch-sv" flag and once without it, showing that we'll get the same key!
</blockquote></details>

Went with the "--fetch-sv" suggestion. Does this look better?

Regarding the log output: It currently doesn't look particularly nice. Would a switch to comma separated lines (instead of tab separated) be an option to improve things at least a little bit?

matzf (approved):
> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marcfrei)

___
*[`demo/drkey/main.go` line 107 at r6](https://reviewable.io/reviews/scionproto/scion/4332#-NQyYF1D-wbuTvP5uqEG:-NRXQb863mAL3tjzQHr-:b-2g1buf) ([raw file](https://github.com/scionproto/scion/blob/3998e87c683c633368b1614b07d1538b8a171639/demo/drkey/main.go#L107)):*
Great, thanks!

> Regarding the log output: It currently doesn't look particularly nice. Would a switch to comma separated lines (instead of tab separated) be an option to improve things at least a little bit?

Sure, go for it. Or perhaps putting the timing info onto separate lines, as suggested by Jordi in the other comment thread, could be even better. Nothing depends on this, you can make it as nice as you want :) Just make sure to update the key-extraction regex in the test.py if necessary.

marcfrei (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: 3 of 5 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`demo/drkey/main.go` line 107 at r6](https://reviewable.io/reviews/scionproto/scion/4332#-NQyYF1D-wbuTvP5uqEG:-NRbOp7wAUAUH_49INEP:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3998e87c683c633368b1614b07d1538b8a171639/demo/drkey/main.go#L107)):*

Great, thanks!

> Regarding the log output: It currently doesn't look particularly nice. Would a switch to comma separated lines (instead of tab separated) be an option to improve things at least a little bit?

Sure, go for it. Or perhaps putting the timing info onto separate lines, as suggested by Jordi in the other comment thread, could be even better. Nothing depends on this, you can make it as nice as you want :) Just make sure to update the key-extraction regex in the test.py if necessary.
</blockquote></details>

Done.

matzf (approved):
> Reviewed 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4332)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marcfrei)

PR #4333, opened 2023-03-16
topology: deprecate unused attributes

matzf:
Deprecate unused attributes `authoritative`, `voting` and `issuing` from the topology.json configuration. These are no longer considered by any service or tool and the functionality has been moved into the cs.toml (containing a `[ca]` section for "issuing" ASes) or into the TRC and the configuration of the corresponding tooling.
Only the `core` attribute option remains. In the long run, this should either disappear entirely from the topology configuration or be changed to a boolean flag.

In the internal `private/topology`, remove the unused `CA() bool` from the interface, and simplify the representation of the state from a list of attributes to a simpler `IsCore`.

The goal here is mainly to avoid confusion caused by the unused attributes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4333)

### DIFF
--- a/acceptance/topo_common/topology.json
+++ b/acceptance/topo_common/topology.json
@@ -2,10 +2,7 @@
   "isd_as": "1-ff00:0:110",
   "mtu": 1400,
   "attributes": [
-    "authoritative",
-    "core",
-    "issuing",
-    "voting"
+    "core"
   ],
   "border_routers": {
     "br1-ff00_0_110-1": {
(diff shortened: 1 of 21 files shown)


oncilla (approved):
> Reviewed 21 of 21 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4333)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4343, opened 2023-05-11
router: fix minimal SCMP checks in dataplane tests

matzf:
There is a small number of of packet processing test cases that cover cases resulting in SCMPs.
The packet processor signals SCMPs by returning an `scmpError`. Previously, the test cases just asserted that any error is returned from processing and so they would pass even if an internal error occurred. In this instance there was an internal error while serializing the SCMP message, as the internal IP of the dataplane, the source address of the SCMP message, was not initialized in the test setup.

Add a dummy internal IP address to the test dataplane setup, and change the type of the dataplane's internal IP so that a missing initialization of the IP leads to a panic instead of a runtime error.

Assert that an `scmpError` is returned for the relevant test cases, and check the SCMP type/code.
The full SCMP message is still not checked.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4343)

### DIFF
--- a/router/dataplane.go
+++ b/router/dataplane.go
@@ -99,7 +99,7 @@ type DataPlane struct {
 	linkTypes         map[uint16]topology.LinkType
 	neighborIAs       map[uint16]addr.IA
 	internal          BatchConn
-	internalIP        net.IP
+	internalIP        *net.IPAddr
 	internalNextHops  map[uint16]*net.UDPAddr
 	svc               *services
 	macFactory        func() hash.Hash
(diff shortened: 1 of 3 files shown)


marcfrei (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4343#-:-NVJLpFxBpu3Z3iSoD1d:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4346, opened 2023-06-01
refactor: replace addr.HostAddr hierarchy with tagged union addr.Host

### LINKED ISSUE #4149: Refactor addr.HostAddr types

matzf:
Replace the `addr.HostAddr` type hierarchy, consisting of the `HostAddr` interface and the `HostNone`, `HostIPv4`, `HostIPv6` and `HostSVC` implementations with a single type `addr.Host`, representing all the different host address types as a tagged union ("sum type"). This uses, and follows the spirit of, the `net/netip.Addr` types to represent IP addresses for both IPv4 and IPv6.

The `addr.Host` type is a simple value type, which can be created and copied without heap allocations and can be used in comparisons and as map keys.
In addition to replacing all uses of `addr.HostAddr`, the new `addr.Host` type is used in the `slayers.SCION.Get/SetSrc/DstAddr` interfaces, as well as in the DRKey infrastructure. Using a consistent representation of the addresses allows to get rid of duplicate functionality and a number of address conversions, e.g. in `snet`.
In contrast to the `addr.HostAddr` interface, the new `addr.Host`  does not implement the `net.Addr` interface. In particular, service addresses were passed through layers requiring a `net.Addr`. Use `snet.SVCAddr` in these cases.

Additionally, add a new `addr.Addr` type representing a full SCION address (ISD, AS and host address), including parsing functionality. This definition is identical to the `snet.SCIONAddress` type, which is now only kept as a type alias for compatibility.

Closes #4149.

Future work:
- `pkg/addr` is a public API and should have a more SCION-specific and less clash-likely name, e.g. `saddr`
- replace `snet.SCIONAddress` by  `addr.Addr`
- adopt using `addr.Addr` in the ping, traceroute etc tools, in place of the ill-fitting use of `snet.UDPAddr`.
- add `addr.MustParse{ISD,AS,IA}` for completeness sake (currently in xtest package)
- get rid of the IP-slice to `netip.Addr` conversions by gradually increasing the adoption of `netip.Addr` and related APIs

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4346)


marcfrei (approved):
:lgtm:

> Reviewed 110 of 116 files at r1, 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4346#-:-NXEpNJZDiqmgh_Its3J:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`pkg/addr/host.go` line 95 at r2](https://reviewable.io/reviews/scionproto/scion/4346#-NXEfJNFCs0EfJNj9CH4:-NXEfJNG8jDMITSb06MJ:b-6k2jkb) ([raw file](https://github.com/scionproto/scion/blob/d44237c12abb42bb26af3c0ed11ad415263b472b/pkg/addr/host.go#L95)):*
> ```Go
> // IP address.
> // TODO(matzf): return ok or ...?
> func HostIPFromSlice(ip net.IP) Host {
> ```

- Just `HostFromSlice` instead of `HostIPFromSlice`?
- Re "return ok or ...?": Browsing through the existing code, I see the convenience argument, nevertheless I lean towards passing the ok value to the caller instead of returning an invalid address.
___
*[`pkg/addr/svc.go` line 40 at r2](https://reviewable.io/reviews/scionproto/scion/4346#-NXEhgEqDuNMxZxBniP-:-NXEhgEqDuNMxZxBniP0:b-47bet5) ([raw file](https://github.com/scionproto/scion/blob/d44237c12abb42bb26af3c0ed11ad415263b472b/pkg/addr/svc.go#L40)):*
> ```Go
> // SVC is a SCION service address.
> // A service address is a short identifier for the service type, and a
> // flag-bit to for multicast.
> ```

"flag-bit for multicast"?



_Code quote:_
```Go
flag-bit to for multicast
```

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4346)* status: 95 of 125 files reviewed, all discussions resolved (waiting on @marcfrei and @oncilla)

___
*[`pkg/addr/host.go` line 95 at r2](https://reviewable.io/reviews/scionproto/scion/4346#-NXEfJNFCs0EfJNj9CH4:-NXnoxwS2czzJhQOfSPt:bch8b7q) ([raw file](https://github.com/scionproto/scion/blob/d44237c12abb42bb26af3c0ed11ad415263b472b/pkg/addr/host.go#L95)):*

- Just `HostFromSlice` instead of `HostIPFromSlice`?
- Re "return ok or ...?": Browsing through the existing code, I see the convenience argument, nevertheless I lean towards passing the ok value to the caller instead of returning an invalid address.
</blockquote></details>

Done. I've added the handling of the `ok` result by raising errors where appropriate. This looks a bit clunky in some places now, but as pointed out in the PR description, the plan is to get rid of these type conversions by adopting the netip.Addr etc more broadly.
When the caller needs to handle the `ok` result, this `HostIPFromSlice` is really no longer any more convenient than using `netip.AddrFromSlice` directly, and so I've removed the function entirely.

marcfrei (approved):
:lgtm:

> Reviewed 5 of 17 files at r3, 25 of 25 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4346#-:-NXowxH13oGUfg4bItD9:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

lukedirtwalker (changes requested):
> Reviewed 95 of 116 files at r1, 5 of 17 files at r3, 25 of 25 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4346)* status: all files reviewed, 15 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`pkg/addr/addr.go` line 63 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc3hW_42XXLMnZcrKx:-NYc3hW_42XXLMnZcrKy:b-67sj3j) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/addr.go#L63)):*
> ```Go
> 	return fmt.Sprintf("%s,%s", a.IA, a.Host)
> }
> ```

I think it would be nice if we'd also implement [TextMarshaler](https://pkg.go.dev/encoding@go1.20.5#TextMarshaler) and [TextUnmarshaler](https://pkg.go.dev/encoding@go1.20.5#TextUnmarshaler)
___
*[`pkg/addr/addr.go` line 85 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc3UXe8WFb6pNNIddi:-NYc3UXe8WFb6pNNIddj:b-buyvhr) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/addr.go#L85)):*
> ```Go
> // EXPERIMENTAL: This API is experimental. It may be changed to return a
> // combined AddrPort type instead.
> func ParseAddrPort(s string) (Addr, uint16, error) {
> ```

should we also have an `JoinAddrPort`?
___
*[`pkg/addr/addr.go` line 88 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc4ogU9Q0If-9O3wW8:-NYc4ogV0e-EXZKT0rD7:b-hi14sm) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/addr.go#L88)):*
> ```Go
> 	host, port, err := net.SplitHostPort(s)
> 	if err != nil {
> 		return Addr{}, 0, serrors.WrapStr("invalid address: split host:port", err, "addr", s)
> ```

I wonder if it would be desirable to get rid of the serrors dependency on this package and if it is I would suggest to use `fmt.Errorf` here.
___
*[`pkg/addr/host.go` line 51 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc5iLrC7U8-oLp9FLr:-NYc5iLrC7U8-oLp9FLs:b-5orjcn) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/host.go#L51)):*
> ```Go
> //
> // The zero value is a valid object with Host{}.Type() == HostTypeNone.
> type Host struct {
> ```

I think we should sort the members by size to get the better alignment (and smaller struct size): `ip, svc, t`, I wonder why SVC needs to be a uint16 and not just a uint8 (but it doesn't matter much anyway)
___
*[`pkg/addr/svc.go` line 43 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc94FR3gwJ1lImw_At:-NYc94FR3gwJ1lImw_Au:bjqw5z) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/svc.go#L43)):*
> ```Go
> // The package's SVC constant values are defined without multicast.
> // The Multicast and Base methods set/unset the multicast flag.
> type SVC uint16
> ```

If we anyway rework, can we get rid of the `SVC multicast` ? :P 
___
*[`pkg/addr/svc.go` line 85 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc9l3K3c9qQqhsHeKj:-NYc9l3K3c9qQqhsHeKk:b-6g393p) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/svc.go#L85)):*
> ```Go
> }
> 
> // XXX(matzf): change this to the format accepted by ParseSVC?
> ```

hm I think it's not nice that we don't have a matching pair. I would have expected that whatever I print with `String()` I can also parse.
___
*[`tools/braccept/cases/bfd.go` line 106 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcCjo28XJjKKVI_Nsb:-NYcCjo28XJjKKVI_Nsc:b-pwkgj0) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/bfd.go#L106)):*
> ```Go
> 		SrcIA:        remoteIA,
> 	}
> 	err := scionL.SetSrcAddr(addr.HostIP(netip.MustParseAddr("192.168.13.3")))
> ```

why not use `addr.MustParseHost()` here and below?
___
*[`tools/braccept/cases/child_to_child_xover.go` line 112 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcCrGWBRvTnyOOlli4:-NYcCrGWBRvTnyOOlli5:btbuqx8) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/child_to_child_xover.go#L112)):*
> ```Go
> 	}
> 
> 	if err := scionL.SetSrcAddr(addr.HostIP(netip.MustParseAddr("172.16.5.1"))); err != nil {
> ```

MustParseHost ?
___
*[`tools/braccept/cases/child_to_internal.go` line 101 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcCuMD5P5XTF3cZLDT:-NYcCuMD5P5XTF3cZLDU:b8qllll) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/child_to_internal.go#L101)):*
> ```Go
> 		Path:         sp,
> 	}
> 	if err := scionL.SetSrcAddr(addr.HostIP(netip.MustParseAddr("172.16.4.1"))); err != nil {
> ```

MustParseHost? here and below?
___
*[`tools/braccept/cases/child_to_parent.go` line 109 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcCxReEU8DJ0cAj-MV:-NYcCxReEU8DJ0cAj-MW:b-7ropln) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/child_to_parent.go#L109)):*
> ```Go
> 		Path:         sp,
> 	}
> 	if err := scionL.SetSrcAddr(addr.HostIP(netip.MustParseAddr("172.16.4.1"))); err != nil {
> ```

ditto
___
*[`tools/braccept/cases/internal_to_child.go` line 104 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcCzDX66dnUajVmunf:-NYcCzDX66dnUajVmung:b-7ropln) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/internal_to_child.go#L104)):*
> ```Go
> 		Path:         sp,
> 	}
> 	if err := scionL.SetSrcAddr(addr.MustParseHost("192.168.0.51")); err != nil {
> ```

ditto
___
*[`tools/braccept/cases/jumbo.go` line 108 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcD0nzB99Gd4qZ6oDu:-NYcD0nzB99Gd4qZ6oDv:b-7ropln) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/jumbo.go#L108)):*
> ```Go
> 		Path:         sp,
> 	}
> 	if err := scionL.SetSrcAddr(addr.MustParseHost("172.16.3.1")); err != nil {
> ```

ditto
___
*[`tools/braccept/cases/onehop.go` line 82 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcD2_w669WLpb2SNGS:-NYcD2_w669WLpb2SNGT:b-7ropln) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/onehop.go#L82)):*
> ```Go
> 		Path:         ohp,
> 	}
> 	if err := scionL.SetSrcAddr(addr.MustParseHost("172.16.4.1")); err != nil {
> ```

ditto
___
*[`tools/braccept/cases/parent_to_child.go` line 108 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcD4hn0iVzoAosbzUe:-NYcD4hn0iVzoAosbzUf:b-7ropln) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/parent_to_child.go#L108)):*
> ```Go
> 		Path:         sp,
> 	}
> 	if err := scionL.SetSrcAddr(addr.MustParseHost("172.16.3.1")); err != nil {
> ```

ditto
___
*[`tools/braccept/cases/parent_to_internal.go` line 95 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcD5XTDpbgt0MGD6ov:-NYcD5XTDpbgt0MGD6ow:b-7ropln) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/parent_to_internal.go#L95)):*
> ```Go
> 		Path:         sp,
> 	}
> 	if err := scionL.SetSrcAddr(addr.MustParseHost("172.16.3.1")); err != nil {
> ```

ditto

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4346)* status: all files reviewed, 15 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/addr/addr.go` line 63 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc3hW_42XXLMnZcrKx:-NYvvwiy-QLhaPv6-UEq:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/addr.go#L63)):*

I think it would be nice if we'd also implement [TextMarshaler](https://pkg.go.dev/encoding@go1.20.5#TextMarshaler) and [TextUnmarshaler](https://pkg.go.dev/encoding@go1.20.5#TextUnmarshaler)
</blockquote></details>

Done.
___
*[`pkg/addr/addr.go` line 85 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc3UXe8WFb6pNNIddi:-NYvxxUbBglqY6iiIkNK:b-jndwij) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/addr.go#L85)):*

should we also have an `JoinAddrPort`?
</blockquote></details>

Done. 
I thought that the "Join" name is a bit close to `net.JoinHostPort` which also works on scion addresses (in the form of strings). I would have chosen `FmtAddrPort` but this package already contains a few `FormatXY` functions, so I went with `FormatAddrPort`.
___
*[`pkg/addr/addr.go` line 88 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc4ogU9Q0If-9O3wW8:-NYw6jVj17RINuKam0xy:b-eqr992) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/addr.go#L88)):*

I wonder if it would be desirable to get rid of the serrors dependency on this package and if it is I would suggest to use `fmt.Errorf` here.
</blockquote></details>

Ok to look into this as a follow up? The serrors package is throughout `addr` and this would all need to change as we don't gain anything by using a different error mechanism only in the new functionality. 
___
*[`pkg/addr/host.go` line 51 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc5iLrC7U8-oLp9FLr:-NYvyo5766m60wHeELcs:blic3tx) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/host.go#L51)):*

I think we should sort the members by size to get the better alignment (and smaller struct size): `ip, svc, t`, I wonder why SVC needs to be a uint16 and not just a uint8 (but it doesn't matter much anyway)
</blockquote></details>

Excellent, done. 

The uint16 is really a bit weird. The address field in the packet header is 4 bytes wide and it currently only uses the upper two bytes.
___
*[`pkg/addr/svc.go` line 43 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc94FR3gwJ1lImw_At:-NYw3ZLH3cBh_vGW9jLP:b2w8qhk) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/svc.go#L43)):*

If we anyway rework, can we get rid of the `SVC multicast` ? :P 
</blockquote></details>

I'm happy to work more on this, and I agree that the service addresses can be used better. But such a functional change to the service addresses themselves should perhaps be discussed separately.
___
*[`pkg/addr/svc.go` line 85 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc9l3K3c9qQqhsHeKj:-NYw15Cf9UzWuavXRskn:b-3cgupj) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/svc.go#L85)):*

hm I think it's not nice that we don't have a matching pair. I would have expected that whatever I print with `String()` I can also parse.
</blockquote></details>

Done. It now only contains the hex representation of the service identifier in cases where it's unknown. This format cannot currently be parsed back.
___
*[`tools/braccept/cases/bfd.go` line 106 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcCjo28XJjKKVI_Nsb:-NYw2rs53_9VPS4RxVe3:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/bfd.go#L106)):*

why not use `addr.MustParseHost()` here and below?
</blockquote></details>

Done.
___
*[`tools/braccept/cases/child_to_child_xover.go` line 112 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcCrGWBRvTnyOOlli4:-NYw38ePAwpDf0vdDUDn:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/child_to_child_xover.go#L112)):*

MustParseHost ?
</blockquote></details>

Done.
___
*[`tools/braccept/cases/child_to_internal.go` line 101 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcCuMD5P5XTF3cZLDT:-NYw39gIDWj0NNVUl_tY:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/child_to_internal.go#L101)):*

MustParseHost? here and below?
</blockquote></details>

Done.
___
*[`tools/braccept/cases/child_to_parent.go` line 109 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcCxReEU8DJ0cAj-MV:-NYw3M1k3mgxhOYJPRNY:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/child_to_parent.go#L109)):*

ditto
</blockquote></details>

Done.
___
*[`tools/braccept/cases/internal_to_child.go` line 104 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcCzDX66dnUajVmunf:-NYw3N71Ew9Do5XabBH8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/internal_to_child.go#L104)):*

ditto
</blockquote></details>

Done.
___
*[`tools/braccept/cases/jumbo.go` line 108 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcD0nzB99Gd4qZ6oDu:-NYw3RI86my9EQExFBiy:b-ln9uub) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/jumbo.go#L108)):*

ditto
</blockquote></details>

Done (was already MustParseHost here)
___
*[`tools/braccept/cases/onehop.go` line 82 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcD2_w669WLpb2SNGS:-NYw3UPR5pf8ggKOLTN6:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/onehop.go#L82)):*

ditto
</blockquote></details>

Done.
___
*[`tools/braccept/cases/parent_to_child.go` line 108 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcD4hn0iVzoAosbzUe:-NYw3UzW62eRPK_7oRQV:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/parent_to_child.go#L108)):*

ditto
</blockquote></details>

Done.
___
*[`tools/braccept/cases/parent_to_internal.go` line 95 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYcD5XTDpbgt0MGD6ov:-NYw3Vb-9dp5hl5074m-:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/tools/braccept/cases/parent_to_internal.go#L95)):*

ditto
</blockquote></details>

Done.

lukedirtwalker (approved):
:lgtm: 

> Reviewed 11 of 11 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4346#-:-NYwkI6v4KuJt-IgGLcH:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

___
*[`pkg/addr/svc.go` line 43 at r4](https://reviewable.io/reviews/scionproto/scion/4346#-NYc94FR3gwJ1lImw_At:-NYwjrm38Alg6x2xjR9L:b-pbdqz3) ([raw file](https://github.com/scionproto/scion/blob/62ab7dec211b419f48a3a81e9b0302a46cac1d12/pkg/addr/svc.go#L43)):*

I'm happy to work more on this, and I agree that the service addresses can be used better. But such a functional change to the service addresses themselves should perhaps be discussed separately.
</blockquote></details>

Agreed.

lukedirtwalker (approved):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4346)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

matzf:
Thanks for the reviews and feedback so far! 
Since this PR implements a proposal that has not been formally accepted, I will add the "final comment period" label on #4149 and leave this PR open until the 10 day period has elapsed.

lukedirtwalker (approved):
> Reviewed 30 of 30 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4346)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

PR #4347, opened 2023-06-01
doc: consistent naming of the Traffic Class field

matzf:
The Traffic Class field used to be abbreviated by QoS, an unnecessary alias. Traffic Class is also the name used for the equivalent field in the IPv6 header.

Also, fix the documented NextHeader value for the SCMP protocol.


oncilla (approved):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4347)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4368, opened 2023-07-19
border router: implement slow path and configurability

rohrerj:
This is the second pullrequest that also includes the part of the originally planned third pullrequest with the goal to restructure the border router as described in the [Design Document](https://github.com/scionproto/scion/blob/master/doc/dev/design/BorderRouter.rst).

This pullrequest contains the implementation for the slow path and the configurability. All code paths that end up calling packSCMP() are treated as slow-path. The SCMP state together with the raw packet is handed over to slow-path processors that will then parse the header again and then pack the SCMP packet and forward them to the forwarder.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4368)


rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4368)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`router/dataplane.go` line 546 at r1](https://reviewable.io/reviews/scionproto/scion/4368#-N_hoe2-7l4KfH4OrlP9:-N_hoe2-7l4KfH4OrlPA:b5sut2m) ([raw file](https://github.com/scionproto/scion/blob/ee93b2727baae27f9914f68aaad1c71adcc479f3/router/dataplane.go#L546)):*
> ```Go
> 	slowQs := make([]chan packet, cfg.NumSlowPathProcessors)
> 	for i := 0; i < cfg.NumSlowPathProcessors; i++ {
> 		slowQs[i] = make(chan packet, 16)
> ```

What queue size do you consider appropriate here?

PR #4369, opened 2023-07-21
cppki: limit segment expiration by certificate lifetime

### LINKED ISSUE #4286: cppki: require segment lifetime fully covered by certificate chain
# Proposal

## Summary

Enforce the full lifetime of a segment/beacon is covered by the validity time of
at least one certificate chain.

### Problem

Currently, we do not enforce that the certificate chain that is used to verify a
segment is covering the full lifetime of the segment. Rather, we check that the
certificate chain is valid at verification time. This can cause issues if the
signing entity is failing to renew its certificate chains and issues segments
that with a lifetime that is not covered by any certificate chain. If the
segment life time is covered partially by a certificate chain, it can be verified
and will be inserted into segment/beacon storage. If there is a path segment request
that matches this segment, it will be part of the response. As soon as the
certificate chain expires, the segment is no longer verifiable. This causes
all future requests to fail with signature verification errors.

### How did we end up here?

Initially, there was an idea floating around that we could renew a certificate
chain for the same public key. In that world, having more lax verification rules
is nice, because the signing entity can create segments with longer lifetimes
even though its certificate chain has not been renewed yet. If the signing
entity succeeds to renew the certificate chain before the previous one expired,
the segment can be verified the whole time, without any issues.

However, this is mostly a theoretical advantage that comes with complexity
drawbacks. There are many more combinations between segment lifetime and
certificate chain validity time. The system also becomes a lot harder to reason
about. Furthermore, in practice each certificate chain authenticates a fresh
public key for better key hygiene.

## Solution

There is two sides to this problem. The signing entity and the relying party. To
make this a non-breaking change, we can start with adapting the behavior of
signing entity to ensure that no segments are created with lifetimes that are
not covered by a certificate chain. At a later stage, we can update relying
parties to not accept segments that are not covered by a certificate chain.

### Signing entity

The change to the signing entity is fairly small. When creating the hop fields
and peer fields for the AS entry, the expiration time should be chosen such that
the full lifetime is covered by the available certificate chain.

### Relying party

The change to the relying party is again fairly small. When a segment is
verified, the relying party additionally checks that the lifetime is fully
covered by the validity of the certificate chain. In theory, this is a breaking
change. However, in practice, segments that will be rejected by this change only
stem from cases where the certificate renewal fails.

## Alternatives

As an alternative, we could change the segment lookup and beaconing behavior to
filter out segments that can not be verified at the current time. This would
limit the effects of a missing certificate chain for a segment to one entity.

(Side-note: This check would be good even if we go with the proposal.)

However, this is not enough to behave correctly. Entities need to proactively
fetch certificate chains to fill in the gaps, as they might not learn the new
certificate chains passively. This is a lot of additional complexity that is
currently not implemented and will be hard to get right. It is also hard to
justify this additional complexity if in practice that feature will never be
used due to the rotated private keys on every certificate.

uniquefine:
Update the `beaconing.DefaultExtender` to not create Segments with expiration date that is later than the current AS certificate expiration.

Instead the extender uses `min(segmentExpiration,certificateExpiration)` when creating an AS Hop.

Contributes to #4286

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)

### DIFF
--- a/control/beaconing/BUILD.bazel
+++ b/control/beaconing/BUILD.bazel
@@ -37,6 +37,7 @@ go_library(
         "//private/segment/verifier:go_default_library",
         "//private/topology:go_default_library",
         "//private/tracing:go_default_library",
+        "//private/trust:go_default_library",
         "@com_github_opentracing_opentracing_go//:go_default_library",
     ],
 )
@@ -70,6 +71,7 @@ go_test(
         "//pkg/scrypto/signed:go_default_library",
         "//pkg/segment:go_default_library",
         "//pkg/segment/extensions/staticinfo:go_default_library",
+        "//pkg/slayers/path:go_default_library",
         "//pkg/slayers/path/scion:go_default_library",
         "//pkg/snet:go_default_library",
         "//pkg/snet/addrutil:go_default_library",
--- a/control/beaconing/extender.go
+++ b/control/beaconing/extender.go
@@ -23,13 +23,21 @@ import (
 	"github.com/scionproto/scion/control/ifstate"
 	"github.com/scionproto/scion/pkg/addr"
 	"github.com/scionproto/scion/pkg/log"
+	"github.com/scionproto/scion/pkg/metrics"
 	"github.com/scionproto/scion/pkg/private/serrors"
 	"github.com/scionproto/scion/pkg/private/util"
 	seg "github.com/scionproto/scion/pkg/segment"
 	"github.com/scionproto/scion/pkg/segment/extensions/digest"
 	"github.com/scionproto/scion/pkg/segment/extensions/epic"
 	"github.com/scionproto/scion/pkg/slayers/path"
+	"github.com/scionproto/scion/private/trust"
 )
+
+// SignerGen generates signers and returns their expiration time.
+type SignerGen interface {
+	// Generate generates a signer it.
+	Generate(ctx context.Context) (trust.Signer, error)
+}
 
 // Extender extends path segments.
 type Extender interface {
@@ -44,8 +52,8 @@ type Extender interface {
 type DefaultExtender struct {
 	// IA is the local IA
 	IA addr.IA
-	// Signer is used to sign path segments.
-	Signer seg.Signer
+	// SignerGen is used to sign path segments.
+	SignerGen SignerGen
 	// MAC is used to calculate the hop field MAC.
 	MAC func() hash.Hash
 	// Intfs holds all interfaces in the AS.
@@ -60,6 +68,11 @@ type DefaultExtender struct {
 	StaticInfo func() *StaticInfoCfg
 	// EPIC defines whether the EPIC authenticators should be added when the segment is extended.
 	EPIC bool
+
+	// SegmentExpirationDeficient is a gauge that is set to 1 if the expiration time of the segment
+	// is below the maximum expiration time. This happens when the signer expiration time is lower
+	// than the maximum segment expiration time.
+	SegmentExpirationDeficient metrics.Gauge
 }
 
 // Extend extends the beacon with hop fields.
@@ -85,8 +98,27 @@ func (s *DefaultExtender) Extend(
 	}
 	ts := pseg.Info.Timestamp
 
+	signer, err := s.SignerGen.Generate(ctx)
+	if err != nil {
+		return serrors.WrapStr("getting signer", err)
+	}
+	// Make sure the hop expiration time is not longer than the signer expiration time.
+	expTime := s.MaxExpTime()
+	if ts.Add(path.ExpTimeToDuration(expTime)).After(signer.Expiration) {
+		metrics.GaugeSet(s.SegmentExpirationDeficient, 1)
+		var err error
+		expTime, err = path.ExpTimeFromDuration(signer.Expiration.Sub(ts))
+		if err != nil {
+			return serrors.WrapStr(
+				"calculating expiry time from signer expiration time", err,
+				"signer_expiration", signer.Expiration,
+			)
+		}
+	} else {
+		metrics.GaugeSet(s.SegmentExpirationDeficient, 0)
+	}
 	hopBeta := extractBeta(pseg)
-	hopEntry, epicHopMac, err := s.createHopEntry(ingress, egress, ts, hopBeta)
+	hopEntry, epicHopMac, err := s.createHopEntry(ingress, egress, expTime, ts, hopBeta)
 	if err != nil {
 		return serrors.WrapStr("creating hop entry", err)
 	}
@@ -104,7 +136,7 @@ func (s *DefaultExtender) Extend(
 	// is traversed.
 
 	peerBeta := hopBeta ^ binary.BigEndian.Uint16(hopEntry.HopField.MAC[:2])
-	peerEntries, epicPeerMacs, err := s.createPeerEntries(egress, peers, ts, peerBeta)
+	peerEntries, epicPeerMacs, err := s.createPeerEntries(egress, peers, expTime, ts, peerBeta)
 	if err != nil {
 		return err
 	}
@@ -143,7 +175,7 @@ func (s *DefaultExtender) Extend(
 		}
 	}
 
-	if err := pseg.AddASEntry(ctx, asEntry, s.Signer); err != nil {
+	if err := pseg.AddASEntry(ctx, asEntry, signer); err != nil {
 		return err
 	}
 	if egress == 0 {
@@ -153,12 +185,12 @@ func (s *DefaultExtender) Extend(
 }
 
 func (s *DefaultExtender) createPeerEntries(egress uint16, peers []uint16,
-	ts time.Time, beta uint16) ([]seg.PeerEntry, [][]byte, error) {
+	expTime uint8, ts time.Time, beta uint16) ([]seg.PeerEntry, [][]byte, error) {
 
 	peerEntries := make([]seg.PeerEntry, 0, len(peers))
 	peerEpicMacs := make([][]byte, 0, len(peers))
 	for _, peer := range peers {
-		peerEntry, epicMac, err := s.createPeerEntry(peer, egress, ts, beta)
+		peerEntry, epicMac, err := s.createPeerEntry(peer, egress, expTime, ts, beta)
 		if err != nil {
 			log.Debug("Ignoring peer link upon error",
 				"task", s.Task, "peer_interface", peer, "err", err)
@@ -170,15 +202,20 @@ func (s *DefaultExtender) createPeerEntries(egress uint16, peers []uint16,
 	return peerEntries, peerEpicMacs, nil
 }
 
-func (s *DefaultExtender) createHopEntry(ingress, egress uint16, ts time.Time,
-	beta uint16) (seg.HopEntry, []byte, error) {
+func (s *DefaultExtender) createHopEntry(
+	ingress,
+	egress uint16,
+	expTime uint8,
+	ts time.Time,
+	beta uint16,
+) (seg.HopEntry, []byte, error) {
 
 	remoteInMTU, err := s.remoteMTU(ingress)
 	if err != nil {
 		return seg.HopEntry{}, nil, serrors.WrapStr("checking remote ingress interface (mtu)", err,
 			"interfaces", ingress)
 	}
-	hopF, epicMac := s.createHopF(ingress, egress, ts, beta)
+	hopF, epicMac := s.createHopF(ingress, egress, expTime, ts, beta)
 	return seg.HopEntry{
 		IngressMTU: int(remoteInMTU),
 		HopField: seg.HopField{
@@ -190,7 +227,7 @@ func (s *DefaultExtender) createHopEntry(ingress, egress uint16, ts time.Time,
 	}, epicMac, nil
 }
 
-func (s *DefaultExtender) createPeerEntry(ingress, egress uint16, ts time.Time,
+func (s *DefaultExtender) createPeerEntry(ingress, egress uint16, expTime uint8, ts time.Time,
 	beta uint16) (seg.PeerEntry, []byte, error) {
 
 	remoteInIA, remoteInIfID, remoteInMTU, err := s.remoteInfo(ingress)
@@ -198,7 +235,7 @@ func (s *DefaultExtender) createPeerEntry(ingress, egress uint16, ts time.Time,
 		return seg.PeerEntry{}, nil, serrors.WrapStr("checking remote ingress interface", err,
 			"ingress_interface", ingress)
 	}
-	hopF, epicMac := s.createHopF(ingress, egress, ts, beta)
+	hopF, epicMac := s.createHopF(ingress, egress, expTime, ts, beta)
 	return seg.PeerEntry{
 		PeerMTU:       int(remoteInMTU),
 		Peer:          remoteInIA,
@@ -259,10 +296,9 @@ func (s *DefaultExtender) remoteInfo(ifid uint16) (
 	return topoInfo.IA, topoInfo.RemoteID, topoInfo.MTU, nil
 }
 
-func (s *DefaultExtender) createHopF(ingress, egress uint16, ts time.Time,
+func (s *DefaultExtender) createHopF(ingress, egress uint16, expTime uint8, ts time.Time,
 	beta uint16) (path.HopField, []byte) {
 
-	expTime := s.MaxExpTime()
 	input := make([]byte, path.MACBufferSize)
 	path.MACInput(beta, util.TimeToSecs(ts), expTime, ingress, egress, input)
 
--- a/control/beaconing/extender_test.go
+++ b/control/beaconing/extender_test.go
@@ -36,7 +36,9 @@ import (
 	cryptopb "github.com/scionproto/scion/pkg/proto/crypto"
 	"github.com/scionproto/scion/pkg/scrypto"
 	seg "github.com/scionproto/scion/pkg/segment"
+	"github.com/scionproto/scion/pkg/slayers/path"
 	"github.com/scionproto/scion/private/topology"
+	"github.com/scionproto/scion/private/trust"
 )
 
 func TestDefaultExtenderExtend(t *testing.T) {
@@ -98,8 +100,8 @@ func TestDefaultExtenderExtend(t *testing.T) {
 				intfs.Get(peer).Activate(peerRemoteIfs[peer])
 			}
 			ext := &beaconing.DefaultExtender{
-				IA:     topo.IA(),
-				Signer: testSigner(t, priv, topo.IA()),
+				IA:        topo.IA(),
+				SignerGen: testSignerGen{Signer: testSigner(t, priv, topo.IA())},
 				MAC: func() hash.Hash {
 					mac, err := scrypto.InitMac(make([]byte, 16))
 					require.NoError(t, err)
@@ -171,8 +173,8 @@ func TestDefaultExtenderExtend(t *testing.T) {
 		intfs := ifstate.NewInterfaces(interfaceInfos(topo), ifstate.Config{})
 		require.NoError(t, err)
 		ext := &beaconing.DefaultExtender{
-			IA:     topo.IA(),
-			Signer: testSigner(t, priv, topo.IA()),
+			IA:        topo.IA(),
+			SignerGen: testSignerGen{Signer: testSigner(t, priv, topo.IA())},
 			MAC: func() hash.Hash {
 				mac, err := scrypto.InitMac(make([]byte, 16))
 				require.NoError(t, err)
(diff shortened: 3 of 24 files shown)


matzf (changes requested):
> Reviewed 1 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: 1 of 13 files reviewed, 4 unresolved discussions (waiting on @oncilla and @uniquefine)

___
*[`control/tasks.go` line 59 at r1](https://reviewable.io/reviews/scionproto/scion/4369#-Na5nuIm7yozteI-n92L:-Na5nuIm7yozteI-n92M:b-hpqpjg) ([raw file](https://github.com/scionproto/scion/blob/89863593e52945f7504be45484a1cb88ce099f6b/control/tasks.go#L59)):*
> ```Go
> 	SegmentRegister       beaconing.RPC
> 	BeaconStore           Store
> 	Signer                seg.Signer
> ```

Remove this `Signer`? Looks like it's no longer used (`Signer` in `Originator` and `Propagator` is also unused).
___
*[`control/beaconing/extender.go` line 37 at r1](https://reviewable.io/reviews/scionproto/scion/4369#-Na5qGMu-glCX_IsHbxH:-Na5qGMu-glCX_IsHbxI:b-911eob) ([raw file](https://github.com/scionproto/scion/blob/89863593e52945f7504be45484a1cb88ce099f6b/control/beaconing/extender.go#L37)):*
> ```Go
> type SignerGen interface {
> 	// Generate generates a signer and returns its expiration time.
> 	Generate(ctx context.Context) (seg.Signer, time.Time, error)
> ```

Wouldn't it make sense to use `private/trust.Signer` in this file, instead of the `seg.Signer` interface? Then we can access the `signer.Expiration` directly and the existing `control/trust.SignerGen` interface suffices.
___
*[`control/trust/signer.go` line 38 at r1](https://reviewable.io/reviews/scionproto/scion/4369#-Na5sMtz-t-wY0k7JNRC:-Na5sMtz-t-wY0k7JNRD:bn0zzfz) ([raw file](https://github.com/scionproto/scion/blob/89863593e52945f7504be45484a1cb88ce099f6b/control/trust/signer.go#L38)):*
> ```Go
> 
> // Sign signs the message with the latest available Signer.
> func (s RenewingSigner) Sign(
> ```

This function no longer seems to be used. While the `SignCMS` one is still in use, I'm not sure this `RenewingSigner` abstraction is still worth keeping. I'd propose to remove this and pass the `SignerGen` around directly -- this would make the very few call sites for SignCMS just a few lines longer, and reduce the "magic".
We can also keep it for now and do look at this as a follow up.
___
*[`pkg/slayers/path/hopfield.go` line 152 at r1](https://reviewable.io/reviews/scionproto/scion/4369#-Na5g68c-6YYA0lg9bDx:-Na5g68d3Db08djn3I_5:bucsddk) ([raw file](https://github.com/scionproto/scion/blob/89863593e52945f7504be45484a1cb88ce099f6b/pkg/slayers/path/hopfield.go#L152)):*
> ```Go
> // can be used in a HopField. The returned value is guaranteed to be >= 1.
> // For durations that are out of range, an error is returned.
> func ExpTimeFromSeconds(seconds float64) (uint8, error) {
> ```

The commit "avoid using duration.Seconds()" changes this function from taking a `Duration` parameter to a `float64`. In my opinion,`Duration` should be preferred.
If you want to avoid the `duration.Seconds()` call (which has just been moved out of this function now),  keep the calculations in nanoseconds (i.e. multiply `MaxTTL` by `time.Second`, or even better, change `MaxTTL` to a duration constant).

The documentation is wrong: for values `expTimeUnit` to `2*expTimeUnit` it will (correctly) return 0.
The "fits into" phrasing is not very obvious, perhaps explicitly say "less than or equal". Maybe even write it out in (pseudo-)code:
```
//   d <= ExpTimeToDuration(ExpTimeFromDuration(d))
```

uniquefine (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: 0 of 24 files reviewed, 4 unresolved discussions (waiting on @lukedirtwalker, @matzf, and @oncilla)

___
*[`control/tasks.go` line 59 at r1](https://reviewable.io/reviews/scionproto/scion/4369#-Na5nuIm7yozteI-n92L:-Na7SIQO0G5ccGV2iIbk:bs2vhuc) ([raw file](https://github.com/scionproto/scion/blob/89863593e52945f7504be45484a1cb88ce099f6b/control/tasks.go#L59)):*

Remove this `Signer`? Looks like it's no longer used (`Signer` in `Originator` and `Propagator` is also unused).
</blockquote></details>

Good point, removed.
___
*[`control/beaconing/extender.go` line 37 at r1](https://reviewable.io/reviews/scionproto/scion/4369#-Na5qGMu-glCX_IsHbxH:-Na7STWw-IzjvC_v9Hat:bae4wr6) ([raw file](https://github.com/scionproto/scion/blob/89863593e52945f7504be45484a1cb88ce099f6b/control/beaconing/extender.go#L37)):*

Wouldn't it make sense to use `private/trust.Signer` in this file, instead of the `seg.Signer` interface? Then we can access the `signer.Expiration` directly and the existing `control/trust.SignerGen` interface suffices.
</blockquote></details>

True, I changed it to use trust.Signer directly. I didn't realize that we are using trust.Signer in all tests already.
___
*[`control/trust/signer.go` line 38 at r1](https://reviewable.io/reviews/scionproto/scion/4369#-Na5sMtz-t-wY0k7JNRC:-Na7Ta3TEQTI375lHrc2:b-o0tp7v) ([raw file](https://github.com/scionproto/scion/blob/89863593e52945f7504be45484a1cb88ce099f6b/control/trust/signer.go#L38)):*

This function no longer seems to be used. While the `SignCMS` one is still in use, I'm not sure this `RenewingSigner` abstraction is still worth keeping. I'd propose to remove this and pass the `SignerGen` around directly -- this would make the very few call sites for SignCMS just a few lines longer, and reduce the "magic".
We can also keep it for now and do look at this as a follow up.
</blockquote></details>

I agree that this doesn't add much. But as you mentioned, it's not really related to this PR. I'll do it in a followup PR instead.
___
*[`pkg/slayers/path/hopfield.go` line 152 at r1](https://reviewable.io/reviews/scionproto/scion/4369#-Na5g68c-6YYA0lg9bDx:-NaAnCim1z9hfZtT4Pz4:bubypjv) ([raw file](https://github.com/scionproto/scion/blob/89863593e52945f7504be45484a1cb88ce099f6b/pkg/slayers/path/hopfield.go#L152)):*

The commit "avoid using duration.Seconds()" changes this function from taking a `Duration` parameter to a `float64`. In my opinion,`Duration` should be preferred.
If you want to avoid the `duration.Seconds()` call (which has just been moved out of this function now),  keep the calculations in nanoseconds (i.e. multiply `MaxTTL` by `time.Second`, or even better, change `MaxTTL` to a duration constant).

The documentation is wrong: for values `expTimeUnit` to `2*expTimeUnit` it will (correctly) return 0.
The "fits into" phrasing is not very obvious, perhaps explicitly say "less than or equal". Maybe even write it out in (pseudo-)code:
```
//   d <= ExpTimeToDuration(ExpTimeFromDuration(d))
```
</blockquote></details>

I totally agree with what you said about the doc comment.

I now changed the MaxTTL to duration type. This makes some calculations more precise (hence the updated tests).
My issue with the duration is that the gobra check in CI fails. What can/should I do to resolve the failing check?

uniquefine (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: 0 of 24 files reviewed, 4 unresolved discussions (waiting on @lukedirtwalker, @matzf, and @oncilla)

___
*[`pkg/slayers/path/hopfield.go` line 152 at r1](https://reviewable.io/reviews/scionproto/scion/4369#-Na5g68c-6YYA0lg9bDx:-NaB9L3QDctzDohwrxRq:bc0ew0h) ([raw file](https://github.com/scionproto/scion/blob/89863593e52945f7504be45484a1cb88ce099f6b/pkg/slayers/path/hopfield.go#L152)):*
<details><summary><i>Previously, uniquefine (Sebastian Leisinger) wrote…</i></summary><blockquote>

I totally agree with what you said about the doc comment.

I now changed the MaxTTL to duration type. This makes some calculations more precise (hence the updated tests).
My issue with the duration is that the gobra check in CI fails. What can/should I do to resolve the failing check?
</blockquote></details>

Oh NVM, now it passed the check.

matzf (changes requested):
The PR description seems a bit inaccurate; this is changing the signer side only now, so the title should be something like "cppki: limit segment expiration by certificate lifetime"

This _contributes_ to the linked issue, but the issue is not completed until the relying party is also updated to require this, in a second step.

> Reviewed 2 of 13 files at r1, 7 of 23 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369#-:-NaBoQJq6Hl0wTJpk9B3:bfussy3)* status: 9 of 25 files reviewed, 4 unresolved discussions (waiting on @lukedirtwalker, @oncilla, and @uniquefine)

___
*[`control/beaconing/extender.go` line 192 at r2](https://reviewable.io/reviews/scionproto/scion/4369#-NaBFR1kAfUPJfqt85Il:-NaBFR1kAfUPJfqt85Im:b47n2t2) ([raw file](https://github.com/scionproto/scion/blob/d483a0952c1721b2e5b0809d81b34b82f9c5c3be/control/beaconing/extender.go#L192)):*
> ```Go
> 	}
> 	hopF, epicMac := s.createHopF(ingress, egress, expTime, ts, beta)
> 	// make sure
> ```

`// make sure` what? :)
___
*[`control/trust/signer.go` line 61 at r2](https://reviewable.io/reviews/scionproto/scion/4369#-NaBH1y20MwXNw0ImDyP:-NaBH1y20MwXNw0ImDyQ:b-qxuang) ([raw file](https://github.com/scionproto/scion/blob/d483a0952c1721b2e5b0809d81b34b82f9c5c3be/control/trust/signer.go#L61)):*
> ```Go
> func (s RenewingSigner) Generate(ctx context.Context) (trust.Signer, error) {
> 	return s.SignerGen.Generate(ctx)
> }
> ```

This method seems redundant. Just use the `SignerGen` directly (pass `SignerGen: signer.SignerGen` instead of `SignerGen: signer` in `control/cmd/control/main.go`).
___
*[`pkg/slayers/path/hopfield.go` line 34 at r2](https://reviewable.io/reviews/scionproto/scion/4369#-NaBRH4O64Du5OVwBUu1:-NaBRH4O64Du5OVwBUu2:b-n7cnp) ([raw file](https://github.com/scionproto/scion/blob/d483a0952c1721b2e5b0809d81b34b82f9c5c3be/pkg/slayers/path/hopfield.go#L34)):*
> ```Go
> 
> // MaxTTL is the maximum age of a HopField in seconds.
> const MaxTTL = 24 * 60 * 60 * time.Second // One day in seconds
> ```

Nit: comments are no longer accurate. 
Can use `time.Hour` to make the definition obvious enough that a comment is not required



_Suggestion:_
```Go
// MaxTTL is the maximum age of a HopField.
const MaxTTL = 24 * time.Hour
```
___
*[`pkg/slayers/path/hopfield.go` line 36 at r2](https://reviewable.io/reviews/scionproto/scion/4369#-NaBQypZ4kf4Kkp1vS4T:-NaBQypZ4kf4Kkp1vS4U:b-uuac9j) ([raw file](https://github.com/scionproto/scion/blob/d483a0952c1721b2e5b0809d81b34b82f9c5c3be/pkg/slayers/path/hopfield.go#L36)):*
> ```Go
> const MaxTTL = 24 * 60 * 60 * time.Second // One day in seconds
> 
> const expTimeUnit = MaxTTL / 256 // ~5m38s
> ```

Nit: // 5m38.5s

uniquefine (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: 6 of 25 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker, @matzf, and @oncilla)

___
*[`control/beaconing/extender.go` line 192 at r2](https://reviewable.io/reviews/scionproto/scion/4369#-NaBFR1kAfUPJfqt85Il:-NaCGPGq6XPCvxiWP_QT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d483a0952c1721b2e5b0809d81b34b82f9c5c3be/control/beaconing/extender.go#L192)):*

`// make sure` what? :)
</blockquote></details>

Done.
___
*[`control/trust/signer.go` line 61 at r2](https://reviewable.io/reviews/scionproto/scion/4369#-NaBH1y20MwXNw0ImDyP:-NaCGnZEAOScwrIVKEKI:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d483a0952c1721b2e5b0809d81b34b82f9c5c3be/control/trust/signer.go#L61)):*

This method seems redundant. Just use the `SignerGen` directly (pass `SignerGen: signer.SignerGen` instead of `SignerGen: signer` in `control/cmd/control/main.go`).
</blockquote></details>

Done.

matzf (changes requested):
> Reviewed 15 of 23 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: all files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker, @oncilla, and @uniquefine)

___
*[`private/path/combinator/__debug_bin` line 0 at r3](https://reviewable.io/reviews/scionproto/scion/4369#-NaC_Ery1MHXLXp_kLVg:-NaC_Erz5ni4BPjEBYs0:b-ja1qh) ([raw file](https://github.com/scionproto/scion/blob/7fe407b0f616d84c658960ad284c6ab627df4ce0/private/path/combinator/__debug_bin#L0)):*
Accidentally committed?

uniquefine (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: 24 of 25 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker, @matzf, and @oncilla)

___
*[`private/path/combinator/__debug_bin` line  at r3](https://reviewable.io/reviews/scionproto/scion/4369#-NaC_Ery1MHXLXp_kLVg:-NaCbYimAkYOGZBqF_A_:b-6r5n8n) ([raw file](https://github.com/scionproto/scion/blob/7fe407b0f616d84c658960ad284c6ab627df4ce0/private/path/combinator/__debug_bin#L)):*

Accidentally committed?
</blockquote></details>

Yes, reverted

matzf (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @lukedirtwalker and @oncilla)

matzf:
@lukedirtwalker, @oncilla, this PR is waiting for your review.

oncilla (changes requested):
> Reviewed 2 of 13 files at r1, 18 of 23 files at r2, 4 of 4 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: all files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @uniquefine)

___
*[`control/beaconing/extender.go` line 101 at r4](https://reviewable.io/reviews/scionproto/scion/4369#-Nazmqt13zrn1CvmF9K1:-NazmVqiETaLWXYtUpqy:b-ihgj8t) ([raw file](https://github.com/scionproto/scion/blob/d708fc927e3fe295dadb631440a70ee6fec35327/control/beaconing/extender.go#L101)):*
> ```Go
> 	// Make sure the hop expiration time is not longer than the signer expiration time.
> 	expTime := s.MaxExpTime()
> 	if ts.Add(path.ExpTimeToDuration(expTime)).After(signer.Expiration) {
> ```

This currently silently reduces the expiration time until we go to zero.

Let's add a gauge that indicates that the configured maximum expiration time can no longer be supported.
Operators should be able to create an alert based on that guage.
___
*[`pkg/slayers/path/hopfield.go` line 160 at r4](https://reviewable.io/reviews/scionproto/scion/4369#-Nazp5BS6PJaIe2oFEC4:-Nazp5BS6PJaIe2oFEC5:b-9mxgtj) ([raw file](https://github.com/scionproto/scion/blob/d708fc927e3fe295dadb631440a70ee6fec35327/pkg/slayers/path/hopfield.go#L160)):*
> ```Go
> 		return 0, serrors.New("duration too large", "seconds", d)
> 	}
> 	return uint8(d*256/MaxTTL - 1), nil
> ```

add brackets to make it unambiguous even if the reader does not know the precedence order.



_Code quote:_
```Go
d*256/MaxTTL
```

lukedirtwalker:
given both Matthias and Dominik already review(ed) it, I'll pass on doing a review.

oncilla (changes requested):
> Reviewed 7 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: all files reviewed, 4 unresolved discussions (waiting on @uniquefine)

___
*[`control/observability.go` line 249 at r5](https://reviewable.io/reviews/scionproto/scion/4369#-NfACM2k1Xe85f_QWMhA:-NfACM2k1Xe85f_QWMhB:undefined) ([raw file](https://github.com/scionproto/scion/blob/5898c1d975f12ff2077a7086d53dc7f1f38119b8/control/observability.go#L249)):*
> ```Go
> 			prometheus.GaugeOpts{
> 				Name: "control_segment_expiration_deficient",
> 				Help: "Indicates whether the expiration time of the segment is below the maximum." +
> ```





_Suggestion:_
```Go
the configured maximum." +
```
___
*[`control/observability.go` line 252 at r5](https://reviewable.io/reviews/scionproto/scion/4369#-NfAByt6BJAYsFfwtwpF:-NfAByt6BJAYsFfwtwpG:ba4brop) ([raw file](https://github.com/scionproto/scion/blob/5898c1d975f12ff2077a7086d53dc7f1f38119b8/control/observability.go#L252)):*
> ```Go
> 					" This happens when the signer expiration time is lower than the maximum segment expiration time.",
> 			},
> 			[]string{"src"},
> ```

This type of label is usually named `"isd_as"`. Let's use that for consistency.
___
*[`control/tasks.go` line 219 at r5](https://reviewable.io/reviews/scionproto/scion/4369#-NfABFQA4F-0G52fcG2Z:-NfABFQB-CXw0ozybYUK:bi4g4oh) ([raw file](https://github.com/scionproto/scion/blob/5898c1d975f12ff2077a7086d53dc7f1f38119b8/control/tasks.go#L219)):*
> ```Go
> 		EPIC:       t.EPIC,
> 	}
> 	if t.Metrics != nil {
> ```

Personally, I prefer the following:

```go
...
SegementExpirationDeficient: func() metrics.Gauge {
    if t.Metrics == nil {
        return nil
    }
    return metrics.GaugeWith(
			metrics.NewPromGauge(t.Metrics.SegmentExpirationDeficient),
			"isd_as", ia.String(),
		)

}(),
```

This way, initialization is not split across multiple lines just because of a conditional.

(non-blocking) because it is a matter of taste.
___
*[`control/beaconing/extender.go` line 101 at r4](https://reviewable.io/reviews/scionproto/scion/4369#-Nazmqt13zrn1CvmF9K1:-NfAAaNW5YXDCMzp_dDx:bijuczu) ([raw file](https://github.com/scionproto/scion/blob/d708fc927e3fe295dadb631440a70ee6fec35327/control/beaconing/extender.go#L101)):*
<details><summary><i>Previously, uniquefine (Sebastian Leisinger) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

You also need a to unset it once it is not deficient anymore.
___
*[`pkg/slayers/path/hopfield.go` line 160 at r4](https://reviewable.io/reviews/scionproto/scion/4369#-Nazp5BS6PJaIe2oFEC4:-NfA7S4d-ORhbQIe6Or8:btq3ep0) ([raw file](https://github.com/scionproto/scion/blob/d708fc927e3fe295dadb631440a70ee6fec35327/pkg/slayers/path/hopfield.go#L160)):*
<details><summary><i>Previously, uniquefine (Sebastian Leisinger) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

The crucial part is the * and /. Please add the brackets there.
I think go formating already does a good job at indicating precedence between */ and - by adding the whitespace

```
uint8((d*256)/MaxTTL - 1)
```

oncilla (changes requested):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: all files reviewed, 2 unresolved discussions (waiting on @uniquefine)

___
*[`control/observability.go` line 252 at r5](https://reviewable.io/reviews/scionproto/scion/4369#-NfAByt6BJAYsFfwtwpF:-NfHZ6Wa0BA12EuonAXF:b-96af7g) ([raw file](https://github.com/scionproto/scion/blob/5898c1d975f12ff2077a7086d53dc7f1f38119b8/control/observability.go#L252)):*
<details><summary><i>Previously, uniquefine (Sebastian Leisinger) wrote…</i></summary><blockquote>

Done, This service doesn't use "isd_as" as label on any of the other metrics though.
</blockquote></details>

It does not, but it does use src with very different semantics :wink:
But come to think of it, it does not add this label in other places because it is single ISD-AS anyway. I think we should drop it for consistency.
___
*[`pkg/slayers/path/hopfield.go` line 160 at r4](https://reviewable.io/reviews/scionproto/scion/4369#-Nazp5BS6PJaIe2oFEC4:-NfHZynb4xYQHQzCPp-Z:b-gvul65) ([raw file](https://github.com/scionproto/scion/blob/d708fc927e3fe295dadb631440a70ee6fec35327/pkg/slayers/path/hopfield.go#L160)):*
<details><summary><i>Previously, uniquefine (Sebastian Leisinger) wrote…</i></summary><blockquote>

:bulb: I didn't think that it makes a difference
</blockquote></details>

For the record, the formula was correct in the beginning already

But time.Duration is an integer behind the scenes. You need to be careful when dealing with divisions and time.Duration.
That's why it is extra helpful to the reader if they do not need to worry about precedence rules.

uniquefine (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: 22 of 27 files reviewed, 2 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`control/observability.go` line 252 at r5](https://reviewable.io/reviews/scionproto/scion/4369#-NfAByt6BJAYsFfwtwpF:-NgPChV4DxOvpytAPuyl:b2cm3ml) ([raw file](https://github.com/scionproto/scion/blob/5898c1d975f12ff2077a7086d53dc7f1f38119b8/control/observability.go#L252)):*

It does not, but it does use src with very different semantics :wink:
But come to think of it, it does not add this label in other places because it is single ISD-AS anyway. I think we should drop it for consistency.
</blockquote></details>

Removed, Done.
___
*[`pkg/slayers/path/hopfield.go` line 160 at r4](https://reviewable.io/reviews/scionproto/scion/4369#-Nazp5BS6PJaIe2oFEC4:-NgPCkQnD771X8YJugbX:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/d708fc927e3fe295dadb631440a70ee6fec35327/pkg/slayers/path/hopfield.go#L160)):*

For the record, the formula was correct in the beginning already

But time.Duration is an integer behind the scenes. You need to be careful when dealing with divisions and time.Duration.
That's why it is extra helpful to the reader if they do not need to worry about precedence rules.
</blockquote></details>

Done.

matzf (approved):
> Reviewed 2 of 7 files at r5, 2 of 5 files at r6, 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: all files reviewed, 2 unresolved discussions (waiting on @oncilla)

oncilla (approved):
> Reviewed 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4369)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @uniquefine)

PR #4371, opened 2023-07-24
build: support fedora

jiceatscion:
Changes can be summarized as follows:
1 . Update the scripts in tools/ so the denecessary packages are installed (in the right order).
2 . Bottleneck references the sqlite lib through a single source file so it's easy to use different implementations of that lib.
3 . Replace mattn_sqlite with modenrc_sqlite to remove dependency on a specific version of glibc (which breaks integration tests).

Also did some minor makefile simplifications: no need to "make go-mod-tidy" and "make go_deps.blz" explicitly. It happens as needed upon "make gazelle".

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)


matzf (changes requested):
> Reviewed 40 of 40 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: all files reviewed, 7 unresolved discussions (waiting on @jiceatscion)

___
*[`BUILD.bazel` line 22 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaN3VV_3w7EgKndFAv8:-NaN3VV_3w7EgKndFAv9:b-vxnar7) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/BUILD.bazel#L22)):*
> ```Starlark
> 
> # This is simplistic but the complete, by-the-everchanging-bazel-book, solution
> # is ludicrously complicated. Go there if and when needed.
> ```

Could you describe first in the comment what this tries to achieve and a hint on how to use it? 

IIUC, this `config_setting`  _matches_ the gotags defined elsewhere, in the bazelrc or on the command line. Initially I had expected it to _define_ values, like the attribute is called, i.e. to set the gotags. I was very confused for a moment...
___
*[`Makefile` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMwu4tA2PD6nQZseu5:-NaMwu4tA2PD6nQZseu6:bvn1yop) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/Makefile#L1)):*
> ```Makefile
> .PHONY: all antlr bazel clean docker-images gazelle go.mod licenses mocks protobuf scion-topo test test-integration write_all_source_files
> ```

`go.mod` is not phony :)
___
*[`Makefile` line 63 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMx7qvBlzRvPpPmHMG:-NaMx7qw3e_smLYmgvCF:b73g4r5) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/Makefile#L63)):*
> ```Makefile
> gazelle: go_deps.bzl
> 	@# call gazelle with -args, which appends our arguments to those from the gazelle() rule
> 	bazel run //:gazelle --verbose_failures --config=quiet -- -args -mode=$(GAZELLE_MODE) $(GAZELLE_DIRS)
> ```

Maybe it would be worth getting rid of these args entirely? The GAZELLE_MODE is only used in `lint-go-gazelle` below. There, it could be replaced by creating a corresponding `gazelle_diff` rule in the `BUILD` file (analogous to the gazelle_update_repos rule above).  The `antlr` thing ignores the GAZELLE_MODE, so we can just remove it there.
___
*[`Makefile` line 66 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMzjymA-Be6KRIHr4S:-NaMzjymA-Be6KRIHr4T:b-t5fp45) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/Makefile#L66)):*
> ```Makefile
> 
> licenses: bazel
> 	tools/licenses.sh
> ```

Should we also remove the `bazel build` from the tools/licenses.sh?
___
*[`nogo.json` line 113 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaN-vhNAfApr92dinQ7:-NaN-vhNAfApr92dinQ8:b-bqq1mx) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/nogo.json#L113)):*
> ```JSON
>             "/com_github_uber_jaeger_client_go": "",
> 	    "/org_modernc_sqlite": "",
> 	    "/org_modernc_mathutil": ""
> ```

Nit: inconsistent indent (tab)
___
*[`private/storage/db/sqlite_modernc.go` line 33 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMvGg31ERzJ1KFcKIf:-NaMvGg31ERzJ1KFcKIg:bqyune9) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/private/storage/db/sqlite_modernc.go#L33)):*
> ```Go
> // no database exists a new database is be created. If the schema version of the
> // stored database is different from schemaVersion, an error is returned.
> func NewSqlite(path string, schema string, schemaVersion int) (*sql.DB, error) {
> ```

Instead of replicating this entire file, we could move only the _imports_ for the  sqlite implementations into two separate files. So these two new files would be empty, except for the anonymous import and the build tags.

Perhaps we could add the same constant definition to both files to ensure the build fails if the gotags are not set correctly:
```
const buildtag_guard = "must choose an sqlite implementation to build, by defining exactly one of the gotags 'sqlite_modernc' or 'sqlite_mattn'"
```
___
*[`private/storage/db/sqlite_mattn.go` line 131 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMu9J93ZTVY7-GT7bI:-NaMu9J93ZTVY7-GT7bJ:b87hx3y) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/private/storage/db/sqlite_mattn.go#L131)):*
> ```Go
> // that we do not care about. Currently we do not care about ErrConstraint.
> // If the error is filtered, we return nil.
> func FilterError(err error) error {
> ```

FilterError and IsSqliteError is no longer needed after #4372.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: 36 of 40 files reviewed, 5 unresolved discussions (waiting on @matzf)

___
*[`Makefile` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMwu4tA2PD6nQZseu5:-NaQpCkD6AkVQ4Zzx9xw:b-wjpwud) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/Makefile#L1)):*

`go.mod` is not phony :)
</blockquote></details>

That a make quirck that I had long forgotten. Yes go.mod is phony: it needs to be built regardless of the existence of a file by that name. That is in fact what phony does. Without phony, it doesn't get updated because it has no explicit dependency.
___
*[`Makefile` line 63 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMx7qvBlzRvPpPmHMG:-NaQv6Zi6R1OoCwATDt-:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/Makefile#L63)):*

Maybe it would be worth getting rid of these args entirely? The GAZELLE_MODE is only used in `lint-go-gazelle` below. There, it could be replaced by creating a corresponding `gazelle_diff` rule in the `BUILD` file (analogous to the gazelle_update_repos rule above).  The `antlr` thing ignores the GAZELLE_MODE, so we can just remove it there.
</blockquote></details>

done
___
*[`Makefile` line 66 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMzjymA-Be6KRIHr4S:-NaQwKpL5mO2aMbmPJEL:b-qfyurw) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/Makefile#L66)):*

Should we also remove the `bazel build` from the tools/licenses.sh?
</blockquote></details>

Yes, done.
___
*[`nogo.json` line 113 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaN-vhNAfApr92dinQ7:-NaQwgbOETCKRgPVacqR:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/nogo.json#L113)):*

Nit: inconsistent indent (tab)
</blockquote></details>

done
___
*[`private/storage/db/sqlite_modernc.go` line 33 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMvGg31ERzJ1KFcKIf:-NaR0RAMA0PSqy42LFx5:b8mzvim) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/private/storage/db/sqlite_modernc.go#L33)):*

Instead of replicating this entire file, we could move only the _imports_ for the  sqlite implementations into two separate files. So these two new files would be empty, except for the anonymous import and the build tags.

Perhaps we could add the same constant definition to both files to ensure the build fails if the gotags are not set correctly:
```
const buildtag_guard = "must choose an sqlite implementation to build, by defining exactly one of the gotags 'sqlite_modernc' or 'sqlite_mattn'"
```
</blockquote></details>

The two aren't really identical, the differences are slight but hard to represent as meaningful abstractions.  Still, I'll see what I can do about the duplication without making outright spaghetti. Stay tuned.

Re. the constant. Good idea, the error will be clearer than "oh you have two functions named NewSqlite". So, done
___
*[`private/storage/db/sqlite_mattn.go` line 131 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMu9J93ZTVY7-GT7bI:-NaR-pJN59vnJRk3NQVw:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/private/storage/db/sqlite_mattn.go#L131)):*

FilterError and IsSqliteError is no longer needed after #4372.
</blockquote></details>

Done.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: 36 of 40 files reviewed, 4 unresolved discussions (waiting on @matzf)

___
*[`BUILD.bazel` line 22 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaN3VV_3w7EgKndFAv8:-NaRHN712gMjP-xypKVa:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/BUILD.bazel#L22)):*

Could you describe first in the comment what this tries to achieve and a hint on how to use it? 

IIUC, this `config_setting`  _matches_ the gotags defined elsewhere, in the bazelrc or on the command line. Initially I had expected it to _define_ values, like the attribute is called, i.e. to set the gotags. I was very confused for a moment...
</blockquote></details>

Done.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: 30 of 41 files reviewed, 4 unresolved discussions (waiting on @matzf)

___
*[`private/storage/db/sqlite_modernc.go` line 33 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMvGg31ERzJ1KFcKIf:-NaRrdXIDisElpK-HwXi:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/private/storage/db/sqlite_modernc.go#L33)):*

The two aren't really identical, the differences are slight but hard to represent as meaningful abstractions.  Still, I'll see what I can do about the duplication without making outright spaghetti. Stay tuned.

Re. the constant. Good idea, the error will be clearer than "oh you have two functions named NewSqlite". So, done

</blockquote></details>

done

matzf (changes requested):
> Reviewed 3 of 4 files at r2, 7 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: 40 of 41 files reviewed, 3 unresolved discussions (waiting on @jiceatscion)

___
*[`BUILD.bazel` line 22 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaN3VV_3w7EgKndFAv8:-Naf12u_62ot2FRjRETR:b6zo1sm) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/BUILD.bazel#L22)):*

Done.
</blockquote></details>

:100: thanks!
___
*[`Makefile` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMwu4tA2PD6nQZseu5:-Naf19giCjcn8zbG58la:b2xc7mm) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/Makefile#L1)):*

That a make quirck that I had long forgotten. Yes go.mod is phony: it needs to be built regardless of the existence of a file by that name. That is in fact what phony does. Without phony, it doesn't get updated because it has no explicit dependency.
</blockquote></details>

Argh, of course, sorry.
___
*[`private/storage/db/sqlite.go` line 22 at r3](https://reviewable.io/reviews/scionproto/scion/4371#-Naf2bew2UDr6xvRQkIY:-Naf2bew2UDr6xvRQkIZ:b-72hx92) ([raw file](https://github.com/scionproto/scion/blob/1240f440370e7fc579afc885c13f7ab2f6f8a1f5/private/storage/db/sqlite.go#L22)):*
> ```Go
> 	"net/url"
> 
> 	_ "modernc.org/sqlite"
> ```

This is not supposed to be here, or is it?
___
*[`private/storage/db/sqlite_modernc.go` line 33 at r1](https://reviewable.io/reviews/scionproto/scion/4371#-NaMvGg31ERzJ1KFcKIf:-Naf2hJ_-Bk_Yw8K3kj1:b7c58zb) ([raw file](https://github.com/scionproto/scion/blob/3abef90be2ae09d1a72baff867a7ce639e9c1856/private/storage/db/sqlite_modernc.go#L33)):*

done
</blockquote></details>

Ah sorry, I missed the differences. Looks good :+1: 
___
*[`private/storage/db/sqlite_mattn.go` line 1 at r3](https://reviewable.io/reviews/scionproto/scion/4371#-Naf1wcb6SWxDi-9gRlo:-Naf1wcb6SWxDi-9gRlp:undefined) ([raw file](https://github.com/scionproto/scion/blob/1240f440370e7fc579afc885c13f7ab2f6f8a1f5/private/storage/db/sqlite_mattn.go#L1)):*
> ```Go
> // Copyright 2023 Scion Association
> ```





_Suggestion:_
```Go
// Copyright 2023 SCION Association
```
___
*[`private/storage/db/sqlite_mattn.go` line 31 at r3](https://reviewable.io/reviews/scionproto/scion/4371#-Naf2-cE-Gm6tYMrcfYu:-Naf2-cE-Gm6tYMrcfYv:b-m0yyjn) ([raw file](https://github.com/scionproto/scion/blob/1240f440370e7fc579afc885c13f7ab2f6f8a1f5/private/storage/db/sqlite_mattn.go#L31)):*
> ```Go
> // connection path for this sqlite implementation. The modifications turn on 
> // foreign keys and WAL journal mode for every SQL query.
> func AddPragmas(q url.Values) {
> ```

Should very likely not be exported, `addPragmas`, and `driverName`.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: 35 of 43 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`private/storage/db/sqlite.go` line 22 at r3](https://reviewable.io/reviews/scionproto/scion/4371#-Naf2bew2UDr6xvRQkIY:-Naf9-eL1DCAiG8NL2d_:b-wb2jd1) ([raw file](https://github.com/scionproto/scion/blob/1240f440370e7fc579afc885c13f7ab2f6f8a1f5/private/storage/db/sqlite.go#L22)):*

This is not supposed to be here, or is it?
</blockquote></details>

Indeed. That was a leftover. Gone.
___
*[`private/storage/db/sqlite_mattn.go` line 1 at r3](https://reviewable.io/reviews/scionproto/scion/4371#-Naf1wcb6SWxDi-9gRlo:-Naf98I37iSjgcly_W_z:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1240f440370e7fc579afc885c13f7ab2f6f8a1f5/private/storage/db/sqlite_mattn.go#L1)):*
> ```Go
> // Copyright 2023 Scion Association
> ```

Done.
___
*[`private/storage/db/sqlite_mattn.go` line 31 at r3](https://reviewable.io/reviews/scionproto/scion/4371#-Naf2-cE-Gm6tYMrcfYu:-Naf9AF1-XPaeWBF7ThZ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/1240f440370e7fc579afc885c13f7ab2f6f8a1f5/private/storage/db/sqlite_mattn.go#L31)):*

Should very likely not be exported, `addPragmas`, and `driverName`.
</blockquote></details>

Done.

matzf (commented):
> Reviewed 3 of 7 files at r4, 2 of 4 files at r5, 1 of 1 files at r6, 4 of 4 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: 45 of 46 files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4371#-NavXdOa5Kpzd-DwT4hg:-NavXdObDSYyOg7Z5eJY:bftgt27) (no related file):*
The section on building with `go build` in the documentation should now explain that a tag must be given to select the sqlite implementation, here: 
https://github.com/scionproto/scion/blob/2ce7932c0b0ae94679e3fdc08425332d49dbc904/doc/dev/setup.rst?plain=1#L116

___

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: 42 of 47 files reviewed, 1 unresolved discussion (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4371#-NavXdOa5Kpzd-DwT4hg:-NavqE8W0j_pEq0AreV7:b-896fix) (no related file):*

The section on building with `go build` in the documentation should now explain that a tag must be given to select the sqlite implementation, here: 
https://github.com/scionproto/scion/blob/2ce7932c0b0ae94679e3fdc08425332d49dbc904/doc/dev/setup.rst?plain=1#L116
</blockquote></details>

Done.

___

matzf (commented):
> Reviewed 1 of 2 files at r7, 4 of 4 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: all files reviewed, 2 unresolved discussions (waiting on @jiceatscion)

___
*[`doc/dev/setup.rst` line 115 at r10](https://reviewable.io/reviews/scionproto/scion/4371#-NavtiBT9s0uj7Y-zjg_:-NavtiBUEJoZKqF_9j76:b-k6fiqe) ([raw file](https://github.com/scionproto/scion/blob/1573c026f29b9571c4d2732c54bfab714b4f66da/doc/dev/setup.rst#L115)):*
> ```reStructuredText
> 
>    - `mattn`: A cgo implementation. It is well established but can makes go
>      executables dependent on a minimum glibc version.
> ```

Also, maybe "but requires `CGo <... link to cgo>`_ and makes go executables depend ..."



_Suggestion:_
```reStructuredText
   - `mattn`: A cgo implementation. It is well established but makes go
     executables dependent on a minimum glibc version.
```

matzf (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: all files reviewed, 2 unresolved discussions (waiting on @jiceatscion)

oncilla (changes requested):
> Reviewed 30 of 41 files at r1, 3 of 4 files at r2, 2 of 8 files at r3, 3 of 7 files at r4, 1 of 2 files at r7, 4 of 4 files at r9, 3 of 4 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: all files reviewed, 3 unresolved discussions (waiting on @jiceatscion and @matzf)

___
*[`private/storage/db/sqlite_modernc.go` line 26 at r11](https://reviewable.io/reviews/scionproto/scion/4371#-NazwHTR53cC0UOAhxOQ:-NazwHTR53cC0UOAhxOR:b-7ropln) ([raw file](https://github.com/scionproto/scion/blob/cf084ae1c79f876692de214682410d7bdcf02f90/private/storage/db/sqlite_modernc.go#L26)):*
> ```Go
> )
> 
> const buildtag_guard_either_sqlite_mattn_or_sqlite_modernc = "must choose an sqlite " +
> ```

ditto
___
*[`private/storage/db/sqlite_mattn.go` line 26 at r11](https://reviewable.io/reviews/scionproto/scion/4371#-NazvopPEA_17B-SRELW:-NazvopPEA_17B-SRELX:b-prc07m) ([raw file](https://github.com/scionproto/scion/blob/cf084ae1c79f876692de214682410d7bdcf02f90/private/storage/db/sqlite_mattn.go#L26)):*
> ```Go
> )
> 
> const buildtag_guard_either_sqlite_mattn_or_sqlite_modernc = "must choose an sqlite " +
> ```

variable names in go should always be camelCase instead of snake_case.

See https://go.dev/doc/effective_go#mixed-caps

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: all files reviewed, 3 unresolved discussions (waiting on @jiceatscion and @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4371#-NavXdOa5Kpzd-DwT4hg:-NazwurG5R5E2o-3zHVX:b601o4) (no related file):*

Done.
</blockquote></details>

I think having to specify a build flag in all cases is a bit unfortunate.
It also goes against the Go credo that every repository should compile and be able to run `go test ./...` out of the box.

I would prefer if we agree on a default (modernc or mattn) that is used when no build tag is specified, and enable the other via a build tag instead.

___

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4371#-NavXdOa5Kpzd-DwT4hg:-Nb-B4Ja2LwP6ckxvCc0:bvmmjc0) (no related file):*

I think having to specify a build flag in all cases is a bit unfortunate.
It also goes against the Go credo that every repository should compile and be able to run `go test ./...` out of the box.

I would prefer if we agree on a default (modernc or mattn) that is used when no build tag is specified, and enable the other via a build tag instead.
</blockquote></details>

I don't think go test ./ works out of the box already. There are quite a few preparatory steps required. That said, I am not entirely against having a default choice. (If we want the integration tests to work out of the box for the majority of development environment that would have to be modernc.). However I would find it a bit unfortunate that there be no way to make either choice explicit and to assume there will ever be only two choices.

Because go build constraints are so simplistic I do not know of a way to match the absence of a given category of tags. Should there be more than 2 options for something, we would need to express the default as a negation of every other variants. For example, we would need to put the following in sqlite_modernc.go:

//go:build sqlite_modernc || !(sqlite_mattn || sqlite_x || sqlite_y)

We're really missing a key-value oriented syntax; something like
//go:build sqlite=modernc || sqlite=''

... Unles there's regex matching? Does something like this work?:
//go:build sqlite_modernc || !sqlite_.* 

I haven't actually tried.

Opinions?

___
___
*[`private/storage/db/sqlite_mattn.go` line 26 at r11](https://reviewable.io/reviews/scionproto/scion/4371#-NazvopPEA_17B-SRELW:-Nb-9wvv-gVHv1r7ST20:b-3ne0n2) ([raw file](https://github.com/scionproto/scion/blob/cf084ae1c79f876692de214682410d7bdcf02f90/private/storage/db/sqlite_mattn.go#L26)):*

variable names in go should always be camelCase instead of snake_case.

See https://go.dev/doc/effective_go#mixed-caps


</blockquote></details>

This is meant to be easy to read as part of an error message, thereby providing an obvious hint to the user. We'd loose some of that readability by making it CamelCase. I don't see the point of applying the go convention in this case. This variable isn't referenced except by assigning it to "_" to ensure we get an error when it is undefined, so it has no effect on the readability of the rest of the code.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`private/storage/db/sqlite_modernc.go` line 26 at r11](https://reviewable.io/reviews/scionproto/scion/4371#-NazwHTR53cC0UOAhxOQ:-Nb-JInm44adidF9tZ4w:b-rqxvp7) ([raw file](https://github.com/scionproto/scion/blob/cf084ae1c79f876692de214682410d7bdcf02f90/private/storage/db/sqlite_modernc.go#L26)):*

ditto
</blockquote></details>

ditto

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4371#-NavXdOa5Kpzd-DwT4hg:-Nb-UAKQ4T7Fu-AzEcyQ:b-s64wrc) (no related file):*

I don't think go test ./ works out of the box already. There are quite a few preparatory steps required. That said, I am not entirely against having a default choice. (If we want the integration tests to work out of the box for the majority of development environment that would have to be modernc.). However I would find it a bit unfortunate that there be no way to make either choice explicit and to assume there will ever be only two choices.

Because go build constraints are so simplistic I do not know of a way to match the absence of a given category of tags. Should there be more than 2 options for something, we would need to express the default as a negation of every other variants. For example, we would need to put the following in sqlite_modernc.go:

//go:build sqlite_modernc || !(sqlite_mattn || sqlite_x || sqlite_y)

We're really missing a key-value oriented syntax; something like
//go:build sqlite=modernc || sqlite=''

... Unles there's regex matching? Does something like this work?:
//go:build sqlite_modernc || !sqlite_.* 

I haven't actually tried.

Opinions?

</blockquote></details>

After a quick check of the spec... there's no such thing as wildcard matching in go tags and no practical way of implementing anything approximating key-val mapping. So that leaves us with enumerating all alternatives in the default file :-(

___

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4371#-NavXdOa5Kpzd-DwT4hg:-Nb-jlhP9MYd1QjEwFSa:bxg5yfq) (no related file):*

After a quick check of the spec... there's no such thing as wildcard matching in go tags and no practical way of implementing anything approximating key-val mapping. So that leaves us with enumerating all alternatives in the default file :-(
</blockquote></details>

So, within these limitations, here's the change that would give you a default sqlite impl:

```
diff --git a/private/storage/db/sqlite\_modernc.go b/private/storage/db/sqlite\_modernc.go  
index d0b3d0d8c..a621ed286 100644  
\--- a/private/storage/db/sqlite\_modernc.go  
+++ b/private/storage/db/sqlite\_modernc.go  
@@ -12,8 +12,14 @@  
// See the License for the specific language governing permissions and  
// limitations under the License.  
   
\-//go:build sqlite\_modernc  
\-// +build sqlite\_modernc  
+//go:build sqlite\_modernc || !sqlite\_mattn  
+  
+// Note that above go:build expression makes modernc the default by matching  
+// the absence of sqlite\_mattn. Should there be more alternatives, update  
+// the expression to match their absence too.  
+// Also note that this default is overriden by a build configuration  
+// in .bazelrc, so this is only useful when building with "go build" and  
+// may not match what bazel build does.  
   
package db
```

Pay attention to the caveats that I had to warn about in the following comments. Choose your poison?

___

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: 45 of 47 files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4371#-NavXdOa5Kpzd-DwT4hg:-NbJKIcR5DLLAEr410Az:b-896fix) (no related file):*

I'm pretty sure we will not add more than these two sqlite implementations.
I would be more than happy to switch to modernc, but maybe let's do that after we have gained some confidence that it works.
Thus, I think the latest suggestion works and is good enough.

As for tests not running with `go test ./...`. That is an oversight.
I opened #4375 to resolve this. Please let me know if you find any other go unit tests that do not work out of the box.


</blockquote></details>

Done.

___

oncilla (approved):
> Reviewed 2 of 2 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371#-:-NbV1SWLDeWkCOMDeoAO:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

oncilla (approved):
> Dismissed @matzf from a discussion.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4371)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4361, opened 2023-07-07
doc: control service manual

matzf:
Add a conceptual description of the control plane and a reference manual for the control service.

The control plane description is mostly based on the [IETF SCION control plane specification draft](https://www.ietf.org/archive/id/draft-dekater-scion-controlplane-00.html).

The reference manual for the control service describes the many command line flags, environment variables, configuration files and APIs. This documentation highlights a variety of usability shortcomings, redundancies and inconsistencies that should be addressed eventually.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4361)

PR #4374, opened 2023-07-27
router: configuration for async packet processing

rohrerj:
This pullrequest is the second out of 3 with the goal to restructure the border router as described in the [Design Document](https://github.com/scionproto/scion/blob/master/doc/dev/design/BorderRouter.rst).

This pullrequest contains the implementation for the configurability.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4374)


rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4374)* status: 2 of 12 files reviewed, 4 unresolved discussions (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4374#-Nag2syE7JXvamqUNiol:-NahVXR-9Z2fNplsZVpN:b-896fix) (no related file):*

Please update the documentation for these variables (in `doc/manuals/router.rst`).
Follow the existing format, and the format in the manual for the control service ( #4361 ). Roughly, this would look like this:

```rst
.. program:: router-conf-toml

.. object:: general
   # ... 

# new stuff starts here:
.. object:: router
   .. option:: router.receive_buffer_size = <int> (Default: 1048576)
      # description
   .. option:: router.write_buffer_size = <int> (Default: ...)
      # ....
   .. option:: router.num_processors = <int> (Default: GOMAXPROCS)
      # ....
   # etc.
```
</blockquote></details>

Done.

___
___
*[`router/config/config.go` line 75 at r1](https://reviewable.io/reviews/scionproto/scion/4374#-NagQm1NCMdQFKdVWQqL:-NahJC7u2oNhHzo50MRI:b-t6pl28) ([raw file](https://github.com/scionproto/scion/blob/411ee8718bd62c3797172793eebcae358c15e433/router/config/config.go#L75)):*

Shouldn't we be treating the receive and send buffer sizes somewhat consistently? I can't think of a good reason why 0 doesn't mean the same thing for both. I'd change that 0 for receive buffer size to also not set a specific receive buffer size. (This needs a small change in `private/underlay/conn/conn.go`, `initConnUDP`).
</blockquote></details>

Right. They should be consistent. Because '0' means now "use system default" we cannot use '0' to to set the receive buffer to 1<<20. Either the new default value would be to use system default (0) or set the value to 1<<20 with the python topology generator. Or do you see some third option?
___
*[`router/config/sample.go` line 25 at r1](https://reviewable.io/reviews/scionproto/scion/4374#-NagML1ADQiM0SK0L7DC:-NahDUK3C0_dGxtiZJ1P:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/411ee8718bd62c3797172793eebcae358c15e433/router/config/sample.go#L25)):*

`num_processors` is missing here.
</blockquote></details>

Done.
___
*[`tools/topology/go.py` line 106 at r1](https://reviewable.io/reviews/scionproto/scion/4374#-Nag19IH79-t_yEN5PYy:-NahK3Zq7UrrQs-pdg6K:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/411ee8718bd62c3797172793eebcae358c15e433/tools/topology/go.py#L106)):*

If these are all set to the defaults, it may be better not to specify anything at all -- fewer places to change the default value.
</blockquote></details>

Done.

matzf (commented):
> Reviewed 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4374)* status: all files reviewed, 2 unresolved discussions (waiting on @rohrerj)

___
*[`doc/manuals/router.rst` line 132 at r2](https://reviewable.io/reviews/scionproto/scion/4374#-NauWpCIC0yteafrmQZN:-NauWpCIC0yteafrmQZO:bj0la4w) ([raw file](https://github.com/scionproto/scion/blob/0ef1524406570d97daa20ba954ab533d8cb491c3/doc/manuals/router.rst#L132)):*
"processors" is a bit too vague IMO. Perhaps expand this, and also expand a little bit what goroutines are (apart from this feature, an operator does not need to be familiar with this concept). 

Proposal:

> option:: router.num_processors = <int> (Default: GOMAXPROCS)
>
> Number of _goroutines_ started for SCION packets processing.
> These goroutines make the routing decision for the SCION packets by inspecting, validating and updating the path information in the packet header. Packets are processed asynchronously from the corresponding read/write operations on the individual interface sockets.
> 
> `Goroutines <https://en.wikipedia.org/wiki/Go_(programming_language)#Concurrency:_goroutines_and_channels>`_ are the Go pramming language's light-weight user-space concurrency primitives. Go's runtime schedules goroutines on top of a fixed number of kernel threads. The number of kernel threads is controlled by the ``GOMAXPROCS`` environment variable. See also `https://pkg.go.dev/runtime#hdr-Environment_Variables`_.
> 
> By default, the router uses ``GOMAXPROCS`` packet processor goroutines, i.e. exactly one goroutine for each kernel thread created by the runtime.
___
*[`doc/manuals/router.rst` line 136 at r2](https://reviewable.io/reviews/scionproto/scion/4374#-NauWV012kP4gKxteyIE:-NauWV021EKIhatlXp-w:b8qrxr1) ([raw file](https://github.com/scionproto/scion/blob/0ef1524406570d97daa20ba954ab533d8cb491c3/doc/manuals/router.rst#L136)):*
> ```reStructuredText
>    .. option:: router.num_slow_processors = <int> (Default: 1)
> 
>       The number of slow-path processors.
> ```

Ah, I just realized that this is configuring something that is not yet implemented. My preference would be to leave this away entirely for now (i.e. remove num_slow_processors from the documentation, and all the configuration structs) and add it back together with the implementation of the feature -- this will be very easy, as all the other stuff for the config is already in place.
Alternatively, it would perhaps be ok to have note here saying that this is not doing anything yet.
___
*[`router/config/config.go` line 75 at r1](https://reviewable.io/reviews/scionproto/scion/4374#-NagQm1NCMdQFKdVWQqL:-NauTmI57CT4x-1ouR7_:biy82lr) ([raw file](https://github.com/scionproto/scion/blob/411ee8718bd62c3797172793eebcae358c15e433/router/config/config.go#L75)):*

Right. They should be consistent. Because '0' means now "use system default" we cannot use '0' to to set the receive buffer to 1<<20. Either the new default value would be to use system default (0) or set the value to 1<<20 with the python topology generator. Or do you see some third option?
</blockquote></details>

Default 0 seems good enough. 
FWIW, that can't set anything other than 0 as default here is a shortcoming of the configuration "framework" -- we'd need a chance to set the default _before_ unmarshalling the config. No big deal here though.

The python generator is just a development tool, this doesn't really play a role here.

rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4374)* status: 11 of 12 files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`doc/manuals/router.rst` line 132 at r2](https://reviewable.io/reviews/scionproto/scion/4374#-NauWpCIC0yteafrmQZN:-NavnAbW4_laAG1wXy5i:bu66vk5) ([raw file](https://github.com/scionproto/scion/blob/0ef1524406570d97daa20ba954ab533d8cb491c3/doc/manuals/router.rst#L132)):*

"processors" is a bit too vague IMO. Perhaps expand this, and also expand a little bit what goroutines are (apart from this feature, an operator does not need to be familiar with this concept). 

Proposal:

> option:: router.num_processors = <int> (Default: GOMAXPROCS)
>
> Number of _goroutines_ started for SCION packets processing.
> These goroutines make the routing decision for the SCION packets by inspecting, validating and updating the path information in the packet header. Packets are processed asynchronously from the corresponding read/write operations on the individual interface sockets.
> 
> `Goroutines <https://en.wikipedia.org/wiki/Go_(programming_language)#Concurrency:_goroutines_and_channels>`_ are the Go pramming language's light-weight user-space concurrency primitives. Go's runtime schedules goroutines on top of a fixed number of kernel threads. The number of kernel threads is controlled by the ``GOMAXPROCS`` environment variable. See also `https://pkg.go.dev/runtime#hdr-Environment_Variables`_.
> 
> By default, the router uses ``GOMAXPROCS`` packet processor goroutines, i.e. exactly one goroutine for each kernel thread created by the runtime.

</blockquote></details>

Right. I didn't know how detailed the description there should be. Thank you for the proposal, I copied it.
___
*[`doc/manuals/router.rst` line 136 at r2](https://reviewable.io/reviews/scionproto/scion/4374#-NauWV012kP4gKxteyIE:-NavmrMyDTFJIK8FO1rj:brvvsef) ([raw file](https://github.com/scionproto/scion/blob/0ef1524406570d97daa20ba954ab533d8cb491c3/doc/manuals/router.rst#L136)):*

Ah, I just realized that this is configuring something that is not yet implemented. My preference would be to leave this away entirely for now (i.e. remove num_slow_processors from the documentation, and all the configuration structs) and add it back together with the implementation of the feature -- this will be very easy, as all the other stuff for the config is already in place.
Alternatively, it would perhaps be ok to have note here saying that this is not doing anything yet.

</blockquote></details>

Since that part is already reviewed and the slow-path should be implemented by end of the month, I think it should be enough to just add a note saying that it's not doing anything at the moment.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4374)* status: all files reviewed, 1 unresolved discussion (waiting on @rohrerj)

___
*[`doc/manuals/router.rst` line 132 at r2](https://reviewable.io/reviews/scionproto/scion/4374#-NauWpCIC0yteafrmQZN:-Naw0fg-EcDE8iQM49T7:b-pwskbb) ([raw file](https://github.com/scionproto/scion/blob/0ef1524406570d97daa20ba954ab533d8cb491c3/doc/manuals/router.rst#L132)):*

Right. I didn't know how detailed the description there should be. Thank you for the proposal, I copied it.
</blockquote></details>

Ok. If you look at the rendered result, it looks a bit dense now: https://scion--4374.org.readthedocs.build/en/4374/manuals/router.html#cmdoption-router-conf-toml-arg-router.num_processors



- Please add the blank lines for paragraphs.

- The `See also.` link looks a bit odd. I'd either keep the full URL in the text, or give  it some title like 

  ```
  See also the `go runtime documentation <link>`_.
  ```
- Format `GOMAXPROC` as "code" by enclosing it in *double* backticks.

rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4374)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`doc/manuals/router.rst` line 132 at r2](https://reviewable.io/reviews/scionproto/scion/4374#-NauWpCIC0yteafrmQZN:-Naxns4L3u9iJNp9YL75:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/0ef1524406570d97daa20ba954ab533d8cb491c3/doc/manuals/router.rst#L132)):*

Ok. If you look at the rendered result, it looks a bit dense now: https://scion--4374.org.readthedocs.build/en/4374/manuals/router.html#cmdoption-router-conf-toml-arg-router.num_processors



- Please add the blank lines for paragraphs.

- The `See also.` link looks a bit odd. I'd either keep the full URL in the text, or give  it some title like 

  ```
  See also the `go runtime documentation <link>`_.
  ```
- Format `GOMAXPROC` as "code" by enclosing it in *double* backticks.
</blockquote></details>

Done.

matzf (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4374#-:-Nb-mn_A1pLDXCVTY0dm:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @rohrerj)

___
*[`doc/manuals/router.rst` line 141 at r4](https://reviewable.io/reviews/scionproto/scion/4374#-Nb-ldzCD4A0aeZso2q9:-Nb-ldzCD4A0aeZso2qA:b17s54j) ([raw file](https://github.com/scionproto/scion/blob/e361454b6027356deb27ad6c995037c9ae77a9a7/doc/manuals/router.rst#L141)):*
> ```reStructuredText
>       goroutines on top of a fixed number of kernel threads. The number of kernel threads is controlled by
>       the ``GOMAXPROCS`` environment variable. See also the `go runtime documentation <https://pkg.go.dev/runtime#hdr-Environment_Variables>`_.
>       
> ```

Nit. Trailing whitespace.

PR #4377, opened 2023-08-16
slayers: unmap IPv4-mapped IPv6 addresses

oncilla:
The Go standard library can produce IPv4-mapped IPv6 addresses when resolving IP addresses. These IP addresses need to be unmapped before putting them on the wire.

Before this patch, we could observe the following with tshark:

    Len=1304 SCION 1-ff00:0:110,[::ffff:172.20.2.2] -> 1-ff00:0:111,[::ffff:172.20.3.2] UDP 32769 -> 32768 1208

The regression was introduced in #4346, which removed the unmapping behavior in slayers.PackAddr. This patch restores the behavior to ensure only unmapped IPv4 addresses make it on the wire.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4377)


marcfrei (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

___
*[`pkg/slayers/scion.go` line 409 at r1](https://reviewable.io/reviews/scionproto/scion/4377#-NbxsanOD4yC3cwaezqp:-NbxsanOD4yC3cwaezqq:b-g3fm5e) ([raw file](https://github.com/scionproto/scion/blob/4069fad297f6bda0e4ac5be7b647b93002e92d9b/pkg/slayers/scion.go#L409)):*
> ```Go
> 			t = T16Ip
> 		}
> 		return t, ip.AsSlice(), nil
> ```

With the change above the following would now maybe look a little nicer:
```
if ip.Is6() {
	return T16Ip, ip.AsSlice(), nil
}
return T4Ip, ip.AsSlice(), nil
```



_Code quote:_
```Go
		t := T4Ip
		if ip.Is6() {
			t = T16Ip
		}
		return t, ip.AsSlice(), nil
```

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @marcfrei and @matzf)

___
*[`pkg/slayers/scion.go` line 409 at r1](https://reviewable.io/reviews/scionproto/scion/4377#-NbxsanOD4yC3cwaezqp:-NcMmm7v3-ITsa3r83dV:b-tmjmb9) ([raw file](https://github.com/scionproto/scion/blob/4069fad297f6bda0e4ac5be7b647b93002e92d9b/pkg/slayers/scion.go#L409)):*

With the change above the following would now maybe look a little nicer:
```
if ip.Is6() {
	return T16Ip, ip.AsSlice(), nil
}
return T4Ip, ip.AsSlice(), nil
```
</blockquote></details>

I simplified the logic a bit more.
Unwrap can be called on everything, even invalid IP addresses.

marcfrei (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

___
*[`pkg/slayers/scion.go` line 409 at r1](https://reviewable.io/reviews/scionproto/scion/4377#-NbxsanOD4yC3cwaezqp:-NcN1WpcE4-6zQ8VtZy8:btmapzq) ([raw file](https://github.com/scionproto/scion/blob/4069fad297f6bda0e4ac5be7b647b93002e92d9b/pkg/slayers/scion.go#L409)):*

I simplified the logic a bit more.
Unwrap can be called on everything, even invalid IP addresses.
</blockquote></details>

Indeed. This reads even nicer.

marcfrei (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

matzf (commented):
Ouch. This is obviously not good, and my mistake in #4346.
IMO the fix is in the wrong place now though. If we want to put in a 4in6 address as a destination address, this should be possible at this layer.
The bug is that unmapped addresses are created and passed as destination address here, likely the result of a direct conversation from a `net.IP` using `netip.AddrFromSlice`. In particular, the call in `snet/writer.go` should unmap the IP address that comes out of `UDPAddr.Host`. Perhaps we need to revisit all `netip.AddrFromSlice` calls.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377#-:-NcMuXuw1rAIJjaBbOdB:b-fwajuv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

oncilla (commented):
Hm, I guess you have a point given that slayers should just be about encoding/decoding.

I just assumed that there is no valid use case where you would put a mapped IPv4 address on the wire, given they should never be used as source or destination of IPv6 packets.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377#-:-NcNHyQA0e0NF9mHhkay:bvv60jm)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

matzf (commented):
Ah interesting, I was not aware that these addresses are never supposed to be used on the wire (note: specified in RFC 4213, nodes SHOULD silently discard packets with IPv4-mapped IPv6 source addresses and a few other addresses).

Ideally we'd fix the callers to do the unmapping and raise an error for 4in6 addresses in `PackAddr` -- we could even raise the error _and_ allow `SetSrc/DstAddr` to continue, so that `slayers` could still be used to create bogus packets if you really, really want.

The difference is (almost) only semantics, so if you think this is too much effort, I'm also ok with keeping this as it is now.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377#-:-NcNZLJk8sR7jer1Jr2e:b75w19k)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

oncilla (commented):
I'm not 100% sure how we need to adjust the code to handle dual-stack correctly in a dispatcher-less world.
I think the reason we even get the mapped  IPv4 addresses is because we are technically in a dual-stacked world.

I guess unmapping the address in snet before invoking slayers *should* be safe. But I'm not sure if we can do this broadly for all the netip.AddrFromSlice calls. I think some of them need to keep this information (e.g. next hop needs to keep being mapped in a dual stack env AFAIU, but here I lack knowledge) 

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377#-:-NcRuTOE1crgm6ir-Mm5:bmuf95m)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

matzf (commented):
> I guess unmapping the address in snet before invoking slayers _should_ be safe. But I'm not sure if we can do this broadly for all the netip.AddrFromSlice calls

Ah, sure, I didn't mean to unmap this in for `netip.AddrFromSlice` calls. I meant to _check_ all of the calls. Those that will be turned into a SCION address should be unmapped. The ones that are underlay or other local IP connectivity, we'd leave "mapped" (guessing that this is the right thing to do there).

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377#-:-Nd-2urqBl1PXfpCbenv:b-lhe981)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

oncilla (commented):
Ack. I vetted some locations where the netip.AddrFromSlice is called. But they are so distant from the place where it is passed to snet and in turn to slayers that it seems very easy to forget to do the unmap.

We could do it in snet, but implementations not using snet can easily run into this issue.
I have a feeling that we save us and the users of the slayers package a lot of trouble by doing the unmapping in slayers.

I wonder if we should have a follow-up PR that validates in the router that we do not forward mapped addresses too?

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377#-:-NdjZMD10Fqbc6NcWYcl:b-g9m4s6)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

matzf (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4377)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

matzf:
> I wonder if we should have a follow-up PR that validates in the router that we do not forward mapped addresses too?

Created #4415

PR #4381, opened 2023-08-18
dispatcher: support auto-assigned port for SVC registrations

### LINKED ISSUE #4379: Service cannot listen on dynamically assigned port with Svc ID
ran into this while experimenting with registering the resolver service port==0  (as a way to not occupy the port from the public address).

It appears that a port from the dynamic range is properly assigned by the dispatcher but that the registration of the SVC address fails with Cause=="zero port".

This is because the dynamically assigned port isn't passed to the service registration. The original port 0 is passed instead.

This looks like a bug: there is no obvious benefit to rejecting such a configuration and, for better or for worse. if I fix that, the resolver works just fine.

jiceatscion:
fixes #4379

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4381)

### DIFF
--- a/dispatcher/internal/registration/table.go
+++ b/dispatcher/internal/registration/table.go
@@ -60,7 +60,7 @@ func (t *Table) Register(public *net.UDPAddr, bind net.IP, svc addr.SVC,
 	if bind == nil {
 		bind = public.IP
 	}
-	svcRef, err := t.insertSVCIfRequested(svc, bind, public.Port, value)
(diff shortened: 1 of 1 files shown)


oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4381#-:-Nc8HBn_DCvhMTnIfH2U:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4383, opened 2023-08-23
router: slow-path in async packet processing

rohrerj:
This pullrequest is the third out of 3 with the goal to restructure the border router as described in the [Design Document](https://github.com/scionproto/scion/blob/master/doc/dev/design/BorderRouter.rst).

This pullrequest contain the implementation of the slow-path.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)


matzf (changes requested):
> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: 1 of 5 files reviewed, 6 unresolved discussions (waiting on @rohrerj)

___
*[`router/dataplane.go` line 572 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd9qx6jDUAy1U-j3P3N:-Nd9qx6k7rEYQ9_4Ge_T:b-o82idc) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L572)):*
> ```Go
> 	packet        packet
> 	slowPathState slowPathState
> }
> ```

- camelCase, not snake_case
- embed `packet` to make the access somewhat shorter?
___
*[`router/dataplane.go` line 807 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd5YzTk0YSkrhslKH6E:-Nd5YzTk0YSkrhslKH6F:b-dnhp9i) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L807)):*
> ```Go
> 	// DRKey key derivation for SCMP authentication
> 	drkeyProvider drkeyProvider
> }
> ```

-  unused `mac` and `cachedMac`
- the `macBuffers` whcih is shared with `scionPacketProcessor` is a bit silly now -- the slow processor only uses `drkeyInput`, the normal processor uses the other two buffers. I think we can simplify this be getting rid of the struct and just using a `macInputBuffer []byte`, initializing it to the max size of the different buffers. This seems safe.
___
*[`router/dataplane.go` line 1203 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4nh4R8QHVRpPPu8dM:-Nd4nh4S9BfXuvuFmNsW:by3d7pv) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L1203)):*
> ```Go
> 	bfdLayer layers.BFD
> 
> 	slowPathState slowPathState
> ```

Don't make this a member here. This can be part of the return value from `processPkt`. Either in the `processingResult` (and do _not_ return an error) or in the error type. Conceptually, I think both ways are ok, but I believe if we pack it in the error, we cannot avoid the heap allocation.
___
*[`router/dataplane.go` line 1211 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4qMvb83Cmpt2-09as:-Nd4qMvb83Cmpt2-09at:bdlhmi5) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L1211)):*
> ```Go
> 	code       slayers.SCMPCode
> 	parameters []interface{}
> 	cause      error
> ```

- `slowPathState`: I don't think "state" is a good name for this. Something like "request", or "metadata", or something along these lines would be more fitting.
- mode:
  - "mode" seems an odd name, perhaps "type"?
  - this should be an "enum" (`type int bla; const bla ... = iota; ....`)
  - maybe the name for traceroute should rather be more generic "router alert" 
- parameters: this will always allocate and needs ugly casts. Instead, just add the union of all parameters that we'll need. It's not our fault this looks stupid, it's go's fault for not having ergonomic sum types.
- cause: ok to pass a reference to a predefined global variable, but we should be clean up the places that allocate new errors
___
*[`router/dataplane.go` line 2155 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4tdxF-VX9wgncY-3X:-Nd4tdxF-VX9wgncY-3Y:bn8nmji) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L2155)):*
> ```Go
> 	}
> 
> 	return p.buffer.Bytes(), scmpError{TypeCode: scmpH.TypeCode, Cause: cause}
> ```

IMO we should remove the scmpError now. In the slowPathProcessor, an scmp error is the expected result, I don't think this should be treated as an error anymore. 
It's also not used for anything useful, only a debug log statement uses this information. This, we can just move somewhere else 
We can just as well move the Debug log statement somewhere else.
___
*[`router/export_test.go` line 83 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4xT0N0GdOwTekwGgO:-Nd4xT0N0GdOwTekwGgP:bkgics2) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/export_test.go#L83)):*
> ```Go
> 	}
> 	result, err := p.processPkt(m.Buffers[0], srcAddr, ifID)
> 	if errors.Is(err, slowPathRequired) {
> ```

IMO squeezing the slow path processing into the same tests is the wrong approach here. Instead, I'd rather adapt the existing test cases so that they only check that the slow path processing is triggered when necessary. Add separate test cases for the actual slow path processing.

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: 1 of 5 files reviewed, 7 unresolved discussions (waiting on @rohrerj)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4383#-Nd9rrpi8KZy3YeWHgo4:-Nd9rrpi8KZy3YeWHgo5:b-2xtrje) (no related file):*
The overall structure looks good to me :thumbsup: 

As we move some methods from one type to another, the resulting "physical" organization of the definitions etc. in the `dataplane.go` file is rather messy now. Perhaps it is good to keep this as is for this PR to keep the diff small, and do a separate cleanup pass later.

___

rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: 1 of 5 files reviewed, 6 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 572 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd9qx6jDUAy1U-j3P3N:-NdC26as-MXPPwaV9c9v:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L572)):*

- camelCase, not snake_case
- embed `packet` to make the access somewhat shorter?
</blockquote></details>

Done.
___
*[`router/dataplane.go` line 807 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd5YzTk0YSkrhslKH6E:-NdC3xwL3DzDdv-_MQeJ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L807)):*

-  unused `mac` and `cachedMac`
- the `macBuffers` whcih is shared with `scionPacketProcessor` is a bit silly now -- the slow processor only uses `drkeyInput`, the normal processor uses the other two buffers. I think we can simplify this be getting rid of the struct and just using a `macInputBuffer []byte`, initializing it to the max size of the different buffers. This seems safe.
</blockquote></details>

Done.
___
*[`router/dataplane.go` line 1203 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4nh4R8QHVRpPPu8dM:-NdCcWnECBs8kAMuMtC2:b-kvdzse) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L1203)):*

Don't make this a member here. This can be part of the return value from `processPkt`. Either in the `processingResult` (and do _not_ return an error) or in the error type. Conceptually, I think both ways are ok, but I believe if we pack it in the error, we cannot avoid the heap allocation.
</blockquote></details>

I included the slowPathRequest in the processingResult but I still had to return an error because the entire logic of processPkt works like "continue if no error". To make it work without returning an error it would require a lot of changes in the packet processing logic.
___
*[`router/dataplane.go` line 1211 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4qMvb83Cmpt2-09as:-NdCMBIYA-mMVIUAXt7G:b-ysaxw7) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L1211)):*

- `slowPathState`: I don't think "state" is a good name for this. Something like "request", or "metadata", or something along these lines would be more fitting.
- mode:
  - "mode" seems an odd name, perhaps "type"?
  - this should be an "enum" (`type int bla; const bla ... = iota; ....`)
  - maybe the name for traceroute should rather be more generic "router alert" 
- parameters: this will always allocate and needs ugly casts. Instead, just add the union of all parameters that we'll need. It's not our fault this looks stupid, it's go's fault for not having ergonomic sum types.
- cause: ok to pass a reference to a predefined global variable, but we should be clean up the places that allocate new errors
</blockquote></details>

The problem with the "cause" field is that if we have a parametrized error, we either have to create a new error here or pass all parameters such that we can reconstruct the error. But the union of all those parameters would get rather large and we would need a lot of additional logic to reconstruct the correct errors. So what would you suggest here?
___
*[`router/dataplane.go` line 2155 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4tdxF-VX9wgncY-3X:-NdCP4mZC4kVIMzvIOM4:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L2155)):*

IMO we should remove the scmpError now. In the slowPathProcessor, an scmp error is the expected result, I don't think this should be treated as an error anymore. 
It's also not used for anything useful, only a debug log statement uses this information. This, we can just move somewhere else 
We can just as well move the Debug log statement somewhere else.
</blockquote></details>

Done.
___
*[`router/export_test.go` line 83 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4xT0N0GdOwTekwGgO:-NdCdG3d-LgP8mIoR5pW:b4xsv5t) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/export_test.go#L83)):*

IMO squeezing the slow path processing into the same tests is the wrong approach here. Instead, I'd rather adapt the existing test cases so that they only check that the slow path processing is triggered when necessary. Add separate test cases for the actual slow path processing.
</blockquote></details>

I created a separate function in export_test for slow path processing. Some already existing tests require slow path and because of that I moved them to a new test function for slow path.

matzf (changes requested):
It only occurred to me now that "path" is a heavily used term in SCION, so calling this "slow-path" might be puzzling. If someone has a good idea for a better term, it might still be worth quickly renaming this.

> Reviewed 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383#-:-NdZDC4E7nbFULnZGd81:b-fl1cqx)* status: 2 of 5 files reviewed, 5 unresolved discussions (waiting on @rohrerj)

___
*[`router/dataplane.go` line 1203 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4nh4R8QHVRpPPu8dM:-NdZ4-Hz3ABu8ta5-u7c:brxp9sq) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L1203)):*

I included the slowPathRequest in the processingResult but I still had to return an error because the entire logic of processPkt works like "continue if no error". To make it work without returning an error it would require a lot of changes in the packet processing logic.
</blockquote></details>

Sounds good.
___
*[`router/dataplane.go` line 1211 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4qMvb83Cmpt2-09as:-NdVkFpt1EfhWsNwZ40W:b6wkckg) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L1211)):*

The problem with the "cause" field is that if we have a parametrized error, we either have to create a new error here or pass all parameters such that we can reconstruct the error. But the union of all those parameters would get rather large and we would need a lot of additional logic to reconstruct the correct errors. So what would you suggest here?
</blockquote></details>

Right, ignore the point on the "cause". It's not really related, we can clean this up separately. The idea would be to just drop the additional information in the error. This is only used for debug logging, so we can instead just do the debug logging in place (if it is enabled), and return simple error objects.
___
*[`router/dataplane.go` line 676 at r2](https://reviewable.io/reviews/scionproto/scion/4383#-NdVcg2I10LvfHWW6y5i:-NdVcg2J0GsyxA8Fb406:b-b7bd36) ([raw file](https://github.com/scionproto/scion/blob/14502311a2f5082d3fd1edfb627aba6978ca8d5f/router/dataplane.go#L676)):*
> ```Go
> 				packet:          p,
> 				slowPathRequest: result.SlowPathRequest,
> 			}:
> ```

Nit: bad formatting by gofmt here, closing brace visually matches up with the one from select. It's weird...
This can fit on a single line though, I think that makes it much more readable.



_Suggestion:_
```Go
			case slowQ <- slowPacket{p, result.SlowPathRequest}:
```
___
*[`router/dataplane.go` line 849 at r2](https://reviewable.io/reviews/scionproto/scion/4383#-NdVjqf1-GHxynmnDU9W:-NdVjqf1-GHxynmnDU9X:bw5c4y3) ([raw file](https://github.com/scionproto/scion/blob/14502311a2f5082d3fd1edfb627aba6978ca8d5f/router/dataplane.go#L849)):*
> ```Go
> 		return p.handleSCMPTraceRouteRequest(pkt.slowPathRequest.interfaceId)
> 	default:
> 		return processResult{}, serrors.New("Unsupported slow-path type", "type",
> ```

panic instead?
___
*[`router/dataplane.go` line 1179 at r2](https://reviewable.io/reviews/scionproto/scion/4383#-NdVjzkc4y8WOMJrDCbJ:-NdVjzkc4y8WOMJrDCbK:undefined) ([raw file](https://github.com/scionproto/scion/blob/14502311a2f5082d3fd1edfb627aba6978ca8d5f/router/dataplane.go#L1179)):*
> ```Go
> 
> const slowPathSCMP slowPathType = 0
> const slowPathRouterAlert slowPathType = 1
> ```





_Suggestion:_
```Go
const (
	slowPathSCMP slowPathType = iota
	slowPathRouterAlert
)
```
___
*[`router/dataplane_test.go` line 653 at r2](https://reviewable.io/reviews/scionproto/scion/4383#-NdZAGjzAMzi60o3Ex8v:-NdZAGjzAMzi60o3Ex8w:b-i9k52t) ([raw file](https://github.com/scionproto/scion/blob/14502311a2f5082d3fd1edfb627aba6978ca8d5f/router/dataplane_test.go#L653)):*
> ```Go
> 			result, err := dp.ProcessPkt(tc.srcInterface, input)
> 			assert.ErrorIs(t, err, router.SlowPathRequired)
> 			_, err = dp.ProcessSlowPath(tc.srcInterface, input, result.SlowPathRequest)
> ```

This test currently doesn't check very much other than running the code a bit. It's not worse than before, but it seems like this would be an opportunity to to improve this.
I'd split the test; one part checks that the `ProcessPkt` creates the expected `slowPathRequest` -- like the test here, but instead of calling `ProcessSlowPath`, compare the parameters of the `result.SlowPathRequest` with some expected values.
A separate part defines `slowPathRequest` as inputs and checks that the resulting packet is the expected SCMP message.

rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 1211 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4qMvb83Cmpt2-09as:-Ndbm3sa8GIOdAXiRhar:bfly5r7) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L1211)):*

Right, ignore the point on the "cause". It's not really related, we can clean this up separately. The idea would be to just drop the additional information in the error. This is only used for debug logging, so we can instead just do the debug logging in place (if it is enabled), and return simple error objects.
</blockquote></details>

I changed the code to only use errors without parameters for the cause and added a debug log statement for all those scmp cases.
___
*[`router/dataplane.go` line 849 at r2](https://reviewable.io/reviews/scionproto/scion/4383#-NdVjqf1-GHxynmnDU9W:-NdbR8RG9Qy9Qw0IVzQ3:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/14502311a2f5082d3fd1edfb627aba6978ca8d5f/router/dataplane.go#L849)):*

panic instead?
</blockquote></details>

Done.
___
*[`router/dataplane.go` line 1179 at r2](https://reviewable.io/reviews/scionproto/scion/4383#-NdVjzkc4y8WOMJrDCbJ:-NdbPQuF9D8x4Ri4mJ21:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/14502311a2f5082d3fd1edfb627aba6978ca8d5f/router/dataplane.go#L1179)):*
> ```Go
> 
> const slowPathSCMP slowPathType = 0
> const slowPathRouterAlert slowPathType = 1
> ```

Done.
___
*[`router/dataplane_test.go` line 653 at r2](https://reviewable.io/reviews/scionproto/scion/4383#-NdZAGjzAMzi60o3Ex8v:-NdbmQPp4emT_miDhORE:b-h924d3) ([raw file](https://github.com/scionproto/scion/blob/14502311a2f5082d3fd1edfb627aba6978ca8d5f/router/dataplane_test.go#L653)):*

This test currently doesn't check very much other than running the code a bit. It's not worse than before, but it seems like this would be an opportunity to to improve this.
I'd split the test; one part checks that the `ProcessPkt` creates the expected `slowPathRequest` -- like the test here, but instead of calling `ProcessSlowPath`, compare the parameters of the `result.SlowPathRequest` with some expected values.
A separate part defines `slowPathRequest` as inputs and checks that the resulting packet is the expected SCMP message.
</blockquote></details>

I modified the test to check whether the returned slowPathRequest is as expected. But it is hard to check whether the out packet corresponds to a expected packet because we don't have nice functions to create / parse SCMP packets. (or at least I was not able to find one). And hardcode a byte array for comparison is hard to read / understand.

matzf (commented):
> Reviewed 1 of 5 files at r1, 2 of 3 files at r2, 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @rohrerj)

___
*[`router/dataplane.go` line 1211 at r1](https://reviewable.io/reviews/scionproto/scion/4383#-Nd4qMvb83Cmpt2-09as:-NdjgIjwEvZxOaKYkHg0:bmi8ppe) ([raw file](https://github.com/scionproto/scion/blob/bc00f274f157775090b43242a6d1c7e7fcd188dc/router/dataplane.go#L1211)):*

I changed the code to only use errors without parameters for the cause and added a debug log statement for all those scmp cases.
</blockquote></details>

:100: 
___
*[`router/dataplane.go` line 138 at r3](https://reviewable.io/reviews/scionproto/scion/4383#-NdjljDy0fyk0ltNsOpV:-NdjljDy0fyk0ltNsOpW:bx362ug) ([raw file](https://github.com/scionproto/scion/blob/650a640d1e8e540ac4a031ad17463f056ca34354/router/dataplane.go#L138)):*
> ```Go
> 	macVerificationFailed         = serrors.New("MAC verification failed")
> 	badPacketSize                 = serrors.New("bad packet size")
> 	slowPathRequired              = serrors.New("slow-path required")
> ```

Somewhat aside (non-blocking): these `serrors` errors all include a rather pointless stacktrace which just clutters the log. We can easily improve this by creating the values with `errors.New` instead.
___
*[`router/dataplane_test.go` line 653 at r2](https://reviewable.io/reviews/scionproto/scion/4383#-NdZAGjzAMzi60o3Ex8v:-NdjfQGX7Udt-V3FYlQ-:b-n3ji6m) ([raw file](https://github.com/scionproto/scion/blob/14502311a2f5082d3fd1edfb627aba6978ca8d5f/router/dataplane_test.go#L653)):*

I modified the test to check whether the returned slowPathRequest is as expected. But it is hard to check whether the out packet corresponds to a expected packet because we don't have nice functions to create / parse SCMP packets. (or at least I was not able to find one). And hardcode a byte array for comparison is hard to read / understand.
</blockquote></details>

Ok.  Well there is `slayers.SCMP` and the corresponding messages like `slayers.SCMPParameterProblem` if you want to try. There are a load of examples on how to use these, e.g. in the `braccept/cases`.
___
*[`router/export_test.go` line 105 at r3](https://reviewable.io/reviews/scionproto/scion/4383#-Ndjsq9_8WY32ONn2E0_:-Ndjsq9_8WY32ONn2E0a:bsn56pf) ([raw file](https://github.com/scionproto/scion/blob/650a640d1e8e540ac4a031ad17463f056ca34354/router/export_test.go#L105)):*
> ```Go
> }
> 
> type SlowPathRequest struct {
> ```

Just move it to the "internal" tests instead of jumping through hoops to pretend exposing this?

rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`router/dataplane.go` line 138 at r3](https://reviewable.io/reviews/scionproto/scion/4383#-NdjljDy0fyk0ltNsOpV:-Ne0VasK552Y2uDtzJWL:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/650a640d1e8e540ac4a031ad17463f056ca34354/router/dataplane.go#L138)):*

Somewhat aside (non-blocking): these `serrors` errors all include a rather pointless stacktrace which just clutters the log. We can easily improve this by creating the values with `errors.New` instead.
</blockquote></details>

Done
___
*[`router/dataplane_test.go` line 653 at r2](https://reviewable.io/reviews/scionproto/scion/4383#-NdZAGjzAMzi60o3Ex8v:-Ne0VdXi5XUhGfK46iEA:b-z7k6wu) ([raw file](https://github.com/scionproto/scion/blob/14502311a2f5082d3fd1edfb627aba6978ca8d5f/router/dataplane_test.go#L653)):*

Ok.  Well there is `slayers.SCMP` and the corresponding messages like `slayers.SCMPParameterProblem` if you want to try. There are a load of examples on how to use these, e.g. in the `braccept/cases`.
</blockquote></details>

I added checks that the serialized scmp type code matches the expectation and that the correct scmp error layer type is contained in the layers (e.g. the slayers.LayerTypeSCMPParameterProblem)
___
*[`router/export_test.go` line 105 at r3](https://reviewable.io/reviews/scionproto/scion/4383#-Ndjsq9_8WY32ONn2E0_:-Ne0WEk26LXJ1KOKocvD:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/650a640d1e8e540ac4a031ad17463f056ca34354/router/export_test.go#L105)):*

Just move it to the "internal" tests instead of jumping through hoops to pretend exposing this?
</blockquote></details>

Done.

matzf (changes requested):
> Reviewed 3 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: 4 of 5 files reviewed, 7 unresolved discussions (waiting on @rohrerj)

___
*[`router/dataplane_internal_test.go` line 46 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne2EStf9zavE9dUyThL:-Ne2EStf9zavE9dUyThM:b-4n1zen) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L46)):*
> ```Go
> )
> 
> var key = []byte("testkey_xxxxxxxx")
> ```

`testKey`, or `macKey` or `hopfieldKey` or so.
Btw, if you make this a `string` it can be defined as const. Just pass a string to `computeMAC` and cast to `[]byte` internally.
___
*[`router/dataplane_internal_test.go` line 424 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne27fWXAyWhRfAjfxdF:-Ne27fWXAyWhRfAjfxdG:bklr9pm) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L424)):*
> ```Go
> 	ctrl := gomock.NewController(t)
> 	defer ctrl.Finish()
> 	key := []byte("testkey_xxxxxxxx")
> ```

Use global `key` instead.
___
*[`router/dataplane_internal_test.go` line 505 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne29yjZ9_00-jHFdXC5:-Ne29yjZ9_00-jHFdXC6:b-77cc77) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L505)):*
> ```Go
> 			dp := tc.prepareDP(ctrl)
> 			input := tc.mockMsg()
> 			result, err := dp.ProcessPkt(tc.srcInterface, input)
> ```

If we invoke the processor / slow processor directly, we can avoid the dance with the ipv4 message and the test-exported `dp.ProcessSlowPath`.
I'm not entirely sure if this really ends up as more readable & maintanable. Please take another look and decide.
___
*[`router/dataplane_internal_test.go` line 513 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne2CBsR0sX7aK3bAcw2:-Ne2CBsR0sX7aK3bAcw3:blk8qbg) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L513)):*
> ```Go
> 				cause:    result.SlowPathRequest.cause,
> 			}
> 			assert.Equal(t, tc.expectedSlowPathRequest, selectedResult)
> ```

Just `result.SlowPathRequest` instead of building `selectedResult`.
___
*[`router/dataplane_internal_test.go` line 521 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne2DG-DE5syiFK0Ww_w:-Ne2DG-DE5syiFK0Ww_x:brmvgg2) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L521)):*
> ```Go
> 			packet := gopacket.NewPacket(result.OutPkt, slayers.LayerTypeSCION, gopacket.Default)
> 			scmpLayer := packet.Layer(slayers.LayerTypeSCMP)
> 			scmp := scmpLayer.(*slayers.SCMP)
> ```

Condense



_Suggestion:_
```Go
			scmp := packet.Layer(slayers.LayerTypeSCMP).(*slayers.SCMP)
```
___
*[`router/dataplane_internal_test.go` line 532 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne2CgmY89G82InO2W14:-Ne2CgmY89G82InO2W15:b-9s2h0s) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L532)):*
> ```Go
> 				}
> 			}
> 			assert.True(t, layerFound)
> ```

Use `packet.Layer()` to find the expected layer. Returns `nil` if not found.



_Suggestion:_
```Go
			assert.NotNil(t, packet.Layer(tc.expectedLayerType))
```
___
*[`router/dataplane_internal_test.go` line 540 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne2BdodD8DYWyiOY8We:-Ne2BdodD8DYWyiOY8Wf:b-8xl37j) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L540)):*
> ```Go
> 	t.Helper()
> 	ret := &ipv4.Message{}
> 	spkt.Path = dpath
> ```

Remove this effectively unused `dpath` parameter.

rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: 3 of 5 files reviewed, 7 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane_internal_test.go` line 46 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne2EStf9zavE9dUyThL:-NeHrKdl773WpDOq-ef3:b4lrlf0) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L46)):*

`testKey`, or `macKey` or `hopfieldKey` or so.
Btw, if you make this a `string` it can be defined as const. Just pass a string to `computeMAC` and cast to `[]byte` internally.
</blockquote></details>

I renamed it to testKey. The problem with making this a string is that it is not only used for computeMAC but also for the NewDP function and since we use it at multiple places and everywhere in form of a byte array, I think it makes sense to leave it like that.
___
*[`router/dataplane_internal_test.go` line 424 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne27fWXAyWhRfAjfxdF:-NeHqCD3AaGxLRQJocdu:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L424)):*

Use global `key` instead.
</blockquote></details>

Done.
___
*[`router/dataplane_internal_test.go` line 505 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne29yjZ9_00-jHFdXC5:-NeI8BjR8bRh5sTYbkQV:bx3rbvh) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L505)):*

If we invoke the processor / slow processor directly, we can avoid the dance with the ipv4 message and the test-exported `dp.ProcessSlowPath`.
I'm not entirely sure if this really ends up as more readable & maintanable. Please take another look and decide.
</blockquote></details>

Since we moved it to the internal test I think it makes sense to change that.
___
*[`router/dataplane_internal_test.go` line 513 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne2CBsR0sX7aK3bAcw2:-NeI8o_W03wUxfIgMggd:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L513)):*

Just `result.SlowPathRequest` instead of building `selectedResult`.
</blockquote></details>

Done.
___
*[`router/dataplane_internal_test.go` line 521 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne2DG-DE5syiFK0Ww_w:-NeI9-bUE2RDpTNG0f-P:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L521)):*

Condense
</blockquote></details>

Done.
___
*[`router/dataplane_internal_test.go` line 532 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne2CgmY89G82InO2W14:-NeI98oOAdNp9r7jAThW:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L532)):*

Use `packet.Layer()` to find the expected layer. Returns `nil` if not found.
</blockquote></details>

Done.
___
*[`router/dataplane_internal_test.go` line 540 at r4](https://reviewable.io/reviews/scionproto/scion/4383#-Ne2BdodD8DYWyiOY8We:-NeI9J082WPzWAbENqIu:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/edc85232a24f303445feb5a7b03f0c5177d3becb/router/dataplane_internal_test.go#L540)):*

Remove this effectively unused `dpath` parameter.
</blockquote></details>

Done.

matzf (approved):
:lgtm:

> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383#-:-NfGNVUq3kK9Od8paxP_:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rohrerj)

matzf (changes requested):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: all files reviewed, 3 unresolved discussions (waiting on @rohrerj)

___
*[`router/dataplane.go` line 139 at r6](https://reviewable.io/reviews/scionproto/scion/4383#-NfPcn6R-de5ShDWcibS:-NfPcn6R-de5ShDWcibT:b-rlqj9k) ([raw file](https://github.com/scionproto/scion/blob/96d2886f00384dc3cb9a620d3691f24a5e436a61/router/dataplane.go#L139)):*
> ```Go
> 	errPeeringNonemptySeg2        = serrors.New("non-zero-length segment[2] in peering path")
> 	errShortPacket                = serrors.New("Packet is too short")
> 	errBFDSessionDown             = serrors.New("bfd session down")
> ```

Bad merge, you wanted to change this all to `errors.New`.
___
*[`router/dataplane.go` line 820 at r6](https://reviewable.io/reviews/scionproto/scion/4383#-NfPj6kdBowBiYFpHTcg:-NfPj6kdBowBiYFpHTch:bh5rkpx) ([raw file](https://github.com/scionproto/scion/blob/96d2886f00384dc3cb9a620d3691f24a5e436a61/router/dataplane.go#L820)):*
> ```Go
> 	drkeyProvider drkeyProvider
> 
> 	peering bool
> ```

I see this is just copied over from the "scionPacketProcessor", but I think here this structure is (even more) sub-optimal. There is only a single place that uses this variable. I think this would be much clearer if we'd just move the `determinePeer` call into `prepareSCMP`, where this is used.
(non-blocking -- it's okay-ish for now)
___
*[`router/dataplane.go` line 2119 at r6](https://reviewable.io/reviews/scionproto/scion/4383#-NfPjYOl6Yudd8iaRHVp:-NfPjYOl6Yudd8iaRHVq:b-9tmoxc) ([raw file](https://github.com/scionproto/scion/blob/96d2886f00384dc3cb9a620d3691f24a5e436a61/router/dataplane.go#L2119)):*
> ```Go
> 		return nil, serrors.Wrap(cannotRoute, err, "details", "reversing path for SCMP")
> 	}
> 	revPath := revPathTmp.(*scion.Decoded)
> ```

See comment on `slowPathPacketProcessor.peering`; add `peering, err := determinePeer(revPath)` e.g. here.

rohrerj (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 139 at r6](https://reviewable.io/reviews/scionproto/scion/4383#-NfPcn6R-de5ShDWcibS:-NfPzF5R6rXP2DRbo8bH:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/96d2886f00384dc3cb9a620d3691f24a5e436a61/router/dataplane.go#L139)):*

Bad merge, you wanted to change this all to `errors.New`.
</blockquote></details>

Done.
___
*[`router/dataplane.go` line 820 at r6](https://reviewable.io/reviews/scionproto/scion/4383#-NfPj6kdBowBiYFpHTcg:-NfQ4m6c8kR-jnxUWw1D:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/96d2886f00384dc3cb9a620d3691f24a5e436a61/router/dataplane.go#L820)):*

I see this is just copied over from the "scionPacketProcessor", but I think here this structure is (even more) sub-optimal. There is only a single place that uses this variable. I think this would be much clearer if we'd just move the `determinePeer` call into `prepareSCMP`, where this is used.
(non-blocking -- it's okay-ish for now)
</blockquote></details>

Done.
___
*[`router/dataplane.go` line 2119 at r6](https://reviewable.io/reviews/scionproto/scion/4383#-NfPjYOl6Yudd8iaRHVp:-NfQ4pX_1UruvTmdUWu-:b-r05lwx) ([raw file](https://github.com/scionproto/scion/blob/96d2886f00384dc3cb9a620d3691f24a5e436a61/router/dataplane.go#L2119)):*

See comment on `slowPathPacketProcessor.peering`; add `peering, err := determinePeer(revPath)` e.g. here.
</blockquote></details>

I had to make some changes to determinePeer() because revPath was not of type scion.Raw. But now it should work.

matzf (approved):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4383)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rohrerj)

___
*[`router/dataplane.go` line 820 at r6](https://reviewable.io/reviews/scionproto/scion/4383#-NfPj6kdBowBiYFpHTcg:-NfQmgFHCKMVZD8iJaSh:bmi8ppe) ([raw file](https://github.com/scionproto/scion/blob/96d2886f00384dc3cb9a620d3691f24a5e436a61/router/dataplane.go#L820)):*

Done.
</blockquote></details>

:100:

PR #4299, opened 2022-11-25
protocol: implement peering links 

benthor:
This is an attempt to address Issue #4093

Since the feature branch referenced by the original PR #4249 wasn't correctly rebased, we'll start over with a clean slate as suggested by @oncilla .

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4299)

PR #4390, opened 2023-09-11
Peering links continued

jiceatscion:
Restarting where https://github.com/scionproto/scion/pull/4299 left off.

Added test cases.
Applied reviewer's suggestions.
Changed a field name that was miss-leading.
Added some comments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)


matzf (changes requested):
> Reviewed 5 of 10 files at r1, 13 of 13 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: all files reviewed, 13 unresolved discussions (waiting on @jiceatscion)

___
*[`control/beaconing/extender.go` line 106 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne7m9X6-gIZmjD6c1xB:-Ne7m9X74IQThuBHY6yl:b-73n6h6) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/control/beaconing/extender.go#L106)):*
> ```Go
> 
> 	// FIXME: why do we compute Beta(pseg) a second time? It hasn't changed since last
> 	// time.
> ```

+1, reusing the `extractBeta(pseg)` value would make this a lot clearer.
___
*[`router/dataplane.go` line 1035 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne7sXfBErW5-8EWXEM7:-Ne7sXfBErW5-8EWXEM8:bak45yn) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane.go#L1035)):*
> ```Go
> 	// effectiveXover indicates if a cross-over segment change was done during processing.
> 	effectiveXover bool
> 	// peering indicates that the packet is inside of a peering AS
> ```

Nit: i find the term "peering AS" to be a bit inaccurate, I'd prefer "peering hop field".  E.g. "peering indicates that the current hop is a peering hop field"
___
*[`router/dataplane.go` line 1291 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne7pTY46kUi0j5q-vW9:-Ne7pTY46kUi0j5q-vWA:b8anga6) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane.go#L1291)):*
> ```Go
> 	case ingress == topology.Child && egress == topology.Child:
> 		return processResult{}, nil
> 	case ingress == topology.Unset && egress == topology.Peer:
> ```

The comment above seems to say this that this case is an error, and think I agree. 
I don't think a peering link could ever be a valid egress interface after a cross-over.
___
*[`router/dataplane.go` line 1367 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8EPgTEFaySxovNF3t:-Ne8EPgTEFaySxovNF3u:b-2ipixl) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane.go#L1367)):*
> ```Go
> 	// We are the egress router and if we go in construction direction we
> 	// need to update the SegID (unless we are effecting a peering hop).
> 	// When we're at a peering hop, the SegID for this HOP and for the next
> ```

Nit. HOP -> hop (looks like an acronym otherwise)
___
*[`router/dataplane_test.go` line 727 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8HaYKB9eZANAdw0Du:-Ne8HaYKB9eZANAdw0Dv:b-rnvhvv) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane_test.go#L727)):*
> ```Go
> 				// info field because computeMAC makes that easy.
> 				dpath.HopFields[1].Mac = computeMAC(t, key, dpath.InfoFields[1], dpath.HopFields[1])
> 				dpath.HopFields[2].Mac = computeMAC(t, key, dpath.InfoFields[1], dpath.HopFields[2])
> ```

Nit. Just to make sure there is no confusion here, the MAC for hop field 2 would in practice be computed with the secret key of a different AS. It might be good to illustrate this, by just using a different key.
___
*[`router/dataplane_test.go` line 741 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8Rkh7BO-yjI3frnQv:-Ne8Rkh7BO-yjI3frnQw:b-5jzrsx) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane_test.go#L741)):*
> ```Go
> 				return ret
> 			},
> 			srcInterface: 1, // from peering link
> ```

Somewhat aside: it's a bit strange that this test harness does not assert that the egress interface (`result.EgressID`) is correct.
___
*[`tools/braccept/main.go` line 132 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8Z8er4WZ5-Mkgma7Y:-Ne8Z8er4WZ5-Mkgma7Z:b-9j3bdk) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/main.go#L132)):*
> ```Go
> 		cases.JumboPacket(artifactsDir, hfMAC),
> 		cases.ChildToChildPeeringOut(artifactsDir, hfMAC),
> 		cases.ChildToChildPeeringIn(artifactsDir, hfMAC),
> ```

Should we perhaps add (a) test(s) for processing the hop(s) before/after a peering hop? 
The processing in these cases is identical to the various ChildToParent / ParentToChild test cases, the only difference is that the path segments have the peer flag set, which the router _should_ effectively ignore.

Another case that might be interesting to cover would be `PeerToInternalHost` / `InternalHostToPeer`, i.e. a path ending/starting in a peering hop.

Note: all of these cases are also covered by the running the "end2end_integration" and scion_integration tests in the default topology, but having an explicit test could still be helpful.
___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 1 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8TJLU8P668mYOqohh:-Ne8TJLU8P668mYOqohi:btcovqk) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L1)):*
> ```Go
> ```

// Copyright 2023 SCION Association
___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8TPuT4i1CJaxgoZEg:-Ne8TPuUBfNK9vGh6l50:b-bnir3f) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L41)):*
> ```Go
> // The peering hop is the first hop on the second segment: it crosses from a peering
> // interface to a child interface.
> func ChildToChildPeeringIn(artifactsDir string, mac hash.Hash) runner.Case {
> ```

Following the naming scheme for the other braccept cases, this would be `PeerToChild`. The name does not describe the global path, but the hop that is processed at the router.
___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 47 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8V7Dp-FmVEWPWI12y:-Ne8V7Dp-FmVEWPWI12z:b-qog00s) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L47)):*
> ```Go
> 	}
> 
> 	// We inject the packet into A (at I/F 121) as if coming from 2 (at I/F 211)
> ```

Nit. I believe we've usually abbreviated interface as "IF", not "I/F". 
___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 107 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8UOUI19C1INoT7EFA:-Ne8UOUI19C1INoT7EFB:bkyk9yz) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L107)):*
> ```Go
> 	// in construction direction, HF[2]. Therefore, SEG[1]'s SegID.
> 	sp.HopFields[1].Mac = path.MAC(mac, sp.InfoFields[1], sp.HopFields[1], nil)
> 	sp.HopFields[2].Mac = path.MAC(mac, sp.InfoFields[1], sp.HopFields[2], nil)
> ```

Same comment as in the unit test; computing the MAC with the same key (`mac` object contains key) may be misleading. If we want to set up the different MACs for illustration, we should use separate keys for the different ASes.
___
*[`tools/braccept/cases/child_to_child_peering_out.go` line 1 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8U-RYCa3Lmf-0gcka:-Ne8U-RYCa3Lmf-0gckb:btcovqk) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_out.go#L1)):*
> ```Go
> ```

// Copyright 2023 SCION Association
___
*[`tools/braccept/cases/child_to_child_peering_out.go` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8U24Y3BfSa-wQRL9s:-Ne8U24Y3BfSa-wQRL9t:b-uw5um) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_out.go#L41)):*
> ```Go
> // The peering hop is the second hop on the first segment: it crosses from a child
> // interface to a peering interface.
> func ChildToChildPeeringOut(artifactsDir string, mac hash.Hash) runner.Case {
> ```

`ChildToPeer`

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: all files reviewed, 8 unresolved discussions (waiting on @matzf)

___
*[`control/beaconing/extender.go` line 106 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne7m9X6-gIZmjD6c1xB:-NeChhiWB3AbuwRLnBkw:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/control/beaconing/extender.go#L106)):*

+1, reusing the `extractBeta(pseg)` value would make this a lot clearer.
</blockquote></details>

done
___
*[`router/dataplane.go` line 1035 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne7sXfBErW5-8EWXEM7:-NeChkzKE34sG3k54fol:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane.go#L1035)):*

Nit: i find the term "peering AS" to be a bit inaccurate, I'd prefer "peering hop field".  E.g. "peering indicates that the current hop is a peering hop field"
</blockquote></details>

done
___
*[`router/dataplane.go` line 1367 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8EPgTEFaySxovNF3t:-NeChyp625gY1Qb2qhO5:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane.go#L1367)):*

Nit. HOP -> hop (looks like an acronym otherwise)
</blockquote></details>

done
___
*[`router/dataplane_test.go` line 727 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8HaYKB9eZANAdw0Du:-NeCkbAh0xw8KemKORAy:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane_test.go#L727)):*

Nit. Just to make sure there is no confusion here, the MAC for hop field 2 would in practice be computed with the secret key of a different AS. It might be good to illustrate this, by just using a different key.
</blockquote></details>

done
___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 1 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8TJLU8P668mYOqohh:-NeCwp8K6WSFv3_cuMzI:bb7f2mf) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L1)):*

// Copyright 2023 SCION Association
</blockquote></details>

done
___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8TPuT4i1CJaxgoZEg:-NeCwkpm0m0-_axqERCB:bb7f2mf) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L41)):*

Following the naming scheme for the other braccept cases, this would be `PeerToChild`. The name does not describe the global path, but the hop that is processed at the router.
</blockquote></details>

done
___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 47 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8V7Dp-FmVEWPWI12y:-NeCwj6t9mrtdMnI9871:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L47)):*

Nit. I believe we've usually abbreviated interface as "IF", not "I/F".
</blockquote></details>

done
___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 107 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8UOUI19C1INoT7EFA:-NeCwsxbBviAsMTOzXUQ:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L107)):*

Same comment as in the unit test; computing the MAC with the same key (`mac` object contains key) may be misleading. If we want to set up the different MACs for illustration, we should use separate keys for the different ASes.
</blockquote></details>

done
___
*[`tools/braccept/cases/child_to_child_peering_out.go` line 1 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8U-RYCa3Lmf-0gcka:-NeCwwgW0hs1KczxCVE0:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_out.go#L1)):*

// Copyright 2023 SCION Association
</blockquote></details>

done
___
*[`tools/braccept/cases/child_to_child_peering_out.go` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8U24Y3BfSa-wQRL9s:-NeCwxo80FpU_d7MzyNB:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_out.go#L41)):*

`ChildToPeer`
</blockquote></details>

done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 11 of 18 files reviewed, 8 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 1291 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne7pTY46kUi0j5q-vW9:-NeD5Qd_7NCDu2E0Bzzn:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane.go#L1291)):*

The comment above seems to say this that this case is an error, and think I agree. 
I don't think a peering link could ever be a valid egress interface after a cross-over.
</blockquote></details>

done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 11 of 18 files reviewed, 8 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane_test.go` line 741 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8Rkh7BO-yjI3frnQv:-NeD7jf92_3zDqEZrEdp:bxrizu7) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane_test.go#L741)):*

Somewhat aside: it's a bit strange that this test harness does not assert that the egress interface (`result.EgressID`) is correct.
</blockquote></details>

True, but if we change this we need to update every test case to provide the expected egress ID. Should we?

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 11 of 18 files reviewed, 8 unresolved discussions (waiting on @matzf)

___
*[`tools/braccept/main.go` line 132 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8Z8er4WZ5-Mkgma7Y:-NeD81Q72oQro7TV9sbq:b-7gpq3u) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/main.go#L132)):*

Should we perhaps add (a) test(s) for processing the hop(s) before/after a peering hop? 
The processing in these cases is identical to the various ChildToParent / ParentToChild test cases, the only difference is that the path segments have the peer flag set, which the router _should_ effectively ignore.

Another case that might be interesting to cover would be `PeerToInternalHost` / `InternalHostToPeer`, i.e. a path ending/starting in a peering hop.

Note: all of these cases are also covered by the running the "end2end_integration" and scion_integration tests in the default topology, but having an explicit test could still be helpful.
</blockquote></details>

Actually that's the case covered by the braccept test, while the unit test covers the two-routers with 4 hops case. But yeah, the later does not look at what happens downstream/upstream for the same packet. I guess I can add that, or did you mean something else?

Also, I'm not sure how the braccept and the unit tests really differ for the router. They seems to be testing the same thing, albeit with the accept test having to contend with a bunch of layers that are not the focus. So, not sure where adding a test case is more desirable.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 11 of 18 files reviewed, 8 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane_test.go` line 741 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8Rkh7BO-yjI3frnQv:-NeDJNOUEqHnEy06SJd1:b-yp6kfa) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane_test.go#L741)):*

True, but if we change this we need to update every test case to provide the expected egress ID. Should we?
</blockquote></details>

Ah, never mind I just did. And, yes we should. I think it found a couple of bugs...

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 11 of 18 files reviewed, 7 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane_test.go` line 741 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8Rkh7BO-yjI3frnQv:-NeDOsUg5iDxqDsi9frT:b-zg1quv) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/router/dataplane_test.go#L741)):*

Ah, never mind I just did. And, yes we should. I think it found a couple of bugs...
</blockquote></details>

Nah, my test was wrong. Fixed.
and so...
done

matzf (changes requested):
> Reviewed 3 of 7 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: all files reviewed, 3 unresolved discussions (waiting on @jiceatscion)

___
*[`control/beaconing/extender.go` line 288 at r3](https://reviewable.io/reviews/scionproto/scion/4390#-NeD3xFRC8HmFRD7DHDQ:-NeD3xFS3JY7EIFffJpf:b-gzv94s) ([raw file](https://github.com/scionproto/scion/blob/a6f5f61cb9449d98c289ddff42eec6c246db0c4b/control/beaconing/extender.go#L288)):*
> ```Go
> // be added at the end of the segment.
> // FIXME(jice): keeping an accumulator would be just as easy to do as it is during
> // forwarding. What's the benefit of re-calculating the whole chain every time?
> ```

"... to be added to be added ..." 

Regarding the FIXME: first, I think there would be virtually no benefit in passing around an accumulator in the path segment. My gut feeling is that computing this value is just as cheap as having to include it in the protobuf message which would result in (tiny, but still) overheads in encoding, decoding, signature computations and storage.
Also, I think that trusting such a value would weaken the routing security properties in SCION, so we'd have to always check it by computing the XOR. Just now, I can't really put my finger on what the exact security issue would be though, to be honest.
___
*[`tools/braccept/main.go` line 132 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8Z8er4WZ5-Mkgma7Y:-NeDdYkL9Ku_1hFTw3OS:bprh50i) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/main.go#L132)):*
> But yeah, the later does not look at what happens downstream/upstream for the same packet. I guess I can add that, or did you mean something else?

Yes that is exactly what I meant.

The braccept vs unittests is an excellent question, and I don't really know the right answer. I can't see very strong reasons for either choice. It sounds like the unit test would be the more convenient option, so perhaps pick that one?
___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 107 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8UOUI19C1INoT7EFA:-NeDYuhq3Ca4NqkJmh2Q:bqus20i) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L107)):*

done
</blockquote></details>

X and Y are the same keys. Could we make these keys obviously different? 
Btw, I think using `scrypto.InitMac` would be a tiny bit more obvious than the `HFMacFactory` thing.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`tools/braccept/main.go` line 132 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8Z8er4WZ5-Mkgma7Y:-NeDwsvN3H0pJ9019Gim:btd740w) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/main.go#L132)):*

> But yeah, the later does not look at what happens downstream/upstream for the same packet. I guess I can add that, or did you mean something else?

Yes that is exactly what I meant.

The braccept vs unittests is an excellent question, and I don't really know the right answer. I can't see very strong reasons for either choice. It sounds like the unit test would be the more convenient option, so perhaps pick that one?
</blockquote></details>

I agree the unit test is more convenient. You got it!

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: all files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`control/beaconing/extender.go` line 288 at r3](https://reviewable.io/reviews/scionproto/scion/4390#-NeD3xFRC8HmFRD7DHDQ:-NeEA_AE31Qc8mCYgjNP:bpqbd24) ([raw file](https://github.com/scionproto/scion/blob/a6f5f61cb9449d98c289ddff42eec6c246db0c4b/control/beaconing/extender.go#L288)):*

"... to be added to be added ..." 

Regarding the FIXME: first, I think there would be virtually no benefit in passing around an accumulator in the path segment. My gut feeling is that computing this value is just as cheap as having to include it in the protobuf message which would result in (tiny, but still) overheads in encoding, decoding, signature computations and storage.
Also, I think that trusting such a value would weaken the routing security properties in SCION, so we'd have to always check it by computing the XOR. Just now, I can't really put my finger on what the exact security issue would be though, to be honest.
</blockquote></details>

Ah, no. I didn't mean to pass it around during beaconing (although we do seem to find it economical during forwarding - but I also have unspecific security concerns anyway). I just meant to keep it as non-wire meta-data in paths structures so we don't compute it again and again while constructing a path object. Ok, maybe appending more than one hop at the same control service never happens. In that case, I'll remove my comment.

Fixed the echo.

matzf (approved):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

___
*[`control/beaconing/extender.go` line 288 at r3](https://reviewable.io/reviews/scionproto/scion/4390#-NeD3xFRC8HmFRD7DHDQ:-NeEGRWg-NwKRnxk8bV3:bfv61kf) ([raw file](https://github.com/scionproto/scion/blob/a6f5f61cb9449d98c289ddff42eec6c246db0c4b/control/beaconing/extender.go#L288)):*

Ah, no. I didn't mean to pass it around during beaconing (although we do seem to find it economical during forwarding - but I also have unspecific security concerns anyway). I just meant to keep it as non-wire meta-data in paths structures so we don't compute it again and again while constructing a path object. Ok, maybe appending more than one hop at the same control service never happens. In that case, I'll remove my comment.

Fixed the echo.
</blockquote></details>

Ah, I see, sure. Yes, we _do_ extend the same path segments multiple times in the same control service.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 17 of 18 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 107 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8UOUI19C1INoT7EFA:-NeEWP6vDeb4v0LSoGmT:b-ql2r89) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L107)):*

X and Y are the same keys. Could we make these keys obviously different? 
Btw, I think using `scrypto.InitMac` would be a tiny bit more obvious than the `HFMacFactory` thing.
</blockquote></details>

Oops, I was meaning to change one the keys and forgot. Re. InitMac(). Can't.  path.Mac() wants the generator, not the key. (WTF do we need a generator made by a factory? Hopefully we'll never feel the need for a FactoryGenerator)

Anyway.

done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 17 of 18 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`tools/braccept/main.go` line 132 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8Z8er4WZ5-Mkgma7Y:-NeEXLVS9bUKS_yHT1C4:b-7rl0jd) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/main.go#L132)):*

I agree the unit test is more convenient. You got it!
</blockquote></details>

done.

matzf (approved):
> Reviewed 1 of 1 files at r8, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 107 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8UOUI19C1INoT7EFA:-NecAtxfE80_OpifHBxf:bo20bhk) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L107)):*
> Can't. path.Mac() wants the generator, not the key

No, this does not seem right. `path.MAC`  wants a `hash.Hash`, the thing returned from `scrypto.InitMac()`. The `HFMacFactory` returns the generator, `func() hash.Hash`, which we now (need to) invoke in the call to `path.MAC`.
The generator thing is quite useless...

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 16 of 18 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`tools/braccept/cases/child_to_child_peering_in.go` line 107 at r2](https://reviewable.io/reviews/scionproto/scion/4390#-Ne8UOUI19C1INoT7EFA:-NecyYJo8jVbh5FgNZuL:b3ducur) ([raw file](https://github.com/scionproto/scion/blob/3f8391f0e54f9df788c6bd74370b38951e7c5818/tools/braccept/cases/child_to_child_peering_in.go#L107)):*

> Can't. path.Mac() wants the generator, not the key

No, this does not seem right. `path.MAC`  wants a `hash.Hash`, the thing returned from `scrypto.InitMac()`. The `HFMacFactory` returns the generator, `func() hash.Hash`, which we now (need to) invoke in the call to `path.MAC`.
The generator thing is quite useless...
</blockquote></details>

Uh...that's right. I got lost through the redirections. Fixed.

matzf (approved):
> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

___
*[`router/dataplane.go` line 1112 at r10](https://reviewable.io/reviews/scionproto/scion/4390#-NegHAz7CXTkjBJ9WTDG:-NegHAz7CXTkjBJ9WTDH:b-ndv8gj) ([raw file](https://github.com/scionproto/scion/blob/168107f7eab9997454178b919fa2c0c753986e8e/router/dataplane.go#L1112)):*
> ```Go
> 
> 	// TODO: proper error
> 	err := serrors.New("TODO: segment length error (peering)")
> ```

Cross-checking with the review comments in #4299; this was commented on to be fixed.
The default approach in this package is to define the error in the corresponding global `var` block e.g. as

`invalidPeerPath = serrors.New("invalid segment lengths for peering path")`

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 17 of 18 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`router/dataplane.go` line 1112 at r10](https://reviewable.io/reviews/scionproto/scion/4390#-NegHAz7CXTkjBJ9WTDG:-Nem9cPp0Y7sbSC4Lsqx:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/168107f7eab9997454178b919fa2c0c753986e8e/router/dataplane.go#L1112)):*

Cross-checking with the review comments in #4299; this was commented on to be fixed.
The default approach in this package is to define the error in the corresponding global `var` block e.g. as

`invalidPeerPath = serrors.New("invalid segment lengths for peering path")`
</blockquote></details>

Done.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 17 of 18 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`router/dataplane.go` line 1112 at r10](https://reviewable.io/reviews/scionproto/scion/4390#-NegHAz7CXTkjBJ9WTDG:-NemAJJA71GWRO6_BdVU:b-nh97rs) ([raw file](https://github.com/scionproto/scion/blob/168107f7eab9997454178b919fa2c0c753986e8e/router/dataplane.go#L1112)):*

Done.
</blockquote></details>

Uh... at least I thought that's what I'd done. Does this still need more changes? I do not understand what you are referring to by "in the corresponding global var block". There's only one block.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: 17 of 18 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`router/dataplane.go` line 1112 at r10](https://reviewable.io/reviews/scionproto/scion/4390#-NegHAz7CXTkjBJ9WTDG:-NemBKaR6H6fGYAf7xmz:b-uw2iw3) ([raw file](https://github.com/scionproto/scion/blob/168107f7eab9997454178b919fa2c0c753986e8e/router/dataplane.go#L1112)):*

Uh... at least I thought that's what I'd done. Does this still need more changes? I do not understand what you are referring to by "in the corresponding global var block". There's only one block.
</blockquote></details>

Ah, got it. You were commenting on revision 10.
So, done.

matzf (approved):
> Reviewed 1 of 1 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4390)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4392, opened 2023-09-15
scion/ping: wait for send go routine to terminate

oncilla:
Wait for the ping sender go routine to terminate before we return the stats. With bad luck, the receiver go routine receives all the replies before the sender go routine has terminated. Because only the sender go routine increments the stats.Sent count at the end, we can end up with wrong stats that claim more packets were received then sent.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4392)

### DIFF
--- a/scion/ping/ping.go
+++ b/scion/ping/ping.go
@@ -20,6 +20,7 @@ import (
 	"encoding/binary"
 	"math/rand"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/scionproto/scion/pkg/addr"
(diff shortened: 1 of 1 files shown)


matzf (approved):
:lgtm:

Just to slightly expand on the description of the problem (because I had initially misunderstood this to be about spurious echo packets); the race condition is that `WriteTo` in the last `pinger.send` succeeds, but then the writer go-routine is not scheduled to increment the `stats.Sent` counter before these `stats` are returned from `pinger.Ping`. This only seems likely to happen with very low latency (e.g. ping localhost).

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4392#-:-NebxxEFCeq5A624YXrF:bffmach)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oncilla)

PR #4394, opened 2023-09-19
build: prevent go from linking libresolv dynamically.

jiceatscion:
This is default (https://pkg.go.dev/net@go1.21.1#hdr-Name_Resolution). It creates an indirect dependency on a fresh version of glibc. As a result, it prevents our binaries from running on not-so-fesh docker images.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -17,7 +17,7 @@ build --incompatible_default_to_explicit_init_py
 
 # include one of "--define gotags=sqlite_mattn" or "--define gotags=sqlite_modernc"
 # cannot be in common, because query chokes on it.
-build --define gotags=sqlite_modernc
+build --define gotags=sqlite_modernc,no_libresolv
 
 ### options for test
 test --build_tests_only --print_relative_test_log_paths --test_output=errors


oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4394#-NewjJCy7sJfP-9FwIGW:-NewjJCy7sJfP-9FwIGX:b-e64wul) (no related file):*
Is this really required?
I think you can just build against an older glibc version that is supported by your target base image.
E.g., you could use https://github.com/aspect-build/gcc-toolchain which uses glibc 2.26.

Which docker base image are you trying to use?

___

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4394#-NewjJCy7sJfP-9FwIGW:-NewkUSL1ta_oWEwi_P-:bxto8hc) (no related file):*

Is this really required?
I think you can just build against an older glibc version that is supported by your target base image.
E.g., you could use https://github.com/aspect-build/gcc-toolchain which uses glibc 2.26.

Which docker base image are you trying to use?
</blockquote></details>

(I guess the underlying issue is that your locally installed c compiler is using newer glibc version than what is supported)

___

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4394#-NewjJCy7sJfP-9FwIGW:-Newy2kfCN_dT6jm3i9P:b-sjpkja) (no related file):*

(I guess the underlying issue is that your locally installed c compiler is using newer glibc version than what is supported)
</blockquote></details>

This is no more required than switching to sqlite_modernc was. Same reason. Yes, there are other ways to solve the problem: cross build to target older run-times, or build docker images compatible with the build system. The former would make sense if we intended to build for a specific set of platforms. The latter would be more of the right thing to do if we intend to deliver target-agnostic source-code that one can build on and for one's choice's system.

At the moment, I think that target-agnostic is our focus. We (SCION Association) have no near-term plan to include support packages for specific target platforms. So, adding cross-build support right now wouldn't be a great use of my time (but possibly that's something Anapaya would have a vested interest in). 

Making docker images compatible with the build system however isn't immediate either. In the meantime, I, for one, cannot run integration tests anymore. A problem that I can solve for myself and everyone else similarly affected by removing the issue's trigger at basically zero cost to anybody. Any particular reason to *want* libresolv?

___

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4394#-NewjJCy7sJfP-9FwIGW:-NexF22Z4dUH9toit61O:bj7pqak) (no related file):*
> (but possibly that's something Anapaya would have a vested interest in).

Not particularly, we build and distribute the packages in our custom way. But internally, we use the linked aspect project to guarantee hermetic builds and do not rely on the host installed compiler.

> At the moment, I think that target-agnostic is our focus.

I remember OSX having issues with the Go only DNS resolver.
(https://danp.net/posts/macos-dns-change-in-go-1-20/)
Maybe this has been resolved in the meantime (but I'm also not an OSX user...)

___

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4394#-NewjJCy7sJfP-9FwIGW:-NexbpmBEv8xHHOomNLF:bjtj0cd) (no related file):*

> (but possibly that's something Anapaya would have a vested interest in).

Not particularly, we build and distribute the packages in our custom way. But internally, we use the linked aspect project to guarantee hermetic builds and do not rely on the host installed compiler.

> At the moment, I think that target-agnostic is our focus.

I remember OSX having issues with the Go only DNS resolver.
(https://danp.net/posts/macos-dns-change-in-go-1-20/)
Maybe this has been resolved in the meantime (but I'm also not an OSX user...)
</blockquote></details>

That'd be a drag if that had an effect on our own users, but I don't see the use case here. Are there people building SCION for MacOS? If that works, then I guess they won't be using the containers, so indeed my change isn't even useful. Unfortunately there doesn't seem to be a way to put conditional expressions in .bazelrc. However, may be someone building on MacOS could override that line in their own ~/.bazelrc?

___

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4394#-NewjJCy7sJfP-9FwIGW:-NexfmgX2Hn0t6QyfpRa:b-vy5n7z) (no related file):*

That'd be a drag if that had an effect on our own users, but I don't see the use case here. Are there people building SCION for MacOS? If that works, then I guess they won't be using the containers, so indeed my change isn't even useful. Unfortunately there doesn't seem to be a way to put conditional expressions in .bazelrc. However, may be someone building on MacOS could override that line in their own ~/.bazelrc?
</blockquote></details>

There have been in the past. I'm not sure if they are still around. I just brought it up because your mentioning being target agnostic, but AFAIU using Go only supports a subset of the targets 

___

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4394#-NewjJCy7sJfP-9FwIGW:-NexorBn2I596HL2jAj0:b-ip0ei5) (no related file):*

Yeah... I wasn't saying super-portable. It's still mostly Linux, I think.
</blockquote></details>

Good news. I found how one puts platform specific flags in .bazelrc. Problem solved. No MacOS user harmed by this change.

___

matzf (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4394#-NewjJCy7sJfP-9FwIGW:-NfA9tk3Ee2lvoRQDruo:bxmo2rs) (no related file):*

Good news. I found how one puts platform specific flags in .bazelrc. Problem solved. No MacOS user harmed by this change.
</blockquote></details>

This seems ok to me. 

For what it's worth, I don't think DNS is critical in any of the SCION binaries. While some configuration files do ostensibly support host names, these names are only resolved at startup. Functionally, this is mostly equivalent to resolving the names manually at the time of configuration.
I'm not sure the special case for OSX is really necessary, but as you found a very simple solution it doesn't seem to hurt keeping it now.

___

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4394#-NewjJCy7sJfP-9FwIGW:-NfALhz088Vs6cPyYdTm:b-v9i9w9) (no related file):*

This seems ok to me. 

For what it's worth, I don't think DNS is critical in any of the SCION binaries. While some configuration files do ostensibly support host names, these names are only resolved at startup. Functionally, this is mostly equivalent to resolving the names manually at the time of configuration.
I'm not sure the special case for OSX is really necessary, but as you found a very simple solution it doesn't seem to hurt keeping it now.
</blockquote></details>

Thanks! Dominik?

___

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion and @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4394#-NewjJCy7sJfP-9FwIGW:-NfAWsVb8oHxUbotr-d7:bm2lni7) (no related file):*

Thanks! Dominik?
</blockquote></details>

It's not the route I would go, but I also don't have strong opinions.
Resolved and :lgtm: 

___

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

matzf (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4394)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

matzf:
Danger danger: I'm merging despite failing buildkite run; the failing is only due to the updated buildkite runner stack, which will require some small fixes in our pipeline to work (see #4397). As this change here is also required to make the tests run on the new stack, I'm merging this now.

PR #4396, opened 2023-09-22
tools: specify project name in docker-compose files

matzf:
Simplify the usage of the various docker-compose configurations by including the project name in the configuration file. This has been supported for a while now in docker-compose v2. This allows to drop the `-p`/`--project-name` from all `docker-compose` incantations.

Also, streamline the docker-compose files generated by the topogen scripts; remove the explicit `container_name` configurations and drop all the explicit `scion_` prefixes -- managing these prefixes is docker-compose's job. 
Shut up the warnings on "SCION_EXPERIMENTAL_... variable is not set. Defaulting to a blank string." by using the variable expansion syntax to explicitly default to a blank string.

Also, drop the `docker-compose` --compatibility flag. The compatibility flag affects what word separator is used in the
container name and it has long been deprecated. We don't usually rely on specific container names, so this should be
fine. In some exception cases where expecting specific container names seems more practical (containers for bazel-remote-cache and go-module-proxy) due to special casing in the CI scripts, container_name is set explicitly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4396)


jiceatscion (commented):
:lgtm:

> Reviewed 22 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4396#-:-NiFoH8X3hD7iM5fiunF:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

Issue #4215, opened 2022-06-10
E2E failing links test is flaky


matzf:
The E2E failing links test has been flaky since #4168.
The general problem seems to be that the `await-connectivity` script waits for _a_ path to be established between any two ASes, but the "E2E failing links" test requires _specific_ paths to have been discovered before continuing to disable routers.
Consequently, possible fixes could be:
- wait for a fixed amount of time
- extend the `await-connectivity` script to wait for specific path segments
- adapt the `await-connectivity` script to wait for _all_ paths to be established, e.g. by waiting until there are no new segments after one full  beaconing period.

matzf:
Update: I did some investigation (#4356), suspecting that the problem mainly lies with the error and timeout handling end-to-end test utility. However, while I found that these can be improved, this is not the (main) issue here.

The problem is related to the caching of path segments in the control service; once path segments are queried, the result is kept in the cache for a relatively long time (`path.query_interval`, defaults to 5 minutes). The "Refresh" flag of the sciond path requests does not propagate to the path segment requests in the control service.
If the "surviving" segments of the failing link scenario have not been registered before the first path segment request is made, the test will fail.
This suggests a possible alternative, going beyond just fixing the test; rethink the path query caching strategy, to better take into account the interaction between the different caching levels and the usage requirements from SCION applications.

PR #4398, opened 2023-09-26
scion.sh, topogen: adopt docker compose plugin and cleanup

jiceatscion:
Main points of this PR:
Make the build and test environment easier to setup and use:
* Obsolete (and slow) docker-compose V1 replaced with V2:
  * V1 all gone
  * Made scion.sh mstatus robust wrt docker-compose output formatting and make output more intelligible.
  * Work-around issue in some  V2 versions: set some safe $HOME for python tests that use docker-compose.
* Jeager and prometheus now started/stopped in one command and neither is started nor stopped by default. Documentation updated accordingly too.
* Scion.sh commands are start/stop, not run.stop.
* Doc is a bit clearer regarding supervisor vs docker.
* Fixed revocation test to pass -d to end2end-integration when required.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)


matzf (changes requested):
> Reviewed 1 of 20 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 1 of 26 files reviewed, 9 unresolved discussions (waiting on @jiceatscion)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4398#-NfPua9C1hD8E68N8EvV:-NfPua9D5wJG3JFgQfj8:b-bzd23m) (no related file):*
Reading through the docker docs, I could not find out what the `--compatibility` flag really does. Is this just defensive, or do we need this?
And I believe this is related, but I'm not sure: if we are now doing a big pass over all the compose files, would it make sense to change to the compose file version 3?

___
___
*[`-- commits` line 61 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPu7Tf0G9_FULSXrh1):*
FWIW, I don't get that either. The behavior is the same for supervisor and the status subcommand. Personally, I've never liked this; when everything is Ok, the output is empty and that is just not very informative. 
___
*[`scion.sh` line 145 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPpbsfDMXkP0CHBOhx:-NfPpbsfDMXkP0CHBOhy:b-gcel6t) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/scion.sh#L145)):*
> ```Shell
>         rscount=$(./tools/dc scion ps --status=running --format "{{.Name}}" $services | wc -l)
>         tscount=$(echo "$services" | wc -w) # Number of all globed services
> 	./tools/dc scion ps -a \
> ```

This file seems to use space for indention
___
*[`scion.sh` line 279 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPq7gpElTzA1HpFe_R:-NfPq7gpElTzA1HpFe_S:b-rxt3o8) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/scion.sh#L279)):*
> ```Shell
> 	    $PROGRAM help
> 	        Show this text.
> 	    $PROGRAM traces [folder]
> ```

I'm not entirely sure, but isn't `traces`/`stop_traces` now "covered" by the start/stop monitoring subcommands?
___
*[`scion.sh` line 294 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPmpHd9v14NCV1IhNs:-NfPmpHd9v14NCV1IhNt:bv3bbwq) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/scion.sh#L294)):*
> ```Shell
> 
> case "$COMMAND" in
>     help|start|start_monitoring|mstart|mstatus|mstop|stop|stop_monitoring|status|topology|sciond-addr|traces|stop_traces|topo_clean|topodot|bazel_remote)
> ```

Can we please use kebab-case for cli subcommands and flags?
___
*[`acceptance/cert_renewal/test.py` line 62 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPrduS5WNptSWbNgIO:-NfPrduT-0Q8l10rmy-Q:bhr3gvm) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/acceptance/cert_renewal/test.py#L62)):*
> ```Python
> 
>         # Services in their containers need time to start.
>         time.sleep(5)
> ```

I hate these sleeps. IMO, the right way to go about this is to explicitly wait for the required condition. It's usually _much_ quicker, more reliable and explicitly documents what needs to happen to continue.
In many cases (e.g. here), a sensible condition is that the path discovery process has run for enough time so that at least one path between any two ASes have been found. This is implemented in `tools/await-connectivity`, but so far none of the acceptance tests have been modifed to use this. Note that the bazel harness requires to jump through some hoops to use that script. So, unless the flakyness is so bad that we cannot continue without this, I'd prefer to keep this flaky and fix it properly in a separate PR.
___
*[`tools/dc` line 40 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPqYte89tM6Jq1PMoB:-NfPqYte89tM6Jq1PMoC:b-1q9p25) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/tools/dc#L40)):*
> ```
> 	    - [GROUP]:
> 	        - scion:    For all scion services.
> 	        - monitoring:     For the prometheus service.
> ```

Nit: "For the monitoring services, i.e. prometheus and jaeger."
___
*[`tools/integration/revocation_test.sh` line 41 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPnORG2nP86mEml2ho:-NfPnORG2nP86mEml2hp:b-15l03b) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/tools/integration/revocation_test.sh#L41)):*
> ```Shell
> # Was there a time when end2end_integration would start the
> # scion service containers?
> ./scion.sh mstart $REV_BRS
> ```

No this is the point of this test, that the routers are down :smile:

The test is slightly flaky though, see #4215.
___
*[`tools/topology/monitoring.py` line 162 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPnnxmDT6screMlLrp:-NfPnnxmDT6screMlLrq:bb76mu4) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/tools/topology/monitoring.py#L162)):*
> ```Python
>     def _write_dc_file(self):
>         # Merged yeager and prometheus files.
>         # FIXME: should be a single container
> ```

Why?

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 1 of 26 files reviewed, 7 unresolved discussions (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4398#-NfPua9C1hD8E68N8EvV:-NfPyL6IDF_MSQpTUjwP:b9vv5ox) (no related file):*

Reading through the docker docs, I could not find out what the `--compatibility` flag really does. Is this just defensive, or do we need this?
And I believe this is related, but I'm not sure: if we are now doing a big pass over all the compose files, would it make sense to change to the compose file version 3?
</blockquote></details>

Yeah, it is mostly defensive. Some doc says that this is needed to preserve some container naming compatibility (_ vs - something like that). It's included in the shim script that has replaced /usr/bin/docker-compose. Eventually we'll want to figure out the naming business and adjust if needed.

___
___
*[`-- commits` line 61 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPu7Tf0G9_FULSXrh1):*

FWIW, I don't get that either. The behavior is the same for supervisor and the status subcommand. Personally, I've never liked this; when everything is Ok, the output is empty and that is just not very informative.
</blockquote></details>

Any concern about fixing it? May be outputing to stderr instead of stdout, too. To protect invokers that expect silence when stuff works.
___
*[`scion.sh` line 145 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPpbsfDMXkP0CHBOhx:-NfPzod4AsTO8aDPQIua:bb7f2mf) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/scion.sh#L145)):*

This file seems to use space for indention
</blockquote></details>

done
___
*[`scion.sh` line 279 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPq7gpElTzA1HpFe_R:-NfPzsjo4CjaSNsnMMGj:b-jbtwt2) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/scion.sh#L279)):*

I'm not entirely sure, but isn't `traces`/`stop_traces` now "covered" by the start/stop monitoring subcommands?
</blockquote></details>

I think not. This seems to start/stop jaeger in some ad-hoc manner to retrieve traces. It existed independently of run_yaeger/stop_yeager which was called automatically by run_scion/stop_scion. So, I figured it was orthogonal business and left it alone. However, its usefullness probably depends  yaeger having been started  when starting scion. May be some added doc would be needed. I haven't seen any mention of the start_traces in the doc.
___
*[`scion.sh` line 294 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPmpHd9v14NCV1IhNs:-NfQ6fKC8TatBT88hNis:b-u7bqy4) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/scion.sh#L294)):*

Can we please use kebab-case for cli subcommands and flags?
</blockquote></details>

skewered everything.
done
___
*[`acceptance/cert_renewal/test.py` line 62 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPrduS5WNptSWbNgIO:-NfQ6v2o8q_2tflBW2Sj:bi1jhrg) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/acceptance/cert_renewal/test.py#L62)):*

I hate these sleeps. IMO, the right way to go about this is to explicitly wait for the required condition. It's usually _much_ quicker, more reliable and explicitly documents what needs to happen to continue.
In many cases (e.g. here), a sensible condition is that the path discovery process has run for enough time so that at least one path between any two ASes have been found. This is implemented in `tools/await-connectivity`, but so far none of the acceptance tests have been modifed to use this. Note that the bazel harness requires to jump through some hoops to use that script. So, unless the flakyness is so bad that we cannot continue without this, I'd prefer to keep this flaky and fix it properly in a separate PR.
</blockquote></details>

The flakyness is really bad. At least one test had a near 100% failure rate for me. If you file a bug about these added delays and assign to me, I'll try and fix it. In the meantime we do need a work-around.
___
*[`tools/dc` line 40 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPqYte89tM6Jq1PMoB:-NfQ7q4yDpBPVZA9lgIQ:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/tools/dc#L40)):*

Nit: "For the monitoring services, i.e. prometheus and jaeger."
</blockquote></details>

done
___
*[`tools/integration/revocation_test.sh` line 41 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPnORG2nP86mEml2ho:-NfQEAaCAawJX8n-_RZv:bhszy49) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/tools/integration/revocation_test.sh#L41)):*

No this is the point of this test, that the routers are down :smile:

The test is slightly flaky though, see #4215.
</blockquote></details>

Oops. Had not understood the test design. Removed that.
___
*[`tools/topology/monitoring.py` line 162 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPnnxmDT6screMlLrp:-NfQEOJA-IYRbgP1ai6X:bnx3sm0) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/tools/topology/monitoring.py#L162)):*

Why?
</blockquote></details>

Uh...why not. Just a small resource saving. I'm happy to give up on that though.
done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 0 of 32 files reviewed, 7 unresolved discussions (waiting on @matzf)

___
*[`acceptance/cert_renewal/test.py` line 62 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPrduS5WNptSWbNgIO:-NfRhVwV-Zm-Gg54m8OW:b-artseo) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/acceptance/cert_renewal/test.py#L62)):*

The flakyness is really bad. At least one test had a near 100% failure rate for me. If you file a bug about these added delays and assign to me, I'll try and fix it. In the meantime we do need a work-around.
</blockquote></details>

Discussion now moot. Issue fixed better in truck. My fix removed.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 0 of 32 files reviewed, 7 unresolved discussions (waiting on @matzf)

___
*[`-- commits` line 61 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPu7Tf0G9_FULSXrh1):*

Any concern about fixing it? May be outputing to stderr instead of stdout, too. To protect invokers that expect silence when stuff works.
</blockquote></details>

Any objection to fixing that way?

matzf (changes requested):
> Reviewed 5 of 9 files at r1, 8 of 20 files at r2, 1 of 3 files at r3, 3 of 8 files at r4, 2 of 6 files at r6, 8 of 8 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 28 of 32 files reviewed, 7 unresolved discussions (waiting on @jiceatscion)

___
*[`-- commits` line 61 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPu7Tf0G9_FULSXrh1):*

Any objection to fixing that way?
</blockquote></details>

This seems like a fine idea. If you want to fix this, I'd perhaps go even further and simplify the status/mstatus commands to just print the (filtered) output of supervisor or docker compose.
I'm not aware of anything that uses the output of "./scion.sh [m]status" in a script. The only usage in this repo is in the revocation test, but that one only looks at exit status, as a sanity check to verify that a service exists at all.
___
*[`scion.sh` line 279 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPq7gpElTzA1HpFe_R:-NfUsRSC4dG1IfL_C_Db:b-4donsa) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/scion.sh#L279)):*

I think not. This seems to start/stop jaeger in some ad-hoc manner to retrieve traces. It existed independently of run_yaeger/stop_yeager which was called automatically by run_scion/stop_scion. So, I figured it was orthogonal business and left it alone. However, its usefullness probably depends  yaeger having been started  when starting scion. May be some added doc would be needed. I haven't seen any mention of the start_traces in the doc.
</blockquote></details>

Right, I do remember now -- I think the idea was that the `traces` subcommand helps to look at traces that were collected somewhere else.
The difference to the jaeger started in `start-monitoring` is that this one (the `traces` one) does _not_ open the port to allow collecting new traces.
I don't think this really belongs here, it would make more sense as a separate script eg. in tools/, or get rid of it entirely, idk.
At minimum, I'd suggest to add a short sentence here in the documentation to clarify that this is intended to be used to look at the traces gathered elsewhere. To look at the traces collected locally, just open `http://localhost:16868` while `monitoring` is up.
___
*[`scion.sh` line 294 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPmpHd9v14NCV1IhNs:-NfUlBXN-36fg9NlC7MT:butdm0a) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/scion.sh#L294)):*

skewered everything.
done
</blockquote></details>

Excellent, bon appetit!
___
*[`scion.sh` line 50 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfUzOcdEI0of8-ng1xY:-NfUzOceAb46mx3IgZkB:b-p3gy1k) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/scion.sh#L50)):*
> ```Shell
>     start_scion
>     echo "Note that jaeger is not included anymore."
>     echo "To run jaeger and prometheus, use run_monitoring."
> ```

"To run jaeger and prometheus, use '$PROGRAM start-monitoring'."

Also nit: "not started automatically" instead of "not included"? Otherwise it makes it sound like we've removed it entirely.
___
*[`scion.sh` line 66 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfV05cB19UNd8UVOISn:-NfV05cB19UNd8UVOISo:bw9kgjd) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/scion.sh#L66)):*
> ```Shell
>     fi
>     echo "Running monitoring..."
>     ./tools/quiet ./tools/dc monitoring up -d
> ```

Maybe this could echo the URLs for the local web-UIs of prometheus and jaeger, so people can actually find this?
```
Running monitoring...
Jaeger UI: http://localhost:16686
Prometheus UI: http://locahlhost:<port?>
```
___
*[`scion.sh` line 122 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfV-TtI4L2fSLgdNHWP:-NfV-TtI4L2fSLgdNHWQ:b-x9kqsi) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/scion.sh#L122)):*
> ```Shell
>     stop_scion
>     echo "Note that jaeger is not included anymore."
>     echo "To stop jaeger and prometheus, use stop-monitoring."
> ```

Ditto, same as cmd_start.
___
*[`doc/dev/run.rst` line 19 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfUmbB22Pn65CCSH8wR:-NfUmbB22Pn65CCSH8wS:b-lnd7jf) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/doc/dev/run.rst#L19)):*
> ```reStructuredText
> - `docker compose <https://docs.docker.com/compose/>`_. Runs individual processes in separate containers connected with docker network bridges. Only this mode supports running a "SCION-IP gateway".
> 
> Before attempting to use the docker compose mode, be sure to build the necessary docker images with:
> ```

Nit: maybe indent this so it is part of the `docker compose` bullet point. Could make it a "hint" admonition

```rst
- `docker compose <...>`_. blabla

   .. hint:: Before attempting blablabl run `make docker-images`
```

(I'm suggesting to use inline code style instead of a code-block, because the  admonition looks oddly big otherwise).
___
*[`doc/dev/run.rst` line 91 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfUm9YM1OfXjtsi7Puu:-NfUm9YM1OfXjtsi7Puv:bdjc5p6) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/doc/dev/run.rst#L91)):*
> ```reStructuredText
> corresponding :doc:`scion daemon </manuals/daemon>` instance.
> 
> If  :option:`scion.sh topology -d` command is used, additional configuration files are created to
> ```

Nit: I think "additional" is not completely accurate. It's more like selecting one of two variants (docker vs supervisor), both of which generate somewhat different things. I'd just leave the word "additional" away.
___
*[`doc/dev/run.rst` line 198 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfUoppN2PImg88h3dGo:-NfUoppN2PImg88h3dGp:b-bf0wo5) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/doc/dev/run.rst#L198)):*
> ```reStructuredText
> 
>    Stop the monitoring services.
> 	    
> ```

Nit. Trailing whitespace

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 28 of 32 files reviewed, 8 unresolved discussions (waiting on @jiceatscion)

___
*[`doc/dev/run.rst` line 93 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfVAhHl5_QclU7ZUW9i:-NfVAhHl5_QclU7ZUW9j:b-ezq73k) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/doc/dev/run.rst#L93)):*
> ```reStructuredText
> If  :option:`scion.sh topology -d` command is used, additional configuration files are created to
> enable running the SCION services in docker containers (see `docker`_). Otherwise,  a configuration
> file is created to enable running the SCION services as plain processes (see `supervisor`_)
> ```

Oops, sphinx creates a link to `supervisor.org`!?
Add labels and use `:ref:` instead to ensure this stuff points to where we want (see https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref)

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 28 of 32 files reviewed, 4 unresolved discussions (waiting on @matzf)

___
*[`-- commits` line 61 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPu7Tf0G9_FULSXrh1):*

This seems like a fine idea. If you want to fix this, I'd perhaps go even further and simplify the status/mstatus commands to just print the (filtered) output of supervisor or docker compose.
I'm not aware of anything that uses the output of "./scion.sh [m]status" in a script. The only usage in this repo is in the revocation test, but that one only looks at exit status, as a sanity check to verify that a service exists at all.
</blockquote></details>

Ok, then I keep the output on stdout. Doing a formated ps of everything in scope, running or otherwise. Made the output for the docker case similar to that of the supervisor case. I cannot make them strictly equivalent though. There's nothing relevant I can output for a pid and the names aren't structured the same way.
___
*[`scion.sh` line 279 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPq7gpElTzA1HpFe_R:-NfVS33G2KhsH53hAMd8:bril6jh) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/scion.sh#L279)):*

Right, I do remember now -- I think the idea was that the `traces` subcommand helps to look at traces that were collected somewhere else.
The difference to the jaeger started in `start-monitoring` is that this one (the `traces` one) does _not_ open the port to allow collecting new traces.
I don't think this really belongs here, it would make more sense as a separate script eg. in tools/, or get rid of it entirely, idk.
At minimum, I'd suggest to add a short sentence here in the documentation to clarify that this is intended to be used to look at the traces gathered elsewhere. To look at the traces collected locally, just open `http://localhost:16868` while `monitoring` is up.
</blockquote></details>

Now that I paid attention to it, I agree. It doesn't really belong there. Moved it to it's own tool.
done
___
*[`scion.sh` line 50 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfUzOcdEI0of8-ng1xY:-NfVKJxv3YeA-dEsK_zf:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/scion.sh#L50)):*

"To run jaeger and prometheus, use '$PROGRAM start-monitoring'."

Also nit: "not started automatically" instead of "not included"? Otherwise it makes it sound like we've removed it entirely.
</blockquote></details>

done
___
*[`scion.sh` line 66 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfV05cB19UNd8UVOISn:-NfVKjB98CrwtCYrtE00:bb7f2mf) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/scion.sh#L66)):*

Maybe this could echo the URLs for the local web-UIs of prometheus and jaeger, so people can actually find this?
```
Running monitoring...
Jaeger UI: http://localhost:16686
Prometheus UI: http://locahlhost:<port?>
```
</blockquote></details>

done
___
*[`scion.sh` line 122 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfV-TtI4L2fSLgdNHWP:-NfVKkW15Hwo4qCo1-CM:bb7f2mf) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/scion.sh#L122)):*

Ditto, same as cmd_start.
</blockquote></details>

done
___
*[`doc/dev/run.rst` line 19 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfUmbB22Pn65CCSH8wR:-NfVT5IA8ZAQZbvn6AlC:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/doc/dev/run.rst#L19)):*

Nit: maybe indent this so it is part of the `docker compose` bullet point. Could make it a "hint" admonition

```rst
- `docker compose <...>`_. blabla

   .. hint:: Before attempting blablabl run `make docker-images`
```

(I'm suggesting to use inline code style instead of a code-block, because the  admonition looks oddly big otherwise).
</blockquote></details>

done
___
*[`doc/dev/run.rst` line 91 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfUm9YM1OfXjtsi7Puu:-NfVTQZb8XSuyKMV5agd:b-aakknf) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/doc/dev/run.rst#L91)):*

Nit: I think "additional" is not completely accurate. It's more like selecting one of two variants (docker vs supervisor), both of which generate somewhat different things. I'd just leave the word "additional" away.
</blockquote></details>

Yeah. I hesitated on that one. I wanted to convey that there are additional files (different ones) depending on the docker/supervisor option, but if I used "additional" twice it became even more confusing. Giving up. Better without it.
___
*[`doc/dev/run.rst` line 198 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfUoppN2PImg88h3dGo:-NfVWn8iD2UX9V8eXd51:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/doc/dev/run.rst#L198)):*

Nit. Trailing whitespace
</blockquote></details>

done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 27 of 33 files reviewed, 4 unresolved discussions (waiting on @matzf)

___
*[`doc/dev/run.rst` line 93 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfVAhHl5_QclU7ZUW9i:-NfVX5DUDJyd5_Ek6wSR:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/doc/dev/run.rst#L93)):*

Oops, sphinx creates a link to `supervisor.org`!?
Add labels and use `:ref:` instead to ensure this stuff points to where we want (see https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref)
</blockquote></details>

done

matzf (changes requested):
> Reviewed 1 of 8 files at r4, 3 of 4 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 31 of 33 files reviewed, 2 unresolved discussions (waiting on @jiceatscion)

___
*[`scion.sh` line 279 at r3](https://reviewable.io/reviews/scionproto/scion/4398#-NfPq7gpElTzA1HpFe_R:-Nfjm9Jd1IJ578GIvltE:b87jjfs) ([raw file](https://github.com/scionproto/scion/blob/bd3bb9baf645da7d486c66b084ff08ecb419b9a3/scion.sh#L279)):*

Now that I paid attention to it, I agree. It doesn't really belong there. Moved it to it's own tool.
done
</blockquote></details>

Ah sweet, thanks!
___
*[`doc/dev/run.rst` line 93 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfVAhHl5_QclU7ZUW9i:-NfjmXPkARRVRQ9hagEM:b-jq37n7) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/doc/dev/run.rst#L93)):*

done
</blockquote></details>

The trailing underscore is a different syntax, it should be removed now (`doc/dev/run.rst:87: WARNING: Mismatch: both interpreted text role prefix and reference suffix.`)

```
:ref:`docker-section`

NOT

:ref:`docker-section`_
```

Note: you can see the rendered page and the build logs linked from the PR; in the "Checks" section with the statuses from the different integrated systems click on the "Details" link for read the docs.
Note 2: I need to review the settings for read the docs builds; these warnings should _fail_ the build.
___
*[`doc/dev/run.rst` line 93 at r11](https://reviewable.io/reviews/scionproto/scion/4398#-NfjnEgaCHwLWudsT4Xz:-NfjnEgaCHwLWudsT4Y-:b-6rmcl9) ([raw file](https://github.com/scionproto/scion/blob/0bfc34efe6eb71a8f7f7ac9c6bc59459093e9d24/doc/dev/run.rst#L93)):*
> ```reStructuredText
> 
> .. _supervisor-section:
> supervisor
> ```

Needs a blank line between the label definition and the section. (`doc/dev/run.rst:93: WARNING: Explicit markup ends without a blank line; unexpected unindent`)
Nit: we've

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 31 of 33 files reviewed, 3 unresolved discussions (waiting on @jiceatscion)

___
*[`tools/integration/revocation_test.sh` line 22 at r11](https://reviewable.io/reviews/scionproto/scion/4398#-NfjpEs30ya7E9fXqh8p:-NfjpEs40nxfu3n5y3fD:b-otds2p) ([raw file](https://github.com/scionproto/scion/blob/0bfc34efe6eb71a8f7f7ac9c6bc59459093e9d24/tools/integration/revocation_test.sh#L22)):*
> ```Shell
> 
> for br in $REV_BRS; do
>     if ! ./scion.sh mstatus "$br"; then
> ```

This currently fails, not sure why. Is the exit code different now for some reason?

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 30 of 33 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`doc/dev/run.rst` line 93 at r9](https://reviewable.io/reviews/scionproto/scion/4398#-NfVAhHl5_QclU7ZUW9i:-NfjsU3T-tBECMnEHYoE:b-6xgs8j) ([raw file](https://github.com/scionproto/scion/blob/927085b78fe54d895323b982f57edaa8687212da/doc/dev/run.rst#L93)):*

The trailing underscore is a different syntax, it should be removed now (`doc/dev/run.rst:87: WARNING: Mismatch: both interpreted text role prefix and reference suffix.`)

```
:ref:`docker-section`

NOT

:ref:`docker-section`_
```

Note: you can see the rendered page and the build logs linked from the PR; in the "Checks" section with the statuses from the different integrated systems click on the "Details" link for read the docs.
Note 2: I need to review the settings for read the docs builds; these warnings should _fail_ the build.
</blockquote></details>

oops, forgot to remove it.
done.
___
*[`doc/dev/run.rst` line 93 at r11](https://reviewable.io/reviews/scionproto/scion/4398#-NfjnEgaCHwLWudsT4Xz:-Nfjsf1r2T_ggj6lwpr9:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/0bfc34efe6eb71a8f7f7ac9c6bc59459093e9d24/doc/dev/run.rst#L93)):*

Needs a blank line between the label definition and the section. (`doc/dev/run.rst:93: WARNING: Explicit markup ends without a blank line; unexpected unindent`)
Nit: we've
</blockquote></details>

done
___
*[`tools/integration/revocation_test.sh` line 22 at r11](https://reviewable.io/reviews/scionproto/scion/4398#-NfjpEs30ya7E9fXqh8p:-Nfjsh9t4xjxB-g6THpE:bduscio) ([raw file](https://github.com/scionproto/scion/blob/0bfc34efe6eb71a8f7f7ac9c6bc59459093e9d24/tools/integration/revocation_test.sh#L22)):*

This currently fails, not sure why. Is the exit code different now for some reason?
</blockquote></details>

not sure either...works for me. Digging.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 30 of 33 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`tools/integration/revocation_test.sh` line 22 at r11](https://reviewable.io/reviews/scionproto/scion/4398#-NfjpEs30ya7E9fXqh8p:-NfjuPwk6gbq6ztNgV0r:b-rz3l2p) ([raw file](https://github.com/scionproto/scion/blob/0bfc34efe6eb71a8f7f7ac9c6bc59459093e9d24/tools/integration/revocation_test.sh#L22)):*

not sure either...works for me. Digging.
</blockquote></details>

Ah...it fails only when using supervisor, and not with docker... stay tuned.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 29 of 33 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`tools/integration/revocation_test.sh` line 22 at r11](https://reviewable.io/reviews/scionproto/scion/4398#-NfjpEs30ya7E9fXqh8p:-Nfk6QVB8_7v4Ms0eZ4b:b-rnraql) ([raw file](https://github.com/scionproto/scion/blob/0bfc34efe6eb71a8f7f7ac9c6bc59459093e9d24/tools/integration/revocation_test.sh#L22)):*

Ah...it fails only when using supervisor, and not with docker... stay tuned.
</blockquote></details>

done.

matzf (commented):
> Reviewed 1 of 1 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: 31 of 33 files reviewed, all discussions resolved

matzf (approved):
> Reviewed 1 of 1 files at r13, 5 of 5 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

matzf (approved):
> Reviewed 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4398)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4399, opened 2023-09-28
acceptance: de-flake cert_renewal and hidden_paths

### LINKED ISSUE #4301: ci: AT hidden_paths is flaky
The hidden_paths integration test is flaky on the CI infrastructure. As far as I remember, this had not been flaky on the old buildkite AWS stack, only on the new infrastructure. Possibly it's just a small thing like adjusting a timeout value.

matzf:
These tests have become even more flaky with the updated CI runners; it seems that this is triggered due to using the significantly quicker docker-compose v2.

Use the existing await-connectivity script to determine when a test can safely start to run, instead of sleeping for arbitrary number of seconds.

Closes #4301

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4399)

### DIFF
--- a/acceptance/cert_renewal/test.py
+++ b/acceptance/cert_renewal/test.py
@@ -53,6 +53,7 @@ class Test(base.TestTopogen):
     """
 
     def _run(self):
+        self.await_connectivity()
 
         isd_ases = scion.ASList.load("%s/gen/as_list.yml" %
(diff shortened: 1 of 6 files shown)


jiceatscion (approved):

matzf:
Closes #4301

PR #4422, opened 2023-10-16
router: extend packet processing metrics

### LINKED ISSUE #4409: ## Define router performance metrics reproducible in (at least) a given developer env

jiceatscion:
Improve metrics for router packet processing:

* Classify packets by size bucket: log2(size)
* Classify packets by traffic type: incoming, outgoing, as-transit-in/out, br-transit.
* Add process-level on-cpu and sleep time accounting

closes #4409

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)


matzf (changes requested):
> Reviewed 7 of 15 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 7 of 16 files reviewed, 6 unresolved discussions (waiting on @jiceatscion)

___
*[`pkg/private/processmetrics/BUILD.bazel~` line 0 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-Nh0SUhzAWvr1aweNLjK:-Nh0SUi-Du8Cpw5VQu1u:bdxyjtm) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/BUILD.bazel~#L0)):*
Accidentally commited backup file? Same for `processmetrics.go~`. Btw, feel free to add a gitignore pattern for emacs.
___
*[`pkg/private/processmetrics/processmetrics_linux.go` line 43 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-NhAtoDUANKCFNwJC9EZ:-NhAtoDUANKCFNwJC9E_:bl4jhb7) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/processmetrics_linux.go#L43)):*
> ```Go
> 	// runningTime should be the same as uTime+sTime reported in a variety of other ways,
> 	// but when doing calculations, better use the two data from the same source. So, collect them
> 	// both.
> ```

Since go 1.20, the [runtime/metrics](https://pkg.go.dev/runtime/metrics) package has a bunch of CPU metrics. In particular `/cpu/classes/idle:cpu-seconds` and `/cpu/classes/total` appear to be the value that we're after. It appears to be tracked by the go runtime and thus is likely not exactly the same value as the one exposed by the OS scheduler. 
I believe this data _should_ be exported by prometheus as `go_cpu_classes_idle_cpu_seconds_total` and `go_cpu_classes_total_cpu_seconds_total`. For some reason, I couldn't make this happen though yet, I'll have another look. If this works, it seems that we don't even need this custom collector here.
___
*[`pkg/private/processmetrics/processmetrics_linux.go` line 70 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-Nh0ZlqpD3a0XEnk2W02:-Nh0ZlqpD3a0XEnk2W03:bffzox5) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/processmetrics_linux.go#L70)):*
> ```Go
> 	// FIXME (if we can): AllThreads builds a list and it
> 	// ends-up on the garbage pile.
> 	myProcs, err := procfs.AllThreads(c.myPid)
> ```

I think it would be safe to get this list of threads during the initialization and then keep using it. As far as I understand, the go runtime spins up threads during the initialization
___
*[`pkg/private/processmetrics/processmetrics_linux.go` line 80 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-Nh0Wp7HAidFR0bvV6xQ:-Nh0Wp7IEUqGP0mNXHQ-:bcn5mvx) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/processmetrics_linux.go#L80)):*
> ```Go
> 	// and remove only the unupdated ones, but Go makes map values un-addressible. We'd need
> 	// to use pointers instead, which would probably make things worse. TBD.
> 	for pid := range c.lastSchedstats {
> ```

Could use `clear(c.lastSchedstats)` (new in go 1.21).

If garbage is a concern, couldn't we just some over the threads already here, instead of keeping the map and then summing later?
___
*[`pkg/private/processmetrics/processmetrics_linux.go` line 133 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-Nh0V-TcCvy70HzXx6bz:-Nh0V-TcCvy70HzXx6c-:bsz4u3i) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/processmetrics_linux.go#L133)):*
> ```Go
> // It is safe to ignore errors from this but prometheus may lack some
> // metrics.
> func NewProcStatCollector() error {
> ```

I'd expect that `New...` returns something. If this directly registers the collector, perhaps`Init..` or `Register...` would be clearer.
___
*[`pkg/private/processmetrics/processmetrics_linux.go` line 149 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-Nh0Vrw0DlITRD49ZEXg:-Nh0Vrw0DlITRD49ZEXh:bhy8hpo) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/processmetrics_linux.go#L149)):*
> ```Go
> 	err = prometheus.Register(c)
> 	if err != nil {
> 		log.Error("NewProcStatCollector", "registration failed", err)
> ```

IMO, ideally we'd move the logging outside the library in the calling code.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 7 of 16 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`pkg/private/processmetrics/BUILD.bazel~` line  at r4](https://reviewable.io/reviews/scionproto/scion/4422#-Nh0SUhzAWvr1aweNLjK:-NhQeGpyCq1Z3WwqM05v:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/BUILD.bazel~#L)):*

Accidentally commited backup file? Same for `processmetrics.go~`. Btw, feel free to add a gitignore pattern for emacs.
</blockquote></details>

Done.
___
*[`pkg/private/processmetrics/processmetrics_linux.go` line 43 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-NhAtoDUANKCFNwJC9EZ:-NhQfbUUEj_sREF4tlEs:b-6orinw) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/processmetrics_linux.go#L43)):*

Since go 1.20, the [runtime/metrics](https://pkg.go.dev/runtime/metrics) package has a bunch of CPU metrics. In particular `/cpu/classes/idle:cpu-seconds` and `/cpu/classes/total` appear to be the value that we're after. It appears to be tracked by the go runtime and thus is likely not exactly the same value as the one exposed by the OS scheduler. 
I believe this data _should_ be exported by prometheus as `go_cpu_classes_idle_cpu_seconds_total` and `go_cpu_classes_total_cpu_seconds_total`. For some reason, I couldn't make this happen though yet, I'll have another look. If this works, it seems that we don't even need this custom collector here.
</blockquote></details>

/cpu/classes/total:cpu-seconds seemed promising, but it is defined as "GOMAXPROCS integrated over the wall-clock duration this process has been executing for." Which means that includes all periods of "Runnable" state. Where the process is kept off the CPU by preemption. Every existing metric I looked at, other than "schedstats" has that problem. The Go doc is even brutally honest about the disconnection with reality: "Compare only with other /cpu/classes metrics". So, I'd rather use something the meaning of which I'm pretty sure is what I need.
___
*[`pkg/private/processmetrics/processmetrics_linux.go` line 70 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-Nh0ZlqpD3a0XEnk2W02:-NhQiE0UBqZyqT0xzRGp:bem591n) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/processmetrics_linux.go#L70)):*

I think it would be safe to get this list of threads during the initialization and then keep using it. As far as I understand, the go runtime spins up threads during the initialization
</blockquote></details>

I have found differently. The line-up increases in 2 or 3 increments before stabilizing. The number of thread never goes down, though. That is confirmed by several feature requests since 2016, never addressed. As you see in my latest revision. I now refresh the list only when needed. That's 2 or 3 times total.
___
*[`pkg/private/processmetrics/processmetrics_linux.go` line 80 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-Nh0Wp7HAidFR0bvV6xQ:-NhQsLvSBcN_G2gzeHkU:bm18k5y) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/processmetrics_linux.go#L80)):*

Could use `clear(c.lastSchedstats)` (new in go 1.21).

If garbage is a concern, couldn't we just some over the threads already here, instead of keeping the map and then summing later?
</blockquote></details>

You're right. I was trying to stick to a more general pattern of gathering everything and be free to report what we want later. However, the cost greatly exceeds the benefits. Simplified the map out.
done
___
*[`pkg/private/processmetrics/processmetrics_linux.go` line 133 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-Nh0V-TcCvy70HzXx6bz:-NhR1Fkd40kQWexNuq61:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/processmetrics_linux.go#L133)):*

I'd expect that `New...` returns something. If this directly registers the collector, perhaps`Init..` or `Register...` would be clearer.
</blockquote></details>

Done.
___
*[`pkg/private/processmetrics/processmetrics_linux.go` line 149 at r4](https://reviewable.io/reviews/scionproto/scion/4422#-Nh0Vrw0DlITRD49ZEXg:-NhQt-5_8snyAB_X7ebt:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3e72e8116ff3c8bd018e2f7acd73d5518aa4d2ff/pkg/private/processmetrics/processmetrics_linux.go#L149)):*

IMO, ideally we'd move the logging outside the library in the calling code.
</blockquote></details>

Done.

matzf (changes requested):
> Reviewed 1 of 15 files at r2, 6 of 7 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 11 of 17 files reviewed, 4 unresolved discussions (waiting on @jiceatscion)

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 10 of 17 files reviewed, 3 unresolved discussions (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4422#-NhS8wI-40tAuAgi0X5L:-NhXHtbMAY5j__HGRz0E:b-bt6vzh) (no related file):*

As discussed offline, it seems clearer to additionally record the "ingress" interface only for output packets, instead of this crossing label. This makes the label "schemas" different for processed/output packets vs input and dropped packets though (but this does seem to make conceptual sense).
</blockquote></details>

I have refactored accordingly and then some. There's now a fine grain traffic type label for output packets/bytes metrics only. 

___
___
*[`router/dataplane.go` line 986 at r6](https://reviewable.io/reviews/scionproto/scion/4422#-NhRnlts3NI6EO_HSM6Y:-NhXIGts7FJtle71Plw7:b-x8ob4j) ([raw file](https://github.com/scionproto/scion/blob/d7c0b23367f66516f956a50df4667d5b421b7fd2/router/dataplane.go#L986)):*

This seems to miss the "transport" of the corresponding `ingressID` slice, so the accounting will be wrong.

This code was "optimized" for least amount of mental overhead, that's why only the IPv4 messages are tracked here. If we now want to track packet metadata, it's probably much better to reorganize this so that:
- `readUpTo` fills a list of `packet` (instead of `ipv4.Message` and new also `ingressIDs`)
- we only keep track of the `packet`s to be written
- translate `packet` to `ipv4.Message` immediately before the WriteBatch call and dont use the `ipv4.Message` otherwise
</blockquote></details>

I have fixed the bug. I have yet to implement the refactoring you suggest....tomorrow.
___
*[`router/dataplane.go` line 2465 at r6](https://reviewable.io/reviews/scionproto/scion/4422#-NhRb1EFEQ4NxGoteHKq:-NhSAMiB4BqMCnl-yeB-:bg4z0gb) ([raw file](https://github.com/scionproto/scion/blob/d7c0b23367f66516f956a50df4667d5b421b7fd2/router/dataplane.go#L2465)):*

The largest supported packet size is `private/comnmon.SupportedMTU == 9188`, so `maxSizeClass` can be `14`.
</blockquote></details>

If I do that, then I have to keep track of SupportedMTU. Which, unlike the maximum supported by the protocol, could change easily.
___
*[`router/dataplane.go` line 2488 at r6](https://reviewable.io/reviews/scionproto/scion/4422#-NhRbu_hCD_-2Bwv2-Qo:-NhXIYGRDe_Wa39P8FCR:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/d7c0b23367f66516f956a50df4667d5b421b7fd2/router/dataplane.go#L2488)):*

Would it make sense to also introduce a `minSizeClass` or so? Distinguishing the tiny packets probably does not give any benefit, but they make up the bulk of the different size classes. Everything below 128 bytes or so is just "very small".
</blockquote></details>

done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 10 of 17 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 2465 at r6](https://reviewable.io/reviews/scionproto/scion/4422#-NhRb1EFEQ4NxGoteHKq:-NhaAzJG7pU4aT3NE5tP:b-fy2ev2) ([raw file](https://github.com/scionproto/scion/blob/d7c0b23367f66516f956a50df4667d5b421b7fd2/router/dataplane.go#L2465)):*

If I do that, then I have to keep track of SupportedMTU. Which, unlike the maximum supported by the protocol, could change easily.
</blockquote></details>

Ok. I added an init function to assert for the size. maxSizeClass is now 15 (it is the first not-supported size).
Done.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 10 of 17 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 986 at r6](https://reviewable.io/reviews/scionproto/scion/4422#-NhRnlts3NI6EO_HSM6Y:-NhbdDO4D-NSRAVD5lLg:b-y1irp7) ([raw file](https://github.com/scionproto/scion/blob/d7c0b23367f66516f956a50df4667d5b421b7fd2/router/dataplane.go#L986)):*

I have fixed the bug. I have yet to implement the refactoring you suggest....tomorrow.
</blockquote></details>

Done.
I'll let you judge if it's really better than before. As long as we can't pass []Packet to writeBatch, there isn't a whole lot we can do to avoid managing Packets and Messages in parallel (without sacrificing performance).

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 10 of 17 files reviewed, 7 unresolved discussions (waiting on @jiceatscion)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4422#-NhS8wI-40tAuAgi0X5L:-Nhf1tPR3Fp09VMwh0SM:b-8fx2u7) (no related file):*

I have refactored accordingly and then some. There's now a fine grain traffic type label for output packets/bytes metrics only.
</blockquote></details>

:+1: 

___
___
*[`router/dataplane.go` line 986 at r6](https://reviewable.io/reviews/scionproto/scion/4422#-NhRnlts3NI6EO_HSM6Y:-Nhf-XaE27v1Xa4VmsVK:b-rnm1bn) ([raw file](https://github.com/scionproto/scion/blob/d7c0b23367f66516f956a50df4667d5b421b7fd2/router/dataplane.go#L986)):*
I think it would help if we get rid of updating the `writeMsgs` in this loop here and just stupidly "copy" the entire `readPkts` into `writeMsgs` everytime. I doubt this will hurt.

> // Note: we could use a disposable array for that, but we'd have to redo the copies for the packets that we retry.

Also, can we name it just `pkts` or so? "read" being this stupid verb without proper participle makes this a bit confusing...
___
*[`router/dataplane.go` line 2465 at r6](https://reviewable.io/reviews/scionproto/scion/4422#-NhRb1EFEQ4NxGoteHKq:-NheuWtXET0eBrYn9cbb:bxqsw0e) ([raw file](https://github.com/scionproto/scion/blob/d7c0b23367f66516f956a50df4667d5b421b7fd2/router/dataplane.go#L2465)):*

Ok. I added an init function to assert for the size. maxSizeClass is now 15 (it is the first not-supported size).
Done.
</blockquote></details>

Nice. There's a (somewhat cryptic) way to make this a compile time check instead:

```
const _ = uint(1 << (maxSizeClass-1) - bufSize) // compile time assert: ensure that 2 ^ (maxSizeClass-1) >= bufSize
```

Checking at compile time avoids that this shows up only very late in the refactoring process, after someone has already declared success when things compile.
Not blocking, if you think it's too obscure just leave it as is.
___
*[`router/dataplane.go` line 953 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-NhfABTlD--YaEh03t2-:-NhfABTlD--YaEh03t20:b963vf4) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L953)):*
> ```Go
> 		}
> 
> 		// We need to collect stats by traffic type and size class.
> ```

This accounting got a bit verbose now. Would it be possible to move this into a separate function? `accountWrittenPkts(metrics, pkts[:written])` or so?
The call to `returnPacketToPool` should perhaps still be here though for clarity.
___
*[`router/dataplane.go` line 958 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-Nhf7sKk6_epFWTXTi3d:-Nhf7sKk6_epFWTXTi3e:bjt67vh) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L958)):*
> ```Go
> 		writtenPkts := [ttMax][maxSizeClass]int{}
> 		writtenBytes := [ttMax][maxSizeClass]int{}
> 		for i := range writeMsgs[:written] {
> ```

Nit Replace `writeMsgs` here with `readPkts` in the loop.
___
*[`router/dataplane.go` line 1001 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-NhfB3SA5sWPEPHLLSLL:-NhfB3SA5sWPEPHLLSLM:bo8oqmk) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L1001)):*
> ```Go
> 
> func readUpTo(c <-chan packet,
> 	n int, needsBlocking bool, pkts []packet) int {
> ```

Nit: styleguide for multiline function declarations is: "If a function declaration uses more than 1 line, each parameter should be declared on a separate line and the first line of the function body should be empty"

This one probably still fits on a single line though :)
___
*[`router/dataplane.go` line 1078 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-Nhf-CVbEVIBTrwBv6nb:-Nhf-CVbEVIBTrwBv6nc:b-54vyxq) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L1078)):*
> ```Go
> 
> 	pld := p.lastLayer.LayerPayload()
> 	minResult := processResult{}
> ```

No longer useful, please revert to returning `processResult{}`.
___
*[`router/dataplane.go` line 2424 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-Nhf24DzDiLtVxhNhUKa:-Nhf24DzDiLtVxhNhUKb:b-bmr8y5) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L2424)):*
> ```Go
> }
> 
> // interfaceMetrics is the set of metrics that are relevant for one given interface.
> ```

We could move all of metrics stuff (~things below this line) into `metrics.go`. I think the `(*Dataplane).initMetrics` should stay, but for the rest this would seem fairly appropriate, and we'd avoid further growing this gargantuan `dataplane.go` thing. What do you think?
___
*[`router/dataplane.go` line 2523 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-NhezHgy-WoFebdJn5kB:-NhezHgy-WoFebdJn5kC:b-hl1lbx) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L2523)):*
> ```Go
> 		return "inTransit"
> 	case ttOutTransit:
> 		return "outTransit"
> ```

We use `snake_case` for the metric labels and values.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 10 of 17 files reviewed, 4 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 986 at r6](https://reviewable.io/reviews/scionproto/scion/4422#-NhRnlts3NI6EO_HSM6Y:-NhfEQ-z4UtnRJI7GX3y:bmlhkk4) ([raw file](https://github.com/scionproto/scion/blob/d7c0b23367f66516f956a50df4667d5b421b7fd2/router/dataplane.go#L986)):*

I think it would help if we get rid of updating the `writeMsgs` in this loop here and just stupidly "copy" the entire `readPkts` into `writeMsgs` everytime. I doubt this will hurt.

> // Note: we could use a disposable array for that, but we'd have to redo the copies for the packets that we retry.

Also, can we name it just `pkts` or so? "read" being this stupid verb without proper participle makes this a bit confusing...
</blockquote></details>

Done
I came to the same conclusion. Did some more variable renaming.
___
*[`router/dataplane.go` line 2465 at r6](https://reviewable.io/reviews/scionproto/scion/4422#-NhRb1EFEQ4NxGoteHKq:-NhfbTDV8Nu4JtYXLude:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/d7c0b23367f66516f956a50df4667d5b421b7fd2/router/dataplane.go#L2465)):*

Nice. There's a (somewhat cryptic) way to make this a compile time check instead:

```
const _ = uint(1 << (maxSizeClass-1) - bufSize) // compile time assert: ensure that 2 ^ (maxSizeClass-1) >= bufSize
```

Checking at compile time avoids that this shows up only very late in the refactoring process, after someone has already declared success when things compile.
Not blocking, if you think it's too obscure just leave it as is.
</blockquote></details>

Done
___
*[`router/dataplane.go` line 953 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-NhfABTlD--YaEh03t2-:-Nhfih6f4zo81miasJL4:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L953)):*

This accounting got a bit verbose now. Would it be possible to move this into a separate function? `accountWrittenPkts(metrics, pkts[:written])` or so?
The call to `returnPacketToPool` should perhaps still be here though for clarity.
</blockquote></details>

Done
___
*[`router/dataplane.go` line 958 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-Nhf7sKk6_epFWTXTi3d:-NhfDqNGEaRMBSEQDCJU:b-896fjx) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L958)):*

Nit Replace `writeMsgs` here with `readPkts` in the loop.
</blockquote></details>

Done
___
*[`router/dataplane.go` line 1001 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-NhfB3SA5sWPEPHLLSLL:-NhfF0qw6GZFf6Ra41Ls:b-afv3f5) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L1001)):*

Nit: styleguide for multiline function declarations is: "If a function declaration uses more than 1 line, each parameter should be declared on a separate line and the first line of the function body should be empty"

This one probably still fits on a single line though :)
</blockquote></details>

Done
fits one line again now.
___
*[`router/dataplane.go` line 1078 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-Nhf-CVbEVIBTrwBv6nb:-NhfFHus3Dy2D36H7_dv:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L1078)):*

No longer useful, please revert to returning `processResult{}`.
</blockquote></details>

Done
___
*[`router/dataplane.go` line 2424 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-Nhf24DzDiLtVxhNhUKa:-NhfIFahEfE5RhwnClI9:bop324c) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L2424)):*

We could move all of metrics stuff (~things below this line) into `metrics.go`. I think the `(*Dataplane).initMetrics` should stay, but for the rest this would seem fairly appropriate, and we'd avoid further growing this gargantuan `dataplane.go` thing. What do you think?
</blockquote></details>

Done
I agree.
___
*[`router/dataplane.go` line 2523 at r7](https://reviewable.io/reviews/scionproto/scion/4422#-NhezHgy-WoFebdJn5kB:-NhfbpNE9rGYZ1jBOBuA:b-896fjx) ([raw file](https://github.com/scionproto/scion/blob/231c59a069784548d864587bb4703f4d0c3f9f36/router/dataplane.go#L2523)):*

We use `snake_case` for the metric labels and values.
</blockquote></details>

Done

matzf (changes requested):
> Reviewed 1 of 15 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 11 of 17 files reviewed, 10 unresolved discussions (waiting on @jiceatscion)

___
*[`router/dataplane.go` line 626 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfsNx9CUwZdPNL9w_K:-NhfsNxA6SCTPd8EO0DQ:b-optnkm) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/dataplane.go#L626)):*
> ```Go
> 		// For non-broken packets, we defer the ingress-side metrics accounting
> 		// until we have a chance to figure the traffic type.
> 		// That's in runProcessor and processPkt.
> ```

This comment is no longer accurate.
___
*[`router/dataplane.go` line 927 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfqxfH5qLHEYcFF7cJ:-NhfqxfH5qLHEYcFF7cK:b-zb5pph) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/dataplane.go#L927)):*
> ```Go
> 	}
> 	for t := ttOther; t < ttMax; t++ {
> 		for sc := sizeClass(0); sc < maxSizeClass; sc++ {
> ```

`for sc := minSizeClass; ...`
___
*[`router/dataplane.go` line 930 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfrQS_CO2QQP12DVt0:-NhfrQS_CO2QQP12DVt1:b-3tplat) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/dataplane.go#L930)):*
> ```Go
> 			if writtenPkts[t][sc] > 0 {
> 				metrics[sc].Output[t].OutputPacketsTotal.Add(
> 					float64(writtenPkts[t][sc]))
> ```

Nit: fits on one line
___
*[`router/dataplane.go` line 941 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhfqe5PCf2M0UZj83jV:-Nhfqe5PCf2M0UZj83jW:b-1nh43r) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/dataplane.go#L941)):*
> ```Go
> 	cfg *RunConfig, c <-chan packet) {
> 
> 	fmt.Println("Initialize forwarder for", "interface")
> ```

Leftover printf
___
*[`router/dataplane.go` line 959 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfsuyR626aFaWHJg5G:-NhfsuyR626aFaWHJg5H:bas14ob) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/dataplane.go#L959)):*
> ```Go
> 	for d.running {
> 		toWrite += readUpTo(c, cfg.BatchSize-toWrite, toWrite == 0,
> 			pkts[toWrite:])
> ```

Nit: also seems to fit onto a single line now
___
*[`router/export_test.go` line 70 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhflbx2-mw1s2GkdThB:-Nhflbx2-mw1s2GkdThC:b-nks9p4) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/export_test.go#L70)):*
> ```Go
> 
> func InitDPMetrics(dp *DataPlane) {
> 	dp.initMetrics()
> ```

Move this into `NewDP`? It seems we always call this now
___
*[`router/metrics.go` line 220 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhfmf1r1QNIy4OVltc5:-Nhfmf1r1QNIy4OVltc6:b-rocxxk) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/metrics.go#L220)):*
> ```Go
> func (sc sizeClass) String() string {
> 	return strings.Join(
> 		[]string{strconv.Itoa((1 << sc) >> 1), strconv.Itoa(1<<sc - 1)},
> ```

Nit: this is not accurate for `minSizeClass` which would now show as `32_63` (or so.. off by some), instead of `0_63`.
Btw, `Sprintf()` might be simpler here.
___
*[`router/metrics.go` line 227 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfpHpd43I5fAKbwjFi:-NhfpHpd43I5fAKbwjFj:b-1h3vkv) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/metrics.go#L227)):*
> ```Go
> // trafficMetrics groups all the metrics instances that all share the same interface AND
> // sizeClass label values (but have different names - i.e. they count different things).
> type trafficMetrics struct {
> ```

Nit: I think it would help readability if we keep the type definitions of these metrics structs close together:
```
type interfaceMetrics map[sizeClass]trafficMetrics
struct trafficMetrics { ... }
struct outputMetrics { ... }

... size class, traffic type, ...
```
___
*[`router/metrics.go` line 235 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhfovmc1Y6C9NnESeFW:-Nhfovmc1Y6C9NnESeFX:b3himxe) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/metrics.go#L235)):*
> ```Go
> 	DroppedPacketsBusySlowPath  prometheus.Counter
> 	ProcessedPackets            prometheus.Counter
> 	Output                      map[trafficType]outputMetrics
> ```

Not sure if it's worth optimizing this, but this could be an array instead of a map, `Output [ttMax]outputMetrics`.
Same for the size classes, actually, but there it's a bit more awkward because of the min offset.

I _think_ this would allow to turn `outputMetrics` into an anonymous inline struct definition, which potentially makes this easier to navigate.
___
*[`router/metrics.go` line 285 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfnmEn3AE6SZTQtoq-:-NhfnmEn3AE6SZTQtoq0:bxbxqxk) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/metrics.go#L285)):*
> ```Go
> // counters are already instantiated for all the relevant interfaces so this
> // will not have to be repeated during packet forwarding.
> func (d *DataPlane) initMetrics() {
> ```

My preference would be to keep this function together with all the other`Dataplane` methods in the `dataplane.go` file.

If we slightly amend the interface of `initInterfaceMetrics` so that it takes `id uint16, localIA addr.IA, neighbors map[uint16]addr.IA` instead of `labels` and does the call to `interfaceToMetricsLabels` internally, this can be the only function that is exposed from this metrics.go file.

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 11 of 17 files reviewed, 11 unresolved discussions (waiting on @jiceatscion)

___
*[`tools/topology/monitoring.py` line 140 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhfvfzt4tCTMx2mersP:-Nhfvfzu2K_vFYWRFbb1:bk88gvr) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/tools/topology/monitoring.py#L140)):*
> ```Python
>             if relabels is not None:
>                 job_scrape_config['metric_relabel_configs'] = relabels
>             scrape_configs.append(job_scrape_config)
> ```

Left over?

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: 11 of 17 files reviewed, 8 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 626 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfsNx9CUwZdPNL9w_K:-Nhfw6Yr6ScxH0ly2cfC:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/dataplane.go#L626)):*

This comment is no longer accurate.
</blockquote></details>

Done
___
*[`router/dataplane.go` line 927 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfqxfH5qLHEYcFF7cJ:-Nhfw80RCVB2FkqhtcKT:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/dataplane.go#L927)):*

`for sc := minSizeClass; ...`
</blockquote></details>

Done
___
*[`router/dataplane.go` line 930 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfrQS_CO2QQP12DVt0:-Nhg7nuq7ihn9cskGiOI:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/dataplane.go#L930)):*

Nit: fits on one line
</blockquote></details>

Done
___
*[`router/dataplane.go` line 941 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhfqe5PCf2M0UZj83jV:-Nhg7pMd0Ke3nbCXxATm:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/dataplane.go#L941)):*

Leftover printf
</blockquote></details>

Done
___
*[`router/dataplane.go` line 959 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfsuyR626aFaWHJg5G:-Nhg7qZa8G2ovoxIVY43:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/dataplane.go#L959)):*

Nit: also seems to fit onto a single line now
</blockquote></details>

Done
___
*[`router/export_test.go` line 70 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhflbx2-mw1s2GkdThB:-Nhg9voH2Mz5rMqC5Zm6:b-rb06bv) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/export_test.go#L70)):*

Move this into `NewDP`? It seems we always call this now
</blockquote></details>

Done
(Yeah, looks that way, now)
___
*[`router/metrics.go` line 220 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhfmf1r1QNIy4OVltc5:-NhgBVS72qas-XO0zwRu:bfn7ezs) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/metrics.go#L220)):*

Nit: this is not accurate for `minSizeClass` which would now show as `32_63` (or so.. off by some), instead of `0_63`.
Btw, `Sprintf()` might be simpler here.
</blockquote></details>

Done
Ah, right. Here you are. Upper bound also promoted to "inf".
___
*[`router/metrics.go` line 227 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfpHpd43I5fAKbwjFi:-NhgCRnpDHJMtDa6GIkq:bsuclo3) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/metrics.go#L227)):*

Nit: I think it would help readability if we keep the type definitions of these metrics structs close together:
```
type interfaceMetrics map[sizeClass]trafficMetrics
struct trafficMetrics { ... }
struct outputMetrics { ... }

... size class, traffic type, ...
```
</blockquote></details>

K. Sorted differently.
___
*[`router/metrics.go` line 235 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhfovmc1Y6C9NnESeFW:-NhgNC1SCFLmd6CGpbx_:b7eawcp) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/metrics.go#L235)):*

Not sure if it's worth optimizing this, but this could be an array instead of a map, `Output [ttMax]outputMetrics`.
Same for the size classes, actually, but there it's a bit more awkward because of the min offset.

I _think_ this would allow to turn `outputMetrics` into an anonymous inline struct definition, which potentially makes this easier to navigate.
</blockquote></details>

1. Done
2. We'll talk.
___
*[`router/metrics.go` line 285 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-NhfnmEn3AE6SZTQtoq-:-NhgYvx26EumNVfoDeOW:b-zetgum) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/router/metrics.go#L285)):*

My preference would be to keep this function together with all the other`Dataplane` methods in the `dataplane.go` file.

If we slightly amend the interface of `initInterfaceMetrics` so that it takes `id uint16, localIA addr.IA, neighbors map[uint16]addr.IA` instead of `labels` and does the call to `interfaceToMetricsLabels` internally, this can be the only function that is exposed from this metrics.go file.
</blockquote></details>

Done. Good idea. Worked out nicely. I took the opportunity to rename the init(.*)Metrics to new(\1)Metrics since that's what these functions ~are. Unfortunately your wish isn't entirely granted: serviceMetricLabels() is still referenced directly somewhere in dataplane.go
___
*[`tools/topology/monitoring.py` line 140 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhfvfzt4tCTMx2mersP:-NhgchyAE0dqc7qjsjtK:b-qorvf1) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/tools/topology/monitoring.py#L140)):*

Left over?
</blockquote></details>

No. I left it there deliberately as I predict that relabeling will be back soon, just for other reasons. I don't want to throw away support for it.

matzf (approved):
> Reviewed 2 of 4 files at r7, 4 of 4 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

___
*[`tools/topology/monitoring.py` line 140 at r10](https://reviewable.io/reviews/scionproto/scion/4422#-Nhfvfzt4tCTMx2mersP:-Nhhb9_71yKcTb0O593g:b-8fx2u7) ([raw file](https://github.com/scionproto/scion/blob/7d174547fdcb02284dc6a0230050b4a1b34abfb9/tools/topology/monitoring.py#L140)):*

No. I left it there deliberately as I predict that relabeling will be back soon, just for other reasons. I don't want to throw away support for it.
</blockquote></details>

:+1:

matzf (approved):
> Reviewed 3 of 3 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4422)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4432, opened 2023-11-02
deps: update oapi-codegen to v2

matzf:
The oapi-codegen libraries went through a reorganization; the repositories for the oapi-codegen build tool was separated from the runtime libraries.
In our normal code dependencies, the oapi-codegen build tool no longer shows up. Import it in a dummy .go file as an easy way to ensure that we have all the appropriate dependencies (via the `go.mod`/`go_deps.bzl`) . This leaves some clutter in the `go.mod` file which is not aesthetically pleasing, but is not a real issue due to dependency graph pruning.
The alternative solution would have been to manage the dependencies of the `rules_openapi` bazel rules explicitly in a `rules_openapi_dependencies` macro. This turns out to be rather cumbersome, and can lead to unexpected or even invalid dependency version combinations, because `rules_go` does not have a way to separate the dependencies of the build tools from the dependencies of the code.

Aside; fix `cfg = "exec"` for the oapi-codegen buildtool so that this tool is built for the environment of the builder, not the target platform.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4432)


oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4432)* status: 0 of 36 files reviewed, 1 unresolved discussion (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4432#-NiFWwwU-nZSEjTwvFQi:-NiFWwwVEP-K6nUh6CJa:b-nplqsc) (no related file):*
Have you considered adding oapi-codgen to [tools.go](https://github.com/scionproto/scion/blob/master/tools/tools.go)?

I think it would be a nicer solution than manually maintaining the required go dependencies.

___

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4432)* status: 0 of 36 files reviewed, 1 unresolved discussion (waiting on @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4432#-NiFWwwU-nZSEjTwvFQi:-NiFYWSG3xeboija1sVG:b1em7y) (no related file):*

Have you considered adding oapi-codgen to [tools.go](https://github.com/scionproto/scion/blob/master/tools/tools.go)?

I think it would be a nicer solution than manually maintaining the required go dependencies.
</blockquote></details>

Interesting, no I haven't considered this, I'll give it a shot.

Generally I find it awkward that the `go.mod` file is polluted with build tools and all kinds of random garbage, so I wouldn't agree that is a _nicer_ solution, but it certainly is the _easiest_.

I was happy to remove things from `go.mod` (I was only disappointed that it wasn't more). What I would have preferred is to keep the dependencies of the oapi-codegen tool entirely separate from the other `go_deps.bzl` -- it's strange to me that rules_go seems to insist that everything in the workspace must be compiled with the same dependency versions.

___

oncilla (changes requested):
> Reviewed 30 of 36 files at r1, 14 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4432)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4432#-NiFWwwU-nZSEjTwvFQi:-NiFvuel0P8KpryoKX98:b-8gqxbq) (no related file):*

Done.
</blockquote></details>

I think with dependency graph pruning, there is less of an issue with having a lot of dependencies in go.mod file.

alternative is to have a multi-module repo. But that has another can of worms that we don't want to worry about yet IMO.

___

oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4432#-:-NiFwQ1ZD--_hKfGzjOP:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

Issue #4411, opened 2023-10-02
## Add router performance regression tests (reproducible, at least, in the CI context)


matzf:

jiceatscion:
This is in progress too. The difference with the raw performance measurement was slim, so I included them both in the same PR.

Issue #4410, opened 2023-10-02
## Create router benchmark automation, with performance data collection


matzf:

jiceatscion:
This has started. Rough plan:

* Create end2end-style test program that generates a router load (in a specific packet size class):
  * The test generates load of a given packet size and number between every host pair in the available topology.
  * Each kind of load test comes with a custom topology such that exercising all pairs produces a relevant traffic mix.
* Add automation to run the test and Collect and report packets/available-cpu-seconds and bytes/available-cpu-seconds:
  * Run the test over an add-hoc topology where each type of traffic (incoming/outgoing/local/transit) is processed by one identified router.
  * At the end of the test collect metrics for each traffic type and  the relevant size class between two TBD points in time (eg. 1 second after start and 1 second before end), and calculates and logs the average packet/available-cpu-seconds and byte/per-available-cpu-seconds.
* Add acceptance test that runs the above and compares the result with some expectation. Expectations (and the resulting success/failure of the test) would be dependent of the platform where the test is run. So, this may be a CI-only test.
* Cherry on the cake: Extrapolate some approximate abstract, platform independent, value, and have expectations for that, instead of the real numbers.

The first point is done.

PR #4437, opened 2023-11-06
testing: add router benchmark program and use it as integration test

jiceatscion:
Summary of changes:

* Enhance the end2end_integration programs as follows:
  * end2end_integration has a slightly finer-grain role-based pair selection capability.
  * end2end_integration outputs an extra result line that provides precise start/stop timestamps for the whole test.
* process_metrics exposes the number of cores used by Go.
* Add  end2endblast. Similar to end2end but plays a game other than pingpong: packetblast. It keeps logging and tracing to a minimum. The server side responds to very few of the pings (1/256). Just enough to prove that the circuit works.
* Add an integration test called router_benchmark:
  * relies on an especially crafted topology: router_bm.topo
  * uses end2end_integration with command end2endblast over that topology as a load test for the router
  * extracts and logs relevant performance metrics after the test
  * compares performances with a set of expectations if executed by CI.

As a benchmark program, this is only a first approximation. Extending the end2end suite provided a quick solution but isn't ideal: it uses too many cores at once due to requiring 5 routers actively running for one of the tests. More work is in progress to replace this with some leaner setup where only one router is actually running.

Fixes: #4410
Fixes: #4411

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)


matzf (changes requested):
> Reviewed 2 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)* status: 2 of 15 files reviewed, 6 unresolved discussions (waiting on @jiceatscion)

___
*[`.buildkite/pipeline_lib.sh` line 20 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioZmCq-vTrU3qv8sZJ:-NioZmCq-vTrU3qv8sZK:bdwzdev) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/.buildkite/pipeline_lib.sh#L20)):*
> ```Shell
>         if [[ "$test" =~ router_benchmark ]]; then
>             args="--test_arg=--ci"
> 	fi
> ```

There should already be an env variable `CI` which could be used instead of this flag. If I understand the plumbum API correctly, we could even set the flag implicitly based on an envvar using `cli.Flag(...., envname="CI")`.
___
*[`acceptance/router_benchmark/BUILD.bazel` line 8 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioYT4l40T2m_eajAWB:-NioYT4l40T2m_eajAWC:boij2qv) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/acceptance/router_benchmark/BUILD.bazel#L8)):*
> ```Starlark
>     args = [
>         "--executable=end2end_integration:$(location //tools/end2end_integration)",
>     ],
> ```

I'm actually a bit confused how this can even run the `./bin/end2end` as a subcommand binary. How is this even built and how come this is on the appropriate path to run here? Usually bazel is such a stickler about this...
___
*[`acceptance/router_benchmark/test.py` line 3 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-Nio_6LkCeJzOrNlDU3E:-Nio_6LkCeJzOrNlDU3F:b-u8uu7d) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/acceptance/router_benchmark/test.py#L3)):*
> ```Python
> #!/usr/bin/env python3
> 
> # Copyright 2020 Anapaya Systems
> ```

Nit copy pasted copyright
___
*[`pkg/snet/packet_conn.go` line 118 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioHL1Z3BeUFI2mbzSn:-NioHL1Z3BeUFI2mbzSo:bptslzd) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/pkg/snet/packet_conn.go#L118)):*
> ```Go
> 	// FlowID: arbitrary value to distribute connections over router CPUs.
> 	// Initialize to 1 if backward compatible (no-distribution) behavior is desired.
> 	FlowID uint32
> ```

Could we deal with this Flow ID change in a separate PR, so that this gets the appropriate visibility?

IMO, `PacketConn` this is not the right API level to expose this. The `PacketConn` is sort of a "raw" connection. A user of this API can be expected to set the FlowID directly in the `Packet`.

I would add this to the `snet.Conn` instead, setting it when creating a packet in `scionConnWriter.WriteTo`. We can e.g. set this based on a hash of src/dst ISD, AS, Host address and Port (see e.g. https://www.rfc-editor.org/rfc/rfc6437#appendix-A).
I don't think this requires a backwards compatibility layer. From an application perspective, with using a constant FlowID, we would have possibly had an ordering guarantee between packets sent from different `snet.Conn`s. I doubt that this would even have worked in the first place (dispatcher / sending host might not guarantee the send order already). (Ab-)Using this in an application would probably have been tricky enough that I'm sure someone would have noticed that it's a bad idea. 
___
*[`tools/end2end/main.go` line 17 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioQe87DoKbxE55Kz6P:-NioQe87DoKbxE55Kz6Q:b-y1eijb) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/tools/end2end/main.go#L17)):*
> ```Go
> 
> // This is a general purpose client/server code for end2end tests. It plays
> // ping-pong with some variantions depending on command line arguments.
> ```

I'm not sure about generalizing this end2end thing. It feels to me like this makes things much more complicated than they need to be.

Wouldn't a separate "blaster" be much shorter and simpler than this? We only want run one test at a time, and so we don't seem to benefit much from the `end2end_integration` harness. The server side becomes trivial if we don't send any replies and just report received packets out of band.

Note that we could also use the `end2end_integration` harness with a different command than end2end -- it just assumes that some flags are supported -- so this could still be used with a simpler "blaster" test.

I don't know, what do you think?
___
*[`tools/integration/integrationlib/common.go` line 150 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioO_7t9eL7ReI3uDWJ:-NioO_7t9eL7ReI3uDWK:b-qd5ben) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/tools/integration/integrationlib/common.go#L150)):*
> ```Go
> 		}
> 		log.Info("Retrying...")
> 		time.Sleep(integration.RetryTimeout)
> ```

Nit. This is now printing "Retrying..." after the last attempt.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)* status: 2 of 15 files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`.buildkite/pipeline_lib.sh` line 20 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioZmCq-vTrU3qv8sZJ:-NioulmRCOpQmZ3jLq-d:b-y86gpn) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/.buildkite/pipeline_lib.sh#L20)):*

There should already be an env variable `CI` which could be used instead of this flag. If I understand the plumbum API correctly, we could even set the flag implicitly based on an envvar using `cli.Flag(...., envname="CI")`.
</blockquote></details>

Done.
I wasn't aware. Works like a charm.
___
*[`acceptance/router_benchmark/BUILD.bazel` line 8 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioYT4l40T2m_eajAWB:-Niov5iJ85cHyArSrkFx:b-5g242n) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/acceptance/router_benchmark/BUILD.bazel#L8)):*

I'm actually a bit confused how this can even run the `./bin/end2end` as a subcommand binary. How is this even built and how come this is on the appropriate path to run here? Usually bazel is such a stickler about this...
</blockquote></details>

I honestly don't know. I didn't invent this; I cargo-culted it from trc_update/BUILD.bazel. Apparently the value of "--executable" is actually a mapping: "end2end_integration" -> path_of(tools/end2end_integration). The rest is the business of common/base.py (which builds the map) and <whatevertest>/test.py that peruses it. Bazel is content because the executable file is listed as a data dependency of the test.
___
*[`acceptance/router_benchmark/test.py` line 3 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-Nio_6LkCeJzOrNlDU3E:-NioxMJGB476izJ8DzIo:b-896fjx) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/acceptance/router_benchmark/test.py#L3)):*

Nit copy pasted copyright
</blockquote></details>

Done
___
*[`pkg/snet/packet_conn.go` line 118 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioHL1Z3BeUFI2mbzSn:-NioxRgq5AeQb2nPwcpl:baairo1) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/pkg/snet/packet_conn.go#L118)):*

Could we deal with this Flow ID change in a separate PR, so that this gets the appropriate visibility?

IMO, `PacketConn` this is not the right API level to expose this. The `PacketConn` is sort of a "raw" connection. A user of this API can be expected to set the FlowID directly in the `Packet`.

I would add this to the `snet.Conn` instead, setting it when creating a packet in `scionConnWriter.WriteTo`. We can e.g. set this based on a hash of src/dst ISD, AS, Host address and Port (see e.g. https://www.rfc-editor.org/rfc/rfc6437#appendix-A).
I don't think this requires a backwards compatibility layer. From an application perspective, with using a constant FlowID, we would have possibly had an ordering guarantee between packets sent from different `snet.Conn`s. I doubt that this would even have worked in the first place (dispatcher / sending host might not guarantee the send order already). (Ab-)Using this in an application would probably have been tricky enough that I'm sure someone would have noticed that it's a bad idea.
</blockquote></details>

I suspected you'd object to such an ad-hoc API change. I just did a couple of runs with and without... the performance difference is actually unknown because I just discovered while deleting the code that I had forgotten to use the passed flowID value when serializing! So, I guess we can live without that until we had time to add the feature properly.
___
*[`tools/end2end/main.go` line 17 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioQe87DoKbxE55Kz6P:-Nip4sKY74HQF-o_PvcC:b-jctg8y) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/tools/end2end/main.go#L17)):*

I'm not sure about generalizing this end2end thing. It feels to me like this makes things much more complicated than they need to be.

Wouldn't a separate "blaster" be much shorter and simpler than this? We only want run one test at a time, and so we don't seem to benefit much from the `end2end_integration` harness. The server side becomes trivial if we don't send any replies and just report received packets out of band.

Note that we could also use the `end2end_integration` harness with a different command than end2end -- it just assumes that some flags are supported -- so this could still be used with a simpler "blaster" test.

I don't know, what do you think?
</blockquote></details>

A separate blaster may be a "simpler" solution to execute, but not to code, because end2end is already written. I was about to write a separate tool when I realized that the changes I needed to make to end2end to satisfy my needs where very simple. In the long run I agree with you. We'll eventually make a specialized tool. Not because it's easier but because it can test fewer components at a time (i.e. one), which will give more accurate readings. For now, I'm happy that I could meter all 5 traffic types in two runs with one topology. I find it pretty convenient to get some early results. So, if it's ok with you, I'd prefer to keep this for now and do another round of improvements in a later PR.

 
___
*[`tools/integration/integrationlib/common.go` line 150 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioO_7t9eL7ReI3uDWJ:-Nip8YUD8vIVURdBjtZw:bb6un1j) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/tools/integration/integrationlib/common.go#L150)):*

Nit. This is now printing "Retrying..." after the last attempt.
</blockquote></details>

Done

matzf (changes requested):
> Reviewed 1 of 14 files at r1, 5 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)* status: 7 of 16 files reviewed, 4 unresolved discussions (waiting on @jiceatscion)

___
*[`acceptance/router_benchmark/BUILD.bazel` line 8 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioYT4l40T2m_eajAWB:-NjD8ppq1NRnTShaeCbJ:b-p2s5v3) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/acceptance/router_benchmark/BUILD.bazel#L8)):*

I honestly don't know. I didn't invent this; I cargo-culted it from trc_update/BUILD.bazel. Apparently the value of "--executable" is actually a mapping: "end2end_integration" -> path_of(tools/end2end_integration). The rest is the business of common/base.py (which builds the map) and <whatevertest>/test.py that peruses it. Bazel is content because the executable file is listed as a data dependency of the test.
</blockquote></details>

Yup, I saw that it's the same in `trc_update`. I do understand the `end2end_integration`, but I'm confused about `end2end` which is invoked by `end2end_integration`. I think this is escapes the bazel sandbox somehow, which probably means that it only works because we build into `bin/` before.
___
*[`doc/manuals/router.rst` line 182 at r5](https://reviewable.io/reviews/scionproto/scion/4437#-NjD2Bn19DxxiV3OXGzZ:-NjD2Bn19DxxiV3OXGz_:bkmgnok) ([raw file](https://github.com/scionproto/scion/blob/180bbc73549dcd193f6169c8c5209e4d66527105/doc/manuals/router.rst#L182)):*
> ```reStructuredText
>       The send buffer size in bytes. 0 means use system default.
> 
>    .. option:: router.num_cores = <int> (Default: ``GOMAXPROCS``)
> ```

I would prefer not to create a redundant option to control this. The existing `GOMAXPROCS` already does what we need. I think it's a great idea to document better that this (and other GO...) environment variables apply.

In the test setup, adding an environment variable is just as easy as editing the toml, it can be set in the docker-compose file.
(Btw, the compose file also has some options to limit cpu usage. AFAIU, setting the `cpuset` option would result in an affinity mask which the go runtime does take into account for the default GOMAXPROC value. The go runtime does not, however, take into account CPU quotas with `cpus` value.)
___
*[`pkg/snet/dispatcher.go` line 59 at r5](https://reviewable.io/reviews/scionproto/scion/4437#-NjDFUnV8ioeZHn6crUn:-NjDFUnWA4cSGlvFtyNk:b-bedeu2) ([raw file](https://github.com/scionproto/scion/blob/180bbc73549dcd193f6169c8c5209e4d66527105/pkg/snet/dispatcher.go#L59)):*
> ```Go
> 	ia addr.IA,
> 	registration *net.UDPAddr,
> 	svc addr.SVC) (PacketConn, uint16, error) {
> ```

Nit: maybe revert this file and `snet/packet_conn.go` entirely to avoid any git blame and or merge issues.
___
*[`pkg/snet/packet_conn.go` line 118 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioHL1Z3BeUFI2mbzSn:-NjD3-AYAAAp1rt_D_ob:b2nhlkw) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/pkg/snet/packet_conn.go#L118)):*

I suspected you'd object to such an ad-hoc API change. I just did a couple of runs with and without... the performance difference is actually unknown because I just discovered while deleting the code that I had forgotten to use the passed flowID value when serializing! So, I guess we can live without that until we had time to add the feature properly.
</blockquote></details>

Ok!
___
*[`tools/end2end/main.go` line 17 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioQe87DoKbxE55Kz6P:-NjD3M7y5HCPWA3JB23C:b-hqfh4n) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/tools/end2end/main.go#L17)):*
> A separate blaster may be a "simpler" solution to execute, but not to code, because end2end is already written.

The end2end tool grew by about 40% lines of code to include the packetflood. My gut feeling is that with a copy-pasted and stream-lined version of end2end that _only_ does the "packetflood", we would end up with not much more code in total.
I don't like that this now conflates two mostly separate code paths just to share some of the boilerplate, but I think I don't feel strongly enough about this end2end test to object. Especially, as you already have plans to replace this. @scionproto/scion-core-team opinions?
___
*[`topology/router_bm.topo` line 1 at r5](https://reviewable.io/reviews/scionproto/scion/4437#-NjD8Lei6ukj7ogjuxk2:-NjD8Lei6ukj7ogjuxk3:bl8lazh) ([raw file](https://github.com/scionproto/scion/blob/180bbc73549dcd193f6169c8c5209e4d66527105/topology/router_bm.topo#L1)):*
> ```
> --- # Bespoke Topology for router benchmarking:
> ```

Nit. For other tests that expect specific topology, we put this .topo file into the testdata, i.e. `acceptance/router_benchmark/testdata/`. Probably a good way to keep things tidy.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)* status: 7 of 16 files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`acceptance/router_benchmark/BUILD.bazel` line 8 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioYT4l40T2m_eajAWB:-NjDN6tO5xllKgOEVD5x:bxs3bbk) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/acceptance/router_benchmark/BUILD.bazel#L8)):*

Yup, I saw that it's the same in `trc_update`. I do understand the `end2end_integration`, but I'm confused about `end2end` which is invoked by `end2end_integration`. I think this is escapes the bazel sandbox somehow, which probably means that it only works because we build into `bin/` before.
</blockquote></details>

mmm, end2end_integration might need to have bin/end2end as a data dependency too. It seems to be enough. Although I don't know precisely how bazel manages it.
___
*[`doc/manuals/router.rst` line 182 at r5](https://reviewable.io/reviews/scionproto/scion/4437#-NjD2Bn19DxxiV3OXGzZ:-NjDhEqr6_L0SYUeg-2Y:bjgc1fb) ([raw file](https://github.com/scionproto/scion/blob/180bbc73549dcd193f6169c8c5209e4d66527105/doc/manuals/router.rst#L182)):*

I would prefer not to create a redundant option to control this. The existing `GOMAXPROCS` already does what we need. I think it's a great idea to document better that this (and other GO...) environment variables apply.

In the test setup, adding an environment variable is just as easy as editing the toml, it can be set in the docker-compose file.
(Btw, the compose file also has some options to limit cpu usage. AFAIU, setting the `cpuset` option would result in an affinity mask which the go runtime does take into account for the default GOMAXPROC value. The go runtime does not, however, take into account CPU quotas with `cpus` value.)
</blockquote></details>

Well, yes it occured to me that we could use the env var directly. I just loathe env vars. They pop in and out of existence with no dicernable origin and no scope. I thought I was doing us a favor by adding an explicit config... but ok. I used the env var instead.
___
*[`pkg/snet/dispatcher.go` line 59 at r5](https://reviewable.io/reviews/scionproto/scion/4437#-NjDFUnV8ioeZHn6crUn:-NjDcXHB7Z5bwMlTTmvt:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/180bbc73549dcd193f6169c8c5209e4d66527105/pkg/snet/dispatcher.go#L59)):*

Nit: maybe revert this file and `snet/packet_conn.go` entirely to avoid any git blame and or merge issues.
</blockquote></details>

done
___
*[`tools/end2end/main.go` line 17 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioQe87DoKbxE55Kz6P:-NjDco2T1wzfENr5LjSz:b-6shb4r) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/tools/end2end/main.go#L17)):*

> A separate blaster may be a "simpler" solution to execute, but not to code, because end2end is already written.

The end2end tool grew by about 40% lines of code to include the packetflood. My gut feeling is that with a copy-pasted and stream-lined version of end2end that _only_ does the "packetflood", we would end up with not much more code in total.
I don't like that this now conflates two mostly separate code paths just to share some of the boilerplate, but I think I don't feel strongly enough about this end2end test to object. Especially, as you already have plans to replace this. @scionproto/scion-core-team opinions?
</blockquote></details>

Yeah, in terms of code quantity it ended-up about the same. It was just an easier (seeming) path to follow. Ideally I would just ditch this and replace it with the packet blaster approach, but something tells me it isn't going to be a 2 days job. The nitty-gritty is rather messy. I really don't want to drag this PR on forever. The test itself may not be great but the pattern is established and produces result. So I'd like to submit it. *However* I have no problem moving the packetflood game into its own binary. So...stay tuned for that one.
___
*[`topology/router_bm.topo` line 1 at r5](https://reviewable.io/reviews/scionproto/scion/4437#-NjD8Lei6ukj7ogjuxk2:-NjDdqmJ67GUFosDHYAG:bs2w4vt) ([raw file](https://github.com/scionproto/scion/blob/180bbc73549dcd193f6169c8c5209e4d66527105/topology/router_bm.topo#L1)):*

Nit. For other tests that expect specific topology, we put this .topo file into the testdata, i.e. `acceptance/router_benchmark/testdata/`. Probably a good way to keep things tidy.
</blockquote></details>

router_benchmark is like end2end_integration. You can tun it by hand outside the test framework. That's why I put the topo file in the same location. But yeah, I can move the file. Done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)* status: 4 of 17 files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`doc/manuals/router.rst` line 182 at r5](https://reviewable.io/reviews/scionproto/scion/4437#-NjD2Bn19DxxiV3OXGzZ:-NjE0JjD1zgoQf8F1TY7:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/180bbc73549dcd193f6169c8c5209e4d66527105/doc/manuals/router.rst#L182)):*

Well, yes it occured to me that we could use the env var directly. I just loathe env vars. They pop in and out of existence with no dicernable origin and no scope. I thought I was doing us a favor by adding an explicit config... but ok. I used the env var instead.
</blockquote></details>

done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)* status: 4 of 21 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`tools/end2end/main.go` line 17 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioQe87DoKbxE55Kz6P:-NjI33PfEMjY30eI3UU0:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/tools/end2end/main.go#L17)):*

Yeah, in terms of code quantity it ended-up about the same. It was just an easier (seeming) path to follow. Ideally I would just ditch this and replace it with the packet blaster approach, but something tells me it isn't going to be a 2 days job. The nitty-gritty is rather messy. I really don't want to drag this PR on forever. The test itself may not be great but the pattern is established and produces result. So I'd like to submit it. *However* I have no problem moving the packetflood game into its own binary. So...stay tuned for that one.
</blockquote></details>

done

matzf (commented):
> Reviewed 1 of 14 files at r1, 7 of 14 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)* status: 13 of 21 files reviewed, 5 unresolved discussions (waiting on @jiceatscion)

___
*[`acceptance/router_benchmark/BUILD.bazel` line 8 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioYT4l40T2m_eajAWB:-NjM2T9h47i28a6kTYjk:bhz6xjp) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/acceptance/router_benchmark/BUILD.bazel#L8)):*

mmm, end2end_integration might need to have bin/end2end as a data dependency too. It seems to be enough. Although I don't know precisely how bazel manages it.
</blockquote></details>

For the record; as you found out, the end2end is baked into the tester docker image, that's how it ends up in the right place.
___
*[`doc/manuals/router.rst` line 182 at r5](https://reviewable.io/reviews/scionproto/scion/4437#-NjD2Bn19DxxiV3OXGzZ:-NjM2llr4EiUy4hBXiIp:bhgybtt) ([raw file](https://github.com/scionproto/scion/blob/180bbc73549dcd193f6169c8c5209e4d66527105/doc/manuals/router.rst#L182)):*

done
</blockquote></details>

Ah, I see. I appreciate that you want to improve the configuration, sorry for rudely shooting it down! 
As we cannot get rid of the environment variable even if we wanted,  also exposing this as a configuration option means that the settings can conflict and one takes precedence, which, IMO, is even nastier than the environment variable alone. Hope this makes some sense.
___
*[`doc/manuals/router.rst` line 214 at r7](https://reviewable.io/reviews/scionproto/scion/4437#-NjM38LABWcdMvA5DZ7_:-NjM38LABWcdMvA5DZ7a:b-37xvit) ([raw file](https://github.com/scionproto/scion/blob/65a4fe584ed8745b489072558bf2c31b4e40cf31/doc/manuals/router.rst#L214)):*
> ```reStructuredText
> 
>       The number of kernel threads that go creates depends on the number of usable cores, which is
>       controlled by the environment variable ``GOMAXPROCS``. See :ref:`GOMAXPROCS <gomaxprocs>`.
> ```

Nit: use the `:env:` role, this directly creates a reference and the right formatting. Then we don't need the separate anchor `_gomaxprocs`.



_Suggestion:_
```reStructuredText
controlled by the environment variable :env:`GOMAXPROCS`.
```
___
*[`tools/end2end/main.go` line 17 at r2](https://reviewable.io/reviews/scionproto/scion/4437#-NioQe87DoKbxE55Kz6P:-NjM7ixc-Qj-a-oGDgD9:bcq5nwn) ([raw file](https://github.com/scionproto/scion/blob/dbcf003fd259c69af732c4596ab53a5a47ef9c53/tools/end2end/main.go#L17)):*

done
</blockquote></details>

Ok, got it. Seems good to me.
___
*[`tools/end2end/main.go` line 360 at r7](https://reviewable.io/reviews/scionproto/scion/4437#-NjM8zCm9ItC3-t-r6at:-NjM8zCm9ItC3-t-r6au:bfvxkvs) ([raw file](https://github.com/scionproto/scion/blob/65a4fe584ed8745b489072558bf2c31b4e40cf31/tools/end2end/main.go#L360)):*
> ```Go
> }
> 
> func (c *client) ping(ctx context.Context, n int, path snet.Path, logIfOk bool) error {
> ```

Nit: remove the `logIfOk` (`true` here in `end2end`, `false` in `end2endblast`).
___
*[`tools/end2endblast/main.go` line 117 at r7](https://reviewable.io/reviews/scionproto/scion/4437#-NjM9Czj2FNhAKbf6Gv9:-NjM9Czj2FNhAKbf6GvA:b-tqpxg) ([raw file](https://github.com/scionproto/scion/blob/65a4fe584ed8745b489072558bf2c31b4e40cf31/tools/end2endblast/main.go#L117)):*
> ```Go
> 	flag.Var(timeout, "timeout", "The timeout for each attempt")
> 	flag.BoolVar(&epic, "epic", false, "Enable EPIC")
> 	flag.BoolVar(&traces, "traces", true, "Enable Jaeger traces")
> ```

Nit: as far as I understand, the traces as they are implemented here never make much sense for the blaster. I think it would make sense to remove this flag here and drop all the related code. In the end2end, on the other hand, we could revert to always enable this.
___
*[`tools/end2endblast/main.go` line 306 at r7](https://reviewable.io/reviews/scionproto/scion/4437#-NjMBPqnAm4V3eiYxing:-NjMBPqnAm4V3eiYxinh:bnbg9k0) ([raw file](https://github.com/scionproto/scion/blob/65a4fe584ed8745b489072558bf2c31b4e40cf31/tools/end2endblast/main.go#L306)):*
> ```Go
> 	defer c.sdConn.Close()
> 	c.errorPaths = make(map[snet.PathFingerprint]struct{})
> 	pong_out := make(chan int)
> ```

Nit: `pongOut` (also `pingResult`, `pongResult`, ...)
___
*[`tools/end2endblast/main.go` line 359 at r7](https://reviewable.io/reviews/scionproto/scion/4437#-NjMAZZtEPILcLwyebL4:-NjMAZZtEPILcLwyebL5:b-io0ppp) ([raw file](https://github.com/scionproto/scion/blob/65a4fe584ed8745b489072558bf2c31b4e40cf31/tools/end2endblast/main.go#L359)):*
> ```Go
> 			return true
> 		}
> 		c.path = p // struct fields cannot be assigned with :=
> ```

Nit: in contrast to `end2end`, this does not switch to alternative paths on errors. We could clean this up by moving this `getRemote` call out, before the first call of the `blindPing`.
We can also remove everything related to `errorPaths`

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)* status: 13 of 21 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`doc/manuals/router.rst` line 182 at r5](https://reviewable.io/reviews/scionproto/scion/4437#-NjD2Bn19DxxiV3OXGzZ:-NjMU2PV3sJLBX0Ts_8l:b796uc3) ([raw file](https://github.com/scionproto/scion/blob/180bbc73549dcd193f6169c8c5209e4d66527105/doc/manuals/router.rst#L182)):*

Ah, I see. I appreciate that you want to improve the configuration, sorry for rudely shooting it down! 
As we cannot get rid of the environment variable even if we wanted,  also exposing this as a configuration option means that the settings can conflict and one takes precedence, which, IMO, is even nastier than the environment variable alone. Hope this makes some sense.
</blockquote></details>

Well, yeah. For the config approach to be an actual improvement, we would need to also suppress the effect of the  env var (which we can, btw: call GOMAXPROC(NumCPU()) by default). But that's added mess for too little benefit. I do find that there are too many sources of configuration in general; not just for this one thing, so if it gets addressed it better be as a separate more comprehensive effort.
___
*[`tools/end2end/main.go` line 360 at r7](https://reviewable.io/reviews/scionproto/scion/4437#-NjM8zCm9ItC3-t-r6at:-NjMXGE00LaNvL6Vd3Uk:bb7f2mf) ([raw file](https://github.com/scionproto/scion/blob/65a4fe584ed8745b489072558bf2c31b4e40cf31/tools/end2end/main.go#L360)):*

Nit: remove the `logIfOk` (`true` here in `end2end`, `false` in `end2endblast`).
</blockquote></details>

done
___
*[`tools/end2endblast/main.go` line 117 at r7](https://reviewable.io/reviews/scionproto/scion/4437#-NjM9Czj2FNhAKbf6Gv9:-NjMaY_O8d1CoPUMw3gF:b-nc886a) ([raw file](https://github.com/scionproto/scion/blob/65a4fe584ed8745b489072558bf2c31b4e40cf31/tools/end2endblast/main.go#L117)):*

Nit: as far as I understand, the traces as they are implemented here never make much sense for the blaster. I think it would make sense to remove this flag here and drop all the related code. In the end2end, on the other hand, we could revert to always enable this.
</blockquote></details>

Done. Might as well now.
___
*[`tools/end2endblast/main.go` line 306 at r7](https://reviewable.io/reviews/scionproto/scion/4437#-NjMBPqnAm4V3eiYxing:-NjMgUzm8amHTb1nQ1eb:b-rdy7ru) ([raw file](https://github.com/scionproto/scion/blob/65a4fe584ed8745b489072558bf2c31b4e40cf31/tools/end2endblast/main.go#L306)):*

Nit: `pongOut` (also `pingResult`, `pongResult`, ...)
</blockquote></details>

Done. Sorry for letting snakes in the camel pen.
___
*[`tools/end2endblast/main.go` line 359 at r7](https://reviewable.io/reviews/scionproto/scion/4437#-NjMAZZtEPILcLwyebL4:-NjN4CaL1_qsJWChy_qb:b-xmfd1j) ([raw file](https://github.com/scionproto/scion/blob/65a4fe584ed8745b489072558bf2c31b4e40cf31/tools/end2endblast/main.go#L359)):*

Nit: in contrast to `end2end`, this does not switch to alternative paths on errors. We could clean this up by moving this `getRemote` call out, before the first call of the `blindPing`.
We can also remove everything related to `errorPaths`
</blockquote></details>

Done... and then some.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)* status: 12 of 21 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`doc/manuals/router.rst` line 214 at r7](https://reviewable.io/reviews/scionproto/scion/4437#-NjM38LABWcdMvA5DZ7_:-NjN4QXT6mNON5WRdihv:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/65a4fe584ed8745b489072558bf2c31b4e40cf31/doc/manuals/router.rst#L214)):*

Nit: use the `:env:` role, this directly creates a reference and the right formatting. Then we don't need the separate anchor `_gomaxprocs`.
</blockquote></details>

done

matzf (approved):
> Reviewed 1 of 14 files at r1, 1 of 14 files at r6, 9 of 9 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4437)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4444, opened 2023-11-23
An improved router benchmark that only runs one router

jiceatscion:
This is an alternative to router_benchmark and meant to eventually replace it.
It's main feature is that it only runs a single router. As a result it's measured performance is far less dependent on the number of cores available on the test system.

It is heavily inspired by the braccept test, in that:
* it relies on the test harness to create an isolated network and veth interfaces to connect the test driver to the router.
* it manufactures packets that are fed directly into the router's interfaces and it optionally captures packets for verification in the same manner.
* it assumes a custom topology and cannot be usefully run on any other topology.

It differs from braccept in that:
* It does far less verification of the forwarded packets (it's braccept's job to do that).
* It is considerably simplified, mainly because of the lighter verifications.
* It relies on a simple address-assignment scheme so code maintainers don't have to keep tens of addresses in mind.
* It does not assume that the router runs in a container network; it could also talk to a real router if given the names and mac addresses of the relevant interfaces. The test harness (or the user) is responsible for supplying the interface names and mac addresses to the test driver.
* It does not requires the test harness (or the user) to have any understanding of the topology (only to configure the router with it). Knowledge of the specifics of the captive network to be configured are supplied by the test driver.

Currently there's only one test case. I'll wait until the review process irons any kinks before I write more tests.

The similarity between this and braccept means that, in the future, we could re-converge them. Notably, the benefits of the simple addressing scheme would make the braccept test cases easier to maintain or expand.


matzf (commented):
> Reviewed 4 of 13 files at r1, 2 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: 6 of 14 files reviewed, 11 unresolved discussions (waiting on @jiceatscion)

___
*[`acceptance/router_newbenchmark/BUILD.bazel` line 74 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njx1HyC2WBQhRRpoUHM:-Njx1HyC2WBQhRRpoUHN:bgrmw3e) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/BUILD.bazel#L74)):*
> ```Starlark
> 
> go_binary(
>     name = "router_newbenchmark",
> ```

I think this is redundant (same as `brload`).
___
*[`acceptance/router_newbenchmark/main.go` line 71 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwxMDk-_4XqnJ5ExJQ:-NjwxMDk-_4XqnJ5ExJR:bmlktid) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L71)):*
> ```Go
> // initDevices inventories the available network interfaces, picks the ones that a case may inject
> // traffic into, and associates them with a AF Packet interface.
> func initDevices() error {
> ```

Nit: `handles` does not need to be a gloabal, return it from here.
___
*[`acceptance/router_newbenchmark/main.go` line 105 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njwx_IE86TCroVe5CJf:-Njwx_IF-ozPHYVxaQg6:b57scfp) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L105)):*
> ```Go
> 		fmt.Println(listInterfaces())
> 		return 0
> 	}
> ```

The flags of this tool are sufficiently complex that the `flags` package becomes ugly. I'd suggest to use `cobra`, which we use in other places already, and which can handle subcommands etc.
___
*[`acceptance/router_newbenchmark/main.go` line 122 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njwp_-0-cZo5N6361Y9:-Njwp_-0-cZo5N6361YA:b9t97xx) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L122)):*
> ```Go
> 	}
> 
> 	artifactsDir, err := os.MkdirTemp("", "brload_")
> ```

Nit: clean this up?
Also, it seems we only use this if `dir` / TEST_ARTIFACTS_DIR are not set, so only create when needed?
___
*[`acceptance/router_newbenchmark/main.go` line 201 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njww3liAgGJZhvBRywB:-Njww3liAgGJZhvBRywC:b-yfd9uu) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L201)):*
> ```Go
> 			}
> 		}
> 	}()
> ```

Nit: move the main body of this `go func()` into a separate function, passing in the `readPktFrom` handle; something like `func receivePkts(handle *afpacket.TPacket, expectedPayload string) int`.
___
*[`acceptance/router_newbenchmark/main.go` line 223 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwlSC69R6qEQz3qxHo:-NjwlSC69R6qEQz3qxHp:b3srjah) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L223)):*
> ```Go
> 	close(packetChan)
> 
> 	outcome := <-listenerChan
> ```

Perhaps a bit more elegant; instead of Sleep + close + read, read from the listenerChan with a timeout.
___
*[`acceptance/router_newbenchmark/test.py` line 106 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwghMX1na1Cpj3RkV3:-NjwghMX1na1Cpj3RkV4:b-5dk2i) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/test.py#L106)):*
> ```Python
>     # Accepts an IntfReq records names and mac addresses into an Intf and associates it with
>     # the request label.
>     def create_interface(self, req: IntfReq, ns: str) -> Intf:
> ```

Nit: type, does not actually return anything?
___
*[`acceptance/router_newbenchmark/test.py` line 165 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwjDkv1YD5_PcdfvSs:-NjwjDkv1YD5_PcdfvSt:b-sifwoe) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/test.py#L165)):*
> ```Python
>         brload = self.get_executable("brload")
>         output = exec_sudo(f"{brload.executable} -artifacts {self.artifacts} "
>                            "-show_interfaces")
> ```

Nit: don't run as sudo.
___
*[`acceptance/router_newbenchmark/test.py` line 183 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njx1fVp8Z5l74O2X6Jo:-Njx1fVp8Z5l74O2X6Jp:b2f1rdg) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/test.py#L183)):*
> ```Python
>                     "--network container:prometheus "
>                     "--name router "
>                     "bazel/acceptance/router_newbenchmark:router")
> ```

Is this the name of the docker image? If so, do we then not need this clunky "container-loader" argument passed to the test in the BUILD file?
___
*[`acceptance/router_newbenchmark/topo.go` line 174 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwrJTFCl1uDCGS4epO:-NjwrJTFCl1uDCGS4epP:b-2f4ybl) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/topo.go#L174)):*
> ```Go
> 		sb.WriteString(i.peerIP.String())
> 		sb.WriteString("\n")
> 	}
> ```

This is quite hard to understand. Document what the expected format is?
___
*[`acceptance/router_newbenchmark/conf/topology.json` line 14 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwsiZJ-fa-T64vJc7c:-NjwsiZJ-fa-T64vJc7d:bau6zk1) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/conf/topology.json#L14)):*
> ```JSON
>   "discovery_service": {
>     "cs1-ff00_0_1-1": {
>       "addr": "172.20.0.28:30252"
> ```

Nit; Is it legal to leave control_service/discovery_service out? We don't use them.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: 6 of 15 files reviewed, 6 unresolved discussions (waiting on @matzf)

___
*[`acceptance/router_newbenchmark/BUILD.bazel` line 74 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njx1HyC2WBQhRRpoUHM:-Njx5DoGE7nAUNaV97I7:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/BUILD.bazel#L74)):*

I think this is redundant (same as `brload`).
</blockquote></details>

Done.
___
*[`acceptance/router_newbenchmark/main.go` line 71 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwxMDk-_4XqnJ5ExJQ:-Njx6Lv0BqZna_zlz0DD:b-896fjx) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L71)):*

Nit: `handles` does not need to be a gloabal, return it from here.
</blockquote></details>

Done
___
*[`acceptance/router_newbenchmark/main.go` line 122 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njwp_-0-cZo5N6361Y9:-Njx6bJkCnl6QPsF4Dws:b-kkou2j) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L122)):*

Nit: clean this up?
Also, it seems we only use this if `dir` / TEST_ARTIFACTS_DIR are not set, so only create when needed?
</blockquote></details>

I was also wondering why that was created unconditionally. The only things that came to mind were that it could double as cheap mutual exclusion between parallel runs (can't see why that would be needed) or that it was just a nicer looking code flow. In addition, it seems that the only thing we want out of that dir is to read files from it. There's no alternative if the files don't exist. So, creating the dir serves no purpose. I think this is code left-over from something that used to be more complicated. Removed it entirely.

Done
___
*[`acceptance/router_newbenchmark/main.go` line 201 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njww3liAgGJZhvBRywB:-NjxG5wS6JrJ0ko6j51C:b-sxsy7e) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L201)):*

Nit: move the main body of this `go func()` into a separate function, passing in the `readPktFrom` handle; something like `func receivePkts(handle *afpacket.TPacket, expectedPayload string) int`.
</blockquote></details>

done
Not sure it looks better, but no worse. I kinda like running go routines in the context of the function that creates them. Saves on boilerplate.
___
*[`acceptance/router_newbenchmark/main.go` line 223 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwlSC69R6qEQz3qxHo:-NjxFyEFDzCgXoKhZtgX:bdytsuo) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L223)):*

Perhaps a bit more elegant; instead of Sleep + close + read, read from the listenerChan with a timeout.
</blockquote></details>

Yeah. Here's the prettier version.
___
*[`acceptance/router_newbenchmark/test.py` line 106 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwghMX1na1Cpj3RkV3:-NjxH-ikBsLyjrMMTa6K:b-zhvk5i) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/test.py#L106)):*

Nit: type, does not actually return anything?
</blockquote></details>

Ah, yes, no more. It used to.
___
*[`acceptance/router_newbenchmark/test.py` line 165 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwjDkv1YD5_PcdfvSs:-NjxIFH6Aj1XnUbUidBJ:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/test.py#L165)):*

Nit: don't run as sudo.
</blockquote></details>

done
___
*[`acceptance/router_newbenchmark/test.py` line 183 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njx1fVp8Z5l74O2X6Jo:-NjxJFVB-8yMDXHVENLg:bcv45pt) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/test.py#L183)):*

Is this the name of the docker image? If so, do we then not need this clunky "container-loader" argument passed to the test in the BUILD file?
</blockquote></details>

Not sure if that's necessary or not. May be we don't have to build one specificly for the test? Will look tomorrow.
___
*[`acceptance/router_newbenchmark/conf/topology.json` line 14 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwsiZJ-fa-T64vJc7c:-NjxK2hW4-xLxx-ShYeV:b5ogntx) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/conf/topology.json#L14)):*

Nit; Is it legal to leave control_service/discovery_service out? We don't use them.
</blockquote></details>

I have been wondering. I haven't tried to remove them. I will.
___
*[`acceptance/router_newbenchmark/topo.go` line 174 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwrJTFCl1uDCGS4epO:-NjxK-OTDQ8NdK8ATVRo:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/topo.go#L174)):*

This is quite hard to understand. Document what the expected format is?
</blockquote></details>

Done.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: 6 of 15 files reviewed, 6 unresolved discussions (waiting on @matzf)

___
*[`acceptance/router_newbenchmark/test.py` line 183 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njx1fVp8Z5l74O2X6Jo:-Nk-rjgVEEXU--Xc2y_C:bh3n08g) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/test.py#L183)):*

Not sure if that's necessary or not. May be we don't have to build one specificly for the test? Will look tomorrow.
</blockquote></details>

I think I figured out how it's supposed to work. Yes the build file needs to pass container-loader. I also named just the same as everywhere else: posix-router:latest and made test.py run that. Deleted the custom router image.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: 6 of 15 files reviewed, 5 unresolved discussions (waiting on @matzf)

___
*[`acceptance/router_newbenchmark/main.go` line 105 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njwx_IE86TCroVe5CJf:-Nk-z6q2EAF8hD2CSSK_:bb7f2mf) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L105)):*

I don't find cobra any prettier, but if it's our standard, I'll comply. Stay tuned.
</blockquote></details>

done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: 6 of 15 files reviewed, 4 unresolved discussions (waiting on @matzf)

___
*[`acceptance/router_newbenchmark/conf/topology.json` line 14 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwsiZJ-fa-T64vJc7c:-Nk0r_ri92tl1074lCv_:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/conf/topology.json#L14)):*

I have been wondering. I haven't tried to remove them. I will.
</blockquote></details>

done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: 9 of 15 files reviewed, 6 unresolved discussions (waiting on @matzf)

___
*[`acceptance/router_newbenchmark/main.go` line 201 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njww3liAgGJZhvBRywB:-NkGR_wz8Tts0rjCiY49:b-ygjayo) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/main.go#L201)):*

I got rather confused re-reading this, because the code looks like it's counting packets, but it does return after the first one (as documented, I know, but i still misread it). Could we make this return a boolean instead of an int?
I think it would help to keep the channel logic in the context of the caller like so:

```
go func() {
    defer log.HandlePanic()
    gotone := receivePackets(packetChan, payloadString)
    listenerChan <- gotone
    close(listenerChan)
}
```
</blockquote></details>

As I said I do like the possiblity of counting all packets via a simple change. But even then I can do as you say. It hadn't occured to me and it does read better. So...
done
___
*[`acceptance/router_newbenchmark/main.go` line 84 at r5](https://reviewable.io/reviews/scionproto/scion/4444#-NkFnYXtBjzb5N2pPuYK:-NkGPGG5BeN41_3JOLDd:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/40adc9ea08a26c423df4f5230bf40beb9c51047c/acceptance/router_newbenchmark/main.go#L84)):*

kebab-case for commands and subcommands and flags please.
</blockquote></details>

Done.
___
*[`acceptance/router_newbenchmark/main.go` line 217 at r5](https://reviewable.io/reviews/scionproto/scion/4444#-NkFuVjP8LH1yeNV6KJM:-NkGQwWw4YQXb4YDY7kx:b-epq6n1) ([raw file](https://github.com/scionproto/scion/blob/40adc9ea08a26c423df4f5230bf40beb9c51047c/acceptance/router_newbenchmark/main.go#L217)):*

> // .... Close the channel ourselves. After this, the next loop is guaranteed an outcome.

True, but the outcome is also guaranteed to be 0, no? Meaning, we don't seem to need the loop and the `close`; if we run into timeout, we can directly say it didn't work.
</blockquote></details>

That is true only if the listener outputs its result and terminates at the first packet (something you suggested in another comment). But I like to retain the ability to count all the packets without rewriting the code, by just removing the return. In that case the listener can still be waiting for more packets and yet have some non-zero outcome in store for us.
___
*[`acceptance/router_newbenchmark/test.py` line 59 at r5](https://reviewable.io/reviews/scionproto/scion/4444#-NkGH-YX02HTmjZXgU8K:-NkGVw1v3oQtAWQhuxkD:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/40adc9ea08a26c423df4f5230bf40beb9c51047c/acceptance/router_newbenchmark/test.py#L59)):*

Would it make sense to pick a different arbitrary prefix here, to make sure that we don't accidentally rely on this being the same as the `macAddr` func in go?
</blockquote></details>

Done.
___
*[`acceptance/router_newbenchmark/test.py` line 120 at r5](https://reviewable.io/reviews/scionproto/scion/4444#-NkG287B32MLUzE8PaBk:-NkGXSkx0_7OMINHxwHt:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/40adc9ea08a26c423df4f5230bf40beb9c51047c/acceptance/router_newbenchmark/test.py#L120)):*

Here and elsewhere; single value formatting string would be shorter and more readable passing just a normal variable: `f"{brIntf}"` -> just `brIntf`. 

For cases where something is not already a string we still need to apply the conversion (plumbum does not do it automatically I believe). For this, I'd prefer `str(val)` over `f"{val}"`. But it seems most things here are strings anyway...
</blockquote></details>

Done.
___
*[`acceptance/router_newbenchmark/test.py` line 169 at r5](https://reviewable.io/reviews/scionproto/scion/4444#-NkG3SoA78AyD52NuwIq:-NkK9bNa5y7Q87ocolwI:b-xp2yq) ([raw file](https://github.com/scionproto/scion/blob/40adc9ea08a26c423df4f5230bf40beb9c51047c/acceptance/router_newbenchmark/test.py#L169)):*

Remove the `'` in the format string, then you don't need to replace it in the output either.
</blockquote></details>

done
No idea why these were there. Probably inherited from earlier invocations that somehow caused the removal of one layer of quotes.
___
*[`acceptance/router_newbenchmark/topo.go` line 174 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwrJTFCl1uDCGS4epO:-NkKFUfx0lqjq8OokjF1:b7kfxvp) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/topo.go#L174)):*

I don't really get this 24 prefix length. Is this important?
</blockquote></details>

Meh, it has to be supplied to ip addr add. Since the test is supplying the IPs, it should decide on the subnets. We could hard-code it in test.py, but why break the model?
___
*[`acceptance/router_newbenchmark/cases/topo.go` line 102 at r6](https://reviewable.io/reviews/scionproto/scion/4444#-NkG-llaEALsCR_mzUNr:-NkKEvGo0mYxJVmQ4Ii7:bilt41x) ([raw file](https://github.com/scionproto/scion/blob/55fd64b632097c2509a3f42dc8dd709d64a711bb/acceptance/router_newbenchmark/cases/topo.go#L102)):*

Nit: maybe call this "device names" to distinguish it from "AS/router interface"? It's easy to get confused here. Same for the `interfaceName` function.
</blockquote></details>

done . Indeed it is clearer.
___
*[`acceptance/router_newbenchmark/cases/topo.go` line 108 at r6](https://reviewable.io/reviews/scionproto/scion/4444#-NkG-b-G-bo8YvvKaI0c:-NkKF0CG1dyAOomiCULd:bb7f2mf) ([raw file](https://github.com/scionproto/scion/blob/55fd64b632097c2509a3f42dc8dd709d64a711bb/acceptance/router_newbenchmark/cases/topo.go#L108)):*

Nit: Might be better to initialize the maps in `InitInterfaces` to make it obvious that this must be called before the case function.
</blockquote></details>

done
___
*[`acceptance/router_newbenchmark/cases/topo.go` line 125 at r6](https://reviewable.io/reviews/scionproto/scion/4444#-NkG0lUP0O3ndVArVzc1:-NkKF53R8GUwrK-MAeGf:b1x4vnz) ([raw file](https://github.com/scionproto/scion/blob/55fd64b632097c2509a3f42dc8dd709d64a711bb/acceptance/router_newbenchmark/cases/topo.go#L125)):*

As this is an internal tool, I think we can let this slide, but panic for input checking is pretty gross.
</blockquote></details>

I disagree. I did this deliberately. Crashing in this case is the most helpful thing we can do. It attracts attention and provides the cause.

matzf (commented):
> Reviewed 2 of 6 files at r5, 1 of 4 files at r7, 1 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: 13 of 15 files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

___
*[`acceptance/router_newbenchmark/BUILD.bazel` line 74 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-Njx1HyC2WBQhRRpoUHM:-NkZVLmKBrCNp0zXiBHk:bbb09e2) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/BUILD.bazel#L74)):*

Done.
</blockquote></details>

This unused `router_newbenchmark` go binary is still here though.
___
*[`acceptance/router_newbenchmark/topo.go` line 174 at r2](https://reviewable.io/reviews/scionproto/scion/4444#-NjwrJTFCl1uDCGS4epO:-NkZV1g57CESF5uClMwe:b-usv013) ([raw file](https://github.com/scionproto/scion/blob/171acadacea2ec67a7020abe9be1b017decadc8e/acceptance/router_newbenchmark/topo.go#L174)):*

Meh, it has to be supplied to ip addr add. Since the test is supplying the IPs, it should decide on the subnets. We could hard-code it in test.py, but why break the model?
</blockquote></details>

Ok, I see. 
___
*[`acceptance/router_newbenchmark/cases/topo.go` line 125 at r6](https://reviewable.io/reviews/scionproto/scion/4444#-NkG0lUP0O3ndVArVzc1:-NkZUbgXCGoP9ll_ArUT:b-47r6uc) ([raw file](https://github.com/scionproto/scion/blob/55fd64b632097c2509a3f42dc8dd709d64a711bb/acceptance/router_newbenchmark/cases/topo.go#L125)):*

I disagree. I did this deliberately. Crashing in this case is the most helpful thing we can do. It attracts attention and provides the cause.
</blockquote></details>

Ok. I disagree with that, too. :wink: How is this better than returning an error explaining that the argument was malformed? If I see a panic, I assume that the program is broken, not my invocation of it. Never mind though, leave it.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: 19 of 20 files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`acceptance/router_newbenchmark/cases/topo.go` line 94 at r11](https://reviewable.io/reviews/scionproto/scion/4444#-NkoXIPGDS6wr_gO6w4Y:-NkpOtBc9tQTKC2-nRWN:bluwxlv) ([raw file](https://github.com/scionproto/scion/blob/4166d27e8cc35fadf9a9b2662b38dc0941ae6190/acceptance/router_newbenchmark/cases/topo.go#L94)):*

Nit: unexport this, it seems strictly internal (only used to initialize `intfMap`).
</blockquote></details>

done
(and the Subnet = max() thing too).

matzf (approved):
> Reviewed 6 of 7 files at r13, 3 of 3 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: all files reviewed, 2 unresolved discussions (waiting on @jiceatscion)

___
*[`go.mod` line 133 at r14](https://reviewable.io/reviews/scionproto/scion/4444#-NkyL2bR-Tr4Q_tGUP3r:-NkyL2bR-Tr4Q_tGUP3s:b-ogez9l) ([raw file](https://github.com/scionproto/scion/blob/c1e1abc9c94ef50dd497e37f698bd7e4913a1831/go.mod#L133)):*
> ```XML
> )
> 
> go 1.18
> ```

Can you please mention this in the PR description, so we won't forget to mention this in the next release notes (affects downstream).
___
*[`acceptance/router_newbenchmark/cases/topo.go` line 70 at r14](https://reviewable.io/reviews/scionproto/scion/4444#-NkyLw2X-RuvSdJg-wtW:-NkyLw2X-RuvSdJg-wtX:biiqdlg) ([raw file](https://github.com/scionproto/scion/blob/21f6013a8f541dfb092a739fbf133a9cffdb393c/acceptance/router_newbenchmark/cases/topo.go#L70)):*
> ```Go
> // the local AS. This works if there are no cycles. Else there could be subnet number collisions.
> func PublicIP(localAS byte, remoteAS byte) netip.Addr {
> 	subnetNr := byte(max(int(remoteAS), int(localAS)))
> ```

The type conversions are not necessary, `max` is generic.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`go.mod` line 133 at r14](https://reviewable.io/reviews/scionproto/scion/4444#-NkyL2bR-Tr4Q_tGUP3r:-NkyWTSd1fyjxl_5ud1Q:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/c1e1abc9c94ef50dd497e37f698bd7e4913a1831/go.mod#L133)):*

Can you please mention this in the PR description, so we won't forget to mention this in the next release notes (affects downstream).
</blockquote></details>

done
___
*[`acceptance/router_newbenchmark/cases/topo.go` line 70 at r14](https://reviewable.io/reviews/scionproto/scion/4444#-NkyLw2X-RuvSdJg-wtW:-NkyVgEh5PHv9tusacYf:bbyl009) ([raw file](https://github.com/scionproto/scion/blob/21f6013a8f541dfb092a739fbf133a9cffdb393c/acceptance/router_newbenchmark/cases/topo.go#L70)):*

The type conversions are not necessary, `max` is generic.
</blockquote></details>

Yeah, some of the random things I tried when the linter was screaming at me.
done

matzf (approved):
> Reviewed 2 of 2 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4444)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4448, opened 2023-11-24
dist: build debian packages for multiple platforms

matzf:
Build debian packages for amd64, arm64, i386 and armel.
- There are separate packages for router, control, daemon, dispatcher, gateway and tools (scion and scion-pki).
- The packages include systemd unit files to run the services. For the daemon, dispatcher and gateway one instance per host is supported by the systemd service, and a default configuration file is included. For router and control, multiple instances per host are supported, and as a consequence of this, no default configuration file is provided.
- Currently, there is no man page contained in the packages. We should be able to build these from our existing manuals, but it seems to require a significant amount of fiddling to get something useful.

Building the .deb packages uses bazel with `rules_pkg`.
The target `//dist:deb_all` cross-builds packages for the default set of target platforms. Alternatively, the target `//dist:deb` allows to build (all) packages for the current target platform. This current platform can be set with the `--platforms`  bazel option (see https://github.com/bazelbuild/rules_go#how-do-i-cross-compile for more details).
- To increase reuse of build results while cross-building, some internal targets related to openapi forcibly ignore the target platform
- The package version is based on the current git tag. `rules_pkg` can include this version _in_ the package metadata, but bazel _cannot_ spit out appropriately named package files. This is addressed by copying and renaming the package files after build in a make target `make dist-deb`.

Contributes to #4425.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4448)

PR #4449, opened 2023-11-24
build: allow building without scion-ci

mlegner:
My motivation for this is an error on ARM64 machines when building `scion-ci`:
```
(13:40:17) ERROR: /home/ubuntu/.cache/bazel/_bazel_ubuntu/d590279010957ac1dca373a0e19d0c3c/external/io_bazel_rules_go/BUILD.bazel:86:17: While resolving toolchains for target @io_bazel_rules_go//:cgo_context_data: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type.
```

However, it may be interesting irrespective of this to just build the main binaries without the CI tools.

Maybe the naming could be improved, let me know what you think.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,11 @@ bazel:
 	tar -kxf bazel-bin/scion.tar -C bin
 	tar -kxf bazel-bin/scion-ci.tar -C bin
 
+bazel-no-ci:
+	rm -f bin/*
+	bazel build //:scion
+	tar -kxf bazel-bin/scion.tar -C bin
+
 test:
 	bazel test --config=unit_all
 


matzf (commented):
Thanks!

For context: the error occurs because building some of the tools requires cgo. Specifically, this is `braccept` which uses the cgo-based `gopacket/afpacket` library to write raw packets.
As I see it, the `braccept` utility is solely built used for the `router_multi` test, which is triggered from within bazel, and does not need the output of the `scion-ci` target. As an alternative approach, I think it could make sense to just drop `braccept` from the `scion-ci` target.

If we decide to go for the approach with different make targets, I think it could make sense to call the non-ci variant just something plain like `build`, and the variant with the CI tools something like `build-dev`.
Or, something like `dist-bin` could make sense for the no-CI variant; I was planning to add a make target `dist-deb` for building debian packages (see WIP #4448), so this could also fit.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4449#-:-Nk0nlrb3cMQclk4TN9D:b-bytv15)* status: 0 of 1 files reviewed, all discussions resolved

jiceatscion (commented):
gopacket is also used by brload now.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4449#-:-Nk120aLD9klM6yTEsvU:b-osuf82)* status: 0 of 1 files reviewed, all discussions resolved

mlegner:
> If we decide to go for the approach with different make targets, I think it could make sense to call the non-ci variant just something plain like build, and the variant with the CI tools something like build-dev.
Or, something like dist-bin could make sense for the no-CI variant; I was planning to add a make target dist-deb for building debian packages (see WIP https://github.com/scionproto/scion/pull/4448), so this could also fit.

All of these sound fine to me, no strong preference either way. I'm happy to update this PR, but feel free to just close it and include this in a separate PR.

matzf (commented):
Ok. Let's just add the make target that you propose then. 
My preference for the name would be something like `dist`, `dist-bin`, or `build` (and `build-dev` for the "ci" variant) -- nothing with "bazel".
Please mark the target as `.PHONY`.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4449#-:-NkFI_0F2T60nxcUKMU6:bdmy1x3)* status: 0 of 1 files reviewed, all discussions resolved

matzf (approved):
:100: 

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4449#-:-NkFkFJjB0_BQhBXuIG0:bmi8ppe)* status: all files reviewed, 1 unresolved discussion (waiting on @mlegner)

___
*[`doc/dev/setup.rst` line 75 at r2](https://reviewable.io/reviews/scionproto/scion/4449#-NkFkLIBCVXvrMKnfrRN:-NkFkLIBCVXvrMKnfrRO:bx9w607) ([raw file](https://github.com/scionproto/scion/blob/e0798583839da6267a0621e7ffb756d1689ec4a0/doc/dev/setup.rst#L75)):*
> ```reStructuredText
>       make
> 
>   .. hint:: This installs tools for tests in addition to the main SCION services (e.g., `end2end`);
> ```

Nit: I think "installs" could be mistaken here, just "build" or "compile" might be better.

matzf (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4449)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mlegner)

PR #4450, opened 2023-11-30
prometheus: fix copying of metrics labels

sustrik:
When copying metric labels we assumed that `append` will always create a copy of the label array. This is not necessarily the case. In such case two metrics may end up with the same underlaying array of labels and change to one of them also overwrites the labels in the other one.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4450)


matzf (approved):
Classic. :wink: 

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4450#-:-NkUFW7mErw0u9f_t_Bq:b-my5p2c)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sustrik)

PR #4456, opened 2023-12-08
router: make the benchmark and the router more adaptive to core count

jiceatscion:
In the router. Compute a better default for the number of processors based on multiple observations: given enough streams to occupy all processors, performance is degraded if there are as many processors as cores. Best performance is obtained by setting aside two cores for other tasks. Making this the default simplifies the benchmarking process by not having to configure an override for each type of target machine being tested.

In the benchmark. Run the test with many streams. It reflects reality and allows a very parallel router to perform at its best without distorting the performance of a less-parallel one. Make sure that the number of streams is likely to be a multiple of the number of cores so the cores are loaded evenly in spite of there being typically fewer streams than in real life. (TBD, learn the router's internal processors count and just use that).

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4456)

### DIFF
--- a/.buildkite/pipeline_lib.sh
+++ b/.buildkite/pipeline_lib.sh
@@ -56,5 +56,7 @@ gen_bazel_test_steps() {
         echo "          automatic:"
         echo "            - exit_status: -1 # Agent was lost"
         echo "            - exit_status: 255 # Forced agent shutdown"
+	echo "            - exit_status: 1 # Test may be flaky or it just didn't pass"
+	echo "              limit: 2
     done
 }
--- a/.buildkite/pipeline_lib.sh
+++ b/.buildkite/pipeline_lib.sh
@@ -57,6 +57,6 @@ gen_bazel_test_steps() {
         echo "            - exit_status: -1 # Agent was lost"
         echo "            - exit_status: 255 # Forced agent shutdown"
 	echo "            - exit_status: 1 # Test may be flaky or it just didn't pass"
-	echo "              limit: 2
+	echo "              limit: 2"
     done
 }
--- a/.buildkite/pipeline_lib.sh
+++ b/.buildkite/pipeline_lib.sh
@@ -56,7 +56,7 @@ gen_bazel_test_steps() {
         echo "          automatic:"
         echo "            - exit_status: -1 # Agent was lost"
         echo "            - exit_status: 255 # Forced agent shutdown"
-	echo "            - exit_status: 1 # Test may be flaky or it just didn't pass"
+	echo "            - exit_status: 3 # Test may be flaky or it just didn't pass"
 	echo "              limit: 2"
     done
 }
--- a/.buildkite/pipeline_lib.sh
+++ b/.buildkite/pipeline_lib.sh
@@ -53,12 +53,12 @@ gen_bazel_test_steps() {
         echo "          - \"bazel-testlogs.tar.gz\""
         echo "        timeout_in_minutes: 20"
         echo "        retry:"
-	echo "          manual:"
-	echo "            permit_on_passed: true"
+        echo "          manual:"
+        echo "            permit_on_passed: true"
         echo "          automatic:"
         echo "            - exit_status: -1 # Agent was lost"
         echo "            - exit_status: 255 # Forced agent shutdown"
-	echo "            - exit_status: 3 # Test may be flaky or it just didn't pass"
-	echo "              limit: 2"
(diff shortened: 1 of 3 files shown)


matzf (approved):
:lgtm:

Makes sense, and the numbers look good. Should we update the "expected" performance for the CI runs to reflect this tuning?


> Make sure that the number of streams is likely to be a multiple of the number of cores so the cores are loaded evenly in spite of there being typically fewer streams than in real life. (TBD, learn the router's internal processors count and just use that).

Alternative approach could be to organize the test packet generation slightly differently, such that we don't pre-fill the flow ID in the "case" function. As far as I remember, the flow ID does not figure into any checksums etc., so with only knowing the offset of the field, the packet sender code could set the flow ID on every packet individually, allowing to use virtually unlimited number of flows.

> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4456#-:-NlD-UmAAengLsp63_KG:b-7sezhd)* status: 1 of 2 files reviewed, all discussions resolved

___
*[`router/config/config.go` line 80 at r1](https://reviewable.io/reviews/scionproto/scion/4456#-NlN5TbW1fj6DkST5IG7:-NlN5TbW1fj6DkST5IG8:b-y5o3zi) ([raw file](https://github.com/scionproto/scion/blob/f76ee3cc4a19effdfa01a6d7bba9a7ffb140577c/router/config/config.go#L80)):*
> ```Go
> 	// counterproductive. We get much better results by setting two cores aside for other go
> 	// routines (such as for input and output). It remains to be seen if even more cores need to be
> 	// set aside on large core-count systems.
> ```

Note: the benchmark setup intentionally generates a single type of traffic flow at any time. The router is only receiving and transmitting on a single interface. In more realistic scenarios where multiple interfaces experience some load, it could be possible that we should optimally set aside more threads for IO, e.g. one for each active interface or so. 
Perhaps we should later also add a traffic mix benchmark case to find this out and tune the performance also in a somewhat more realistic scenario.

jiceatscion (commented):
Yeah, the multiple prebuilt packets was a bit of a lazyarse solution, but even 840 pre-built packets haven't caused any trouble (it's still pocket-change in the end). If I put additional effort there, I'd rather have the test learn the processor-count from the router and create just that many streams (which will be a very small number).

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4456#-:-NlNB0wv9iCVHDuwt-6G:b-mesnsx)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @matzf)

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4456)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`router/config/config.go` line 80 at r1](https://reviewable.io/reviews/scionproto/scion/4456#-NlN5TbW1fj6DkST5IG7:-NlNCTmd2nhuZR0PbX-2:b7gcpg3) ([raw file](https://github.com/scionproto/scion/blob/f76ee3cc4a19effdfa01a6d7bba9a7ffb140577c/router/config/config.go#L80)):*

Note: the benchmark setup intentionally generates a single type of traffic flow at any time. The router is only receiving and transmitting on a single interface. In more realistic scenarios where multiple interfaces experience some load, it could be possible that we should optimally set aside more threads for IO, e.g. one for each active interface or so. 
Perhaps we should later also add a traffic mix benchmark case to find this out and tune the performance also in a somewhat more realistic scenario.
</blockquote></details>

Yes, you're right. More work needed in that area. Subsequent PR?

matzf (approved):
> Reviewed 2 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4456)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

___
*[`router/config/config.go` line 80 at r1](https://reviewable.io/reviews/scionproto/scion/4456#-NlN5TbW1fj6DkST5IG7:-NlRz7ne1htdtJtFA6yD:b-lnn24b) ([raw file](https://github.com/scionproto/scion/blob/f76ee3cc4a19effdfa01a6d7bba9a7ffb140577c/router/config/config.go#L80)):*

Yes, you're right. More work needed in that area. Subsequent PR?
</blockquote></details>

Yep :+1:

PR #4459, opened 2023-12-21
doc: update buildkite instance type

matzf:
Documenting that we've bumped the buildkite instance type from t3 to m6i.
The main reason for the switch is that the "burstable" nature of the t*-instances makes them extremely unsuitable for running benchmarks. The m6i are also much faster, but not much more expensive.
Also, the agent pool is now at most 6 agents. Spinning up 10 agents and having to warm up all the caches, and then scaling back in after a single build, seemed excessive.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4459)

### DIFF
--- a/doc/dev/testing/buildkite.rst
+++ b/doc/dev/testing/buildkite.rst
@@ -89,7 +89,8 @@ Cluster configuration
 
    The agent cluster is operated by the SCION Association, in the AWS account ``scion-association``.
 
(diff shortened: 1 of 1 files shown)


oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4459#-:-NmC94T3C54Gu-Ruapgx:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4468, opened 2024-02-19
testing: router benchmark, multiplex external br interfaces on one veth pair

jiceatscion:
This makes no difference, performance-wise as each test case uses exactly ones of the external interfaces. However, it reduces the infrastructure required to run benchmarks on real devices: minimum 4 switch ports instead of minimum 6.

- - -
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4468)


matzf (approved):
The representation of this interface grouping in topology information with this exclusive bit does not seem entirely straight forward when reading this. The alternatives do seem just as complex though.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4468#-:-NrZwpRd6oMglrMHdOYS:b-t1728j)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4470, opened 2024-02-21
cmd/scion: add one-hop support to ping

oncilla:
Add support to create one-hop paths in the ping command. This can be used to send SCION pings to direct neighbors over interfaces for which no path exists currently.


jiceatscion (approved):
Some non-binding suggestions only.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4470#-:-NrB454z4UlgZmexUula:b-xugwwh)* status: all files reviewed, 3 unresolved discussions (waiting on @oncilla)

___
*[`dispatcher/underlay.go` line 357 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrAytt6C_AJjw3d53Yb:-NrAytt6C_AJjw3d53Yc:b-w56qtn) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/dispatcher/underlay.go#L357)):*
> ```Go
> 		return serrors.WrapStr("reversing path", err)
> 	}
> 	pkt.SCION.PathType = pkt.SCION.Path.Type()
> ```

A bit obscure to me. Why would PathType need to change after reversal? Add a comment for the naive readers like me?
___
*[`private/keyconf/keyconf.go` line 58 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrB38dc4GURS1cWPRvq:-NrB38dc4GURS1cWPRvr:b-gynt08) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/private/keyconf/keyconf.go#L58)):*
> ```Go
> 	}
> 	dbuf = dbuf[:n]
> 	if strings.ToLower(algo) != RawKey {
> ```

Not that this is added by this PR, but...WTF ever was the purpose of the "algo" parameter? Since this was an unexported function, may be it the algo param could go and the function just just become exported. No?
___
*[`scion/cmd/scion/ping.go` line 173 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrB01n17j57HYL2r5CA:-NrB01n17j57HYL2r5CB:b-fwbn96) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L173)):*
> ```Go
> 				path.WithEPIC(flags.epic),
> 			}
> 			if flags.egress != 0 || flags.nextHop.IP != nil || flags.forwardingKey != "" {
> ```

I understood the intent but, along with the next three tests, it reads weird. Could it make more sense to check that all three conditions have the same truth value and, if not, report an error to that effect? Then use whatever that value is to decide on doing the rest. Might not be shorter but a bit more obvious.

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4470)* status: all files reviewed, 3 unresolved discussions (waiting on @jiceatscion)

___
*[`scion/cmd/scion/ping.go` line 173 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrB01n17j57HYL2r5CA:-NrBKLzf01TaweKTb-Bk:bg5q0dw) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L173)):*
> Could it make more sense to check that all three conditions have the same truth value and, if not, report an error to that effect? Then use whatever that value is to decide on doing the rest.

I'm not sure I understand exactly what you mean here.

The intention is the following:
- If any of these flags is set, we assume the user wants to run in one-hop mode.
- In one-hop mode, all the flags are required.

I guess I could change it to:

- If any of the flags is set, we assume the user wants to run in one-hop mode.
- Check if all are set, and if not, report the flags that are missing.

Is that what you had in mind?

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4470)* status: all files reviewed, 3 unresolved discussions (waiting on @oncilla)

___
*[`scion/cmd/scion/ping.go` line 173 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrB01n17j57HYL2r5CA:-NrBMIIH8RrZ9Q0b8mET:b-nzaigu) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L173)):*

> Could it make more sense to check that all three conditions have the same truth value and, if not, report an error to that effect? Then use whatever that value is to decide on doing the rest.

I'm not sure I understand exactly what you mean here.

The intention is the following:
- If any of these flags is set, we assume the user wants to run in one-hop mode.
- In one-hop mode, all the flags are required.

I guess I could change it to:

- If any of the flags is set, we assume the user wants to run in one-hop mode.
- Check if all are set, and if not, report the flags that are missing.

Is that what you had in mind?
</blockquote></details>

Ok, the intent is what I understood. What I had in mind was something like: if a and b and c { one_hop } elif !a and !b and !c { not_one_hop } else { error } ... May be all that's needed is what you propose, only replacing your three different errors with a single one; like: if (!b or !b or !c) { error("all three flags must be set, or none must be") }. It would make everything clearer; both to readers and users.

matzf (commented):
> Reviewed 2 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4470)* status: all files reviewed, 6 unresolved discussions (waiting on @jiceatscion and @oncilla)

___
*[`dispatcher/underlay.go` line 357 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrAytt6C_AJjw3d53Yb:-NrF0KlV6gSJamvm5CWJ:b-xmih4v) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/dispatcher/underlay.go#L357)):*

A bit obscure to me. Why would PathType need to change after reversal? Add a comment for the naive readers like me?
</blockquote></details>

It's because Reverse is (too) magic and can return a path of _different_ type.  In particular, reversing a one-hop path returns a SCION path. It wouldn't have to, but it does. It's sort of a convenience in higher levels of the inter-AS RPC stack.
___
*[`private/keyconf/keyconf.go` line 58 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrB38dc4GURS1cWPRvq:-NrEzbVp5zW9wZkYUt6r:b-ighbkr) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/private/keyconf/keyconf.go#L58)):*

Not that this is added by this PR, but...WTF ever was the purpose of the "algo" parameter? Since this was an unexported function, may be it the algo param could go and the function just just become exported. No?
</blockquote></details>

Historically, this function was also used to load elliptic curve keys, which now are all stored in PEM and use a different code path. I guess this algorithm could be removed indeed.
___
*[`scion/cmd/scion/ping.go` line 397 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrEv1TO-Kvnd-L9Y7X7:-NrEv1TO-Kvnd-L9Y7X8:b-ahdm84) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L397)):*
> ```Go
> 	cmd.Flags().StringVar(&flags.format, "format", "human",
> 		"Specify the output format (human|json|yaml)")
> 	cmd.Flags().Uint16Var(&flags.egress, "onehop.egress", 0, "Egress interface for one-hop path")
> ```

The onehop flags, in particular the egress parameter, is incompatible with any other path option flags, namely
- `--interactive`
- `--refresh`
- `--healthy-only`
- `--sequence`
- (and also `--epic`)

A somewhat smoother integration might be to add onehop as only a bool flag instead of specifying the egress interface. 
This could create the entire set of one hop paths to the neighboring AS (if there are any), among which the user can then choose with the existing options (interactive prompt, sequence match, arbitrary default).
As epic and onehop are mutually exclusive, perhaps these flags could even be combined in a single enum "path type" parameter.

If this seems too much additional work for only cosmetic benefit, I'd sueggest to mark the new onehop flags as experimental, so that we can tackle this later without too much friction.
___
*[`scion/cmd/scion/ping.go` line 398 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrExjU39OK6Eow7BFWj:-NrExjU39OK6Eow7BFWk:b-ou3759) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L398)):*
> ```Go
> 		"Specify the output format (human|json|yaml)")
> 	cmd.Flags().Uint16Var(&flags.egress, "onehop.egress", 0, "Egress interface for one-hop path")
> 	cmd.Flags().Var(&flags.nextHop, "onehop.next-hop", "NextHop for one-hop path")
> ```

The next hop can be inferred from the topology. The daemon's `Interfaces()` provides the corresponding mapping from the `egress` interface.
With this, we can make this parameter optional, or perhaps even remove it entirely -- for the other path types, we don't support overriding the next hop either.
___
*[`scion/cmd/scion/ping.go` line 400 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrEyTpi5-t6wNzFPBq1:-NrEyTpi5-t6wNzFPBq2:b-k2x7eg) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L400)):*
> ```Go
> 	cmd.Flags().Var(&flags.nextHop, "onehop.next-hop", "NextHop for one-hop path")
> 	cmd.Flags().StringVar(&flags.forwardingKey, "onehop.forwarding-key", "",
> 		"Forwarding key for one-hop path",
> ```

As we don't have a separate user manual for this tool, perhaps add a bit more information to this doc string here.



_Suggestion:_
```Go
	cmd.Flags().StringVar(&flags.forwardingKey, "onehop.forwarding-key", "master0.key",
		"Forwarding secret key file for one-hop path. Note that this is typically only available to the operator of the SCION AS.",
```

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4470)* status: all files reviewed, 6 unresolved discussions (waiting on @jiceatscion and @matzf)

___
*[`dispatcher/underlay.go` line 357 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrAytt6C_AJjw3d53Yb:-NrF7kbd9fjWJsiIK291:b-cx1zdp) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/dispatcher/underlay.go#L357)):*

It's because Reverse is (too) magic and can return a path of _different_ type.  In particular, reversing a one-hop path returns a SCION path. It wouldn't have to, but it does. It's sort of a convenience in higher levels of the inter-AS RPC stack.
</blockquote></details>

Agreed. The Reverse method is on the wrong level. On the level of the path we do not have enough info. This is also why the EPIC above needs special handling.

I will add a note here, but this is a more fundamental issue with the path itself.
___
*[`private/keyconf/keyconf.go` line 58 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrB38dc4GURS1cWPRvq:-NrF8TNd3khvySfuvdS6:b-jml0m6) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/private/keyconf/keyconf.go#L58)):*

Historically, this function was also used to load elliptic curve keys, which now are all stored in PEM and use a different code path. I guess this algorithm could be removed indeed.
</blockquote></details>

Yeah, we could also drop the master1.key :laughing: 

I think we can clean it up when we finally support forwarding key rollover.
___
*[`scion/cmd/scion/ping.go` line 397 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrEv1TO-Kvnd-L9Y7X7:-NrFAoj_8YGojXj4_hx3:bgbayax) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L397)):*

The onehop flags, in particular the egress parameter, is incompatible with any other path option flags, namely
- `--interactive`
- `--refresh`
- `--healthy-only`
- `--sequence`
- (and also `--epic`)

A somewhat smoother integration might be to add onehop as only a bool flag instead of specifying the egress interface. 
This could create the entire set of one hop paths to the neighboring AS (if there are any), among which the user can then choose with the existing options (interactive prompt, sequence match, arbitrary default).
As epic and onehop are mutually exclusive, perhaps these flags could even be combined in a single enum "path type" parameter.

If this seems too much additional work for only cosmetic benefit, I'd sueggest to mark the new onehop flags as experimental, so that we can tackle this later without too much friction.
</blockquote></details>

Ack, I can do some clean ups here.
___
*[`scion/cmd/scion/ping.go` line 398 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrExjU39OK6Eow7BFWj:-NrFACAe1nm4P8xPMphb:b-z9rlp8) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L398)):*

The next hop can be inferred from the topology. The daemon's `Interfaces()` provides the corresponding mapping from the `egress` interface.
With this, we can make this parameter optional, or perhaps even remove it entirely -- for the other path types, we don't support overriding the next hop either.
</blockquote></details>

That requires that the scion command has access to the topology file. I would prefer to not spread the topology file any further than what it has already infected.

For regular paths, the next hop is part of the response of the Daemon. We can extend the Daemon API to allow querying the next hop for a certain interface, but that seems like an overkill to me.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4470)* status: all files reviewed, 6 unresolved discussions (waiting on @jiceatscion and @oncilla)

___
*[`scion/cmd/scion/ping.go` line 398 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrExjU39OK6Eow7BFWj:-NrFW32KCoP-LEr-eZbI:b-aagtjc) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L398)):*

That requires that the scion command has access to the topology file. I would prefer to not spread the topology file any further than what it has already infected.

For regular paths, the next hop is part of the response of the Daemon. We can extend the Daemon API to allow querying the next hop for a certain interface, but that seems like an overkill to me.
</blockquote></details>

I fully agree about the topology file, accessing this directly would not be nice.

But the API in the daemon already exists: `pkg/daemon.Connector.Interfaces(ctx context.Context) (map[uint16]*net.UDPAddr, error)`.

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4470)* status: all files reviewed, 6 unresolved discussions (waiting on @jiceatscion and @matzf)

___
*[`scion/cmd/scion/ping.go` line 398 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrExjU39OK6Eow7BFWj:-NrFnV-c2Ju2Y0Hxv-Am:buh0enl) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L398)):*

I fully agree about the topology file, accessing this directly would not be nice.

But the API in the daemon already exists: `pkg/daemon.Connector.Interfaces(ctx context.Context) (map[uint16]*net.UDPAddr, error)`.
</blockquote></details>

Oh wow, I forgot that even existed :100: 

Will do!

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4470)* status: all files reviewed, 4 unresolved discussions (waiting on @matzf and @oncilla)

___
*[`private/keyconf/keyconf.go` line 58 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrB38dc4GURS1cWPRvq:-NrZnUogDrzZqgEb_iY6:b-ia70ng) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/private/keyconf/keyconf.go#L58)):*

Yeah, we could also drop the master1.key :laughing: 

I think we can clean it up when we finally support forwarding key rollover.
</blockquote></details>

Add a todo/file a bug, so it gets done one day?
___
*[`scion/cmd/scion/ping.go` line 173 at r1](https://reviewable.io/reviews/scionproto/scion/4470#-NrB01n17j57HYL2r5CA:-NrZno3PCI6q9FZSiW2v:bqhccsr) ([raw file](https://github.com/scionproto/scion/blob/be5bb988a076ace0f5f34ed66586add9a2437e98/scion/cmd/scion/ping.go#L173)):*

Ok, the intent is what I understood. What I had in mind was something like: if a and b and c { one_hop } elif !a and !b and !c { not_one_hop } else { error } ... May be all that's needed is what you propose, only replacing your three different errors with a single one; like: if (!b or !b or !c) { error("all three flags must be set, or none must be") }. It would make everything clearer; both to readers and users.
</blockquote></details>

We understood each-other. Not blocking anyway.

jiceatscion (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4470#-:-NrZoXX53OtvoV5_jI7C:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @matzf and @oncilla)

PR #4472, opened 2024-02-26
testing: Make router_benchmark use a less popular address block.

jiceatscion:
In case one wants to run the benchmark outside containers, 192.168.[1-10].* can be inconvenient as it is a very popular address range. The surrounding network (or the operator's workstation itself), may well be using some of it.
Ultimately the non-containerized work-flow will need to become more real-world friendly, but this will help in the short term.


matzf (approved):
:lgtm:

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4472#-:-Nr_UabF9Ad5OiGdhhHy:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4478, opened 2024-03-05
Move /share/conf to /etc/scion and rename br.toml to router.toml.

jiceatscion:
Motivation: /share is not standard for config files and is only used in test containers. The packages we produce have the config files in /etc/scion, which is more in line with unix practice. Using the same path names makes it easier to reuse some test config files as-is on real devices. For similar reasons, Renamed br.toml to router.toml; the name we use on openwrt. It makes it easier to generate the corresponding config as an openwrt package and otherwise makes no difference. In passing, delete /share/crypto, which is no-longer used at all.


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4478)

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4478)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lukedirtwalker)

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4478)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4479, opened 2024-03-07
testing: simplify the process of configuring routers, brload for benchmarking

jiceatscion:
Summary of changes:

* Package the benchmark router configuration as an openwrt package.
* Add the global BFD config to the router config schema. Retire the corresponding env vars.
* Make brload open the host interfaces that it is configured with regardless of their names; as opposed to all interfaces named veth_\*_host
* No-longer pass MAC addresses on the brload command line (one still optional - used for CI). brload finds them.
* No-longer require static arp entries on the router's side. brload responds to ARP.


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4479)

matzf (changes requested):
> Reviewed 8 of 20 files at r1, 5 of 7 files at r3, 2 of 14 files at r4.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4479)* status: 7 of 28 files reviewed, 2 unresolved discussions (waiting on @jiceatscion)

___
*[`acceptance/router_benchmark/test.py` line 348 at r6](https://reviewable.io/reviews/scionproto/scion/4479#-NtVaNu928y6zhzPOvY7:-NtVaNu928y6zhzPOvY8:b-95mc91) ([raw file](https://github.com/scionproto/scion/blob/6a23f19d0bc5a465767ed673940c09403a784c88/acceptance/router_benchmark/test.py#L348)):*
> ```Python
>         sudo("ip", "netns", "exec", ns,
>              "ip", "addr", "add", f"{req.ip}/{req.prefixLen}", "broadcast",
>              bcast_for_ip(req.ip, int(req.prefixLen)), "dev", brIntf)
> ```

Use `ipaddress` standard library instead: `ipaddress.ip_network(f"{req.ip}/{req.prefixLen}", strict=False).broadcast_address`
___
*[`doc/manuals/router.rst` line 185 at r4](https://reviewable.io/reviews/scionproto/scion/4479#-NtGc4ii5q6vRs6p251z:-NtGc4ii5q6vRs6p252-:b-qowgte) ([raw file](https://github.com/scionproto/scion/blob/655ecd86905957f412c1910b14b74755a467fe8c/doc/manuals/router.rst#L185)):*
> ```reStructuredText
> 
>    .. object:: bfd
>    
> ```

Whitespace: here trailing spaces. Below, there are some tabs that should be spaces instead.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4479)* status: 7 of 28 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`acceptance/router_benchmark/test.py` line 348 at r6](https://reviewable.io/reviews/scionproto/scion/4479#-NtVaNu928y6zhzPOvY7:-NtVep9h8RTxDveWwmAe:bg0k5do) ([raw file](https://github.com/scionproto/scion/blob/6a23f19d0bc5a465767ed673940c09403a784c88/acceptance/router_benchmark/test.py#L348)):*

Use `ipaddress` standard library instead: `ipaddress.ip_network(f"{req.ip}/{req.prefixLen}", strict=False).broadcast_address`
</blockquote></details>

Gladly. Done
___
*[`doc/manuals/router.rst` line 185 at r4](https://reviewable.io/reviews/scionproto/scion/4479#-NtGc4ii5q6vRs6p251z:-NtVfw588dpZDKAV3jMd:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/655ecd86905957f412c1910b14b74755a467fe8c/doc/manuals/router.rst#L185)):*

Whitespace: here trailing spaces. Below, there are some tabs that should be spaces instead.
</blockquote></details>

done

matzf (commented):
> Reviewed 1 of 7 files at r3, 3 of 14 files at r4, 1 of 2 files at r5, 14 of 15 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4479)* status: all files reviewed, 8 unresolved discussions (waiting on @jiceatscion)

___
*[`acceptance/router_benchmark/cases/topo.go` line 283 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtocGcW4hjRVA1WcuDi:-NtocGcW4hjRVA1WcuDj:bded0uq) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/acceptance/router_benchmark/cases/topo.go#L283)):*
> ```Go
> 	}
> 	as4 := ip.As4()
> 	return net.HardwareAddr{0xf0, 0x0d, 0xca, 0xfe, as4[2], as4[3]}
> ```

I don't know if I'm confusing things now, I thought this was intentionally different from the addresses generated in the test script, to avoid they'd accidentally match.
As a quick test, I replaced this with a panic and the tests did not seem to require this code path. Perhaps panic-ing would be clearer here?
___
*[`doc/manuals/common.rst` line 299 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-Ntou-8w4udU19QjdRMT:-Ntou-8w4udU19QjdRMU:b-2r2m1g) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/doc/manuals/common.rst#L299)):*
> ```reStructuredText
>             Disable BFD, unconditionally consider the connection alive.
> 
>             If true, takes precedence over :option:`router.bfd.disabled <router-conf-toml disabled>`.
> ```

Non-blocking: to make the prcedence behavior more useful, I think we could relatively easily keep track of whether the `disable` is overridden in the topology file, by changing the type of `private/topology/json.BFD.Disable` from `bool` to `*bool`. This is a rather clumsy tri-state, but the json parser handles this and it would mostly just work. 
We'd need to adapt the types in a few more places, and some of the logging statements become a bit awkward, as we need to negate a boolean pointer to log "bfd_enabled". 
Can also defer this to a separate PR though.
___
*[`router/connector.go` line 44 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtoYXAC5QXqMuIOe2HB:-NtoYXAC5QXqMuIOe2HC:bqdldos) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/connector.go#L44)):*
> ```Go
> 	ReceiveBufferSize int
> 	SendBufferSize    int
> 	BfdConfig         control.BFD
> ```

Nit: `BFDConfig` (or just `BFD` as in LinkInfo?).
___
*[`router/connector.go` line 214 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtosdmU2NbEd1getMwQ:-NtosdmU2NbEd1getMwR:b-fe4w1m) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/connector.go#L214)):*
> ```Go
> // setting prevails if it is true. (ie. bfd is disabled as soon as either the link or the global
> // setting says so).
> func (c *Connector) applyBFDdefaults(cfg control.BFD) control.BFD {
> ```

nit `applyBFDDefaults`.
___
*[`router/dataplane.go` line 242 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtoqRi7EUiOebQBAeSi:-NtoqRi7EUiOebQBAeSj:b-owvh9e) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/dataplane.go#L242)):*
> ```Go
> 	err := d.addExternalInterfaceBFD(ifID, conn, src, dst, cfg)
> 	if err != nil {
> 		return serrors.WrapStr("adding external BFD", err, "if_id", ifID)
> ```

Nit: `ifID` 
___
*[`router/dataplane.go` line 250 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtoqepmBSBJ-v0nEzm1:-NtoqepmBSBJ-v0nEzm2:bqye51o) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/dataplane.go#L250)):*
> ```Go
> 		d.interfaces = make(map[uint16]BatchConn)
> 	}
> 	if _, exists := d.external[ifID]; exists {
> ```

Shouldn't this be checked before calling `addExternalInterfaceBFD`?
___
*[`router/config/config.go` line 51 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtoZgRb2CdI-MzG7_Ib:-NtoZgRb2CdI-MzG7_Ic:b-pc5vi2) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/config/config.go#L51)):*
> ```Go
> 	NumSlowPathProcessors int      `toml:"num_slow_processors,omitempty"`
> 	BatchSize             int      `toml:"batch_size,omitempty"`
> 	Bfd                   json.BFD `toml:"bfd,omitempty"`
> ```

Nit: BFD
___
*[`router/config/config.go` line 51 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-Ntojw_bBVvgMX90XUbk:-Ntojw_bBVvgMX90XUbl:b-tuhze7) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/config/config.go#L51)):*
> ```Go
> 	NumSlowPathProcessors int      `toml:"num_slow_processors,omitempty"`
> 	BatchSize             int      `toml:"batch_size,omitempty"`
> 	Bfd                   json.BFD `toml:"bfd,omitempty"`
> ```

Using the `json.BFD` struct type does not work well here. The names of the fields are not defined for the toml reader, so that the fields have  "non-compliant" names like `desiredMinTxInterval`.
Best approach is probably to be define a separate struct here with `toml` tags. This also allows to add a separate `Validate`, `InitDefaults` and `Sample` implementation.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4479)* status: all files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`acceptance/router_benchmark/cases/topo.go` line 283 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtocGcW4hjRVA1WcuDi:-NtzKXaf--sbp74BjO5K:b-q1v0wa) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/acceptance/router_benchmark/cases/topo.go#L283)):*

I don't know if I'm confusing things now, I thought this was intentionally different from the addresses generated in the test script, to avoid they'd accidentally match.
As a quick test, I replaced this with a panic and the tests did not seem to require this code path. Perhaps panic-ing would be clearer here?
</blockquote></details>

No, in the end this does not need to be different from what the test script uses. It is only used for fictional hosts, that is those that are not simulated by brload, nor executing the router (i.e. they only exist in the topology). It happens to not be needed by any test case, but I do not want to create an unnecessary restriction. How would I describe that? "This module provides information about hosts in the topology file, but only for br1a and its immediate neighbors?"

I have added a comment in the code to explain that.
___
*[`doc/manuals/common.rst` line 299 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-Ntou-8w4udU19QjdRMT:-NtzdUFKCy2KNBdRXth0:buq9pgj) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/doc/manuals/common.rst#L299)):*

Non-blocking: to make the prcedence behavior more useful, I think we could relatively easily keep track of whether the `disable` is overridden in the topology file, by changing the type of `private/topology/json.BFD.Disable` from `bool` to `*bool`. This is a rather clumsy tri-state, but the json parser handles this and it would mostly just work. 
We'd need to adapt the types in a few more places, and some of the logging statements become a bit awkward, as we need to negate a boolean pointer to log "bfd_enabled". 
Can also defer this to a separate PR though.
</blockquote></details>

Nah, I did it since there was a lot of cleanup around that area to do anyway. I had completely overlooked the json/toml discrepancy. Sucks.
___
*[`router/connector.go` line 44 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtoYXAC5QXqMuIOe2HB:-NtzdoU1A4L8IM7YJJ18:b-ffmzvk) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/connector.go#L44)):*

Nit: `BFDConfig` (or just `BFD` as in LinkInfo?).
</blockquote></details>

The latter. Done
___
*[`router/connector.go` line 214 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtosdmU2NbEd1getMwQ:-Ntzfa-9-5EIIuyBlsPk:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/connector.go#L214)):*

nit `applyBFDDefaults`.
</blockquote></details>

done
___
*[`router/dataplane.go` line 242 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtoqRi7EUiOebQBAeSi:-Ntzfy-R3EwMnItBXN4e:b-9xqxhq) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/dataplane.go#L242)):*

Nit: `ifID`
</blockquote></details>

Won't do. This follows the existing practice in all of dataplane.go. ... unless you want me to change all log messages to use camel case for labels.
___
*[`router/dataplane.go` line 250 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtoqepmBSBJ-v0nEzm1:-NtzgZM3DXYLRVzmL3Pk:b-gfx5t7) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/dataplane.go#L250)):*

Shouldn't this be checked before calling `addExternalInterfaceBFD`?
</blockquote></details>

I didn't change that. I just moved to *after* checking if d.external is nil :-) So, the check for d.external would need to move with it. The tests expect the check to be done here. It is considered an exceptional situation, not silent precaution. I think this is fine.
___
*[`router/config/config.go` line 51 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtoZgRb2CdI-MzG7_Ib:-NtzhprkEp1IHWsJQ-LI:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/config/config.go#L51)):*

Nit: BFD
</blockquote></details>

done
___
*[`router/config/config.go` line 51 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-Ntojw_bBVvgMX90XUbk:-NtzhryA0t71gnRuvkwF:b-rh5rjx) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/config/config.go#L51)):*

Using the `json.BFD` struct type does not work well here. The names of the fields are not defined for the toml reader, so that the fields have  "non-compliant" names like `desiredMinTxInterval`.
Best approach is probably to be define a separate struct here with `toml` tags. This also allows to add a separate `Validate`, `InitDefaults` and `Sample` implementation.
</blockquote></details>

done
...sadly. What was the deal with non-compliant names? I don't see any difference.

matzf (changes requested):
> Reviewed 14 of 14 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4479)* status: all files reviewed, 3 unresolved discussions (waiting on @jiceatscion)

___
*[`doc/manuals/common.rst` line 295 at r8](https://reviewable.io/reviews/scionproto/scion/4479#-Nu3EKMWExmdrrWvCbN6:-Nu3EKMWExmdrrWvCbN7:b-wn79rw) ([raw file](https://github.com/scionproto/scion/blob/7a152f7cdfad84d91f6e9e2e6436a5572bc7c2ab/doc/manuals/common.rst#L295)):*
> ```reStructuredText
>          :option:`<router-id> <topology-json <router-id>>`.
> 
>          .. option:: disable = <bool>, default :option:`router.bfd.disabled <router-conf-toml disabled>`
> ```

This is one of the places where RST/sphinx breaks down; the option "headline" here is not styled, unfortunately this is rendered verbatim. We need to keep the link in the body, below.
___
*[`router/dataplane.go` line 242 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtoqRi7EUiOebQBAeSi:-Nu3HkHY5duJteEmorNR:b-k7fx64) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/dataplane.go#L242)):*

Won't do. This follows the existing practice in all of dataplane.go. ... unless you want me to change all log messages to use camel case for labels.
</blockquote></details>

You're right, sorry!
___
*[`router/dataplane_test.go` line 122 at r8](https://reviewable.io/reviews/scionproto/scion/4479#-Nu3KLBF8d8qQa5eeMsq:-Nu3KLBF8d8qQa5eeMsr:b-2yjqye) ([raw file](https://github.com/scionproto/scion/blob/7a152f7cdfad84d91f6e9e2e6436a5572bc7c2ab/router/dataplane_test.go#L122)):*
> ```Go
> 	}
> 	disable := true
> 	nobfd := control.BFD{Disable: &disable}
> ```

Maybe we could add something like this to `xtest/helpers.go`:
```go
func PointerTo[T any](v T) *T {
    return &v
}
```
___
*[`router/config/config.go` line 51 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-Ntojw_bBVvgMX90XUbk:-Nu3LFvuEZ-TyacxxzqW:b-ky33wt) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/router/config/config.go#L51)):*
> What was the deal with non-compliant names? I don't see any difference.

Ah, I just meant to say that without the `toml` tags, the toml parser was looking for the wrong field names. All good now.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4479)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`acceptance/router_benchmark/cases/topo.go` line 283 at r7](https://reviewable.io/reviews/scionproto/scion/4479#-NtocGcW4hjRVA1WcuDi:-Nu3R_xFD04tITIfODJ4:bkr28tp) ([raw file](https://github.com/scionproto/scion/blob/3615cea24361da581e593c27266e5316fd6522ce/acceptance/router_benchmark/cases/topo.go#L283)):*

No, in the end this does not need to be different from what the test script uses. It is only used for fictional hosts, that is those that are not simulated by brload, nor executing the router (i.e. they only exist in the topology). It happens to not be needed by any test case, but I do not want to create an unnecessary restriction. How would I describe that? "This module provides information about hosts in the topology file, but only for br1a and its immediate neighbors?"

I have added a comment in the code to explain that.
</blockquote></details>

...and I also reverted to 0xdeadbeef. Because it makes it more obvious which entity the address was assigned by. As we discussed.
___
*[`doc/manuals/common.rst` line 295 at r8](https://reviewable.io/reviews/scionproto/scion/4479#-Nu3EKMWExmdrrWvCbN6:-Nu3g2bU9rvuEVBqCU1R:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/7a152f7cdfad84d91f6e9e2e6436a5572bc7c2ab/doc/manuals/common.rst#L295)):*

This is one of the places where RST/sphinx breaks down; the option "headline" here is not styled, unfortunately this is rendered verbatim. We need to keep the link in the body, below.
</blockquote></details>

Done.
___
*[`router/dataplane_test.go` line 122 at r8](https://reviewable.io/reviews/scionproto/scion/4479#-Nu3KLBF8d8qQa5eeMsq:-Nu3b1iu-YRoMZ0P_J0c:b47pla0) ([raw file](https://github.com/scionproto/scion/blob/7a152f7cdfad84d91f6e9e2e6436a5572bc7c2ab/router/dataplane_test.go#L122)):*

Maybe we could add something like this to `xtest/helpers.go`:
```go
func PointerTo[T any](v T) *T {
    return &v
}
```
</blockquote></details>

Done. Nice one.

matzf (changes requested):
> Reviewed 8 of 8 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4479)* status: all files reviewed, 5 unresolved discussions (waiting on @jiceatscion)

___
*[`control/beacon/policy.go` line 24 at r9](https://reviewable.io/reviews/scionproto/scion/4479#-Nu49-JfDoEEqf_9_SVW:-Nu49-JfDoEEqf_9_SVX:brlhaf6) ([raw file](https://github.com/scionproto/scion/blob/1abedb9390e656c953e2ba224a5955d4d90266b4/control/beacon/policy.go#L24)):*
> ```Go
> 	"github.com/scionproto/scion/pkg/addr"
> 	"github.com/scionproto/scion/pkg/private/serrors"
> 	"github.com/scionproto/scion/pkg/private/xtest/ptr"
> ```

I think we should not use xtest/... packages in non-test code.
For now I would just revert the change in this file.
If there is ever a bigger use of this pattern, we can "promote" `pkg/private/xtest/ptr` to `pkg/private/ptr`.
___
*[`doc/manuals/common.rst` line 295 at r8](https://reviewable.io/reviews/scionproto/scion/4479#-Nu3EKMWExmdrrWvCbN6:-Nu4Alav6ilz62Yew1Rh:b19lsh7) ([raw file](https://github.com/scionproto/scion/blob/7a152f7cdfad84d91f6e9e2e6436a5572bc7c2ab/doc/manuals/common.rst#L295)):*

Done.
</blockquote></details>

There needs to be an empty line before `See :option: ...`. 
___
*[`doc/manuals/common.rst` line 295 at r9](https://reviewable.io/reviews/scionproto/scion/4479#-Nu4AySL7ThZ_HAOTvFw:-Nu4AySL7ThZ_HAOTvFx:undefined) ([raw file](https://github.com/scionproto/scion/blob/1abedb9390e656c953e2ba224a5955d4d90266b4/doc/manuals/common.rst#L295)):*
> ```reStructuredText
>          :option:`<router-id> <topology-json <router-id>>`.
> 
>          .. option:: disable = <bool>, default the router.bfd.disable.
> ```





_Suggestion:_
```reStructuredText
router.bfd.disable
```
___
*[`pkg/private/xtest/ptr/ptr.go` line 1 at r9](https://reviewable.io/reviews/scionproto/scion/4479#-Nu49HZo2CpONoSc3tGg:-Nu49HZo2CpONoSc3tGh:b-o9g1uh) ([raw file](https://github.com/scionproto/scion/blob/1abedb9390e656c953e2ba224a5955d4d90266b4/pkg/private/xtest/ptr/ptr.go#L1)):*
> ```Go
> package ptr
> ```

Missing license text.
___
*[`pkg/private/xtest/ptr/ptr.go` line 3 at r9](https://reviewable.io/reviews/scionproto/scion/4479#-Nu49Jqj1lMRWETcjpzc:-Nu49Jqk21jeYtEdqQ5-:bdszonx) ([raw file](https://github.com/scionproto/scion/blob/1abedb9390e656c953e2ba224a5955d4d90266b4/pkg/private/xtest/ptr/ptr.go#L3)):*
> ```Go
> package ptr
> 
> // PointerTo generates a pointer to a variable containing the given value.
> ```

Nit: To ...

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4479)* status: 28 of 38 files reviewed, 4 unresolved discussions (waiting on @matzf)

___
*[`control/beacon/policy.go` line 24 at r9](https://reviewable.io/reviews/scionproto/scion/4479#-Nu49-JfDoEEqf_9_SVW:-Nu4uS213VmEdPLvvyFs:b-2olo0s) ([raw file](https://github.com/scionproto/scion/blob/1abedb9390e656c953e2ba224a5955d4d90266b4/control/beacon/policy.go#L24)):*

I think we should not use xtest/... packages in non-test code.
For now I would just revert the change in this file.
If there is ever a bigger use of this pattern, we can "promote" `pkg/private/xtest/ptr` to `pkg/private/ptr`.
</blockquote></details>

Promoted ptr out of xtest. Just as easy.
___
*[`doc/manuals/common.rst` line 295 at r8](https://reviewable.io/reviews/scionproto/scion/4479#-Nu3EKMWExmdrrWvCbN6:-Nu4uHcE6yUa5mGTNKKl:bb7f2mf) ([raw file](https://github.com/scionproto/scion/blob/7a152f7cdfad84d91f6e9e2e6436a5572bc7c2ab/doc/manuals/common.rst#L295)):*

There needs to be an empty line before `See :option: ...`.
</blockquote></details>

done
___
*[`doc/manuals/common.rst` line 295 at r9](https://reviewable.io/reviews/scionproto/scion/4479#-Nu4AySL7ThZ_HAOTvFw:-Nu4u_m_0jkhFxL-vklg:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/1abedb9390e656c953e2ba224a5955d4d90266b4/doc/manuals/common.rst#L295)):*
> ```reStructuredText
>          :option:`<router-id> <topology-json <router-id>>`.
> 
>          .. option:: disable = <bool>, default the router.bfd.disable.
> ```

done
___
*[`pkg/private/xtest/ptr/ptr.go` line 1 at r9](https://reviewable.io/reviews/scionproto/scion/4479#-Nu49HZo2CpONoSc3tGg:-Nu4tocKA5e-LFchQUaF:b-896fjx) ([raw file](https://github.com/scionproto/scion/blob/1abedb9390e656c953e2ba224a5955d4d90266b4/pkg/private/xtest/ptr/ptr.go#L1)):*

Missing license text.
</blockquote></details>

Done
___
*[`pkg/private/xtest/ptr/ptr.go` line 3 at r9](https://reviewable.io/reviews/scionproto/scion/4479#-Nu49Jqj1lMRWETcjpzc:-Nu4tnfG6yF24BX_vbEV:b-896fjx) ([raw file](https://github.com/scionproto/scion/blob/1abedb9390e656c953e2ba224a5955d4d90266b4/pkg/private/xtest/ptr/ptr.go#L3)):*

Nit: To ...
</blockquote></details>

Done

matzf (approved):
> Reviewed 10 of 10 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4479)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4483, opened 2024-03-12
paths: Add check for hopfield count <= 64 when deserializing a scion path

### LINKED ISSUE #4482: Question about `router`: bound-checking for hop fields possibly missing
Hi everyone,

First, I apologize if this report/question is a bit too hand-wavy at times, but while proving the correctness of the router in the context of VerifiedSCION, I noticed that something looks weird.

More concretely, in file `pkg/slayers/path/scion/decoded.go`, there are the following definitions, together with comments expressing a few assumptions about the structure of the scion packets:
```go
const (
	// MaxINFs is the maximum number of info fields in a SCION path.
	MaxINFs = 3
	// MaxHops is the maximum number of hop fields in a SCION path.
	MaxHops = 64
)
```
While the restriction on `MaxINFs` looks OK, the restriction that we have at most 64 hop fields seems to not be enforced: when we decode a `Base`, we stored in the field `NumHops` the sum of the lengths of all segments. The length of each segment cannot exceed 64, but their sums can. As far as I can tell, there is no check performed while forwarding a packet that `NumINFs` is less than 64. While processing a SCION packet, we eventually call `IncPath`, which has the effect of incrementing the field `CurrHF` in `PathMeta`. As far as I can tell, nothing so far prohibits this field from getting to value 64 (if its original value was 63). The problem here is that, when we serialize the output packet (including its packet), if its value is at least 64, we will only serialize its 6 least significant bits, instead of storing its full value (the operation that serializes the `CurrHF` is `uint32(m.CurrHF&0x3F)<<24`). As such, I believe that, as is, the router would allow sending packets whose paths are too long until a point when they become invalid and are dropped. Is my understanding correct?

jiceatscion:
Added unit test for deserialization and unit test for router ingress.

Fixes #4482

### DIFF
--- a/pkg/slayers/path/scion/BUILD.bazel
+++ b/pkg/slayers/path/scion/BUILD.bazel
@@ -24,6 +24,7 @@ go_test(
     ],
     deps = [
         ":go_default_library",
+        "//pkg/private/serrors:go_default_library",
         "//pkg/slayers/path:go_default_library",
         "@com_github_stretchr_testify//assert:go_default_library",
         "@com_github_stretchr_testify//require:go_default_library",
--- a/pkg/slayers/path/scion/base.go
+++ b/pkg/slayers/path/scion/base.go
@@ -66,13 +66,6 @@ func (s *Base) DecodeFromBytes(data []byte) error {
 		}
 		s.NumHops += int(s.PathMeta.SegLen[i])
 	}
-	// We must check the validity of NumHops. It is possible to fit more than 64 hops in
-	// the length of a scion header. Yet a path of more than 64 hops cannot be followed to
-	// the end because CurrHF is only 6 bits long.
-	if s.NumHops > 64 {
-		return serrors.New(
-			fmt.Sprintf("NumHops > 64"))
-	}
 	return nil
 }
 
--- a/pkg/slayers/path/scion/base.go
+++ b/pkg/slayers/path/scion/base.go
@@ -22,10 +22,17 @@ import (
 	"github.com/scionproto/scion/pkg/slayers/path"
 )
 
-// MetaLen is the length of the PathMetaHeader.
-const MetaLen = 4
+const (
+	// MaxINFs is the maximum number of info fields in a SCION path.
+	MaxINFs = 3
+	// MaxHops is the maximum number of hop fields in a SCION path.
+	MaxHops = 64
 
-const PathType path.Type = 1
+	// MetaLen is the length of the PathMetaHeader.
+	MetaLen = 4
+
+	PathType path.Type = 1
+)
 
 func RegisterPath() {
 	path.RegisterPath(path.Metadata{
@@ -66,6 +73,14 @@ func (s *Base) DecodeFromBytes(data []byte) error {
 		}
 		s.NumHops += int(s.PathMeta.SegLen[i])
 	}
+
+	// We must check the validity of NumHops. It is possible to fit more than 64 hops in
+	// the length of a scion header. Yet a path of more than 64 hops cannot be followed to
+	// the end because CurrHF is only 6 bits long.
+	if s.NumHops > MaxHops {
+		return serrors.New("NumHops too large", "NumHops", s.NumHops, "Maximum", MaxHops)
+	}
+
 	return nil
 }
 
--- a/pkg/slayers/path/scion/base.go
+++ b/pkg/slayers/path/scion/base.go
@@ -66,6 +66,13 @@ func (s *Base) DecodeFromBytes(data []byte) error {
 		}
 		s.NumHops += int(s.PathMeta.SegLen[i])
 	}
+	// We must check the validity of NumHops. It is possible to fit more than 64 hops in
+	// the length of a scion header. Yet a path of more than 64 hops cannot be followed to
+	// the end because CurrHF is only 6 bits long.
+	if s.NumHops > 64 {
+		return serrors.New(
+			fmt.Sprintf("NumHops > 64"))
+	}
 	return nil
 }
 
@@ -113,7 +120,9 @@ func (s *Base) infIndexForHF(hf uint8) uint8 {
 	}
 }
 
-// Len returns the length of the path in bytes.
+// Len returns the length of the path in bytes. That is, the number of byte required to
+// store it, based on the metadata. The actual number of bytes available to contain it
+// can be inferred from the common header field HdrLen. It may or may not be consistent.
 func (s *Base) Len() int {
 	return MetaLen + s.NumINF*path.InfoLen + s.NumHops*path.HopLen
 }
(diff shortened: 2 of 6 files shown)


matzf (changes requested):
> Reviewed 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4483)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @jiceatscion)

___
*[`pkg/slayers/path/scion/decoded.go` line 52 at r2](https://reviewable.io/reviews/scionproto/scion/4483#-Nsx9i3p8V7HPZLvJtka:-Nsx9i3p8V7HPZLvJtkb:bf1788t) ([raw file](https://github.com/scionproto/scion/blob/861721c3367c61bd39e992c946e2fd8d448e43b4/pkg/slayers/path/scion/decoded.go#L52)):*
> ```Go
> 	// the length of a scion header. Yet a path of more than 64 hops cannot be followed to
> 	// the end because CurrHF is only 6 bits long.
> 	if s.NumHops > MaxHops {
> ```

This check now only applies to the `Decoded` variant of the SCION path representation, but we mainly use the `Raw` one e.g. in the router. The cleanest way to check in both cases, is to move this (back) to  `Base`.
___
*[`router/dataplane_test.go` line 646 at r2](https://reviewable.io/reviews/scionproto/scion/4483#-NsxDVtb5HIh1Bxf6GdK:-NsxDVtb5HIh1Bxf6GdL:b-6sumc6) ([raw file](https://github.com/scionproto/scion/blob/861721c3367c61bd39e992c946e2fd8d448e43b4/router/dataplane_test.go#L646)):*
> ```Go
> 				dpath.Base.PathMeta.SegLen = [3]uint8{24, 24, 17}
> 				dpath.Base.NumINF = 3
> 				dpath.Base.NumHops = 65
> ```

As discussed, this seems to result in a broken path (possibly because only one info field is written).

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4483)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`pkg/slayers/path/scion/decoded.go` line 52 at r2](https://reviewable.io/reviews/scionproto/scion/4483#-Nsx9i3p8V7HPZLvJtka:-Nsxez7l-pqQq2BRHTWD:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/861721c3367c61bd39e992c946e2fd8d448e43b4/pkg/slayers/path/scion/decoded.go#L52)):*

This check now only applies to the `Decoded` variant of the SCION path representation, but we mainly use the `Raw` one e.g. in the router. The cleanest way to check in both cases, is to move this (back) to  `Base`.
</blockquote></details>

Done.
___
*[`router/dataplane_test.go` line 646 at r2](https://reviewable.io/reviews/scionproto/scion/4483#-NsxDVtb5HIh1Bxf6GdK:-Nsxf0of-G1sMyDTR85t:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/861721c3367c61bd39e992c946e2fd8d448e43b4/router/dataplane_test.go#L646)):*

As discussed, this seems to result in a broken path (possibly because only one info field is written).
</blockquote></details>

done

matzf (approved):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4483)* status: all files reviewed, 2 unresolved discussions (waiting on @jiceatscion)

___
*[`pkg/slayers/path/scion/raw_test.go` line 91 at r3](https://reviewable.io/reviews/scionproto/scion/4483#-NtFkU0U7BkmtYQRb8Ob:-NtFkU0U7BkmtYQRb8Oc:b-pvuplz) ([raw file](https://github.com/scionproto/scion/blob/2c63d759221269b15c871bfa0b71a9c87994434f/pkg/slayers/path/scion/raw_test.go#L91)):*
> ```Go
> 	assert.NoError(t, overlongPath.SerializeTo(b)) // permitted, if only to enable this test.
> 	s := &scion.Raw{}
> 	assert.Error(t, s.DecodeFromBytes(b)) // invalid raw packet.
> ```

Nit: Maybe also add the `assert.ErrorIs` check here to ensure we get the expected error and that the path is not broken in some other way.
___
*[`router/dataplane_test.go` line 670 at r3](https://reviewable.io/reviews/scionproto/scion/4483#-NtFjKRoE1tZyLUVIpft:-NtFjKRpCTXFX_ZRbkeV:b-6p9u3e) ([raw file](https://github.com/scionproto/scion/blob/2c63d759221269b15c871bfa0b71a9c87994434f/router/dataplane_test.go#L670)):*
> ```Go
> 				expected := serrors.New("NumHops too large",
> 					"NumHops", 65, "Maximum", scion.MaxHops).Error()
> 				return assert.Equal(t, expected, err.Error())
> ```

Nit:  "serrors" can be compared with `Is`, and so `assert.ErrorIs(t, serrors.New("...",...), err)` can be used to shorten this.



_Suggestion:_
```Go
				expected := serrors.New("NumHops too large",
					"NumHops", 65, "Maximum", scion.MaxHops)
				return assert.ErrorIs(t, expected, err)
```

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4483)* status: 3 of 6 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`pkg/slayers/path/scion/raw_test.go` line 91 at r3](https://reviewable.io/reviews/scionproto/scion/4483#-NtFkU0U7BkmtYQRb8Ob:-NtGNhD_C4AC06FKR-59:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/2c63d759221269b15c871bfa0b71a9c87994434f/pkg/slayers/path/scion/raw_test.go#L91)):*

Nit: Maybe also add the `assert.ErrorIs` check here to ensure we get the expected error and that the path is not broken in some other way.
</blockquote></details>

done
___
*[`router/dataplane_test.go` line 670 at r3](https://reviewable.io/reviews/scionproto/scion/4483#-NtFjKRoE1tZyLUVIpft:-NtGIDs981asHF7ixGAV:bfph9vc) ([raw file](https://github.com/scionproto/scion/blob/2c63d759221269b15c871bfa0b71a9c87994434f/router/dataplane_test.go#L670)):*

Nit:  "serrors" can be compared with `Is`, and so `assert.ErrorIs(t, serrors.New("...",...), err)` can be used to shorten this.
</blockquote></details>

It is indeed nicer. Unfortunately we just found that a bug in serror prevents doing this. Filed # #4486

matzf (approved):
:lgtm:

> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4483#-:-NtG_RMlDTAoyG592NDM:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

Issue #3935, opened 2020-11-24
acceptance: sig_short_exp_time is a false positive


matzf:
sig_short_exp_time is failing (at least) since the new SIG was introduced in #3932. According to the logs, the script exits with status 1, but for some reason the acceptance test framework still sees a successful test.

Note: The setup for this test is using a fake sciond connection, configured by `fake_sciond.json` data files. Such a fake sciond connection was supported explicitly by the old SIG implementation, specifically for this test. The new SIG does not (yet) have support for this.

matzf:
In #4182

> Fix the false-positive result of sig_short_exp_time, and explicitly skip the now failing test from running. 

The issue was that `ping` failed (as it should) but the exit code was ignored.

PR #4487, opened 2024-03-18
build: replace rules_docker with rules_oci

### LINKED ISSUE #4481: build: replace rules_docker with rules_oci
rules_docker is no longer maintained and causes issues with particular versions of docker (see e.g. #4480). Replace our use of rules_docker by https://github.com/bazel-contrib/rules_oci.

matzf:
rules_docker is no longer maintained and causes issues with particular versions of docker (see e.g. https://github.com/scionproto/scion/issues/4480).
Replace our use of rules_docker by https://github.com/bazel-contrib/rules_oci.

Due to differences in the rules, the build targets in `//docker` have slightly changed behaviour:
- `//docker:<component>` is an unpacked oci layer that can be packed into other layers.
- `//docker:<component>.tarball` is the target producing the docker image .tar
- `//docker:<component>.load` is a runnable target to load the tarball into docker
  The distinction between `.tar` and `.load` is due to a weird behavior of bazel related to targets that are both runnable and return files (`$(location ..)` does not seem to work properly, see comment in `docker/scion_apps.bzl`, in `scion_app_image()`).
- `//docker:prod` and `//docker:test` are no longer runnable, they are just a filegroups of the tarballs. The runnable target could previously be used to load all images. This would now require an additional external rule (multirun) which seems unnecessary (loading images is trivial).
- `//docker:gateway` creates a plain binary without setcap.
  This simplifies the build process; it avoids having to install setcap tools which would be more complicated now with rules_oci/rules_debian_packages.
- `//docker:tester` is now built with rules_debian_packages (instead of the no longer compatible rules_deb_packages).

Introduce consistent naming and tagging of all produced docker images. All images are now tagged `scion/<component>:latest`. For router and gateway, the names were simplified by dropping the `posix-` prefix. 
For internal images used for testing we use nested names e.g. `scion/tools/udpproxy` or `scion/acceptance/topo_reload_cs`. 
Furthermore, all images are annotated with a docker label `org.scion` (no value). This allows cleaning up with a relatively simple match, as is now done in `make clean`. Additionally, the label `org.scion.version` will be set when building with `--stamp`.
The motivation to be able to clean up is to have a simple way to ensure that tests actually pick up the intended images; I managed to thoroughly confuse myself by building new images with a misspelled name, causing old images to be picked up unnoticedly.
I looked for different options to make this even stricter, but alternatives seemed either heavy handed (start a separate docker daemon, or temporarily rename all existing images), or toothless (e.g. use a randomized tag `:test-xxx` for all images, which can be bypassed by tests too easily).

More details:
- simplified acceptance test infrastructure: `--container-loader=<tag>#<path-to-shellscript-loading-a-docker-image>` is now just `--docker-image=<path-to-tar>`. The naming is assumed to be consistent, no `docker tag` is done anymore (see discussion above).
- acceptance/sig_short_exp_time, which is broken #3935: update the setup so it keeps running until the known failure. Instead of building various helper images, which is now rather verbose with rules_oci, use the "stock" images and mount the necessary config as volumes.
- acceptance/topo_daemon_reload and topo_cs_reload: add cleanup so that helper images are removed after test

------

Recommendation: remove any old scion-y images.
```sh
# enable commented section after double checking that these images are safe to remove
docker image ls --format '{{.ID}}\t{{.Repository}}:{{.Tag}}' | grep -E 'scion|dispatcher|daemon|control|tester|udpproxy|posix.router|posix.gateway|bazel\/' # | cut -f 1 | xargs docker image rm
```

Closes #4481


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4487)

oncilla (approved):
:lgtm:

> Reviewed 38 of 39 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4487#-:-NtHeQme5FyciebKUIXX:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @matzf)

___
*[`Makefile` line 45 at r2](https://reviewable.io/reviews/scionproto/scion/4487#-NtHcjR3BHrZrsuHp7co:-NtHcjR3BHrZrsuHp7cp:bxj2qjr) ([raw file](https://github.com/scionproto/scion/blob/72a55463efea49df6e5fa384d2eff8653f69e6fd/Makefile#L45)):*
> ```Makefile
> 	bazel clean
> 	rm -f bin/*
> 	docker image ls --filter reference='scion/*' --filter reference='scion/*/*' -q | xargs --no-run-if-empty docker image rm
> ```

non-blocking: instead of filtering by image name, you could use docker labels.

But the risk is pretty small that we remove images that we do not intend in this way, thus, non-blocking.

lukedirtwalker (commented):
> Reviewed 36 of 39 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4487)* status: all files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`WORKSPACE` line 180 at r3](https://reviewable.io/reviews/scionproto/scion/4487#-NtKqPp53eBH9dqhWP3x:-NtKqPp53eBH9dqhWP3y:b4tskap) ([raw file](https://github.com/scionproto/scion/blob/847dde3fc6e54e14ad8baac09f21bd9a8c87f78a/WORKSPACE#L180)):*
> ```Starlark
>     sha256 = "0ae3b332f9d894e57693ce900769d2bd1b693e1f5ea1d9cdd82fa4479c93bcc8",
>     strip_prefix = "rules_debian_packages-0.2.0",
>     url = "https://github.com/betaboon/rules_debian_packages/releases/download/v0.2.0/rules_debian_packages-v0.2.0.tar.gz",
> ```

I assume you actually use https://github.com/bazel-contrib/rules_debian_packages maybe add a comment here?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4487)* status: 39 of 40 files reviewed, all discussions resolved (waiting on @lukedirtwalker and @oncilla)

___
*[`Makefile` line 45 at r2](https://reviewable.io/reviews/scionproto/scion/4487#-NtHcjR3BHrZrsuHp7co:-NtKwbF01JastE_WfA25:bevios0) ([raw file](https://github.com/scionproto/scion/blob/72a55463efea49df6e5fa384d2eff8653f69e6fd/Makefile#L45)):*

non-blocking: instead of filtering by image name, you could use docker labels.

But the risk is pretty small that we remove images that we do not intend in this way, thus, non-blocking.
</blockquote></details>

Ah that seems like a nice idea, the label also propagates to the containers giving an easy way to filter these too :thumbsup:
I've also added a version label that can be included with --stamp -- not sure what for, but it seemed easy.
___
*[`WORKSPACE` line 180 at r3](https://reviewable.io/reviews/scionproto/scion/4487#-NtKqPp53eBH9dqhWP3x:-NtKueiv1oQ_4bNgx-fN:b-3ox6nk) ([raw file](https://github.com/scionproto/scion/blob/847dde3fc6e54e14ad8baac09f21bd9a8c87f78a/WORKSPACE#L180)):*

I assume you actually use https://github.com/bazel-contrib/rules_debian_packages maybe add a comment here?
</blockquote></details>

Good catch, I blindly copied the installation instructions. The bazel-contrib version is identical.
Done.

PR #4489, opened 2024-03-19
topology: simplify local address configuration

matzf:
The router interface section of the topology.json file previously had two fields to specify the local address of the router:
- public: IP:port
- bind: IP

The idea was that normally only `public` is supplied. When behind NAT, then `bind` could override the IP address on which the router binds locally, and `public` would be the address that is visible to the remote router. The only place where this could be relevant is the router BFD keep-alive messages, as these contain the local address of the router in the SCION address header. There is, however, no check that enforces that this matches the underlay address.
In practice, if `bind` is supplied, the IP of `public` has simply been ignored.

This change replaces `public` and `bind` with a new `local` field. The old fields are deprecated, but still processed for backwards compatibility.
The `local` support omitting the IP to bind on any local IP. This was previously possible only by explicitly configuring a zero IP address.

Aside:
- change the internal representation from `net.UDPAddr` to `netip.AddrPort`. Note that this changes the representation in the json dump of the `/topology` status page.
- remove unused BRNames


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4489)

oncilla (approved):
> Reviewed 36 of 36 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4489)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4492, opened 2024-03-22
cppki/segverifer: enforce AS certificate to cover AS entry

### LINKED ISSUE #4286: cppki: require segment lifetime fully covered by certificate chain

shitz:
When verifying an AS entry the verifier will now only use an AS certificate that is valid until the expiration of the AS entry. Before, the verifier would use any certificate that was valid at the time of verification but not necessarily for the entire validity of the AS entry, i.e., the hop field expiration contained in the entry.
The verifier now requests certificates that cover at least the entire validity period of the AS entry. To that end, the `ChainsRequest.date` field has been renamed to `ChainsRequest.at_least_valid_until` and a new `ChainsRequest.at_least_valid_since` field has been introduced.
Legacy clients of the API will only set the `at_least_valid_until` field (for them it's still the `date` field, but the name of the field does not matter for protobuf wire enconding). New servers will simply set `at_least_valid_since` to be equal to `at_least_valid_until` to keep the same semantics.
Legacy servers will ignore the `at_least_valid_since` field, since they don't know about it. 
This only enforces that the AS certificate is still valid at the AS entry expiration, but not that it was also valid at the start of the hop field validity. Once all trust material servers and clients have been upgraded we automatically have the desired behavior.

Fixes #4286


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4492)

matzf (commented):
> Reviewed 9 of 22 files at r1, 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4492)* status: 12 of 25 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker, @oncilla, and @shitz)

___
*[`proto/control_plane/v1/cppki.proto` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4492#-Ntp93Xv67xI1IEfsGzP:-Ntp93Xv67xI1IEfsGzQ:b-y7b9he) ([raw file](https://github.com/scionproto/scion/blob/076abf8ae10c2e96d282a2ae07ff5f08e0209e8d/proto/control_plane/v1/cppki.proto#L41)):*
> ```proto
>     // be alternatively used to the date field. If both are set, the validity
>     // takes precedence.
>     Validity validity = 4;
> ```

Have considered this alternative approach?
Instead of keeping the "old" `date` _and_ the new `not_before`/`not_after`, we could _rename_ `date` to `not_after`. The semantics for requests to legacy servers would be identical to what is now implemented explicitly (by setting date = not_after in the request). To handle legacy clients, we can say that  if either `not_before` or `not_after` is missing, it is assumed to be identical to the other field.

The advantage would be that we won't have to revisit this later to clean up old behavior. I admit it's also slightly more obscure.



_Suggestion:_
```proto
    // Point in time that must be covered by the AS certificate's validity
    // period. In seconds since UNIX epoch.
    google.protobuf.Timestamp not_after = 3;
    google.protobuf.Timestamp not_before = 4;
```

shitz (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4492)* status: 12 of 25 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker and @oncilla)

___
*[`proto/control_plane/v1/cppki.proto` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4492#-Ntp93Xv67xI1IEfsGzP:-NtuB5lH7iP_GaJs-FKy:b-4lcqm4) ([raw file](https://github.com/scionproto/scion/blob/076abf8ae10c2e96d282a2ae07ff5f08e0209e8d/proto/control_plane/v1/cppki.proto#L41)):*

Have considered this alternative approach?
Instead of keeping the "old" `date` _and_ the new `not_before`/`not_after`, we could _rename_ `date` to `not_after`. The semantics for requests to legacy servers would be identical to what is now implemented explicitly (by setting date = not_after in the request). To handle legacy clients, we can say that  if either `not_before` or `not_after` is missing, it is assumed to be identical to the other field.

The advantage would be that we won't have to revisit this later to clean up old behavior. I admit it's also slightly more obscure.
</blockquote></details>

Yeah that would work I guess, although I'm not sure how common it is to rename a protobuf field eventhough it does not affect the wireformat in any way. I don't have a strong preference.
Maybe @lukedirtwalker or @oncilla have something to say on this?

lukedirtwalker (changes requested):
> Reviewed 21 of 22 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4492)* status: all files reviewed, 1 unresolved discussion (waiting on @oncilla)

___
*[`proto/control_plane/v1/cppki.proto` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4492#-Ntp93Xv67xI1IEfsGzP:-Ntua30fCz829RVM-hWl:bv0byxz) ([raw file](https://github.com/scionproto/scion/blob/076abf8ae10c2e96d282a2ae07ff5f08e0209e8d/proto/control_plane/v1/cppki.proto#L41)):*

Yeah that would work I guess, although I'm not sure how common it is to rename a protobuf field eventhough it does not affect the wireformat in any way. I don't have a strong preference.
Maybe @lukedirtwalker or @oncilla have something to say on this?
</blockquote></details>

I like the proposal, it also reduces the state space a bit, i.e. you can't have a state where both validity and date is set at the same time. Additionally, I like that the migration plan is easier, no need to deprecate a field. So overall I think there are more reasons for it, than against.

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4492)* status: all files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

___
*[`proto/control_plane/v1/cppki.proto` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4492#-Ntp93Xv67xI1IEfsGzP:-Ntvmi1O1KcocqHkZL2s:b2nhrg9) ([raw file](https://github.com/scionproto/scion/blob/076abf8ae10c2e96d282a2ae07ff5f08e0209e8d/proto/control_plane/v1/cppki.proto#L41)):*

I like the proposal, it also reduces the state space a bit, i.e. you can't have a state where both validity and date is set at the same time. Additionally, I like that the migration plan is easier, no need to deprecate a field. So overall I think there are more reasons for it, than against.
</blockquote></details>

Wfm

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4492)* status: all files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @shitz)

___
*[`proto/control_plane/v1/cppki.proto` line 46 at r2](https://reviewable.io/reviews/scionproto/scion/4492#-Ntvmr-rAABMx3F_eydK:-Ntvmr-rAABMx3F_eydL:bm3tal4) ([raw file](https://github.com/scionproto/scion/blob/076abf8ae10c2e96d282a2ae07ff5f08e0209e8d/proto/control_plane/v1/cppki.proto#L46)):*
> ```proto
> message Validity {
>     // NotBefore time of the certificate. In seconds since UNIX epoch.
>     google.protobuf.Timestamp not_before = 1;
> ```

Naming and comment are a bit misleading.

We are looking for certificates that contain the start and end time, not for certificates with exactly that not_before and not_after times

shitz (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4492)* status: 14 of 35 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker, @matzf, and @oncilla)

___
*[`proto/control_plane/v1/cppki.proto` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4492#-Ntp93Xv67xI1IEfsGzP:-Nu0rkxN-sPkmz4Ihi64:boj4qc7) ([raw file](https://github.com/scionproto/scion/blob/076abf8ae10c2e96d282a2ae07ff5f08e0209e8d/proto/control_plane/v1/cppki.proto#L41)):*

Wfm
</blockquote></details>

Done.

I chose the names `AtLeastValidUntil` and `AtLeastValidSince` to better reflect the semantics of the fields. Open for better suggestions.
___
*[`proto/control_plane/v1/cppki.proto` line 46 at r2](https://reviewable.io/reviews/scionproto/scion/4492#-Ntvmr-rAABMx3F_eydK:-Nu0rzI53VTkfQPn04C0:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/076abf8ae10c2e96d282a2ae07ff5f08e0209e8d/proto/control_plane/v1/cppki.proto#L46)):*

Naming and comment are a bit misleading.

We are looking for certificates that contain the start and end time, not for certificates with exactly that not_before and not_after times
</blockquote></details>

Done.

matzf (approved):
> Reviewed 4 of 22 files at r1, 21 of 21 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4492)* status: all files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

lukedirtwalker (approved):
> Reviewed 21 of 21 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4492)* status: all files reviewed, 1 unresolved discussion (waiting on @oncilla)

oncilla (approved):
> Reviewed 11 of 22 files at r1, 3 of 4 files at r2, 21 of 21 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4492)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @shitz)

___
*[`proto/control_plane/v1/cppki.proto` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4492#-Ntp93Xv67xI1IEfsGzP:-Nu4XLF34WZ8w0B4WQPn:b-6ew71a) ([raw file](https://github.com/scionproto/scion/blob/076abf8ae10c2e96d282a2ae07ff5f08e0209e8d/proto/control_plane/v1/cppki.proto#L41)):*

Done.

I chose the names `AtLeastValidUntil` and `AtLeastValidSince` to better reflect the semantics of the fields. Open for better suggestions.
</blockquote></details>

very explicit, wfm :)

PR #4505, opened 2024-04-12
fix #4504: ensure acl default rule matches all

### LINKED ISSUE #4504: pathpol `UnmarshalJSON` and `UnmarshalYAML` should check for default action
If one does unmarshal a json or yaml path policy like `["+ 42", "-"]` it is not guaranteed that the last entry is going to be a default action.

It should have a similar check like the in [NewACL](https://github.com/scionproto/scion/blob/v0.9.1/private/path/pathpol/acl.go#L36-L42):

```go
// NewACL creates a new entry and checks for the presence of a default action
func NewACL(entries ...*ACLEntry) (*ACL, error) {
	if len(entries) == 0 || !entries[len(entries)-1].Rule.matchesAll() {
		return nil, ErrNoDefault
	}
	return &ACL{Entries: entries}, nil
}
```

Otherwise, using `Eval` on the malformed ACL is hitting this [panic](https://github.com/scionproto/scion/blob/v0.9.1/private/path/pathpol/acl.go#L90).

fbuetler:
fixes #4504

### DIFF
--- a/private/path/pathpol/acl.go
+++ b/private/path/pathpol/acl.go
@@ -61,7 +61,11 @@ func (a *ACL) MarshalJSON() ([]byte, error) {
 }
 
 func (a *ACL) UnmarshalJSON(b []byte) error {
-	return json.Unmarshal(b, &a.Entries)
+	err := json.Unmarshal(b, &a.Entries)
+	if len(a.Entries) == 0 || !a.Entries[len(a.Entries)-1].Rule.matchesAll() {
+		return ErrNoDefault
+	}
+	return err
 }
 
 func (a *ACL) MarshalYAML() (interface{}, error) {
@@ -69,7 +73,11 @@ func (a *ACL) MarshalYAML() (interface{}, error) {
 }
 
 func (a *ACL) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	return unmarshal(&a.Entries)
+	err := unmarshal(&a.Entries)
+	if len(a.Entries) == 0 || !a.Entries[len(a.Entries)-1].Rule.matchesAll() {
+		return ErrNoDefault
+	}
+	return err
 }
 
 func (a *ACL) evalPath(pm *snet.PathMetadata) ACLAction {
--- a/private/path/pathpol/acl_test.go
+++ b/private/path/pathpol/acl_test.go
@@ -60,6 +60,38 @@ func TestNewACL(t *testing.T) {
 	}
 }
 
+func TestUnmarshalJSON(t *testing.T) {
+	tests := map[string]struct {
+		Input       []byte
+		ExpectedErr error
+	}{
+		"No entry": {
+			Input:       []byte{},
+			ExpectedErr: ErrNoDefault,
+		},
+		"No default entry": {
+			Input:       []byte(`"+ 42"`),
+			ExpectedErr: ErrNoDefault,
+		},
+		"Entry without rule": {
+			Input: []byte(`["+"]`),
+		},
+		"Entry with hop predicates": {
+			Input: []byte(`["+ 42", "-"]`),
+		},
+	}
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			var acl ACL
+			err := acl.UnmarshalJSON(test.Input)
+			assert.ErrorIs(t, err, test.ExpectedErr)
+			if test.ExpectedErr == nil {
+				assert.NotNil(t, acl)
+			}
+		})
+	}
+}
+
 func TestACLEntryLoadFromString(t *testing.T) {
 	tests := map[string]struct {
 		String         string


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4505)

matzf (changes requested):
Thank you! I only have a few nitpicks below.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4505#-:-Nv_w4-6CmxAnyAUfINt:bwbdoyn)* status: all files reviewed, 6 unresolved discussions (waiting on @fbuetler)

___
*[`private/path/pathpol/acl.go` line 68 at r1](https://reviewable.io/reviews/scionproto/scion/4505#-Nv_rCJ7ELSwclzfsUQw:-Nv_rCJ7ELSwclzfsUQx:bmti93m) ([raw file](https://github.com/scionproto/scion/blob/fd01f12ab1407cd14059ca50b5f7f523627b4f3b/private/path/pathpol/acl.go#L68)):*
> ```Go
> 		return ErrNoDefault
> 	}
> 	return err
> ```

This will shadow any serialization error; if the json is malformed, it will result in an empty list and we say `ErrNoDefault`.
Check for the serialization error first. (Same for the yaml below).



_Suggestion:_
```Go
	if err := json.Unmarshal(b, &a.Entries); err != nil {
		return err
    }
	if len(a.Entries) == 0 || !a.Entries[len(a.Entries)-1].Rule.matchesAll() {
		return ErrNoDefault
	}
	return nil
```
___
*[`private/path/pathpol/acl.go` line 68 at r1](https://reviewable.io/reviews/scionproto/scion/4505#-Nv_rx3UCrdgLMxPC8Vp:-Nv_rx3UCrdgLMxPC8Vq:b567pr1) ([raw file](https://github.com/scionproto/scion/blob/fd01f12ab1407cd14059ca50b5f7f523627b4f3b/private/path/pathpol/acl.go#L68)):*
> ```Go
> 		return ErrNoDefault
> 	}
> 	return err
> ```

It could make sense to add a shared function `func validateACL(entries []*ACLEntry) error` which does the check for the default entry. Then the above code snippet becomes:
```go
    if err := json.Unmarshal(b, &a.Entries); err != nil {
		return err
    }
    return validateACL(a.Entries)
```

The validate function can perform exactly the check that we currently apply -- or it can be a good place to make this a bit more helpful; enforcing the _last_ entry to be a default match seems unnecessarily restrictive. The evaluation just needs _any_ entry that is a default match. Sure, the subsequent entries are "dead code"; if we want to forbid this (to be "helpful"), we could have a separate error for this so the user can more easily understand the problem.
___
*[`private/path/pathpol/acl_test.go` line 70 at r1](https://reviewable.io/reviews/scionproto/scion/4505#-Nv_toh4ECAqAqTmoR4d:-Nv_toh5BfRcPdoQK6N3:b-vtt2a2) ([raw file](https://github.com/scionproto/scion/blob/fd01f12ab1407cd14059ca50b5f7f523627b4f3b/private/path/pathpol/acl_test.go#L70)):*
> ```Go
> 		"No entry": {
> 			Input:       []byte{},
> 			ExpectedErr: ErrNoDefault,
> ```

This should be a json serialization error (see my comment on the "shadowing" of the error).
Perhaps you meant to check `[]` instead?
___
*[`private/path/pathpol/acl_test.go` line 73 at r1](https://reviewable.io/reviews/scionproto/scion/4505#-Nv_uWyu3iiojuO_YrkD:-Nv_uWyu3iiojuO_YrkE:b-ftljgv) ([raw file](https://github.com/scionproto/scion/blob/fd01f12ab1407cd14059ca50b5f7f523627b4f3b/private/path/pathpol/acl_test.go#L73)):*
> ```Go
> 		},
> 		"No default entry": {
> 			Input:       []byte(`"+ 42"`),
> ```

You probably meant `["+ 42"]`

Perhaps change the type of `Input` to `string` and just apply the cast to `[]byte` when invoking the unmarshal function below; then this becomes slightly more readable here.
___
*[`private/path/pathpol/acl_test.go` line 89 at r1](https://reviewable.io/reviews/scionproto/scion/4505#-Nv_xzrpA30Ce1OWDw8_:-Nv_xzrq6QSejuPTRy8D:by0zhsw) ([raw file](https://github.com/scionproto/scion/blob/fd01f12ab1407cd14059ca50b5f7f523627b4f3b/private/path/pathpol/acl_test.go#L89)):*
> ```Go
> 			assert.ErrorIs(t, err, test.ExpectedErr)
> 			if test.ExpectedErr == nil {
> 				assert.NotNil(t, acl)
> ```

This does not check anything, `acl` cannot be `nil` (it's a struct, not a pointer to a struct).
Maybe `assert.NotEmpty(t, acl.Entries)` is a good check?
___
*[`private/path/pathpol/acl_test.go` line 92 at r1](https://reviewable.io/reviews/scionproto/scion/4505#-Nv_wPi30O4_Xh6lO70j:-Nv_wPi30O4_Xh6lO70k:bvdl505) ([raw file](https://github.com/scionproto/scion/blob/fd01f12ab1407cd14059ca50b5f7f523627b4f3b/private/path/pathpol/acl_test.go#L92)):*
> ```Go
> 			}
> 		})
> 	}
> ```

We can use the exact same test inputs for also checking the YAML unmarshal (all the inputs are valid YAML too).

fbuetler:
Thanks a lot for taking the time to review this PR and very good catches!

I applied most of your suggestions. However, I don't quite understand how the `UnmarshalYAML` function works. Where is the input coming from? Is it hidden within the supplied function? Is there a reason to not use `yaml.Unmarshal` analogously to `UnmarshalJSON`?

oncilla (commented):
Yes. The yaml unmarshler interface is a bit of a mind bender.

The provided function should be used to unmarshal into a variable.
See https://github.com/search?q=repo%3Ascionproto%2Fscion%20unmarshalyaml&type=code for some examples

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4505#-:-NvfWiBx6GgSq2enZjJl:b-wpbvl5)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @fbuetler and @matzf)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4505)* status: all files reviewed, 3 unresolved discussions (waiting on @fbuetler)

___
*[`private/path/pathpol/acl.go` line 32 at r3](https://reviewable.io/reviews/scionproto/scion/4505#-NvkCOQCCKNfDclhVG7w:-NvkCOQCCKNfDclhVG7x:bqu6mc1) ([raw file](https://github.com/scionproto/scion/blob/e67b7ba7208c63497c090b0018e4d46f32df6bd3/private/path/pathpol/acl.go#L32)):*
> ```Go
> 	// ErrObsoleteEntries indicates that there is a default acl entry posisioned in
> 	// the middle, making the following acl entries obsolete.
> 	ErrObsoleteEntries = errors.New("ACL has a default entry posisioned in the middle, making the following entries obsolete")
> ```

Nit. Typo "posisioned" -> "positioned". 
Maybe "obsolete" is not the ideal term here (it means outmoded, out of date), I'd suggest "unused" or "extra": `ErrExtraEntries = .. "ACL has unused extra entries after a default entry"`.
___
*[`private/path/pathpol/acl_test.go` line 94 at r3](https://reviewable.io/reviews/scionproto/scion/4505#-NvkDiWh-50HHXWxyAPB:-NvkDiWh-50HHXWxyAPC:b-v4lkws) ([raw file](https://github.com/scionproto/scion/blob/e67b7ba7208c63497c090b0018e4d46f32df6bd3/private/path/pathpol/acl_test.go#L94)):*
> ```Go
> 		t.Run(name, func(t *testing.T) {
> 			var acl ACL
> 			err := acl.UnmarshalJSON([]byte(test.Input))
> ```

Note, perhaps this can help with your question regarding using UnmarshalJSON/YAML. 

Here, calling `acl.UnmarshalJSON(buffer)` works and is perfectly fine, but the more "generic" invocation would be `json.Unmarshal(buffer, &acl)`. This variant works for all "unmarshallable" types, not only those types that directly implement a custom `UnmarshalJSON` -- for example, `json.Unmarshal()` would also let you unmarshal strings, numbers, structs with default unmarshaller, or a list of `ACL` by `var v []ACL; json.Unmarshal(buffer, &v)`.

The yaml interface is similar, use `yaml.Unmarshal(buffer, &v)`. Here, the customized UnmarshalYAML cannot be invoked directly so easily as with the JSON counterpart.

fbuetler:
So the idea is to do something like this?

```go
			var acl ACL
			err := acl.UnmarshalYAML(func(entries interface{}) error {
				return yaml.Unmarshal([]byte(test.Input), entries)
			})
			assert.ErrorIs(t, err, test.ExpectedErr)
```

matzf (commented):
No, just `err := yaml.Unmarshal([]byte(test.Input), &acl)`.

(see also the note in my previous comment in acl_test.go a few moments ago)

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4505#-:-NvkNsuq9qAsEp5j1k1D:b-rw13fu)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @fbuetler)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4505)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @fbuetler)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4505#-NvkPTpU0wRRIUZ8u1mk:-NvkPTpU0wRRIUZ8u1ml:b-841iu) (no related file):*
There is a failing test now, [pathpol/policy_test.go TestPathJsonConversion](https://buildkite.com/scionproto/scion/builds/4340#018ef064-3e33-4226-8b74-f28eca992c2e/182).
This fails because the policy contains a default deny entry after a test entry that is "accidentally" a default entry (predicate `0-0#0` is a match all). Please change the test there; perhaps it would make sense to create the ACL with `NewACL` to ensure that the test does not create a bogus object in the first place.

___

fbuetler:
Ah, this works because of the [Unmarshaler interface](https://pkg.go.dev/gopkg.in/yaml.v2#Unmarshaler). This was the missing puzzle piece for me.

matzf (approved):
> Reviewed 1 of 2 files at r5, 2 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4505)* status: all files reviewed, 2 unresolved discussions (waiting on @fbuetler)

___
*[`private/path/pathpol/policy_test.go` line 592 at r8](https://reviewable.io/reviews/scionproto/scion/4505#-Nvl6jPg2q4wiARN-O9G:-Nvl6jPg2q4wiARN-O9H:b-xrzfn3) ([raw file](https://github.com/scionproto/scion/blob/05fe2d6def4ade0a895f0294ddbf28efa5aa7d42/private/path/pathpol/policy_test.go#L592)):*
> ```Go
> 		{Action: Allow, Rule: mustHopPredicate(t, "42-0#0")},
> 		denyEntry,
> 	}...)
> ```

Nit: no real problem it works, but it's more clumsy than necessary. The slice and then unpacking with ellipsis can be left away:



_Suggestion:_
```Go
	acl, err := NewACL(
		&ACLEntry{Action: Allow, Rule: mustHopPredicate(t, "42-0#0")},
		denyEntry,
	)
```

matzf (approved):
:lgtm:

> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4505#-:-Nvp5Lxc2nBvLQQxqTGy:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @fbuetler)

matzf (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4505)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @fbuetler)

PR #4499, opened 2024-04-03
private/env: remove RunsInDocker during startup log

JordiSubira:
This PR tries to remove a Linux-only dependency for running SCION applications of other platforms. The `RunsUnDocker()` method is the Linux-only call, thus we check whether we are running in Docker only in Linux platform. I am in favor of even removing this log line, supporting @matzf comment on https://reviewable.io/reviews/scionproto/scion/4344#-Nt_IGRG4RbvC-PCVToV.

PR #4507, opened 2024-04-18
router: factor out platform-dependent underlay conn flags

### LINKED ISSUE #4046: Remove Linux-Only Features
I'm using SCION on Darwin OSes (iOS and macOS) and dispatcher and sciond can be _almost_ compiled straight out of the box if it weren't for one unsupported sycall flag in `underlay/conn.go`: [`MSG_WAITFORONE`](https://github.com/scionproto/scion/blob/master/go/lib/underlay/conn/conn.go#L94)

In this case the affected code can simply be commented out because it appears to be unused anyway. But going forward I suggest removing this specific syscall flag so that dispatcher can be compiled with no modifications on macOS and other OSes that aren't Linux. Surely there must be some way to work without this flag.

Generally it should be avoided to use [OS-specific features such as this `MSG_WAITFORONE` flag](https://man7.org/linux/man-pages/man2/recvmmsg.2.html) – I think it just unnecessarily hurts the possibilities for broader adoption.

marcfrei:
The use of the platform-dependent flag syscall.MSG_WAITFORONE prevents building the router on platforms other than Linux. This change isolates this dependency. Behavior on non-Linux platforms will be less efficient but still semantically correct because ReadBatch will only read a single message on those platforms anyway.

Together with the previous PR #4499 and the dispatcher removal PR #4344 this set of changes will, e.g., enable running local development topologies on platforms other than Linux.

Fixes #4046.


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4507)

matzf (approved):
:lgtm:

> Reviewed 2 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4507#-:-Nvkvovc-W7xz3K1wkL2:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @marcfrei)

PR #4511, opened 2024-04-22
deps: update lestrrat-go/jwx, x/net, x/crypto

matzf:
Update github.com/lestrrat-go/jwx from v1.2.25 to v1.2.29.

This includes security patches for vulnerabilities reported to us via dependabot:
- lestrrat-go/jwx's malicious parameters in JWE can cause a DOS
- Parsing JSON serialized payload without protected field can lead to segfault
- JWX vulnerable to a denial of service attack using compressed JWE message

As far as I understand, these issues affect the JWE and JWS modules, which we don't use. We use the JWA and JWT modules to perform JWT authorization via Bearer tokens as part of the internal management API.

The update to jwx triggers version bumps for the x/net and x/crypto dependencies. If we bump these, we might as well use the latest version of these:
Update x/net from v0.14.0 to v0.24.0.
Update x/crypto from v0.12.0 to v0.22.0.

This, in turn, addresses a number of dependabot vulnerability alerts.
- HTTP/2 Stream Cancellation Attack
- Prefix Truncation Attack against ChaCha20-Poly1305 and Encrypt-then-MAC aka Terrapin
- HTTP/2 rapid reset can cause excessive work in net/http
- net/http, x/net/http2: close connections when receiving too many headers

As far as I'm aware, the internal management APIs do not use HTTP/2, and the gRPC used for the control plane messaging has a separate implementation of HTTP/2. Thus, I believe the vulnerable code is not actually used by us.


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4511)

PR #4516, opened 2024-04-26
control: select a signer that covers the segment

lukedirtwalker:
When extending a segment we should select a signer that covers the whole validity of the segment. So far we only checked the end date, but with the introduction of enforcing that the AS certificate covers the whole AS entry (#4492) the verifier checks that the start date of the certificate also covers the whole segment lifetime. However for signing we so far only used the latest certificate, which might have a later start date than the timestamp of the segment.

This is a problem because between creation of a segment and the signing of a AS entry in a downstream AS there can be quite some time.

This commit now makes sure to select a signer/certificate that covers the start of the segments.


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4516)

oncilla (approved):
> Reviewed 11 of 18 files at r1, 13 of 13 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4516)* status: all files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker and @matzf)

___
*[`control/mgmtapi/api.go` line 698 at r2](https://reviewable.io/reviews/scionproto/scion/4516#-NwdacQa5Kn8wCvhjOvK:-NwdacQa5Kn8wCvhjOvL:bw0va9d) ([raw file](https://github.com/scionproto/scion/blob/3e9c0508aa2ea7de9a5cde288981d6cecd918cc8/control/mgmtapi/api.go#L698)):*
> ```Go
> 		return
> 	}
> 	now := time.Now()
> ```

nit: I think that should use s.now() for consistency

oncilla (commented):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4516)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @matzf)

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4516)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #4518, opened 2024-05-03
testing: incremental improvements to the router benchmark tooling

jiceatscion:
Summary of changes:
* The packet size is now configurable. CI uses a constant while the benchmark command accepts an optional argument.
* The benchmark command sets the mtu to 9000 on the host side and instructs the user to do the same on the router side. In the CI test variant: an mtu of 9000 (not 8000) is used on both sides.
* The performance index formula gained a tentative value for the NIC_CONSTANT constant.
* The benchmark's total number of packets (per testcase) is no longer tunable. Instead, the duration is what is configurable. Both the CI test and the benchmark command use 13 seconds per test case.
* The benchmark's output includes a bit more information: bit rate and raw (not corrected for preemptions) packet rate.
* A useless and noisy error message was silenced in brload: that which was output for every foreign packet received (turns out that some Realtek hardware sends proprietary management traffic as Ethernet frames).
* Improved the relevance of mmbm by deliberately misaligning buffers. Also make sure to not use more buffers than can fit in L3 cache when probing for L2 cache + TLB misses.
* Added zip as a dependency after finding it isn't part of Debian by default.
* Added an option (via constant for now) to profile the router during the CI test.


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4518)

matzf (commented):
Title: `testing: incremental improvements to router benchmark tooling`

> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4518#-:-O-_vmN16kfBV6Pxuv6N:b-432i93)* status: all files reviewed, 5 unresolved discussions (waiting on @jiceatscion)

___
*[`acceptance/router_benchmark/brload/main.go` line 169 at r1](https://reviewable.io/reviews/scionproto/scion/4518#-O-_shF74dmC_saH1byX:-O-_shF74dmC_saH1byY:b-lyqc0r) ([raw file](https://github.com/scionproto/scion/blob/78c94047c5f53fecf09abad41aee522310db5174/acceptance/router_benchmark/brload/main.go#L169)):*
> ```Go
> 		packetSize = 154
> 	}
> 	payloadSize := packetSize - 154
> ```

The SCION header size is not really constant, even for the existing cases the sizes differ by some bytes. Unlikely that this makes a big difference, but it could be confusing. If we want this to be accurate, we could pass the selected packet size to the case functions and generate an appropriate payload size for each individual case.
___
*[`acceptance/router_benchmark/brload/main.go` line 209 at r1](https://reviewable.io/reviews/scionproto/scion/4518#-O-_fXLY68ejec9SaXAX:-O-_fXLY68ejec9SaXAY:beyq5xb) ([raw file](https://github.com/scionproto/scion/blob/78c94047c5f53fecf09abad41aee522310db5174/acceptance/router_benchmark/brload/main.go#L209)):*
> ```Go
> 	binary.BigEndian.PutUint16(rawPkt[40:42], 0)
> 
> 	for time.Now().Unix() < metricsEnd {
> ```

Nit. `time.Since(begin) >= testDuration` could be slightly faster because it only accesses the monotonic clock (and also robust against wall clock changes). Only mentioning this because this seems expensive enough to move it out of the hot loop.
___
*[`acceptance/router_benchmark/brload/main.go` line 211 at r1](https://reviewable.io/reviews/scionproto/scion/4518#-O-_gQqs4MabKrl5EojT:-O-_gQqs4MabKrl5EojU:b-hypc8s) ([raw file](https://github.com/scionproto/scion/blob/78c94047c5f53fecf09abad41aee522310db5174/acceptance/router_benchmark/brload/main.go#L211)):*
> ```Go
> 	for time.Now().Unix() < metricsEnd {
> 		// Check the time only once every 10000 packets
> 		for i := 0; i < 10000; i++ {
> ```

Nit: you took great care to pick 840 streams, but now introduce this modulo bias here? :wink:
___
*[`acceptance/router_benchmark/brload/main.go` line 278 at r1](https://reviewable.io/reviews/scionproto/scion/4518#-O-_h82s2ToZnvf82c81:-O-_h82s2ToZnvf82c82:b-5qhb5u) ([raw file](https://github.com/scionproto/scion/blob/78c94047c5f53fecf09abad41aee522310db5174/acceptance/router_benchmark/brload/main.go#L278)):*
> ```Go
> 		}
> 		checkLen := len(payload)
> 		if checkLen > 20 {
> ```

What is this 20?
___
*[`tools/mmbm/main.go` line 44 at r1](https://reviewable.io/reviews/scionproto/scion/4518#-O-_u5hg2hjnUSlEDDwX:-O-_u5hg2hjnUSlEDDwY:bveh2j3) ([raw file](https://github.com/scionproto/scion/blob/78c94047c5f53fecf09abad41aee522310db5174/tools/mmbm/main.go#L44)):*
> ```Go
> // Block is a packet buffer representation arranged to make it easy to copy 1, 172 or 4096
> // bytes. It is exported to make sure go cannot optimize fields out. Missalign is exported
> // so lint doesn't complain.
> ```

Nit: can also just `//nolint:unused`

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4518)* status: 4 of 11 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`acceptance/router_benchmark/brload/main.go` line 169 at r1](https://reviewable.io/reviews/scionproto/scion/4518#-O-_shF74dmC_saH1byX:-O-fTCci815vzJSYKmbT:b-896fjx) ([raw file](https://github.com/scionproto/scion/blob/78c94047c5f53fecf09abad41aee522310db5174/acceptance/router_benchmark/brload/main.go#L169)):*

The SCION header size is not really constant, even for the existing cases the sizes differ by some bytes. Unlikely that this makes a big difference, but it could be confusing. If we want this to be accurate, we could pass the selected packet size to the case functions and generate an appropriate payload size for each individual case.
</blockquote></details>

Done
___
*[`acceptance/router_benchmark/brload/main.go` line 209 at r1](https://reviewable.io/reviews/scionproto/scion/4518#-O-_fXLY68ejec9SaXAX:-O-fUpp71_eSxJLTWhpj:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/78c94047c5f53fecf09abad41aee522310db5174/acceptance/router_benchmark/brload/main.go#L209)):*

Nit. `time.Since(begin) >= testDuration` could be slightly faster because it only accesses the monotonic clock (and also robust against wall clock changes). Only mentioning this because this seems expensive enough to move it out of the hot loop.
</blockquote></details>

done
___
*[`acceptance/router_benchmark/brload/main.go` line 211 at r1](https://reviewable.io/reviews/scionproto/scion/4518#-O-_gQqs4MabKrl5EojT:-O-fVy-T7OSuUsk4proP:boqj2yz) ([raw file](https://github.com/scionproto/scion/blob/78c94047c5f53fecf09abad41aee522310db5174/acceptance/router_benchmark/brload/main.go#L211)):*

Nit: you took great care to pick 840 streams, but now introduce this modulo bias here? :wink:
</blockquote></details>

Ooops. i used to be the total pkt count. Just forgot to adjust.
Done
___
*[`acceptance/router_benchmark/brload/main.go` line 278 at r1](https://reviewable.io/reviews/scionproto/scion/4518#-O-_h82s2ToZnvf82c81:-O-fWC7C6kpYo9cG3e27:bz2zyvu) ([raw file](https://github.com/scionproto/scion/blob/78c94047c5f53fecf09abad41aee522310db5174/acceptance/router_benchmark/brload/main.go#L278)):*

What is this 20?
</blockquote></details>

It doesn't really mater. Just a small sample. The sample payload string can change. ("actualpayloadbytes" is 18 bytes; I am throwing in 2 extra zeros). I don't what I could use besides a magic constant. I do not have access to the sample payload string any more. It is in the test cases.
___
*[`tools/mmbm/main.go` line 44 at r1](https://reviewable.io/reviews/scionproto/scion/4518#-O-_u5hg2hjnUSlEDDwX:-O-fZl3N3U2CZiH2Z3qe:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/78c94047c5f53fecf09abad41aee522310db5174/tools/mmbm/main.go#L44)):*

Nit: can also just `//nolint:unused`
</blockquote></details>

done

jiceatscion (commented):
done

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4518#-:-O-fjGLxCsQfK7caaet8:b-7rl0kd)* status: 4 of 11 files reviewed, all discussions resolved (waiting on @matzf)

jiceatscion:
Matthias, I think I have addressed your comments. I also added the profiling. Is this good to go?

matzf (commented):
Looks good to me, just one more nit pick.

> Reviewed 7 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4518#-:-O09aEx5EcdBizeYTAJW:b-vj1gpi)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

___
*[`acceptance/router_benchmark/cases/helpers.go` line 23 at r4](https://reviewable.io/reviews/scionproto/scion/4518#-O09Suhq1a_YPDTYjO9p:-O09Suhq1a_YPDTYjO9q:biuwgtl) ([raw file](https://github.com/scionproto/scion/blob/047f7d71e27dcc9ca3d2bb5d0059c92a3611b52c/acceptance/router_benchmark/cases/helpers.go#L23)):*
> ```Go
> func headerLength(options gopacket.SerializeOptions, layers ...gopacket.SerializableLayer) int {
> 	sb := gopacket.NewSerializeBuffer()
> 	if err := gopacket.SerializeLayers(sb, options, layers...); err != nil {
> ```

Ideally we'd set both SerializeOptions flags (FixLengths and ComputeChecksums) to false; now, this is called with the options from each test case, where both of these are set to true. With these options on, SerializeLayers modifies the layers, which seems awkward for this function `headerLength` that could be expected to only return a value.

My suggestion would be to not expose this options here; create a new options in the function with both flags set to false.

In the same vain, `mkPacket` could drop the parameter also and set both flags to `true`. There isn't really a good reason  to leave this up to the caller of this convenience helper function.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4518)* status: 7 of 13 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`acceptance/router_benchmark/cases/helpers.go` line 23 at r4](https://reviewable.io/reviews/scionproto/scion/4518#-O09Suhq1a_YPDTYjO9p:-O09iv41513CTn2ObOQa:b-ebfjay) ([raw file](https://github.com/scionproto/scion/blob/047f7d71e27dcc9ca3d2bb5d0059c92a3611b52c/acceptance/router_benchmark/cases/helpers.go#L23)):*

Ideally we'd set both SerializeOptions flags (FixLengths and ComputeChecksums) to false; now, this is called with the options from each test case, where both of these are set to true. With these options on, SerializeLayers modifies the layers, which seems awkward for this function `headerLength` that could be expected to only return a value.

My suggestion would be to not expose this options here; create a new options in the function with both flags set to false.

In the same vain, `mkPacket` could drop the parameter also and set both flags to `true`. There isn't really a good reason  to leave this up to the caller of this convenience helper function.
</blockquote></details>

you go it!

jiceatscion (commented):
done

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4518#-:-O0A-zO1BUoTgI_NQX3p:b-7rl0kd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4520, opened 2024-05-03
router: optimize computeProcID

### LINKED ISSUE #4519: router: computeProcID consumes 2.5% of packet processing time

jiceatscion:
Fixes #4519

### DIFF
--- a/router/BUILD.bazel
+++ b/router/BUILD.bazel
@@ -5,6 +5,7 @@ go_library(
     srcs = [
         "connector.go",
         "dataplane.go",
+        "fnv1aCheap.go",
         "metrics.go",
         "svc.go",
     ],
--- a/router/dataplane.go
+++ b/router/dataplane.go
@@ -678,7 +678,7 @@ func computeProcID(data []byte, numProcRoutines int, base uint16) (uint32, error
 	// inject the flowID
 	s ^= uint32(data[1] & 0xF) // The left 4 bits aren't part of the flowID.
 	s *= prime32
-	for _, c := range data[2:3] {
+	for _, c := range data[2:4] {
 		s ^= uint32(c)
 		s *= prime32
 	}
--- a/router/dataplane.go
+++ b/router/dataplane.go
@@ -23,7 +23,6 @@ import (
 	"errors"
 	"fmt"
 	"hash"
-	"hash/fnv"
 	"math/big"
 	"net"
 	"net/netip"
@@ -598,16 +597,10 @@ func (d *DataPlane) runReceiver(ifID uint16, conn BatchConn, cfg *RunConfig,
 	procQs []chan packet) {
 
 	log.Debug("Run receiver for", "interface", ifID)
-	randomValue := make([]byte, 16)
-	if _, err := rand.Read(randomValue); err != nil {
-		panic("Error while generating random value")
-	}
 
 	msgs := underlayconn.NewReadMessages(cfg.BatchSize)
 	numReusable := 0                     // unused buffers from previous loop
 	metrics := d.forwardingMetrics[ifID] // If receiver exists, fw metrics exist too.
-	flowIDBuffer := make([]byte, 3)
-	hasher := fnv.New32a()
 
 	enqueueForProcessing := func(pkt ipv4.Message) {
 		srcAddr := pkt.Addr.(*net.UDPAddr)
@@ -616,8 +609,7 @@ func (d *DataPlane) runReceiver(ifID uint16, conn BatchConn, cfg *RunConfig,
 		metrics[sc].InputPacketsTotal.Inc()
 		metrics[sc].InputBytesTotal.Add(float64(size))
 
-		procID, err := computeProcID(pkt.Buffers[0],
-			cfg.NumProcessors, randomValue, flowIDBuffer, hasher)
+		procID, err := computeProcID(pkt.Buffers[0], cfg.NumProcessors, ifID)
 		if err != nil {
 			log.Debug("Error while computing procID", "err", err)
 			d.returnPacketToPool(pkt.Buffers[0])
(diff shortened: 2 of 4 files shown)


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4520)

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4520)* status: all files reviewed, 2 unresolved discussions (waiting on @jiceatscion)

___
*[`router/dataplane.go` line 604 at r2](https://reviewable.io/reviews/scionproto/scion/4520#-NxBi0DyAmlZTKtc7789:-NxBi0DyAmlZTKtc778A:bwky3u2) ([raw file](https://github.com/scionproto/scion/blob/f82aabe41beaf72569d78f28e83f932b5554a700/router/dataplane.go#L604)):*
> ```Go
> 	if _, err := rand.Read(randomValue); err != nil {
> 		panic("Error while generating random value")
> 	}
> ```

Adding this randomness into the hash is a security mechanism, it makes it harder for an attacker to attack the hash function. E.g. an attacker could compute a flow ID to target processing on the same processing queue as a specific victim flow. Note that the output of the hash function is very difficult to observe for an attacker, and so recovering the random value, even if it's short, would be very hard.

That said, 16 bytes of randomness is possibly excessive; 4 bytes seem to suffice to sufficiently randomize over most of the "state space" (I ran a quick experiment, a bit under 2^31 distinct outputs are reached from the full set of 2^32 four byte inputs). If hashing this same random value repeatedly has a significant cost, it could also make sense to precompute the hash state for the random input.
___
*[`router/dataplane.go` line 693 at r2](https://reviewable.io/reviews/scionproto/scion/4520#-NxBlk3f-wROZ2UCtXNS:-NxBlk3f-wROZ2UCtXNT:b-d392xp) ([raw file](https://github.com/scionproto/scion/blob/6b35ac4d2783d1e839cecc857de24d2211f1609a/router/dataplane.go#L693)):*
> ```Go
> 
> 	return s % uint32(numProcRoutines), nil
> }
> ```

Would be nice to still implement the main body of the hash function in a separate function, perhaps also in a separate file (this dataplane.go is growing to contain everything).
Something like `func hashFNV1a(state uint32, c byte) uint32`, then use it as `s = hashFNV1a(s, c)` . This should be inlined and so have no overhead. The definition of the `prime32` constant can be moved into that function body. The initial offset is needed for the initialization, I'd define it as something like `fnv1Offset32`.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4520)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 604 at r2](https://reviewable.io/reviews/scionproto/scion/4520#-NxBi0DyAmlZTKtc7789:-NxllaYC0dxwm4ZQfr1T:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/f82aabe41beaf72569d78f28e83f932b5554a700/router/dataplane.go#L604)):*

Adding this randomness into the hash is a security mechanism, it makes it harder for an attacker to attack the hash function. E.g. an attacker could compute a flow ID to target processing on the same processing queue as a specific victim flow. Note that the output of the hash function is very difficult to observe for an attacker, and so recovering the random value, even if it's short, would be very hard.

That said, 16 bytes of randomness is possibly excessive; 4 bytes seem to suffice to sufficiently randomize over most of the "state space" (I ran a quick experiment, a bit under 2^31 distinct outputs are reached from the full set of 2^32 four byte inputs). If hashing this same random value repeatedly has a significant cost, it could also make sense to precompute the hash state for the random input.
</blockquote></details>

Done.
___
*[`router/dataplane.go` line 693 at r2](https://reviewable.io/reviews/scionproto/scion/4520#-NxBlk3f-wROZ2UCtXNS:-NxlldqVBPqhaudw8mrk:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/6b35ac4d2783d1e839cecc857de24d2211f1609a/router/dataplane.go#L693)):*

Would be nice to still implement the main body of the hash function in a separate function, perhaps also in a separate file (this dataplane.go is growing to contain everything).
Something like `func hashFNV1a(state uint32, c byte) uint32`, then use it as `s = hashFNV1a(s, c)` . This should be inlined and so have no overhead. The definition of the `prime32` constant can be moved into that function body. The initial offset is needed for the initialization, I'd define it as something like `fnv1Offset32`.
</blockquote></details>

Done.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4520)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 675 at r3](https://reviewable.io/reviews/scionproto/scion/4520#-NxmSZ46-b8ssqql7uGr:-Nxm_dyN2siJbpQSTdUt:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3bf323912c834f2ae3d8bec0fc11cb0be39e314b/router/dataplane.go#L675)):*

`s := hashSeed`
</blockquote></details>

Done.
___
*[`router/fnv1aCheap.go` line 19 at r3](https://reviewable.io/reviews/scionproto/scion/4520#-NxmDi4_2vA-ntMzkMVH:-Nxm_gBA5wd6vYooOghD:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/3bf323912c834f2ae3d8bec0fc11cb0be39e314b/router/fnv1aCheap.go#L19)):*

Make this uint32, `const fnv1aOffset32 uint32 = ...`, and remove the cast where the value is referenced.
</blockquote></details>

Done.

matzf:
For whatever it's worth, comparing the router benchmark results [with this PR](https://buildkite.com/scionproto/scion/builds/4411#018f7297-e4eb-4bd3-ab69-bf2ab1d0f9f6) against [nightly](https://buildkite.com/scionproto/scion-nightly/builds/2070#018f7543-889c-48f2-a4d9-28e468787665), I see an improvement of 0.09%. Not a whole lot gained, but not a lot of complexity either.

jiceatscion (commented):
Yeah. Most likely the time removed from computing was added to the time sleeping :-( Oh well, when we improve the I/O side we'll be able to profit from this.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4520#-:-NxqDo4_AIU29PvGvGkN:b-99tnti)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #4521, opened 2024-05-10
Nat address discovery

tzaeschke:
Draft for proposal on NAT address/port discovery


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4521)

PR #4526, opened 2024-05-16
Router: remove most uses of net.UDPAddr in favor of netip.AddrPort

jiceatscion:
This stemmed from an effort at reducing the cost of ResolveLocal, which comes mostly from the net.UDPAddr API causing lots of garbage to collect. In the end that change is a bit viral and many more less costly uses were removed as well.


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4526)

PR #4527, opened 2024-05-16
pkg/metrics: add simple metrics

lukedirtwalker:
Introduce simple version of the pkg/metrics objects. The simpler versions don't have a With method. The With version can have quite some overhead if used in a hot path. For now the simple counters are used in snet.

### DIFF
--- a/control/observability.go
+++ b/control/observability.go
@@ -82,7 +82,7 @@ type Metrics struct {
 	TrustTRCFileWritesTotal                *prometheus.CounterVec
 	SCIONNetworkMetrics                    snet.SCIONNetworkMetrics
 	SCIONPacketConnMetrics                 snet.SCIONPacketConnMetrics
-	SCMPErrors                             metrics.Counter
+	SCMPErrors                             metrics.SimpleCounter
 	TopoLoader                             topology.LoaderMetrics
 	DRKeySecretValueQueriesTotal           *prometheus.CounterVec
 	DRKeyLevel1QueriesTotal                *prometheus.CounterVec
--- a/gateway/metrics.go
+++ b/gateway/metrics.go
@@ -304,7 +304,7 @@ type Metrics struct {
 
 	// Scion Network Metrics
 	SCIONNetworkMetrics    snet.SCIONNetworkMetrics
-	SCMPErrors             metrics.Counter
+	SCMPErrors             metrics.SimpleCounter
 	SCIONPacketConnMetrics snet.SCIONPacketConnMetrics
 }
 
--- a/gateway/pathhealth/pathwatcher.go
+++ b/gateway/pathhealth/pathwatcher.go
@@ -68,7 +68,7 @@ type DefaultPathWatcherFactory struct {
 	// remote.
 	ProbesSendErrors func(remote addr.IA) metrics.Counter
 
-	SCMPErrors             metrics.Counter
+	SCMPErrors             metrics.SimpleCounter
 	SCIONPacketConnMetrics snet.SCIONPacketConnMetrics
 }
 
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -53,14 +53,22 @@ import (
 // Counter describes a metric that accumulates values monotonically.
 // An example of a counter is the number of received HTTP requests.
 type Counter interface {
+	SimpleCounter
 	With(labelValues ...string) Counter
+}
+
+type SimpleCounter interface {
 	Add(delta float64)
 }
 
 // Gauge describes a metric that takes specific values over time.
 // An example of a gauge is the current depth of a job queue.
 type Gauge interface {
+	SimpleGauge
 	With(labelValues ...string) Gauge
+}
+
+type SimpleGauge interface {
 	Set(value float64)
 	Add(delta float64)
 }
@@ -70,13 +78,17 @@ type Gauge interface {
 // typically expressed as quantiles or buckets. An example of a histogram is
 // HTTP request latencies.
 type Histogram interface {
+	SimpleHistogram
 	With(labelValues ...string) Histogram
+}
+
+type SimpleHistogram interface {
 	Observe(value float64)
 }
 
 // CounterAdd increases the passed in counter by the amount specified.
 // This is a no-op if c is nil.
-func CounterAdd(c Counter, delta float64) {
+func CounterAdd(c SimpleCounter, delta float64) {
 	if c != nil {
 		c.Add(delta)
 	}
@@ -84,7 +96,7 @@ func CounterAdd(c Counter, delta float64) {
 
 // CounterInc increases the passed in counter by 1.
 // This is a no-op if c is nil.
-func CounterInc(c Counter) {
+func CounterInc(c SimpleCounter) {
 	CounterAdd(c, 1)
 }
 
@@ -98,7 +110,7 @@ func CounterWith(c Counter, labelValues ...string) Counter {
 
 // GaugeSet sets the passed in gauge to the value specified.
 // This is a no-op if g is nil.
-func GaugeSet(g Gauge, value float64) {
+func GaugeSet(g SimpleGauge, value float64) {
 	if g != nil {
 		g.Set(value)
 	}
@@ -106,7 +118,7 @@ func GaugeSet(g Gauge, value float64) {
 
 // GaugeSetTimestamp sets the passed gauge to the specified time stamp.
 // This is a no-op if g is nil.
-func GaugeSetTimestamp(g Gauge, ts time.Time) {
+func GaugeSetTimestamp(g SimpleGauge, ts time.Time) {
 	if g != nil {
 		g.Set(Timestamp(ts))
 	}
@@ -114,13 +126,13 @@ func GaugeSetTimestamp(g Gauge, ts time.Time) {
 
 // GaugeSetCurrentTime sets the passed gauge to the current time.
 // This is a no-op if g is nil.
-func GaugeSetCurrentTime(g Gauge) {
+func GaugeSetCurrentTime(g SimpleGauge) {
 	GaugeSetTimestamp(g, time.Now())
 }
 
 // GaugeAdd increases the passed in gauge by the amount specified.
 // This is a no-op if g is nil.
-func GaugeAdd(g Gauge, delta float64) {
+func GaugeAdd(g SimpleGauge, delta float64) {
 	if g != nil {
 		g.Add(delta)
 	}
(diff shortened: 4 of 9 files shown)


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4527)

matzf (commented):
Constructing the `SimpleCounter` (same for other SimpleX, I'll just stick to Counter as an example) types with `NewCounter` seems very restricted now, it can only be used when all the label values have been set via the prometheus library.  Like this, I don't see much of a point in using this wrapper library, we might as well just use prometheus.Counter instead of metrics.SimpleCounter. 
Instead of this special case, we could add an additional method to "bake" a `Counter` into a `SimpleCounter` once all label values have been "With-ed".


```go
type Counter interface {
   SimpleCounter                  // compatibility / shortcut for To().Add()
   With(labelValues ...string) Counter
   To() SimpleCounter          // "bake" 
}

type SimpleCounter interface {
   Add(delta float64)
}


// .... in prometheus.go:

func (c *counter) To() SimpleCounter {
  return &simpleCounter {
    c.cv.With(makeLabels(c.lvs...))
  }
}

func (c *counter) Add(delta float64) {
  c.To().Add(delta)
}
```
With this, we don't need the new `NewCounter` constructor.

I wonder if `With` could directly return `SimpleCounter`, or if we have many use cases where we incrementally apply the label values.
Ultimately, I think I'd remove the `SimpleCounter` interface from `Counter`, requiring that the caller makes this "baking" explicit. But this would be rather lot of type changes, so maybe could make sense to do this later.

"Simple" might not be an ideal name under this proposal. I find prometheus terminology quite sensible, `CounterVec` corresponding to our `Counter`, and `Counter` corresponding to `SimpleCounter`.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4527#-:-Ny0FAE24Z3H70N4ohYp:bx0lwqr)* status: 0 of 9 files reviewed, all discussions resolved

matzf (commented):
I see what you mean, and I agree that the API is not ideal. To me, one of the biggest weaknesses of the current API is that there is no way to tell from the type if the counter has been "fully" initialized with label values. 
But I don't think that just adding a completely separate way to do metrics improves the situation much. The newest approach to do metrics is then to handle "provisioning" the counters with all label values somewhere in the main application modules and pass only these to the libraries. This seems a good approach to me, but re-organizing the existing libraries to follow this exact approach will be a very large amount of effort (and thus will virtually never be completed). My suggestion thus is to improve the situation for both new code and the old libraries. I thought it was too daring to suggest before, but I actually thought that restricting the `Counter` interface even more, and make it act as _only_ a "builder" would be a good goal:
```go
type Counter interface {           // rename -> CounterVec / CounterBuilder / CounterWither / CounterLabeler? 
   With(labelValues ...string) SimpleCounter
}
type SimpleCounter interface {   // rename -> Counter?
     Add(delta float64)
}
```

I tried this as an experiment and it was surprisingly straight forward. There are some cases in the gateway that do require some simple reorganization because the same struct is (ab)-used for both the "builder" and the "built" counters, but otherwise it was a rather easy pass, just changing the types.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4527#-:-Ny0z0pW7RYri7O5X4O_:bcwoh0m)* status: 0 of 9 files reviewed, all discussions resolved

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4527)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

___
*[`pkg/metrics/prometheus.go` line 115 at r1](https://reviewable.io/reviews/scionproto/scion/4527#-Ny4bRDH5-iIBM5FuCp9:-Ny4bRDH5-iIBM5FuCpA:b5udzaj) ([raw file](https://github.com/scionproto/scion/blob/03bb6ddd5f3f11081c1abc62e120b3b1eab68185/pkg/metrics/prometheus.go#L115)):*
> ```Go
> }
> 
> func (g *simpleGauge) Set(value float64) {
> ```

Do we even need these wrapper structs? It seems that the prometheus interfaces already implement the SimpleX interfaces.
If we keep the structs, I think I'd make them value types, the additional pointer indirection seems unnecessary.

jiceatscion (commented):
I don't think this API was every designed with the intent that the client constructs the labeled counter reference on-the-fly, although it's certainly a prototyping simplification. However, the ability to build the label sets incrementally is convenient and I am no too sure what we gain by trying to remove it. For example in the router we have things like: 

```
metrics.DroppedPacketsTotal.MustCurryWith(ifLabels).MustCurryWith(scLabels).With(reasonMap)
```

Sure, one can always build a single label dic from all the input dics, but that's just extra boiler-plate code for the same result. Also, the fact that the router only uses the incremental building in fluent-style one-liners is a side effect of refactoring done for other reasons. Earlier there *was* code that built the label dics incrementally by passing curried vectors to other functions. Not vital, but why to we want to remove that ability? I still don't get what the cost of it is (unless used clearly improperly).

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4527#-:-Ny4gAlDBeGi0yKrAkjr:b-6zlz1m)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker and @matzf)

jiceatscion (commented):
I had not realized that using with() caused counters to become expensive during use. Or are you referring to cases of using with() every time the counter is changed? I guess what you really want is to make that faux-pas impossible to commit. Right?

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4527#-:-NyQxq_j4YbwJN7IO2eq:b1kpaga)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker and @matzf)

lukedirtwalker (commented):
No even counters initialized with "With" once and re-used will cause allocations:
So for example the counter Add implementation:
https://github.com/scionproto/scion/blob/15bdf63eab3b7e2191e14cb6bfac36637b24f00a/pkg/metrics/prometheus.go#L160-L162
makeLabels creates a dictionary which is then passed to prometheus.
Also in prometheus the label values have to be hashed we call the With function (via Add).

Use the following to compare the 2:
```go
func BenchmarkMetrics(b *testing.B) {
	auto := promauto.With(prometheus.NewRegistry())
	probesSent := auto.NewCounterVec(
		prometheus.CounterOpts{
			Name: "gateway_path_probes_sent_" + b.Name(),
			Help: "Number of path probes being sent.",
		},
		[]string{"isd_as", "remote_isd_as"},
	)
	create1 := func(local, remote addr.IA) metrics.Counter {
		return metrics.CounterWith(
			metrics.NewPromCounter(probesSent),
			"isd_as", local.String(),
			"remote_isd_as", remote.String(),
		)
	}
	create2 := func(local, remote addr.IA) metrics.SimpleCounter {
		return metrics.NewCounter(probesSent.With(prometheus.Labels{
			"isd_as":        local.String(),
			"remote_isd_as": remote.String(),
		}))
	}
	mustIA := func(s string) addr.IA {
		ia, err := addr.ParseIA(s)
		if err != nil {
			b.Fatal(err)
		}
		return ia
	}
	var c metrics.Counter
	_ = create1(mustIA("1-ff00:0:1"), mustIA("2-ff00:0:2"))
	c = create2(mustIA("1-ff00:0:1"), mustIA("2-ff00:0:2"))
	for n := 0; n < b.N; n++ {
		metrics.CounterInc(c)
	}
}
```

`create1` has 2 allocations, `create2` 0.

Yes I think we should prevent this in general, metrics should not cause allocations/or high computation overhead.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4527#-:-NyYxKjb3d7Gl0mH58fE:bbkbvk2)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @matzf)

matzf (commented):
We should not confuse the suboptimal implementation of the current interface with the flaws of the interface.
Even the current interface could be easily tweaked to avoid the `makeLabels` call in `Add`, by implementing `With` by means prometheus' built in `Curry` functions instead of tracking the label value set ourselves. Furthermore, the first call to `Add` could cache the internal counter object to avoid prometheus' internal map lookups for subsequent invocations on the same Counter object.

That being said, it seems that we mostly agree that the interface _is_ flawed on its own; the types do not convey any information about whether all the label values have already been set (and thus whether `Add` can be used), requiring the programmer to keep track of the state of the initialization. And, as @oncilla pointed out, it encourages stringly-typed, tight coupling of the initialization and the usage of the metrics. Without such a `With`, setting the label values is forced to be closer to the initialization of the metrics, using explicit (typed!) call backs where dynamic label values are required.

> The reason why I added a separate set of types was to prevent breaking changes. If we are fine with breaking changes, I would just do aways with the whole `With` method and let that be handled outside.

I think breaking changes here are ok; even this is on the periphery of the public API, we don't need to be overly conservative here IMO.
Removing the `With` generally sounds like a good plan to me. My impression was that this will require a lot of restructuring. If you think this is feasible, I'd fully agree that this would be ideal. The  suggestion with a split of `CounterBuilder` and `SimpleCounter` intended to at least partially disentangle the With/Add without having fully restructure this immediately.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4527#-:-NyZG9883QUxXwmgLphL:bw5eoxa)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @lukedirtwalker)

lukedirtwalker (commented):
let's see, I'll try it out, when I have a bit of time.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4527#-:-NyZo30w7iAJASDaxbve:bgk8k28)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @matzf)

lukedirtwalker (commented):
ok I wonder if we should just introduce a metrics/v2 package for the time being and gradually migrate to it. I fear that migrating everything in one pass will be too large of a PR to review at once.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4527#-:-Ny_I8fT8DdqxBalNdRN:bia19dy)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @matzf)

lukedirtwalker (commented):
I created a v2 package: https://github.com/scionproto/scion/pull/4530

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4527#-:-Ny_MylK19b8LEcFGn95:b-24qt33)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @matzf)

PR #4529, opened 2024-05-23
doc: fixup generated documentation files after removing dispatcher flags

matzf:
Run `bazel run //:write_all_source_files` to fix up generated documentation after removing the dispatcher flag from scion/scion-pki binaries.

This was missed during the PR because the existing check for this was not active in the CI. Extend the bazel configuration `test:lint` to enable running this in `make lint`.


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4529)

PR #4530, opened 2024-05-23
pkg/metrics: introduce v2

lukedirtwalker:
Introduce a v2 version of pkg/metrics. The current metrics package is not optimal:
- Using objects that have been called with With allocated on each increment/set value call.
- Having the option to call With ties consumer code with initialization code.
- There is no way whether a counter is fully initialized with all labels.

A full discussion can be found in
https://github.com/scionproto/scion/pull/4527.

The new v2 version is compatible with prometheus, so no explicit conversion functions are offered, instead the prometheus creation functions can be used directly.

So far the new library has been plugged in selected places. The intention is that it will eventually replace the current library completely, however doing so in a single sweep would be too extensive.


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)

oncilla (approved):
> Reviewed 28 of 28 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed, 3 unresolved discussions (waiting on @lukedirtwalker)

___
*[`pkg/metrics/v2/fakes.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_q4GX5nD7BK9Hr2z_:-Ny_q4GX5nD7BK9Hr2za:b-b09hwg) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/pkg/metrics/v2/fakes.go#L1)):*
> ```Go
> ```

nit: AFAICT, this is the same as metrics/v1.

You could move everything here and simply create an alias in metrics/v1
___
*[`pkg/metrics/v2/helper.go` line 20 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_pr0zBF6cfoOdmrse:-Ny_pr0zBF6cfoOdmrsf:bv86fng) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/pkg/metrics/v2/helper.go#L20)):*
> ```Go
> 
> // Timestamp expresses a timestamp in seconds since UNIX epoch.
> func Timestamp(ts time.Time) float64 {
> ```

nit: metrics/v1 could be an alias to this function to reduce duplication.
___
*[`private/periodic/legacy_metrics.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_qXAE1Qh_NSHbxBIH:-Ny_qXAE1Qh_NSHbxBII:bu3huyw) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/private/periodic/legacy_metrics.go#L1)):*
> ```Go
> package periodic
> ```

nit: license header.

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed, 3 unresolved discussions (waiting on @lukedirtwalker)

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed, 5 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/metrics/v2/fakes.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_q4GX5nD7BK9Hr2z_:-Nya5VL57VXZuhSbgRdK:bwt7g31) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/pkg/metrics/v2/fakes.go#L1)):*

nit: AFAICT, this is the same as metrics/v1.

You could move everything here and simply create an alias in metrics/v1
</blockquote></details>

Alternatively, remove the `With` from the `TestCounter`, and also remove any dealing with labels and metrics hierarchies from this file. Instead of `With`, a caller should just instantiate new counter objects directly, as that's the pattern that we generally want to see.
___
*[`pkg/metrics/v2/helper.go` line 20 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_pr0zBF6cfoOdmrse:-Nya5DuK7oNJkkAWOs7t:b-9timqf) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/pkg/metrics/v2/helper.go#L20)):*

nit: metrics/v1 could be an alias to this function to reduce duplication.
</blockquote></details>

IMO this can be hidden from the API, as there is already `GaugeSetTimestamp`.
___
*[`pkg/metrics/v2/metrics.go` line 43 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_iBLb9wS_eanbTDDv:-Ny_iBLb9wS_eanbTDDw:b-ya17xo) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/pkg/metrics/v2/metrics.go#L43)):*
> ```Go
> // metric implementation (e.g., Prometheus) to the application main function.
> //
> // The types are taken from the metrics interfaces in the go-kit/kit project
> ```

Nit: this is no longer accurate.
___
*[`pkg/metrics/v2/metrics_test.go` line 47 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_iVhk3qFoui8Fod6I:-Ny_iVhk3qFoui8Fod6J:bxkqzy9) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/pkg/metrics/v2/metrics_test.go#L47)):*
> ```Go
> 		}),
> 	}
> 	giant.MagicBeansEaten.Add(4)
> ```

I think it could be worth adding an example also for the suggested usage pattern for cases with "dynamic" label values.

Btw, the go style guide `doc/dev/style.go` should be updated to reference this example instead of the one in v1.

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed, 5 unresolved discussions (waiting on @lukedirtwalker)

lukedirtwalker (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: 21 of 29 files reviewed, all discussions resolved (waiting on @matzf and @oncilla)

___
*[`pkg/metrics/v2/fakes.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_q4GX5nD7BK9Hr2z_:-O-2AlCd6LdqQRKE3SWO:b-yyqe2b) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/pkg/metrics/v2/fakes.go#L1)):*

Alternatively, remove the `With` from the `TestCounter`, and also remove any dealing with labels and metrics hierarchies from this file. Instead of `With`, a caller should just instantiate new counter objects directly, as that's the pattern that we generally want to see.
</blockquote></details>

Done. Removed the `With` functionality
___
*[`pkg/metrics/v2/metrics.go` line 43 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_iBLb9wS_eanbTDDv:-O-2BNkT-OJA2McNceY9:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/pkg/metrics/v2/metrics.go#L43)):*

Nit: this is no longer accurate.
</blockquote></details>

Done.
___
*[`pkg/metrics/v2/metrics_test.go` line 47 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_iVhk3qFoui8Fod6I:-O-2Conm0XwSfB5uFL4q:bogvkmb) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/pkg/metrics/v2/metrics_test.go#L47)):*

I think it could be worth adding an example also for the suggested usage pattern for cases with "dynamic" label values.

Btw, the go style guide `doc/dev/style.go` should be updated to reference this example instead of the one in v1.
</blockquote></details>

Updated the doc. I didn't add another example. I think this one covers the most relevant things.
___
*[`private/periodic/legacy_metrics.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_qXAE1Qh_NSHbxBIH:-O-2Dl8W56CxgpUGPT73:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/private/periodic/legacy_metrics.go#L1)):*

nit: license header.
</blockquote></details>

Done.
___
*[`pkg/metrics/v2/helper.go` line 20 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_pr0zBF6cfoOdmrse:-O-2B2cC09OrJqHAqJTH:b-2y7b2c) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/pkg/metrics/v2/helper.go#L20)):*

IMO this can be hidden from the API, as there is already `GaugeSetTimestamp`.
</blockquote></details>

Done. Inlined to `GaugeSetTimestamp`

oncilla (approved):
> Reviewed 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lukedirtwalker)

___
*[`private/periodic/legacy_metrics.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_qXAE1Qh_NSHbxBIH:-O-2SmoZ86CyEZkNl-PK:b-1fjuhe) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/private/periodic/legacy_metrics.go#L1)):*

Done.
</blockquote></details>

Add newline.

Otherwise it is a package doc string and gofmt will mess with indents

matzf (approved):
:lgtm:

> Reviewed 21 of 28 files at r1, 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530#-:-O-57X1L3pJhzin-_MxS:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lukedirtwalker)

lukedirtwalker (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: 27 of 29 files reviewed, all discussions resolved (waiting on @matzf and @oncilla)

___
*[`private/periodic/legacy_metrics.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4530#-Ny_qXAE1Qh_NSHbxBIH:-O-6bJDu2GWIVJZGSLKk:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/691122f176ac14a80aa7f6edc6e9a9b289b65879/private/periodic/legacy_metrics.go#L1)):*

Add newline.

Otherwise it is a package doc string and gofmt will mess with indents
</blockquote></details>

Done.

matzf (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lukedirtwalker)

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @lukedirtwalker)

___
*[`pkg/metrics/v2/metrics.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4530#-O-eVgYC6U6A3t1qmOZp:-O-eVgYC6U6A3t1qmOZq:b2sl16n) ([raw file](https://github.com/scionproto/scion/blob/ce54ccac628e9f0baf7473fcbb903b793c2258a4/pkg/metrics/v2/metrics.go#L50)):*
> ```Go
> 	"time"
> )
> ```

General remark on this API. It seems semantically confusing (Non-blocking, just point me at the relevant doc if this has already been discussed):
* If a counter accumulates values monotonically, should it have a general Add(signed) method? That seems wrong, or a Counter is not actually monotonic.
* If Counter.Add and Gauge.Add have the same signature (and even semantics), why do we need two different Add interfaces? Can't an instance of Gauge also implement the Counter interface?
* What is the added value of these function wrappers on top of just calling the interface methods?

lukedirtwalker (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @jiceatscion)

___
*[`pkg/metrics/v2/metrics.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4530#-O-eVgYC6U6A3t1qmOZp:-O-eYU-B2PjeYpAtR79T:b-cy7567) ([raw file](https://github.com/scionproto/scion/blob/ce54ccac628e9f0baf7473fcbb903b793c2258a4/pkg/metrics/v2/metrics.go#L50)):*

General remark on this API. It seems semantically confusing (Non-blocking, just point me at the relevant doc if this has already been discussed):
* If a counter accumulates values monotonically, should it have a general Add(signed) method? That seems wrong, or a Counter is not actually monotonic.
* If Counter.Add and Gauge.Add have the same signature (and even semantics), why do we need two different Add interfaces? Can't an instance of Gauge also implement the Counter interface?
* What is the added value of these function wrappers on top of just calling the interface methods?
</blockquote></details>

Fair points. Here my replies:
- As said in the comment, we try to be as close to the prometheus API as possible. Those use float64 for counters. AFAIK there is no unsigned float type in Go, but for metrics it should be possible to add non integer increments.
- We can also embed the counter interface in the Gauge interface if you prefer. In the current way it's a bit easier to read because you immediately see the methods when looking at the Gauge interface and don't need the additional indirection to the counter.
- The function wrappers provide nil safety. You can pass in a nil counter without risking a panic.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @lukedirtwalker)

___
*[`pkg/metrics/v2/metrics.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4530#-O-eVgYC6U6A3t1qmOZp:-O-ec7lp8Og6_382KVil:bexu8jh) ([raw file](https://github.com/scionproto/scion/blob/ce54ccac628e9f0baf7473fcbb903b793c2258a4/pkg/metrics/v2/metrics.go#L50)):*

Fair points. Here my replies:
- As said in the comment, we try to be as close to the prometheus API as possible. Those use float64 for counters. AFAIK there is no unsigned float type in Go, but for metrics it should be possible to add non integer increments.
- We can also embed the counter interface in the Gauge interface if you prefer. In the current way it's a bit easier to read because you immediately see the methods when looking at the Gauge interface and don't need the additional indirection to the counter.
- The function wrappers provide nil safety. You can pass in a nil counter without risking a panic.
</blockquote></details>

So, it would seem that a counter and a Gauge have exactly the same interface. If any increment is valid, then both Counters and Gauges might as well implement both Set and Add. As for the nil-handling wrappers, this should not be required. Go methods can be called with a nil receiver just fine; up to them to check for nil. As far as I know that *is* the Go-ish way of doing it.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @lukedirtwalker)

___
*[`pkg/metrics/v2/metrics.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4530#-O-eVgYC6U6A3t1qmOZp:-O-efgWO59__3TKafZG8:bmpxyi0) ([raw file](https://github.com/scionproto/scion/blob/ce54ccac628e9f0baf7473fcbb903b793c2258a4/pkg/metrics/v2/metrics.go#L50)):*

So, it would seem that a counter and a Gauge have exactly the same interface. If any increment is valid, then both Counters and Gauges might as well implement both Set and Add. As for the nil-handling wrappers, this should not be required. Go methods can be called with a nil receiver just fine; up to them to check for nil. As far as I know that *is* the Go-ish way of doing it.
</blockquote></details>

Not mentioning that in a performance critical code path, do we really want to rely on, and so require, a nil-check?

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @jiceatscion and @lukedirtwalker)

___
*[`pkg/metrics/v2/metrics.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4530#-O-eVgYC6U6A3t1qmOZp:-O-eodbb9FXDVDAEx-Z2:b-ya914m) ([raw file](https://github.com/scionproto/scion/blob/ce54ccac628e9f0baf7473fcbb903b793c2258a4/pkg/metrics/v2/metrics.go#L50)):*

Not mentioning that in a performance critical code path, do we really want to rely on, and so require, a nil-check?
</blockquote></details>

Counters do not have set (see https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Counter)

In performance critical code, no one forces you to use the safe helper function

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @jiceatscion, @lukedirtwalker, and @oncilla)

___
*[`pkg/metrics/v2/metrics.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4530#-O-eVgYC6U6A3t1qmOZp:-O-eob7OAVvdop-kAqyX:br3y4qb) ([raw file](https://github.com/scionproto/scion/blob/ce54ccac628e9f0baf7473fcbb903b793c2258a4/pkg/metrics/v2/metrics.go#L50)):*
> Go methods can be called with a nil receiver just fine.

This is not the case for interfaces, unfortunately.

oncilla (commented):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @jiceatscion and @lukedirtwalker)

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @lukedirtwalker, @matzf, and @oncilla)

___
*[`pkg/metrics/v2/metrics.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4530#-O-eVgYC6U6A3t1qmOZp:-O-f5ddAAiE0Wqo4paAs:bkpfap7) ([raw file](https://github.com/scionproto/scion/blob/ce54ccac628e9f0baf7473fcbb903b793c2258a4/pkg/metrics/v2/metrics.go#L50)):*

> Go methods can be called with a nil receiver just fine.

This is not the case for interfaces, unfortunately.
</blockquote></details>

But you never do that, right. You have nil pointers to some conrete type or another and the interface object is created when you try to pass it to something that expects an interface type. Then the interface object isn't nil.

lukedirtwalker (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @jiceatscion, @matzf, and @oncilla)

___
*[`pkg/metrics/v2/metrics.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4530#-O-eVgYC6U6A3t1qmOZp:-O-fZDqd6D-dPL9v1Sf-:b-xbxjt2) ([raw file](https://github.com/scionproto/scion/blob/ce54ccac628e9f0baf7473fcbb903b793c2258a4/pkg/metrics/v2/metrics.go#L50)):*

But you never do that, right. You have nil pointers to some conrete type or another and the interface object is created when you try to pass it to something that expects an interface type. Then the interface object isn't nil.
</blockquote></details>

yes but if you have something like this:
```
type MyWorker struct {
     Executions metrics.Counter
}

func (w MyWorker) Run() {
    metrics.CounterInc(w.Executions, 1)
    // do some logic here
}


func main() {
     // for this worker I don't care about the metrics
     w := MyWorker{}
     for i := 0; i < 100; i1++ {
          w.Run()
     }
     // for this worker I want metrics
     w2 := MyWorker{Executions: prometheus.NewCounter(...)}
     // use w2
}
```

you will have a nil pointer in `w`.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @lukedirtwalker, @matzf, and @oncilla)

___
*[`pkg/metrics/v2/metrics.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4530#-O-eVgYC6U6A3t1qmOZp:-O-fa-nr6t1rnk9ZXi9P:b-zifovt) ([raw file](https://github.com/scionproto/scion/blob/ce54ccac628e9f0baf7473fcbb903b793c2258a4/pkg/metrics/v2/metrics.go#L50)):*

yes but if you have something like this:
```
type MyWorker struct {
     Executions metrics.Counter
}

func (w MyWorker) Run() {
    metrics.CounterInc(w.Executions, 1)
    // do some logic here
}


func main() {
     // for this worker I don't care about the metrics
     w := MyWorker{}
     for i := 0; i < 100; i1++ {
          w.Run()
     }
     // for this worker I want metrics
     w2 := MyWorker{Executions: prometheus.NewCounter(...)}
     // use w2
}
```

you will have a nil pointer in `w`.
</blockquote></details>

But isn't that considered an incorrect use of interfaces? Shouldn't w.Executions always be a concrete type? Well, ok, banning all cases of interface types in structures is asking a bit much, but that's what go commits to with the design of interfaces... That sucks; they violated their own principle that the zero value should be useful. There would be a way to prevent this accidental nil interface though: In your example: unexport MyWorker and have a default constructor. The default ctor can fill a default value for Executions that's a nil ptr to a concrete type. But I can't well demand that you do that. So objection retracted.

lukedirtwalker (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lukedirtwalker)

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4530)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lukedirtwalker)

PR #4539, opened 2024-06-05
router: avoid gc and copy

### LINKED ISSUE #4553: Router code makes unnecessary copies and tmp objects.
There are two main sources of data copy in the router:
* ProcessResults is a fairly large data structure (~20 bytes) that is passed by value up the stack to describe the disposition of a packet after processing. Every time we return one such we have to make a copy.
* Packet buffers are slices (therefore 24 bytes) passed by value many times per packet.
* A packet's internalized representation (srcAddr, dstAddr, ingress intf, and traffic type - 24 bytes) is passed around through channels and copied on the stack, many times per packet.
* SlowPathRequest is also a large data structure passed by value up the stack (~ 40 bytes) for all packets on the slow path (ie. requiring some scmp-related processing - constgituting a DOS exposure).

In spite of prior efforts, there is still a frequent source of allocating and freeing:
* srcAddr and dstAddr are net.UDPAddr passed and updated by reference without recycling. Every time a src or dst addr is updated, the old object is discarded, along with the array backing the embedded byte slice for the IP address. This internal reference is difficult to control. By contrast, the type netip.AddrPort, which contains the same information, is immutable, safe to pass by reference and garbage free when overwritten.

Overall, profiling showed that the simple ResolveLocal() function alone did consume 1% of the router's time, mostly due to allocation and freeing. Other allocations and copy account for at least 4% of time spent.

jiceatscion:
This is a branch from a branch. Just because the changes were getting a bit too many.
I am ok to merge this one directly, or to do it in two steps. Either way.

Fixes #4553


matzf:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)

jiceatscion:
These changes together bring about 20% throughput increase. The brload improvement was needed as it wasn't able to saturate the router any more. Not only did this cap the measured throughput, but also made the CI check fail.

matzf (changes requested):
> Reviewed 4 of 8 files at r2, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 4 of 18 files reviewed, 4 unresolved discussions (waiting on @jiceatscion)

___
*[`acceptance/router_benchmark/brload/mmsg.go` line 54 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DlKCv-OR1EiNWIHz7:-O0DlKCv-OR1EiNWIHz8:b-x23fha) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/acceptance/router_benchmark/brload/mmsg.go#L54)):*
> ```Go
> 
> 	for i := 0; i < numP; i++ {
> 		p := ps[i] // p = one packet
> ```

Nit: for i, p := range ps {
___
*[`private/underlay/conn/conn.go` line 42 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DlnOFE4Ku4Lt3U2V-:-O0DlnOFE4Ku4Lt3U2V0:b-q5xsey) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/private/underlay/conn/conn.go#L42)):*
> ```Go
> // Conn describes the API for an underlay socket
> // ReadFrom is the only remaining conn method that depends on net.UDPAddr. Since it is only used
> // by the soon-to-be-deprecated dispatcher, leaving it alone for now. Else we'd end-up
> ```

dispatcher is gone, ReadFrom and Write can be dropped.
___
*[`private/underlay/conn/conn.go` line 163 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DnxKtCHb30tqUTm56:-O0DnxKtCHb30tqUTm57:bf1omy4) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/private/underlay/conn/conn.go#L163)):*
> ```Go
> type connUDPBase struct {
> 	conn   *net.UDPConn
> 	Listen *netip.AddrPort
> ```

I find this a bit weird, I think AddrPort was really meant to be passed by value.
___
*[`private/underlay/conn/conn.go` line 281 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DnUWKDqcKBvO15Iog:-O0DnUWKDqcKBvO15Ioh:bkoutbd) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/private/underlay/conn/conn.go#L281)):*
> ```Go
> 	}
> 	// POSSIBLY EXPENSIVE CONVERSION: temp net.UDPAddr is allocated from heap.
> 	return c.conn.WriteTo(b, net.UDPAddrFromAddrPort(*dst))
> ```

use `c.conn.WriteToAddrPort`

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 4 of 18 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`acceptance/router_benchmark/brload/mmsg.go` line 54 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DlKCv-OR1EiNWIHz7:-O0DyGbOD0uANMbC7XFH:b-q7ab17) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/acceptance/router_benchmark/brload/mmsg.go#L54)):*

Nit: for i, p := range ps {
</blockquote></details>

Yeah, I've gotten overly defiant of for/range. This one seems fine.
___
*[`private/underlay/conn/conn.go` line 42 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DlnOFE4Ku4Lt3U2V-:-O0DvAeY5QgGYV3ZnuFA:b-1e1tba) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/private/underlay/conn/conn.go#L42)):*

dispatcher is gone, ReadFrom and Write can be dropped.
</blockquote></details>

done
yeay!
___
*[`private/underlay/conn/conn.go` line 163 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DnxKtCHb30tqUTm56:-O0DpVNk5lzQAu_n8iei:b-jvzhd1) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/private/underlay/conn/conn.go#L163)):*

I find this a bit weird, I think AddrPort was really meant to be passed by value.
</blockquote></details>

AddrPort is better than UDPAddr in more than one way. The fact that it doesn't have internal pointers and is smaller, make it more economical to pass by value, but it is also immutable, so it is safe to share. Passing by reference is still cheaper than by value, just due to size. In this case it matters little, performance-wise, but I just thought I'd use the same style everywhere, as much as feasible.
___
*[`private/underlay/conn/conn.go` line 281 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DnUWKDqcKBvO15Iog:-O0DwdFB1xD79i7XRYvS:b1pteyi) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/private/underlay/conn/conn.go#L281)):*

use `c.conn.WriteToAddrPort`
</blockquote></details>

Cool. I had completely forgotten about that one.

matzf (changes requested):
> Reviewed 1 of 13 files at r1, 1 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 5 of 18 files reviewed, 7 unresolved discussions (waiting on @jiceatscion)

___
*[`router/dataplane.go` line 790 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0E1RtZ45GYlteOjnhT:-O0E1Rt_33E7lgsDxF7I:bc4t69h) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/router/dataplane.go#L790)):*
> ```Go
> 			continue
> 		case DISCARD: // Packets that don't need more processing (e.g. BFD), plus all errors.
> 			metrics.DroppedPacketsInvalid.Inc()
> ```

BFD is not invalid.  
___
*[`router/dataplane.go` line 890 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0E1g319I2WlAVDIqQF:-O0E1g319I2WlAVDIqQG:b8ea60i) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/router/dataplane.go#L890)):*
> ```Go
> func (p *slowPathPacketProcessor) reset() {
> 	if err := p.buffer.Clear(); err != nil {
> 		// TODO(jiceatscion): ...and we don't care?
> ```

I think this could panic, it really cannot happen and if it does happen it's sort of unrecoverable.
___
*[`router/dataplane.go` line 1099 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0EWYbr4rk8YKp4evrU:-O0EWYbr4rk8YKp4evrV:bu63fxl) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/router/dataplane.go#L1099)):*
> ```Go
> // Convenience function to log an error and return the DISCARD disposition.
> // We do almost nothing with errors, so, we shouldn't invest in creating them.
> func errorDiscard(err error) PacketDisp {
> ```

great idea to have this centralized. Should we change the errors into strings?
___
*[`router/dataplane.go` line 1184 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0EW5we4ngbbaSNK-fj:-O0EW5we4ngbbaSNK-fk:b13cr4h) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/router/dataplane.go#L1184)):*
> ```Go
> 	if v, ok := p.d.bfdSessions[ifID]; ok {
> 		v.ReceiveMessage(bfd)
> 		return DISCARD // All's fine. That packet's journey ends here.
> ```

Sepaate disposition could make sense. E.g. just OK, or DONE.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 5 of 18 files reviewed, 7 unresolved discussions (waiting on @matzf)

___
*[`router/dataplane.go` line 790 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0E1RtZ45GYlteOjnhT:-O0E_2K8BliBFwtMQ0Lm:b7pzdpv) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/router/dataplane.go#L790)):*

BFD is not invalid.
</blockquote></details>

Yeah, you're right. Took me a while to realize what you were objecting to. Fixed.
___
*[`router/dataplane.go` line 890 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0E1g319I2WlAVDIqQF:-O0EeI6t-x_D4J-u-f1E:b-m808vp) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/router/dataplane.go#L890)):*

I think this could panic, it really cannot happen and if it does happen it's sort of unrecoverable.
</blockquote></details>

Yup. Done, here and the other spot.
___
*[`router/dataplane.go` line 1099 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0EWYbr4rk8YKp4evrU:-O0EfBkDByRZ4Ly9Ymi4:bzejp17) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/router/dataplane.go#L1099)):*

great idea to have this centralized. Should we change the errors into strings?
</blockquote></details>

Meh...any benefit? The error interfaces only get String()'ed if the log level is debug, so it doesn't mater. Does it? At least that way, the messages have a distinct type and are different even if we give them the same description string. FWIW.
___
*[`router/dataplane.go` line 1184 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0EW5we4ngbbaSNK-fj:-O0Eg7B_5uFt5ysIY6uL:b-4q4aai) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/router/dataplane.go#L1184)):*

Sepaate disposition could make sense. E.g. just OK, or DONE.
</blockquote></details>

Oh... right. I understand your comment earlier about DISCARD'ed BFD packets; yes, the statistics are wrong. Yes, I've added a "DONE" disposition for that.

matzf (commented):
> Reviewed 1 of 13 files at r1, 1 of 1 files at r3, 1 of 2 files at r6, 1 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 9 of 18 files reviewed, 8 unresolved discussions (waiting on @jiceatscion)

___
*[`private/underlay/conn/conn.go` line 163 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DnxKtCHb30tqUTm56:-O0Naqou-WFFVt8oRdYI:b-a9fun1) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/private/underlay/conn/conn.go#L163)):*

AddrPort is better than UDPAddr in more than one way. The fact that it doesn't have internal pointers and is smaller, make it more economical to pass by value, but it is also immutable, so it is safe to share. Passing by reference is still cheaper than by value, just due to size. In this case it matters little, performance-wise, but I just thought I'd use the same style everywhere, as much as feasible.
</blockquote></details>

Good point. 
The style consistency that you also mention seems important though. Everywhere else, we pass Addr and AddrPort by value, and so does the std lib.
Regarding performance, I don't think the trade off is so clear. Sure, AddrPort is fatter than a pointer, but pointer chasing has its own cost. And probably more importantly, anything that we pass around as a reference has the potential to escape to the heap.
Passing by reference in a few places where the overhead of copying is more important makes sense, but as a default passing by value seems the safer bet to me.
___
*[`router/dataplane.go` line 1099 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0EWYbr4rk8YKp4evrU:-O0OWfgZAQXS7G_5gWmZ:baxvbma) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/router/dataplane.go#L1099)):*

Meh...any benefit? The error interfaces only get String()'ed if the log level is debug, so it doesn't mater. Does it? At least that way, the messages have a distinct type and are different even if we give them the same description string. FWIW.
</blockquote></details>

To avoid the temptation of creating an expensive error with `serrors.With`  just for logging.
But yeah with just a single string, the temptation might then be to create formatted strings instead... To make this useful even for the case where we want to include debug info without too much overhead, we could  make the interface of this function look like the structured logging APIs;
```
func errorDiscard(reason string, ctx ... any) PacketDisp {
  log.Debug("Discarding packet", "reason", reason, ctx...);
  return DISCARD
}
```

With this it's fine to append additional context information about the packet with basically no overhead when the debug log is disabled.

Not so important though, if you're not convinced just leave it as it is.
___
*[`router/dataplane.go` line 95 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0OrW9I0wlkamZ-_ZM9:-O0OrW9I0wlkamZ-_ZMA:bthlqk3) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane.go#L95)):*
> ```Go
> 	SLOWPATH
> 	DONE
> )
> ```

PacketDisp needs not be exported --> `packetDisp`. Btw, it's not entirely obvious what the abbreviation stands for (dispatch? dispute? disparage?). If it's not too annoying, I'd suggest to spell it out `packetDisposition`. Otherwise a tiny comment explaing this could help.

Personally I quite like the upper case names here,  seems appropriate, even if the style guide explicitly wants `mixedCaps`. There is semantics to this though (exported symbol) so perhaps prepend a lowercase type abbreviation (`pdDISCARD` )?

(go enums _still_  just suck)
___
*[`router/dataplane.go` line 114 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0OuiAa5LnQ-6RNPUtV:-O0OuiAa5LnQ-6RNPUtW:bb4e4mv) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane.go#L114)):*
> ```Go
> 	dstAddr *net.UDPAddr
> 	// Additional metadata in case the packet is put on the slow path. Updated in-place.
> 	slowPathRequest *slowPathRequest
> ```

(related to comment elsewhere about alignment / padding)

slowPathRequest is actually quite small. If we drop the useless `cause and `ia` fields,  it will be  smaller than 16 bytes. With this, it seems that inlining the slowPathRequest here could be an ideal way to "pad" the `packet` struct to 64 bytes.

The `cause` field is simply redundant; it's just used as a boolean to figure out if this SCMP message was triggered by a processing error or not. `cause` is non-nil if and only if `typ == slowPathSCMP`.
The `ia` field is always the local and references can be replaced with `p.d.localIA`.
___
*[`router/dataplane.go` line 137 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0NnNfO0a7uEeunbDwV:-O0NnNfO0a7uEeunbDwW:b-5xjsa8) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane.go#L137)):*
> ```Go
> // * Economical: clearing dstAddr (otherwise it would need to be cleared in many code paths).
> // * Economical: clearing egress ID (otherwise it would need to be cleared in many code paths).
> // * Economical: setting disp to DISCARD (for the same reason).
> ```

Nit, outdated, disp no longer here.
Maybe flip this around, reset everything and highlight the exceptions.
___
*[`router/dataplane.go` line 188 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0NoHF04PiLBOc3xnq0:-O0NoHF04PiLBOc3xnq1:b2z24eg) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane.go#L188)):*
> ```Go
> 	//   by value.
> 	// Conclusion: put all the meta-data that we need in a struct along with the byte buffer and
> 	// make a pool for all of it. The pool oinly contains references.
> ```

The referenced design document does not go into details of how the packet metadata is passed around, so there is no need to point this out as difference from that document. I'd suggest to summarize this, e.g.: "packets passed as around as references to avoid repeated copying of even this relatively small struct"
___
*[`router/dataplane.go` line 631 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0Npjq61LN_XbF-LxyG:-O0Npjq61LN_XbF-LxyH:b-tarw95) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane.go#L631)):*
> ```Go
> 	d.packetPool = make(chan *packet, poolSize)
> 	pktBuffers := make([][bufSize]byte, poolSize)
> 	pktStructs := make([]packet, poolSize)
> ```

Allocating these packets in a compact buffer potentially leads to structs straddling multiple cache lines. We write to these structs a lot from many different threads, so false sharing of cache lines is a concern. 

Allocations on the heap do seem to be 64 byte aligned; i didn't find any real references for this, so this could be unreliable and might need to be asserted e.g. in tests.
We could either allocate the packets individually, or, if we allocate the packets in a contiguous slice, we should consider padding the packet struct to 64 bytes (currently 56).
___
*[`router/dataplane.go` line 1006 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0OWZ4n2ELhYHBOCJ4s:-O0OWZ4n2ELhYHBOCJ4t:b-w5gdtc) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane.go#L1006)):*
> ```Go
> 		// in making a temporary copy. More crucially, we want each message to reference the
> 		// destination address, packet.dstAddr, not a copy (and surely not a local variable!). So,
> 		// a local copy of each item we traverse is thoroughly useless.
> ```

This change no longer makes sense, the pkts slice only contains references.

matzf (commented):
> Reviewed 3 of 13 files at r1, 3 of 4 files at r4, 1 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 16 of 18 files reviewed, 13 unresolved discussions (waiting on @jiceatscion)

___
*[`router/dataplane_internal_test.go` line 200 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0P7KAE6AJtc43OzZJ2:-O0P7KAE6AJtc43OzZJ3:bmhgrev) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane_internal_test.go#L200)):*
> ```Go
> 		pkt.srcAddr = &net.UDPAddr{} // Receiver always sets this.
> 		pkt.dstAddr = dstAddr
> 		pkt.ingress = 0
> ```

I may be mistaken, but I believe that the forwarder (that we're testing here) never looks at ingress and srcAddr.  I think these can be dropped here to declutter the test (it's confusing enough).

It might make sense, however, to fill dstAddr inplace to simulate what the actual processor is doing, to ensure we test the case with setting IP to an empty slice.

Also, please keep the lines initializing the payload together. So, 

```
pkt := <-dp.packetPool
pkt.reset()
if i < 100 {
   pkt.dstAddr.IP = pkt.dstAddr.IP[:4]
   copy(pkt.dstAddr.IP, testIP)
}
pkt.rawPacket[0] = byte(i)
pkt.rawPacket = pkt.rawPacket[:1]
```
___
*[`router/dataplane_internal_test.go` line 535 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0P5XfV1gpxiwmKNsf0:-O0P5XfV1gpxiwmKNsf1:bml9pig) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane_internal_test.go#L535)):*
> ```Go
> 			pkt.srcAddr = &net.UDPAddr{} // The receiver always sets this.
> 			pkt.rawPacket = pkt.rawPacket[:len(rp)]
> 			copy(pkt.rawPacket, rp)
> ```

Seems somewhat verbose.



_Suggestion:_
```Go
			pkt := packet{
			  rawPacket: rp,
			  ingress: tc.srcInterface,
			  srcAddr: &net.UDPAddr{},	
			}
```
___
*[`router/dataplane_test.go` line 1602 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0P0dnW5MJ9eKenMJCm:-O0P0dnW5MJ9eKenMJCn:b-qfx2bd) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane_test.go#L1602)):*
> ```Go
> 				return
> 			}
> 			out := pkt.ToIpv4Msg()
> ```

I think at this point, this ip.Message type is just getting in the way here. It used to be the input and output of the processPkt function, but now were translating both ways and just obscure what we're testing.

I think we can simplify the setup here; we don't need to export the Packet type at all. We can do all of that in  the ProcessPkt test-wrapper, if we change it's signature to something like `func (dp *dataplane) ProcessPkt(srcAddr netip.AddrPort, srcInterface int, packet []byte) (disposition, dstAddr netip.AddrPort, dstInterface int, outPacket []byte)`

The `mockMsg` function becomes `func(bool) []byte` and source/dest address can be defined as fields in the test case struct, just like the ingress/egress interface.

It may seem like a lot of churn, but all the test cases use the same helper functions, so I think it's actually quite quick.
___
*[`router/export_test.go` line 123 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0P3YtC5x0LHcaxVDY_:-O0P3YtC5x0LHcaxVDYa:bgn6s0v) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/export_test.go#L123)):*
> ```Go
> }
> 
> func (d *DataPlane) ProcessPkt(pkt *Packet) PacketDisp {
> ```

As noted in `dataplane_test.go`, I think we don't need to expose the Packet type at all, if we instead expand the definition of this `ProcessPkt` function to take and return all the metadata explicitly.
___
*[`router/svc.go` line 72 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0P3vx98CRgtaJ0neY3:-O0P3vx98CRgtaJ0neY4:bams37z) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/svc.go#L72)):*
> ```Go
> func (s *services) index(a netip.AddrPort, addrs []netip.AddrPort) (int, bool) {
> 	for i, o := range addrs {
> 		if a.Addr() == o.Addr() && a.Port() == o.Port() {
> ```

`a == o`

(or, just use the newfangled `slices.Index(addrs, a)`)

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 11 of 18 files reviewed, 3 unresolved discussions (waiting on @matzf)

___
*[`private/underlay/conn/conn.go` line 163 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DnxKtCHb30tqUTm56:-O0il9x0-zxZdHm0drm4:b-yay9l0) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/private/underlay/conn/conn.go#L163)):*

Ok. It's true, following pointers costs too. I have a hard time figuring how to compare one cost to the other, including cache cost, the number of times the thing is copied, and how many times it is accessed. I think I'll try both and just see what the benchmark says. I'll push the rest first. Stay tuned.
</blockquote></details>

Taking a closer look, I think that on the critical path there isn't that much passing around of AddrPort left, I don't remember what the code looked like before we made all these changes to remove ProcessResult and shrink slowpathRequest. May be there was more of it, then. Anyway, you're right, there's no reason to bother with the passing by reference anymore. I'll fix it.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 7 of 19 files reviewed, 2 unresolved discussions (waiting on @matzf)

___
*[`private/underlay/conn/conn.go` line 163 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DnxKtCHb30tqUTm56:-O0jTL7c-331JUx_VNp_:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/private/underlay/conn/conn.go#L163)):*

Taking a closer look, I think that on the critical path there isn't that much passing around of AddrPort left, I don't remember what the code looked like before we made all these changes to remove ProcessResult and shrink slowpathRequest. May be there was more of it, then. Anyway, you're right, there's no reason to bother with the passing by reference anymore. I'll fix it.
</blockquote></details>

done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 6 of 19 files reviewed, 1 unresolved discussion (waiting on @matzf)

___
*[`router/dataplane_test.go` line 1602 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0P0dnW5MJ9eKenMJCm:-O0nxA3v1mrRuasxKVHQ:bhj9wrh) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane_test.go#L1602)):*

I tried to make sense of that spaghetti... I don't see clearly what to best do about it yet. What you propose isn't obvious to me and I don't much like we'd end-up testing an exported function that is not a straight wrapper of the original. I'll push the rest first, so I can work just on this. Stay tuned.
</blockquote></details>

Done. Squeaky clean.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 6 of 19 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`router/export_test.go` line 123 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0P3YtC5x0LHcaxVDY_:-O0nxV-QDddJWoJX09ay:bisya65) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/export_test.go#L123)):*

As noted in `dataplane_test.go`, I think we don't need to expose the Packet type at all, if we instead expand the definition of this `ProcessPkt` function to take and return all the metadata explicitly.
</blockquote></details>

done. Albeit not that way.

matzf (commented):
> Reviewed 1 of 5 files at r8, 6 of 11 files at r9, 4 of 5 files at r10, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: all files reviewed, 3 unresolved discussions (waiting on @jiceatscion)

___
*[`private/underlay/conn/conn.go` line 163 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0DnxKtCHb30tqUTm56:-O0t-b2r4uWQ5zmxoTKU:b-u87p7v) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/private/underlay/conn/conn.go#L163)):*

done
</blockquote></details>

Neat, thanks!
___
*[`private/underlay/conn/conn.go` line 71 at r11](https://reviewable.io/reviews/scionproto/scion/4539#-O0t-uMe6yQhfBhmiEND:-O0t-uMe6yQhfBhmiENE:b-510j80) ([raw file](https://github.com/scionproto/scion/blob/801096c49258b86a05e599282d021b0e5930bb09/private/underlay/conn/conn.go#L71)):*
> ```Go
> 		a = remote
> 	}
> 	if !a.IsValid() {
> ```

Did you consciously  change the logic here? This was `if listen == nil && remote == nil {`, which would tranlsate to `if !listen.IsValid() && !remote.IsValid() {`.
___
*[`router/dataplane.go` line 1099 at r5](https://reviewable.io/reviews/scionproto/scion/4539#-O0EWYbr4rk8YKp4evrU:-O0t3IJx2U4Z4uejcNOL:b-ihar3c) ([raw file](https://github.com/scionproto/scion/blob/a340ee4e6662bf232d099ea4cfac8997ab5467f4/router/dataplane.go#L1099)):*

It makes sense. I wasn't able to write it the way you did. For some reason the Go compiler isn't able to call a variadic function with a mixture of explicit arguments and a list. I would have had to generate a new slice, I think. Yuck. Instead, I did something a bit simpler, at the cost of passing the extra "error" string in al the existing cases.
</blockquote></details>

Argh, I had naively assumed that the variadic args support this. Bah... Your solution looks good :thumbsup: 
___
*[`router/dataplane.go` line 95 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0OrW9I0wlkamZ-_ZM9:-O0t0pQC1kPl14EGwjBn:b4c9lyr) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane.go#L95)):*

Yeah, changed the names. Looking into the allcaps issue, I found that the directive to use mixedcaps instead is strong (and taken seriously by zealots) in the style guide. Given that using something else is only a matter of taste, I'll stay away from dogmatic arguments and follow the style guide. That's what it's for. I did include a "p" prefix so there can be at least one capital without unwanted exporting.

As a redittor put it. You should say "go has no enums" and that's the issue, right there. :-) As far as the naming issue goes, the real problem is more that "exporting" is done by capitalizing. That's so f*g sloppy!
</blockquote></details>

:thumbsup: 
___
*[`router/dataplane.go` line 137 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0NnNfO0a7uEeunbDwV:-O0t1CWsCTsYa8CaEbQI:b4medmw) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane.go#L137)):*

removed the outdated info. It's still more concise than describing what is left out.
</blockquote></details>

Leave fewer things out from resetting then? :wink:
It just seems very dangerous to me, and extremely hard to keep track of. Ideally we'd ensure that we cannot forget to clean up a field, like:
```go
func (p *packet) reset() {
    p.dstAddr.IP = p.dstAddr.IP[0:0]
    *p := packet {
        rawPacket: p.rawPacket[:cap(p.rawPacket)],
        dstAddr: p.dstAddr,
        srcAddr: p.srcAddr,
    }
}
```

Looking at the output of this, it seems that the compiler does not optimize this well and pessimistically inserts an unnecessary gc write barrier. 
Resetting all the fields explicitly looks better, and I think this would be a reasonable compromise; overhead over the current `reset` can't be significant, and I believe it still decreases the overall risk of forgetting to set a field before using it.
```go
func (p *packet) reset() {
    p.rawPacket = p.rawPacket[:cap(p.rawPacket)]
    p.dstAddr.IP = p.dstAddr.IP[0:0]
    p.slowPathRequest = slowPathRequest{}
    p.trafficType = 0
    p.ingress = 0
    p.egress = 0
}
```

(See reset0 vs reset1 vs reset2 in this [godbolt](https://godbolt.org/z/zxdGo9cj3))
___
*[`router/dataplane.go` line 631 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0Npjq61LN_XbF-LxyG:-O0t3ouDAqYmRVrDtRdk:b-86wrvg) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane.go#L631)):*
Cool.

> Note that recent CPUs seem to make cache line optimization surprisingly irrelevant, but I think it doesn't hurt either.

I guess I don't have a recent CPU :wink: Anyway I didn't  necessarily mention this as a big optimization directly, but more avoid the pitfall of false sharing once we crank up parallelism (if i recall correctly, the benchmarks are run on relatively small number of cores).
___
*[`router/dataplane.go` line 133 at r11](https://reviewable.io/reviews/scionproto/scion/4539#-O0t2Y-B-j2MGvvxMnNq:-O0t2Y-B-j2MGvvxMnNr:b-1fwi1z) ([raw file](https://github.com/scionproto/scion/blob/801096c49258b86a05e599282d021b0e5930bb09/router/dataplane.go#L133)):*
> ```Go
> 	p.rawPacket = buffer[:]
> 	p.dstAddr = &net.UDPAddr{IP: make(net.IP, net.IPv6len)}
> 	p.slowPathRequest = slowPathRequest{}
> ```

Nit: this initialization of slowPathRequest is now redundant.
___
*[`router/dataplane.go` line 1296 at r11](https://reviewable.io/reviews/scionproto/scion/4539#-O0t8Td25mpWCwChksDi:-O0t8Td25mpWCwChksDj:b-1mbeav) ([raw file](https://github.com/scionproto/scion/blob/801096c49258b86a05e599282d021b0e5930bb09/router/dataplane.go#L1296)):*
> ```Go
> // Golang gives precious little guarantees about alignment (and so padding) that it uses.
> // Best to round-up the size to something predictable (given non-absurd Go added constraints)
> // We make this struct 16 bytes long.
> ```

I don't quite see how the explicit padding is predictable if the alignment of the other fields is unclear; for all we know they might all be 8 byte aligned. 
Either way, it seems very easy to forget this when this or the packet struct is updated, so perhaps move the two struct definitions closer together "physically".
And I would add a test that asserts the sizeof for both structs.
___
*[`router/dataplane_internal_test.go` line 200 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0P7KAE6AJtc43OzZJ2:-O0tBhxQBem6SduEG9aV:b-hl70ui) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane_internal_test.go#L200)):*

Moved the pool length assert as you requested and fill addr in place. I disagree with not setting some fields. I've been irritated so many times by tests that broke on me just because of an unnecessary (and all of a sudden untrue) assumption that I try to avoid them when I can.
</blockquote></details>

Fair enough. There is still this `rp` var at the beginning, and then the rawPacket is set / truncated at the end. Please move this together, and avoid the unnecessary `rp`.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

___
*[`private/underlay/conn/conn.go` line 71 at r11](https://reviewable.io/reviews/scionproto/scion/4539#-O0t-uMe6yQhfBhmiEND:-O0tbAR0AdnzPx6yYoWd:bl4dd3n) ([raw file](https://github.com/scionproto/scion/blob/801096c49258b86a05e599282d021b0e5930bb09/private/underlay/conn/conn.go#L71)):*

Did you consciously  change the logic here? This was `if listen == nil && remote == nil {`, which would tranlsate to `if !listen.IsValid() && !remote.IsValid() {`.
</blockquote></details>

Nah, it's the same thing. We where checking the values of both remote and listen, while what we actually care about is the value of a, which we just calculated: a is invalid IFF both listen and remote are.
___
*[`router/dataplane.go` line 137 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0NnNfO0a7uEeunbDwV:-O0te0MJEQRWK7q3CrTm:b-gyghit) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane.go#L137)):*

Leave fewer things out from resetting then? :wink:
It just seems very dangerous to me, and extremely hard to keep track of. Ideally we'd ensure that we cannot forget to clean up a field, like:
```go
func (p *packet) reset() {
    p.dstAddr.IP = p.dstAddr.IP[0:0]
    *p := packet {
        rawPacket: p.rawPacket[:cap(p.rawPacket)],
        dstAddr: p.dstAddr,
        srcAddr: p.srcAddr,
    }
}
```

Looking at the output of this, it seems that the compiler does not optimize this well and pessimistically inserts an unnecessary gc write barrier. 
Resetting all the fields explicitly looks better, and I think this would be a reasonable compromise; overhead over the current `reset` can't be significant, and I believe it still decreases the overall risk of forgetting to set a field before using it.
```go
func (p *packet) reset() {
    p.rawPacket = p.rawPacket[:cap(p.rawPacket)]
    p.dstAddr.IP = p.dstAddr.IP[0:0]
    p.slowPathRequest = slowPathRequest{}
    p.trafficType = 0
    p.ingress = 0
    p.egress = 0
}
```

(See reset0 vs reset1 vs reset2 in this [godbolt](https://godbolt.org/z/zxdGo9cj3))
</blockquote></details>

awrite, I'm being pennywise... Fixed.
___
*[`router/dataplane.go` line 133 at r11](https://reviewable.io/reviews/scionproto/scion/4539#-O0t2Y-B-j2MGvvxMnNq:-O0tbxobEXJ2uIWxrPSL:b-6mv472) ([raw file](https://github.com/scionproto/scion/blob/801096c49258b86a05e599282d021b0e5930bb09/router/dataplane.go#L133)):*

Nit: this initialization of slowPathRequest is now redundant.
</blockquote></details>

done thanks
___
*[`router/dataplane.go` line 1296 at r11](https://reviewable.io/reviews/scionproto/scion/4539#-O0t8Td25mpWCwChksDi:-O0teQZkBDlk6L-d-ibv:b-3kjjl1) ([raw file](https://github.com/scionproto/scion/blob/801096c49258b86a05e599282d021b0e5930bb09/router/dataplane.go#L1296)):*

I don't quite see how the explicit padding is predictable if the alignment of the other fields is unclear; for all we know they might all be 8 byte aligned. 
Either way, it seems very easy to forget this when this or the packet struct is updated, so perhaps move the two struct definitions closer together "physically".
And I would add a test that asserts the sizeof for both structs.
</blockquote></details>

The assert is good. Added it. I make alignment predictable mostly by giving Go no choice. Far as I know structs are aligned on 8 byte boundaries (and I couldn't even find that documented) and nothing else is clearly stated. I think there's a bug in the doc they said "structs are aligned on the greatest of their element's alignment" where they meant "greatest element size". So, I can only assume that Go does not add absurd paddings. Not sure how to explain that concisely for a comment. Brought the two structs together, too. Ah and... that assert would have found that I was off by 3. I had forgotten to enlarge the size of trafficType like the comment promised.
___
*[`router/dataplane_internal_test.go` line 200 at r7](https://reviewable.io/reviews/scionproto/scion/4539#-O0P7KAE6AJtc43OzZJ2:-O0tklvlDMdLfwXpX0Wf:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/75556c39f88dd91dc418a4665a9fdd59e97d9628/router/dataplane_internal_test.go#L200)):*

Fair enough. There is still this `rp` var at the beginning, and then the rawPacket is set / truncated at the end. Please move this together, and avoid the unnecessary `rp`.
</blockquote></details>

done

matzf (commented):
> Reviewed 2 of 3 files at r12.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 19 of 20 files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

___
*[`private/underlay/conn/conn.go` line 71 at r11](https://reviewable.io/reviews/scionproto/scion/4539#-O0t-uMe6yQhfBhmiEND:-O1GNogT8U1lUkhLRpGy:b-1qidc2) ([raw file](https://github.com/scionproto/scion/blob/801096c49258b86a05e599282d021b0e5930bb09/private/underlay/conn/conn.go#L71)):*

Nah, it's the same thing. We where checking the values of both remote and listen, while what we actually care about is the value of a, which we just calculated: a is invalid IFF both listen and remote are.
</blockquote></details>

Right, of course, sorry.
___
*[`router/dataplane.go` line 1296 at r11](https://reviewable.io/reviews/scionproto/scion/4539#-O0t8Td25mpWCwChksDi:-O1GX7Tx5w61D1L26pKP:bqteu80) ([raw file](https://github.com/scionproto/scion/blob/801096c49258b86a05e599282d021b0e5930bb09/router/dataplane.go#L1296)):*
> Far as I know structs are aligned on 8 byte boundaries

I don't believe this is generally true. E.g. a struct with only byte fields will be 1-byte aligned. As you note, the spec is vague and does not specify alignment for individual primitive types. But still, to me the statement "structs are aligned on the greatest of their element's alignment" seems to make perfect sense. The size of the primitive type will typically correspond directly to its alignment, but it seems they wanted to keep that open (or maybe avoid having to put architecture specific details into the specs). 

My point was, if we consider the alignment rules and thus the padding to be unpredictable, then we cannot predictably insert the padding field explicitly either. If the compiler would decide to align the [5]byte field to, say, 8 bytes (which is allowed according to the specs which require alignment _at least_ 1) it would insert padding before the padding field.
___
*[`router/dataplane.go` line 639 at r12](https://reviewable.io/reviews/scionproto/scion/4539#-O1GSaqV8X4fnKQJoP2Y:-O1GSaqV8X4fnKQJoP2Z:bg2zquc) ([raw file](https://github.com/scionproto/scion/blob/3018e91e61f66a2683d9d3e75c56786c95f54c78/router/dataplane.go#L639)):*
> ```Go
> 	if unsafe.Sizeof(packet{}) != 64 {
> 		panic("The type packet struct is incorrectly defined")
> 	}
> ```

The runtime check scares me a bit because we cross-compile but don't currently test our builds on all platforms.
We can turn this into a compile time assertion sort of hack. This fails if sizeof != 64 because negative constants cannot be assigned to unsigned ints.
```go
// assert sizeof(packet) == 64:
const _ uintptr = 64 - unsafe.Sizeof(packet{})  // assert 64 >= sizeof(packet)
const _ uintptr = unsafe.Sizeof(packet{}) - 64  // assert sizeof(packet) >= 64
```

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 19 of 20 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`router/dataplane.go` line 639 at r12](https://reviewable.io/reviews/scionproto/scion/4539#-O1GSaqV8X4fnKQJoP2Y:-O1GhUa9EyUWP1Kjxpz0:b-6uja5o) ([raw file](https://github.com/scionproto/scion/blob/3018e91e61f66a2683d9d3e75c56786c95f54c78/router/dataplane.go#L639)):*

The runtime check scares me a bit because we cross-compile but don't currently test our builds on all platforms.
We can turn this into a compile time assertion sort of hack. This fails if sizeof != 64 because negative constants cannot be assigned to unsigned ints.
```go
// assert sizeof(packet) == 64:
const _ uintptr = 64 - unsafe.Sizeof(packet{})  // assert 64 >= sizeof(packet)
const _ uintptr = unsafe.Sizeof(packet{}) - 64  // assert sizeof(packet) >= 64
```
</blockquote></details>

Good idea and done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4539)* status: 19 of 20 files reviewed, all discussions resolved (waiting on @matzf)

___
*[`router/dataplane.go` line 1296 at r11](https://reviewable.io/reviews/scionproto/scion/4539#-O0t8Td25mpWCwChksDi:-O1GhkZc9_SaIL-9PEKj:b-fgtu10) ([raw file](https://github.com/scionproto/scion/blob/801096c49258b86a05e599282d021b0e5930bb09/router/dataplane.go#L1296)):*

> Far as I know structs are aligned on 8 byte boundaries

I don't believe this is generally true. E.g. a struct with only byte fields will be 1-byte aligned. As you note, the spec is vague and does not specify alignment for individual primitive types. But still, to me the statement "structs are aligned on the greatest of their element's alignment" seems to make perfect sense. The size of the primitive type will typically correspond directly to its alignment, but it seems they wanted to keep that open (or maybe avoid having to put architecture specific details into the specs). 

My point was, if we consider the alignment rules and thus the padding to be unpredictable, then we cannot predictably insert the padding field explicitly either. If the compiler would decide to align the [5]byte field to, say, 8 bytes (which is allowed according to the specs which require alignment _at least_ 1) it would insert padding before the padding field.
</blockquote></details>

You are right, we can't be sure of much. We can't even be sure 64bit data is 64bit aligned because some architectures don't care at all and I remember seeing some comments to that effect in some platform-dependent code of the Go runtime. That's why the assert is a good thing. What I'm doing is making very conservative alignments explicitly. As a result, the compiler can only add conterproductive padding. And I am betting it doesn't do that.

PR #4557, opened 2024-07-01
doc: update deploy tutorial to release v0.11.0

matzf:
Update link to download v0.11.0 packages.
Update topology.json files:
- Use simpler "local" address configuration without explicitly binding to a specific local IP.
- Use hostnames for remote hosts to avoid having to explicitly patch the placeholder IP addresses in the topology files.


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4557)

PR #4560, opened 2024-07-01
doc: proposal for nat address discovery

tzaeschke:
Proposal for NAT address discovery. Original discussion: #4517


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4560)

tzaeschke:
The build currently fails because the proposal is not in any TOC-tree. As I understand it, this is by design because, following `index.rst`, a proposal is only added to a TOC tree once the `change proposal converges and a design document is approved`.
  
Please let me know if the build problem should be fixed before proceeding.

jiceatscion:
I wonder if it could make sense to add a WIP section to the index and insert those proposals that are in active discussion. If only to avoid the rrt build failed noise. Thumbs up, thumbs down?

jiceatscion:
You can now link your doc to the WIP section to get rid of the CI grumbling.

jiceatscion (approved):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4560)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tzaeschke)

tzaeschke:
> (waiting on @tzaeschke)

@jiceatscion Can you tell me what I am meant to do?

jiceatscion:
I am a bit unclear about the process as well to be honest. Matthias had defined it very well on an abstract basis: Your design is in the final review state. The process says: "
Design:

    If the proposal has been accepted, the authors submit a design document and submit it to the repository ([doc/dev/design](https://github.com/scionproto/scion/blob/26103217c8074dba85330905ee7d4fae4127fb6f/doc/dev/design)) in the form of a pull request. See [Design Documents](https://docs.scion.org/en/latest/dev/design/index.html) for details.
"

However it doesn't describe the mechanics. Should we use the PR as a vehicle for discussions and refinement?  That would mean that the doc should never be submitted before the final approval. Not sure that would be the best. As long as the design is marked W-I-P submitting it would be OK, so more than one person can make changes to the design text.

Thoughts? Anyone?

PR #4563, opened 2024-07-03
snet: move SCMPHandler to cooked UDP Conn, don't propagate SCMP errors by default

matzf:
- snet.PacketConn is a "raw" socket. Having a special hook to handle SCMP packets, the SCMPHandler, seems unnecessary here, as any user of this raw socket can (and often wants to) just handle these packets explicitly.
  The SCMPHandler is thus moved from the "raw" PacketConn, to the "cooked" Conn and invoked only there. This makes use of the raw socket for things like sending and reacting to SCMPs much more straight forward.
- Options like the SCMPHandler and ReplyPather can be set in each Dial/Listen call, overriding the default set in the snet.SCIONNetwork.
- The `snet.DefaultSCMPHandler` no longer propagates any errors up. Path revocations are registered, but not directly reported back to the caller of Read/ReadFrom. That is, it's now equivalent to using the (removed) `snet.SCMPPropagationStopper`. See https://github.com/scionproto/scion/issues/4389
- Misc related snet API tweaks/cleanup; remove some unused structs, remove OpError, expose setting buffer sizes from snet.Conn (used by quic-go if available).

WIP: simply ignoring the SCMP errors seems undesirable in some places. The "end2end revocation test" relied on these errors and is no longer working. In the end2end test application itself, tighter timeouts for the ping/pong roundtrip might be sufficient to detect path failures quickly enough. However, I also noticed that path lookups themselves are not functioning well here and I suspect that in some place, the CS is also using SCMP errors for rapid detection of path failures. I need to investigate where this is, and what would be a good way to address this.

Perhaps this PR could be split into two parts; 1) moving the SCMPHandler from PacketConn to Conn, and 2) stopping the error propagation by default.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -94,13 +94,12 @@ func (dpf DataplaneSessionFactory) New(id uint8, policyID int,
 type PacketConnFactory struct {
 	Network *snet.SCIONNetwork
 	Addr    *net.UDPAddr
-	Options []snet.ConnOption
 }
 
 func (pcf PacketConnFactory) New() (net.PacketConn, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
 	defer cancel()
-	conn, err := pcf.Network.Listen(ctx, "udp", pcf.Addr, pcf.Options...)
+	conn, err := pcf.Network.Listen(ctx, "udp", pcf.Addr)
 	if err != nil {
 		return nil, serrors.WrapStr("creating packet conn", err)
 	}
@@ -408,14 +407,14 @@ func (g *Gateway) Run(ctx context.Context) error {
 	}
 
 	scionNetwork := &snet.SCIONNetwork{
-		Topology:          g.Daemon,
+		Topology: g.Daemon,
+		SCMPHandler: snet.DefaultSCMPHandler{
+			RevocationHandler: revocationHandler,
+			SCMPErrors:        g.Metrics.SCMPErrors,
+			Log:               log.FromCtx(ctx).Debug,
+		},
 		PacketConnMetrics: g.Metrics.SCIONPacketConnMetrics,
 		Metrics:           g.Metrics.SCIONNetworkMetrics,
-	}
-	scmpHandler := snet.DefaultSCMPHandler{
-		RevocationHandler: revocationHandler,
-		SCMPErrors:        g.Metrics.SCMPErrors,
-		Log:               log.FromCtx(ctx).Debug,
 	}
 
 	// Initialize the UDP/SCION QUIC conn for outgoing Gateway Discovery RPCs and outgoing Prefix
@@ -424,7 +423,6 @@ func (g *Gateway) Run(ctx context.Context) error {
 		context.TODO(),
 		"udp",
 		&net.UDPAddr{IP: g.ControlClientIP},
-		snet.WithSCMPHandler(scmpHandler),
 	)
 	if err != nil {
 		return serrors.WrapStr("unable to initialize client QUIC connection", err)
@@ -507,7 +505,6 @@ func (g *Gateway) Run(ctx context.Context) error {
 		context.TODO(),
 		"udp",
 		g.ControlServerAddr,
-		snet.WithSCMPHandler(scmpHandler),
 	)
 	if err != nil {
 		return serrors.WrapStr("unable to initialize server QUIC connection", err)
@@ -555,8 +552,7 @@ func (g *Gateway) Run(ctx context.Context) error {
 	// received from the session monitors of the remote gateway.
 	// *********************************************************************************
 
-	probeConn, err := scionNetwork.Listen(context.TODO(), "udp", g.ProbeServerAddr,
-		snet.WithSCMPHandler(scmpHandler))
+	probeConn, err := scionNetwork.Listen(context.TODO(), "udp", g.ProbeServerAddr)
 	if err != nil {
 		return serrors.WrapStr("creating server probe conn", err)
 	}
@@ -571,16 +567,8 @@ func (g *Gateway) Run(ctx context.Context) error {
 	}()
 
 	// Start dataplane ingress
-	dataplaneServerConn, err := scionNetwork.Listen(
-		context.TODO(),
-		"udp",
-		g.DataServerAddr,
-		snet.WithSCMPHandler(scmpHandler),
-	)
-	if err != nil {
-		return serrors.WrapStr("creating ingress conn", err)
-	}
-	if err := StartIngress(ctx, dataplaneServerConn, deviceManager, g.Metrics); err != nil {
+	if err := StartIngress(ctx, scionNetwork, g.DataServerAddr, deviceManager,
+		g.Metrics); err != nil {
 
 		return err
 	}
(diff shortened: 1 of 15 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4563)

jiceatscion (commented):
If the places where the default stopper behaviour hurts are known, I guess a non-default handler can used, right? So, is your concern about backward compatibility?

> Reviewed 14 of 15 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4563#-:-O0tGH0s2uug0faMX5p_:b-y8v8fx)* status: all files reviewed, 5 unresolved discussions (waiting on @matzf)

___
*[`gateway/pathhealth/pathwatcher.go` line 241 at r2](https://reviewable.io/reviews/scionproto/scion/4563#-O0t-bXr1UAgFIwWy3sD:-O0t-bXr1UAgFIwWy3sE:b-teycda) ([raw file](https://github.com/scionproto/scion/blob/eb310ce88a45d3a23d811c80f301564c0e3cbe14/gateway/pathhealth/pathwatcher.go#L241)):*
> ```Go
> 		}
> 		if err != nil {
> 			logger.Info("Unexpected error when reading probe reply", "err", err)
> ```

Do we intend to log for all cases of error, even the SCMP errors that we handle? The old code seemed to only log non-scmp errors. Something to do with what the built-in handler did?
___
*[`pkg/snet/conn.go` line 166 at r2](https://reviewable.io/reviews/scionproto/scion/4563#-O0t2viDDeFUolZRHf7u:-O0t2viDDeFUolZRHf7v:b2we079) ([raw file](https://github.com/scionproto/scion/blob/eb310ce88a45d3a23d811c80f301564c0e3cbe14/pkg/snet/conn.go#L166)):*
> ```Go
> 	}
> 	if o.replyPather == nil {
> 		o.replyPather = DefaultReplyPather{}
> ```

Ok, so this would seem to be the only opportunity to apply the default if WithReplyPather wasn't used, but... what is it that enforces the use of apply(). Using it seems optional (and frankly I don't understand why it exists). Just chaining calls to "WithXyz()" would be my natural way of building options (eventhough I don't really see the point of that pattern in Go). I must be mis-understanding the intended usage pattern.
___
*[`pkg/snet/reader.go` line 78 at r2](https://reviewable.io/reviews/scionproto/scion/4563#-O0t6dITBzojLb3O5Zdx:-O0t6dITBzojLb3O5Zdy:b9wgng1) ([raw file](https://github.com/scionproto/scion/blob/eb310ce88a45d3a23d811c80f301564c0e3cbe14/pkg/snet/reader.go#L78)):*
> ```Go
> 	}
> 
> 	udp := pkt.Payload.(UDPPayload)
> ```

Why do this earlier than before? The rest could still fail.
___
*[`scion/ping/ping.go` line 315 at r2](https://reviewable.io/reviews/scionproto/scion/4563#-O0tDmTR2ppypc-Kp5dr:-O0tDmTR2ppypc-Kp5ds:bmm2mlm) ([raw file](https://github.com/scionproto/scion/blob/eb310ce88a45d3a23d811c80f301564c0e3cbe14/scion/ping/ping.go#L315)):*
> ```Go
> 	switch r := pkt.Payload.(type) {
> 	case snet.SCMPEchoReply:
> 		if r.Identifier != expectedId {
> ```

is that really the way it works? We don't just drop it? It could be a late reply that we had given up on getting, could it not? Why must we care?
___
*[`scion/traceroute/traceroute.go` line 322 at r2](https://reviewable.io/reviews/scionproto/scion/4563#-O0tF-72BB5DweCmhqG5:-O0tF-73BbzX3DCCe6Vo:b-m0vt36) ([raw file](https://github.com/scionproto/scion/blob/eb310ce88a45d3a23d811c80f301564c0e3cbe14/scion/traceroute/traceroute.go#L322)):*
> ```Go
> 	switch r := pkt.Payload.(type) {
> 	case snet.SCMPTracerouteReply:
> 		if r.Identifier != expectedId {
> ```

same question as before about echo.

JordiSubira (commented):
> Reviewed 11 of 15 files at r1, 1 of 1 files at r2, 14 of 14 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4563)* status: all files reviewed, 8 unresolved discussions (waiting on @jiceatscion and @matzf)

___
*[`control/revhandler.go` line 29 at r3](https://reviewable.io/reviews/scionproto/scion/4563#-O1LQZ2r9zR1oC3wcHRn:-O1LQZ2r9zR1oC3wcHRo:bu26q33) ([raw file](https://github.com/scionproto/scion/blob/5dadff6ff6cb923825b70d904453aa5788d488f5/control/revhandler.go#L29)):*
> ```Go
> 
> // RevocationHandler handles raw revocations from the snet stack and inserts
> // them into the
> ```

them into the RevCache



_Code quote:_
```Go
them into the
```
___
*[`pkg/snet/scmp.go` line 54 at r3](https://reviewable.io/reviews/scionproto/scion/4563#-O1LUwpC9IKmbOtweufB:-O1LUwpC9IKmbOtweufC:bw3ce5g) ([raw file](https://github.com/scionproto/scion/blob/5dadff6ff6cb923825b70d904453aa5788d488f5/pkg/snet/scmp.go#L54)):*
> ```Go
> // revocation handler is configured, it is informed of any received interface
> // down messages.
> // It never returns an error "in line", so snet.Conn.Read/ReadFrom will not be
> ```

My doubts are twofold:
- Firstly, the current implementation of the `RevocationHandler` interface, do not propagate any error. So I guess we keep the error because some future implementation could propagate errors?
- Assuming that at some point the `RevocationHandler` will propagate internal errors when handling SCMP packets, I do not fully understand this statement. If  `self.RevocationHandler.Revoke` returns an error, I see this as an "SCMP error" or error related to SCMP, so I think we should clarify what this means.
___
*[`pkg/snet/scmp.go` line 69 at r3](https://reviewable.io/reviews/scionproto/scion/4563#-O1LT6KOEGH71I_Pg_qQ:-O1LT6KOEGH71I_Pg_qR:bnjgmqm) ([raw file](https://github.com/scionproto/scion/blob/5dadff6ff6cb923825b70d904453aa5788d488f5/pkg/snet/scmp.go#L69)):*
> ```Go
> 	scmp, ok := pkt.Payload.(SCMPPayload)
> 	if !ok {
> 		return nil
> ```

Related to my lack of understanding in the description. Is this a SCMP error that we shouldn't propagate? I only see that Handle is called for "cases" matching 'SCMPPayload', do we want to  just ignore if at some point we introduce some execution flow that provides other packets?
___
*[`scion/ping/ping.go` line 315 at r2](https://reviewable.io/reviews/scionproto/scion/4563#-O0tDmTR2ppypc-Kp5dr:-O1LrRIcAWIH8BRrOlbM:batli8t) ([raw file](https://github.com/scionproto/scion/blob/eb310ce88a45d3a23d811c80f301564c0e3cbe14/scion/ping/ping.go#L315)):*

is that really the way it works? We don't just drop it? It could be a late reply that we had given up on getting, could it not? Why must we care?
</blockquote></details>

I tnink the identifier now must be equal to the source port used for sending out the echo request, in the first place. So it should match the value. I guess considering this, the logic is to end up delivering this packet to the `ErrHandler` that then decides what to do.

PR #4575, opened 2024-07-16
monitoring: give a timeout to our prometheus request handler

### LINKED ISSUE #1971: Upgrade prometheus client_golang lib to fix timeout issues
The version of https://github.com/prometheus/client_golang/ that we're currently using has no timeout in its http handlers. This means if there are connectivity issues, it's possible for it to get stuck forever trying to read from a tcp connection that the other side has already torn down. This means a socket and goroutine get leaked forever. As of https://github.com/prometheus/client_golang/pull/379, there is now a way to specify a timeout for prometheus http handlers, plus it will automatically gain http keepalives because promhttp now uses `http.RoundTripper`.

So, we should upgrade to a recent client_golang version, and configure it to use a sane timeout (e.g. 5mins).

jiceatscion:
Fixes #1971

### DIFF
--- a/private/env/BUILD.bazel
+++ b/private/env/BUILD.bazel
@@ -19,6 +19,7 @@ go_library(
         "//pkg/scrypto:go_default_library",
         "//private/config:go_default_library",
         "@com_github_opentracing_opentracing_go//:go_default_library",
+        "@com_github_prometheus_client_golang//prometheus:go_default_library",
         "@com_github_prometheus_client_golang//prometheus/promhttp:go_default_library",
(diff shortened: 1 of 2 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4575)

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4575)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

Issue #4094, opened 2021-07-28
slayers: possible panic in the `SerializeTo` with receiver of type `Raw`


jcp19:
Hi :)
In the following excerpt, it seems to me that there is the implicit assumption that `len(s.Raw)` is at least `MetaLen`. Is this always true? if not, then it is possible that a panic occurs in the slicing operation (`s.Raw[:MetaLen]`).
https://github.com/scionproto/scion/blob/e4b728b87a7860d8901502efc77978fd97347d50/go/lib/slayers/path/scion/raw.go#L52-L56

If this is indeed a problem, then the simplest solution seems to be to remove the slicing operation because `s.PathMeta.SerializeTo` already performs a bounds check before writing to the buffer passed as argument.

oncilla:
Hi @jcp19 

There is an implicit assumption here. The only place that raw is set, here:

https://github.com/scionproto/scion/blob/e4b728b87a7860d8901502efc77978fd97347d50/go/lib/slayers/path/scion/raw.go#L39

This means that `s.Raw` is non-nil, it is guaranteed to be of at least MetaLen length.
However, to protect against future changes, I think we should just drop the slicing as you suggested.

Care to create a PR?

jcp19:
Yes, I will take care of this issue and the other occurrences of `s.Raw[:MetaLen]`.

jiceatscion:
Since we haven't heard from jcp19, attempting to refactor as suggested.

PR #4576, opened 2024-07-16
pkg/slayers: remove unnecessary sub-slicing.

jiceatscion:
Fixes: #4094

### DIFF
--- a/pkg/slayers/path/scion/decoded.go
+++ b/pkg/slayers/path/scion/decoded.go
@@ -64,7 +64,7 @@ func (s *Decoded) SerializeTo(b []byte) error {
 		return serrors.New("buffer too small to serialize path.", "expected", s.Len(),
 			"actual", len(b))
 	}
-	if err := s.PathMeta.SerializeTo(b[:MetaLen]); err != nil {
+	if err := s.PathMeta.SerializeTo(b); err != nil {
 		return err
 	}
 	offset := MetaLen
--- a/pkg/slayers/path/scion/raw.go
+++ b/pkg/slayers/path/scion/raw.go
@@ -50,8 +50,8 @@ func (s *Raw) SerializeTo(b []byte) error {
 		return serrors.New("buffer too small", "expected", minLen, "actual", len(b))
 	}
 	// XXX(roosd): This modifies the underlying buffer. Consider writing to data
-	// directly.
-	if err := s.PathMeta.SerializeTo(s.Raw[:MetaLen]); err != nil {
+	// directly. XXX(jiceatscion): Do we not want to update s.Raw anyway?
+	if err := s.PathMeta.SerializeTo(s.Raw); err != nil {
 		return err
 	}
 	copy(b, s.Raw)
@@ -82,7 +82,7 @@ func (s *Raw) Reverse() (path.Path, error) {
 // ToDecoded transforms a scion.Raw to a scion.Decoded.
 func (s *Raw) ToDecoded() (*Decoded, error) {
 	// Serialize PathMeta to ensure potential changes are reflected Raw.
-	if err := s.PathMeta.SerializeTo(s.Raw[:MetaLen]); err != nil {
+	if err := s.PathMeta.SerializeTo(s.Raw); err != nil {
 		return nil, err
 	}
 	decoded := &Decoded{}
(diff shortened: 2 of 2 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4576)

matzf (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4576)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion and @oncilla)

___
*[`pkg/slayers/path/scion/raw.go` line 53 at r1](https://reviewable.io/reviews/scionproto/scion/4576#-O1vRDkB4ShgZDc2qY36:-O1vRDkB4ShgZDc2qY37:b-31lqy0) ([raw file](https://github.com/scionproto/scion/blob/4d294f1f9168210708779638d901f5fc699561ac/pkg/slayers/path/scion/raw.go#L53)):*
>  XXX(jiceatscion): Do we not want to update s.Raw anyway?

Not necessarily; in the (somewhat contrived) scenario that we read and parse a packet, modify the PathMeta header and then use SerializeTo to a new buffer to create a copy of the packet, this would inadvertently modify the original packet.

IMO, the root cause for this to even be a question is the redundancy of the path meta information which is contained both in PathMeta and Raw.
A solution to this would be to cleanly separate PathMeta and Raw; during `DecodeFromBytes` we could strip off the first MetaLen bytes from Raw, then the information is only in one place. During SerializeTo the only option will be 

```
s.PathMeta.SerializeTo(b) // Serialize the PathMeta to b
copy(b[MetaLen:], s.Raw) // Serialize the rest of the path
```

Such a change would obviously affect a bunch of offset calculations in `raw.go`. Probably it would affect some test cases that create these objects by initializing the fields directly -- in a good way though, because it's currently possible to have very non-obviously conflicting information in the byte slices vs the PathMeta header.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4576)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @oncilla)

___
*[`pkg/slayers/path/scion/raw.go` line 53 at r1](https://reviewable.io/reviews/scionproto/scion/4576#-O1vRDkB4ShgZDc2qY36:-O1wLHKG9NrsUiLARL6e:b-idhlom) ([raw file](https://github.com/scionproto/scion/blob/4d294f1f9168210708779638d901f5fc699561ac/pkg/slayers/path/scion/raw.go#L53)):*

>  XXX(jiceatscion): Do we not want to update s.Raw anyway?

Not necessarily; in the (somewhat contrived) scenario that we read and parse a packet, modify the PathMeta header and then use SerializeTo to a new buffer to create a copy of the packet, this would inadvertently modify the original packet.

IMO, the root cause for this to even be a question is the redundancy of the path meta information which is contained both in PathMeta and Raw.
A solution to this would be to cleanly separate PathMeta and Raw; during `DecodeFromBytes` we could strip off the first MetaLen bytes from Raw, then the information is only in one place. During SerializeTo the only option will be 

```
s.PathMeta.SerializeTo(b) // Serialize the PathMeta to b
copy(b[MetaLen:], s.Raw) // Serialize the rest of the path
```

Such a change would obviously affect a bunch of offset calculations in `raw.go`. Probably it would affect some test cases that create these objects by initializing the fields directly -- in a good way though, because it's currently possible to have very non-obviously conflicting information in the byte slices vs the PathMeta header.
</blockquote></details>

Thanks! Since you approved, I take it that you agree with me that it can be treated as a separate issue. I will, add something to that effect to Dominik's comment instead of my question.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4576)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @matzf and @oncilla)

___
*[`pkg/slayers/path/scion/raw.go` line 53 at r1](https://reviewable.io/reviews/scionproto/scion/4576#-O1vRDkB4ShgZDc2qY36:-O2-6Erx7WWdC7aMTn1D:b-imrd6p) ([raw file](https://github.com/scionproto/scion/blob/4d294f1f9168210708779638d901f5fc699561ac/pkg/slayers/path/scion/raw.go#L53)):*

Thanks! Since you approved, I take it that you agree with me that it can be treated as a separate issue. I will, add something to that effect to Dominik's comment instead of my question.
</blockquote></details>

I think my edited comment addresses your point.

matzf (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4576#-:-O2-gwg4-i8z58AkZ8LH:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @oncilla)

PR #4578, opened 2024-07-16
router: implement a proper shutdown mechanism

### LINKED ISSUE #4116: Possibly missing synchronization in the `DataPlane.Run` method
Hi all,
in the `Run` method of the file [`dataplane.go`](https://github.com/scionproto/scion/blob/master/go/pkg/router/dataplane.go), it looks like every goroutine running the `read` closure consults the value of `d.running` (Line 455) to decide when to stop processing and forwarding messages. However, it seems to me that this is done without any kind of synchronization. According to the [Go memory model](https://golang.org/ref/mem) (section **Incorrect synchronization**), it can happen that a goroutine running `read` might not ever observe any changes to `d.running` made by other threads, and thus, it might not stop when it should.
https://github.com/scionproto/scion/blob/764d6e2afe4765acf24492746c197c32417a57c9/go/pkg/router/dataplane.go#L444-L456

Is this a bug or am I missing something here?

jiceatscion:
The "running" flag is now accessed as an atomic (so with sequential ordering per the Go memory model). Added a shutdown entry point which cannot race with initialization.
Fixes #4116

### DIFF
--- a/router/dataplane.go
+++ b/router/dataplane.go
@@ -27,6 +27,7 @@ import (
 	"net"
 	"net/netip"
 	"sync"
+	"sync/atomic"
 	"time"
 	"unsafe"
 
@@ -184,7 +185,7 @@ type DataPlane struct {
 	bfdSessions         map[uint16]bfdSession
 	localIA             addr.IA
 	mtx                 sync.Mutex
-	running             bool
+	running             atomic.Bool
 	Metrics             *Metrics
 	forwardingMetrics   map[uint16]interfaceMetrics
 	dispatchedPortStart uint16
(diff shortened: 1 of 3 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4578)

matzf (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4578#-:-O1wPw1v7r-GYLGhIr-O:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @oncilla)

jiceatscion:
I agree. I don't much see a real use case for sharing the management of a router between multiple threads but it is theoretically possible. Multiple threads could also fight over interfaces during initialization, which is also unlikely to produce sensible results. Note that after a call to Shutdown, the router is committed to terminating, so there's not much incentive in using Shutdown() as a way to allow reconfigurations. As things stand at the moment, the shutdown mechanism only exists so unit tests can terminate the router. I almost did what you described (i.e. more than two states) and then stopped myself...YAGNI.

But if that loose end bothers you, you are mot welcome to file a new issue (or re-open yours) and propose such an improvement. It shouldn't add much complexity, so I won't object on those grounds.

PR #4579, opened 2024-07-17
router: drop packets with v4 mapped v6 src/dst or unspecified dst hosts

### LINKED ISSUE #4415: router: filter IPv4-mapped IPv6 addresses
An IPv6 address can represent an IPv4 address, e.g. the IPv6 address `::ffff:172.20.2.2` represents the IPv4 address `172.20.2.2`. This is called IPv4-mapped IPv6 address and is intended as a compatibility mechanism in dual stack hosts. These addresses are not intended to be used on the wire.
The SCION address header has explicit representation for IPv4 or IPv6 addresses with an explicit type discriminator. An IPv4 address can be encoded as either IPv4 type, or as IPv6 type containing an IPv4-mapped IPv6 address.
The latter choice is wasteful (16 bytes instead of 4) and not practically useful. The dual representation could potentially be abused to bypass naive or faulty packet filter implementations.

As there does not appear to be an upside in allowing IPv4-mapped IPv6 addresses in the SCION address header. To reduce the risk of abusive or accidental (as in #4377) use of these mapped addresses, the router should filter these packets. The following two processing rules should be added to the router:

- packets destined for the local AS with an IPv4-mapped IPv6 destination address should be dropped (optional SCMP error message)
- packets originating from the local AS with an IPv4-mapped IPV6 source address should be dropped (optional SCMP error message)

### LINKED ISSUE #4558: router: drop packets with invalid IPv4/6 destination address
The router forwards packets relying on whatever the system call, i.e., `unix.Sendmsg` does with UDP messages with destination address `0.0.0.0`. One could expect that the syscall returns an error, however, what it does is the following (based on a toy example I creted for trying this out):

- If internal BR address (or local_IP) uses IPv4 and destination address is 0.0.0.0: Sends a packet with destination_IP = local_IP
- If internal BR address (or local_IP) uses IPv6 and destination address is 0.0.0.0: Yields an error Error sending message: network is unreachable
-  If internal BR address (or local_IP) uses IPv6 and destination address is [::]: Sends a packet with destination_IP = local_IP
-  If internal BR address (or local_IP) uses IPv4 and destination address is [::]: Yields an error Error sending message: network is unreachable

(One can reproduce it using this simple code snippet https://gist.github.com/JordiSubira/4171baa5b0535878bebf4c93eb1688e5).

Based on an offline discussion, Anapaya's BR drops packets with 0.0.0.0, [::], loopback/localhost or ipv4-mapped-ipv6 destination addresses. Thus, the common conclusion is that we should reproduce this in the OS-BR.

jiceatscion:
The check is done on the first hop for src and on the last hop for dst.

Fixes #4415
Fixes #4558

### DIFF
--- a/router/dataplane.go
+++ b/router/dataplane.go
@@ -217,6 +217,7 @@ var (
 	unsupportedPathType           = errors.New("unsupported path type")
 	unsupportedPathTypeNextHeader = errors.New("unsupported combination")
 	unsupportedV4MappedV6Address  = errors.New("unsupported v4mapped IP v6 address")
+	unsupportedUnspecifiedAddress = errors.New("unsupported unspecified address")
 	noBFDSessionFound             = errors.New("no BFD session was found")
 	noBFDSessionConfigured        = errors.New("no BFD sessions have been configured")
 	errPeeringEmptySeg0           = errors.New("zero-length segment[0] in peering path")
@@ -1656,7 +1657,7 @@ func (p *scionPacketProcessor) resolveInbound() disposition {
 			code:     slayers.SCMPCodeNoRoute,
 		}
 		return pSlowPath
-	case invalidDstAddr, unsupportedV4MappedV6Address:
+	case invalidDstAddr, unsupportedV4MappedV6Address, unsupportedUnspecifiedAddress:
 		log.Debug("SCMP response", "cause", err)
 		p.pkt.slowPathRequest = slowPathRequest{
 			scmpType: slayers.SCMPTypeParameterProblem,
@@ -1855,8 +1856,6 @@ func (p *scionPacketProcessor) validateSrcHost() disposition {
 	if p.scionLayer.SrcIA != p.d.localIA {
 		return pForward
 	}
-	// TODO(jiceatscion): That is the second time we parse the src address
-	// we might want to cache it.
 	src, err := p.scionLayer.SrcAddr()
 	if err == nil && src.IP().Is4In6() {
 		err = unsupportedV4MappedV6Address
@@ -2103,6 +2102,9 @@ func (d *DataPlane) resolveLocalDst(
 		dstIP := dst.IP()
 		if dstIP.Is4In6() {
 			return unsupportedV4MappedV6Address
+		}
+		if dstIP.IsUnspecified() { // IsInvalid() not possible, we initialized it from wire bits.
+			return unsupportedUnspecifiedAddress
 		}
 		return d.addEndhostPort(resolvedDst, lastLayer, dstIP)
 	default:
--- a/router/dataplane.go
+++ b/router/dataplane.go
@@ -2104,7 +2104,7 @@ func (d *DataPlane) resolveLocalDst(
 		if dstIP.Is4In6() {
 			return unsupportedV4MappedV6Address
 		}
-		return d.addEndhostPort(resolvedDst, lastLayer, dst.IP())
+		return d.addEndhostPort(resolvedDst, lastLayer, dstIP)
 	default:
 		panic("unexpected address type returned from DstAddr")
 	}
@@ -2446,18 +2446,13 @@ func (p *slowPathPacketProcessor) prepareSCMP(
 	scionL.Path = revPath
 	scionL.DstIA = p.scionLayer.SrcIA
 	scionL.SrcIA = p.d.localIA
-	srcA, err := p.scionLayer.SrcAddr()
-	if err != nil {
-		return nil, serrors.Wrap(cannotRoute, err, "details", "extracting src addr")
-	}
-	if err := scionL.SetDstAddr(srcA); err != nil {
-		return nil, serrors.Wrap(cannotRoute, err, "details", "setting dest addr")
-	}
+	scionL.DstAddrType = p.scionLayer.SrcAddrType
+	scionL.RawDstAddr = p.scionLayer.RawSrcAddr
+	scionL.NextHdr = slayers.L4SCMP
+
 	if err := scionL.SetSrcAddr(addr.HostIP(p.d.internalIP)); err != nil {
 		return nil, serrors.Wrap(cannotRoute, err, "details", "setting src addr")
 	}
-	scionL.NextHdr = slayers.L4SCMP
-
 	typeCode := slayers.CreateSCMPTypeCode(typ, code)
 	scmpH := slayers.SCMP{TypeCode: typeCode}
 	scmpH.SetNetworkLayerForChecksum(&scionL)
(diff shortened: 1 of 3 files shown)


matzf (approved):
LGTM, but maybe you can directly add filtering for zero addresses  ("unspecified" addresses) too, to fix https://github.com/scionproto/scion/issues/4558?

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4579#-:-O23mMNA6Qbs6OY85gqc:btru7oj)* status: all files reviewed, 3 unresolved discussions (waiting on @jiceatscion and @JordiSubira)

___
*[`router/dataplane.go` line 1858 at r1](https://reviewable.io/reviews/scionproto/scion/4579#-O23nGOS05EiPndLXNyT:-O23nGOS05EiPndLXNyU:b-crhrfj) ([raw file](https://github.com/scionproto/scion/blob/d63e202e61333a93238d6ff6e97f0847850e8089/router/dataplane.go#L1858)):*
> ```Go
> 		return pForward
> 	}
> 	// TODO(jiceatscion): That is the second time we parse the src address
> ```

Nit: Is it? I only see this in the "slow path" related to creating SCMPs.

Btw, in `prepareSCMP` parsing the source address only to set the destination address is unnecessary and constrains the supported address types to what this implementation understands (could even be considered a bug). It could just set `DstAddrType, RawDstAddr = SrcAddrType, RawSrcAddr` without parsing instead.
___
*[`router/dataplane.go` line 2107 at r1](https://reviewable.io/reviews/scionproto/scion/4579#-O23oxwq9tyH0V62jEZW:-O23oxwq9tyH0V62jEZX:b9fn8u0) ([raw file](https://github.com/scionproto/scion/blob/d63e202e61333a93238d6ff6e97f0847850e8089/router/dataplane.go#L2107)):*
> ```Go
> 			return unsupportedV4MappedV6Address
> 		}
> 		return d.addEndhostPort(resolvedDst, lastLayer, dst.IP())
> ```

Nit: use`dstIP`?
___
*[`router/dataplane_internal_test.go` line 728 at r1](https://reviewable.io/reviews/scionproto/scion/4579#-O23s4H09Cnw9IZl14_6:-O23s4H09Cnw9IZl14_7:bou74bc) ([raw file](https://github.com/scionproto/scion/blob/d63e202e61333a93238d6ff6e97f0847850e8089/router/dataplane_internal_test.go#L728)):*
> ```Go
> 		},
> 	}
> 	dpath.HopFields[2].Mac = computeMAC(t, testKey, dpath.InfoFields[0], dpath.HopFields[2])
> ```

Nit: this is not the current hop field (current is 0). This would fail in MAC validation (which comes after validating the src address, so the tests are passing).



_Suggestion:_
```Go
	dpath.HopFields[0].Mac = computeMAC(t, testKey, dpath.InfoFields[0], dpath.HopFields[0])
```

jiceatscion (commented):
Good idea. 1 stone, 2 bugs.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4579#-:-O247zvEEVvihjyMM2hX:b4syra6)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @JordiSubira and @matzf)

___
*[`router/dataplane.go` line 1858 at r1](https://reviewable.io/reviews/scionproto/scion/4579#-O23nGOS05EiPndLXNyT:-O23xKVG6ghML-g56m7J:b-896fjb) ([raw file](https://github.com/scionproto/scion/blob/d63e202e61333a93238d6ff6e97f0847850e8089/router/dataplane.go#L1858)):*

Nit: Is it? I only see this in the "slow path" related to creating SCMPs.

Btw, in `prepareSCMP` parsing the source address only to set the destination address is unnecessary and constrains the supported address types to what this implementation understands (could even be considered a bug). It could just set `DstAddrType, RawDstAddr = SrcAddrType, RawSrcAddr` without parsing instead.
</blockquote></details>

Done 
___
*[`router/dataplane.go` line 2107 at r1](https://reviewable.io/reviews/scionproto/scion/4579#-O23oxwq9tyH0V62jEZW:-O23xfgPDHKVHJT1kLrx:bb7f2mf) ([raw file](https://github.com/scionproto/scion/blob/d63e202e61333a93238d6ff6e97f0847850e8089/router/dataplane.go#L2107)):*

Nit: use`dstIP`?
</blockquote></details>

done
___
*[`router/dataplane_internal_test.go` line 728 at r1](https://reviewable.io/reviews/scionproto/scion/4579#-O23s4H09Cnw9IZl14_6:-O23y3bD-t1gDbD2-LNL:b8vmmaa) ([raw file](https://github.com/scionproto/scion/blob/d63e202e61333a93238d6ff6e97f0847850e8089/router/dataplane_internal_test.go#L728)):*

Nit: this is not the current hop field (current is 0). This would fail in MAC validation (which comes after validating the src address, so the tests are passing).
</blockquote></details>

done oops.

matzf (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4579#-:-O24KU-P4iBEmU-zz9oG:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JordiSubira)

JordiSubira (commented):
Just one minor think, in #4558 we also mention loopback/localhost address which are not taken care of in this PR. So, either we create a new issue only with this case, or I wouldn't state that this PR completely fixes #4558 .

Besides, should we also filter unspecified src addresses?

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4579#-:-O24N7nhAihB43fp9NKH:b-n2n56e)* status: all files reviewed, 1 unresolved discussion

___
*[`router/dataplane.go` line 2135 at r2](https://reviewable.io/reviews/scionproto/scion/4579#-O24OcxV8SYA4DZ-gIy7:-O24OcxV8SYA4DZ-gIy8:bmt1xy7) ([raw file](https://github.com/scionproto/scion/blob/f229da17d258831bec53bfca1fc0e3912f5ad70c/router/dataplane.go#L2135)):*
> ```Go
> 			return unsupportedV4MappedV6Address
> 		}
> 		if dstIP.IsUnspecified() { // IsInvalid() not possible, we initialized it from wire bits.
> ```

Is this a function of netip.Addr. I do not completely understand the comment otherwise.



_Code quote:_
```Go
IsInvalid()
```

jiceatscion:
Ah yes, netip.Addr has IsInvalid() and IsUnspecified() and they have different meanings. IsInvalid() means "uninitialized" i.e. the Go zero value, which is neither "0.0.0.0" nor "::". I was trying to convey that concisely in the comment but I'm afraid I overdid the concisely part.

PR #4586, opened 2024-07-26
serrors: delete error context merging and code cleanup

### LINKED ISSUE #4486: serrors isn't treating by value or by reference errors consistently
serrors.New returns a pointer. all the wrapper constructors return values.
Is() expects a value. It will fail to recognize a pointer as a basicError. However, if given a value, it will panic for any error that happens to be a wrapper around basicErrors, as it will try to compare these objects which are not comparable.

WithCtx() expects only a value and will fail to downcast a basicError ptr, resulting in always wrapping instead of merging ctx.
A number of test cases have been crafted to expect that un-planned behaviour. For example private/mgmtapi/cppki/api/api_test.go expect a wrapped error while serror() was designed to produce the same error in merged form. (i.e. fixing serrors) will break those tests.

Either we fix serrors and the tests or, may be simpler, we remove that accidentally unused merging mechanism from serrors.

In either case we should decide if we want serrrors to handle values, or references, or both and implement whatever we chose consistently. I suspect we really want references since that's what we always use (otherwise Is() would be panicking).

jiceatscion:
There were a number of intertwined issues difficult to characterize in isolation.
This proposes something that has a more consistent behavior and interface, with shims to ease transition.

Most issues stemmed from the withCtx() function which (in rare cases) created hybrid objects that required complicated support code and, in some cases could cause Is() to panic.

Removed the feature, replaced WithCtx() with a shim that doesn't create hybrids. Removed all the complicated code that resulted. The code is now more obvious but not shorter. There are now two subclasses of errors:
* basicError: dedicated to errors of the form string+cause - Constructed by Wrap()
* joinedError: dedicated to errors of the form error+cause - Constructed by Join()
This allows for Is() and Unwrap() to be different and do the right thing in all cases.

Introduced new constructors with obvious semantics to replace those already marked deprecated.

I still plan on editing the calls to deprecated constructors (and common.ErrMsg) out of everywhere, but in a subsequent PR.

Fixes #4486


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)

oncilla (changes requested):
> Reviewed 2 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 2 of 13 files reviewed, 4 unresolved discussions (waiting on @jiceatscion and @lukedirtwalker)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4586#-O330JnG9TNu80lAW0_O:-O330JnG9TNu80lAW0_P:b-y2xftn) (no related file):*
note: only high level comments so far. I need to dig into the logic some more and give a more thorough look.

___
___
*[`pkg/private/serrors/errors.go` line 75 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O32r1LR0l-wf-JY-zVq:-O32r1LR0l-wf-JY-zVr:b-wjk2h4) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L75)):*
> ```Go
> // representation.
> func (e basicError) MarshalLogObject(enc zapcore.ObjectEncoder) error {
> 	enc.AddString("msg", e.msg.Error())
> ```

I think here you should still check whether the wrapped error implements the zapcore.ObjectMarshaler interface, and use it if so.

Otherwise, this does not play nicely with JSON encoded logging.
___
*[`pkg/private/serrors/errors.go` line 245 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O33-z8f5mDvFG7ofndO:-O33-z8f5mDvFG7ofndP:bb8u4bk) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L245)):*
> ```Go
> }
> 
> // Wrap() is deprecated. Use FromErr() instead. This shim has almost the same result.
> ```

Use the syntax described in https://go.dev/wiki/Deprecated
This will mark it as deprecated in documentation, and will add squiggly lines in editors.
___
*[`pkg/private/serrors/errors.go` line 252 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O33064b7Aon1iRNOBxe:-O33064b7Aon1iRNOBxf:b-89a4l7) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L252)):*
> ```Go
> }
> 
> // WrapStr() is deprecated. Use FromStrWithStack() instead.
> ```

Ditto

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 2 of 13 files reviewed, 4 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/private/serrors/errors.go` line 75 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O32r1LR0l-wf-JY-zVq:-O33Bv4--GtGuRqMd9Sf:bqjvzxv) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L75)):*

I think here you should still check whether the wrapped error implements the zapcore.ObjectMarshaler interface, and use it if so.

Otherwise, this does not play nicely with JSON encoded logging.
</blockquote></details>

Not calling it was deliberate in that I wanted to treat e.msg as a generic error. Specifically, if it happens to be a basicError, and if it happens to have a stack and a cause, I think those ought to be ignored; the ones that are meaningful are e.cause and e.stack. So that leaves us with e.msg.Error() as the only representation we can get. Ok, we can still get a context in the string. Is that what ruins the json representation? If so, I should probably encode it since we can't know what e.msg.Error() looks like (in cases it doesn't implement objectMarshaler, should we check for it).
___
*[`pkg/private/serrors/errors.go` line 245 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O33-z8f5mDvFG7ofndO:-O33Bt-f7blZFmdGMuUy:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L245)):*

Use the syntax described in https://go.dev/wiki/Deprecated
This will mark it as deprecated in documentation, and will add squiggly lines in editors.
</blockquote></details>

Done.
___
*[`pkg/private/serrors/errors.go` line 252 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O33064b7Aon1iRNOBxe:-O33BtYq4SjavQr0v7iJ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L252)):*

Ditto
</blockquote></details>

Done.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 2 of 13 files reviewed, 2 unresolved discussions (waiting on @jiceatscion and @lukedirtwalker)

___
*[`pkg/private/serrors/errors.go` line 75 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O32r1LR0l-wf-JY-zVq:-O36ri930rVlVFloQXxn:bi9bswd) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L75)):*

Not calling it was deliberate in that I wanted to treat e.msg as a generic error. Specifically, if it happens to be a basicError, and if it happens to have a stack and a cause, I think those ought to be ignored; the ones that are meaningful are e.cause and e.stack. So that leaves us with e.msg.Error() as the only representation we can get. Ok, we can still get a context in the string. Is that what ruins the json representation? If so, I should probably encode it since we can't know what e.msg.Error() looks like (in cases it doesn't implement objectMarshaler, should we check for it).
</blockquote></details>

argh. I think I read the code wrong. I thought his was about cause rather than msg. Ignore me :bow: : 
___
*[`pkg/private/serrors/errors.go` line 245 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O33-z8f5mDvFG7ofndO:-O36vUGlBTGL5fXcBOMq:bal8e84) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L245)):*

Done.
</blockquote></details>

Don't see it. Did you push your changes?
___
*[`pkg/private/serrors/errors.go` line 252 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O33064b7Aon1iRNOBxe:-O36vX5x3aC1OJH5QgT8:b-89a4l7) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L252)):*

Done.
</blockquote></details>

Ditto

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 2 of 13 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/private/serrors/errors.go` line 245 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O33-z8f5mDvFG7ofndO:-O3WXLNy46G19GBxq4mq:b-hyzdkt) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L245)):*

Don't see it. Did you push your changes?
</blockquote></details>

Indeed I hadn't. Done now.
___
*[`pkg/private/serrors/errors.go` line 252 at r3](https://reviewable.io/reviews/scionproto/scion/4586#-O33064b7Aon1iRNOBxe:-O3WXhjsEv81rR7Lp5Ql:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/60bc8694337c5af0a0421ef3ef7d7284a95f41df/pkg/private/serrors/errors.go#L252)):*

Ditto
</blockquote></details>

Done.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 2 of 13 files reviewed, 3 unresolved discussions (waiting on @jiceatscion and @lukedirtwalker)

___
*[`pkg/private/serrors/errors.go` line 173 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3g_9M5D4Y74y8Orkhc:-O3g_9M5D4Y74y8Orkhd:bq2ten3) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L173)):*
> ```Go
> 	}
> 	if cause != nil {
> 		// In the odd case where err already has a causes, the new one takes precedence.
> ```

nit:  I think that branch is unnecessary. You can always assign cause in the variable initialization.

My only guess for having this if condition is that you want to protect against typed nil, but that does not work:
https://go.dev/play/p/SYObtbU3eyk
___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3gb34B2_Yq5Z6VtpYC:b-2xxdrs) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*
> ```Go
> // The returned error implements Is. Is(err) returns true. Is(cause) returns true if cause is not
> // nil.
> func FromErr(err, cause error, errCtx ...interface{}) error {
> ```

This PR introduced 4 error constructors:

- func FromErr
- func FromMsgWithStack
- func FromStr
- func FromStrWithStack

And deprecates the following:

-  Wrap
-  WrapStr
-  WithCtx

I have two main concerns:

1. IMO it becomes even harder to understand what should be the go to mechanism for wrapping errors. In 99% of the cases, we want to go with WrapStr/FromStrWithStack. 

2. The common case becomes unreadable -> our code will be messy to read as we have serrors.FromStrWithStack ~everywhere now.

I think we should take a step back. What is it that we really want to achieve?

1. We want a way to easily wrap errors and attach additional context. The error wrapping and additional context should play nicely with JSON encoding.

2. We want a way to use sentinel errors in response to other errors such that client code can change their behavior. (Here, in most cases we do not even care about additional context, most call sites do not add context)

I think we get ourselves in a bit of a mess because we try to stick both of these goals in the same type.

What if we do the following:

```go
type basicError struct {
    msg string
    cause error
    stack *stack
}

func New(msg string, errCtx ...any) error {}

// Replaces WrapStr
func Wrap(msg string, cause error, errCtx ...any) error {}

type sentinelError struct {
    sentinel error
    cause error
    stack *stack
    // potentially add ctx?
}

// Replaces Wrap
func Sentinel(sentinel error, cause error) error {}
```

This simplifies the package API to 3 functions:
- New
- Wrap
- Sentinel

it should also cover all the use cases that we have.
WithCtx can be covered by combining Sentinel and Wrap together AFAICT.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3mSG0fEGHbkPQWnjKg:b-i37wuz) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*

I don't agreee with this proposal: Sentinel() is confusing. It gives the impression the we're building a sentinel error while we're building FROM a sentinel error. Likewise for the name of the struct, of course. Besides, we really do not need two different structures to construct errors wrapped around sentinel errors. We can obtain the same result by requiring sentinel errors to be of some explicit type, or just not care about the underlying type of the msg field besides it implementing error. What's wrong about that? Costs next to nothing and gives us freedom to refactor the using code incrementally. So, I;m afraid I don't really understand why you want a Sentinel error type. How would its behavior compare with that of basicError?
</blockquote></details>

Done as per our oob discussion. There are more than 3 new constructors because I insist on enabling the creation of stack-dump-free errors, but no-one needs to be aware of them. The short names are for the stack-full ones. It isn't possible to be rid of Is(). Our error objects aren't comparable. We'd need to replace the context map by a pointer-to-same, and even then, I'm not 100% sure that the default comparison would do what we want (but I can test it in case we want that badly).

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @jiceatscion and @lukedirtwalker)

___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3mYKD239_eovUdFetH:bgzluv) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*
> Our error objects aren't comparable.

I fail to see why they would need to be comparable.
As we have discussed OOB, our errors are solely there to wrap or join.
They should not be used for constructing sentinel errors.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @jiceatscion and @lukedirtwalker)

___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3mZ9vq8ik-Lq1bH7tN:b-kmv80a) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*

> Our error objects aren't comparable.

I fail to see why they would need to be comparable.
As we have discussed OOB, our errors are solely there to wrap or join.
They should not be used for constructing sentinel errors.
</blockquote></details>

In other words:

```go

s1 := errors.New("s1")
s2 := errors.New("s2")

wrapped := serrors.Wrap("wrapped", s1)
joined := serrors.Join(s1, s2)

errors.Is(wrapped, s1) // true
errors.Is(wrapped, s2) // false
errors.Is(wrapped, wrapped) // can be false, because we do not care :shrug:

erors.Is(joined, s1) // true
erors.Is(joined, s2) // true
erors.Is(joined, joined) // can be false, because we do not care :shrug:
```

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3maF3j8-ZSN7ifF0mG:b83u4c7) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*

In other words:

```go

s1 := errors.New("s1")
s2 := errors.New("s2")

wrapped := serrors.Wrap("wrapped", s1)
joined := serrors.Join(s1, s2)

errors.Is(wrapped, s1) // true
errors.Is(wrapped, s2) // false
errors.Is(wrapped, wrapped) // can be false, because we do not care :shrug:

erors.Is(joined, s1) // true
erors.Is(joined, s2) // true
erors.Is(joined, joined) // can be false, because we do not care :shrug:
```
</blockquote></details>

You might think that our entire code base should be a certain way but that doesn't mean it is. For now we rely on Is() behaving the way it does and I did not sign up to refactor everything so it no longer matters. Not in this PR. Also, I would be careful when stating that something shouldn't matter based on a current usage pattern. It is too easy to create opportunities for inconsistencies and really absurd behaviors that are only revealed when someone later uses the API in a new, yet legitimate way. Aaanyway... I did make one quick check. If we force our errors to be comparable by making the map a pointer, then, error behaves as we want even without the Is() methods. If it matters very much to you that we do not implement Is(), then it can be had. That saves ~30 LoCs.

oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @jiceatscion and @lukedirtwalker)

___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3mc_cr7DPmPmfTRHFo:b-gerh8l) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*

You might think that our entire code base should be a certain way but that doesn't mean it is. For now we rely on Is() behaving the way it does and I did not sign up to refactor everything so it no longer matters. Not in this PR. Also, I would be careful when stating that something shouldn't matter based on a current usage pattern. It is too easy to create opportunities for inconsistencies and really absurd behaviors that are only revealed when someone later uses the API in a new, yet legitimate way. Aaanyway... I did make one quick check. If we force our errors to be comparable by making the map a pointer, then, error behaves as we want even without the Is() methods. If it matters very much to you that we do not implement Is(), then it can be had. That saves ~30 LoCs.
</blockquote></details>

I do not care about the lines of code.

I care about the code being correct, and it is not obvious to me that the following snippet does the right thing.

```go
func (e joinedError) Is(err error) bool {
	switch other := err.(type) {
	case joinedError:
		// Two joinedError values. They're not comparable, that's why errors.Is() called this.
		// (That's unlikely given how joinedError is used, but entirely feasible).
		if e.error != nil && other.error != nil && !(reflect.TypeOf(e.error).Comparable() &&
			reflect.TypeOf(other.error).Comparable()) {
			return false
		}
		return e.error == other.error
	}
	// Otherwise only possible equality is if e.cause.Is(err) or e.error.Is(err). Those'll be tested
	// after Unwrap().
	return false
}
```

I would rather not bother and just let errors.Is deal with it.

We can use a map pointer, or we could return pointer instead of value types in our New,Wrap,Join.

FWIW, we can completely drop the map and just have a pointer to a slice instead. Saves a bit of garbage too.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3mdhey-XuqGkROS1gD:b-nm0e5l) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*

I do not care about the lines of code.

I care about the code being correct, and it is not obvious to me that the following snippet does the right thing.

```go
func (e joinedError) Is(err error) bool {
	switch other := err.(type) {
	case joinedError:
		// Two joinedError values. They're not comparable, that's why errors.Is() called this.
		// (That's unlikely given how joinedError is used, but entirely feasible).
		if e.error != nil && other.error != nil && !(reflect.TypeOf(e.error).Comparable() &&
			reflect.TypeOf(other.error).Comparable()) {
			return false
		}
		return e.error == other.error
	}
	// Otherwise only possible equality is if e.cause.Is(err) or e.error.Is(err). Those'll be tested
	// after Unwrap().
	return false
}
```

I would rather not bother and just let errors.Is deal with it.

We can use a map pointer, or we could return pointer instead of value types in our New,Wrap,Join.

FWIW, we can completely drop the map and just have a pointer to a slice instead. Saves a bit of garbage too.
</blockquote></details>

I just said I can give you a solution where error.Is() is enough to satisfy all the requirements that we test for, it does remove this code that you have doubts about AND relies only on trusted errors.Is(). Is that still not acceptable? Regarding the means, a slice will not be better than a map. They're both non-comparable.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3ml9evALslvDc7EDhJ:b-inlx7i) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*

I just said I can give you a solution where error.Is() is enough to satisfy all the requirements that we test for, it does remove this code that you have doubts about AND relies only on trusted errors.Is(). Is that still not acceptable? Regarding the means, a slice will not be better than a map. They're both non-comparable.
</blockquote></details>

But yeah, ptr or not, a slice would be enough. That seems to be completely orthogonal to this PR's focus, though, but if you'll review it, I'll do it.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3qFnye-CnGNA5stZMZ:beqzbgo) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*

But yeah, ptr or not, a slice would be enough. That seems to be completely orthogonal to this PR's focus, though, but if you'll review it, I'll do it.
</blockquote></details>

And here it is. I think it is cleaner and simpler than the map. ctxPairs is what we'd turn it into for encodeContext anyway. Might as well skip the map. Is that a keeper?

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

*[a discussion](https://reviewable.io/reviews/scionproto/scion/4586#-O330JnG9TNu80lAW0_O:-O3qI-BE7QPma7WqAQAd:b2nhysk) (no related file):*

note: only high level comments so far. I need to dig into the logic some more and give a more thorough look.
</blockquote></details>

ack

___

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3qI5ho6nwxh6Zr6viY:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*

And here it is. I think it is cleaner and simpler than the map. ctxPairs is what we'd turn it into for encodeContext anyway. Might as well skip the map. Is that a keeper?
</blockquote></details>

done

oncilla (changes requested):
> Reviewed 2 of 5 files at r1, 6 of 9 files at r2, 1 of 3 files at r3, 4 of 5 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: all files reviewed, 3 unresolved discussions (waiting on @jiceatscion and @lukedirtwalker)

___
*[`pkg/private/serrors/errors.go` line 198 at r4](https://reviewable.io/reviews/scionproto/scion/4586#-O3gb34B2_Yq5Z6VtpYB:-O3qOdHW7nUP6lrE0GkM:b8q77cv) ([raw file](https://github.com/scionproto/scion/blob/13ea17f8210f497b863129cb76a9c8eaa50bbb6b/pkg/private/serrors/errors.go#L198)):*
> I just said I can give you a solution where error.Is() is enough to satisfy all the requirements that we test for, it does remove this code that you have doubts about AND relies only on trusted errors.Is(). Is that still not acceptable?

It is acceptable, I did not mean to imply it wasn't. It is actually my preferred way. Sorry, it looks like I did not phrase it clearly enough.


> But yeah, ptr or not, a slice would be enough. That seems to be completely orthogonal to this PR's focus, though, but if you'll review it, I'll do it.

Orthogonal,  yes. Hence the FWIW. But again, it is the preferable option, especially since we are touching all of this code already anyway.

> Is that a keeper?

Indeed.
___
*[`pkg/private/serrors/errors.go` line 79 at r8](https://reviewable.io/reviews/scionproto/scion/4586#-O3qPKY6807nnVYiDbI8:-O3qPKY6807nnVYiDbI9:bmy5nk0) ([raw file](https://github.com/scionproto/scion/blob/239b7684a0996697222bd41d977733f3da0d0f62/pkg/private/serrors/errors.go#L79)):*
> ```Go
> 	}
> 	for _, pair := range *e.ctx {
> 		if err := enc.AddReflected(pair.Key, pair.Value); err != nil {
> ```

Reminder: include the changes from #4594 here too after rebase.
___
*[`pkg/private/serrors/errors.go` line 134 at r8](https://reviewable.io/reviews/scionproto/scion/4586#-O3qOPx1A_gt_h-SGceq:-O3qOPx1A_gt_h-SGcer:b-7bnyuw) ([raw file](https://github.com/scionproto/scion/blob/239b7684a0996697222bd41d977733f3da0d0f62/pkg/private/serrors/errors.go#L134)):*
> ```Go
> 	// basicError was without a stack trace, then there'll never be one. That's to avoid looking
> 	// for it in every level or every constructor. TB revisisted if necessary.
> 	if r.cause == nil || !errors.As(cause, &existingVal) && !errors.As(cause, &existingPtr) {
> ```

I _think_ you need to check both basic error or joined error here.
Otherwise we have two stack traces if we wrap a joined error.
___
*[`pkg/private/serrors/errors.go` line 257 at r8](https://reviewable.io/reviews/scionproto/scion/4586#-O3qPuIU6jGrAOUPd6YY:-O3qPuIU6jGrAOUPd6YZ:b-k1qov3) ([raw file](https://github.com/scionproto/scion/blob/239b7684a0996697222bd41d977733f3da0d0f62/pkg/private/serrors/errors.go#L257)):*
> ```Go
> }
> 
> // Deprecated: WithCtx should never have existed. Depending on intent, use:
> ```

nit: (Feel free to ignore)

To make this look nice in pkg.go.dev, you can use go doc supported formatting: https://tip.golang.org/doc/comment#lists

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: 13 of 14 files reviewed, 2 unresolved discussions (waiting on @lukedirtwalker and @oncilla)

___
*[`pkg/private/serrors/errors.go` line 79 at r8](https://reviewable.io/reviews/scionproto/scion/4586#-O3qPKY6807nnVYiDbI8:-O3rzqMj2EReBNhayxRe:b-rmg13o) ([raw file](https://github.com/scionproto/scion/blob/239b7684a0996697222bd41d977733f3da0d0f62/pkg/private/serrors/errors.go#L79)):*

Reminder: include the changes from #4594 here too after rebase.
</blockquote></details>

I wasn't going to miss that one. It conflicted nicely with the map->slice change :-)
___
*[`pkg/private/serrors/errors.go` line 134 at r8](https://reviewable.io/reviews/scionproto/scion/4586#-O3qOPx1A_gt_h-SGceq:-O3s--SP8ccoc5CiQCdr:b-de4tqb) ([raw file](https://github.com/scionproto/scion/blob/239b7684a0996697222bd41d977733f3da0d0f62/pkg/private/serrors/errors.go#L134)):*

I _think_ you need to check both basic error or joined error here.
Otherwise we have two stack traces if we wrap a joined error.
</blockquote></details>

Yup, I think you're right. I'm considering adding a "stacktracer" implemented by errorInfo, but may be that's overkill.
___
*[`pkg/private/serrors/errors.go` line 257 at r8](https://reviewable.io/reviews/scionproto/scion/4586#-O3qPuIU6jGrAOUPd6YY:-O3s-U720SudjIcgRZJH:b-896fjx) ([raw file](https://github.com/scionproto/scion/blob/239b7684a0996697222bd41d977733f3da0d0f62/pkg/private/serrors/errors.go#L257)):*

nit: (Feel free to ignore)

To make this look nice in pkg.go.dev, you can use go doc supported formatting: https://tip.golang.org/doc/comment#lists
</blockquote></details>

Done

oncilla (changes requested):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion and @lukedirtwalker)

___
*[`pkg/private/serrors/errors.go` line 134 at r8](https://reviewable.io/reviews/scionproto/scion/4586#-O3qOPx1A_gt_h-SGceq:-O44M1_540TDJnrHq4S5:b83psmq) ([raw file](https://github.com/scionproto/scion/blob/239b7684a0996697222bd41d977733f3da0d0f62/pkg/private/serrors/errors.go#L134)):*

Yup, I think you're right. I'm considering adding a "stacktracer" implemented by errorInfo, but may be that's overkill.
</blockquote></details>

I think it is fine as is for now.

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4586)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lukedirtwalker)

jiceatscion:
Took the liberty of bypassing review for the last commit: reformating a comment due to long line.

Issue #4598, opened 2024-08-14
tools: wireshark plugin does not dissect first SCION packet


FR4NK-W:
After the switch to the heuristic SCION traffic identification in #4582 , the first SCION packet in a "conversation" (4-tuple) does not get dissected as SCION traffic, even though the packet is matched by the filter heuristics for SCION.

FR4NK-W:
[repro_dissector_convo.pcap.zip](https://github.com/user-attachments/files/16612369/repro_dissector_convo.pcap.zip)

FR4NK-W:
Behavior before:
![repro_dissector_convo_before](https://github.com/user-attachments/assets/f0885aab-cb15-458c-aa93-b29cb2834f24)

Fixed behavior:
![repro_dissector_convo_fixed](https://github.com/user-attachments/assets/ddfc2eb8-1b1f-4047-8bbf-a72992c45e63)

PR #4599, opened 2024-08-14
tools: fixup the wireshark lua packet dissector heuristic filter

FR4NK-W:
This fixes bug #4598  where the first SCION packet in a capture is never dissected as a SCION packet.

The filter function is (only) called on the first packet of the conversation (4-tuple) since we set the dissector for the conversation.

To also dissect the initial packet of a conversation for which a heuristic dissector function is registered, the dissector has to be called explicitly at the end of the filter function when a packet matches.

From the docs:
https://www.wireshark.org/docs/wsdg_html/#lua_fn_proto_register_heuristic_listname__func_
>The function should perform as much verification as possible to ensure
the payload is for it, and **dissect the packet** (including setting TreeItem info and such) only if the payload is for it, before returning true or false.

Also look at the comments there:
https://github.com/wireshark/wireshark/blob/master/test/lua/dissector.lua#L523 https://github.com/wireshark/wireshark/blob/afff4e02/test/lua/dissector.lua#L523

### DIFF
--- a/tools/wireshark/scion.lua
+++ b/tools/wireshark/scion.lua
@@ -143,11 +143,17 @@ local function scion_proto_filter(tvbuf, pktinfo, root)
     local addr_type_src_valid = (addrTypes[bit.band(tvbuf(9, 1):uint(), 0xf)] ~= nil)
     local rsv_valid = (tvbuf(10, 2):uint() == 0)
     
-    if version_valid and next_hdr_valid and path_type_valid and addr_type_dst_valid and addr_type_src_valid and rsv_valid then
-       pktinfo.conversation = scion_proto
-       return true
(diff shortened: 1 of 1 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4599)

jiceatscion (approved):
Yes, looks like I had it wrong and never noticed that the first packet was un-dissected.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4599#-:-O4G-xyMA56ns0SIzAr3:b7yepyt)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @FR4NK-W)

FR4NK-W:
Closed #4598

PR #4600, opened 2024-08-14
private/app/path: include IA in Sort function

### LINKED ISSUE #4596: private/app/path: IA ignored in Sort function
I came accross the `Sort` function in `path.go` and as I understand it, it should sort paths (in decreasing priority) by length,  IA, ID and expiry time. Currently the `IA` step is skipped due to a minor bug in the following line: https://github.com/scionproto/scion/blob/ad503c1fb89ba583093e2ea8004c7acd5027e010/private/app/path/path.go#L49

I assume the second `intfA[i].IA` should be `intfB[i].IA`.

The more I think about it, I think the whole IA check might be unnecessary. As a given interface always leads to the same IA, sorting the interface IDs also sorts the next IAs. I think there are only two scenarios in which not doing the IA check would change the order and I'm not sure if they are even possible:
- Paths with different starting IAs,
- The same interface leading to two different IAs.

oncilla:
fixes #4596

### DIFF
--- a/private/app/path/path.go
+++ b/private/app/path/path.go
@@ -46,7 +46,7 @@ func Sort(paths []snet.Path) {
 			return len(intfA) < len(intfB)
 		}
 		for i := range intfA {
-			if iaA, iaB := intfA[i].IA, intfA[i].IA; iaA != iaB {
+			if iaA, iaB := intfA[i].IA, intfB[i].IA; iaA != iaB {
 				return iaA < iaB
 			}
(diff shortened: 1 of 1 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4600)

FR4NK-W (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4600)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4607, opened 2024-09-02
daemon: improve revcache performance

lukedirtwalker:
When dealing with a lot of paths the current implementation is not optimal. Specifically filterRevoked showed up quite prominently in a pprof. See image below:

image

- Use a typed cache, to prevent string key conversion.
- Only support single key lookups to prevent map allocation on lookup.

Overall this even simplifies the code.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,6 +54,7 @@ require (
 	google.golang.org/protobuf v1.34.1
 	gopkg.in/yaml.v2 v2.4.0
 	modernc.org/sqlite v1.29.9
+	zgo.at/zcache/v2 v2.1.0
 )
 
 require (
--- a/go.sum
+++ b/go.sum
@@ -497,3 +497,5 @@ modernc.org/strutil v1.2.0/go.mod h1:/mdcBmfOibveCTBxUl5B5l6W+TTH1FXPLHZE6bTosX0
 modernc.org/token v1.1.0 h1:Xl7Ap9dKaEs5kLoOQeQmPWevfnk/DM5qcLcYlA8ys6Y=
 modernc.org/token v1.1.0/go.mod h1:UGzOrNV1mAFSEB63lOFHIpNRUVMvYTc6yu1SMY/XTDM=
 rsc.io/pdf v0.1.1/go.mod h1:n8OzWcQ6Sp37PL01nO98y4iUCRdTGarVfzxY20ICaU4=
+zgo.at/zcache/v2 v2.1.0 h1:USo+ubK+R4vtjw4viGzTe/zjXyPw6R7SK/RL3epBBxs=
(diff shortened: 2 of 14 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4607)

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4607)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion and @lukedirtwalker)

___
*[`private/segment/segfetcher/pather.go` line 138 at r1](https://reviewable.io/reviews/scionproto/scion/4607#-O5lwPaL-9xe8mdIeMja:-O5lwPaL-9xe8mdIeMjb:blbt5rw) ([raw file](https://github.com/scionproto/scion/blob/c9f2a79e490815b7335bdb6d29942e2435230a70/private/segment/segfetcher/pather.go#L138)):*
> ```Go
> 	debugOn := logger.Enabled(log.DebugLevel)
> 	revokedInterfaces := make(map[snet.PathInterface]struct{})
> 	for _, path := range paths {
> ```

I wonder if it would make sense to deduplicate the requests to the revcache. But on the other hand it is also essentially a map lookup in the revcache too.

lukedirtwalker (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4607)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion and @oncilla)

___
*[`private/segment/segfetcher/pather.go` line 138 at r1](https://reviewable.io/reviews/scionproto/scion/4607#-O5lwPaL-9xe8mdIeMja:-O5lxQPe4oSFvhlRX2yw:b-rfl6r7) ([raw file](https://github.com/scionproto/scion/blob/c9f2a79e490815b7335bdb6d29942e2435230a70/private/segment/segfetcher/pather.go#L138)):*

I wonder if it would make sense to deduplicate the requests to the revcache. But on the other hand it is also essentially a map lookup in the revcache too.
</blockquote></details>

I would say not. You would need some map to keep track of what requests to do, which means you need to do allocation for the map. And as you say it's only a map lookup behind, so IMO we don't need to dedupe.

oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4607#-:-O5lzytg1yyjpqJZyqgA:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4607)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

PR #4619, opened 2024-09-15
tutorial: topology source host missing causes scion-router@br.service to crash

thomasgruebl:
Hi,

I've tried to setup the freestanding SCION network as per the [tutorial](https://docs.scion.org/en/latest/tutorials/deploy.html) and ran into a few issues:

Using the default topology files from the website, the scion-router service always crashes because of an invalid source IP (see screenshot below):

image

@tobiasmoldan realized that this can be resolved by adding the source IP (or host name) to the local field in the topology:

`"underlay": {
            "local": "scion01:50012",
            "remote": "scion02:50012"
          },`

Secondly, I was wondering if all ASes need to be up and running for the network to be fully functional? I have set up and configured scion01 and scion02 as per the tutorial. All services (daemon, router, control, dispatcher) are active and there is a direct link between the two VMs. The machines are pingable (via normal ICMP), however, there is no reachability via SCMP. I assume that, even without scion03-05, they should have connectivity?

I got scion addresses for scion01 and scion02 respectively,

`42-ffaa:1:1,127.0.0.1`

`42-ffaa:1:2,127.0.0.1`

and I have generated the certificate material on scion01 and shared it with scion02. Is there anything I'm missing - e.g., does path exploration need to be triggered somehow or should it happen automatically?

Thanks a lot for your help on this!

Best regards,

Thomas


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4619)

jiceatscion (approved):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4619#-:-O6zCPkJ8XlrJs6_GQ3Q:b-ucfknh)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @thomasgruebl)

thomasgruebl:
Hi,

thanks for your response! Yes it's odd indeed. I have now configured scion03 as well (so all three core ASes are up and running). In Wireshark, I can even see path propagation packets being sent (and received) between all 3 ASes via UDP ports 50012, 50013, 50023, however, scion ping, showpaths and traceroute are still not working and outputting:

`Error: fetching paths: no path available`

I was wondering if you'd be available for a quick remote discussion sometime - then I could show you the setup?

Thanks,

Thomas

thomasgruebl:
Hi @jiceatscion ,

I've now installed the SCION protocol dissector for Wireshark and I'm seeing the following messages being exchanged between the two ASes.

Screenshot 2024-09-21 at 13 14 23

This is a request from scion01 -> scion02:

Screenshot 2024-09-21 at 13 16 06


This is a response from scion02 -> scion01


Screenshot 2024-09-21 at 13 16 28


Can you infer something from the BFD control messages?

Thanks!

Thomas

jiceatscion:
Not sure if we can learn much from the BFD messages. Those don't call for a response; they're just mutual keep-alive packets. If they weren't there, then we would have a hint of why paths aren't available.

Recently we found that Go v1.23 was including a ~bug in timers behavior that was breaking the control service RPCs. If you're building the last release, you shouldn't be affected. But if you're building from head, then make sure you've pulled past the fix that downgrades Go to 1.22.7

thomasgruebl:
Hi @jiceatscion ,

Thanks for your reply! Unfortunately, that's not the case because I'm building the last release v0.11.0.

PR #4624, opened 2024-09-17
docs: add interactive TRC signing ceremony builder

oncilla:
And an interactive TRC signing ceremony builder.
It supports all three types of TRC ceremony: base, regular, and sensitive. The user can select between scion-pki and openssl. Furthermore, support for pkcs11 is available for openssl. In a future iteration, support for scion-pki kms will be added.

The builder is intended to be used by a voting (or root CA) participant. The adminstrator role is not supported in this version. (Administrators should be very knowledgable about the TRC ceremony anyway.)

The builder is based on aplinejs and tailwindcss. This allows us to included it in our RTD page with zero dependency and build steps.

image

[doc]


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4624)

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4624)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @oncilla)

PR #4629, opened 2024-09-20
details: nits from SCIONLab

juagargi:
While in the process of merging FABRID into SCIONLab, we have found these very small nits to probably be beneficial to merge upstream (here):
- Allow the parsing of ASes from string using a different separator than `:`
- Improve the comment on the package variable `HostAddr` in package `tools/integration`.

### DIFF
--- a/pkg/addr/fmt.go
+++ b/pkg/addr/fmt.go
@@ -65,7 +65,7 @@ func ParseFormattedAS(as string, opts ...FormatOption) (AS, error) {
 		}
 		as = trimmed
 	}
-	return parseAS(as, o.separator)
+	return ParseASSep(as, o.separator)
 }
 
 // FormatIA formats the ISD-AS.
--- a/pkg/addr/fmt.go
+++ b/pkg/addr/fmt.go
@@ -65,7 +65,7 @@ func ParseFormattedAS(as string, opts ...FormatOption) (AS, error) {
 		}
 		as = trimmed
 	}
-	return ParseASSep(as, o.separator)
+	return parseAS(as, o.separator)
 }
 
 // FormatIA formats the ISD-AS.
--- a/pkg/addr/isdas.go
+++ b/pkg/addr/isdas.go
@@ -77,7 +77,7 @@ type AS uint64
 // ParseAS parses an AS from a decimal (in the case of the 32bit BGP AS number
 // space) or ipv6-style hex (in the case of SCION-only AS numbers) string.
 func ParseAS(as string) (AS, error) {
-	return parseAS(as, ":")
+	return ParseASSep(as, ":")
 }
 
 // MustParseAS parses s and returns the corresponding addr.AS object. It panics
@@ -90,7 +90,7 @@ func MustParseAS(s string) AS {
 	return as
 }
 
-func parseAS(as string, sep string) (AS, error) {
+func ParseASSep(as string, sep string) (AS, error) {
 	parts := strings.Split(as, sep)
 	if len(parts) == 1 {
 		// Must be a BGP AS, parse as 32-bit decimal number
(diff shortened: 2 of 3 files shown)


oncilla (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4629)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @juagargi)

___
*[`pkg/addr/fmt.go` line 59 at r2](https://reviewable.io/reviews/scionproto/scion/4629#-O7EZl-hAlX5bjMNo3RH:-O7EZl-i3gPdMzZy_fbA:b-qvnkm9) ([raw file](https://github.com/scionproto/scion/blob/22c5c91441b5c0ea732b77f95fb72edde078c68c/pkg/addr/fmt.go#L59)):*
> ```Go
> // ParseFormattedAS parses an AS number that was formatted with the FormatAS
> // function. The same options must be provided to successfully parse.
> func ParseFormattedAS(as string, opts ...FormatOption) (AS, error) {
> ```

Is this not sufficient for your use case?

If not, what is blocking you from using this instead of the newly introduced function?

juagargi:
> > _[Reviewable](https://reviewable.io/reviews/scionproto/scion/4629)_ status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @juagargi)
> 
> _[`pkg/addr/fmt.go` line 59 at r2](https://reviewable.io/reviews/scionproto/scion/4629#-O7EZl-hAlX5bjMNo3RH:-O7EZl-i3gPdMzZy_fbA:b-qvnkm9) ([raw file](https://github.com/scionproto/scion/blob/22c5c91441b5c0ea732b77f95fb72edde078c68c/pkg/addr/fmt.go#L59)):_
> 
> > ```go
> > // ParseFormattedAS parses an AS number that was formatted with the FormatAS
> > // function. The same options must be provided to successfully parse.
> > func ParseFormattedAS(as string, opts ...FormatOption) (AS, error) {
> > ```
> 
> Is this not sufficient for your use case?
> 
> If not, what is blocking you from using this instead of the newly introduced function?

I'm tagging @marcodermatt so that he sees this PR and discussion.
In my opinion, `fmt.ParseFormattedAS` should be enough, but I don't know enough about FABRID or the performance they would need to have a canonical answer ...

jeltevanbommel:
*[`pkg/addr/fmt.go` line 59 at r2](https://reviewable.io/reviews/scionproto/scion/4629#-O7EZl-hAlX5bjMNo3RH:-O7EdjQn1WxB0UwBWHWH:b-8tgne) ([raw file](https://github.com/scionproto/scion/blob/22c5c91441b5c0ea732b77f95fb72edde078c68c/pkg/addr/fmt.go#L59)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

I'm tagging @marcodermatt so that he sees this PR and discussion.
In my opinion, `fmt.ParseFormattedAS` should be enough, but I don't know enough about FABRID or the performance they would need to have a canonical answer ...

</blockquote></details>

We will use fmt.ParseFormattedAS indeed, so the change is not necessary. Thanks @oncilla!

oncilla (approved):
> Reviewed 1 of 2 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4629)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @juagargi)

oncilla (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4629#-:-O7EkMbfEdvUFJSp3L1i:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @juagargi)

PR #4650, opened 2024-11-07
router: refactor the WriteTo method out of existence

jiceatscion:


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4650)

PR #4652, opened 2024-11-18
build: work around issues between pkg_rpm and rpmbuild after 4.20

### LINKED ISSUE #4653: buildrpm 4.20 and later are not supported
No one at the moment can run make test-integration on fedora 41, and likely not on any platform using rpm v 4.20.

This is a result of incompatible changes in rpmbuild colliding with the assumptions made by the pgk_rpm rules.

jiceatscion:
Since rpmbuild v4.20, the _builddir variable is forcefully set to <chosen_builddir>/<pkg_name> (which is created if it doesn't exist). Any files that are not built/copied by the %prep% phase must be already present in <chosen_builddir>/<pkg_name>/[%buildsubdir].

pkg_rpm doesn't expect that and places the files to be packaged in <chosen_builddir>. Since %buildsubdir is not used (=="."), we compensate by setting it to "..". As a result <chosen_builddir>/<pkg_name>/[%buildsubdir] evaluates to <chosen_buildir> thereby restoring pre-4.20 behavior.

Since 4.20, debug-info rpms are built by default. We don't support or want that, so we now set "don't-do-it" explicitly.

Fixes #4653

### DIFF
--- a/dist/package.bzl
+++ b/dist/package.bzl
@@ -121,6 +121,17 @@ def scion_pkg_rpm(name, package, executables = {}, systemds = [], configs = [], 
     if post:
         kwargs.pop("postinst")
 
+    # defines are work-arounds for pkg_rpm issues:
+    # _smp_build_ncpus: used to be default.
+    # buildsubdir: starting with rpmbuild 4.20, builddir is forcefully replaced with:
+    #              <chosen_builddir>/<package_name> (which is created if it doesn't exist).
+    #              pkg_rpm assumes that one can place ready-to-package files directly into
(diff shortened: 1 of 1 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4652)

oncilla (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4652)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

Issue #4593, opened 2024-08-06
router: refactor code to enable multiple packet I/O implementations


jiceatscion:
To implement fast packet IOs consistent with 10G+ NICs, we will need a very different underlay networking code from what we have today. For example using raw sockets, XDP or DPDK. Some of these frameworks are only available on Linux or different APIs on other platforms, and then not necessarily for all types of NICs. Since we want the reference SCION implementation to be highly portable, we need to support more than one implementation of the underlay network and allow the best applicable one to be used on a given platform. The SCION router code originally makes the assumption that there is a single underlay implementation.

In addition, there has been experiments at using different, more economical, underlay protocols; such as MPLS, which is permitted by the SCION design but not currently feasible due to the single underlay implementation assumption as well as a hardcoded UDP/IP underlay addressing assumption.

At least the first, but preferably both of these issues must be address in order to support high performance underlay implementations.

Note: This issue's scope has been reduced to router refactoring. The other tasks are now tracked by https://github.com/orgs/scionproto/projects/1

jiceatscion:
In order to preserve portability to non-linux platforms, we'll need to preserve the plain IP/UDP socket code. Likewise, it is possible that the AF_XDP code isn't as portable as the AF_PACKET with RINGS code, so we probably want to keep both if AF_XDP is much better. Action plan:

[ ] Refactor code to enable multiple packet I/O implementations. Cut just high enough to move the plain-socket related code below that interface.
[ ] Add a second implementation that uses AF_PACKET with RX/TX rings.
[ ] Add a third implementation that uses AF_XDP.

PR #4658, opened 2024-11-29
router: make room for multiple underlay impls phase 1

jiceatscion:
The goal is to eventually remove all underlay-specific code from the main router code and place it in a plugin, such that there can then be multiple underlay plugins. This PR isn't attempting to fully accomplish that. It avoids moving some code, so that the diff is easier to read. The price to pay is a few extra spaghetti left between main code and plugin.

The code that should eventually move to the underlay is:
* runReceiver and runForwarder.
* likely some of the BFD code.
* opening of connections (currently done by connector.go)

Other changes being planned:
* Stop reusing the internal connection for sibling links (so we can take advantage of bound connections).
* Add knowledge of multiple underlay into the configurator.
* Make underlay addresses opaque to the router.
* Demultiplex to links on ingest, so ifID and srcAddress (when they are defined by the link) are obtained in the most efficient way (directly from the link's fields for example).

Contributes to: #4593


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4658)

tzaeschke (commented):
Iḿ not really familiar with this module and I'm probably missing a lot of context, so my review is almost exclusively about comments and naming.

:lgtm:

> Reviewed 1 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4658#-:-OHbEuPT88Hw3Zup00sF:bdiwaku)* status: 1 of 11 files reviewed, 22 unresolved discussions (waiting on @jiceatscion)

___
*[`-- commits` line 16 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb4W3M-_cXSsGfafAN:-OHb4W3M-_cXSsGfafAO:b-79qu3x):*
Since it is incomplete, maybe add some points about what exactly still needs to be done?
___
*[`router/underlay.go` line 33 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbCFDC69cn-rspDJjY:-OHbCFDC69cn-rspDJjZ:b-xijwt9) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlay.go#L33)):*
> ```Go
> 
> // Link embodies the router's idea of a point to point connection. A link associates the underlay
> // connection, with a bfdSession, a destination address, etc. It also allows the concrete send
> ```

connection with



_Code quote:_
```Go
connection, with
```
___
*[`router/underlay.go` line 48 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbCVmgCBVITj-pAfuT:-OHbCVmgCBVITj-pAfuU:b-80z19o) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlay.go#L48)):*
> ```Go
> 	IsUp() bool
> 	GetIfID() uint16
> 	GetRemote() netip.AddrPort // incremental refactoring: using code will move to underlay.
> ```

TODO?
___
*[`router/underlay.go` line 59 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbCaMJ4Jgp22TTEjz_:-OHbCaMJ4Jgp22TTEjza:b-80z19o) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlay.go#L59)):*
> ```Go
> //
> // Incremental refactoring: addresses are still explicitly IP/port. In the next step, we have to
> // make them opaque; to be interpreted only by the underlay implementation.
> ```

TODO?
___
*[`router/underlayproviders/udpip.go` line 27 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbD4gKE0UFOdHWlci4:-OHbD4gL2NBH4yxfMMwW:b2nhp4s) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlayproviders/udpip.go#L27)):*
> ```Go
> // This is currently the only implementation. The goal of splitting out this code from the router
> // is to enable other implementations. However, as a first step, we continue assuming that the
> // batchConn is given to us and is a Udp socket and that, in the case of externalLink, it is bound.
> ```

UDP



_Code quote:_
```Go
Udp
```
___
*[`router/underlayproviders/udpip.go` line 49 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbDNWPAV1B39e5Dyt9:-OHbDNWPAV1B39e5DytA:b3krg) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlayproviders/udpip.go#L49)):*
> ```Go
> 
> func (u *provider) GetConnections() map[netip.AddrPort]router.UnderlayConnection {
> 	// a map of interfaces and a map of concrete implementations aren't compatible.
> ```

A



_Code quote:_
```Go
a
```
___
*[`router/underlayproviders/udpip.go` line 50 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbDTW1AYLtxtFxuo-d:-OHbDTW1AYLtxtFxuo-e:b-pyansp) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlayproviders/udpip.go#L50)):*
> ```Go
> func (u *provider) GetConnections() map[netip.AddrPort]router.UnderlayConnection {
> 	// a map of interfaces and a map of concrete implementations aren't compatible.
> 	// For the same reason, we cannot have the map of concrete as our return type; it
> ```

concrete implementations as



_Code quote:_
```Go
concrete as
```
___
*[`router/underlayproviders/udpip.go` line 55 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbDkZ42r6W10p1J8Oi:-OHbDkZ52QcmZS-xCfQ2:b-30d7r4) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlayproviders/udpip.go#L55)):*
> ```Go
> 	// Since we do not want to store our own things as interfaces, we have to translate.
> 	// Good thing it doesn't happen often.
> 	m := make(map[netip.AddrPort]router.UnderlayConnection)
> ```

How often is `map()` called? Is it worth caching it?
___
*[`router/underlayproviders/udpip.go` line 90 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbE54a9CrOJbErAK4I:-OHbE54a9CrOJbErAK4J:b-80z19o) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlayproviders/udpip.go#L90)):*
> ```Go
> }
> 
> // Incremental refactoring: The following implements UnderlayConnection so some of the code
> ```

TODO?
___
*[`router/underlayproviders/udpip.go` line 202 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbET4qEwTeYrUABnca:-OHbET4qEwTeYrUABncb:b-7iwoov) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlayproviders/udpip.go#L202)):*
> ```Go
> // to erase the separation between link and connection for this implementation. Side effect
> // of moving the address:link here: the router does not know if there is an existing link. As
> // a result it has to give us a bfdSession in all cases and if we might throuw it away (there
> ```

throw



_Code quote:_
```Go
throuw
```
___
*[`router/underlayproviders/udpip.go` line 204 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbEXD84eNiaveqdmD7:-OHbEXD84eNiaveqdmD8:b-1yafxs) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlayproviders/udpip.go#L204)):*
> ```Go
> // a result it has to give us a bfdSession in all cases and if we might throuw it away (there
> // are no permanent resources attached to it). This will be fixed by moving some bfd related code
> // in-here; but later.
> ```

TODO? Maybe a ticket?
___
*[`router/underlayproviders/udpip.go` line 219 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHbEmB023zOlISQ6q2Y:-OHbEmB023zOlISQ6q2Z:b-80z19o) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/underlayproviders/udpip.go#L219)):*
> ```Go
> 	c, exists := u.allConnections[netip.AddrPort{}]
> 	if !exists {
> 		// That doesn't actually happen and we'll get rid of this soon.
> ```

TODO?
___
*[`router/dataplane.go` line 232 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb4p8sADc-onOyOand:-OHb4p8sADc-onOyOane:b5yppwi) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/dataplane.go#L232)):*
> ```Go
> 	unsupportedUnspecifiedAddress = errors.New("unsupported unspecified address")
> 	noBFDSessionFound             = errors.New("no BFD session was found")
> 	// noBFDSessionConfigured        = errors.New("no BFD sessions have been configured")
> ```

Maybe remove this?
___
*[`router/dataplane.go` line 356 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb4xWSBvmj9dmNdi2B:-OHb4xWT3uWpOhxjGMGZ:b-3j3e4) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/dataplane.go#L356)):*
> ```Go
> 	}
> 
> 	l := d.underlay.NewInternalLink(conn, d.RunConfig.BatchSize)
> ```

rename to "link" or something longer than "l" (which I first read as "1")?
___
*[`router/dataplane.go` line 392 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb5Bua1f73tKCD1Cbx:-OHb5Bua1f73tKCD1Cby:byd4kvm) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/dataplane.go#L392)):*
> ```Go
> 	}
> 	lk := d.underlay.NewExternalLink(conn, d.RunConfig.BatchSize, bfd, dst.Addr, ifID)
> 	d.interfaces[ifID] = lk
> ```

lk -> link?
___
*[`router/dataplane.go` line 648 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb5Srz3vs1zb26Ixz1:-OHb5Srz3vs1zb26Ixz2:by10k4u) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/dataplane.go#L648)):*
> ```Go
> 	}
> 
> 	for a, l := range d.underlay.GetLinks() {
> ```

Use slightly more descriptive variable names?
___
*[`router/dataplane.go` line 741 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb5i_o01QJbejLHnpq:-OHb5i_o01QJbejLHnpr:b-vsp8ck) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/dataplane.go#L741)):*
> ```Go
> 
> 		pkt.rawPacket = pkt.rawPacket[:size] // Update size; readBatch does not.
> 		// Incremental refactoring: We should begin with finding the link and get the ifID
> ```

Is it common in scionproto to insert a TODO here?
___
*[`router/dataplane.go` line 1023 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb6-Od-goS-2gr3ZH4:-OHb6-Od-goS-2gr3ZH5:b-11904j) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/dataplane.go#L1023)):*
> ```Go
> 
> // Big issue with metrics and ifID. If an underlay connection must be shared between links
> // (for example, libling links), then we don't have a specific ifID in the connection per se. It
> ```

sibling



_Code quote:_
```Go
libling
```
___
*[`router/dataplane.go` line 1035 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb6F5W9FOm_akHmpeY:-OHb6F5X1tIpODWjZk-A:b8f2a1n) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/dataplane.go#L1035)):*
> ```Go
> //     metrics... might not be much cheaper than the naive way.
> //   - Use one fw queue per ifID in each connection... but then have to round-robin for fairness....
> //     smaller batches?
> ```

That's also a TODO, isn't it?
___
*[`router/dataplane.go` line 1203 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb6X-s30OAeHckIHtK:-OHb6X-s30OAeHckIHtL:b-fwzidc) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/dataplane.go#L1203)):*
> ```Go
> 	// If this is an inter-AS BFD, it can via an interface we own. So the ifID matches one link
> 	// and the ifID better be valid. In the future that will be checked upstream from here.
> 	l, exists := p.d.interfaces[p.pkt.ingress]
> ```

`l` is an interface? Maybe rename to `intf` (or `link`) or something?
___
*[`router/dataplane.go` line 1528 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb7-AVC8GIRp072zrP:-OHb7-AVC8GIRp072zrQ:b-80z19o) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/dataplane.go#L1528)):*
> ```Go
> // SrcIA checks by disguising packets as transit traffic.
> //
> // Incremental refactoring: All or part of this check should move to the underlay.
> ```

TODO?
___
*[`router/dataplane.go` line 1548 at r3](https://reviewable.io/reviews/scionproto/scion/4658#-OHb77uc6YZG4ctnwcDJ:-OHb77ud8Uuc6AYP2k8j:b-py4ko6) ([raw file](https://github.com/scionproto/scion/blob/a2fb58c1186dbcbafc115163299028f4dc686833/router/dataplane.go#L1548)):*
> ```Go
> }
> 
> // Validates the egress interface referenced by the current hop. This  is not called for
> ```

This is not



_Code quote:_
```Go
This  is not
```

tzaeschke (commented):
:lgtm:

> Reviewed 7 of 11 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4658#-:-OHlIrmw2YVFcR-WGl-i:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

marcfrei (changes requested):
> Reviewed 8 of 11 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4658)* status: all files reviewed, 13 unresolved discussions (waiting on @jiceatscion)

___
*[`router/dataplane.go` line 81 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHngCQD-5mEGieM73vx:-OHngCQD-5mEGieM73vy:bsqvdqf) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/dataplane.go#L81)):*
> ```Go
> )
> 
> type BfdSession interface {
> ```

`BFDSession`

https://go.dev/wiki/CodeReviewComments#initialisms



_Code quote:_
```Go
BfdSession
```
___
*[`router/dataplane.go` line 472 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHn2sRL9DzGJnbs0LnH:-OHn2sRL9DzGJnbs0LnI:b-l4qtn3) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/dataplane.go#L472)):*
> ```Go
> }
> 
> func (d *DataPlane) addBFDController(ifID uint16, s bfd.Sender, cfg control.BFD,
> ```

This doesn't really add a `BFDController` (or BFDSession) anymore. Does it make sense to rename the method and maybe even make it a free-standing function instead of a method on `DataPlane`? The same might also apply to some of the methods using this one.



_Code quote:_
```Go
addBFDController
```
___
*[`router/dataplane.go` line 2474 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHneAQnBGJFaVgP9Qqn:-OHneAQnBGJFaVgP9Qqo:b-6oxt0z) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/dataplane.go#L2474)):*
> ```Go
> 	// If the packet is sent to an external router, we need to increment the
> 	// path to prepare it for the next hop.
> 	// This is an scmp response to pkt, so egress will be pkt.ingress.
> ```

`SCMP`?



_Code quote:_
```Go
scmp
```
___
*[`router/underlayproviders/udpip.go` line 63 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnlHjk63VBNhtsfcE6:-OHnlHjk63VBNhtsfcE7:by6gpeu) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlayproviders/udpip.go#L63)):*
> ```Go
> 
> func (u *provider) GetLinks() map[netip.AddrPort]router.Link {
> 	return u.allLinks
> ```

Should we copy at least the map here? I tend towards 'yes'.



_Code quote:_
```Go
	return u.allLinks
```
___
*[`router/underlay.go` line 44 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnhTDl3-m_VWV6wzTb:-OHnhTDmD_fATugkWqNs:b-66q0m1) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L44)):*
> ```Go
> // the link is a sibling link. If the interface ID is zero, then the link is the internal link.
> type Link interface {
> 	GetScope() LinkScope
> ```

`Scope`



_Code quote:_
```Go
	GetScope
```
___
*[`router/underlay.go` line 45 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnhb131Gir9wluTH8x:-OHnhb140UuGV3QTv3yy:beaqxjj) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L45)):*
> ```Go
> type Link interface {
> 	GetScope() LinkScope
> 	GetBfdSession() BfdSession
> ```

`BFDSession`



_Code quote:_
```Go
	GetBfdSession
```
___
*[`router/underlay.go` line 47 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnhyaLDNCLZXFm0dxs:-OHnhyaLDNCLZXFm0dxt:b-u4zmxf) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L47)):*
> ```Go
> 	GetBfdSession() BfdSession
> 	IsUp() bool
> 	GetIfID() uint16
> ```

`IfID`



_Code quote:_
```Go
	GetIfID
```
___
*[`router/underlay.go` line 48 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHniW2I5Yh0837kcoc9:-OHniW2I5Yh0837kcocA:bp7gjwk) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L48)):*
> ```Go
> 	IsUp() bool
> 	GetIfID() uint16
> 	GetRemote() netip.AddrPort // TODO(multi_underlay): using code will move to underlay.
> ```

`Remote` or `RemoteAddrPort`



_Code quote:_
```Go
	GetRemote
```
___
*[`router/underlay.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnj1d10J3sg_fw8jb0:-OHnj1d10J3sg_fw8jb1:b-c45by3) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L50)):*
> ```Go
> 	GetRemote() netip.AddrPort // TODO(multi_underlay): using code will move to underlay.
> 	Send(p *Packet) bool
> 	BlockSend(p *Packet)
> ```

Should this really be part of the interface? It's only used for testing, right?



_Code quote:_
```Go
	BlockSend(p *Packet)
```
___
*[`router/underlay.go` line 86 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnj_sV1rY94orpOcBz:-OHnj_sV1rY94orpOcC-:b-x6e3ei) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L86)):*
> ```Go
> 	// dataplane code for now. There may be fewer connections than links. For example, right now
> 	// all sibling links and the internal link use a shared un-bound connection.
> 	GetConnections() map[netip.AddrPort]UnderlayConnection
> ```

`Connections`



_Code quote:_
```Go
	GetConnections
```
___
*[`router/underlay.go` line 93 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnjjMHA0eKIhl8OQj2:-OHnjjMHA0eKIhl8OQj3:b-9erc66) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L93)):*
> ```Go
> 	// There may be fewer links than ifIDs. For example, all interfaces owned by one given sibling
> 	// router are connected via the same link because the remote address is the same.
> 	GetLinks() map[netip.AddrPort]Link
> ```

`Links`



_Code quote:_
```Go
	GetLinks
```
___
*[`router/underlay.go` line 101 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnjwLk2ENtFuTbmzBB:-OHnjwLk2ENtFuTbmzBC:bulhskp) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L101)):*
> ```Go
> 	// matched with a link), on ingest by the underlay. That would imply moving a part of the
> 	// runReceiver routine to the underlay. We will do that in the next step.
> 	GetLink(netip.AddrPort) Link
> ```

`Link` or `LinkFor`



_Code quote:_
```Go
	GetLink
```
___
*[`router/underlay.go` line 109 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnkPlh9KH-1PxHPLZO:-OHnkPlh9KH-1PxHPLZP:b-j1vn6j) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L109)):*
> ```Go
> // TODO(multi_underlay): this will eventually be reduced to nothing at all because the sender
> // receiver tasks will be part of the underlay.
> type UnderlayConnection interface {
> ```

`UnderlayConn` and the first method then `BatchConn()`?



_Code quote:_
```Go
UnderlayConnection
```

marcfrei (changes requested):
> Reviewed 2 of 4 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4658)* status: all files reviewed, 7 unresolved discussions (waiting on @jiceatscion)

___
*[`router/dataplane.go` line 472 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHn2sRL9DzGJnbs0LnH:-OHvJCSm5i9XxOP_0pLQ:b60oy57) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/dataplane.go#L472)):*

Done. I agree that this in not a method and not an Add anymore. Renamed "NewBFDController" for the same reason. renamed AddNextHopBFD and AddExternalInterfaceBFD (although they are still methods od DataPlane. One thing I noticed is that newBFDController doesn't seem like it even belong in DataPlane anymore. Should we move it to bfd/session.go ?
</blockquote></details>

Yes, moving to bfd/session.go sounds good. And should we rename it to `newBFDSession`?

The other question would be whether we could also directly use *bfd.Session instead of having the interface `BFDSession`. bfd.Session has already abstracted away the sending and receiving part. But maybe that's for another refactoring.
___
*[`router/underlay.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnj1d10J3sg_fw8jb0:-OHv_Q1l-4vdWWaX0cNj:b-tzsx6b) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L50)):*

Yes, it is only used by tests. Unfortunately, there's no reasonably simple way of implementing that only in test code. It would take a ridiculous amount of plumbing. Anything wrong about actually having a BlockSend method, even if only for testability? Otherwise, any suggestion?
</blockquote></details>

Ok. But I would at least rename it to `SendBlocking`. I think that's more clear and it also shows that it's a variant of `Send
___
*[`router/underlayproviders/udpip.go` line 95 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvcIb528nWJIGEOa4n:-OHvcIb528nWJIGEOa4o:bcbsqkz) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L95)):*
> ```Go
> // next step
> 
> // Name returns the name (for logging) associated with a connection.
> ```

Remove comment or fix it.
___
*[`router/underlayproviders/udpip.go` line 100 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvcT1mATLOQ59UZ3UU:-OHvcT1mATLOQ59UZ3UV:bcbsqkz) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L100)):*
> ```Go
> }
> 
> // Name returns the name (for logging) associated with a connection.
> ```

Remove comment or fix it.
___
*[`router/underlayproviders/udpip.go` line 105 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvcTj4EGjsC3FIrf7L:-OHvcTj4EGjsC3FIrf7M:b-vfv9pl) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L105)):*
> ```Go
> }
> 
> // Name returns the name (for logging) associated with a connection.
> ```

Remove comment?
___
*[`router/underlayproviders/udpip.go` line 110 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvcUrw-jGQuzOGHzB9:-OHvcUrw-jGQuzOGHzBA:bcbsqkz) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L110)):*
> ```Go
> }
> 
> // Name returns the name (for logging) associated with a connection.
> ```

Remove comment or fix it.
___
*[`router/underlayproviders/udpip.go` line 255 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvdAtw9MmL5qWytCzO:-OHvdAtw9MmL5qWytCzP:b-sm9pop) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L255)):*
> ```Go
> 
> func (l *siblingLink) Send(p *router.Packet) bool {
> 	// We use an unbound connection but we offer a connected-oriented service. So, we need to
> ```

`connection-oriented`?



_Code quote:_
```Go
connected-oriented
```
___
*[`router/underlayproviders/udpip.go` line 267 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvdKYQ5F7_t1pbLfOY:-OHvdKYQ5F7_t1pbLfOZ:b-sm9pop) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L267)):*
> ```Go
> 
> func (l *siblingLink) BlockSend(p *router.Packet) {
> 	// We use an unbound connection but we offer a connected-oriented service. So, we need to
> ```

`connection-oriented`?



_Code quote:_
```Go
connected-oriented
```

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4658)* status: 7 of 13 files reviewed, 6 unresolved discussions (waiting on @marcfrei and @tzaeschke)

___
*[`router/dataplane.go` line 472 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHn2sRL9DzGJnbs0LnH:-OHwCNO7El0H9OU72NEq:bc8mlpm) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/dataplane.go#L472)):*

Yes, moving to bfd/session.go sounds good. And should we rename it to `newBFDSession`?

The other question would be whether we could also directly use *bfd.Session instead of having the interface `BFDSession`. bfd.Session has already abstracted away the sending and receiving part. But maybe that's for another refactoring.
</blockquote></details>

With you on both accounts. Moved/renamed, but keeping interface for now.
___
*[`router/underlay.go` line 50 at r4](https://reviewable.io/reviews/scionproto/scion/4658#-OHnj1d10J3sg_fw8jb0:-OHwDPU2--rt3le5cozd:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/2df9f4dc82d5ea16849a63c4ea5d48bc519a3d6e/router/underlay.go#L50)):*

Ok. But I would at least rename it to `SendBlocking`. I think that's more clear and it also shows that it's a variant of `Send
</blockquote></details>

Done.
___
*[`router/underlayproviders/udpip.go` line 95 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvcIb528nWJIGEOa4n:-OHwIHcr-VJdR-mM095G:b-di29ub) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L95)):*

Remove comment or fix it.
</blockquote></details>

Done moved doc strings to the interface and fixed them all.
___
*[`router/underlayproviders/udpip.go` line 100 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvcT1mATLOQ59UZ3UU:-OHwIN_OEsmwrh7Y-qxT:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L100)):*

Remove comment or fix it.
</blockquote></details>

Done.
___
*[`router/underlayproviders/udpip.go` line 105 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvcTj4EGjsC3FIrf7L:-OHwIOrE0RtTtjnlEnL_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L105)):*

Remove comment?
</blockquote></details>

Done.
___
*[`router/underlayproviders/udpip.go` line 110 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvcUrw-jGQuzOGHzB9:-OHwIPfl8Ll5bRNWQNrh:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L110)):*

Remove comment or fix it.
</blockquote></details>

Done.
___
*[`router/underlayproviders/udpip.go` line 267 at r6](https://reviewable.io/reviews/scionproto/scion/4658#-OHvdKYQ5F7_t1pbLfOY:-OHwIbuY0cf_oDlX1-j8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/48fcef7b1f75b7967719a777bb120b49202569d9/router/underlayproviders/udpip.go#L267)):*

`connection-oriented`?
</blockquote></details>

Done.

marcfrei (changes requested):
> Reviewed 3 of 6 files at r7, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4658)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

___
*[`router/bfd/session.go` line 167 at r8](https://reviewable.io/reviews/scionproto/scion/4658#-OIB8UFJ7dk1LOq1drAt:-OIB8UFJ7dk1LOq1drAu:bgqwi7y) ([raw file](https://github.com/scionproto/scion/blob/4f70f2e4f024552ceff555994b36303530bdc81e/router/bfd/session.go#L167)):*
> ```Go
> }
> 
> func NewBFDSession(ifID uint16, s Sender, cfg control.BFD,
> ```

Looking at the function now that it is in package `bfd` I'm not sure anymore whether this was a good idea in the first place. Multiple concerns:
- It should be called `NewSession` now
- Argument `ifID` is not used
- It pulls in the dependency on `github.com/scionproto/scion/router/control`
- How general is this random discriminator functionality?
- Hardcoded `ReceiveQueueSize`

Not sure how unacceptable these points really are, but in doubt I'd probably keep the function (as `newBFDSession`)  in dataplane.go for now. Sorry.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4658)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @marcfrei)

___
*[`router/bfd/session.go` line 167 at r8](https://reviewable.io/reviews/scionproto/scion/4658#-OIB8UFJ7dk1LOq1drAt:-OIBVREdDgLXz6pWMF58:bqv5a01) ([raw file](https://github.com/scionproto/scion/blob/4f70f2e4f024552ceff555994b36303530bdc81e/router/bfd/session.go#L167)):*

Looking at the function now that it is in package `bfd` I'm not sure anymore whether this was a good idea in the first place. Multiple concerns:
- It should be called `NewSession` now
- Argument `ifID` is not used
- It pulls in the dependency on `github.com/scionproto/scion/router/control`
- How general is this random discriminator functionality?
- Hardcoded `ReceiveQueueSize`

Not sure how unacceptable these points really are, but in doubt I'd probably keep the function (as `newBFDSession`)  in dataplane.go for now. Sorry.
</blockquote></details>

I think that it is better to fix these things and leave the code here.
- NewSession: Done. Indeed.
- ifID not used: Gone. It was no-longer used before; I should have removed it anyway.
- dependency on controller (sure, that's normal, it provides the BFD configuration). Controller.go will need a bit of refactor in the next phase anyway, I might want to move the creation of bfd sessions down to underlay.go anyway, so we'll see to the config then, if we need to.
- the discriminator is no worse or better than before. It actually doesn't mater; we don't really use it. This is one thing that could stay in the router, given that it's the router that doesn't need it... but then, having a reasonable built-in is more convenient (and ensures protocol compliance).
- queue size: it is better hard-coded here than in the caller. It is a matter of implementation. The caller should certainly not know what to use.

I think that your concerns stem from the fact that bfd now looks more like a general purpose library than before, but thanks to Go's 1-level package hierarchy, it already was exported; just without a constructor. Doesn't mean that we need to support non-router usage. Given that it was already exposed to such non-router use, this is actually an improvement.

marcfrei (approved):
:lgtm:

> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4658#-:-OIBzMp90dJJmyBE1R5O:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

___
*[`router/bfd/session.go` line 167 at r8](https://reviewable.io/reviews/scionproto/scion/4658#-OIB8UFJ7dk1LOq1drAt:-OIBz05NDYBS03w5t-W1:b-m1nt0i) ([raw file](https://github.com/scionproto/scion/blob/4f70f2e4f024552ceff555994b36303530bdc81e/router/bfd/session.go#L167)):*

I think that it is better to fix these things and leave the code here.
- NewSession: Done. Indeed.
- ifID not used: Gone. It was no-longer used before; I should have removed it anyway.
- dependency on controller (sure, that's normal, it provides the BFD configuration). Controller.go will need a bit of refactor in the next phase anyway, I might want to move the creation of bfd sessions down to underlay.go anyway, so we'll see to the config then, if we need to.
- the discriminator is no worse or better than before. It actually doesn't mater; we don't really use it. This is one thing that could stay in the router, given that it's the router that doesn't need it... but then, having a reasonable built-in is more convenient (and ensures protocol compliance).
- queue size: it is better hard-coded here than in the caller. It is a matter of implementation. The caller should certainly not know what to use.

I think that your concerns stem from the fact that bfd now looks more like a general purpose library than before, but thanks to Go's 1-level package hierarchy, it already was exported; just without a constructor. Doesn't mean that we need to support non-router usage. Given that it was already exposed to such non-router use, this is actually an improvement.
</blockquote></details>

All clear, let's try it like this then.

PR #4659, opened 2024-12-02
Update NAT Address Discovery Design Document 

jdslab:
Updated design document for NAT support as announced in #4630.

### DIFF
--- a/doc/dev/design/NAT-address-discovery.rst
+++ b/doc/dev/design/NAT-address-discovery.rst
@@ -2,8 +2,8 @@
 NAT IP/port discovery
 *********************
 
-- Author(s): Marc Frei, Tilmann Zäschke
-- Last updated: 2024-07-01
+- Author(s): Marc Frei, Jan Luan, Tilmann Zäschke
+- Last updated: 2024-11-25
 - Status: **WIP**
 - Discussion at: :issue:`4517`
 
@@ -15,8 +15,8 @@ This address may not be easy to discover if the sender is separated from the rec
 
 We want to propose a solution that allows SCION endhosts (and endhost libraries) to discover and use
 the address that is visible to the first hop border router as the source host address in outbound packets.
-The most elegant and most reliable solution appears to be to have
-the border router itself detect the NATed IP/port and report it to the client (to the sending endhost).
+The most elegant and most reliable solution appears to be to have the client (the sending endhost)
+detect its NATed IP/port by querying the border router for its publicly visible address.
 
 Background
 ==========
@@ -75,19 +75,76 @@ The implementation on the protocol level could be done in several ways:
 2.  Extend SCMP with a new INFO message
 
     -  Advantages: One less dependency on an external library and protocol
-    -  Disadvantages: More standardization effort? How do we solve authentication?
+    -  Disadvantages: More standardization effort?
 
-3.  Extend SCMP with a new ERROR message: "invalid source address for first hop pkt", similar to error 33.
-    The router can verify that for first hop packets, the IP src address (and L4 port if applicable) matches the SCION
-    src address (and L4 port).
-    If not, it returns an error. The actual source address would need to be attached somewhere, unless we decide
-    to change the payload so it contains the IP header of the offending packet (and the IP header should contain the
-    NATed IP/port).
+An implementation with STUN seems like the better solution, since it is an existing, well known protocol.
+Creating a new SCMP extension that provides essentially the same functionality as STUN seems redundant
+and doesn't provide any obvious advantages apart from the one mentioned above.
 
-    -  Advantages: One less dependency on an external library and protocol. Also one roundtrip less in case an endhost
-       doesn't sit behind a NAT or similar.
-    -  Disadvantages: Conceptually a bit of a hack. The BR would need to check every outbound packet as part of the fast
-       path. More standardization effort? How do we solve authentication?
+Regarding the STUN solution, there are multiple ways to send a STUN message over the wire:
+
+1. STUN/UDP/IP: The standard way of sending STUN packets as proposed in the IETF standard.
+
+   -  Easy to implement. However, the border router must distinguish between STUN packets and SCION packets.
+      This may be done using the magic cookie value, which is part of the STUN header.
+      The optional ``FINGERPRINT`` attribute of STUN may also be used as an aid for distinguishing.
+
+      -  Part of the magic cookie field overlaps with the SCION ``nexthdr`` field.
+         Reserving the value in that part of the magic cookie (33) would make the distinction unambiguous.
+         However, this value is assigned by IANA to "Datagram Congestion Control Protocol",
+         which might complicate standardization should we want to support this protocol over SCION in the future.
+      -  The ``FINGERPRINT`` attribute contains a CRC-32 checksum of the STUN packet, XOR'ed with the value 0x5354554e.
+         This attribute can be checked in addition to the basic check using the magic cookie.
+   -  Disadvantage: We cannot use SCION's Packet Authenticator Option
+      (`SPAO <https://docs.scion.org/en/latest/protocols/authenticator-option.html>`_) for message integrity.
+      If we want to have message integrity/authentication, we need to implement it separately.
+
+      -  The STUN standard provides an optional extension for username/password based authentication.
+         This authentication method is probably impractical to implement for our use case.
+         However, we might be able to "misuse" the ``MESSAGE-INTEGRITY`` STUN attribute of this mechanism,
+         which contains an HMAC of the STUN packet, for our own purpose.
+         We might be able to use DRKey to provide the shared secrets for computing the HMAC.
+
+2. STUN/SCION/UDP/IP: Carry the STUN packet (without UDP headers) inside a SCION packet.
+
+   -  Cleaner solution. We can assign a SCION ``nexthdr`` value to STUN to unambiguously distinguish STUN packets from
+      regular dataplane packets. (This is also how we handle BFD messages.)
+   -  Encapsulating STUN inside a SCION packet makes it possible to use SCION's built-in authentication functionality
+      (SPAO) for message integrity/authentication.
+   -  Conceptually awkward. STUN was designed as a transport layer payload (to be carried over UDP or TCP).
+      If SCION is viewed as a layer-3 protocol (same as IP), carrying STUN messages directly over SCION without
+      encapsulation in a transport layer header would be as if we carried STUN directly over IP without UDP or TCP.
+
+3. STUN/UDP/SCION/UDP/IP: Carry an entire STUN packet with UDP headers inside a SCION packet.
+
+   -  Difficult for BR to distinguish from normal dataplane packets.
+      The BR would need to look inside every UDP over SCION packet.
+   -  Conceptually unclear distinction from normal STUN/UDP messages carried over SCION.
+   -  Useless UDP header between SCION header and STUN packet.
+      The NAT would only rewrite the underlay UDP/IP headers. We would therefore only look at the underlay anyway.
+      The UDP header between SCION and STUN would only be confusing.
+   -  From a conceptual point of view, since the problem we are solving is not NAT detection for SCION,
+      but for the UDP/IP underlay, it is desirable to keep the solution on the layer of the underlay.
+      This way, STUN/UDP over SCION could be reserved for an eventual future use case to detect SCION NATs, if needed.
+
+Remark on message integrity/authentication:
+
+An attacker may spoof NAT address discovery (e.g. STUN) message replies to fool the client into assuming a wrong NAT'ed src address.
+This would cause returning traffic from subsequent communication by the client to be forwarded to the wrong destination.
+In the case of STUN, this attack is mitigated by a 96-bit TxID unique to each request.
+It is very unlikely that an attacker can guess the correct TxID at random,
+and thus send a spoofed STUN reply message that the client actually expects.
+However, an on-path attacker may still be able to modify STUN messages in transit (which have the correct TxID) to cause the same issue.
+This could be mitigated by some form of message integrity/authentication, as described above.
+On the other hand, it is to be noted that an attacker with such far-reaching abilities could also just intercept plain dataplane packets.
+Overall, the threat model is similar to the question about whether we need authentication for intra-AS SCMP messages.
+
+Decision
+--------
+During the open-source contributors meeting on Nov. 19, 2024, it was agreed that the STUN/UDP/IP solution is preferred
+due to its simplicity. However, arguments about message integrity/authentication have not yet been discussed at that time.
+It remains to be discussed whether these arguments pose enough reason to changed the preferred solution in favor of the
+STUN/SCION/UDP/IP variant.
 
 Rationale
 =========
@@ -121,6 +178,18 @@ Alternatives:
    -  This approach may be be problematic with sensitive NATs.
    -  We need to somehow standardize the STUN IP/port and/or communicate it to endhosts, e.g. via the topology.json file
       or the bootstrapping service.
+
+-  Extend SCMP with a new ERROR message: "invalid source address for first hop pkt", similar to error 33.
+   The router can verify that for first hop packets, the IP src address (and L4 port if applicable) matches the SCION
+   src address (and L4 port).
+   If not, it returns an error, with the actual source address attached somewhere, unless we decide
+   to change the payload so it contains the IP header of the offending packet (and the IP header should contain the
+   NATed IP/port).
+
+   -  Advantage: One roundtrip less in case an endhost doesn't sit behind a NAT or similar.
+   -  Disadvantages: Conceptually a bit of a hack. Complicated to implement.
+      The BR would need to check every outbound packet as part of the fast path.
+      The client would need to somehow buffer sent packets in case of errors to resend them with the correct src address.
 
 -  Remove all NATs and use IPv6 instead. This is technically possible but unlikely to happen anytime soon, especially
    because scarcity of IPv4 addresses is not the only reason why NATs are deployed.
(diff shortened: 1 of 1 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4659)

tzaeschke:
*[`doc/dev/design/NAT-address-discovery.rst` line 145 at r1](https://reviewable.io/reviews/scionproto/scion/4659#-ODBYPnOE5T58zZdgH5Y:-ODBYPnP3GeupntnD5TY:b-tumags) ([raw file](https://github.com/scionproto/scion/blob/219dd8553bbb86362c46f2bd82062715d88c3a9e/doc/dev/design/NAT-address-discovery.rst#L145)):*
> ```reStructuredText
> --------
> During the open-source contributors meeting on Nov. 19, 2024, it was agreed that the STUN/UDP/IP solution is preferred
> due to its simplicity. However, arguments about message integrity/authentication have not yet been discussed at that time.
> ```

As I remember it, the advantage wasn't simplicity (IP/UDP/SCION/STUN would also be simple) but a general viewpoint that STUN should operate on the underlay because it solves an underlay problem.

tzaeschke (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659#-:-ODBaLc25dOhyk3cm6pm:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @jdslab)

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659)* status: all files reviewed, 1 unresolved discussion (waiting on @jdslab, @jiceatscion, @JordiSubira, and @oncilla)

lukedirtwalker (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659#-:-ODPHUxV6s6UbHdTkCqu:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @jdslab, @jiceatscion, @JordiSubira, and @oncilla)

tzaeschke (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion, @JordiSubira, and @oncilla)

tzaeschke (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion, @JordiSubira, and @oncilla)

marcfrei (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jdslab, @jiceatscion, @JordiSubira, @lukedirtwalker, @oncilla, and @tzaeschke)

___
*[`doc/dev/design/NAT-address-discovery.rst` line 69 at r3](https://reviewable.io/reviews/scionproto/scion/4659#-ODuGsGu1MNY-wrv--9B:-ODuGsGv8wuOArxEcd-F:bw1kqbf) ([raw file](https://github.com/scionproto/scion/blob/1476173aeebc400502016b47da840d796feab544/doc/dev/design/NAT-address-discovery.rst#L69)):*
> ```reStructuredText
> 
> Also note that the NAT issue only exists for traffic between separate ASes. If both the client and the server are located within
> the same AS, the server can simply send returning packets to the underlay source address of the client (which is visible to the server in this case).
> ```

Add a note that this is unfortunately only possible for servers that are listening on ports which don't get routed through a SHIM dispatcher.



_Code quote:_
```reStructuredText
Also note that the NAT issue only exists for traffic between separate ASes. If both the client and the server are located within
the same AS, the server can simply send returning packets to the underlay source address of the client (which is visible to the server in this case).
```

tzaeschke:
*[`doc/dev/design/NAT-address-discovery.rst` line 244 at r3](https://reviewable.io/reviews/scionproto/scion/4659#-OE0QXNt52GnrALS8S6y:-OE0QXNt52GnrALS8S6z:b-w2b560) ([raw file](https://github.com/scionproto/scion/blob/1476173aeebc400502016b47da840d796feab544/doc/dev/design/NAT-address-discovery.rst#L244)):*
> ```reStructuredText
> - AS local traffic:
>   For communication within the same AS, the endhost should send returning packets to the underlay source address of the sender
>   simply by reversing src/dst addresses of the underlay in addition to reversing the src/dst addresses in the SCION header.
> ```

I would probably not mention any of the reversing, the reversing is just obvious and hasn't changed in any way. Instead, I would probably emphasize the "underlay":, e.g. "For communication...    to the underlay source address instead of the SCION source address."

tzaeschke:
*[`doc/dev/design/NAT-address-discovery.rst` line 237 at r3](https://reviewable.io/reviews/scionproto/scion/4659#-OE0RuGYE6lOXSKyuujD:-OE0RuGYE6lOXSKyuujE:bjp5ypc) ([raw file](https://github.com/scionproto/scion/blob/1476173aeebc400502016b47da840d796feab544/doc/dev/design/NAT-address-discovery.rst#L237)):*
> ```reStructuredText
> STUN/UDP/IP, STUN/SCION/UDP/IP, and SCMP message extension.
> It was agreed that a PR would be created for the STUN/UDP/IP variant.
> The STUN specification used is the newer specification (RFC5389).
> ```

Insert space: "RFC 5389"'? 
And possibly a link?

tzaeschke:
*[`doc/dev/design/NAT-address-discovery.rst` line 69 at r3](https://reviewable.io/reviews/scionproto/scion/4659#-ODuGsGu1MNY-wrv--9B:-OE0SGJm8V3zOfOnAaQH:b-rqfv8v) ([raw file](https://github.com/scionproto/scion/blob/1476173aeebc400502016b47da840d796feab544/doc/dev/design/NAT-address-discovery.rst#L69)):*
<details><summary><i>Previously, tzaeschke (Tilmann) wrote…</i></summary><blockquote>

By the time this goes into a release, there are hopefully no more dispatchers :-)
</blockquote></details>

I think the NAT issue also exists with intra-AS, it is just solved differently. I would rephrase it to something like: "For intra-AS traffic, the NAT problem can be solved be changing the implementation servers such that return packets are always sent to the underlay address instead (which may be a NATed address)of the SCION source address."

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659)* status: all files reviewed, 3 unresolved discussions (waiting on @jdslab, @jiceatscion, @JordiSubira, @marcfrei, and @oncilla)

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659)* status: all files reviewed, 3 unresolved discussions (waiting on @jdslab, @JordiSubira, @marcfrei, and @oncilla)

___
*[`doc/dev/design/NAT-address-discovery.rst` line 69 at r3](https://reviewable.io/reviews/scionproto/scion/4659#-ODuGsGu1MNY-wrv--9B:-OEURDVU8oUl_L33bY_u:bfe0ejj) ([raw file](https://github.com/scionproto/scion/blob/1476173aeebc400502016b47da840d796feab544/doc/dev/design/NAT-address-discovery.rst#L69)):*
<details><summary><i>Previously, tzaeschke (Tilmann) wrote…</i></summary><blockquote>

I think the NAT issue also exists with intra-AS, it is just solved differently. I would rephrase it to something like: "For intra-AS traffic, the NAT problem can be solved be changing the implementation servers such that return packets are always sent to the underlay address instead (which may be a NATed address)of the SCION source address."
</blockquote></details>

I agree that the NAT issue does exist just as well intra-AS: Sunrise has SCION, but I doubt it does me much good with the non-routable address they give me. However, can't we assume that the exact same STUN tactics can be used intra-AS? That is: use the nearest BR as a stun server - on the inside? I know you'll tell me that if the client is behind a symmetric NAT, then it won't work. But that's fine, because if an AS is configured like that internally, then it just violates the SCION protocol entirely: the underlay address is supposed to be a valid endhost address; not some random thing. Am I missing something? I am not too fond of the idea that we can skirt the issue by reversing the underlay header; that assumes that in all cases the underlay header remains associated with the rest of the SCION packet until such time that a response is warranted. It also excludes any kind of NAT-ed  SCION server, which sucks exactly as much as it does in the IP world.

jiceatscion (commented):
My objection to the intra-AS solution shouldn't be viewed as blocking. Intra-AS NAT support can be introduced later and need not be tied to the inter-As solution.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659#-:-OEUXjPyB_VzqZbiqlym:b-nc43bc)* status: all files reviewed, 3 unresolved discussions (waiting on @jdslab, @JordiSubira, @marcfrei, and @oncilla)

jiceatscion (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659#-:-OEUYBDgERiG2K6ZZUJr:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @jdslab, @JordiSubira, @marcfrei, and @oncilla)

tzaeschke:
> > I think the NAT issue also exists with intra-AS, it is just solved differently. I would rephrase it to something like: "For intra-AS traffic, the NAT problem can be solved be changing the implementation servers such that return packets are always sent to the underlay address instead (which may be a NATed address)of the SCION source address."
> 
> I agree that the NAT issue does exist just as well intra-AS: Sunrise has SCION, but I doubt it does me much good with the non-routable address they give me. However, can't we assume that the exact same STUN tactics can be used intra-AS? That is: use the nearest BR as a stun server - on the inside? I know you'll tell me that if the client is behind a symmetric NAT, then it won't work. But that's fine, because if an AS is configured like that internally, then it just violates the SCION protocol entirely: the underlay address is supposed to be a valid endhost address; not some random thing. Am I missing something? I am not too fond of the idea that we can skirt the issue by reversing the underlay header; that assumes that in all cases the underlay header remains associated with the rest of the SCION packet until such time that a response is warranted. It also excludes any kind of NAT-ed SCION server, which sucks exactly as much as it does in the IP world.

Symmetric NAT may be one issue, but another one is that an AS may be divided into multiple subnets separated by NATs. That means, different border routers may be seeing different NATed addresses. The problem is then that we need to figure out which border router is in the same subnet as the destination host, and there may not be any border router in that subnet. 

The only way I see to avoid using the underlay is having endhosts implement STUN servers, so before I send anything to a destination host, I first STUN them to get my NATed address. Compared to that, it seems simpler to me to use the underlay, but we can discuss this if people disagree?

jiceatscion (commented):
Yes the possibility of adding stun server to every endpoint crossed my mind. Sounds brutal, but might not be that bad. It'd be basically the same code, just enabled in more places. Yet, I have a hard time imagining what internal network topology would make that necessary. May be it'd be worth showing an example? As I said, we don't really need to solve it right now. It could be a subsequent effort... even if it leads to generalizing the stun server to every end-point, it would still reuse the stun server code from the inter-AS case, so not much would be wasted.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659#-:-OEYh5066-YUVzsML5f5:b-a6dd0u)* status: all files reviewed, 3 unresolved discussions (waiting on @jdslab, @JordiSubira, @marcfrei, and @oncilla)

tzaeschke:
@jiceatscion 
Okay, so:
1) Dragging along the underlay address: In JPAN, I currently do the following, when I receive a packet, I attach the inverted path to it. If the packet comes from intra-AS, the path is [], so I store the underlay address in the firstHop field. I don't need an extra attribute. When sending an answer, I simply send to the firstHop underlay, no extra logic required (except for ensuring the correct port range).
2) I would actually solve intra-AS and inter-AS together, if possible. It would be weird if one works and the other one doesn't. What do others think?
3) An example for an AS with multiple subnets is the Geant AS, for example Geant Paris and Geant Frankfurt are in the same AS but in different subnets (thanks to @FR4NK-W for providing the example; and yes there isn't really a NAT inbetween them, as I wrote), that may have caused confusion).

Does that make sense?
Should we solve intra/inter-AS separately?

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659)* status: all files reviewed, 3 unresolved discussions (waiting on @jdslab, @JordiSubira, @marcfrei, and @oncilla)

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659)* status: all files reviewed, 3 unresolved discussions (waiting on @jdslab, @marcfrei, and @oncilla)

___
*[`doc/dev/design/NAT-address-discovery.rst` line 69 at r3](https://reviewable.io/reviews/scionproto/scion/4659#-ODuGsGu1MNY-wrv--9B:-OFYozxVDWgKmfFUJIB5:bsixlgr) ([raw file](https://github.com/scionproto/scion/blob/1476173aeebc400502016b47da840d796feab544/doc/dev/design/NAT-address-discovery.rst#L69)):*

I agree that the NAT issue does exist just as well intra-AS: Sunrise has SCION, but I doubt it does me much good with the non-routable address they give me. However, can't we assume that the exact same STUN tactics can be used intra-AS? That is: use the nearest BR as a stun server - on the inside? I know you'll tell me that if the client is behind a symmetric NAT, then it won't work. But that's fine, because if an AS is configured like that internally, then it just violates the SCION protocol entirely: the underlay address is supposed to be a valid endhost address; not some random thing. Am I missing something? I am not too fond of the idea that we can skirt the issue by reversing the underlay header; that assumes that in all cases the underlay header remains associated with the rest of the SCION packet until such time that a response is warranted. It also excludes any kind of NAT-ed  SCION server, which sucks exactly as much as it does in the IP world.
</blockquote></details>

I would also add that on the long term, we cannot assume how packets will be dispatched to end hosts. Mainly, if we ever have kernel support, packets may get encapsulated again to a fix port, so that they can be routed internally by intra-AS middleware. In that sense, ideally, this solution should work also in this case.

tzaeschke:
*[`doc/dev/design/NAT-address-discovery.rst` line 69 at r3](https://reviewable.io/reviews/scionproto/scion/4659#-ODuGsGu1MNY-wrv--9B:-OFuviNJ6qUP_xC40M1a:byjuw0r) ([raw file](https://github.com/scionproto/scion/blob/1476173aeebc400502016b47da840d796feab544/doc/dev/design/NAT-address-discovery.rst#L69)):*
> use the nearest BR as a stun server

The problem really is, which is the "nearest" STUN server? And in which subnet is it?
Copied from my comment just to keep thing in order (I should have put my comment here in the first place):
An example for an AS with multiple subnets is the Geant AS, for example Geant Paris and Geant Frankfurt are in the 
same AS but in different subnets (thanks to [@FR4NK-W](https://github.com/FR4NK-W) for providing the example; and yes there isn't really a NAT inbetween them, as I wrote, that may have caused confusion).

tzaeschke (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4659)* status: all files reviewed, 3 unresolved discussions (waiting on @FR4NK-W, @jdslab, and @oncilla)

tzaeschke:
@lukedirtwalker @oncilla @jiceatscion @JordiSubira @marcfrei @FR4NK-W 

Apologies for opening this up again but we came up with a new idea that we may want to discuss.

The idea is to let border routers rewrite the source address/port (replace it with the underlay src address/port) on packets on the way out. Previously this idea was dismissed because it would not work with SPAO where the SCION SRC/DST addresses are protected fields. If we were to change SPAO to allow this kind of rewriting of SRC addresses and SRC ports, the NAT solution could became a whole lot simpler.
- We would not need STUN. THis would remove quite some complexity from clients (keep a table of mappings and handle keepalive etc).
- This would do away with any STUN roudtrips required to detect the external NATed address prior to connection to e border router.
- This would allow easy implementation of SCION-aware NATs that could simply rewrite the SRC address/port.

SPAO may anyway need some changes because it is unclear how it works with DRKeys, specifically how clients behind a NAT can get distinct DRKeys instead of sharing one key (unless sharing is okay).

We think it may be worth discussing this in more detail, maybe in the upcoming open-source meeting?

Comments?

jiceatscion:
> Apologies for opening this up again but we came up with a new idea that we may want to discuss.

No need. Better take the time to arrive at the best possible solution.

> The idea is to let border routers rewrite the source address/port (replace it with the underlay src address/port) on packets on the way out.

That sounds both simple and disturbing. In fact, I found it disturbing for a while that the effect of NAT leaks out of the underlay. Trying to think about it from an abstract networking stack perspective, I wonder if the layering violation isn't already implicit in the SCION forwarding code. We just assume that the link-layer address (since that's what the underlay address is from the SCION network perspective) is contained verbatim in the network address. I guess avoiding any address resolution protocol is a worthy goal, but may be we should accept that there needs to be some form of resolution. In the router, the forwarding to a local host is named "ResolveLocalDst()". so, I'm not the first one to think that some non-trivial "resolution" could be in order. For example we could keep a map that we populate from packets received from local sources. The map would not need to store the common case identity mapping. I can take my answer from next Tuesday's discussion.

> We think it may be worth discussing this in more detail, maybe in the upcoming open-source meeting?

Agree. Consider it added to the agenda.

shitz:
From our perspective, rewriting the SCION host address is not desirable. Having a stable address is hugely beneficial for multiple reasons:

1. It allows for source authentication. As you already mentioned, SPAO and DRKey rely on stable source addresses. Introducing a different stable identifier for this purpose introduces all kinds of complexity.
2. At Anapaya, we are currently building a system that enables SCION ASes to authorize endpoints in their AS to use the local SCION infrastructure. Furthermore, this system also does usage accounting to enable monetization of "SCION subscribers". We believe that such an infrastructure is essential for the widespread adoption and deployment of SCION. Commercial ISPs will only deploy SCION if they can monetize it and today this is only possible on the level of SCION links. This system also relies on relatively stable SCION source addresses otherwise it would need to keep much more state for accounting purposes. Specifically we intend that clients resolve the underlay address with STUN and then register themselves at the router (or SCION network access point - SNAP) with a SCION address and the resolved underlay address.
3. NAT has been a source of many issues for all kind of end-to-end properties in the past. If we can avoid poluting the SCION address space with IP NATs I think we should do so even if it comes at the cost of some complexity on the underlay.

For those reasons, we are still advocating for the STUN-based approach.

N.B., once it is ready, we will also create a proposal for this SCION endhost API.

PR #4663, opened 2024-12-12
app/path: fix path selection for Windows

LencigGaer:
Selecting a path when using the ```-i``` option doesn't work in Windows.

Windows uses ```"\r\n"``` as delimiter, the current code only removes the last character from the user input to select a path (i.e. ```"\n"```).

This PR instead right-trims all appearances of ```\n``` and ```\r``` from the input to make it work on both, Windows and Unix systems.


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4663)

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4663)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @LencigGaer)

PR #4665, opened 2024-12-14
support BGP ASNs in local topologies

lschulz:
Local topologies (created using the scion.sh topology command) do not support ASN < 2^32. If such ASNs are written in decimal notation, validation of the ISD-ASN fails. Writing the ASN in hexadecimal also fails, because the crytographic material is generated in the wrong directory.

All that seems to be missing to support BGP ASNs is allowing the decimal format and making sure that all ASNs < 2^32 are converted to decimal format as expected by the scion-pki tool.

### DIFF
--- a/tools/topology/config.py
+++ b/tools/topology/config.py
@@ -95,6 +95,7 @@ class ConfigGenerator(object):
         Generate all needed files.
         """
         self._ensure_uniq_ases()
+        self._canonicalize_isd_asns()
         topo_dicts, self.all_networks = self._generate_topology()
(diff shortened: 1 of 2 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4665)

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4665)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @lschulz)

PR #4671, opened 2024-12-19
V6 address design

jiceatscion:
This has been hanging around an old branch of mine forever. Since we're not doing it soon but no-one told me it was dumb, I'd rather have it in the main tree; linked the "postponed" category.

### DIFF
--- a/doc/dev/design/V6AddressSupport.rst
+++ b/doc/dev/design/V6AddressSupport.rst
@@ -1,11 +1,11 @@
-*******************************
-[V6 compatible SCION addresses]
-*******************************
+*****************************
+V6 compatible SCION addresses
+*****************************
 
 :Author(s): J-C HUgly
 :Last updated: 2023-08-31
 :Status: **proposal**
-:Discussion: `Discussions/4324_`
+:Discussion: `Discussions/4324`_
 
 
 .. _`Discussions/4324`: https://github.com/scionproto/scion/discussions/4324
@@ -21,15 +21,22 @@ subset of those suggestions in real code.
 This includes all relevant aspects:
 
 * Validation/Identification:
+
   * "Is this V6 address a valid SCION address?"
   * "Is this scion address representable as V6?"
+
 * Structured conversions:
+
   * "Turn this V6 address into a SCION address if you can."
   * "Turn this SCION address into a V6 address if you can."
+
 * String representation:
+
   * "Construct a SCION address from this V6 address string."
   * "Construct a SCION address from this SCION-style address string if you can."
+
 * Viable migration path and future growth support:
+
   * Configurable address structure with support for the existing V6-incompatible
     structure.
   * Support of existing addresses that are convertible to the new structure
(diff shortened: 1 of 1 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4671)

doc/dev/design/V6-address-support.rst
```diff
@@ -0,0 +1,373 @@
...
+
+Instead, what we would like is that applications can use SCION without even knowing
+it, simply as a result of:
+
+* Parsing a normal-looking textual V6 address from a config file or URL.
+* Getting a normal AAAA result from a DNS query.
+* Opening a UDP/TCP connection to said address via the regular socket API. This would imply
+  the existence of SCION-aware code below the socket API that:
+
+  * Recognizes scion addresses and handles the traffic accordingly.
```
romshark:
```suggestion
  * Recognizes SCION addresses and handles the traffic accordingly.
```

doc/dev/design/V6-address-support.rst
```diff
@@ -0,0 +1,373 @@
...
+However, the SCION/IPV6 address encoding is predicated on the fact that the the full IPV6
+address of the destination host is identical to its SCION address; this means that:
+
+* When the new schema is in use, a destination address supplied by the application
+  is partially redundantly encoded as the HOST portion of the wire format SCION address.
+* That address is a real IPV6 address, reachable by the destination border router.
+
+Until we reach the point where scion routing is part of the normal network stack (and therefore
+scion addresses need to be distinguished from other V6 addresses) we can get away with using
+arbitrary IPV6 addresses in the unique local range. After that, we need to start using real
```
romshark:
```suggestion
Until we reach the point where SCION routing is part of the normal network stack (and therefore
SCION addresses need to be distinguished from other V6 addresses) we can get away with using
arbitrary IPV6 addresses in the unique local range. After that, we need to start using real
```

doc/dev/design/V6-address-support.rst
```diff
@@ -0,0 +1,373 @@
...
+:code:`sendto(addr, stuff, [,routespec])`, where :code:`addr` does not need to include
+the next hop and where routespec is entirely optional.
+
+Once we have that, we can start enabling the use of V6 addresses as the addr argument.
+
+The netsec tree also offers a number of scion-enabled apps. Those rely on "pan", a custom
+API-like layer. It would be good to try and converge pan and the new API contemplated
+above, so that eventually, these scion-enabled apps are able to use plain IPV6 addresses
+too.
+
```
romshark:
```suggestion
The netsec tree also offers a number of SCION-enabled apps. Those rely on "pan", a custom
API-like layer. It would be good to try and converge pan and the new API contemplated
above, so that eventually, these SCION-enabled apps are able to use plain IPV6 addresses
too.

```

romshark: same here, just replace `scion`->`SCION`

doc/dev/design/V6-address-support.rst
```diff
@@ -0,0 +1,373 @@
...
+* Do Nothing:
+
+  * SCION can only be used via especially crafted applications or the SCION gateway.
+  * The gateway is inherently limitted by the IP-SCION address mapping. If all of SCION
+    has to be used through it, then it is pointless.
+  * Are we planning on providing replacement for all the applications and libraries
+    using internet today? Is someone else?
+
+* Map IPV6 addresses to scion ISD-AS downstream from the application:
```
romshark:
```suggestion
* Map IPV6 addresses to SCION ISD-AS downstream from the application:
```

PR #4683, opened 2025-01-23
dispatcher: debug log unroutable SCMP error packets

lukedirtwalker:
Instead of logging at error level only log unroutbale SCMP error packets on Debug level. For some of the usecases there is simply not much we can do, so we shouldn't care too much and fload the error log.

### DIFF
--- a/dispatcher/dispatcher.go
+++ b/dispatcher/dispatcher.go
@@ -197,17 +197,32 @@ func (s *Server) processMsgNextHop(
 	switch s.decoded[len(s.decoded)-1] {
 	case slayers.LayerTypeSCMP:
 		// send response to BR
-		if s.scmpLayer.TypeCode.Type() == slayers.SCMPTypeTracerouteRequest ||
-			s.scmpLayer.TypeCode.Type() == slayers.SCMPTypeEchoRequest {
+		switch s.scmpLayer.TypeCode.Type() {
+		case slayers.SCMPTypeEchoRequest, slayers.SCMPTypeTracerouteRequest:
 			dstAddrPort = prevHop
-		} else { // relay to end application
-			dstAddrPort, err = s.getDstSCMP()
+		case slayers.SCMPTypeEchoReply, slayers.SCMPTypeTracerouteReply:
+			dstAddrPort, err = s.getSCMPInfoReplyDst()
 			if err != nil {
-				log.Error("Getting destination for SCMP message", "err", err)
+				log.Error("Getting destination for SCMPInfo reply", "err", err)
 				return nil, netip.AddrPort{}, nil
 			}
 			if dstAddrPort.Addr().Unmap().Compare(underlay.Unmap()) != 0 {
 				log.Error("UDP/IP addr destination different from UDP/SCION addr",
+					"UDP/IP:", underlay.Unmap().String(),
+					"UDP/SCION:", dstAddrPort.Addr().Unmap().String())
+				return nil, netip.AddrPort{}, nil
+			}
+		default:
+			dstAddrPort, err = s.getSCMPErrorDst()
+			if err != nil {
+				log.Debug("Getting destination for SCMP message",
+					"underlay", underlay,
+					"prev_hop", prevHop,
+					"err", err)
+				return nil, netip.AddrPort{}, nil
+			}
+			if dstAddrPort.Addr().Unmap().Compare(underlay.Unmap()) != 0 {
+				log.Debug("UDP/IP addr destination different from UDP/SCION addr",
 					"UDP/IP:", underlay.Unmap().String(),
 					"UDP/SCION:", dstAddrPort.Addr().Unmap().String())
 				return nil, netip.AddrPort{}, nil
@@ -326,7 +341,7 @@ func (s *Server) reverseSCION() error {
 	return nil
 }
 
-func (s *Server) getDstSCMP() (netip.AddrPort, error) {
+func (s *Server) getSCMPInfoReplyDst() (netip.AddrPort, error) {
 	// Check if its SCMPEcho or SCMPTraceroute reply
 	if s.scmpLayer.TypeCode.Type() == slayers.SCMPTypeEchoReply {
 		var scmpEcho slayers.SCMPEcho
@@ -344,18 +359,23 @@ func (s *Server) getDstSCMP() (netip.AddrPort, error) {
 		}
 		return addrPortFromBytes(s.scionLayer.RawDstAddr, scmpTraceroute.Identifier)
 	}
+	return netip.AddrPort{}, serrors.New("invalid code path")
+}
 
+func (s *Server) getSCMPErrorDst() (netip.AddrPort, error) {
+	errorType := s.scmpLayer.TypeCode.Type()
 	// Drop unknown SCMP error messages.
 	if s.scmpLayer.NextLayerType() == gopacket.LayerTypePayload {
 		return netip.AddrPort{}, serrors.New("unsupported SCMP error message",
-			"type", s.scmpLayer.TypeCode.Type())
+			"error_type", s.scmpLayer.TypeCode.Type())
 	}
 	l, err := decodeSCMP(&s.scmpLayer)
 	if err != nil {
 		return netip.AddrPort{}, err
 	}
 	if len(l) != 2 {
-		return netip.AddrPort{}, serrors.New("SCMP error message without payload")
+		return netip.AddrPort{},
+			serrors.New("SCMP error message without payload", "error_type", errorType)
 	}
 	gpkt := gopacket.NewPacket(*l[1].(*gopacket.Payload), slayers.LayerTypeSCION,
 		gopacket.DecodeOptions{
@@ -363,6 +383,12 @@ func (s *Server) getDstSCMP() (netip.AddrPort, error) {
 		},
 	)
 
+	// For debugging print 200 bytes of the quoted payload.
+	var payloadForDebug []byte
+	if log.Root().Enabled(log.DebugLevel) {
+		payloadForDebug = l[1].(*gopacket.Payload).LayerContents()
+		payloadForDebug = payloadForDebug[:min(len(payloadForDebug), 200)]
+	}
 	// If the offending packet was UDP/SCION, use the source port to deliver.
 	if udp := gpkt.Layer(slayers.LayerTypeSCIONUDP); udp != nil {
 		port := udp.(*slayers.UDP).SrcPort
@@ -371,7 +397,9 @@ func (s *Server) getDstSCMP() (netip.AddrPort, error) {
 		// they set the source port to 0. But there is no harm, since those
 		// packets are destined to be dropped anyway.
 		if port == 0 {
-			return netip.AddrPort{}, serrors.New("SCMP error with truncated UDP header")
+			return netip.AddrPort{},
+				serrors.New("SCMP error with truncated UDP header",
+					"error_type", errorType, "payload", payloadForDebug)
 		}
 		return addrPortFromBytes(s.scionLayer.RawDstAddr, port)
 	}
@@ -385,12 +413,14 @@ func (s *Server) getDstSCMP() (netip.AddrPort, error) {
 		if !tc.InfoMsg() {
 			return netip.AddrPort{},
 				serrors.New("SCMP error message in response to SCMP error message",
-					"type", tc.Type())
+					"error_type", errorType, "quote_type", tc.Type(), "payload", payloadForDebug)
 		}
 		// We only support echo and traceroute requests.
 		t := tc.Type()
 		if t != slayers.SCMPTypeEchoRequest && t != slayers.SCMPTypeTracerouteRequest {
-			return netip.AddrPort{}, serrors.New("unsupported SCMP info message", "type", t)
+			return netip.AddrPort{},
+				serrors.New("SCMP error contains unsupported SCMP info message",
+					"error_type", errorType, "quote_type", t, "payload", payloadForDebug)
 		}
 
 		var port uint16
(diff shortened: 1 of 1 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4683)

jiceatscion (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4683)* status: all files reviewed, 3 unresolved discussions (waiting on @JordiSubira and @lukedirtwalker)

___
*[`dispatcher/dispatcher.go` line 200 at r1](https://reviewable.io/reviews/scionproto/scion/4683#-OHHd0HwCSE7Y_cgDvJk:-OHHd0HwCSE7Y_cgDvJl:b-2pt2yp) ([raw file](https://github.com/scionproto/scion/blob/e0ce875c646d242f2ae3f3be1d5696aa5a7bb826/dispatcher/dispatcher.go#L200)):*
> ```Go
> 	case slayers.LayerTypeSCMP:
> 		// send response to BR
> 		switch s.scmpLayer.TypeCode.Type() {
> ```

We get the typecode twice, in 3 places. Maybe it's worth storing it in a var and pass it to the getxxxDst() functions?
___
*[`dispatcher/dispatcher.go` line 210 at r1](https://reviewable.io/reviews/scionproto/scion/4683#-OHHbCp2E5WhTsVfmgne:-OHHbCp2E5WhTsVfmgnf:bkzptc5) ([raw file](https://github.com/scionproto/scion/blob/e0ce875c646d242f2ae3f3be1d5696aa5a7bb826/dispatcher/dispatcher.go#L210)):*
> ```Go
> 			}
> 			if dstAddrPort.Addr().Unmap().Compare(underlay.Unmap()) != 0 {
> 				log.Error("UDP/IP addr destination different from UDP/SCION addr",
> ```

Is the explicit String() necessary here? The log functions do the string conversion themselves if/when logging happens, no? Also, I don't think the ":" at the end of the keys is redundant, since the logging functions do their own "key:value" formatting.
___
*[`dispatcher/dispatcher.go` line 225 at r1](https://reviewable.io/reviews/scionproto/scion/4683#-OHHc-wL7sM4Ytw3cDoh:-OHHc-wL7sM4Ytw3cDoi:b-7ropln) ([raw file](https://github.com/scionproto/scion/blob/e0ce875c646d242f2ae3f3be1d5696aa5a7bb826/dispatcher/dispatcher.go#L225)):*
> ```Go
> 			}
> 			if dstAddrPort.Addr().Unmap().Compare(underlay.Unmap()) != 0 {
> 				log.Debug("UDP/IP addr destination different from UDP/SCION addr",
> ```

ditto

jiceatscion (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4683#-:-OHHmnhRCs9QC67wF5Ll:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JordiSubira)

JordiSubira (changes requested):
I agree some of the log levels on this file should be reconsidered. Probably there are more than the ones we are discussing here (see inline comments)

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4683#-:-OHMWKrJEfbIsDb7bG4L:b-gwx8i3)* status: all files reviewed, 4 unresolved discussions (waiting on @lukedirtwalker)

___
*[`dispatcher/dispatcher.go` line 210 at r2](https://reviewable.io/reviews/scionproto/scion/4683#-OHMSAb6-8QM3j806WrR:-OHMSAb6-8QM3j806WrS:brx1jo) ([raw file](https://github.com/scionproto/scion/blob/0e8be83168fb43c2726be63a156820cca59ace1d/dispatcher/dispatcher.go#L210)):*
> ```Go
> 			}
> 			if dstAddrPort.Addr().Unmap().Compare(underlay.Unmap()) != 0 {
> 				log.Error("UDP/IP addr destination different from UDP/SCION addr",
> ```

This error is equivalent to the one in the default case, so IMO it should be equally treated.
___
*[`dispatcher/dispatcher.go` line 225 at r2](https://reviewable.io/reviews/scionproto/scion/4683#-OHMSOLuEdXpf_f3GsnR:-OHMSOLuEdXpf_f3GsnS:by3hsz9) ([raw file](https://github.com/scionproto/scion/blob/0e8be83168fb43c2726be63a156820cca59ace1d/dispatcher/dispatcher.go#L225)):*
> ```Go
> 			}
> 			if dstAddrPort.Addr().Unmap().Compare(underlay.Unmap()) != 0 {
> 				log.Debug("UDP/IP addr destination different from UDP/SCION addr",
> ```

Indeed this shouldn't be `log.Error` because this error may be caused by a misbehaving/malicious sender. I am wondering actually whether "Debug" is the right level here. For instance, some users may want to log this "warnings" about potential misbehaving senders. We don't have a "Warn" level, perhaps this should be desirable for this type of errors? We may also investigate this in a separate PR.
___
*[`dispatcher/dispatcher.go` line 235 at r2](https://reviewable.io/reviews/scionproto/scion/4683#-OHMW2_w7s2VdGwKqo-h:-OHMW2_w7s2VdGwKqo-i:b-t3gpm8) ([raw file](https://github.com/scionproto/scion/blob/0e8be83168fb43c2726be63a156820cca59ace1d/dispatcher/dispatcher.go#L235)):*
> ```Go
> 		dstAddrPort, err = s.getDstSCIONUDP()
> 		if err != nil {
> 			log.Error("Getting destination for SCION/UDP message", "err", err)
> ```

We may also want to change this log level
___
*[`dispatcher/dispatcher.go` line 239 at r2](https://reviewable.io/reviews/scionproto/scion/4683#-OHMUlsj75e-NlO2cstd:-OHMUlsj75e-NlO2cste:b-pxhtjl) ([raw file](https://github.com/scionproto/scion/blob/0e8be83168fb43c2726be63a156820cca59ace1d/dispatcher/dispatcher.go#L239)):*
> ```Go
> 		}
> 		if dstAddrPort.Addr().Unmap().Compare(underlay.Unmap()) != 0 {
> 			log.Error("UDP/IP addr destination different from UDP/SCION addr",
> ```

This error is equivalent to the one in the "LayerTypeSCMP" case, so IMO it should be equally treated.

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4683)* status: all files reviewed, 4 unresolved discussions (waiting on @JordiSubira and @lukedirtwalker)

___
*[`dispatcher/dispatcher.go` line 225 at r2](https://reviewable.io/reviews/scionproto/scion/4683#-OHMSOLuEdXpf_f3GsnR:-OHMglxy3ns1nDzLsjs6:ba68uoi) ([raw file](https://github.com/scionproto/scion/blob/0e8be83168fb43c2726be63a156820cca59ace1d/dispatcher/dispatcher.go#L225)):*

Indeed this shouldn't be `log.Error` because this error may be caused by a misbehaving/malicious sender. I am wondering actually whether "Debug" is the right level here. For instance, some users may want to log this "warnings" about potential misbehaving senders. We don't have a "Warn" level, perhaps this should be desirable for this type of errors? We may also investigate this in a separate PR.
</blockquote></details>

The problem with logging it under normal operations is that it exposes the system to being DOS'd by deliberately triggering the costly logging code. It might be more appropriate to have metrics for potentially malicious behavior. That would be less expensive than logging, would it not?

JordiSubira (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4683)* status: all files reviewed, 4 unresolved discussions (waiting on @jiceatscion and @lukedirtwalker)

___
*[`dispatcher/dispatcher.go` line 225 at r2](https://reviewable.io/reviews/scionproto/scion/4683#-OHMSOLuEdXpf_f3GsnR:-OHMhrW20RUikwrG_LSq:b-2elwv3) ([raw file](https://github.com/scionproto/scion/blob/0e8be83168fb43c2726be63a156820cca59ace1d/dispatcher/dispatcher.go#L225)):*

The problem with logging it under normal operations is that it exposes the system to being DOS'd by deliberately triggering the costly logging code. It might be more appropriate to have metrics for potentially malicious behavior. That would be less expensive than logging, would it not?
</blockquote></details>

Well, several software solutions log this type of events (for example, apache-http access logs) and they are consumed by other backend services such as splunk log ingestor. tbh, I haven't investigated how good the log implementation we are using is, for instance deferring costly operations and so. Of course, we can have alternatives to logging, but we don't have one right now for this type of events.

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4683)* status: all files reviewed, 4 unresolved discussions (waiting on @JordiSubira and @lukedirtwalker)

___
*[`dispatcher/dispatcher.go` line 225 at r2](https://reviewable.io/reviews/scionproto/scion/4683#-OHMSOLuEdXpf_f3GsnR:-OHMm9_C7Sy1EVM9FIGa:b-ygkkpf) ([raw file](https://github.com/scionproto/scion/blob/0e8be83168fb43c2726be63a156820cca59ace1d/dispatcher/dispatcher.go#L225)):*

Well, several software solutions log this type of events (for example, apache-http access logs) and they are consumed by other backend services such as splunk log ingestor. tbh, I haven't investigated how good the log implementation we are using is, for instance deferring costly operations and so. Of course, we can have alternatives to logging, but we don't have one right now for this type of events.
</blockquote></details>

So, definitely more conversation required. Let's not block this PR over that. File an issue and leave the log level here as it is for now?

dispatcher/dispatcher.go
```diff
@@ -212,6 +212,22 @@ func (s *Server) processMsgNextHop(
 					"UDP/SCION:", dstAddrPort.Addr().Unmap().String())
 				return nil, netip.AddrPort{}, nil
 			}
+		default:
+			dstAddrPort, err = s.getSCMPErrorDst(t)
+			if err != nil {
+				log.Debug("Getting destination for SCMP message",
```
romshark, line 218: I'd like to point out a small issue of our current `Debug` function: variadic params.
Ideally, we would like debug logs to be 1 branch miss-prediction in the worst case when debug logs are disabled. Variadic parameters are just syntax sugar for slices. Which means that you still pay 1 allocation for it even when debug logs are disabled:

```go
go test -run nothing -bench BenchmarkWrapper -benchmem
goos: darwin
goarch: arm64
pkg: github.com/scionproto/scion/pkg/log
BenchmarkWrapper-14     72618313                16.30 ns/op           16 B/op          1 allocs/op
PASS
ok      github.com/scionproto/scion/pkg/log     2.134s
```

```
rsc@Mac log % go test -run nothing -bench BenchmarkWrapper -benchmem
goos: darwin
goarch: arm64
pkg: github.com/scionproto/scion/pkg/log
BenchmarkWrapper-14     72618313                16.30 ns/op           16 B/op          1 allocs/op
PASS
ok      github.com/scionproto/scion/pkg/log     2.134s
```

For reference, just calling `Debug("SCMP response")`:

```
rsc@Mac log % go test -run nothing -bench BenchmarkWrapper -benchmem
goos: darwin
goarch: arm64
pkg: github.com/scionproto/scion/pkg/log
BenchmarkWrapper-14     329259847                3.611 ns/op           0 B/op          0 allocs/op
PASS
ok      github.com/scionproto/scion/pkg/log     1.800s
```

In this particular case this might or might not be an issue. I'm seeing we're also using debug logs on the data plane frequently (a hot-path AFAIK).

P.S.
Sorry if this remark is outside the scope of the PR.

romshark:
@JordiSubira
> Roman's observation about the efficiency of our wrapper calls around the log (I saw the comment even it doesn't show up on the discussion, don't know why...)

I made a stupid mistake while measuring and measured it wrong :)

Hence removed the comment to avoid misleading people.

the Go compiler seems to be able to optimize most of the `Debug` overhead away _somehow_ (no variadic params slice allocation when debug logs are disabled and even if `Debug` is `//go:noinline`'ed)

I won't dig deeper since I think I would first want to gather profiles from the router before trying to optimize anything by _assuming_ cost without having the numbers

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4683)* status: all files reviewed, 4 unresolved discussions (waiting on @JordiSubira and @lukedirtwalker)

___
*[`dispatcher/dispatcher.go` line 225 at r2](https://reviewable.io/reviews/scionproto/scion/4683#-OHMSOLuEdXpf_f3GsnR:-OHOTUld3ewavUy_XpSC:b-mf9pd5) ([raw file](https://github.com/scionproto/scion/blob/0e8be83168fb43c2726be63a156820cca59ace1d/dispatcher/dispatcher.go#L225)):*

Sure, we can move this discussion somewhere else and perhaps Roman's observation about the efficiency of our wrapper calls around the log (I saw the comment even it doesn't show up on the discussion, don't know why...)
</blockquote></details>

Filed #4685 for that purpose.

dispatcher/dispatcher.go
```diff
@@ -326,43 +343,54 @@ func (s *Server) reverseSCION() error {
...
-	if s.scmpLayer.TypeCode.Type() == slayers.SCMPTypeTracerouteReply {
+	case slayers.SCMPTypeTracerouteReply:
 		var scmpTraceroute slayers.SCMPTraceroute
 		err := scmpTraceroute.DecodeFromBytes(s.scmpLayer.Payload, gopacket.NilDecodeFeedback)
 		if err != nil {
 			return netip.AddrPort{}, err
 		}
 		return addrPortFromBytes(s.scionLayer.RawDstAddr, scmpTraceroute.Identifier)
+	default:
+		return netip.AddrPort{}, serrors.New("invalid code path")
```
romshark, line 364: Could this be turned into a sentinel error?

JordiSubira: Yes it could; or directly "panic()" since we are doing something wrong (buggy) if we fall in the default case within this method...

jiceathome: "Invalid code path" sounds like a panic-able offense to me too. On the other hand, if there's even a remote possibility that this happens due to external causes, and given that it is completely recoverable, keeping the component running would be the nice thing to do.

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4683)* status: all files reviewed, 5 unresolved discussions (waiting on @JordiSubira, @lukedirtwalker, and @romshark)

katyatitkova:
Can it be merged?

PR #4686, opened 2025-01-27
refactor: replace deprecated grpc packages

romshark:
- replace deprecated `github.com/golang/protobuf` with `google.golang.org/protobuf`
- replace deprecated `ptypes.TimestampProto` from `github.com/golang/protobuf/ptypes` with `google.golang.org/protobuf/types/known/timestamppb`
- replace deprecated `grpc.WithInsecure()` with `google.golang.org/grpc/credentials/insecure`

Deprecation notice: https://pkg.go.dev/github.com/golang/protobuf

### DIFF
--- a/daemon/internal/servers/grpc.go
+++ b/daemon/internal/servers/grpc.go
@@ -178,7 +178,7 @@ func pathToPB(path snet.Path) *sdpb.Path {
 		},
 		Interfaces:   interfaces,
 		Mtu:          uint32(meta.MTU),
-		Expiration:   timestamppb.New(meta.Expiry.Truncate(time.Second)),
+		Expiration:   &timestamppb.Timestamp{Seconds: meta.Expiry.Unix()},
 		Latency:      latency,
 		Bandwidth:    meta.Bandwidth,
 		Geo:          geo,
--- a/daemon/internal/servers/grpc.go
+++ b/daemon/internal/servers/grpc.go
@@ -20,11 +20,11 @@ import (
 	"net"
 	"time"
 
-	durationpb "github.com/golang/protobuf/ptypes/duration"
-	timestamppb "github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/opentracing/opentracing-go"
 	"golang.org/x/sync/singleflight"
+	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/emptypb"
+	"google.golang.org/protobuf/types/known/timestamppb"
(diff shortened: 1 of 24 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4686)

daemon/internal/servers/grpc.go
```diff
@@ -178,7 +178,7 @@ func pathToPB(path snet.Path) *sdpb.Path {
 		},
 		Interfaces:   interfaces,
 		Mtu:          uint32(meta.MTU),
-		Expiration:   &timestamppb.Timestamp{Seconds: meta.Expiry.Unix()},
+		Expiration:   timestamppb.New(meta.Expiry.Truncate(time.Second)),
```
jiceathome: What is the advantage of time.Truncate(second) compared to time.Unix(). It seems to be a more complicated way of reaching the same result.

jiceatscion (commented):
As before... I should not approve as sole reviewer. But this sounds good. Besides my question regarding unix time. Add the change of timestamp pb to the description?

> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4686#-:-OHbe61H2geua0IOXUPj:bprd18b)* status: all files reviewed, 1 unresolved discussion (waiting on @oncilla and @romshark)

daemon/internal/servers/grpc.go
romshark: Yep, it's the same. I thought it's "cleaner" to use `New` instead of directly assigning fields. But it really doesn't matter and I can revert this change if you prefer.

romshark: I've reverted this change as debatable. Not worth discussing :)

oncilla (approved):
:lgtm:

> Reviewed 12 of 14 files at r1, 10 of 10 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4686#-:-OHcAfrS7g9tlPr3S0FO:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4686)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @romshark)

PR #4690, opened 2025-01-30
drkey: simplify Epoch type

aaronbojarski:
During verification of the drkey package (as part of the `VerifiedSCION` project) we ran into an issue with our verifier. It complains about the embedded struct in `Epoch`, which in this specific case, does not seem to be supported.

(Just not to confuse anyone: The issue is with the verifier (gobra), not with Go or the `drkey` package itself.)

However, to me there does not seem to be a reason why `Epoch` embedds `cppki.Validity`. It could as well just be a type definition. This is what I propose with this PR. It changes the definition of `Epoch` and adapts (and simplifies) the clients of the `drkey` package.


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4690)

romshark:
Please run `make gazelle` and commit the changes to fix the CI issues

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4690)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @JordiSubira)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/scionproto/scion/4690#-OHrBEJB48XME8gF1WI1:-OHrEgyk0d9-LB_GvE60:b-xushyp):*

This would appear to change a protobuf schema. If so that would be a protocol breaking change and would require a planned transition. Even if nothing is changed in the abstract. I have not yet looked closely enough to decide if the protobuf ends-up being compatible.
</blockquote></details>

Looks fine from that PoV.

JordiSubira (approved):
:lgtm: % @romshark comment

> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4690#-:-OHrJ07M9X6nMjFNwpA5:b-i9vzau)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aaronbojarski)

JordiSubira:
> Could you explain why the types Epoch and Validity continue to exist separately? If they were meant to be semantically interchangeable, and are now completely identical, why keep two different definitions? If we want to prevent invisible conversions we could create two different types that both refer to the common one: "type Validity NotBeforeNotAfter" and "type Epoch NotBeforeNotAfter" (note the absence of "=".

Initially Epoch existed independently from Validity was introduced (it existed in old versions of DRKey that didn't even belong to scionproto). I would say that currently they are semantically interchangeable and I do not think this will ever change in the future (or I cannot think about a good reason right now). However, keeping the alias may help with code readability and clarity. In the documentation we always refer to DRKey epochs, while probably TRCs talk about Validity. In that sense, I guess your proposal about creating ``NotBeforeNotAfter`` would also work.

JordiSubira (approved):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4690)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aaronbojarski)

aaronbojarski (commented):
Thank you for your feedback!

I am unsure what to do now. Do I leave it as is or add a `NotBeforeNotAfter` struct? 
And if so, is it fine to put the new struct into the `cppki` package or would you prefere someplace else?

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4690#-:-OHviWb4BpAqUf5O5EBF:bllutyr)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aaronbojarski)

jiceatscion (commented):
My proposal of two types plus NotBeforeNotAfter was just in case we *want* Epoch and Validity to not be implicitly convertible. Given what Jordi says, I don't think it is necessary. We can just choose one to hold the definition (say, Validity), and make the other one a true alias (that is, with "=") for bw compatibility.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4690#-:-OHw0_BjAwHjUyofD4UR:bcsirnf)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aaronbojarski)

jiceatscion (commented):
Only reason to create a 3rd type would be to avoid extraneous package dependencies; in which case I would create two aliases and put the definition in some common package.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4690#-:-OHw25ca9C7gUNAPY52-:b-g6ohti)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aaronbojarski)

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4690)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aaronbojarski)

___
*[`pkg/drkey/drkey.go` line 39 at r3](https://reviewable.io/reviews/scionproto/scion/4690#-OHwUXxB1vBHIom4fsgd:-OHwUXxB1vBHIom4fsge:b5v6ndl) ([raw file](https://github.com/scionproto/scion/blob/eaa585a3a032b98c5d103895c8e9eb1efff32429/pkg/drkey/drkey.go#L39)):*
> ```Go
> 
> // Epoch represents a validity period.
> type Epoch = cppki.Validity
> ```

I had completely missed the fact that you had already defined Epoch as a replica of Validiy. That conversation would have been much simpler. My apologies. Yes. an alias is best.

aaronbojarski (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4690)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aaronbojarski)

___
*[`pkg/drkey/drkey.go` line 39 at r3](https://reviewable.io/reviews/scionproto/scion/4690#-OHwUXxB1vBHIom4fsgd:-OHwWcF3Aa_PGuxPvWOw:b7upxub) ([raw file](https://github.com/scionproto/scion/blob/eaa585a3a032b98c5d103895c8e9eb1efff32429/pkg/drkey/drkey.go#L39)):*

I had completely missed the fact that you had already defined Epoch as a replica of Validiy. That conversation would have been much simpler. My apologies. Yes. an alias is best.
</blockquote></details>

Thank you! Did you see that one of the integration tests failed the first time it ran? Is that common?

PR #4702, opened 2025-02-13
chore: upgrade go to v1.24.0

romshark:
Since the release of [Go v1.24.0](https://tip.golang.org/doc/go1.24), Go 1.22 is no longer supported.

~~We might go for 1.24 for the various performance improvements and the significant changes to the crypto packages but that's up to collective consensus.~~

We can't upgrade to `v1.23.x` because of significant performance regression and instead we need to upgrade all the way to Go `v1.24.0`.

- `private/trust`: Change RSA key generation in `TestSignerGenGenerate` to use 1024-bit keys instead of 512-bit because the latter is no longer accepted, see https://github.com/golang/go/issues/70643
- **test:** slightly lower router benchmark expectations as there seems to be a general performance regression since the Go upgrade to v1.24 (not as much as v1.23.6 though).
- **ci:** remove linter `exportloopref` due to deprecation, see [golangci-lint/CHANGELOG](https://github.com/golangci/golangci-lint/blob/312fdf7a8d0516ced651630c6196da9d6053c2d7/CHANGELOG.md?plain=1#L251)
- **ci:** disable Go linker's new automatic build id setting which causes rpmbuild to fail due to all binaries having the same build id.
- **ci:** temporarily remove `nogo` due to incompatibility with Go 1.24 until further investigation.
- **ci:** upgrade golangci-lint to v1.64.5


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4702)

romshark:
We might go for 1.24 for the various performance improvements and the significant changes to the crypto packages but that's up to collective consensus.

marcfrei (commented):
@oncilla I don't remember all the details any more, but do you think this should be fine now?

https://github.com/scionproto/scion/pull/4620

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702#-:-OIysnwnBA7R9lrsxe_a:bwrufzs)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jiceatscion)

jiceatscion (commented):
It seems that the issue in question has been fixed by 1.23.3; so we should be good with 1.23.6

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702#-:-OIzkxRg3zJNsYepNfNf:b-iujlnj)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @romshark)

jiceatscion (approved):
> Reviewed 2 of 2 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @romshark)

marcfrei (approved):
Great, thanks for looking this up!
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702#-:-OIzmX8w7V6jdyVZKcXy:b8rws1f)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @romshark)

romshark:
As @jiceatscion shared in Slack, there seems to be a consistent performance-regression with Go 1.23, see https://github.com/golang/go/issues/67822

jiceatscion (commented):
Adding my own random findings here regarding the possible use of go 1.24 instead:
* Performance is better; almost the same as 1.22.7
* Until cause is found nogo is broken and has to be disabled (by commenting it out from the WORSPACE go_register_toolchains statement
* golangcilint is also borken but easy to fix: in Makefile the lint-golang-ci target refers to a specific version of golangci-lint. That needs to be upgrades to 1.64.5. Then there's a deprecated check (exportloopref) to be udated in the config.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702#-:-OJ33OTW-BJMg5tmaTKr:b-xbwcmr)* status: 6 of 10 files reviewed, all discussions resolved (waiting on @romshark)

jiceatscion (commented):
I think we can accept that we will be upgrading to 1.24 if we upgrade. So title needs to change.

Regarding the nogo issue. We're not the only ones and relief is afoot: https://github.com/bazel-contrib/rules_go/pull/4272
In the meantime, I think we have no choice but disabling nogo.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702#-:-OJ3VWrB5mBCs0VhA74d:b-y33jf7)* status: 6 of 10 files reviewed, all discussions resolved (waiting on @romshark)

jiceatscion:
Regarding the exportloopref check. Since go1.22 there's a new check that we should use: copyloopvar, which points out all the places where loop variables are copied to work-around the bugprone function scoping of loop variables that go1.22 did away with.

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702)* status: 6 of 10 files reviewed, 1 unresolved discussion (waiting on @jiceatscion and @romshark)

___
*[a discussion](https://reviewable.io/reviews/scionproto/scion/4702#-OJ3u-LX7G5tzhqV2mAO:-OJ3u-LY6gvNr8LrC5Nz:b-cxsqx0) (no related file):*
nit: please update title and description to reflect the changes

jiceatscion (commented):
I have to take one thing back. The update to rules_go v0.53 is indeed not smooth at all (I had forgotten to update the checksum in WORKSPACE so was still using the old one).

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702#-:-OJ3wBZE1UB-TTBd1urd:b-yjaxsw)* status: 6 of 10 files reviewed, 1 unresolved discussion (waiting on @romshark)

romshark:
It appears to me `rules_go` will require upgrading the major version of Bazel we use (v4). But I may be wrong.

jiceatscion (commented):
No we use V6.4 and I just upgraded to the latest v6.5 without any obvious issue.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702#-:-OJ40wXx7gldr51-la3D:b-42mxsn)* status: 6 of 10 files reviewed, 1 unresolved discussion (waiting on @romshark)

jiceatscion (commented):
Looks like we need to upgrade rules_proto_grpc, which no-longer supports WORKSPACE, which means that we'd need to switch to MODULES.bazel... and on and on... that's not a yak; that's a whole herd of them.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702#-:-OJ4NxLK1aRk9dQvMJ2h:b-gn2nq1)* status: 6 of 10 files reviewed, 1 unresolved discussion (waiting on @romshark)

romshark:
Changes in the linker (see https://tip.golang.org/doc/go1.24#linker) now lead to a new error.
All binaries from https://buildkite.com/scionproto/scion/builds/5037 have the same build id:

```
romshark@lima-default:/home/romshark/shared/scion_0.12.0-modified-ci_amd64_linux$ readelf -n router | grep "Build ID"
    Build ID: b78919aff001d8366249403a2544fba2d833084f
romshark@lima-default:/home/romshark/shared/scion_0.12.0-modified-ci_amd64_linux$ readelf -n dispatcher | grep "Build ID"
    Build ID: b78919aff001d8366249403a2544fba2d833084f
...
```

which leads to:
```
rpm -i /rpm/scion-dispatcher_dev_x86_64.rpm
	file /usr/lib/.build-id/b7/8919aff001d8366249403a2544fba2d833084f from install of scion-dispatcher-0.12.0^modified^ci-%autorelease.x86_64 conflicts with file f
```

romshark:
I can confirm that Go 1.23.6 linker doesn't write a build ID to the binary.
We can strip the build ID for now using:
```
objcopy --remove-section .note.gnu.build-id <binary_path>
```

jiceatscion (commented):
> Reviewed 1 of 2 files at r3, 1 of 2 files at r4, 2 of 2 files at r5, 1 of 1 files at r6, 14 of 14 files at r7, 1 of 1 files at r8, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702)* status: all files reviewed, 3 unresolved discussions (waiting on @romshark)

___
*[`.golangcilint.yml` line 13 at r10](https://reviewable.io/reviews/scionproto/scion/4702#-OJOA6dq8NaHuqYAhK4j:-OJOA6dq8NaHuqYAhK4k:bg7uzae) ([raw file](https://github.com/scionproto/scion/blob/09738f2a3582026aa48dbe6cfa7be14acce5a970/.golangcilint.yml#L13)):*
> ```YAML
>     - typecheck
>     - unused
>     # Extended linters.
> ```

I suggest adding the conterpart: copyloopvar, just commented out; so we have a chance to remember and turn it on later.
___
*[`acceptance/router_benchmark/test.py` line 45 at r10](https://reviewable.io/reviews/scionproto/scion/4702#-OJO9EeiBU2O1M3UeyhR:-OJO9EeiBU2O1M3UeyhS:bidwfr) ([raw file](https://github.com/scionproto/scion/blob/09738f2a3582026aa48dbe6cfa7be14acce5a970/acceptance/router_benchmark/test.py#L45)):*
> ```Python
> TEST_CASES = {
>     "in": 700000,
>     "out": 723000,
> ```

I case other reviewers are suspicious: done with my approval. We have been running very close to the limit and the test fails too often. It seems that Go is delivering regular small performance drops that we can't do much about, so no other choice.

jiceatscion (approved):
yeay!

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702#-:-OJOD6ohAH8kt386iwCv:b-7g7z48)* status: all files reviewed, 3 unresolved discussions (waiting on @romshark)

jiceatscion (approved):

jiceatscion (approved):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702)* status: all files reviewed, 3 unresolved discussions (waiting on @romshark)

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702)* status: all files reviewed, 1 unresolved discussion (waiting on @romshark)

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702)* status: all files reviewed, 1 unresolved discussion (waiting on @romshark)

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4702)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @romshark)

___
*[a discussion](https://reviewable.io/reviews/scionproto/scion/4702#-OJ3u-LX7G5tzhqV2mAO:-OJOcGdI87eJpjaDdSBv:b-896fjx) (no related file):*

nit: please update title and description to reflect the changes
</blockquote></details>

Done

PR #4703, opened 2025-02-14
router: add tcp support for dispatching to endhost

### LINKED ISSUE #4698: router: support for handling TCP/SCION packets at the last-mile AS
Currently, the BR does not handle the dispatch of TCP/SCION packets to the endhosts. Mainly we need to extens the switch cases for `DataPlane.addEndhostPort(·)`. Currently we do not have a TCP/SCION implementation, but some frameworks/deployments may still generate TCP/SCION traffic, e.g., https://github.com/netsys-lab/scion-ip-translator.

We may also want this on the shim dispatcher for backwards compatibility (I mention it here although it may deserve its own issue).

JordiSubira:
This PR resolves #4698. 

I followed the same principle that we use for the other protocols, i.e., we dispatch to the mapped port if the destination port is within the dispatched port range. Otherwise, it is sent to the fix `EndhostPort`. 

This makes it compatible with implementations following the "Wireguard"-model, i.e., sending the encapsulated traffic over a fix UDP port, e.g., the [SCION-IP translator]( https://github.com/netsys-lab/scion-ip-translator); or alternative implementations that map the underlay port to the application port, e.g., what we do with SCMP or UDP/SCION, provided proper `dispatched_ports` configuration.

### DIFF
--- a/router/dataplane.go
+++ b/router/dataplane.go
@@ -2127,6 +2127,16 @@ func (d *DataPlane) addEndhostPort(
 		if port < d.dispatchedPortStart || port > d.dispatchedPortEnd {
 			port = topology.EndhostPort
 		}
+	case slayers.L4TCP:
+		if len(lastLayer.LayerPayload()) < 20 {
+			// TODO: Treat this as a parameter problem
+			return serrors.New("SCION/TCP header len too small", "length",
+				len(lastLayer.LayerPayload()))
+		}
+		port = binary.BigEndian.Uint16(lastLayer.LayerPayload()[2:])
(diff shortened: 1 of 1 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4703)

jiceatscion (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4703)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JordiSubira)

PR #4705, opened 2025-02-18
build: upgrade to Bazel 7.5 (without migrating to modules)

jiceatscion:
Changes that were needed:

* Pain: Python binaries are somehow handled a bit differently and cannot import from a path relative to themselves by default. Remedy: One must now add an explicit root path for imports in py_binary rules.
* Pain: The prohibition on generate un-specified targets into a directory is now enforced. Gendocs is broken. Remedy: Replace the gendocs genrules with a fully specified rule that declare a directory output.
* Pain: Bazel uses a new method of sandboxing by default: containers. Tests that have a dockerized component and mount a data directory end-up with an empty data directory. I can only speculate why (because containers end-up nested, or because they don't). Remedy: Use an obscure set of magic arguments to restore the old sandboxing behavior for tests.
* Pain: bazel now creates canonically named directories for external deps. They confuse our licences.sh script so that licence files are produced and expected by the check_generated CI test, but aren't necessarily included in PRs. Remedy: filter those directories out when extracting licences; they are redundant.
* Pain: https://github.com/bazelbuild/rules_python/issues/1675. Remedy: --noenable_bzlmod


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4705)

lukedirtwalker (changes requested):
> Reviewed 9 of 12 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4705)* status: all files reviewed, 6 unresolved discussions (waiting on @jiceatscion)

___
*[`.bazelrc` line 29 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS25VoEvOhg59_sNnY:-OJS25VoEvOhg59_sNnZ:bjy9t9m) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/.bazelrc#L29)):*
> ```
> 
> ### options for test
> test --build_tests_only --print_relative_test_log_paths --test_output=errors --nobuild_python_zip  --spawn_strategy=remote,worker,processwrapper-sandbox,local
> ```

for us this was enough:
```
# disable binding of a per target /tmp directory
build --noincompatible_sandbox_hermetic_tmp
build --sandbox_add_mount_pair=/tmp
```
I think we mostly mount from tmp?

But if what you have also works, then no objection.
___
*[`scion/cmd/scion/gendocs.bzl` line 0 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS2TE7A-Yr-rqnXxu-:-OJS2TE7A-Yr-rqnXxu0:b-mcyrcl) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/scion/cmd/scion/gendocs.bzl#L0)):*
drop again?
___
*[`scion-pki/cmd/scion-pki/BUILD.bazel` line 37 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS0yj10MHR0T9HIKkp:-OJS0yj10MHR0T9HIKkq:b-lk4iw8) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/scion-pki/cmd/scion-pki/BUILD.bazel#L37)):*
> ```Starlark
> gendocs(
>     name = "gendocs",
>     # outs = ["docs"],
> ```

should we drop the comment here?
___
*[`gendocs.bzl` line 14 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS1jFS5mw7eb5AsyCe:-OJS1jFS5mw7eb5AsyCf:b-g8my3m) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/gendocs.bzl#L14)):*
> ```Starlark
>     return [DefaultInfo(files = depset([dir]))]
> 
> gendocs = rule(
> ```

an alternative would be to use https://github.com/bazel-contrib/bazel-lib/blob/main/docs/run_binary.md if you don't want to add our own "rule". (non-blocking as this approach also works)
___
*[`acceptance/common/BUILD.bazel` line 33 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS00Wx8VL6858DxeM5:-OJS00Wx8VL6858DxeM6:b-tc7768) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/acceptance/common/BUILD.bazel#L33)):*
> ```Starlark
>     name = "scion",
>     srcs = ["scion.py"],
>     data = [
> ```

why does that need to be a data dependency?
___
*[`acceptance/router_benchmark/BUILD.bazel` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS2OMNBOj3z-iLkwP0:-OJS2OMNBOj3z-iLkwP1:b-hj5dyd) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/acceptance/router_benchmark/BUILD.bazel#L41)):*
> ```Starlark
>     data = data,
>     homedir = "$(rootpath //docker:router.tarball)",
>     imports = ["."],
> ```

this is the default, so no change needed here.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4705)* status: 11 of 13 files reviewed, 5 unresolved discussions (waiting on @lukedirtwalker)

___
*[`.bazelrc` line 29 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS25VoEvOhg59_sNnY:-OJTdUt6-H69E0uPwjTV:bwd9fdr) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/.bazelrc#L29)):*

for us this was enough:
```
# disable binding of a per target /tmp directory
build --noincompatible_sandbox_hermetic_tmp
build --sandbox_add_mount_pair=/tmp
```
I think we mostly mount from tmp?

But if what you have also works, then no objection.
</blockquote></details>

I wouldn't want someone else to be surprised later when the mount isn't "/tmp". Information about this was pretty hard to find. So, I think I'll keep my variant.
___
*[`gendocs.bzl` line 14 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS1jFS5mw7eb5AsyCe:-OJU18CN4cXvFFzmPgds:bxabnrf) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/gendocs.bzl#L14)):*

an alternative would be to use https://github.com/bazel-contrib/bazel-lib/blob/main/docs/run_binary.md if you don't want to add our own "rule". (non-blocking as this approach also works)
</blockquote></details>

Sure, that sounds like what I wanted. Didn't know it existed. However, I just tried very hard to use that and it doesn't work: the tool is executed without error (I caused one, just to be sure) but the output is nowhere to be found. If you have a suspicion why, please let me know. I do prefer run_binary.

Here's how I attempted to use it:

run_binary(
    name = "gendocs",
    args = [
        "gendocs",
        "docs",
    ],
    out_dirs = ["docs"],
    tool = ":scion",
    visibility = ["//visibility:public"],
)
___
*[`acceptance/common/BUILD.bazel` line 33 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS00Wx8VL6858DxeM5:-OJU7qfF5zAGjueLQt1L:b31ijv0) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/acceptance/common/BUILD.bazel#L33)):*

why does that need to be a data dependency?
</blockquote></details>

it doesn't. I think that might have been required when the overagressive sandboxing was in effect. Not sure.
___
*[`acceptance/router_benchmark/BUILD.bazel` line 41 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS2OMNBOj3z-iLkwP0:-OJUGPxOA_OBpes2CDb_:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/acceptance/router_benchmark/BUILD.bazel#L41)):*

this is the default, so no change needed here.
</blockquote></details>

Done.
___
*[`scion-pki/cmd/scion-pki/BUILD.bazel` line 37 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS0yj10MHR0T9HIKkp:-OJU4s_N7wGdFLx32tVK:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/scion-pki/cmd/scion-pki/BUILD.bazel#L37)):*

should we drop the comment here?
</blockquote></details>

Done.
___
*[`scion/cmd/scion/gendocs.bzl` line  at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS2TE7A-Yr-rqnXxu-:-OJU77Cc5sT4swmD2ned:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/scion/cmd/scion/gendocs.bzl#L)):*

drop again?
</blockquote></details>

Done.

katyatitkova:
*[`gendocs.bzl` line 14 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS1jFS5mw7eb5AsyCe:-OJUuzAkEmyS-0OglzZX:b1qbj5p) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/gendocs.bzl#L14)):*

Sure, that sounds like what I wanted. Didn't know it existed. However, I just tried very hard to use that and it doesn't work: the tool is executed without error (I caused one, just to be sure) but the output is nowhere to be found. If you have a suspicion why, please let me know. I do prefer run_binary.

Here's how I attempted to use it:

run_binary(
    name = "gendocs",
    args = [
        "gendocs",
        "docs",
    ],
    out_dirs = ["docs"],
    tool = ":scion",
    visibility = ["//visibility:public"],
)
</blockquote></details>

Can confirm: I tried run_binary on my own as part of the bzlmod migration before seeing JC's PR, and couldn't make it work as well.

oncilla (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4705)* status: 7 of 13 files reviewed, 5 unresolved discussions (waiting on @jiceatscion, @katyatitkova, and @lukedirtwalker)

___
*[`gendocs.bzl` line 14 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS1jFS5mw7eb5AsyCe:-OJVF61v1VA715jPPsnM:b1w7fgi) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/gendocs.bzl#L14)):*
<details><summary><i>Previously, katyatitkova (Katya Titkova) wrote…</i></summary><blockquote>

Can confirm: I tried run_binary on my own as part of the bzlmod migration before seeing JC's PR, and couldn't make it work as well.
</blockquote></details>

Try

```
run_binary(
    name = "gendocs",
    args = [
        "gendocs",
        "$@",
    ],
    out_dirs = ["docs"],
    tool = ":scion",
    visibility = ["//visibility:public"],
)
```

lukedirtwalker (approved):
> Reviewed 1 of 1 files at r3, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4705)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion and @katyatitkova)

___
*[`.bazelrc` line 29 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS25VoEvOhg59_sNnY:-OJXZ61u2JFcMnKxBUec:b-ny51u5) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/.bazelrc#L29)):*
> I wouldn't want someone else to be surprised later when the mount isn't "/tmp"

I don't understand this statement.

What does the `--spawn_strategy=remote,worker,processwrapper-sandbox,local` actually do? Does it try each spawn strategy until one succeeds?

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4705)* status: 10 of 14 files reviewed, all discussions resolved (waiting on @katyatitkova, @lukedirtwalker, and @oncilla)

___
*[`.bazelrc` line 29 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS25VoEvOhg59_sNnY:-OJXg4D9AihsBO2S-xHJ:b-eugkm2) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/.bazelrc#L29)):*

> I wouldn't want someone else to be surprised later when the mount isn't "/tmp"

I don't understand this statement.

What does the `--spawn_strategy=remote,worker,processwrapper-sandbox,local` actually do? Does it try each spawn strategy until one succeeds?
</blockquote></details>

Yes. It's not my invention; it is the internal default setting for spawn_strategy; except that I replaced linux-sandbox with wrapper-sandbox (as it was before they switched to linux-sandbox). I found this default setting in a bazel blog post dating back to 0.27! Regarding the surprise issue. What I meant is that your solution works only when the test container mounts /tmp. That may be all we need, but it is setting the stage for the issue to re-appear if there's ever a container that mounts something else. If that ever happens the error will be pretty obscure, so I'd rather we don't waste time on this again in the future. There may be a better way, but there's not a whole lot of intelligible documentation about bazel's management of sandboxing; so I'm taking what I can find.
___
*[`gendocs.bzl` line 14 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS1jFS5mw7eb5AsyCe:-OJYgvpa-ZxPoG3bz3s4:bdj44ay) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/gendocs.bzl#L14)):*

Try

```
run_binary(
    name = "gendocs",
    args = [
        "gendocs",
        "$@",
    ],
    out_dirs = ["docs"],
    tool = ":scion",
    visibility = ["//visibility:public"],
)
```
</blockquote></details>

done It works now. Thanks!

lukedirtwalker (approved):
> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4705)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

___
*[`.bazelrc` line 29 at r2](https://reviewable.io/reviews/scionproto/scion/4705#-OJS25VoEvOhg59_sNnY:-OJb7AKpBuAaY3KclhDY:b49ghcq) ([raw file](https://github.com/scionproto/scion/blob/cd1c556aa874b91a8c2654f31b94f09055301b63/.bazelrc#L29)):*

Yes. It's not my invention; it is the internal default setting for spawn_strategy; except that I replaced linux-sandbox with wrapper-sandbox (as it was before they switched to linux-sandbox). I found this default setting in a bazel blog post dating back to 0.27! Regarding the surprise issue. What I meant is that your solution works only when the test container mounts /tmp. That may be all we need, but it is setting the stage for the issue to re-appear if there's ever a container that mounts something else. If that ever happens the error will be pretty obscure, so I'd rather we don't waste time on this again in the future. There may be a better way, but there's not a whole lot of intelligible documentation about bazel's management of sandboxing; so I'm taking what I can find.
</blockquote></details>

there is some documentation on it here: https://bazel.build/versions/7.5.0/docs/sandboxing?hl=en#sandboxing-strategies but I agrees it's quite little.

PR #4714, opened 2025-02-26
fix: enable linter contextcheck and fix potential problems

romshark:
- ci: enable linter `contextcheck`
- fix: potentially incorrect uses of `context.Context`.

### DIFF
--- a/tools/await-connectivity
+++ b/tools/await-connectivity
@@ -76,6 +76,7 @@ check_connectivity() {
 	  echo "$as: Waiting for Upsegs"
 	  ret=1
     fi
+    sleep 2
   done
   return $ret
 }
--- a/tools/await-connectivity
+++ b/tools/await-connectivity
@@ -76,7 +76,6 @@ check_connectivity() {
 	  echo "$as: Waiting for Upsegs"
 	  ret=1
     fi
-    sleep 2
   done
   return $ret
 }
--- a/.golangcilint.yml
+++ b/.golangcilint.yml
@@ -19,6 +19,7 @@ linters:
     - lll
     - misspell
     - goheader
+    - contextcheck
 linters-settings:
   # ...
   forbidigo:
--- a/control/beaconing/originator_test.go
+++ b/control/beaconing/originator_test.go
@@ -1,4 +1,5 @@
 // Copyright 2019 Anapaya Systems
+// Copyright 2025 SCION Association
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -94,10 +95,10 @@ func TestOriginatorRun(t *testing.T) {
 
 				sender := mock_beaconing.NewMockSender(mctrl)
 				sender.EXPECT().Send(gomock.Any(), gomock.Any()).Times(1).DoAndReturn(
-					func(_ context.Context, b *seg.PathSegment) error {
+					func(ctx context.Context, b *seg.PathSegment) error {
 						// Check the beacon is valid and verifiable.
 						assert.NoError(t, b.Validate(seg.ValidateBeacon))
-						assert.NoError(t, b.VerifyASEntry(context.Background(),
+						assert.NoError(t, b.VerifyASEntry(ctx,
 							segVerifier{pubKey: pub}, b.MaxIdx()))
 						// Extract the hop field from the current AS entry to compare.
 						hopF := b.ASEntries[b.MaxIdx()].HopEntry.HopField
--- a/control/beaconing/propagator_test.go
+++ b/control/beaconing/propagator_test.go
@@ -95,12 +95,12 @@ func TestPropagatorRunNonCore(t *testing.T) {
 		gomock.Any()).Times(1).DoAndReturn(
 
 		func(_ context.Context, _ addr.IA, egIfID uint16,
-			nextHop *net.UDPAddr) (beaconing.Sender, error) {
-
+			nextHop *net.UDPAddr,
+		) (beaconing.Sender, error) {
 			sender := mock_beaconing.NewMockSender(mctrl)
 			sender.EXPECT().Send(gomock.Any(), gomock.Any()).Times(3).DoAndReturn(
-				func(_ context.Context, b *seg.PathSegment) error {
-					validateSend(t, b, egIfID, nextHop, pub, topo)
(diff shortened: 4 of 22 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4714)

jiceatscion (approved):
Sounds good to me, but I'd wait for someone with more historical knowledge to weigh on the issue. There might have been reasons why the surrounding context was not being used in some of these cases. The act seemed sometimes a bit too deliberate (i.e. requiring extra effort) to be an oversight.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4714#-:-OK2HIBWDF-4O7pbyikW:b-ot3l7x)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @romshark)

oncilla (commented):
> Reviewed 8 of 12 files at r1, 12 of 12 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4714)* status: all files reviewed, 3 unresolved discussions (waiting on @romshark)

___
*[`daemon/internal/servers/grpc.go` line 211 at r3](https://reviewable.io/reviews/scionproto/scion/4714#-ONoirc87dqA3NpCKerh:-ONoirc87dqA3NpCKeri:br9jlac) ([raw file](https://github.com/scionproto/scion/blob/c401704ce63de88471f10bc7268b57e6e66bc61b/daemon/internal/servers/grpc.go#L211)):*
> ```Go
> 		return
> 	}
> 	ctx, cancelF := context.WithTimeout(origCtx, backgroundTimeout)
> ```

This is the opposite of what we want. We want to decouple the path fetch from the original context here.

This is a background fetch that should continue even in case the original context is cancelled.
___
*[`control/cmd/control/main.go` line 398 at r3](https://reviewable.io/reviews/scionproto/scion/4714#-ONohuCJ7wZ0GdiuS4ZY:-ONohuCJ7wZ0GdiuS4ZZ:bntq4ie) ([raw file](https://github.com/scionproto/scion/blob/c401704ce63de88471f10bc7268b57e6e66bc61b/control/cmd/control/main.go#L398)):*
> ```Go
> 	}
> 
> 	ctx, cancel := context.WithTimeout(ctx, time.Second)
> ```

please name the context something more specific.
The way it is now, we run the risk of reusing it by accident latter on in the function.  (when new code is added)

Alternatively, you can scope it locally in a block.
___
*[`daemon/cmd/daemon/main.go` line 157 at r3](https://reviewable.io/reviews/scionproto/scion/4714#-ONoiWmXEqfU7QqC9EMH:-ONoiWmXEqfU7QqC9EMI:bdx8tw5) ([raw file](https://github.com/scionproto/scion/blob/c401704ce63de88471f10bc7268b57e6e66bc61b/daemon/cmd/daemon/main.go#L157)):*
> ```Go
> 	})
> 	engine, err := daemon.TrustEngine(
> 		errCtx, globalCfg.General.ConfigDir, topo.IA(), trustDB, dialer,
> ```

why do you use the errCtx instead of the regular ctx here?

romshark:
*[`control/cmd/control/main.go` line 398 at r3](https://reviewable.io/reviews/scionproto/scion/4714#-ONohuCJ7wZ0GdiuS4ZY:-ONtioaF40ES_zBFG1a8:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c401704ce63de88471f10bc7268b57e6e66bc61b/control/cmd/control/main.go#L398)):*

please name the context something more specific.
The way it is now, we run the risk of reusing it by accident latter on in the function.  (when new code is added)

Alternatively, you can scope it locally in a block.
</blockquote></details>

Done.

romshark:
*[`daemon/cmd/daemon/main.go` line 157 at r3](https://reviewable.io/reviews/scionproto/scion/4714#-ONoiWmXEqfU7QqC9EMH:-ONtistv4Z_gx8ts7Rp-:b-mlvab4) ([raw file](https://github.com/scionproto/scion/blob/c401704ce63de88471f10bc7268b57e6e66bc61b/daemon/cmd/daemon/main.go#L157)):*

why do you use the errCtx instead of the regular ctx here?
</blockquote></details>

as soon as something in the errgroup goes out the window - daemon.TrustEngine should be canceled. Am I mistaking?

romshark:
*[`daemon/internal/servers/grpc.go` line 211 at r3](https://reviewable.io/reviews/scionproto/scion/4714#-ONoirc87dqA3NpCKerh:-ONtjL-02mUzcyKE5dTH:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/c401704ce63de88471f10bc7268b57e6e66bc61b/daemon/internal/servers/grpc.go#L211)):*

This is the opposite of what we want. We want to decouple the path fetch from the original context here.

This is a background fetch that should continue even in case the original context is cancelled.
</blockquote></details>

Done.

tools/await-connectivity
```diff
@@ -76,6 +76,7 @@ check_connectivity() {
 	  echo "$as: Waiting for Upsegs"
 	  ret=1
     fi
+    sleep 2
```
katyatitkova: Do you mind writing a small comment explaining why this was needed?

romshark: Oh that's debugging leftovers. Removed. Good catch!

oncilla (approved):
> Reviewed 2 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4714)* status: all files reviewed, 1 unresolved discussion (waiting on @katyatitkova)

___
*[`daemon/cmd/daemon/main.go` line 157 at r3](https://reviewable.io/reviews/scionproto/scion/4714#-ONoiWmXEqfU7QqC9EMH:-ONu6Hc12CGF4aJTxrRg:bz7241e) ([raw file](https://github.com/scionproto/scion/blob/c401704ce63de88471f10bc7268b57e6e66bc61b/daemon/cmd/daemon/main.go#L157)):*

as soon as something in the errgroup goes out the window - daemon.TrustEngine should be canceled. Am I mistaking?
</blockquote></details>

I guess it is not wrong, but it is also not intuitive.

resolving my comment, as it doesn't really matter.

PR #4724, opened 2025-03-11
docs: fix tutorial config cs.toml and topology files to reflect tutorial version update to v.0.12.0

guilloboi1917:
Hi,

minor updates to adapt tutorial to v.0.12.0.

Removing "reconnect_to_dispatcher" field from cs.toml and adding "dispatched_ports" field to topology files.

Best,
Noah

### DIFF
--- a/doc/tutorials/deploy/cs.toml
+++ b/doc/tutorials/deploy/cs.toml
@@ -3,7 +3,6 @@
 [general]
 id = "cs"
 config_dir = "/etc/scion"
-reconnect_to_dispatcher = true
 
 [log.console]
 level = "info"
--- a/doc/tutorials/deploy/topology1.json
+++ b/doc/tutorials/deploy/topology1.json
@@ -3,6 +3,7 @@
     "core"
   ],
   "isd_as": "42-ffaa:1:1",
+  "dispatched_ports": "31000-32767",
   "mtu": 1472,
   "control_service": {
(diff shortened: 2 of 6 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4724)

JordiSubira (approved):
In principle the modifications on the config files should work now. However, I would like to mention a couple of things (which perhaps can be discussed elsewhere to avoid blocking this PR):
1. The dispatcher.toml configuration does NOT configure the shim dispatcher to forward packets, i.e., if some user tries to have a SCION application running on a port outside the dispatched range, e.g., 443, the application will not receive SCION traffic. Should we mention this on the documentation?
2. Should we have a test in the pipeline for the freestanding development. At the end, many people end up picking this tutorial option for trying out SCION (perhaps because it is the only one under that section...). Being this the case, I believe it is important that we try out basic connectivity on the described setup...

As said before, I am not blocking this PR.

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4724#-:-OLcYwyNE_TRqgdt9I-_:buazl4b)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @guilloboi1917)

katyatitkova:
@guilloboi1917 Thank you for your first contribution!

PR #4729, opened 2025-03-12
build: update Buildkite's PATH

katyatitkova:
As JC wrote in https://github.com/scionproto/scion/issues/4728, Buildkite v6.34.0 deleted `/usr/sbin` and `/usr/local/sbin` from `PATH`, which breaks our integration tests when using newer Buildkite versions. I set up a temporary pipeline where Buildkite is upgraded to v6.34.0, here's the build for this branch there: https://buildkite.com/scionproto/temporary/builds/30#01958a44-8b24-40e1-8c77-f1baa790db64

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -7,7 +7,7 @@ pathadd() {
 }
 pathadd ~/.local/bin
 pathadd /usr/sbin
-pathadd /sbin
+pathadd /usr/local/sbin
 printenv PATH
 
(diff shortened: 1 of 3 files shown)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4729)

jiceatscion:
We would normally sollicit a non-SCION-association reviewer, but in this case it seems quite superfluous. So, let's save everyone's time.

PR #4737, opened 2025-03-13
router: refactor phase 4

jiceatscion:
Made the router almost entirely oblivious of underlay addresses. The router now understands only
one underlay address: its own local address on the internal network (and still relies on the assumption that the internal link is on the udpip underlay.)

The router still provides storage for the remote address in the packet structure, which is a net.UdpAddr. It does not use it but resets it when taking a free packet from the pool. This will take more effort to eliminate: we'll need to provide underlay agnostic storage for RemoteAddr.

Sorry for the size of the change. Nothing very fundamental but there was a lot a nitty-gritty affected. There is one loosely related change that might be undesirable: I acted the deprecation of the Public and Bind address fields in topology configs. It just reduced the amount of code to adjust but it could be reverted if we know ppl still depend on it.

Contributes to: [#4593](https://github.com/scionproto/scion/issues/4593)


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4737)

router/underlay.go
```diff
@@ -61,9 +66,28 @@ type Link interface {
 // make them opaque; to be interpreted only by the underlay implementation.
 type UnderlayProvider interface {
 
+	// SetConnNewer is a unit testing device: it allows the replacement of the function
+	// that creates new underlay connections. Underlay implementations can, at their
+	// choice, implement this properly, or panic if it is called. The tests have to know
+	// which it is. Only router/export_test invokes this directly.
+	SetConnNewer(newer any)
```
romshark, line 74: From the discussion in person: Maybe there's a better more specific type that could be used instead of `any`, maybe not 🤷🏻‍♂️

router/dataplane.go
```diff
@@ -85,15 +85,17 @@ type BatchConn interface {
...
 
-func AddUnderlay(newProvider func(int) UnderlayProvider) {
-	newUnderlay = newProvider
+// AddUnderlay registers the named factory function.
+func AddUnderlay(name string, newProvider func(int, int, int) UnderlayProvider) {
+	if underlayProviders == nil {
+		underlayProviders = make(map[string]func(int, int, int) UnderlayProvider)
+	}
+	underlayProviders[name] = newProvider
 }
```
romshark:
```suggestion
var NewProviderFn func(batchSize, receiveBufferSize, sendBufferSize int) UnderlayProvider

// underlayProviders is a map of our underlay providers. Each entry associates a name with a factory
// function that creates a new instance of the named provider. A new instance is created every time
// so multiple dataplane instances can co-exist (as is routinely done by tests).
var underlayProviders map[string]NewProviderFn

// AddUnderlay registers the named factory function.
func AddUnderlay(name string, newProvider NewProviderFn) {
	if underlayProviders == nil {
		underlayProviders = make(map[string]NewProviderFn)
	}
	underlayProviders[name] = newProvider
}
```

romshark: Main problem here is the three `int` are not named which makes it error-prone.

router/dataplane.go
```diff
@@ -204,6 +205,8 @@ type dataPlane struct {
...
 
 var (
+	UnsupportedV4MappedV6Address  = errors.New("unsupported v4mapped IP v6 address")
+	UnsupportedUnspecifiedAddress = errors.New("unsupported unspecified address")
```
romshark, line 211: Consistency-wise it makes sense to keep it that way, but correctness-wise it should have the `Err` prefix. We should probably keep this as is for now and then do the cleanup in a separate PR later.

router/dataplane.go
```diff
@@ -353,52 +366,80 @@ func (d *dataPlane) SetPortRange(start, end uint16) {
...
 	}
+	if link.Remote.Addr == "" {
+		return emptyValue
+	}
+
+	underlay, instantiated := d.underlays[link.Provider]
+	if !instantiated {
+		underlayProvider, exists := underlayProviders[link.Provider]
+		if !exists {
+			panic("No provider for underlay " + link.Provider)
```
romshark:
```suggestion
			panic(fmt.Sprintf("no provider for underlay: %q", link.Provider))
```
This way we avoid a potential `no provider for underlay ` and get a more readable `no provider for underlay: ""`

router/dataplane.go
```diff
@@ -511,44 +536,69 @@ func (d *dataPlane) DelSvc(svc addr.SVC, a netip.AddrPort) error {
...
 		return serrors.JoinNoStack(alreadySet, nil, "ifID", ifID)
 	}
+	if link.Remote.Addr == "" {
+		return emptyValue
+	}
+	underlay, instantiated := d.underlays[link.Provider]
+	if !instantiated {
+		underlayProvider, exists := underlayProviders[link.Provider]
+		if !exists {
+			panic("No provider for underlay " + link.Provider)
```
romshark: Same thing here as with the other panic.

router/underlayproviders/udpip/udpip.go
```diff
@@ -29,23 +29,49 @@ import (
...
+	"github.com/scionproto/scion/private/underlay/conn"
 	"github.com/scionproto/scion/router"
 	"github.com/scionproto/scion/router/bfd"
 )
 
+var (
+	resolveOnSiblingLink  = errors.New("unsupported address resolution on sibling link")
+	resolveOnExternalLink = errors.New("unsupported address resolution on external link")
+	invalidServiceAddress = errors.New("invalid service address")
+)
```
romshark: Here I'd actually apply the `err` prefix convention

router/underlayproviders/udpip/udpip.go
```diff
@@ -29,23 +29,49 @@ import (
...
+
+// An interface to enable unit testing.
+type ConnNewer interface {
```
romshark, line 47: How about "TestConnCreator" instead? "ConnNewer" might be confused with "newer connection".

router/underlayproviders/udpip/udpip.go
```diff
@@ -29,23 +29,49 @@ import (
...
+	New(l netip.AddrPort, r netip.AddrPort, c *conn.Config) (router.BatchConn, error)
+}
+
+// The default ConnNewer for this underlay: opens an underlay conn.
+type un struct {
```
romshark, line 52: That name `un` isn't very telling, but I struggle to propose a better alternative.

router/underlayproviders/udpip/udpip.go
```diff
@@ -382,28 +445,38 @@ type externalLink struct {
...
-		// The router doesn't do this. This is an internal error.
-		panic("Zero address not supported")
+	localAddr, err := conn.ResolveAddrPortOrPort(local)
+	if err != nil {
+		return nil, err
+	}
+	remoteAddr, err := conn.ResolveAddrPort(remote)
+	if err != nil {
+		return nil, err
 	}
```
romshark:
```suggestion
	localAddr, err := conn.ResolveAddrPortOrPort(local)
	if err != nil {
		return nil, fmt.Errorf("resolving local addr: %w")
	}
	remoteAddr, err := conn.ResolveAddrPort(remote)
	if err != nil {
		return nil, fmt.Errorf("resolving remote addr: %w")
	}
```
I suppose this code isn't performance-critical and the potential additional wrapping cost is irrelevant but would improve debuggability.

router/underlayproviders/udpip/udpip.go
```diff
@@ -382,28 +445,38 @@ type externalLink struct {
...
 
 	u.mu.Lock()
 	defer u.mu.Unlock()
 
-	if l := u.allLinks[remote]; l != nil {
+	if l := u.allLinks[remoteAddr]; l != nil {
 		// This is a really bad idea. We can't just panic because it may be a configuration error.
 		// So, we have to return an error.
 		return nil, serrors.New("remote address reused")
```
romshark: Consider making this a sentinel error.

router/underlayproviders/udpip/udpip.go
```diff
@@ -519,35 +597,36 @@ type siblingLink struct {
...
-		panic("Zero address not supported")
+	// We don't currently use the address, but that will change. It *must* be valid.
+	_, err := conn.ResolveAddrPortOrPort(local)
+	if err != nil {
+		return nil, err
+	}
+	remoteAddr, err := conn.ResolveAddrPort(remote)
+	if err != nil {
+		return nil, err
 	}
```
romshark: ditto

router/underlayproviders/udpip/udpip.go
```diff
@@ -733,6 +832,42 @@ func (l *internalLink) IsUp() bool {
...
+	case addr.HostTypeIP:
+		dstAddr = dst.IP()
+		if dstAddr.Is4In6() {
+			return router.UnsupportedV4MappedV6Address
+		}
+		if dstAddr.IsUnspecified() {
+			return router.UnsupportedUnspecifiedAddress
+		}
+	default:
+		panic("unexpected address type returned from DstAddr")
```
romshark:
```suggestion
		panic(fmt.Sprintf("unexpected address type returned from DstAddr: %s", dst.Type().String()))
```

router/mgmtapi/api.go
```diff
@@ -119,8 +119,9 @@ func (s *Server) GetInterfaces(w http.ResponseWriter, r *http.Request) {
 
 	for _, intf := range siblingInterfaces {
 		siblingInterface := SiblingInterface{
-			InterfaceId:       int(intf.IfID), // nolint - name from published API.
-			InternalInterface: intf.InternalInterface.String(),
+			InterfaceId: int(intf.IfID), // nolint - name from published API.
+			// The name InternalInterface is poorly chosen but enshrinned in the schema.
```
romshark:
```suggestion
			// The name InternalInterface is poorly chosen but enshrined in the schema.
```
just a typo

router/control/conf.go
```diff
@@ -238,8 +276,16 @@ func confServices(dp Dataplane, cfg *Config) error {
 		sort.Slice(addrs, func(i, j int) bool {
 			return addrs[i].IP.String() < addrs[j].IP.String()
 		})
+
+		// Topo.Multicast returns SCION host addresses (which just happen to be identical to udp/ip
+		// addresses). So, in theory, these are *not* underlay addresses. The topology API
+		// represents them openly as UdpAddr. The router doesn't make that assumption. It does not
```
romshark:
```suggestion
		// represents them openly as UDPAddr. The router doesn't make that assumption. It does not
```
typo?

router/control/conf.go
```diff
@@ -197,21 +206,49 @@ func confExternalInterfaces(dp Dataplane, cfg *Config) error {
...
+			return err
+		}
+		localHost := addr.HostIP(localAddr.Addr())
+
+		remoteAddr, err := netip.ParseAddrPort(linkInfo.Remote.Addr)
+		if err != nil {
+			return err
+		}
+		remoteHost := addr.HostIP(remoteAddr.Addr())
+
```
romshark: ditto. I suppose it'd make sense to wrap those errors.

romshark (commented):
LGTM, aside from my comments and suggestions.

> Reviewed 1 of 10 files at r1, 25 of 25 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4737#-:-OM7xxVTA9jmpebHXSNh:b5qxqgx)* status: all files reviewed, 15 unresolved discussions (waiting on @jiceatscion, @JordiSubira, @katyatitkova, @lukedirtwalker, and @marcfrei)

router/dataplane.go
jiceatscion: Done.

router/dataplane.go
jiceatscion: Agree, the errors need some work. I'd prefer it if underlays had their own error space tool, rather than sharing with the router code.

router/dataplane.go
jiceatscion: Done.

router/dataplane.go
jiceatscion: Done.

router/underlay.go
jiceatscion: I gave it some thought and came to the conclusion that there was nothing to be gained by making the type of the newer more specific. The best we could do would be to require that it has a New method which returns something. That doesn't help. No mater what, the test that uses this needs to implement one specific newer interface of one specific underlay provider. So additional type constraints won't help. A better idea is welcome, but I don't have one.

router/control/conf.go
jiceatscion: Done.

router/control/conf.go
jiceatscion: Done.

router/mgmtapi/api.go
jiceatscion: Done.

router/underlayproviders/udpip/udpip.go
jiceatscion: Done.

router/underlayproviders/udpip/udpip.go
jiceatscion: I was applying Go's convention (for a change) that a single method interface is named after the name of the method. Not my fault if that convention almost universally leads to bad grammar. Any suggestion that accomplishes the miracle of following the Go convention while being intelligible? Should we name it Connecter (per Go's English grammar) and the method be Connect()? Problem is that such vocabulary is terminally oversubscribed. ConnMaker/Make? I cannot call it "TestConnAnything", because its default implementation is used for real connections.

router/underlayproviders/udpip/udpip.go
jiceatscion: So did I. At least, that one goes by Go's conventions: small scope, short name. Let's wait for ideas from other reviewers?

router/underlayproviders/udpip/udpip.go
jiceatscion: Done.

router/underlayproviders/udpip/udpip.go
jiceatscion: Done.

router/underlayproviders/udpip/udpip.go
jiceatscion: Done.

router/underlayproviders/udpip/udpip.go
jiceatscion: Done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4737)* status: all files reviewed, 15 unresolved discussions (waiting on @JordiSubira, @katyatitkova, @lukedirtwalker, @marcfrei, and @romshark)

router/dataplane.go
jiceathome: Done.

router/dataplane.go
jiceathome: Agree, the errors need some work. I'd prefer it if underlays had their own error space too, rather than sharing with the router code.

router/dataplane.go
jiceathome: Done.

router/dataplane.go
jiceathome: Done.

router/underlay.go
jiceathome: I gave it some thought and came to the conclusion that there was nothing to be gained by making the type of the newer more specific. The best we could do would be to require that it has a New method which returns something. That doesn't help. No mater what, the test that uses this needs to implement one specific newer interface of one specific underlay provider. So additional type constraints won't help. A better idea is welcome, but I don't have one.

router/control/conf.go
jiceathome: Done.

router/control/conf.go
jiceathome: Done.

router/mgmtapi/api.go
jiceathome: Done.

router/underlayproviders/udpip/udpip.go
jiceathome: Done.

router/underlayproviders/udpip/udpip.go
jiceathome: I was applying Go's convention (for a change) that a single method interface is named after the name of the method. Not my fault if that convention almost universally leads to bad grammar. Any suggestion that accomplishes the miracle of following the Go convention while being intelligible? Should we name it Connecter (per Go's English grammar) and the method be Connect()? Problem is that such vocabulary is terminally oversubscribed. ConnMaker/Make? I cannot call it "TestConnAnything", because its default implementation is used for real connections.

router/underlayproviders/udpip/udpip.go
jiceathome: So did I. At least, that one goes by Go's conventions: small scope, short name. Let's wait for ideas from other reviewers?

router/underlayproviders/udpip/udpip.go
jiceathome: Done.

router/underlayproviders/udpip/udpip.go
jiceathome: Done.

router/underlayproviders/udpip/udpip.go
jiceathome: Done.

router/underlayproviders/udpip/udpip.go
jiceathome: Done

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4737)* status: all files reviewed, 15 unresolved discussions (waiting on @JordiSubira, @katyatitkova, @lukedirtwalker, @marcfrei, and @romshark)

romshark (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4737#-:-OMHEZKbAONAMeyvq8Lt:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JordiSubira, @katyatitkova, @lukedirtwalker, and @marcfrei)

romshark:
*[`router/export_test.go` line 96 at r6](https://reviewable.io/reviews/scionproto/scion/4737#-OMMzem7C7QZ51cN_5GQ:-OMMzem7C7QZ51cN_5GR:bj4xc9i) ([raw file](https://github.com/scionproto/scion/blob/e200c3817078964a1c34e062a4a83516be3b490a/router/export_test.go#L96)):*
> ```Go
> // New returns a BatchConn as the udpip underlay might have. If the field conn is non-nil, then that
> // is what New returns. That enables tests to supply a specific BatchConn implementation. Else new
> // returns an instance of MockBatchConn that is just a place holder; calling any of the methods will
> ```

BTW  "placeholder", typo?

romshark (commented):
> Reviewed 2 of 4 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4737)* status: 27 of 28 files reviewed, 1 unresolved discussion (waiting on @jiceatscion, @JordiSubira, @katyatitkova, @lukedirtwalker, and @marcfrei)

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4737)* status: 26 of 28 files reviewed, all discussions resolved (waiting on @JordiSubira, @katyatitkova, @lukedirtwalker, @marcfrei, and @romshark)

___
*[`router/export_test.go` line 96 at r6](https://reviewable.io/reviews/scionproto/scion/4737#-OMMzem7C7QZ51cN_5GQ:-OMN1CPW6jtAHy1ANcfm:b-7rl0kd) ([raw file](https://github.com/scionproto/scion/blob/e200c3817078964a1c34e062a4a83516be3b490a/router/export_test.go#L96)):*

BTW  "placeholder", typo?
</blockquote></details>

done

romshark (approved):
:lgtm:

> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4737#-:-OMN7wJA2uyHpNcyaL5Y:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JordiSubira, @katyatitkova, @lukedirtwalker, and @marcfrei)

marcfrei (approved):
:lgtm:

> Reviewed 1 of 10 files at r1, 19 of 25 files at r2, 4 of 6 files at r3, 1 of 4 files at r5, 1 of 1 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4737#-:-OMrFflf3LkUw_Pd5YBE:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @JordiSubira, @katyatitkova, and @lukedirtwalker)

___
*[`router/underlayproviders/udpip/udpip.go` line 847 at r7](https://reviewable.io/reviews/scionproto/scion/4737#-OMrFAwY9NquXGkOYTCr:-OMrFAwY9NquXGkOYTCs:bq6ndt4) ([raw file](https://github.com/scionproto/scion/blob/61c3fca02dac1e7aee173e0efb52863d4bbf35d1/router/underlayproviders/udpip/udpip.go#L847)):*
> ```Go
> 	case addr.HostTypeIP:
> 		dstAddr = dst.IP()
> 		if dstAddr.Is4In6() {
> ```

Would it not be possible to convert this to a proper IPv4 address and to carry on? Maybe not worth the trouble?

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4737)* status: all files reviewed, 1 unresolved discussion (waiting on @JordiSubira, @katyatitkova, @lukedirtwalker, and @marcfrei)

___
*[`router/underlayproviders/udpip/udpip.go` line 847 at r7](https://reviewable.io/reviews/scionproto/scion/4737#-OMrFAwY9NquXGkOYTCr:-OMrPiZ382O30uVYgDqW:b-c2kq5i) ([raw file](https://github.com/scionproto/scion/blob/61c3fca02dac1e7aee173e0efb52863d4bbf35d1/router/underlayproviders/udpip/udpip.go#L847)):*

I guess we could. However I remember one of the specs (may be even the ietf draft) says that this is forbidden. I'll go see if the draft actually says that. ... something to do while I wait for ppl to scream about the topo fields.
</blockquote></details>

So, it was done as part of this PR https://github.com/scionproto/scion/pull/4579 specifically to address a pair of bug filed about not allowing 4in6 addresses. That particular bit of code was originally added to resolveLocalDst, but there are other similar checks in other places. I think it's best not to mess with that anymore. In this specific case, someone did put a v4mapped address on the wire (albeit a SCION Host, but still), so we should not oblige.

marcfrei (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4737)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JordiSubira, @katyatitkova, and @lukedirtwalker)

PR #4740, opened 2025-03-25
router: STUN implementation

jdslab:
Modification to border router dataplane to process STUN requests, as described in the [design document](https://github.com/scionproto/scion/blob/2755a60bbfe9e300abbebd9051d951345a2a2435/doc/dev/design/NAT-address-discovery.rst) and [discussion](https://github.com/scionproto/scion/issues/4630).


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4740)

jiceatscion (changes requested):
Sounds mostly good. I very much appreciated that you implemented the STUN detection as an error case and handle it on the Slow path. That makes the extra cost zero. Maybe you could consider also gating the feature with a command-line flag. I can imagine cases where users do not want to host the stun service, just to reduce attack surfaces no matter how small. Doc-wise, is the STUN/SCION headers overlap explained anywhere? The fact the STUN packet looks like a SCION packet with Next Header type 2 and Header length 1, needs to be memorialized somewhere if it hasn't been, and your code should refer to it, so that we never break that invariant accidentally.

> Reviewed 15 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4740#-:-OMNFJ7c0GwMmeHUEWZp:b-vnvjud)* status: all files reviewed, 4 unresolved discussions (waiting on @jdslab)

___
*[`demo/stun/test-client/main.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4740#-OMN57MAD_NJudBHt_uk:-OMN57MAD_NJudBHt_ul:b-ax1u0v) ([raw file](https://github.com/scionproto/scion/blob/68be7bde23be9acfe0862e8820853522206100ad/demo/stun/test-client/main.go#L1)):*
> ```Go
> package main
> ```

Copyright? License?
___
*[`demo/stun/test-server/main.go` line 1 at r1](https://reviewable.io/reviews/scionproto/scion/4740#-OMN5B_72s43hOX58Yhy:-OMN5B_72s43hOX58Yhz:b-89a4l7) ([raw file](https://github.com/scionproto/scion/blob/68be7bde23be9acfe0862e8820853522206100ad/demo/stun/test-server/main.go#L1)):*
> ```Go
> package main
> ```

Ditto
___
*[`demo/stun/test.py` line 4 at r1](https://reviewable.io/reviews/scionproto/scion/4740#-OMN4zOm6amM547H11Zk:-OMN4zOm6amM547H11Zl:bijuy0b) ([raw file](https://github.com/scionproto/scion/blob/68be7bde23be9acfe0862e8820853522206100ad/demo/stun/test.py#L4)):*
> ```Python
> 
> # Copyright 2024 ETH Zurich
> #
> ```

2025. It's new code, right? 
___
*[`demo/stun/test.py` line 114 at r1](https://reviewable.io/reviews/scionproto/scion/4740#-OMNEaTBC378pNXy23oa:-OMNEaTBC378pNXy23ob:bz7gwqg) ([raw file](https://github.com/scionproto/scion/blob/68be7bde23be9acfe0862e8820853522206100ad/demo/stun/test.py#L114)):*
> ```Python
> 
>     def _run(self):
>         time.sleep(10) # wait for everything to start up
> ```

How reliable is that time guess? Can we probe readiness instead? It is OK for a test take 10 seconds or even much more, but provided we don't have too many of those.

router/dataplane.go
```diff
@@ -798,6 +799,20 @@ func (p *slowPathPacketProcessor) processPacket(pkt *Packet) error {
 
 	p.lastLayer, err = decodeLayers(pkt.RawPacket, &p.scionLayer, &p.hbhLayer, &p.e2eLayer)
 	if err != nil {
+		// Check for and process STUN packet
+		if stun.Is(pkt.RawPacket) {
+			txid, err := stun.ParseBindingRequest(pkt.RawPacket)
+			if err != nil {
+				return serrors.Wrap("processing STUN packet", err)
+			}
+			resp := stun.Response(txid, pkt.SrcAddr.AddrPort())
```
jiceathome: We might need to make some changes to support this nowadays. The connected links do not set the packet's srcAddr because it is (until now) not used for anything. We might need to change that. Must the address really be repeated in the stun packet?

marcfrei: Unfortunately, yes. Sending back the underlay source address (as seen by the router) is at the core of how STUN works:

"The STUN server copies that source transport address into an XOR-MAPPED-ADDRESS attribute in the STUN Binding response and sends the Binding response back to the STUN client."

https://datatracker.ietf.org/doc/html/rfc8489#section-2

jiceathome: Duh! I'm stupid. Yeah, the address must make it back to the client somehow. Should we move the stun server directly into the underlay? After all, IP NAT *is* an underlay business. It would have no reason to exist without an  IP-based underlay at least. Either way. I can re-expose the src to the upper layer if need be.

marcfrei: Right, the underlay seems like the logical place for the STUN functionality, provided we can also achieve some degree of reuse across various UDP/IP-based underlay implementations. I'm not yet deep enough in the details of the new underlay structure to say whether this would work without hitting additional issues, though.

router/underlayproviders/udpip/udpip.go
```diff
@@ -748,17 +748,17 @@ func (l *detachedLink) receive(size int, srcAddr *net.UDPAddr, p *router.Packet)
...
-		log.Debug("Error while computing procID", "err", err)
-		l.pool.Put(p)
-		metrics[sc].DroppedPacketsInvalid.Inc()
-		return
-	}
 
 	p.Link = l
 	// The src address does not need to be recorded in the packet. The link has all the relevant
 	// information.
+
```
tzaeschke, line 755: It seems weird that this function is an almost exact duplicate of the function in line 571. I suppose this is out of scope for this PR?

demo/stun/test-client/main.go
```diff
@@ -0,0 +1,184 @@
...
+	"net/netip"
+
+	"github.com/scionproto/scion/pkg/addr"
+	"github.com/scionproto/scion/pkg/daemon"
+	"github.com/scionproto/scion/pkg/snet"
+
+	"tailscale.com/net/stun"
+)
+
+func main() {
```
tzaeschke, line 35: Maybe add something here like:
"This demo client uses manual STUN commands to demonstrate how STUN can be implemented with SCION.
Normal clients would not do that and instead use a library that performs STUN automatically and transparently."
?

jiceathome:
J-CIn router/underlayproviders/udpip/udpip.go:
>  
 	p.Link = l
 	// The src address does not need to be recorded in the packet. The link has all the relevant
 	// information.
+

It seems weird that this function is an almost exact duplicate of the function in line 571. I suppose this is out of scope for this PR?



Message ID: ***@***.***>
I guess this is referring to the many similarities between the link subtypes implementations. FWIW, I deliberately avoided using inheritance in order to retain the freedom of optimizing each link subtype. Sharing code between them would have encouraged forced commonalities or added the risk of breaking one when intending to change the behavior of the other. I think this flexibility is worth a small amount of redundancy. The link subtypes do not semantically extend each other. For example, in the af packet implementation (how about reviewing that PR?), the link subtypes have different relationships. So I did not want my code to serve as a misleading example to someone implementing a new underlay.Just my $.02 … that kid has left the nest now :-)

demo/stun/test-client/main.go
jdslab: Done.

router/underlayproviders/udpip/udpip.go
```diff
@@ -937,40 +1015,50 @@ func (l *internalLink) receive(size int, srcAddr *net.UDPAddr, p *router.Packet)
...
+// computation. If data is clearly not a valid SCION packet, it returns ok=false. Otherwise, it
+// returns a processor ID smaller than numProcRoutines and ok=true.
+func computeProcID(data []byte, numProcRoutines int, hashSeed uint32) (uint32, bool) {
 	if len(data) < slayers.CmnHdrLen {
-		return 0, errShortPacket
+		return uint32(numProcRoutines), false
+	}
+
+	switch slayers.L4ProtocolType(data[4]) {
+	case slayers.L4TCP, slayers.L4UDP, slayers.L4SCMP, slayers.L4BFD,
```
JordiSubira, line 1051: If we ever have a protocol type assigned to 0x21 we may identify STUN packets as SCION packets (depending on the result of the if below). We do not need to tackle this now, but documenting it in the function description will help if we ever come across this.

router/underlayproviders/udpip/udpip.go
```diff
@@ -937,40 +1015,50 @@ func (l *internalLink) receive(size int, srcAddr *net.UDPAddr, p *router.Packet)
...
+		slayers.HopByHopClass, slayers.End2EndClass,
+		253, 254 /* experimentation and testing */ :
```
JordiSubira: I would register these values as const in `slayers.l4.go` . I guess they haven't been needed in the implementation so far, but since they are already documented it would make more clear that the implementation is following what is expected in the doc for these values.

demo/stun/test.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # run tests (located in test-client/main.go and test-server/main.go)
+        self.dc.execute_detached("tester_1-ff00_0_110", "bash", "-c",
+                                 "test-server -local 1-ff00:0:110,172.20.0.22:31000")
+        time.sleep(3)
+        result = self.dc.execute(
+            "tester_1-ff00_0_111", "bash", "-c",
+            'test-client -daemon 192.168.123.3:30255 -local 1-ff00:0:111,192.168.123.4:31000 '
+            '-remote 1-ff00:0:110,172.20.0.22:31000 -data "abc"')
+        print(result)
```
JordiSubira, line 144: I do not see any ouput when running the test. This is likely because you are using `log` in `test-client`  and `test-server`. Go's log package writes to STDERR by default. `dc.execute()` only captures STDOUT (see docker.py:55 - return res.stdout). 

In addition, I would add an assertion since we have a fixed output that we are expecting `"abc"`.

demo/stun/README.md
```diff
@@ -0,0 +1,61 @@
+# STUN Demo
```
JordiSubira, line 1: I am a bit doubtful how we should better organize the integration test. I believe the main reason for having `demo` was having the file_transfer demo. Later, Matthias suggested that we could incorporate DRKey in  `demo` instead of acceptance, because we were taking it as an application running over SCION, rather than an inherent part of the SCION protocol (i.e., we do not need DRKey for SCION to work, actually DRKey needs SCION to work, at least as it is defined now). In the case of STUN, I do believe it is a necessary component for SCION to work properly. In any case, this is how I had understood the organization so far, but we can always restructure this.

marcfrei: Up till now, the intention of this demo was to show how to use STUN on an end host "manually" given that we have support for it in the border router. Once we have STUN support in snet, we can get rid of this demo application.

JordiSubira: Sounds good, we can leave it then in this PR.

router/underlayproviders/udpip/udpip.go
jdslab: Comment added

router/underlayproviders/udpip/udpip.go
jdslab: Done

router/dataplane_test.go
```diff
@@ -379,7 +379,7 @@ func TestDataPlaneRun(t *testing.T) {
 							// However, prepBaseMsg does pretend that there's a SCIONUDP Header.
 							// Remove that. Since the removal of the dispatcher, the router snoops
 							// into L4 and would mistake our payload for a broken SCION/UDP header.
-							spkt.NextHdr = slayers.L4None
+							spkt.NextHdr = 253 // use for experimentation and testing
```
marcfrei: `slayers.ExperimentationAndTesting`, then remove the comment

JordiSubira (approved):

marcfrei (approved):

demo/stun/test-client/main.go
```diff
@@ -0,0 +1,193 @@
...
+		srcAddr = srcAddr.Unmap()
+		srcPort = uint16(localAddr.Host.Port)
+		nextHop = remoteAddr.Host
+	} else {
+
+		// Generate and send STUN request
+		txID := stun.NewTxID()
+		req := stun.Request(txID)
+
+		var stunAddr = *nextHop
```
romshark, line 104: `sp.UnderlayNextHop()` may return `nil` yet the `nextHop == nil && remoteAddr.IA.Equal(localAddr.IA)` condition may still be false. Is this a potential nil pointer deref?

demo/stun/test-client/main.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/scionproto/scion/pkg/daemon"
+	"github.com/scionproto/scion/pkg/snet"
+
+	"tailscale.com/net/stun"
+)
+
+// This demo handles all STUN requests manually to demonstrate how STUN can be implemented in SCION.
+// Normal clients should use a client library that performs STUN automatically and transparently.
+
+func main() {
```
romshark, line 35: nit: Maybe consider refactoring this into a few smaller sub-functions instead of one big main func. This is just a suggestion, feel free to resolve without action.

demo/stun/test-client/main.go
```diff
@@ -0,0 +1,193 @@
+// Copyright 2025 ETH Zurich
```
romshark:
```suggestion
// Copyright 2026 ETH Zurich
```

Happy new year 🥳
Applies to all other files too.

demo/stun/test-server/main.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		if int(pld.DstPort) == localAddr.Host.Port {
+			log.Printf("Received data: \"%v\"", string(pld.Payload))
+
+			pkt.Destination, pkt.Source = pkt.Source, pkt.Destination
+
+			rp, ok := pkt.Path.(snet.RawPath)
+			if !ok {
+				log.Printf("Failed to reverse path, unecpected path type: %v", pkt.Path)
```
romshark:
```suggestion
				log.Printf("Failed to reverse path, unexpected path type: %v", pkt.Path)
```

router/underlayproviders/udpip/udpip.go
```diff
@@ -843,13 +846,88 @@ func (l *internalLink) start(
 	procQs []chan *router.Packet,
 	pool router.PacketPool,
 ) {
+	maxCap := 0
+	for _, q := range procQs {
+		if cap(q) > maxCap {
+			maxCap = cap(q)
+		}
```
romshark: nit:

```suggestion
		maxCap = max(maxCap, cap(q))
```

router/underlayproviders/udpip/udpip.go
```diff
@@ -937,40 +1015,54 @@ func (l *internalLink) receive(size int, srcAddr *net.UDPAddr, p *router.Packet)
...
 	if len(data) < slayers.CmnHdrLen {
-		return 0, errShortPacket
+		return uint32(numProcRoutines), false
+	}
+
+	switch slayers.L4ProtocolType(data[4]) {
+	case slayers.L4TCP, slayers.L4UDP, slayers.L4SCMP, slayers.L4BFD,
+		slayers.HopByHopClass, slayers.End2EndClass,
+		slayers.ExperimentationAndTesting, slayers.ExperimentationAndTesting2:
+		break
```
romshark:
```suggestion
```

this `break` is superfluous

.golangci.yml
```diff
@@ -91,6 +91,15 @@ linters:
...
           - goheader
         path: pkg/scrypto/cms
+      - linters:
+          - goheader
+        path: pkg/stun/stun.go
+      - linters:
+          - goheader
+          - ineffassign
+          - lll
+          - staticcheck
```
romshark: Please, do not disable these linters.

```suggestion
          - goheader
          - lll
```

Instead, when you don't need a value, like here:

```go
	gotTx, err = stun.ParseBindingRequest(req)
	if err == nil {
		t.Error("Expected error \"malformed STUN packet\". No error thrown.")
	}

	tx = newTxID()
	req = requestWithoutFingerprint(tx)
	gotTx, err = stun.ParseBindingRequest(req)
	if err == nil {
		t.Error("Expected error \"STUN request didn't end in fingerprint\". No error thrown.")
	}
```

Just ignore it explicitly:

```go
_, err = stun.ParseBindingRequest(req)
```

romshark (changes requested):
LGTM, but disabling `ineffassign` and `staticcheck` is wrong.

go.mod
```diff
@@ -57,6 +57,7 @@ require (
 	google.golang.org/protobuf v1.36.6
 	gopkg.in/yaml.v3 v3.0.1
 	modernc.org/sqlite v1.40.1
+	tailscale.com v1.84.3
```
romshark, line 60: Consider upgrading to latest:

https://pkg.go.dev/tailscale.com@v1.92.4

jdslab: The latest version of tailscale uses go version 1.25.5 but we are still on version 1.24.2. Given that this dependency will disappear once we merge the snet modifications and get rid of the current integration test, I think we can keep this as is.

romshark: Yes, we should really bump our version to 1.25.5 ASAP :)
Let's keep it as is then for now.

romshark (approved):
LGTM

PR #4745, opened 2025-03-31
router: refactor phase 5

jiceatscion:
This is the last of the router refactoring series. It cashes in on the previous changes by implementing connected sibling links, reduce address manipulations, make address copies unnecessary, and finish making underlay addresses opaque to the main router code. In passing, add other simplifications that weren't so obvious before: eliminate most map lookups; notably related to metrics.

Unrelated carry-overs from the previous PR: new name connOpener to replace connNewer and document deletion of the Bind and Public address fields.

In passing: moved the Lint and Check-generated tests out of the CI critical path. I'm tired of not getting tests results because lint complained.

Fixes: #4593


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4745)

private/topology/json/json.go
```diff
@@ -118,10 +118,9 @@ type BRInterface struct {
 
 // Underlay is the underlay information for a BR interface.
 type Underlay struct {
-	Local            string `json:"local,omitempty"`
-	DeprecatedBind   string `json:"bind,omitempty"`   // superseded by "local", for backwards compat
```
romshark: Did SCIONLab already fix their use of these deprecated fields?

router/underlay.go
```diff
@@ -61,9 +66,29 @@ type Link interface {
...
+	// redirect. Not all underlays have to provide that service and it might not be meaningful for a
+	// non-ip underlay. In such cases, this method simply has no effect.
+	SetDispatchPorts(start, end, redirect uint16)
+
+	// AddSvc adds the address for the given service. This can be called multiple times per service.
+	AddSvc(svc addr.SVC, a addr.Host, p uint16) error
+
+	// DelSvc deletes the address for the given service.
+	DelSvc(svc addr.SVC, a addr.Host, p uint16) error
+
```
romshark: I can deduce from the type and position that it's the port but spelling it out wouldn't hurt

private/underlay/conn/conn_linux.go
```diff
@@ -0,0 +1,169 @@
...
+// socket when taking to sibling routers.
+func (cc *connUDPBase) initConnUDP(
+	network string,
+	laddr, raddr netip.AddrPort,
+	cfg *Config) error {
+
+	var c *net.UDPConn
+	var err error
+	if !laddr.IsValid() {
+		return serrors.New("listen address must be specified")
```
romshark, line 46: Consider making this a sentinel error.

jiceathome: Considered and rejected. This is the only use of such an error, there isn't code trying to identify it specifically (that I know of) and adding a sentinel error would require an additional serrors.Wrap to include the stack trace and specific message here.

router/underlay.go
jiceathome: Done. Changed signatures to reflect that the API takes a SCION host and a SCION port.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4745)* status: 0 of 36 files reviewed, 3 unresolved discussions (waiting on @romshark)

private/topology/json/json.go
jiceathome: They plan on doing that when they next sync with master.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4745)* status: 0 of 36 files reviewed, 3 unresolved discussions (waiting on @romshark)

romshark:
*[`router/underlayproviders/udpip/udpip.go` line 365 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0mpd66b_LYC9wGBAC:-ON0mpd66b_LYC9wGBAD:b5071wh) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/underlayproviders/udpip/udpip.go#L365)):*
> ```Go
> 			msgs[i].Buffers[0] = p.RawPacket
> 			msgs[i].Addr = nil
> 			// If we're using a connected socket we must not specify the address. It might causes
> ```

it might "cause" ?

romshark:
*[`router/underlayproviders/udpip/udpip.go` line 430 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0nC7B9v8onaaYOCEX:-ON0nC7B9v8onaaYOCEY:b-j3u4m4) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/underlayproviders/udpip/udpip.go#L430)):*
> ```Go
> }
> 
> // NewExternalLink returns an external link over the UdpIpUnderlay. It is always implemented with a
> ```

nit: Shouldn't this be "UDPIPUnderlay"?

romshark:
*[`router/underlayproviders/udpip/udpip.go` line 582 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0vGR1EANcv8STxtUc:-ON0vGR21IlPSuaZ5_3v:b-r387zt) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/underlayproviders/udpip/udpip.go#L582)):*
> ```Go
> 
> 	p.Link = l
> 	// The src address does not need to be recorded in the packet. The link has all the relevant
> ```

nit: That's a bit weird, why not move the entire sentence to the next line?

romshark:
*[`router/underlayproviders/udpip/udpip.go` line 606 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0v_v99vBb08ehFYMN:-ON0v_v99vBb08ehFYMO:b63ob71) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/underlayproviders/udpip/udpip.go#L606)):*
> ```Go
> }
> 
> // newSiblingLink returns a sibling link over the UdpIp underlay. It may be implemented with either
> ```

Most of the time it's referred to as "udpip underlay". Maybe use it everywhere for consistency.

romshark:
*[`router/underlayproviders/udpip/udpip.go` line 647 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0vwZh4WjntNftkdpS:-ON0vwZh4WjntNftkdpT:bptkffv) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/underlayproviders/udpip/udpip.go#L647)):*
> ```Go
> 
> func (u *provider) newDetachedLink(
> 	qSize int,
> ```

qSize appears to be unused.

romshark:
*[`router/dataplane.go` line 126 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0xXoq4gmAcg3KlyvQ:-ON0xXoq4gmAcg3KlyvR:bwpywer) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane.go#L126)):*
> ```Go
> 	// The source address during ingest and the destination during forwarding. We never need both
> 	// src and dst at the same time. The real type is only known to underlay provider that sets it.
> 	RemoteAddr *struct{}
> ```

good old `void*`

romshark:
*[`router/dataplane.go` line 126 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0xXoq4gmAcg3KlyvQ:-ON0xj5L2YywyLai2iYT:b-66sw2c) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane.go#L126)):*

good old `void*`
</blockquote></details>

into the untyped wild west we ride 🤠

romshark:
*[`router/dataplane.go` line 178 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0y2nl2E314wsQw-Kk:-ON0y2nl2E314wsQw-Kl:b6eevi1) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane.go#L178)):*
> ```Go
> 	numInterfaces       int
> 	linkTypes           [65536]topology.LinkType
> 	neighborIAs         [65536]addr.IA
> ```

consider making this a named constant, maybe `MaxPorts` 🤷🏻‍♂️

romshark:
*[`router/dataplane_internal_test.go` line 113 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON13DYe6NkhHw4V_X-j:-ON13DYe6NkhHw4V_X-k:bvn5m0m) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane_internal_test.go#L113)):*
> ```Go
> 				t,
> 				net.UDPAddr{IP: net.IP{10, 0, 200, 200}},
> 				*(*net.UDPAddr)(unsafe.Pointer(pkt.RemoteAddr)),
> ```

Consider making this a type instead of using unsafe.Pointer directly https://go.dev/play/p/TlWpf5F2Vu8

romshark:
*[`router/dataplane_test.go` line 456 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON13iKS0-ACgT_KU4pI:-ON13iKS0-ACgT_KU4pJ:b-baj479) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane_test.go#L456)):*
> ```Go
> 					return buffer.Bytes()
> 				}
> 				_ = ret.SetKey([]byte("randomkeyformacs"))
> ```

any good reason this error is ignored?

romshark:
*[`router/export_test.go` line 111 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON14_1DDZHbVBgt4RF0:-ON14_1DDZHbVBgt4RF1:b-41i831) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/export_test.go#L111)):*
> ```Go
> 
> // We let the udpip underlay create distinct connections for sibling links as sharing a single
> // mock connection between internal and sibling links obscures tests.
> ```

Shouldn't this comment move inside the function? It appears to me it's describing not the method but why the method is always returning true

romshark:
*[`router/underlay.go` line 64 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON15RTG0856XPU0dF4Z:-ON15RTG0856XPU0dF4_:brlctj6) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/underlay.go#L64)):*
> ```Go
> 	// Send queues the packet for sending over this link; discarding if the queue is full.
> 	Send(p *Packet) bool
> 	// Send queues the packet for sending over this link; blocking while the queue is full.
> ```

SendBlocking



_Code quote:_
```Go
Send
```

romshark (commented):
> Reviewed 4 of 31 files at r1, 14 of 30 files at r3, 1 of 2 files at r4, 10 of 11 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4745)* status: 29 of 36 files reviewed, 13 unresolved discussions (waiting on @jiceatscion and @marcfrei)

romshark:
*[`router/dataplane_internal_test.go` line 113 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON13DYe6NkhHw4V_X-j:-ONZiDyj46qzmW83OWul:b620gvw) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane_internal_test.go#L113)):*

Uh, ok. Why not... but why?
</blockquote></details>

After discussing this yesterday I think you're right and we should not wrap it in a type. However I would still consider just keeping it an `unsafe.Pointer` instead of `*struct{}`.

romshark:
*[`router/dataplane_internal_test.go` line 113 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON13DYe6NkhHw4V_X-j:-ONZiiVH3FEI2wzI8iLH:b-ljq0ig) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane_internal_test.go#L113)):*

After discussing this yesterday I think you're right and we should not wrap it in a type. However I would still consider just keeping it an `unsafe.Pointer` instead of `*struct{}`.
</blockquote></details>

for everyone else: the discussion concluded that underlays may use any exotic address type you can think of and there isn't really a list of address types we could pre-define. The router doesn't need to and should not know of the underlay address type anyway.

romshark:
*[`router/dataplane.go` line 126 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0xXoq4gmAcg3KlyvQ:-ONZjIvzDukR73yGk2xc:b-hejl75) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane.go#L126)):*

No way around that one; short of adding an interface. I tried; it's even uglier. Yeeha!
</blockquote></details>

will mention it here too: Consider storing just the `unsafe.Pointer` instead of `*struct{}`

romshark:
*[`router/dataplane.go` line 178 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0y2nl2E314wsQw-Kk:-ONZjWQy1krbJGMdTZVM:b-9fxl5u) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane.go#L178)):*

There's one: math.MaxUint16 + 1
</blockquote></details>

I though a more explicit name like "MaxPorts" might be a bit more readable but I'm fine with this too.

jiceatscion (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4745)* status: 22 of 36 files reviewed, 11 unresolved discussions (waiting on @marcfrei and @romshark)

___
*[`router/dataplane.go` line 126 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0xXoq4gmAcg3KlyvQ:-ON_JpkE7wqKxujrnZ2g:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane.go#L126)):*

will mention it here too: Consider storing just the `unsafe.Pointer` instead of `*struct{}`
</blockquote></details>

Done.
___
*[`router/dataplane.go` line 178 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON0y2nl2E314wsQw-Kk:-ON_Jrgv0eb3ecXE7Ljs:b-zbg404) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane.go#L178)):*

I though a more explicit name like "MaxPorts" might be a bit more readable but I'm fine with this too.
</blockquote></details>

Since you find it acceptable, I prefer this. It makes the array big enough in all cases. No mis-configuration possible.
___
*[`router/dataplane_internal_test.go` line 113 at r5](https://reviewable.io/reviews/scionproto/scion/4745#-ON13DYe6NkhHw4V_X-j:-ON_JmmpEIrkzBEHqUdJ:b-896fix) ([raw file](https://github.com/scionproto/scion/blob/836fecb5952fc434955ecfe8e04f175198ae4578/router/dataplane_internal_test.go#L113)):*

for everyone else: the discussion concluded that underlays may use any exotic address type you can think of and there isn't really a list of address types we could pre-define. The router doesn't need to and should not know of the underlay address type anyway.
</blockquote></details>

Done.

romshark:
*[`router/underlay.go` line 64 at r7](https://reviewable.io/reviews/scionproto/scion/4745#-ON__w3p9E4cc5YgVjYJ:-ON__w3q0qJKSphPaP-s:bb7phvt) ([raw file](https://github.com/scionproto/scion/blob/4bbc3931fb2ab5ad8d53159e112e2612f5daea16/router/underlay.go#L64)):*
> ```Go
> 	// Send queues the packet for sending over this link; discarding if the queue is full.
> 	Send(p *Packet) bool
> 	// SendBlockingh queues the packet for sending over this link; blocking while the queue is full.
> ```

typo

romshark (approved):
> Reviewed 5 of 30 files at r3, 1 of 2 files at r4, 1 of 11 files at r5, 2 of 7 files at r6, 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4745)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion and @marcfrei)

romshark (approved):
aside from the typo :lgtm:

> Reviewed 1 of 1 files at r8, 4 of 4 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4745#-:-OO2pQWz8prIccfuYUeP:b-7osnbk)* status: all files reviewed, 1 unresolved discussion (waiting on @jiceatscion)

romshark (approved):
:lgtm:

> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4745#-:-OO3Av-ADABuKo5lZZrV:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiceatscion)

Issue #4742, opened 2025-03-25
control: automate detection of segment metadata


tzaeschke:
## Summary
SCION should provide tools to collect and specify metadata.

This is for now only a loose collection of ideas. Feel free to add more ideas or discuss listed ones.

## Motivation
Path metadata (bandwidth, latency geolocation, notes, hopcount, ...) is at the core of the SCION promise for path selection.
Providing metadata should make paths selection much more attractive for users.

## Ideas
* `staticInfoConfig.json` templates: simple tool to create template from a topofile
* Reloading updated `staticInfoConfig.json` files: make sure that the control server can (automatically?) reload updated files
* Latency: automate latency measurements (maybe border routers can/should perform measurements?)
* Internal hopcount: automate measurement of internal hopcount. Use `traceroute` as first approximation?
* Bandwidth: may be automatable, e.g. for AWS or Equinix APIs
* Geolocation: Use IP geolocator for border router IPs as first approximation?

tzaeschke:
Moved to proper proposal in #4761

tzaeschke:
Reopened issue to simplify planning and prioritization. Progress is still tracked in the PR #4761.

PR #4761, opened 2025-05-02
doc: proposal for metadata discovery

tzaeschke:
This design document proposed an approach to improve availability of path metadata.
Initial discussion (very short): #4742

### DIFF
--- a/doc/dev/design/metadata-discovery.rst
+++ b/doc/dev/design/metadata-discovery.rst
@@ -137,6 +137,7 @@ we should probably remove the runtime reparsing of the file. Reparsing of the
 file would thus be an interim solution until the management API is available.
 At that point, the file should only be parsed at startup of the metadata service.
 
+
 Rationale
 =========
 


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4761)

tzaeschke (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4761)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @jiceatscion)

___
*[`doc/dev/design/metadata-discovery.rst` line 60 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGlxtN3YeSXwOZE8_l:-OPVXwik6JA9HeQLQIzj:b9704c5) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L60)):*

I would tend to react with "here comes trouble". If an API can update the data, then do these updates also result in writing to the file? What if someone had just modified it? Are the changes lost in the collision? I think it might be best to chose whether the file is for persistence, transmission or input.
</blockquote></details>

Yes that may cause problems.  However, they could also use the API in parallel, or they could work on config file in parallel (one editing it, another copying over shortly afterwards with a locally edited copy,

Not sure how to protect against uncoordinated admin.

However, this is just a proposal, we could also do away with reparsing an updated file? It really depends on whether we think admins prefer editing a file or using an API.

Thoughts?
___
*[`doc/dev/design/metadata-discovery.rst` line 75 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGntOyAJzuixe0nb4N:-OPVUk6DB9L487ZZeQRM:b-x52gbo) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L75)):*

"Store updates to the"?
</blockquote></details>

Thanks, fixed.
___
*[`doc/dev/design/metadata-discovery.rst` line 81 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGn8Uf--t7LylKxwN9:-OPVZyPG7AaqhTEyiyo0:bm07rma) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L81)):*

At the scale of a Tier 1 operator, that might not be realistic. Unless that's for data that changes extremely rarely.
</blockquote></details>

I was primarily thinking about detecting conflicts in the topology file vs the staticInfoConfig file. I clarified this.
I assume changes should be rare, not sure how often an AS adds links or border routers?
___
*[`doc/dev/design/metadata-discovery.rst` line 82 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGnnoG-RUAtKuJHgki:-OPV_zCMDgxkZ3sCPgQR:b-zjh6r) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L82)):*

adopted?
</blockquote></details>

I think I mean adapted, the metadata is adapted to fit the modified topology...
___
*[`doc/dev/design/metadata-discovery.rst` line 121 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGoQXP5iPTJYAKhNOZ:-OPVazqH0taeO2t2e2PZ:b-jnsir8) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L121)):*

If that file is used to convey manual overrides, you would have to define (possibly in another file) which items are manually overridden and so must not be updated from the censor values.
</blockquote></details>

My idea was as follow, but I realize now it is broken, so just for reference: Overriding was not the intention. With "non-measurable data" I meant data the the control server cannot get anywhere else but from what the admin provides. I see that this is a bit problematic because for some data the source of truth is the control server (measured data: latency, hop count), for other data (addresses, notes) it is whatever the admin says.

It is broken because some data needs to be overridden. For example, the geolocation (long/lat) can be derived from a geolocation service but should obviously be overridable. It may also be desirable to override other measurements, e.g. latency.

Request for discussion: How should we handle this?
- We could, as you suggest, specify which attributes have manual values and which may contain measurements.
  - At which granularity? Is it sufficient to specify a section that specify overridability per data type? E.g. `override_geolocation = true`; `override_latency=false`?
  - should we add an optional flag to each value, e.g. `override=true`?
  - As you suggested, should we put override info into a separate file? I assume you mean the actual _values_, not a flag that indicates overrides?

Suggetions?
___
*[`doc/dev/design/metadata-discovery.rst` line 143 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGp9jTAklB2Lln_bdv:-OPVg88_5PjyYMvtm37H:b-3fxkwd) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L143)):*

Not sure I understand what you mean: if these defaults are found, does it not mean that the data was not generated at all?
</blockquote></details>

I'm not sure I understand. I mean that if the default values are found it means that the data _was_ (auto) generated.
The default values are all values that are unlikely to occur due to measurements. For non-measurable text values,
the text could just contain "(auto generated)".

Maybe I misunderstand the question?
___
*[`doc/dev/design/metadata-discovery.rst` line 162 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGpT-c3ptPr-gD7EzK:-OPVh2K-A-USMtsfOyAW:bfhblz3) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L162)):*

I think that distinction is somewhat fictitious. The two things can live in the same server and still fail to cooperate. So, monitoring what the CS is using would always be needed, independently of what the MS is doing.
</blockquote></details>

Yes, but if the metadata service is in the control service, we wouldn't need a separate API because they are the same.
Maybe the misunderstanding is that with "separate" I mean "separate process"?

tzaeschke (commented):
Thanks for the feedback! :)

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4761#-:-OPVi8Fq16bEy0p_VLOJ:b8dg4cr)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @jiceatscion)

jiceatscion (changes requested):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4761)* status: all files reviewed, 5 unresolved discussions (waiting on @tzaeschke)

___
*[`doc/dev/design/metadata-discovery.rst` line 60 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGlxtN3YeSXwOZE8_l:-OQe19q111sFTkbhCJaI:b-i504o9) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L60)):*
<details><summary><i>Previously, tzaeschke (Tilmann) wrote…</i></summary><blockquote>

Yes that may cause problems.  However, they could also use the API in parallel, or they could work on config file in parallel (one editing it, another copying over shortly afterwards with a locally edited copy,

Not sure how to protect against uncoordinated admin.

However, this is just a proposal, we could also do away with reparsing an updated file? It really depends on whether we think admins prefer editing a file or using an API.

Thoughts?
</blockquote></details>

Ok. It makes sense to persist the configuration (after API-driven changes) into a human readable file. That being the case, there's no legitimate way to prevent someone from editing it while the server is down. So, I guess we shouldn't go out of our way to prevent that. What I would do however is to make it clear (through doc, and in a heading comment in the file), that  the file is read only upon startup and is liable for overwrite as long as the config API is active.
___
*[`doc/dev/design/metadata-discovery.rst` line 121 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGoQXP5iPTJYAKhNOZ:-OQe2rKt2p3WKs8tDLbj:bz5kuvy) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L121)):*
<details><summary><i>Previously, tzaeschke (Tilmann) wrote…</i></summary><blockquote>

My idea was as follow, but I realize now it is broken, so just for reference: Overriding was not the intention. With "non-measurable data" I meant data the the control server cannot get anywhere else but from what the admin provides. I see that this is a bit problematic because for some data the source of truth is the control server (measured data: latency, hop count), for other data (addresses, notes) it is whatever the admin says.

It is broken because some data needs to be overridden. For example, the geolocation (long/lat) can be derived from a geolocation service but should obviously be overridable. It may also be desirable to override other measurements, e.g. latency.

Request for discussion: How should we handle this?
- We could, as you suggest, specify which attributes have manual values and which may contain measurements.
  - At which granularity? Is it sufficient to specify a section that specify overridability per data type? E.g. `override_geolocation = true`; `override_latency=false`?
  - should we add an optional flag to each value, e.g. `override=true`?
  - As you suggested, should we put override info into a separate file? I assume you mean the actual _values_, not a flag that indicates overrides?

Suggetions?
</blockquote></details>

What we could do if have configuration keys dedicated to overrides. For example. geolocation = <autogenerated value> and geolocation_override = <admin supplied>. If the latter is present, then it applies and it cannot be auto-modified. We could define those keys for cases that make sense, and not for those that are only editable defaults; those that must be modifiable at run-time. If we find some third case (I don't know, something like an override that can be ignored in some cases), we can come up with some other suffix like "_preferred" for example. ...Aaanyway. it's minor stuff. We can figure it out as we go.
___
*[`doc/dev/design/metadata-discovery.rst` line 143 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGp9jTAklB2Lln_bdv:-OQe4ozG18TqCDykGVyq:bz3gst3) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L143)):*
<details><summary><i>Previously, tzaeschke (Tilmann) wrote…</i></summary><blockquote>

I'm not sure I understand. I mean that if the default values are found it means that the data _was_ (auto) generated.
The default values are all values that are unlikely to occur due to measurements. For non-measurable text values,
the text could just contain "(auto generated)".

Maybe I misunderstand the question?
</blockquote></details>

Well, it's hard for me to point out at what is wrong with the language (or with me). I just do not understand what you're trying to say here. It seems contradictory. How could some data have been autogenerated if it cannot be calculated? I think what you are trying to say is that the filed should say something to the effect that the data is meant to be autogenerated but has not been determined yet. Is that what you meant? At least that is what your example reflects; that, I understand.
___
*[`doc/dev/design/metadata-discovery.rst` line 162 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGpT-c3ptPr-gD7EzK:-OQe5kcI9RmKZR3vfmBr:b-duvssm) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L162)):*
<details><summary><i>Previously, tzaeschke (Tilmann) wrote…</i></summary><blockquote>

Yes, but if the metadata service is in the control service, we wouldn't need a separate API because they are the same.
Maybe the misunderstanding is that with "separate" I mean "separate process"?
</blockquote></details>

Well, I guess the API separation or not is the real concern. It seems to make a great deal more sense for a segment metadata API to be part of the segment API, rather than separate. That would be the main reason to put it all in the same bag. Adding more functionality to the CS is sort of  a downside. It's already big. But hey, if it's for a good reason; so be it.
___
*[`doc/dev/design/metadata-discovery.rst` line 18 at r3](https://reviewable.io/reviews/scionproto/scion/4761#-OQe0pxvBF7-zmvMkgVw:-OQe0pxvBF7-zmvMkgVx:bb7fgkq) ([raw file](https://github.com/scionproto/scion/blob/5c986a4dfe475d56fc41c6d3f2134ee6c1641a0b/doc/dev/design/metadata-discovery.rst#L18)):*
> ```reStructuredText
> 
> We believe that this is an important next step because metadata is essential
> We believe that this is an important next step because metadata is essential
> ```

echo

tzaeschke (commented):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4761)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @jiceatscion)

___
*[`doc/dev/design/metadata-discovery.rst` line 60 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGlxtN3YeSXwOZE8_l:-OQiKifs8qsyeowLvUnp:b-m1g3e8) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L60)):*

Ok. It makes sense to persist the configuration (after API-driven changes) into a human readable file. That being the case, there's no legitimate way to prevent someone from editing it while the server is down. So, I guess we shouldn't go out of our way to prevent that. What I would do however is to make it clear (through doc, and in a heading comment in the file), that  the file is read only upon startup and is liable for overwrite as long as the config API is active.
</blockquote></details>

I added a sentence to that effect in the section `Management API`: 
`If we decide to have a remote monitoring API, in order to avoid concurrency issues  
we should probably remove the runtime reparsing of the file. Reparsing of the  
file would thus be an interim solution until the management API is available.  
At that point, the file should only be parsed at startup of the metadata service.`
___
*[`doc/dev/design/metadata-discovery.rst` line 121 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGoQXP5iPTJYAKhNOZ:-OQiGmvc0stdBlI_Pu2s:b-ly38rl) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L121)):*

What we could do if have configuration keys dedicated to overrides. For example. geolocation = <autogenerated value> and geolocation_override = <admin supplied>. If the latter is present, then it applies and it cannot be auto-modified. We could define those keys for cases that make sense, and not for those that are only editable defaults; those that must be modifiable at run-time. If we find some third case (I don't know, something like an override that can be ignored in some cases), we can come up with some other suffix like "_preferred" for example. ...Aaanyway. it's minor stuff. We can figure it out as we go.
</blockquote></details>

I added a section `File Format` that proposes an `override` attribute for all fields. However, I am currently not sure what the exact usecase would be, it seems most cases can be covered without such an attribute....?
___
*[`doc/dev/design/metadata-discovery.rst` line 143 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGp9jTAklB2Lln_bdv:-OQiGMM-8TiA5dENbnWJ:b-938ddg) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L143)):*

Well, it's hard for me to point out at what is wrong with the language (or with me). I just do not understand what you're trying to say here. It seems contradictory. How could some data have been autogenerated if it cannot be calculated? I think what you are trying to say is that the filed should say something to the effect that the data is meant to be autogenerated but has not been determined yet. Is that what you meant? At least that is what your example reflects; that, I understand.
</blockquote></details>

I removed that block, I'm not sure anymore why I thought `Completeness` was is desirable. 
If there is no value then that is just that.
___
*[`doc/dev/design/metadata-discovery.rst` line 162 at r2](https://reviewable.io/reviews/scionproto/scion/4761#-OPGpT-c3ptPr-gD7EzK:-OQiFW9TD-korJ900zh2:b-diicxl) ([raw file](https://github.com/scionproto/scion/blob/399f767001b4b5a7d7869bb44350c2d3bd8fd28e/doc/dev/design/metadata-discovery.rst#L162)):*
> Adding more functionality to the CS is sort of a downside.

Indeed :)
I summarized that loosely under `Disadvantages`: `Feature overload of the control service`
___
*[`doc/dev/design/metadata-discovery.rst` line 18 at r3](https://reviewable.io/reviews/scionproto/scion/4761#-OQe0pxvBF7-zmvMkgVw:-OQiGH4n26kX53oL4yY1:bt2vr1b) ([raw file](https://github.com/scionproto/scion/blob/5c986a4dfe475d56fc41c6d3f2134ee6c1641a0b/doc/dev/design/metadata-discovery.rst#L18)):*

echo
</blockquote></details>

Fixed.

PR #4766, opened 2025-05-13
tests: let graphupdater generate multiple graphs

katyatitkova:
There's only one graph that is available for tests. It's generated with `graphupdater` tool which doesn't create the second graph properly because the same variable names might be used in the generated code. This PR fixes it, and adds one more graph that is an example of how the code with multiple graphs will look like. This graph will be used in the future for https://github.com/scionproto/scion/issues/4699 implementation.


lukedirtwalker (changes requested):
> Reviewed 17 of 20 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4766)* status: all files reviewed, 1 unresolved discussion (waiting on @katyatitkova)

___
*[a discussion](https://reviewable.io/reviews/scionproto/scion/4766#-OQgiL7y5iqSMrKKq4EV:-OQgiL7y5iqSMrKKq4EW:b-d3ir8h) (no related file):*
have you considered generating separate graph packages for each generated topology? then you wouldn't need to share any information between the different topologies?

katyatitkova:

have you considered generating separate graph packages for each generated topology? then you wouldn't need to share any information between the different topologies?
</blockquote></details>

Yes, I considered it. The reasons why I chose this approach are quite minor and might be just a matter of taste IMO.

1. We'll need some manual work with every new separate graph package, since we need a magic trick with static interface ids. Here, some manual work might be needed, but only if new interfaces are created. And even then it's less of a manual work than creating a new file with new interface ids.
2. In the future, I'll create a new test in propagator_test.go using this new topology. With two separate graph packages, the new test will use the new package, while the older tests will continue to use the old one. Similar thing might happen to other tests as well. This setup might be a bit more error-prone: someone (or their autocomplete) might accidentally use graph.If_110_X_210_X when graph_big.If_110_X_210_X is intended, for example.
3. I didn't want to copy pkg/private/xtest/graph/graph.go between packages. This one is definitely a matter of taste since if this file will be generated, there are no typical concerns over code duplication.

So, from my side, there are no strong reasons why this approach should be used and not the one with separate graphs. The only thing that bothers me in theory is a situation where we have 5-7 different graph packages, but it doesn't look like it happens anytime soon since everyone was using just one graph for years.

lukedirtwalker (approved):
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4766)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @katyatitkova)

PR #4774, opened 2025-05-22
fix: remove systemd from the docker tutorial setup

romshark:
Even though on macOS (surprisingly) the tutorial setup works just fine - on Linux `systemd` is causing problems when running inside Docker, presumably due to cgroups v2. Also, Docker Compose doesn't appear to support cgroup delegation while `--cgroupns=host` could be a potential workaround.

This PR removes systemd from the setup and instead runs the SCION services from a shell script.


jiceatscion:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4774)

katyatitkova (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4774#-:-OQs8w0G5V1aZ8vXo3O4:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @romshark)

FR4NK-W (approved):
> Reviewed 12 of 13 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4774)* status: all files reviewed, 1 unresolved discussion (waiting on @katyatitkova and @romshark)

___
*[`doc/tutorials/deploy/base/start-scion.sh` line 18 at r2](https://reviewable.io/reviews/scionproto/scion/4774#-ORMK9i57c9eX5iMNTOy:-ORMgbkY2aGD1v09O13k:bvsofjl) ([raw file](https://github.com/scionproto/scion/blob/d2b5eb7a8c2df837f2d27a9ebe2b393a80d39690/doc/tutorials/deploy/base/start-scion.sh#L18)):*
>As for the logging level: I'd, on the contrary, expect a lot of logging be shown, as for the tutorial's target audience it's an easy way to see that there is actually some stuff happening inside the containers.

I would find it confusing to see the logs of the different services interleaved, and it is also not representative of how logging looks like in an actual deployment (either separate containers or units).
For the purpose of the tutorial, I'd probably only want to see if the container (and the services started by the startup script) started successfully.
The tutorial should then direct me to the log files for the different services, let me explore the logs.

But maybe indeed for the tutorial the details of the logging are not so important.

romshark:
@FR4NK-W first I tried to make systemd run within the Docker container to keep the setup close to a real one but unfortunately this proved too difficult (couldn't make it work on Linux running Docker). Starting every service in its own Container could be an option, but it's a bit heavier in terms of configuration than just having 1 container per AS. So the log files are just a hack to compensate for the absence of systemd. We probably need to mention these log files in the tutorial documentation though.

romshark:
Maybe using Vagrant instead of Docker could be another option 🤷‍♂️ but I would rather avoid spending more time on this topic.

katyatitkova (approved):

PR #4780, opened 2025-06-06
tools/wireshark: add support for scion tcp

Wylwi:
Note: First time contributing to a big open-source project! I've read the contribution guide, but I'm open to suggestions/remarks/edits/criticisms and so on!

Hello!

This PR serves personal work. I was investigating SCION packets and realized that no support was offered for decrypting TCP over SCION packets. 

This PR implements it.

Please also note that the base script is the one from [NetSec ETHZ](https://github.com/netsec-ethz/scion/blob/scionlab/tools/wireshark/scion.lua) and thus that this PR includes work that is not mine, but it looks like that this was the most developed version of it and is the one featured on SCION's doc.

Before: 
image

image

After:
image

image

Thanks for your time!

### DIFF
--- a/tools/wireshark/scion.lua
+++ b/tools/wireshark/scion.lua
@@ -16,6 +16,7 @@ local addrTypes = {
 }
 
 local hdrTypes = {
+    [6] = "TCP",
     [17] = "UDP",
     [200] = "HOP_BY_HOP",
     [201] = "END_TO_END",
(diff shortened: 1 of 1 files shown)


jiceathome:
Should we take this to mean that, next, you intend to actually implement TCP on-top of SCION? That'd be really cool!

Wylwi:
Sadly, no, I don't have the knowledge for it. Maybe I misunderstood what TCP over Scion was. In my understanding, it was about transmitting original TCP/IP packets encapsulated in SCION UDP packets. This PR is only about decapsulating those TCP packets in Wireshark, to see the content of them. (Decapsulating SCION packets with NextHdr == 6)

Wylwi:
I've just read some of the dispatcher code. Is it even possible to use TCP payload in SCION UDP packets? Reading the doc I thought so, but reading the code I'm now unsure... Is it designed but not implemented?

jiceathome:
> Sadly, no, I don't have the knowledge for it. Maybe I misunderstood what TCP over Scion was. In my understanding, it was about transmitting original TCP/IP packets encapsulated in SCION UDP packets. This PR is only about decapsulating those TCP packets in Wireshark, to see the content of them. (Decapsulating SCION packets with NextHdr == 6)

That is totally fine. However, there will be no packets worth discecting with it. There isn't a single application or library that I know of that sends TCP over SCION traffic.

Wylwi:
I can't tell why for confidentiality reasons, but I'm working on something that also uses TCP over SCION, that's why I was interested in this and developing this script. But I see now that the current implementation doesn't handle it :sweat_smile:

jiceathome:
> I've just read some of the dispatcher code. Is it even possible to use TCP payload in SCION UDP packets? Reading the doc I thought so, but reading the code I'm now unsure... Is it designed but not implemented?

I don't remember seeing anything in the dispatcher code regarding TCP. Where, in the documentation, is it hinted that there might be such a thing? There isn't any current plan to implement TCP/SCION.
One thing that is very unlikely to ever happen is carrying TCP over UDP over SCION. That wouldn't serve any purpose. You might be thinking of TCP over SCION over UDP over IP; which would make sense, since that is how SCION traffic is carried in almost all deployments today: SCION is a network protocol that generally uses UDP over IP as its link layer.

Wylwi:
Yep, I was thinking about TCP over SCION over UDP over IP. But currently (or it's possible that I'm in the dark about something.), the Wireshark script was only able to decrypt UDP over SCION over UDP over IP. 

And it looks like from the dispatcher that TCP over SCION over UDP over IP is not supported. 
That would explain why my SYN TCP packet were dropped without any reason (and I can confirm that they issue a `ERROR	  dispatcher/dispatcher.go:170	Unexpected packet	{"layers decoded": 1}` in the dispatcher)

Wylwi:
> Where, in the documentation, is it hinted that there might be such a thing?

The [Assigned SCION Protocol Numbers ](https://docs.scion.org/en/latest/protocols/assigned-protocol-numbers.html#assigned-protocol-numbers) features a protocol number for TCP over SCION, which is the one I used in the Wireshark script. Thus, I thought it was working.

I just found out in [this design document about removing the dispatcher](https://docs.scion.org/en/latest/dev/design/router-port-dispatch.html) the following paragraph, which might confirm that there is no TCP over SCION support for now: 

> The SCION router in the destination AS delivers a packet to the end host at the specified destination address. The packet is delivered to the end host over UDP/IP, using a fixed end-host data port (30041). On the end host the packet is then delivered to the appropriate application socket determined by the protocol and port number; **only UDP/SCION (and SCMP) are currently supported.** We do not have operating system support for SCION on any platform, so this last step happens in user space, in the “dispatcher”.

lschulz:
The dispatcherless SCION stack has minimal support for TCP over SCION, see #4703 . It's enough to make [SCION-IP Translation](https://github.com/netsys-lab/scion-ip-translator) work.

It would be nice to have support for TCP over SCION over UDP in the Wireguard plugin.

jiceathome:
True, if the so-called TCP traffic comes from a native TCP/IP stack and is delivered to a native TCP/IP stack; being only tunneled through SCION; then there is in some way TCP traffic over SCION. Note that encapsulating the TCP packets directly as SCION payload is an arbitrary choice of whatever pair of tunnel ends is being used. Only the IP translator would do that, because it can be a one-ended tunnel (in the case of UDP - currently).

PR #4782, opened 2025-06-11
control: pluggable segment registration

### LINKED ISSUE #4726: proposal: pluggable segment registration
Authors: @lukedirtwalker, @oncilla

tl;dr: We propose a plugin system that allows the user to configure their custom
segment registration logic.

## Overview

Path registration is currently not customizable. By default all beacons that are
terminated in an AS are registered at the originating core AS. This means all
segments are public by default. This is not always desirable. Previously a
[hidden paths system](https://docs.scion.org/en/latest/hidden-paths.html) was
added. This hidden paths system allows an AS to decide who can get access to
some non-public path segments. However the approach requires additional
infrastructure and has not found wide-spread adoption.

In this proposal, we propose a more generic solution that allows flexible
segment registration policies to be implemented. To allow for this, we add a
plugin system that allows the user to configure their custom segment
registration logic. Plugins need to be compiled into the binary.

## Motivation

The current implementation of the SCION control service does not allow for any
different registration logic than what is currently implemented in the
opensource repository. We believe that by introducing a plugin system, we can
allow for experimentation with different registration policies.

As part of this proposal we want to implement a public registration plugin
(doing what the current public registration does) and a new hidden segment registration
plugin (see "Proposal: Hidden segments without extra infrastructure").

Apart from that, at Anapaya we have a specific desire to implement a custom
registration plugin.

## Proposal

We propose a plugin system. For this we extend the beacon policy to support
custom plugins. A plugin needs to be compiled into the binary. The details are
described below.

### Plugin API

A plugin is a Go struct that implements the following interface:

```go
type SegmentRegistrationPlugin interface {
    // ID returns the unique identifier of the plugin.
    ID() string
    // New creates a new instance of the plugin with the provided configuration.
    New(ctx context.Context, config map[string]any) (SegmentRegistrar, error)
    // Validate validates the configuration of the plugin.
    Validate(config map[string]any) error
}

type SegmentRegistrar interface {
    // RegisterSegments registers the given segments according to the plugin's logic.
    RegisterSegments(ctx context.Context, segments []*seg.PathSegment) (RegistrationStats, error)
}

type RegistrationStats struct {
    // Status contains the registration status for each segment.
    // If the segment was successfully registered, the error is nil.
    Status map[*seg.PathSegment]error
}
```

We need to define a plugin registry as follows:

```go
var segmentRegistrationPlugins = map[string]SegmentRegistrationPlugin{}

func RegisterPlugin(p SegmentRegistrationPlugin) {
    id := p.ID()
    if _, ok := segmentRegistrationPlugins[id]; ok {
        panic(fmt.Sprintf("plugin %q already registered", id))
    }
    segmentRegistrationPlugins[id] = p
}
```

A plugin must be constructed in the main code of the control service and
registered in a plugin registry. The plugin registry is a map from plugin name
to the SegmentRegistrationPlugin interface.

The plugin is then used depending on the beacon policy, which is described below.

We will need to change the implementation of the [SegmentWriters
method](https://github.com/scionproto/scion/blob/e14ce703acf48d950622133c13ae6557f8b49345/control/tasks.go#L125)
to dispatch based on the beacon policy to the different plugins as described
below.

### Beacon policy changes

Extend the
[beacon.Policy](https://github.com/scionproto/scion/blob/dd086a53c7722082a6c4b5dbf9cef57d5a549066/control/beacon/policy.go#L177-L191)
to support attaching a plugin configuration to a subset of beacons.

We propose that beacons are matched based on the combination of a `pathpol.ACL`
and `pathpol.Sequence` (similar to a path policy in the gateway). The operator
can express a list of mappings from matchers to "plugin configurations". All
beacons that do not match are registered with the default public registration
plugin.

Specifically we propose to extend the `beacon.Policy` with a list of
`RegistrationPolicies`. Each `RegistrationPolicy` consists of the following fields:

- `Name`: A descriptive name for the policy.
- `Description`: A description of the policy.
- `Matcher`: Is used to match the beacon, it contains the following fields:
    - `Sequence`: An optional sequence of `pathpol.Sequence` that need to match.
    - `ACL`: An optional `pathpol.ACL` that needs to match.
- `Plugin`: The plugin that should be used for this policy.
- `PluginConfig`: The configuration for the plugin, this is a key value map that
  is passed to the plugin.

An example of this could look as follows:

```yaml
---
Type: DownSegmentRegistration
BestSetSize: 6
CandidateSetSize: 20
MaxExpTime: 42
RegistrationPolicies:
  - Name: Hidden
    Description: Hide segments that end on interface 1 and only make them available to ASes 1-ff00:0:110 and 1-ff00:0:113
    Matcher:
      Sequence: "0* 0#1"
    Plugin: HiddenSegmentRegistration
    PluginConfig:
      AccessList:
        - 1-ff00:0:110
        - 1-ff00:0:113
  - Name: Custom
    Description: Custom registration logic for segments that end on interface 2
    Matcher:
      Sequence: "0* 0#2"
    Plugin: CustomSegmentRegistration
    PluginConfig:
      Registry: 1-ff00:0:112
```

utkn:
This PR implements pluggable segment registration as proposed in #4726. For backwards compatibility, the current functionality is moved under a default plugin that is initialized when the policy has no registration policies defined.

Closes #4726


control/registration.go
```diff
@@ -0,0 +1,287 @@
...
+)
+
+type SegmentRegistrationPlugin interface {
+	// ID returns the unique identifier of the plugin.
+	ID() string
+	// New creates a new instance of the plugin with the provided configuration.
+	New(
+		ctx context.Context,
+		t *TasksConfig,
+		segmentType seg.Type,
```
lukedirtwalker: IMO we should not need those extra parameters. They should be part of the configuration.

control/registration.go
```diff
@@ -0,0 +1,287 @@
...
+	return registrar, nil
+}
+
+type RegistrationStats struct {
+	// Status contains the registration status for each segment.
+	// The key is the segment ID, generated by seg.PathSegment.FullID().
+	// If the segment was successfully registered, the error is nil.
+	Status map[string]error
+	// WritetStats contains the write statistics from the underlying writer.
+	WriteStats beaconing.WriteStats
```
lukedirtwalker: I would only keep the status

control/registration.go
```diff
@@ -0,0 +1,287 @@
...
+	"github.com/scionproto/scion/pkg/private/serrors"
+)
+
+type SegmentRegistrationPlugin interface {
+	// ID returns the unique identifier of the plugin.
+	ID() string
+	// New creates a new instance of the plugin with the provided configuration.
+	New(
+		ctx context.Context,
+		t *TasksConfig,
```
lukedirtwalker: I think you need this argument for the extender. I think it doesn't make sense that the plugin has to extend (i.e. terminate) the beacon. Instead this could be done before actually handing the segment into the plugins logic.

This is why in the issue it is proposed to only have a `segments []*seg.PathSegment` argument.

lukedirtwalker (changes requested):

control/registration.go
utkn: Extension and filtering is now done by the `GroupWriter`.

control/registration.go
utkn: Done. After private discussion we have decided that `realMain` should construct the plugins with the correct parameters.

control/registration.go
utkn: Done. Registrars now only returns a registration summary (can be removed in the future)

control/beacon/policy.go
```diff
@@ -38,6 +42,61 @@ const (
...
+func (p RegPolicyType) SegmentType() segment.Type {
+	switch p {
+	case RegPolicyTypeUp:
+		return segment.TypeUp
+	case RegPolicyTypeDown:
+		return segment.TypeDown
+	case RegPolicyTypeCore:
+		return segment.TypeCore
+	default:
+		panic("unreachable: invalid registration policy type")
```
lukedirtwalker:
```suggestion
		panic("unreachable: invalid registration policy type: %s", p)
```

control/beacon/policy.go
```diff
@@ -38,6 +42,61 @@ const (
...
+func (p RegPolicyType) ToPolicyType() PolicyType {
+	switch p {
+	case RegPolicyTypeUp:
+		return UpRegPolicy
+	case RegPolicyTypeDown:
+		return DownRegPolicy
+	case RegPolicyTypeCore:
+		return CoreRegPolicy
+	default:
+		panic("unreachable: invalid registration policy type")
```
lukedirtwalker:
```suggestion
		panic("unreachable: invalid registration policy type: %s", p)
```

control/beacon/policy.go
```diff
@@ -202,16 +263,106 @@ func (p *Policy) InitDefaults() {
...
+type RegistrationPolicy struct {
+	Name         string                    `yaml:"Name"`
+	Description  string                    `yaml:"Description"`
+	Matcher      RegistrationPolicyMatcher `yaml:"Matcher"`
+	Plugin       string                    `yaml:"Plugin"`
+	PluginConfig map[string]any            `yaml:"PluginConfig"`
+}
+
+// InitDefaults initializes the default values for unset fields in the registration policy.
+func (p *RegistrationPolicy) InitDefaults() {
```
lukedirtwalker: move the closing bracket also up here.

control/beacon/policy.go
```diff
@@ -202,16 +263,106 @@ func (p *Policy) InitDefaults() {
...
+}
+
+// beaconAsPath implements snet.Path with respect to a Beacon.
```
lukedirtwalker, line 324: I would already indicate here that this type just implements a wrapper so that we can use pathpol filtering on beacons

control/beacon/policy.go
```diff
@@ -202,16 +263,106 @@ func (p *Policy) InitDefaults() {
...
+	return &beaconAsPath{
+		beacon: beacon,
+		Path:   nil,
+	}
+}
+
+// Metadata returns the metadata of the beacon.
+// It only sets the PathMetadata.Interfaces field.
+func (b *beaconAsPath) Metadata() *snet.PathMetadata {
+	md := snet.PathMetadata{}
```
lukedirtwalker:
```suggestion
	md := &snet.PathMetadata{}
```

control/beacon/policy.go
```diff
@@ -202,16 +263,106 @@ func (p *Policy) InitDefaults() {
...
+	md.Interfaces = make([]snet.PathInterface, 0)
```
lukedirtwalker: Simply drop this
```suggestion
```

control/beacon/store.go
```diff
@@ -94,15 +132,23 @@ func (s *Store) BeaconsToPropagate(ctx context.Context) ([]Beacon, error) {
 // SegmentsToRegister returns a channel that provides all beacons to register at
 // the time of the call. The selections are based on the configured policy for
 // the requested segment type.
-func (s *Store) SegmentsToRegister(ctx context.Context, segType seg.Type) ([]Beacon, error) {
+func (s *Store) SegmentsToRegister(
+	ctx context.Context, segType seg.Type,
```
lukedirtwalker:
```suggestion
	ctx context.Context,
	segType seg.Type,
```

control/beacon/store.go
```diff
@@ -164,12 +210,17 @@ func (s *CoreStore) BeaconsToPropagate(ctx context.Context) ([]Beacon, error) {
 
 // SegmentsToRegister returns a slice of all beacons to register at the time of the call.
 // The selection is based on the configured policy for the requested segment type.
-func (s *CoreStore) SegmentsToRegister(ctx context.Context, segType seg.Type) ([]Beacon, error) {
-
+func (s *CoreStore) SegmentsToRegister(
+	ctx context.Context, segType seg.Type,
```
lukedirtwalker:
```suggestion
	ctx context.Context,
	segType seg.Type,
```

control/beacon/store.go
```diff
@@ -58,6 +58,44 @@ func applyStoreOptions(opts []StoreOption) storeOptions {
...
+//
+// If the policy defines no registration policy, all the beacons will be put into
+// the default group, indexed by DEFAULT_GROUP.
+func groupBeacons(beacons []Beacon, policy *Policy) GroupedBeacons {
+	if len(policy.RegistrationPolicies) == 0 {
+		return map[string][]Beacon{
+			DEFAULT_GROUP: beacons,
+		}
+	}
+	// Go through every beacon, and group it into the first registration policy
```
lukedirtwalker: Hm I wonder if we should instead put it to every group that matches?
Why only the first?

control/beaconing/writer.go
```diff
@@ -45,7 +46,7 @@ type SegmentProvider interface {
 	// SegmentsToRegister returns the segments that should be registered for the
```
lukedirtwalker, line 45: given the comment mention the return type we should adapt it.

control/beaconing/writer.go
```diff
@@ -66,13 +67,24 @@ type WriteStats struct {
...
+	s.Count += other.Count
+	for ia := range other.StartIAs {
+		s.StartIAs[ia] = struct{}{}
+	}
+}
+
 // Writer writes segments.
 type Writer interface {
-	// Write writes passed slice of segments.Peers indicate the peering
+	// Write writes passed slice of segments. Peers indicate the peering
```
lukedirtwalker:
```suggestion
	// Write writes passed segments. Peers indicate the peering
```

control/beaconing/writer.go
```diff
@@ -366,3 +426,75 @@ func (l writerLabels) WithResult(result string) writerLabels {
...
+) (WriteStats, error) {
+	processedBeacons := w.processSegments(ctx, allBeacons, peers)
+	writeStats := WriteStats{Count: 0, StartIAs: make(map[addr.IA]struct{})}
+	for name, beacons := range processedBeacons {
+		registrar, err := w.Plugins.GetSegmentRegistrar(w.PolicyType, name)
+		if err != nil {
+			return WriteStats{}, serrors.Wrap("getting segment registrar", err,
+				"policy", w.PolicyType, "name", name)
+		}
+		sum := registrar.RegisterSegments(ctx, beacons, peers)
```
lukedirtwalker: I wonder if we should do this in a Go routine to avoid plugins blocking each other.

pkg/experimental/hiddenpath/beaconwriter.go
```diff
@@ -241,6 +247,184 @@ func (l writerLabels) WithResult(result string) writerLabels {
...
+}
+
+type HiddenSegmentRegistrar struct {
+	HiddenSegmentRegistrationPlugin
+	InternalErrors metrics.Counter
+}
+
+var _ registration.SegmentRegistrar = (*HiddenSegmentRegistrar)(nil)
+
+// RegisterSegments iterates the segments channel and for each of the segments:
```
lukedirtwalker: comment seems outdated?

control/registration/ignore.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+type pluginConfig struct {
+	Level   logLevel
+	Message *template.Template
+}
+
+func parseConfig(config map[string]any) (pluginConfig, error) {
+	if config == nil {
```
lukedirtwalker: why do we need this check?

control/registration/registration.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// SegmentRegistrars maps a policy type to a map of registration policy names
+// to their corresponding segment registrars.
+type SegmentRegistrars map[beacon.RegPolicyType]map[string]SegmentRegistrar
+
+// RegisterSegmentRegistrar registers a segment registrar for a given policy type and
+// registration policy.
+func (s SegmentRegistrars) RegisterSegmentRegistrar(
+	policyType beacon.RegPolicyType, registrationPolicy string, registrar SegmentRegistrar,
```
lukedirtwalker:
```suggestion
	policyType beacon.RegPolicyType,
	registrationPolicy string,
	registrar SegmentRegistrar,
```

control/beacon/store.go
utkn: Done. We now put into all the matching groups.

control/beaconing/writer.go
utkn: Done. I only mention that it returns `GroupedBeacons` because `GroupedBeacons` already has a comment explaining what it is.

pkg/experimental/hiddenpath/beaconwriter.go
utkn: Done.

control/registration/ignore.go
utkn: For the case that the user does not provide a configuration, in which case the default value of the map would be used, i.e., `nil`.

control/beacon/store.go
```diff
@@ -58,6 +58,43 @@ func applyStoreOptions(opts []StoreOption) storeOptions {
...
+			DEFAULT_GROUP: beacons,
+		}
+	}
+	// Go through every beacon, and group it into all the registration
+	// policies that match.
+	beaconBuckets := make(GroupedBeacons)
+	for _, b := range beacons {
+		for _, regPolicy := range policy.RegistrationPolicies {
+			if regPolicy.Matcher.Match(b) {
+				beaconBuckets[regPolicy.Name] = append(beaconBuckets[regPolicy.Name], b)
```
oncilla, line 91: AFAIU, this will register all segments with all handlers.

I am not sure these were the intended semantics. E.g., it's going to be tricky to configure hidden path segments that should not be registered publicly

control/beacon/store.go
```diff
@@ -58,6 +58,43 @@ func applyStoreOptions(opts []StoreOption) storeOptions {
...
+// GroupedBeacons is a map where the key is the registration policy name and the value is a
+// slice of beacons that should be handled by that registration policy.
+type GroupedBeacons map[string][]Beacon
+
+// DEFAULT_GROUP defines the default beacon group.
+// This means that if a policy does not define any registration policies,
+// all the beacons should be put into this group.
+//
+// This should also correspond to the ID of the default plugin.
+const DEFAULT_GROUP string = "default"
```
oncilla: We use CapitalSnakeCase for constants in Go

control/registration/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
+load("@rules_go//go:def.bzl", "go_library")
+
+go_library(
+    name = "go_default_library",
+    srcs = [
+        "ignore.go",
+        "registration.go",
+        "summary.go",
+    ],
+    importpath = "github.com/scionproto/scion/control/registration",
```
oncilla: registration is a bit vague and can mean many things.

Maybe it makes sense to move this to the segreg package (directory already exists)

(no strong opinion)

control/beacon/store.go
lukedirtwalker: This registers at all handlers that match, yes.

At first @utkn implemented this as first match wins but I voted for making this to register with all that match.

I think you can find use cases in both directions, maybe you want multiple plugins to handle a single segment.

But I agree this comes at a cost that you need to define your sequences exclusively if you want hidden paths, i.e. you could not have a matcher for a single interface hidden and a fallback that matches everything as non-hidden. Instead you would need to explicitly exclude the hidden interface.

IMO this way is slightly simpler, but it puts a burden on the user. Not sure what is better.

pkg/experimental/hiddenpath/beaconwriter.go
lukedirtwalker: No?

control/registration_test.go
```diff
@@ -0,0 +1,284 @@
...
+		{
+			Name:        "No policies",
+			RegPolicies: []beacon.RegistrationPolicy{},
+			Expected: beacon.GroupedBeacons{
+				beacon.DefaultGroup: beacons,
+			},
+		},
+		{
+			Name: "Empty policy",
+			RegPolicies: []beacon.RegistrationPolicy{
```
lukedirtwalker, line 92: oh, no matcher matches all beacons?

control/registration/ignore.go
lukedirtwalker: which is ok, AFAICT from the code below.

Read access on a nil map is allowed.

pkg/experimental/hiddenpath/beaconwriter.go
utkn: Updated again.

control/registration_test.go
utkn: Yes. This is the current semantics. I am open to proposals, but here are the current semantics:
- ACL matcher is nil and Sequence matcher is nil: Matches ✅ 
- ACL matcher is nil but Sequence matcher matches: Matches ✅ 
- ACL matcher matches but Sequence matcher is nil: Matches ✅ 
- ACL matcher is non-nil and doesn't match: Doesn't match ⛔ 
- Sequence matcher is non-nil and doesn't match: Doesn't match ⛔ 

To implement a proposal, I'd need to know how each of these cases should be handled.

control/registration/BUILD.bazel
utkn: Done. Moved into segreg package.

control/beacon/store.go
utkn: After private discussion, we have decided to implement first match wins. 

To allow multiple plugins to handle a single segment, we could create a `MultiPlugin` plugin that takes in its configuration a list of plugins and dispatches the segments to all of them.

control/registration_test.go
utkn: After private discussion, we have decided to "accept all" semantic to be explicit. Hence, an empty matcher is now an invalid configuration. At least Sequence or ACL matcher must be defined for every registration policy.

lukedirtwalker (approved):

control/beacon/policy.go
```diff
@@ -39,6 +44,61 @@ const (
...
+const (
+	// RegPolicyTypeUp is the registration policy type for up segments.
+	RegPolicyTypeUp RegPolicyType = RegPolicyType(UpRegPolicy)
+	// RegPolicyTypeDown is the registration policy type for down segments.
+	RegPolicyTypeDown RegPolicyType = RegPolicyType(DownRegPolicy)
+	// RegPolicyTypeCore is the registration policy type for core segments.
+	RegPolicyTypeCore RegPolicyType = RegPolicyType(CoreRegPolicy)
+)
+
+// ToRegPolicyType converts a PolicyType to a RegPolicyType if it is a registration policy type.
```
oncilla:
```suggestion
// RegPolicyType converts a PolicyType to a RegPolicyType if it is a registration policy type.
```

control/beacon/policy.go
```diff
@@ -39,6 +44,61 @@ const (
...
+// The second return value indicates whether the conversion was successful.
+func (p PolicyType) ToRegPolicyType() (RegPolicyType, bool) {
```
oncilla:
```suggestion
func (p PolicyType) RegPolicyType() (RegPolicyType, bool) {
```

control/beacon/policy.go
```diff
@@ -39,6 +44,61 @@ const (
...
+		return segment.TypeDown
+	case RegPolicyTypeCore:
+		return segment.TypeCore
+	default:
+		panic(fmt.Sprintf("unreachable: invalid registration policy type: %s", p))
+	}
+}
+
+// ToPolicyType converts a RegPolicyType to a generic PolicyType.
+func (p RegPolicyType) ToPolicyType() PolicyType {
```
oncilla:
```suggestion
func (p RegPolicyType) PolicyType() PolicyType {
```

control/beacon/policy.go
```diff
@@ -39,6 +44,61 @@ const (
...
+	case RegPolicyTypeDown:
+		return segment.TypeDown
+	case RegPolicyTypeCore:
+		return segment.TypeCore
+	default:
+		panic(fmt.Sprintf("unreachable: invalid registration policy type: %s", p))
+	}
+}
+
+// ToPolicyType converts a RegPolicyType to a generic PolicyType.
```
oncilla:
```suggestion
// PolicyType converts a RegPolicyType to a generic PolicyType.
```

control/segreg/ignore.go
```diff
@@ -0,0 +1,182 @@
...
+
+// logLevel defines the logging level for the IgnoreSegmentRegistrationPlugin.
+type logLevel string
+
+const (
+	LOG_LEVEL_DEBUG logLevel = "debug"
+	LOG_LEVEL_INFO  logLevel = "info"
+	LOG_LEVEL_ERROR logLevel = "error"
+)
+
```
oncilla: Use CamelCase and TypeNameValue pattern for constants

```suggestion
const (
	ConfigKeyLogLevel string = "Level"
	ConfigKeyMessage  string = "Message"
)

// logLevel defines the logging level for the IgnoreSegmentRegistrationPlugin.
type logLevel string

const (
	LogLevelDebug logLevel = "debug"
	LogLevelInfo  logLevel = "info"
	LogLevelError logLevel = "error"
)

```

control/segreg/ignore.go
```diff
@@ -0,0 +1,182 @@
...
+type pluginConfig struct {
+	Level   logLevel
+	Message *template.Template
+}
+
+func parseConfig(config map[string]any) (pluginConfig, error) {
+	// Extract the log level and message from the config.
+	level := LOG_LEVEL_DEBUG // Default log level
+	var message *template.Template
+	if val, ok := config[LOG_LEVEL_CONFIG_KEY]; ok {
```
oncilla: Instead of manually mapping, we could use https://github.com/go-viper/mapstructure (which alerady is a transitive dependency)

While it is maybe overkill here, it would serve as a good example for future plugins.

Feel free to ignore, because I don't want to block the PR on this.

control/beacon/store.go
```diff
@@ -58,6 +58,44 @@ func applyStoreOptions(opts []StoreOption) storeOptions {
...
+// This should also correspond to the ID of the default plugin.
+const DefaultGroup string = "default"
+
+// groupBeacons takes a slice of beacons and groups them according to the registration policies
+// defined in the provided policy.
+//
+// The beacons that do not match any registration policy are dropped.
+//
+// If the policy defines no registration policy, all the beacons will be put into
+// the default group, indexed by DEFAULT_GROUP.
```
oncilla:
```suggestion
// the default group, indexed by DefaultGroup.
```

control/segreg/registration.go
```diff
@@ -0,0 +1,140 @@
...
+		return serrors.New("registrar already registered for policy type and registration policy",
+			"policy_type", policyType, "registration_policy", registrationPolicy)
+	}
+	s[policyType][registrationPolicy] = registrar
+	return nil
+}
+
+// RegisterDefaultSegmentRegistrar registers a default registrar for a given policy type.
+// The default registrar should be registered when no specific registration policy is defined
+// for policyType. The default registrar is associated with the name beacon.DEFAULT_GROUP.
```
oncilla:
```suggestion
// for policyType. The default registrar is associated with the name beacon.DefaultGroup.
```

oncilla (changes requested):

control/segreg/ignore.go
utkn: Done.

control/segreg/ignore.go
oncilla: Not done afaict

control/beacon/store.go
```diff
@@ -164,12 +211,18 @@ func (s *CoreStore) BeaconsToPropagate(ctx context.Context) ([]Beacon, error) {
 
 // SegmentsToRegister returns a slice of all beacons to register at the time of the call.
```
katyatitkova: It's not a slice anymore

control/beacon/store.go
```diff
@@ -94,15 +132,24 @@ func (s *Store) BeaconsToPropagate(ctx context.Context) ([]Beacon, error) {
 // SegmentsToRegister returns a channel that provides all beacons to register at
```
katyatitkova: This was confusing even before this PR, but there's no channel being returned

jiceathome:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4782)

pkg/experimental/hiddenpath/beaconwriter.go
```diff
@@ -241,6 +247,183 @@ func (l writerLabels) WithResult(result string) writerLabels {
...
+}
+
+func (p *HiddenSegmentRegistrationPlugin) New(
+	ctx context.Context,
+	policyType beacon.RegPolicyType,
+	config map[string]any,
+) (segreg.SegmentRegistrar, error) {
+	segType := policyType.SegmentType()
+	if segType != seg.TypeDown {
+		return nil, serrors.New("hidden path registration only supports down segments")
```
romshark: Maybe make this a sentinel error value?

control/tasks.go
```diff
@@ -78,6 +79,76 @@ type TasksConfig struct {
...
+					"policy_type", policy.Type, "registration_policy", regPolicy.Name)
+			}
+		}
+	}
+	// For the policy types that do not have any plugins registered, we construct a registrar from
+	// the default plugin.
+	// This is done for the sake of backward compatibility.
+	defaultPlugin, ok := segreg.GetDefaultSegmentRegPlugin()
+	if !ok {
+		return serrors.New("default segment registration plugin not registered")
```
romshark: Same here, maybe a sentinel error value?

control/tasks.go
```diff
@@ -353,7 +388,7 @@ type Store interface {
 	// parsing error) occurs; otherwise, it returns a slice containing the beacons (which
 	// potentially could be empty when no beacon is found) and no error.
 	// The selections is based on the configured propagation policy for the requested segment type.
-	SegmentsToRegister(ctx context.Context, segType seg.Type) ([]beacon.Beacon, error)
+	SegmentsToRegister(ctx context.Context, segType seg.Type) (beacon.GroupedBeacons, error)
```
romshark, line 392: The comment is no longer up to date. Consider changing to something like:
```
// SegmentsToRegister returns an error if for example connection or
// parsing error occurs; otherwise returns grouped beacons (empty when no beacon is found).
// The selection is based on the configured propagation policy for the requested segment type.
```

control/registration_test.go
```diff
@@ -0,0 +1,419 @@
...
+	r.Results = beacons
+	sum := segreg.NewSummary()
+	for _, b := range beacons {
+		sum.RecordBeacon(&b)
+	}
+	return sum
+}
+
+func MustParseSequence(t *testing.T, seq string) *pathpol.Sequence {
+	sequence, err := pathpol.NewSequence(seq)
```
romshark, line 57:
```suggestion
	t.Helper()
	sequence, err := pathpol.NewSequence(seq)
```

control/registration_test.go
```diff
@@ -0,0 +1,419 @@
...
+
+}
+
+type testLogger struct {
+	Output string
+}
+
+var _ log.Logger = (*testLogger)(nil)
+
+func (l *testLogger) set(level string, msg string, ctx ...any) {
```
romshark:
```suggestion
func (l *testLogger) set(level string, msg string, _ ...any) {
```

control/segreg/summary.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (s *RegistrationSummary) GetSrcs() map[addr.IA]struct{} {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	srcs := make(map[addr.IA]struct{})
+	for src := range s.srcs {
+		srcs[src] = struct{}{}
+	}
+	return srcs
```
romshark: Use https://pkg.go.dev/maps#Clone

control/segreg/summary.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (s *RegistrationSummary) GetIfIDs() []uint16 {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	list := make([]uint16, 0, len(s.ifIDs))
+	for ifID := range s.ifIDs {
+		list = append(list, ifID)
+	}
+	sort.Slice(list, func(i, j int) bool { return list[i] < list[j] })
```
romshark, line 86: Use https://pkg.go.dev/slices#Sort

https://go.dev/play/p/OYW9LflsxQ7

romshark: Use `slices.Sorted(maps.Keys(s.ifIDs))`

https://go.dev/play/p/5Q7lmcA8bnz

romshark: I just tested out of curiosity: https://go.dev/play/p/0xrHKCrwu3j
the most efficient combination is actually `range`, `append` and `slices.Sort`
```
goos: darwin
goarch: arm64
pkg: benchmap
cpu: Apple M4 Pro
BenchmarkRegular-14       387828              3063 ns/op             344 B/op          3 allocs/op
BenchmarkMixed-14         770151              1392 ns/op             288 B/op          1 allocs/op
BenchmarkNew-14           678938              1752 ns/op            1104 B/op         11 allocs/op
PASS
ok      benchmap        3.660s
```

control/segreg/ignore.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+	}
+	return pluginConfig{Level: level, Message: message}, nil
+}
+
+type IgnoreSegmentRegistrationPlugin struct{}
+
+var _ SegmentRegistrationPlugin = (*IgnoreSegmentRegistrationPlugin)(nil)
+
+func (p *IgnoreSegmentRegistrationPlugin) ID() string {
```
romshark:
```suggestion
func (IgnoreSegmentRegistrationPlugin) ID() string {
```

these methods don't need to be a pointer receiver since the type is `struct{}`.

control/segreg/ignore.go
```diff
@@ -0,0 +1,177 @@
...
+) *RegistrationSummary {
+	// Create the logger.
+	logger := log.FromCtx(ctx)
+
+	// Build up a registration status that reports success for all segments.
+	startIAs := make(map[addr.IA]struct{})
+	for _, b := range segments {
+		startIA := b.Segment.FirstIA()
+		startIAs[startIA] = struct{}{}
+	}
```
romshark: Is `startIAs` even used anywhere or can it be eliminated?

control/segreg/ignore.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// If no message is configured, we do not log anything.
+	if r.Message == nil {
```
romshark, line 131: Can this check be moved up to the very beginning? It seems like nothing needs to be done if it's true.

control/segreg/ignore.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil
+	}
+
+	summary := NewSummary()
+
+	for _, b := range segments {
+		var sb strings.Builder
```
romshark: I assume reusing the builder with `sb.Reset` instead of allocating a new one on each iteration might eliminate a few unnecessary allocations.

control/beaconing/writer.go
```diff
@@ -131,137 +143,101 @@ func (r *WriteScheduler) run(ctx context.Context) error {
...
-			logger.Error("Unable to terminate beacon", "beacon", b, "err", err)
-			metrics.CounterInc(r.InternalErrors)
-			continue
-		}
+	for _, b := range beacons {
 		toRegister = append(toRegister, &seg.Meta{Type: r.Type, Segment: b.Segment})
-		beacons[b.Segment.GetLoggingID()] = b
+		logBeacons[b.Segment.GetLoggingID()] = b
 	}
 	if len(toRegister) == 0 {
```
romshark: Can this check happen earlier on `beacons`?
It looks like `toRegister`, `logger` and `logBeacons` are all for nothing if `len(beacons) < 1`

control/beaconing/writer.go
```diff
@@ -280,17 +256,107 @@ func (r *LocalWriter) updateMetricsFromStat(s seghandler.SegStats, b map[string]
...
+			summary: summary,
+			wg:      &wg,
+			pather:  r.Pather,
+		}
+
+		// Avoid head-of-line blocking when sending message to slow servers.
+		s.start(ctx, b)
+	}
+	wg.Wait()
+	if expected > 0 && summary.GetCount() <= 0 {
```
romshark: couldn't this just be `len(beacons)`? In that case `expected++` is unnecessary work.

control/beaconing/writer.go
```diff
@@ -365,3 +425,109 @@ func (l writerLabels) WithResult(result string) writerLabels {
...
+			return WriteStats{}, serrors.Wrap("getting segment registrar", err,
+				"policy", w.PolicyType, "name", name)
+		}
+		tasks = append(tasks, task{Beacons: beacons, Reg: registrar})
+	}
+	// Run the tasks concurrently and collect the write stats.
+	allWriteStats := make([]WriteStats, len(tasks))
+	wg := sync.WaitGroup{}
+	for i, task := range tasks {
+		wg.Add(1)
```
romshark:
```suggestion
	wg.Add(len(tasks))
	for i, task := range tasks {
```

control/beaconing/writer.go
```diff
@@ -365,3 +425,109 @@ func (l writerLabels) WithResult(result string) writerLabels {
...
+) (WriteStats, error) {
+	processedBeacons := w.processSegments(ctx, allBeacons, peers)
+	writeStats := WriteStats{Count: 0, StartIAs: make(map[addr.IA]struct{})}
+	// Defines a concurrent task, i.e., registration of a group of segments with a
+	// specific registrar.
+	type task struct {
+		Beacons []beacon.Beacon
+		Reg     segreg.SegmentRegistrar
+	}
+	var tasks []task
```
romshark:
```suggestion
	tasks := make([]task, 0, len(processedBeacons))
```

katyatitkova (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/scionproto/scion/4782#-:-OVNqsRWDP_vMLl9uEJG:bnfp4nl)* status: 0 of 25 files reviewed, 47 unresolved discussions (waiting on @lukedirtwalker, @oncilla, @romshark, and @utkn)

PR #4783, opened 2025-06-17
private/app/launcher: unwrap errors fully to avoid hiding cause

MatthewCroughan:
Before this commit, when a key was removed from the struct that defines available configuration option for the toml file for a given application such as the scion-dispatcher, it would fail without reporting what key is present when it should be removed from the config file, as follows:

`error: loading config from file {file=/foo.toml}: strict mode: fields in the document are missing in the target struct`

After this commit it will fail more informatively as follows:

`error: loading config from file {file=/foo.toml}: strict mode: fields in the document are missing in the target struct &toml.StrictMissingError{Errors:[]t oml.DecodeError{toml.DecodeError{message:"missing field", line:4, column:1, key:toml.Key{"general", "reconnect_to_dispatcher"}, human:"1| [general]\n2| config_dir = \"/etc/scion\"\n3| id = \"sd\"\n4| reconnect_to_dispatcher = true\n | ~~~~~~~~~~~~~~~~~~~~~ ~~ missing field\n5|\n6| [log.console]\n7| level = \"info\""}}}`

You can see the reason for the failure is the presence of the "reconnect_to_dispatcher" key, which no longer exists in 0.12, but did in 0.11.

I do not think this PR is necessarily the best way to accomplish or format this, but it was useful for me and is a good way to discuss the issue. If anyone else has a better idea, then please contribute it!


MatthewCroughan:
It is worth noting that the inverse case (a config option being missing when it is required) is not fixed, and still behaves badly despite this pr. For example when `local_udp_forwarding` is absent from a dispatcher configuration, it will fail as follows:

`fatal error: loading config from file {file=/foo.html}: strict mode: fields in the document are missing in the target struct &toml.StrictMissingError{E}`

private/app/launcher/launcher_unix.go
```diff
@@ -46,7 +47,7 @@ type Application struct {
 // Run will exit the application if it encounters a fatal error.
 func (a *Application) Run() {
 	if err := a.run(); err != nil {
-		fmt.Fprintf(a.getErrorWriter(), "fatal error: %v\n", err)
+		fmt.Fprintf(a.getErrorWriter(), "fatal error: %v\n %#v\n", err, errors.Unwrap(err))
```
oncilla, line 50: IMO, this still produces hard to read error messages.

Instead, I think you should use errors.As together with toml.StrictMissingError and toml.DecodeError

If error matches -> pretty print to stderr

I would also leave this line unchanged.

Issue #4388, opened 2023-09-02
proposal: Resolving the SVC resolution issue


oncilla:
Status: Draft
Note that this proposal is not finished yet. There are some TODOs, and we probably want some PoCs to validate the claims.
----

# Resolving the SVC resolution issue

Currently, we have a process that is called SVC (service) resolution in the
SCION control plane exchanges. Its purpose is to resolve the address of a
SCION control plane service. This is done by sending a packet with an SVC
destination address to the target AS. The response contains the address where
the service is reachable (technically a map from protocol to address, but for
now we have ever only supported one protocol)

TODO: Why SVC resolution is an issue

## Uses

SVC resolution has currently the following two uses:

1. Bootstrapping communication to remote AS

   Given we are at a very deep layer in the networking stack, we cannot rely on
   many other systems during bootstrapping. To establish SCION control plane
   connections, we need to talk to services whose addresses we do not know
   beforehand. The connection is stream oriented, thus we need packets to
   consistently delivered to the same server. SVC resolution is a way to achieve
   this.

2. Bootstrapping SCION path from One-Hop paths

   Currently, during beaconing, One-Hop paths are used to send the SVC resolutions
   the responses are sent using a full SCION path, which allows the SVC resolution
   client to bootstrap a valid SCION dataplane path.

Both of these two uses can be achieved in a different manner. Use 1 can be
solved in various different ways, we do not necessarily need SVC resolution.
Use 2 is not obviously necessary. If we still require it, we could also decouple
it from SVC resolution.

## Background

TODO: Add more background

- How we got here?
- Packet based RPC stack -> move to stream based approach
- Limitations of quic-go

## Proposal 1 - SVC resolution free communication

One key observation is that we are using QUIC in our SCION control plane
protocol. During the design of QUIC, a lot of thought went into connection
migration and resumption. Every QUIC packet carries a [connection
ID](https://datatracker.ietf.org/doc/html/rfc9000#name-connection-id). Different
connections are de-multiplexed based on their connection ID, and not based on
the addresses of the packets. The connection ID can also be used by load
balancers to infer where the packet should be sent to. E.g.,
[quic-lb](https://datatracker.ietf.org/doc/html/draft-ietf-quic-load-balancers)
attempts to standardize a scheme to encode routing data in the connection ID.

We can leverage this fact in our SCION control plane too. In the world of QUIC
connections, resolving an address first to establish the connection is
unnecessary. As long as the packets get routed to the same server, it will
[...]

PR #4786, opened 2025-06-18
beaconing: add discovery extension

lukedirtwalker:
This PR adds a beacon extension to include control and discovery service addresses along with beacons. This will help to reduce the number of SVC resolution requests we need. With this information we can simply use the provided addresses instead of doing SVC resolution when communicating upstream.

Contributes #4388


lukedirtwalker:
The work is now under this branch https://github.com/scionproto/scion/tree/less-svc-resolutions

PR #4788, opened 2025-06-18
connectrpc: add support for server and client side

oncilla:
Add connectrpc server support for all endpoints with exception of the hidden path inter AS RPCs.

On the client side, we use happy eyeballs with currently a fixed delay of 500ms to connect over the legacy gRPC SCION. This might be tuned in a follow up commit.

The intra AS RPCs are also support connectrpc, but the client side is not adapted. Here, we anyway have gRPC/TCP at the moment and moving to connectrpc is less urgent.

Contributes to https://github.com/scionproto/scion/issues/4434

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -89,7 +89,7 @@ use_repo(
     "com_github_stretchr_testify",
     "com_github_uber_jaeger_client_go",
     "com_github_vishvananda_netlink",
-    "in_gopkg_yaml_v2",
+    "in_gopkg_yaml_v3",
     "org_go4_netipx",
     "org_golang_google_grpc",
     "org_golang_google_grpc_examples",
--- a/control/beacon/BUILD.bazel
+++ b/control/beacon/BUILD.bazel
@@ -27,7 +27,7 @@ go_library(
         "//private/segment/verifier:go_default_library",
         "//private/trust:go_default_library",
         "@com_github_patrickmn_go_cache//:go_default_library",
-        "@in_gopkg_yaml_v2//:go_default_library",
+        "@in_gopkg_yaml_v3//:go_default_library",
         "@org_golang_google_protobuf//proto:go_default_library",
     ],
 )
--- a/control/beacon/policy.go
+++ b/control/beacon/policy.go
@@ -15,9 +15,10 @@
 package beacon
 
 import (
+	"io"
 	"os"
 
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 
 	"github.com/scionproto/scion/pkg/addr"
 	"github.com/scionproto/scion/pkg/private/ptr"
@@ -213,9 +214,11 @@ func (p *Policy) initDefaults(t PolicyType) {
 }
 
 // ParsePolicyYaml parses the policy in yaml format and initializes the default values.
-func ParsePolicyYaml(b []byte, t PolicyType) (*Policy, error) {
+func ParsePolicyYaml(r io.Reader, t PolicyType) (*Policy, error) {
+	d := yaml.NewDecoder(r)
+	d.KnownFields(true)
 	p := &Policy{}
-	if err := yaml.UnmarshalStrict(b, p); err != nil {
+	if err := d.Decode(p); err != nil {
 		return nil, serrors.Wrap("Unable to parse policy", err)
 	}
 	if p.Type != "" && p.Type != t {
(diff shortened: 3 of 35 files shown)


pkg/connect/happy/happy.go
```diff
@@ -0,0 +1,143 @@
...
+	return c.Typ
+}
+
+type NoReturn2[I1, I2 any] func(context.Context, I1, I2) error
+
+func (d NoReturn2[I1, I2]) Call(ctx context.Context, i1 I1, i2 I2) (struct{}, error) {
+	return struct{}{}, d(ctx, i1, i2)
+}
+
+func Happy[R any](ctx context.Context, fast, slow Caller[R]) (R, error) {
```
jiceathome: Up to you, but... I am not sure that the fast/slow distinction is relevant. Since there is some provisions for one method being preferred (referring to the 500 ms offset between the two attempts), then, may be "preferred" and "fallback"  would be more general?

pkg/snet/squic/early.go
```diff
@@ -0,0 +1,145 @@
+// Copyright 2020 Anapaya Systems
```
jiceathome: Is this code that Anapaya has kept in the cellar for 5 years?

private/segment/segfetcher/connect/requester.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2023 Anapaya Systems
```
jiceathome: copyright year is correct?

jiceatscion (approved):

jiceatscion (approved):

pkg/connect/happy/happy.go
oncilla: Renamed and adjusted code such that we can easily turn off rpcs for testing

jiceatscion (approved):

control/beaconing/happy/sender.go
```diff
@@ -0,0 +1,112 @@
...
+// NewSender returns a beacon sender that can be used to send beacons to a remote CS.
+func (f *BeaconSenderFactory) NewSender(
+	ctx context.Context,
+	dstIA addr.IA,
+	egIfID uint16,
+	nextHop *net.UDPAddr,
+) (beaconing.Sender, error) {
+	connectSender, err := f.Connect.NewSender(ctx, dstIA, egIfID, nextHop)
+	if err != nil {
+		return nil, err
```
romshark: Does it make sense to wrap those errors to make it easier to understand which of the two (connect sender or grpc sender) failed?

control/beaconing/happy/sender.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (s BeaconSender) Close() error {
+	var errs serrors.List
+	if err := s.Connect.Close(); err != nil {
+		errs = append(errs, err)
+	}
+	if err := s.Grpc.Close(); err != nil {
+		errs = append(errs, err)
+	}
+	return errs.ToError()
```
romshark:
```suggestion
    return errors.Join(s.Connect.Close(), s.Grpc.Close())
```

romshark: I suppose the extra allocation for `errors.Join`'s variadic param isn't a problem.
But this is just a nit-pick. Both is fine for me so feel free to ignore this suggestion.

P.S. keep it mind this change changes imports so don't use the "Commit suggestion" button.

control/drkey/connect/fetcher.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	for _, p := range paths {
+		if _, ok := f.errorPaths[snet.Fingerprint(p)]; ok {
+			continue
+		}
+		f.errorPaths[snet.Fingerprint(p)] = struct{}{}
+		return p, nil
+	}
+	// we've tried out all the paths; we reset the map to retry them.
+	f.errorPaths = make(map[snet.PathFingerprint]struct{})
```
romshark:
```suggestion
	clear(f.errorPaths)
```

pkg/connect/server.go
```diff
@@ -0,0 +1,107 @@
...
+			return http.ErrServerClosed
+		}
+		if err != nil {
+			// If the context has been canceled, we do not report an error.
+			if ctx.Err() != nil {
+				return nil
+			}
+			return err
+		}
+		go func() {
```
romshark: shouldn't this rather be:
```go
if errors.Is(err, context.Canceled) {
    // If the context has been canceled, we do not report an error.
    return nil
}
```

romshark (approved):
LGTM

control/beaconing/happy/sender.go
oncilla: Done

control/beaconing/happy/sender.go
oncilla: Done

control/drkey/connect/fetcher.go
oncilla: Done

pkg/connect/server.go
oncilla: Done

PR #4789, opened 2025-06-18
controlplane: do less SVC resolution

lukedirtwalker:
This PR adds a new beacon extension that contains the control and discovery service addresses of the AS.
All code has been adapted to use the address out of the extension if it's available.

The only place that still relies on SVC resolution in this new stack is the beacon sending,
because we don't know the control service address of the receiving AS.

This implements a variation of #4388

Co-authored-by: JordiSubira <jordi.subira.nieto@gmail.com>


pkg/segment/extensions/discovery/discovery.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, nil
+	}
+	cses, csErr := transformSliceWithError(
+		pb.ControlServiceAddresses, netip.ParseAddrPort,
+	)
+	dses, dsErr := transformSliceWithError(
+		pb.DiscoveryServiceAddresses, netip.ParseAddrPort,
+	)
+	if err := errors.Join(csErr, dsErr); err != nil {
+		// If any of the addresses are invalid, we return nil.
```
JordiSubira: Couldn't we just ignore the invalid ones and parse the new ones. It is not critical, but if we ever drop the "redirection socket", i.e., we always rely on information being present in the beacons, the operation would resist some misconfiguration on the server side or some divergence on the address format.

pkg/snet/addrutil/addrutil.go
```diff
@@ -87,6 +87,26 @@ func (p Pather) GetPath(svc addr.SVC, ps *seg.PathSegment) (*snet.SVCAddr, error
 	if nextHop == nil {
 		return nil, serrors.New("first-hop border router not found", "intf_id", ifID)
 	}
+	if disco := ps.ASEntries[0].Extensions.Discovery; disco != nil {
+		if svc == addr.SvcCS && len(disco.ControlServices) > 0 {
+			// take any (the first) control service address
```
JordiSubira: Should we add TODO for this and similar places, so that in the future we may have some client based load balancing if there are multiple servers available?

pkg/snet/addrutil/addrutil.go
```diff
@@ -131,3 +151,62 @@ func ResolveLocal(dst net.IP) (net.IP, error) {
...
+
+	destination := path.Destination()
+	metadata := path.Metadata()
+	if metadata != nil {
+		discoveredInfo, hasDiscoveryInfo := metadata.DiscoveryInformation[destination]
+		switch a {
+		case addr.SvcCS:
+			if hasDiscoveryInfo && len(discoveredInfo.ControlServices) > 0 {
+				// Use the first control service if available
+				cs := discoveredInfo.ControlServices[0]
```
JordiSubira: Ditto (elsewhere load balancing)

pkg/snet/addrutil/addrutil.go
```diff
@@ -131,3 +151,62 @@ func ResolveLocal(dst net.IP) (net.IP, error) {
...
+						Port: int(cs.Port()),
+					},
+				}
+
+				return ret
+			}
+		case addr.SvcDS:
+			if hasDiscoveryInfo && len(discoveredInfo.DiscoveryServices) > 0 {
+				// Use the first data service if available
+				ds := discoveredInfo.DiscoveryServices[0]
```
JordiSubira: Ditto (elsewhere load balancing)

proto/control_plane/v1/seg_extensions.proto
```diff
@@ -58,6 +60,18 @@ message StaticInfoExtension {
     string note = 6;
 }
 
+message DiscoveryExtension {
+    // Version of the discovery extension. This is set to 1 for now.
+    uint32 version = 1;
+    // The address of the control services for this AS. Note that this
```
JordiSubira: Should we also add some hint about the fact that beacons should carry information about services that are reachable through the "local interface" where the beacon is propagated?

pkg/experimental/hiddenpath/discovery.go
```diff
@@ -88,15 +88,12 @@ func resolve(ctx context.Context, ia addr.IA, discoverer Discoverer, router snet
...
-		NextHop: p.UnderlayNextHop(),
-		Path:    p.Dataplane(),
-		SVC:     addr.SvcDS,
-	}
-	if dsAddr.Path == nil {
-		dsAddr.Path = path.Empty{}
-	}
+
+	dsAddr := addrutil.ExtractDestinationServiceAddress(addr.SvcDS, p)
+	// TODO: Why is this necessary?
```
jiceathome: That rings a bell. I think that's something that I added a long time ago when I finished the hiddenpaths feature. There's corner case somewhere that crashes if Path is nil. I could go dig in old commits if you really need to find out why. It is also possible that the case has disappeared since.

JordiSubira: I would say it is a trade-off of how much effort it is for you. Otherwise, I could also make an uglier implementation that somehow reproduces this behavior, although for me it doesn't make much sense, because in the previous code we were checking for `p.Dataplane() == nil` not `p` and I haven't seen any class implementing `snet.Path` that returns nil when `p.Dataplane()`... but probably I am missing sth.

jiceathome: Ok. I went to dig; out of curiosity. It seems that the only reason for setting an empty path was
my comment from https://github.com/scionproto/scion/pull/4376:  "In one case we were constructing a full SVC address with a nil path instead of an empty path." . Probably referring to the piece of code that you replaced with "addrutil.ExtractDestinationServiceAddress(addr.SvcDS, p)". I must have found one case where p.dataPlane() is nil and the SVC address ended-up with a nil path. Is that permissible? If I found it, that would be by seeing a test crash.

control/beaconing/handler.go
```diff
@@ -127,13 +128,24 @@ func (h Handler) verifySegment(ctx context.Context, segment *seg.PathSegment,
...
+	var remoteAddr net.Addr = &snet.SVCAddr{
 		IA:      peer.IA,
 		Path:    peerPath.Dataplane(),
 		NextHop: peerPath.UnderlayNextHop(),
 		SVC:     addr.SvcCS,
 	}
-	return segverifier.VerifySegment(ctx, h.Verifier, svcToQuery, segment)
+
+	if disco := segment.ASEntries[segment.MaxIdx()].Extensions.Discovery; disco != nil &&
+		len(disco.ControlServices) > 0 {
```
romshark, line 139: 🕺🪩

control/segreq/helpers.go
```diff
@@ -30,7 +30,7 @@ import (
 
 // Pather computes the remote address with a path based on the provided segment.
 type Pather interface {
```
romshark, line 32: nit: IMHO, this interface's name should rather be `PathGetter`.

pkg/segment/export_test.go
```diff
@@ -14,14 +14,7 @@
 
 package segment
 
-import (
-	cppb "github.com/scionproto/scion/pkg/proto/control_plane"
+var (
+	ExtensionsFromPB = extensionsFromPB
+	ExtensionsToPB   = extensionsToPB
 )
```
romshark: IMHO this file can be deleted and the code can be moved to `extensions_test.go`

pkg/segment/extensions_test.go
```diff
@@ -32,6 +32,7 @@ func TestDecodeEncodeEpicDigest(t *testing.T) {
 	ext := Extensions{
 		Digests: dig,
 	}
-	ext2 := ExtensionsFromPB(ExtensionsToPB(ext))
+	ext2, err := ExtensionsFromPB(ExtensionsToPB(ext))
+	assert.NoError(t, err)
```
romshark, line 36: nit: Use `require` instead since `ext2` is garbage if `err != nil`

pkg/slices/transform.go
```diff
@@ -0,0 +1,26 @@
...
+func Transform[In any, Out any](in []In, transform func(In) Out) []Out {
+	if in == nil {
+		return nil
+	}
+	out := make([]Out, 0, len(in))
+	for _, v := range in {
+		out = append(out, transform(v))
+	}
+	return out
+}
```
romshark:
```suggestion
func Transform[S ~[]I, I any, O any](s S, transform func(I) O) []O {
	if s == nil {
		return nil
	}
	out := make([]O, len(s))
	for i, v := range s {
		out[i] = transform(v)
	}
	return out
}
```

- Use `[S ~[]I, I any, O any]` such that any named slice type can be passed without explicit conversion.
- Avoid using `append` to gain a small performance advantage.

```
var SO []uint64

func Benchmark(b *testing.B) {
	in := make([]uint64, 32)
	for b.Loop() {
		SO = Transform(in, func(x uint64) uint64 { return x + 1 })
	}
}

func BenchmarkOpt(b *testing.B) {
	in := make([]uint64, 32)
	for b.Loop() {
		SO = TransformOpt(in, func(x uint64) uint64 { return x + 1 })
	}
}
```

```
rsc@Mac benchmap % go test -bench . -benchmem -count 2 -benchtime 5s
goos: darwin
goarch: arm64
pkg: benchmap
cpu: Apple M4 Pro
Benchmark-14            100000000               55.67 ns/op          256 B/op          1 allocs/op
Benchmark-14            100000000               56.82 ns/op          256 B/op          1 allocs/op
BenchmarkOpt-14         130025940               46.22 ns/op          256 B/op          1 allocs/op
BenchmarkOpt-14         128801870               46.44 ns/op          256 B/op          1 allocs/op
PASS
ok      benchmap        23.501s
```

romshark: One can actually gain even more performance by unrolling the loop 😜

https://go.dev/play/p/LNhzlUlS5IC

```
goos: darwin
goarch: arm64
pkg: benchmap
cpu: Apple M4 Pro
                │   old.txt    │               new.txt               │
                │    sec/op    │   sec/op     vs base                │
/len_0-14          3.444n ± 0%   3.627n ± 1%   +5.31% (p=0.000 n=10)
/len_3-14          13.05n ± 1%   13.26n ± 2%   +1.57% (p=0.042 n=10)
/len_8-14          20.59n ± 1%   19.54n ± 0%   -5.12% (p=0.000 n=10)
/len_9-14          23.15n ± 0%   21.75n ± 1%   -6.03% (p=0.000 n=10)
/len_64-14        120.90n ± 1%   84.90n ± 1%  -29.78% (p=0.000 n=10)
/len_70-14        142.60n ± 1%   95.00n ± 1%  -33.38% (p=0.000 n=10)
/len_1100-14       3.381µ ± 1%   1.183µ ± 1%  -65.01% (p=0.000 n=10)
/len_65536-14     157.70µ ± 2%   66.41µ ± 0%  -57.89% (p=0.000 n=10)
/len_1048576-14   3754.3µ ± 2%   741.0µ ± 0%  -80.26% (p=0.000 n=10)
geomean            448.8n        277.0n       -38.29%
```

romshark:
Even though the new package `slices` is relatively straightforward we should still add tests.

romshark (approved):
LGTM, superficially speaking

pkg/experimental/hiddenpath/discovery.go
JordiSubira: Fair enough, then I will assume that since test are passing this is not necessary anymore.

control/segreq/helpers.go
JordiSubira: I normally prefer not to touch things that aren't specific to the PR feature/fix...

pkg/slices/transform.go
JordiSubira: It make sense although I am not sure we worry to much here about performance and type generalization, I will let @lukedirtwalker decide if he wants to change his code.

proto/control_plane/v1/seg_extensions.proto
lukedirtwalker: moved the comment to the extension and clarified the point

pkg/slices/transform.go
lukedirtwalker: Done, I don't think we should go into unrolling...

pkg/segment/extensions/discovery/discovery.go
lukedirtwalker: Done

pkg/snet/addrutil/addrutil.go
lukedirtwalker: Use random selection now

pkg/snet/addrutil/addrutil.go
lukedirtwalker: Done

pkg/snet/addrutil/addrutil.go
lukedirtwalker: Done

pkg/slices/transform.go
romshark: We should **not** unroll this :) I just had my micro-optimization flashbacks 😄 (hence the "😜")
The unrolling isn't worth it in this case.

pkg/snet/addrutil/addrutil.go
```diff
@@ -131,3 +152,62 @@ func ResolveLocal(dst net.IP) (net.IP, error) {
 	srcIP := udpConn.LocalAddr().(*net.UDPAddr).IP
 	return srcIP, nil
 }
+
+// ExtractDestinationServiceAddress extracts the destination service address
+// from the provided path. If the path contains discovery information, it will
+// use the first available control or discovery service address based on the
```
JordiSubira: Adapt comment to new behaviour, i.e., random

pkg/segment/as.go
```diff
@@ -86,7 +86,10 @@ func ASEntryFromPB(pb *cppb.ASEntry) (ASEntry, error) {
 		peerEntries = append(peerEntries, peerEntry)
 	}
 
-	extensions := extensionsFromPB(entry.Extensions)
+	extensions, err := extensionsFromPB(entry.Extensions)
+	if err != nil {
+		return ASEntry{}, err
```
JordiSubira, line 91: This comment would rather be here: https://github.com/scionproto/scion/blob/less-svc-resolutions/pkg/segment/seg.go#L120, but I cannot add comments to unmodified lines (great system :irony). We bubble up now this error, and the caller returns an error for the whole segment if we weren't able to parse on single entry from the segment. Is this desirable, perhaps we do not even need the extension for that specific AS?

pkg/snet/addrutil/addrutil.go
```diff
@@ -39,7 +40,7 @@ type Pather struct {
 }
 
 // GetPath computes the remote address with a path based on the provided segment.
```
JordiSubira: Also make it explicit here that the returned address if discovery information exist will be a random entry, which is not deterministic.

lukedirtwalker: Done.

pkg/snet/addrutil/addrutil.go
lukedirtwalker: Done.

pkg/segment/as.go
lukedirtwalker: Hm, I'm not sure here TBH. But I would lean towards being strict. i.e. if someone puts garbage in the beacon it should bubble up.
If you take it to the extreme it could be that someone would use a different proto definition for the extensions and put something completely different in place of one of the extensions, at which point I would assume the proto parsing would fail and we would also end up with an error.
But I could also live with simply logging invalid extensions and completely ignore them.

JordiSubira: I do not have a strong opinion in any case, what you say makes sense. Let's then keep then the current behavior (no need to change anything)

JordiSubira (commented):
LGTM

jiceathome:
This change is [Reviewable](https://reviewable.io/reviews/scionproto/scion/4789)

katyatitkova (approved):

oncilla (approved):

katyatitkova (approved):

PR #4790, opened 2025-06-19
[NOT TO BE MERGED] Check SVC resolution only happens with one-hop path

JordiSubira:
This PR intention is to check that only one-hop path traffic is using the SVC resolution mechanism, which includes two additional round trips. In fact this reduces the traffic to beaconing. 

https://github.com/scionproto/scion/pull/4789 contains the actual contribution.

This PR will be closed without merging after passing all the test and merging https://github.com/scionproto/scion/pull/4789


router/dataplane.go
```diff
@@ -1912,6 +1912,10 @@ func (d *dataPlane) resolveLocalDst(
 		}
 	}
 
+	if a.Type() == addr.HostTypeSVC && s.PathType != onehop.PathType {
```
jiceathome, line 1915: One question about this though. Even assuming that this never happens in any test, is there any reason why it couldn't happen in the wild? After all, until removed from the spec, Service Addresses are a generally available address type. People could (not passing judgement on the wisdom of it) decide to use service addressing in application code if they wanted to. So, I am wondering what we can make of that information once we have it. Is it so we know we can begin deprecating svc addresses?

JordiSubira: Well the idea is that it can happen, that's why we are not adding this check whatsover. Additionally, we did not drop the support for the "resolution socket", we haven't touched that part, so that this is backwards compatible. 

This is an "implementation" check, mainly to verify that we are implemented the new resolution method everywhere we intended. As I said, this PR is going to be closed without merging. If I do not understand your point let me know.

jiceathome: Ah, ok. You have answered my question. I was under the impression that you were planning to delete SVC addressing except for the OHP use case (and this was an attempt at verifying that other use cases didn't exist.

PR #4791, opened 2025-06-19
tools: increase local topology cert validity to 365 days

lschulz:
This patch increases the validity period of certificates generated by `scion.sh` for local topologies to 365 days from the current 3 days.

The 3 days expiration time makes working with local topologies quite annoying, especially when one manually changes the generated topology and is forced to repeat the modifications every 3 days.

### DIFF
--- a/doc/dev/run.rst
+++ b/doc/dev/run.rst
@@ -39,7 +39,7 @@ Quickstart
       ./scion.sh topology -c topology/tiny.topo
 
 
-   .. Attention:: The certificates created by this command expire after 3 days if the
+   .. Attention:: The certificates created by this command expire after 365 days if the
       infrastructure is not running for automatic renewal.
 
 #. To start the infrastructure we just generated, run:
--- a/tools/topology/cert.py
+++ b/tools/topology/cert.py
@@ -47,7 +47,8 @@ class CertGenerator(object):
         self.core_count = collections.defaultdict(int)
 
     def generate(self, topo_dicts):
-        self.pki('testcrypto', '-t', self.args.topo_config, '-o', self.args.output_dir)
+        self.pki('testcrypto', '-t', self.args.topo_config, '-o', self.args.output_dir,
+                 '--as-validity', '365d')
         self._master_keys(topo_dicts)
         self._copy_files(topo_dicts)
 


jiceathome (approved):

jiceathome:
If it's not too much to ask, how about updating the warning about certs TTL in the doc: https://docs.scion.org/en/latest/dev/run.html

romshark (approved):
LGTM

romshark:
As JC already pointed out this needs to be updated:

https://github.com/scionproto/scion/blob/4ab4513377141a5978d13ec5232e5fae1dd633da/doc/dev/run.rst?plain=1#L42

lschulz:
> As JC already pointed out this needs to be updated:
> 
> https://github.com/scionproto/scion/blob/4ab4513377141a5978d13ec5232e5fae1dd633da/doc/dev/run.rst?plain=1#L42

Done.

Question: The documentation speaks of running the infrastructure for "automatic renewal", is there some documentation on how to do that (without a lot of work) for a local topology? If so, it would be helpful to link to it directly.

jiceathome (approved):

jiceathome:
> > As JC already pointed out this needs to be updated:
> > https://github.com/scionproto/scion/blob/4ab4513377141a5978d13ec5232e5fae1dd633da/doc/dev/run.rst?plain=1#L42
> 
> Done.
> 
> Question: The documentation speaks of running the infrastructure for "automatic renewal", is there some documentation on how to do that (without a lot of work) for a local topology? If so, it would be helpful to link to it directly.

Not to my knowledge.

oncilla:
> Question: The documentation speaks of running the infrastructure for "automatic renewal", is there some documentation on how to do that (without a lot of work) for a local topology? If so, it would be helpful to link to it directly.

There is no support out of the box in the testing topologies.
But you can easily run: `scion-pki certificate renew` with the appropriate `--backup` and `--expires-in` flag.
(see: https://docs.scion.org/en/latest/command/scion-pki/scion-pki_certificate_renew.html#scion-pki-certificate-renew)

In productive setups, I would suggest a cronjob for this. For local topologies, we never had the need for it.
But you could run a cron like container that does the renewal for the whole topology fwiw.

PR #4792, opened 2025-06-20
docs: improve tutorial

romshark:
Most of this was done during the Hackathon on Wednesday

- Add vagrant for setting up tutorial VMs.
- Improve shell instructions that were causing permission issues.
- Clarify the difference between the core and non-core ASes.
- Upgrade to Ubuntu 24.04 LTS.

co-authored by @nicorusti @vincent10400094 @lsj0410

### DIFF
--- a/doc/tutorials/deploy.rst
+++ b/doc/tutorials/deploy.rst
@@ -56,12 +56,14 @@ The topology of the ISD includes the inter-AS connections to neighboring ASes, a
 The difference between the core and non-core ASes is that core ASes are transit nodes.
 When going from ``ffaa:1:3`` to ``ffaa:1:2`` you will see a direct path and one going via the other core AS ``ffaa:1:1``, but you will NOT see a path going via ``ffaa:1:5`` or ``ffaa:1:4`` because they are non-core.
 
-ubuntu@scion03:~$ scion showpaths 15-ffaa:1:2
-Available paths to 15-ffaa:1:2
-2 Hops:
-[0] Hops: [15-ffaa:1:3 2>2 15-ffaa:1:2] MTU: 1472 NextHop: 127.0.0.1:31002 Status: alive LocalIP: 127.0.0.1
-3 Hops:
-[1] Hops: [15-ffaa:1:3 1>3 15-ffaa:1:1 2>1 15-ffaa:1:2] MTU: 1472 NextHop: 127.0.0.1:31002 Status: alive LocalIP: 127.0.0.1
+.. code-block:: sh
+
+   ubuntu@scion03:~$ scion showpaths 15-ffaa:1:2
+   Available paths to 15-ffaa:1:2
+   2 Hops:
+   [0] Hops: [15-ffaa:1:3 2>2 15-ffaa:1:2] MTU: 1472 NextHop: 127.0.0.1:31002 Status: alive LocalIP: 127.0.0.1
+   3 Hops:
+   [1] Hops: [15-ffaa:1:3 1>3 15-ffaa:1:1 2>1 15-ffaa:1:2] MTU: 1472 NextHop: 127.0.0.1:31002 Status: alive LocalIP: 127.0.0.1
 
 .. _prerequisites:
 
--- a/doc/tutorials/deploy.rst
+++ b/doc/tutorials/deploy.rst
@@ -77,6 +77,7 @@ This deployment requires five virtual machines (VMs) - one for each AS. We recom
 - Using the naming convention for each VM of scion01, scion02, scion03, scion04, and scion05 will help follow along with this tutorial.
 - The VM names scion01-scion05 can be configured in /etc/hosts.
 
+**Note**: You can automatically prepare the VMs using Vagrant by running ``make vm-up``.
 
 Tasks to Perform
 ----------------
--- a/doc/tutorials/deploy.rst
+++ b/doc/tutorials/deploy.rst
@@ -53,7 +53,7 @@ The topology of the ISD includes the inter-AS connections to neighboring ASes, a
 
    *Figure 1 - Topology of the sample SCION demo environment. It consists of 1 ISD, 3 core ASes and 2 non-core ASes.*
 
-The difference between the core and non-core ASes is that core ASes are inter-ISD transit nodes.
+The difference between the core and non-core ASes is that core ASes are transit nodes for the entire ISD while non-core are transit nodes only for their own child ASes.
 When going from ``ffaa:1:3`` to ``ffaa:1:2`` you will see a direct path and one going via the other core AS ``ffaa:1:1``, but you will NOT see a path going via ``ffaa:1:5`` or ``ffaa:1:4`` because they are non-core.
 
 .. code-block:: sh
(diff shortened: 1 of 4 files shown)


doc/tutorials/deploy.rst
```diff
@@ -53,7 +53,15 @@ The topology of the ISD includes the inter-AS connections to neighboring ASes, a
 
    *Figure 1 - Topology of the sample SCION demo environment. It consists of 1 ISD, 3 core ASes and 2 non-core ASes.*
 
+The difference between the core and non-core ASes is that core ASes are transit nodes.
```
jiceathome: "core ASes are transit nodes" -> "core ASes are inter-ISD transit nodes", may be?

jiceatscion (approved):

doc/tutorials/deploy.rst
romshark: Yes, I think that's even better. Thanks!

jiceatscion (commented):
Sorry, I changed my mind... the correct phrasing would be "inter-ISD".

romshark:
So the final suggestion would be:
```
The difference between the core and non-core ASes is that core ASes are transit nodes for the entire ISD while non-core are transit nodes only for their own child ASes.
```

doc/tutorials/deploy.rst
```diff
@@ -53,7 +53,17 @@ The topology of the ISD includes the inter-AS connections to neighboring ASes, a
 
    *Figure 1 - Topology of the sample SCION demo environment. It consists of 1 ISD, 3 core ASes and 2 non-core ASes.*
 
+The difference between the core and non-core ASes is that core ASes are inter-ISD transit nodes.
+When going from ``ffaa:1:3`` to ``ffaa:1:2`` you will see a direct path and one going via the other core AS ``ffaa:1:1``, but you will NOT see a path going via ``ffaa:1:5`` or ``ffaa:1:4`` because they are non-core.
```
jiceathome, line 57: I do have an issue with this whole explanation, though. The restrictions in paths being described has nothing to do with the core/non-core distinction, since there's only one ISD anyway. The restriction is the result of the chosen AS hierarchy, combined with the facts that valley-path aren't permitted (and so not created). The router will discard any packet that comes in from a parent AS and goes out to another parent.

romshark: I propose changing it to:
```
The difference between the core and non-core ASes is that core ASes are transit nodes for the entire ISD while non-core are transit nodes only for their own child ASes.
```

jiceathome: nicely said.

doc/tutorials/deploy.rst
```diff
@@ -92,7 +103,7 @@ OS Setup
 
   The host file (*/etc/hosts*) will need to be updated with the IP addresses of 5 VMs. This will need to be updated on scion01-scion05. Replace the IP addresses with the assigned IP addresses for the VMs deployed.
```
nicorusti: Should this be consistent with the line below? 
```suggestion
  The host file (*/etc/hosts*) will need to be updated with the IP addresses of 5 VMs. Set this up on one of the machines, so that you can then use it as a jump box. This tutorial uses scion01. Replace the IP addresses with the assigned IP addresses for the VMs deployed.
```

doc/tutorials/deploy.rst
```diff
@@ -92,7 +103,7 @@ OS Setup
...
 
-  Set this up on scion01-scion05.
+  Set this up on scion01:
```
nicorusti, line 106: @romshark do the Vagrant VMs already come with a name?  if not, I'd also add some text like: 

You might want to rename the machines with their name `scion0x`  (edit `/etc/hostname`), so that you know on which machine you are operating.

doc/tutorials/deploy.rst
```diff
@@ -153,8 +164,6 @@ On scion01, download the topology1.json file. On scion02, download topology2.jso
 
    wget LINK_TO_TOPOLOGY.JSON_FILE -O /etc/scion/topology.json
 
-The AS topology files reference the hosts scion01-05 by host name.
-Ensure that you have set up the ``/etc/hosts`` file (:ref:`see above <step0>`) or replace the hostnames with IP addresses.
 
 Step 2 - Generate the Required Certificates
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nicorusti, line 169: How about a bit more explanation for the PKI?
```suggestion
Step 2 - Generate the Required Certificates
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

SCION's control plane messages and path information are all authenticated through the SCION PKI. Setting up the PKI in a freshly created Isolation Domain, like in this tutorial, requires an initial trust bootstrapping process to create the initial :term:`TRC` for this ISD. Then, various  certificates need to be generated for the intermediate :term:`CAs <CA>` and for each AS.
```
This text then would need to be removed, but I cannot suggest it from here: 
```
The various cryptographic certificates need to be generated for each of the ASes.
This requires first setting up the :term:`TRC` for this ISD, and then issuing AS-certificates from the :term:`CAs <CA>`.
```

doc/tutorials/deploy.rst
```diff
@@ -240,6 +265,17 @@ Execute the following commands on every AS:
 
 These steps need to be repeated on each host scion01 - scion05.
 
+Make sure your services are running and didn't fail to start:
```
nicorusti, line 269: DO we have any troubleshooting link we could point to?

nicorusti (commented):
I added a feew comments. Besides that, @JordiSubira mentioned that he found the name a bit confusing. Could we call it a bit differently?

doc/tutorials/deploy.rst
romshark: The current Vagrant setup does set the hostname already:

```sh
rsc@Mac deploy % vagrant ssh scion01
Welcome to Ubuntu 24.04.2 LTS (GNU/Linux 6.8.0-53-generic aarch64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/pro

 System information as of Fri Jun 20 01:52:31 PM UTC 2025

  System load:           0.0
  Usage of /:            16.9% of 29.82GB
  Memory usage:          19%
  Swap usage:            0%
  Processes:             92
  Users logged in:       0
  IPv4 address for eth0: 10.0.2.15
  IPv6 address for eth0: fd17:625c:f037:2:a00:27ff:fe71:19d8


This system is built by the Bento project by Chef Software
More information can be found at https://github.com/chef/bento

Use of this system is acceptance of the OS vendor EULA and License Agreements.
vagrant@scion01:~$ cat /etc/hostname 
scion01
```

P.S. don't pay attention to the eth0 address, it's irrelevant. `eth1` is used and has the right IP assigned.

doc/tutorials/deploy.rst
romshark: I don't know of any troubleshooting docs. It seems like we don't have any. As just discussed - we'll add them separately in another PR.

doc/tutorials/deploy.rst
romshark: done

PR #4794, opened 2025-06-23
chore: replace gopkg.in/yaml.v2 with gopkg.in/yaml.v3

romshark:


control/beacon/policy.go
```diff
@@ -229,11 +232,11 @@ func ParsePolicyYaml(b []byte, t PolicyType) (*Policy, error) {
 // LoadPolicyFromYaml loads the policy from a yaml file and initializes the
 // default values.
 func LoadPolicyFromYaml(path string, t PolicyType) (*Policy, error) {
-	b, err := os.ReadFile(path)
+	f, err := os.OpenFile(path, os.O_RDONLY, 0o644)
```
lukedirtwalker, line 235: This file doesn't seem to be closed anywhere.

nogo.json
```diff
@@ -222,4 +218,4 @@
             "/bazel_gazelle/": ""
         }
     }
-}
+}
```
lukedirtwalker: nit: can't we add the newline back?

control/beacon/policy.go
romshark: fixed, good catch!

nogo.json
romshark: done

lukedirtwalker (approved):

PR #4796, opened 2025-06-25
chore: use tablewriter latest version v1.0.7

romshark:
[olekukonko/tablewriter v0.0.5](https://github.com/olekukonko/tablewriter/releases/tag/v0.0.5) is long obsolete. Use latest [v1.0.7](https://github.com/olekukonko/tablewriter/releases/tag/v1.0.7)

old API: https://go.dev/play/p/KRK3OLB_keG
new API: https://go.dev/play/p/IuvksGrAC4b

### DIFF
--- a/gateway/control/engine.go
+++ b/gateway/control/engine.go
@@ -28,6 +28,8 @@ import (
 
 	"github.com/gopacket/gopacket"
 	"github.com/olekukonko/tablewriter"
+	"github.com/olekukonko/tablewriter/renderer"
+	"github.com/olekukonko/tablewriter/tw"
 
 	"github.com/scionproto/scion/gateway/pathhealth"
 	"github.com/scionproto/scion/gateway/pathhealth/policies"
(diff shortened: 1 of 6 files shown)


gateway/control/engine.go
```diff
@@ -266,19 +268,38 @@ func renderPathInfo(p pathhealth.PathInfo, w io.Writer, indent int) {
...
-	table.SetCenterSeparator("")
-	table.SetColumnSeparator("")
-	table.SetRowSeparator("")
-	table.SetHeaderAlignment(tablewriter.ALIGN_LEFT)
-	table.SetAlignment(tablewriter.ALIGN_LEFT)
-	table.SetHeader([]string{"", "STATE", "REVOKED", "PATH"})
-	table.AppendBulk(paths)
-	table.Render()
 
+	table := tablewriter.NewTable(
```
jiceathome, line 272: I don't really see an improvement in the API, but whatever we have to do...

romshark: I agree it's not necessary to upgrade this one but the version we use is _a bit_ dated compared to latest. Eventually I want to:
- be able to `go get -u all` and have all deps updated.
- have stable dependency versions and avoid `v0.x.x` if possible.

PR #4800, opened 2025-06-27
pkg/snet: change snet.Fingerprint to take interface list

oncilla:
Make the input to the snet.Fingerprint take a list of path interfaces, rather than a full snet.Path.

This is a minor breaking change, but will allow the usage of snet.Fingerprint in more scenarios.
A new method `snet.PathMetadata.Fingerprint` is introduced to easily migrate current usages.

The following https://github.com/uber-go/gopatch can be used to migrate:

```diff
@@
var x expression
@@
-snet.Fingerprint(x)
+x.Metadata().Fingerprint()
```


pkg/snet/path.go
```diff
@@ -208,11 +208,16 @@ func (pf PathFingerprint) String() string {
 	return fmt.Sprintf("%x", []byte(pf))
 }
 
+// Fingerprintable is a type that can be fingerprinted based on its PathMetadata.
+type Fingerprintable interface {
```
marcfrei: The name `Fingerprintable` is not completely ideal, in my opinion. Would be nice to express in the name that the type is associated with paths somehow. `PathFingerprintable` wouldn't sound very elegant. `PathMetadataProvider` maybe?

marcfrei: `PathDescriptor` would be another option. But `Fingerprintable` would be fine with me too.

oncilla: That's fair enough.

I was debating to just change the function itself to take metadata directly. But that would be a breaking change that potentially touches a lot of things.

I think the idomatic thing would be to just add "-er" at the end. So, `type Metadater`. Sounded odd at first, but maybe it is the most descriptive name?

cc @romshark for opinions

marcfrei: > I was debating to just change the function itself to take metadata directly. But that would be a breaking change that potentially touches a lot of things.

Had the same thought. Maybe having a breaking change here would be worth the price since it would result in the simplest overall solution.

romshark: I personally tend to name methods as _what they do_ and interfaces accordingly, so IMHO ideally it should be:

```go
// PathMetadataGetter provides access to the path metadata.
type PathMetadataGetter interface {
	GetPathMetadata() *PathMetadata
}
```

This is probably inapplicable in this case since we won't change the method name, but `Metadater` is very off, to me it reads like it _dates meta_ 😅, so how about:

```go
// MetadataProvider provides access to the path metadata.
type MetadataProvider interface {
	Metadata() *PathMetadata
}
```

or alternatively: `MetadataGetter`

oncilla: Hm, we wouldn't want meta to be dated. Settling for MetadataProvider. PTAL

oncilla:
@romshark how do you feel about a breaking change and adapting the function inputs?

On the Anapaya side, this would lead very little churn, and it would make the whole thing cleaner.

marcfrei:
Additional datapoint re adapting the function inputs: in https://github.com/netsec-ethz/scion-apps this would break only 4 places across 3 files.

romshark:
@oncilla If the breaking change adds value and isn't too much work then why not? Go for it :)

lukedirtwalker:
Another alternative could be an anonymous interface?

oncilla:
> Another alternative could be an anonymous interface?

What do you mean by that? Do you want to change the parameter type to any and do a typeswitch?
I would rather introduce a breaking change. From the sounds of it, it will be a minor thing to adjust for our downstreams

lukedirtwalker:
>What do you mean by that? 

This.
```go
func Fingerprint(path interface {Metadata() *PathMetadata}) PathFingerprint {
```

Anyway breaking change is fine as well.

oncilla:
Ah, I was mixing it up with interface smuggling.

pkg/snet/path.go
```diff
@@ -208,17 +217,21 @@ func (pf PathFingerprint) String() string {
 	return fmt.Sprintf("%x", []byte(pf))
 }
 
+// MetadataProvider is the interface that provides access to the path metadata.
+type MetadataProvider interface {
+	Metadata() *PathMetadata
+}
+
```
marcfrei: Do we still need this? In doubt, I would remove it for now.

oncilla: Good catch. Removed

PR #4803, opened 2025-07-09
cmd/scion-pki: improve CLI UX

oncilla:
The `scion-pki` CLI commands now support reading from stdin when the file path is set to `-`, where appropriate. This allows the user to pipe data directly into the command without the need to create temporary files.

Furthermore, the `scion-pki trc extract certificates` command has two additional filtering options: `--type` and `--subject.isd-as`, allowing the user to filter the extracted certificates.

### DIFF
--- a/private/app/helper.go
+++ b/private/app/helper.go
@@ -16,6 +16,7 @@ package app
 
 import (
 	"context"
+	"io"
 	"os"
 	"os/signal"
 	"syscall"
@@ -105,4 +106,13 @@ func (c *Cleanup) Do() error {
 		}
 	}
 	return errs.ToError()
+}
+
+// ReadFileOrStdin reads the content of a file or stdin if the path is "-".
+// It returns the content as a byte slice.
+func ReadFileOrStdin(path string) ([]byte, error) {
+	if path == "-" {
+		return io.ReadAll(os.Stdin)
+	}
+	return os.ReadFile(path)
 }
--- a/scion-pki/trcs/extract.go
+++ b/scion-pki/trcs/extract.go
@@ -19,10 +19,14 @@ import (
 	"encoding/pem"
 	"fmt"
 	"os"
+	"slices"
+	"strings"
 
 	"github.com/spf13/cobra"
 
+	"github.com/scionproto/scion/pkg/addr"
 	"github.com/scionproto/scion/pkg/private/serrors"
+	"github.com/scionproto/scion/pkg/scrypto/cppki"
 	"github.com/scionproto/scion/private/app/command"
 )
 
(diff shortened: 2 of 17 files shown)


private/app/helper.go
```diff
@@ -106,3 +107,12 @@ func (c *Cleanup) Do() error {
 	}
 	return errs.ToError()
 }
+
+// ReadFileOrStdin reads the content of a file or stdin if the path is "-".
+// It returns the content as a byte slice.
+func ReadFileOrStdin(path string) ([]byte, error) {
```
oncilla, line 113: Open for suggestions if there is a more suitable place to put this.

katyatitkova (approved):

scion-pki/trcs/extract.go
```diff
@@ -86,51 +90,131 @@ To inspect the created asn.1 file you can use the openssl tool::
...
+					break
+				}
+				typ, ok := certTypes[t]
+				if !ok {
+					return fmt.Errorf("unknown certificate type %q, valid types are: %s", t, strings.Join(getTypes(), ", "))
+				}
+				types[typ] = true
+			}
+
+			ias := make(map[addr.IA]bool)
```
romshark, line 120:
```suggestion
			ias := make(map[addr.IA]bool, len(flags.ias))
```

scion-pki/trcs/extract.go
```diff
@@ -86,51 +90,131 @@ To inspect the created asn.1 file you can use the openssl tool::
...
 		Use:     "certificates",
-		Aliases: []string{"certs"},
+		Aliases: []string{"certs", "certificate", "cert"},
 		Short:   "Extract the bundled certificates",
 		Example: fmt.Sprintf(`  %[1]s certificates -o bundle.pem input.trc`, pather.CommandPath()),
 		Long:    `'certificates' extracts the certificates into a bundled PEM file.`,
 		Args:    cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if err := runExtractCertificates(args[0], flags.out); err != nil {
+			types := make(map[cppki.CertType]bool)
```
romshark, line 106: nit: sets can be used for `types` and `ias` (like `map[cppki.CertType]struct{}`) to reduce the ambiguity a bit.

The checks can then just be like:

```go
if len(ias) > 0 && !ias[ia] {
	continue
}
```

scion-pki/trcs/extract.go
```diff
@@ -86,51 +90,131 @@ To inspect the created asn.1 file you can use the openssl tool::
...
+
+	// Filter the certificates based on the user input.
+	for _, cert := range signed.TRC.Certificates {
+		// Check certificate type
+		{
+			typ, err := cppki.ValidateCert(cert)
+			if err != nil {
+				return fmt.Errorf("invalid certificate %s: %w", cert.Subject.CommonName, err)
+			}
+			if len(types) > 0 && !types[typ] {
```
romshark, line 168: if `types` contains `typ` then `len(types)` cannot be `0`, right? Hence `len(types) > 0` looks pointless to me, am I missing something?

romshark (approved):
LGTM, just some nitpicks.

I fixed the linter errors. I hope you don't mind the line breaking style.

private/app/helper.go
romshark: I believe it's fine to keep it here.

Maybe worth adding: 

```go
// ReadFileOrStdin reads the content of a file or stdin if the path is "-".
// It returns the content as a byte slice.
// Follows UNIX utility syntax guidelines.
// See https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02
```

PR #4805, opened 2025-07-11
path-combinator: fix path mtu when segment is shortcut

uniquefine:
Fixes a bug in the path construction logic.

If only use a suffix of a segment is used (shortcut path) the ingress MTU of the last ASentry that is actually used should not be included in the path MTU calculation.


private/path/combinator/graph.go
```diff
@@ -384,7 +387,7 @@ func (solution *pathSolution) Path(hashState hashState) Path {
 					ConsEgress:  peer.HopField.ConsEgress,
 					Mac:         peer.HopField.MAC,
 				}
-				forwardingLinkMtu = peer.PeerMTU
+				mtu = minUint16(mtu, uint16(peer.PeerMTU))
```
romshark: Remove `minUint16` and just use the new builtin `min` function.

uniquefine: Done

PR #4806, opened 2025-07-11
build: attempt to address complaint about disk full during CI

jiceatscion:


jiceathome:
Ah never mind. I involuntarily smuggled the same change into another PR which Dominik approved.
So, done.

Issue #4434, opened 2023-11-03
HTTP3 based SCION control plane


oncilla:
# HTTP3 based SCION control plane

Next to SVC resolution (as discussed in #4388), our SCION control plane RPCs are another messy leftover
from long gone times of experimentation and making it work. Currently, we are
hacking gRPC (which is using HTTP/2) on top of a single QUIC stream, which
emulates a TCP connection. To spell it out, our SCION control plane RPC stack
is: gRPC/HTTP2/QUIC/SCION. We are abusing QUIC streams and not leveraging QUIC
effectively. During standardization, it will be very hard to argue that this
is a good way of doing things (because it is not).

We originally decided to use QUIC because it had good user space
implementations. Now, we can profit from this decision. QUIC has become the new
transport of choice. With it, HTTP/3 has become the application transport layer
of choice too. Support for it is growing, due to the synergies with QUIC. If we can
leverage these two technologies for our SCION control plane RPCs, it will be
very easy to argue why we use them, given they are basically the standard for
current networking stacks.

While gRPC has allowed us to have a reliable SCION control plane RPC stack so
far, it [does not look like](https://github.com/grpc/grpc/issues/19126) they
will adopt HTTP/3 any time soon. Even more worrying, The grpc-go implementation
is tightly coupled to [HTTP/2](https://github.com/grpc/grpc-go/issues/3953)
(which they have re-implemented the transport themselves). Custom transports
have been proposed [5 years ago](https://github.com/grpc/proposal/pull/103)
already. Moving forward, I think we might want to adapt a different RPC stack
which builds on top of HTTP/3 and QUIC.

[Connect](https://connectrpc.com/) is an RPC framework that promises the same
benefits as gRPC (API specified in ProtoBuf, boilerplate generation, etc.).
However, it is a very focused project with a small code base. They even offer
gRPC compatible servers (which we do not need though). Connect also defines
their own protocol for [RPC over HTTP](https://connectrpc.com/docs/protocol).
Note that they do not depend on framing details of specific HTTP versions, thus
this protocol is compatible with HTTP/1.1 (except bi-directional streaming),
HTTP/2, and HTTP/3. The protocol is also so simple that it can be easily
implemented in any language that offers an HTTP and Protobuf library.

I propose that we change to connect rpc, which will allow us to do away with
our gRPC-over-QUIC-stream hack, as we can just use regular HTTP/3 over QUIC.

### Keeping Backwards Compatibility - ALPN

[...]

PR #4811, opened 2025-07-16
test: connectrpc/grpc compatibility via  happy eyeballs

jiceathome:
Summary:

* Add optional rpc_server_protocol and rpc_client_protocol flags to the API
  section of the general component config.
* Add code in CS to propagate config to some happy-enabled client code
  (beaconing, drkey, segfetcher)
* Add code in CS' main to disable grpc or connect rpc servers per config
* Add options to topogen to produce CS config files with rpc_client_protocol
  and rpc_server_protocol flags.
* Run 4 versions of the sig_ping test: client and server support both
  protocols, server supports both but client is grpc-only. client supports
  both but server is grpc-only, both client and server are connectrpc only.
  (Only the CS's config is modified).

Contributes to: #4434

### DIFF
--- a/pkg/connect/server.go
+++ b/pkg/connect/server.go
@@ -103,7 +103,7 @@ func (d ConnectionDispatcher) Run(ctx context.Context) error {
 				}
 			} else {
 				log.Error("neither Connect nor Grpc are available")
-				if err := conn.CloseWithError(serveConnError, err.Error()); err != nil {
+				if err := conn.CloseWithError(1, err.Error()); err != nil {
 					log.Error("can not close connection", "err", err)
 				}
 			}


katyatitkova (approved):

control/config/config.go
```diff
@@ -342,3 +343,6 @@ func (cfg *CAService) Sample(dst io.Writer, _ config.Path, _ config.CtxMap) {
 func (cfg *CAService) ConfigName() string {
 	return "service"
 }
+
+// RpcClientConfig informs all CS sub-components that use happy.
+var RpcClientConfig connecthappy.Config
```
oncilla: Globals are generally evil, and we do not need it here.

You can just pass along the config when you initialize the happy fetchers.

control/cmd/control/main.go
```diff
@@ -365,10 +383,15 @@ func realMain(ctx context.Context) error {
...
 	}
-	cppb.RegisterTrustMaterialServiceServer(quicServer, trustServer)
-	connectInter.Handle(cpconnect.NewTrustMaterialServiceHandler(cstrustconnect.MaterialServer{
-		MaterialServer: trustServer,
-	}))
+
+	if !grpcServerDisabled {
+		cppb.RegisterTrustMaterialServiceServer(quicServer, trustServer)
+	}
+	if !connectServerDisabled {
```
oncilla: Same comment as above

control/cmd/control/main.go
```diff
@@ -365,10 +383,15 @@ func realMain(ctx context.Context) error {
 		IA:       topo.IA(),
 		Requests: libmetrics.NewPromCounter(cstrustmetrics.Handler.Requests),
 	}
-	cppb.RegisterTrustMaterialServiceServer(quicServer, trustServer)
-	connectInter.Handle(cpconnect.NewTrustMaterialServiceHandler(cstrustconnect.MaterialServer{
-		MaterialServer: trustServer,
-	}))
+
+	if !grpcServerDisabled {
```
oncilla: Instead of guarding every single registration, couldn't you just not serve grpc server?

tools/topology/go.py
```diff
@@ -126,8 +126,12 @@ def _build_control_service_conf(self, topo_id, ia, base, name, infra_elem, ca):
             },
             'tracing': self._tracing_entry(),
             'metrics': self._metrics_entry(infra_elem, CS_PROM_PORT),
-            'api': self._api_entry(infra_elem, CS_PROM_PORT+700),
             'features': translate_features(self.args.features),
+            'api': {
+                'addr': prom_addr(infra_elem['addr'], CS_PROM_PORT+700),
+                'rpc_client_protocol': self.args.rpc_client_protocol,
```
oncilla: I would expect this to also appear on the gateway service

jiceathome: It didn't seem totally necessary to apply the test config to every component, but If you think it maters, I'll do that.

control/cmd/control/main.go
jiceathome: I had not seriously considered doing that. I was under the impression that the servers would still be expected to run by some stuff (like prometheus for example). I'll give it a try. On the other hand, I kind of like the if blocks: they delimit all the code that can be deleted (in addition to not executing it) once grpc is abandoned. You do want the ifs gone?

control/cmd/control/main.go
jiceathome: Sure. I didn't know it would be confined to the main function when I started. Done.

control/config/config.go
jiceathome: I never take aphorisms too seriously. Globals have their uses. Passing a configuration all the way from the command line down to a somewhat buried component without the stuff in-between having to know about it; that's a fair use in my opinion. Nevertheless, here's what it's like without the global (I added the config to every use of happy while I was at it). Btw, is it just me, or does it looks a bit silly to configure grpc, configure connect, and then configure a third item that disables one of the two. It made slightly more sense when the happy config was a global override.

control/cmd/control/main.go
oncilla: > I was under the impression that the servers would still be expected to run by some stuff (like prometheus for example).

There should nothing depend on it, and if it does, that's a bug.

>  On the other hand, I kind of like the if blocks: they delimit all the code that can be deleted (in addition to not executing it) once grpc is abandoned. You do want the ifs gone?

Yes, I would like the ifs gone. There is no need to delimit the code that can be deleted.
The compiler will remind us to remove all these code lines once the server is removed.

jiceathome: That works for disabling grpc. It doesn't work for disabling connect. If I do that the integration test fails. It seems that there is a dependency of the grpc-only operation upon some of the connect mechanism. I am still trying to figure out what. If you happen to remember creating such a dependency, let me know.

jiceathome: Update; I made it work by reducing the connect disabling to skipping " connectServer.ServeQUICConn(conn)".
So, I guess it is the following code that grpc cannot work without:

```
				if conn.ConnectionState().TLS.NegotiatedProtocol != "h3" {
					grpcConns <- conn
					return
				}
```

private/mgmtapi/config.go
```diff
@@ -15,21 +15,46 @@
...
 	"io"
+	"slices"
 
 	"github.com/scionproto/scion/private/config"
 )
 
 const apiSample = `
 # The address to expose the API on (host:port or ip:port).
-# If not set, the API is not exposed.
+# If not set, the API is not exposed. rpc_client_protocol specifies
```
oncilla: This seems like an odd place to configure the RPC behavior of the service.

mgmtapi is the admin API of the services that is always served on localhost. It is unrelated to the CP APIs that the services expose to the outside world.

github.com/scionproto/scion/private/env seems more appropriate.

Please also make this its own struct. That would also help not having to do the parsing logic in the mains and repeat the switch statements. The logic can be moved to methods on the struct, instead of leaking the configuration string values into the main code.

jiceathome: Yeah... can't argue. That was lazy. Moved to its own section. Now, regarding the factorization of the parsing, do you have a favorite idiom? I've seen that this is sometimes done by the Validate() method. Also, what cooked values do you believe would be most useful (i.e. may be not requiring an ugly switch statement)? May be just booleans? grpcDisabled and connectRpcDisabled? I'll go with the above to start with but let me know if you have other preferences.

oncilla: I was think instead of doing:

```go
switch globalCfg.API.RpcClientProtocol {
	case "all":
	case "grpc":
		rpcClientConfig.NoPreferred = true // Disable our preferred option, connect.
	case "connect":
		rpcClientConfig.NoFallback = true // Disable our fallback option, grpc.
	}
	switch globalCfg.API.RpcServerProtocol {
	case "all":
	case "grpc":
		connectServerDisabled = true
	case "connect":
		grpcServerDisabled = true
	}
```

you can do for example:

```go
type RPC struct{
//...
}

func (c RPC) ConnectServerEnabled() bool {}
func (c RPC) GRPCServerEnabled() bool {}
func (c RPC) ConnectClientEnabled() bool {}
func (c RPC) GRPCClientEnabled() bool {}
```

And then you do not even need variables in the mains.

jiceathome: I have some code that does roughly that but with RPC being a part of the config Item:
```
var _ config.Config = (*RPC)(nil)

type RPC struct {
	ClientProtocol string `toml:"client_protocol,omitempty"`
	ServerProtocol string `toml:"server_protocol,omitempty"`
	// Cooked values
	GrpcClientDisabled       bool
	ConnectrpcClientDisabled bool
	GrpcServerDisabled       bool
	ConnectrpcServerDisabled bool
}

...
func (cfg *RPC) Validate() error {
	switch cfg.ClientProtocol {
	case "all":
	case "grpc":
		cfg.ConnectrpcClientDisabled = true
	case "connectrpc":
		cfg.GrpcClientDisabled = true
	default:
		return fmt.Errorf("invalid value: %s", cfg.ClientProtocol)
	}
...

```

I have no problem splitting the flags and their parsing out. However there would be an extra call in the using code to extract the flag representation.

jiceathome: Side note: double-negatives (of sorts). My version has "*Dsiabled" flags. Yours has "*Enabled" flags. It has occured to me that we end-up almost always testing "if !xyzDisabled {...", which is annoying. This originated in the happy.Config with its NoPreferred and NoFallback flags; propagating the semantics from there. Would you like me to flip the Happy.Config flags and switch everything else to positive? That will mean initializing the positive flags to true, though.

oncilla: happy.Config is perfectly fine IMO.

The default is to do both -> we want the negative version.

Service config helpers I used the Enabled version because of how they are intended to be used.
I would also be fine with doing the Disabled counter part :shrug: 

Double negatives are not annoying to me. I find it much more annoying to need to initialize a field with a non-zero value to do default behavior. Hence, happy.Config is the way it is.

jiceathome: Here's what I had. Let me know what you'd do differently.

jiceathome: As you can see, I too choose the negative semantics over having to a non-zero-valued default. I mentioned it because in your example you seemed to prefer the opposite flags. Let me know about the rest.

jiceathome: how's that now?

control/cmd/control/main.go
```diff
@@ -939,9 +954,24 @@ func realMain(ctx context.Context) error {
 			Grpc: &beaconinggrpc.BeaconSenderFactory{
 				Dialer: dialer,
 			},
+			RpcConfig: rpcClientConfig,
+		},
+		SegmentRegister: &beaconinghappy.Registrar{
```
oncilla: Why do we duplicate the struct and not reuse rpc here?

control/cmd/control/main.go
```diff
@@ -895,7 +910,7 @@ func realMain(ctx context.Context) error {
 		return topoInfo.LinkType == topology.Core || topoInfo.LinkType == topology.Child
 	}
 
-	rpc := &happy.Registrar{
+	rpc := &beaconinghappy.Registrar{
```
oncilla, line 909: Does this not need client config?

doc/manuals/gateway.rst
```diff
@@ -32,3 +31,15 @@ Network prefix pinning
...
+
+Configuration
+=============
+
+In addition to the :ref:`common .toml configuration options <common-conf-toml>`, the gateway service
+considers the following options.
+
+.. object:: rpc
+
+   .. option:: rpc.client_protocol = "grpc"|"connectrpc"|"all" (Default = "all")
```
oncilla, line 43: Shouldn't the server side also be configurable?

private/env/env.go
```diff
@@ -262,3 +262,54 @@ func (cfg *Tracing) NewTracer(id string) (opentracing.Tracer, io.Closer, error)
 		jaegercfg.Extractor(opentracing.Binary, bp),
 		jaegercfg.Injector(opentracing.Binary, bp))
 }
+
+var _ config.Config = (*RPC)(nil)
+
+type RPC struct {
+	ClientProtocol string `toml:"client_protocol,omitempty"`
+	ServerProtocol string `toml:"server_protocol,omitempty"`
+	// Cooked values
```
oncilla, line 271: nit: Why do we used cooked values rather than just method access?

jiceathome: So that we don't have to cook for every access?

doc/manuals/gateway.rst
jiceathome: There is currently no connectrpc server code in-there; is there? At least I didn't see it in the hackathon PR. Did I miss something? In truth, the config will be happy to ingest a server-side setting, but since it has no effect I did not propose it in the doc.

control/cmd/control/main.go
jiceathome: indeed. missed a spot.

control/cmd/control/main.go
jiceathome: Sorry, I don't understand the question. Can you elaborate. what exactly is redundant?

jiceathome: Oh rpc from line 913? Yeah. I see it now. I don't know. It looks like a merge/refactor failure on my part. I'll fix it if that's the case.

jiceathome: done

doc/manuals/gateway.rst
oncilla: https://github.com/scionproto/scion/blob/8364ada874b450fcc53296d1e3ee04fa54acb41d/gateway/gateway.go#L576-L579

private/env/env.go
oncilla: At the risk of the cached value being out of date, and it is accessed maybe twice.

Anyway, I will not block on this, as it is a matter of taste

doc/manuals/gateway.rst
jiceathome: done

jiceathome:
@oncilla : is this now to your liking? If so, please approve the changes and hit the merge button. I lost that one finger privilege.

pkg/connect/server.go
oncilla, line 109: We are leaking the connection here.
I think we need to close it.

katyatitkova: Do you mean that we're leaking it if both `d.Connect` and `d.Grpc` are `nil`?

pkg/connect/server.go
```diff
@@ -93,11 +93,11 @@ func (d ConnectionDispatcher) Run(ctx context.Context) error {
...
-			if conn.ConnectionState().TLS.NegotiatedProtocol == "h3" {
+			if d.Connect != nil && conn.ConnectionState().TLS.NegotiatedProtocol == "h3" {
 				if err := d.Connect.ServeQUICConn(conn); err != nil {
 					d.Error(err)
 				}
-			} else {
+			} else if d.Grpc != nil {
 				if err := d.Grpc.ServeQUICConn(conn); err != nil {
 					d.Error(err)
 				}
```
katyatitkova, line 108:
```suggestion
				} else {
					log.Error("neither Connect nor Grpc are available")
					if err := conn.CloseWithError(serveConnError, err.Error()); err != nil {
						log.Error("can not close connection", "err", err)
					}
				}
```

katyatitkova: @oncilla did you mean a fix like this? It doesn't look like I can push to the branch directly unfortunately

pkg/connect/server.go
oncilla: Yes. Applied the change

oncilla (approved):

katyatitkova (approved):

pkg/connect/server.go
```diff
@@ -93,14 +93,19 @@ func (d ConnectionDispatcher) Run(ctx context.Context) error {
...
+			} else {
+				log.Error("neither Connect nor Grpc are available")
+				if err := conn.CloseWithError(serveConnError, err.Error()); err != nil {
```
katyatitkova:
```suggestion
				if err := conn.CloseWithError(1, err.Error()); err != nil {
```

PR #4815, opened 2025-07-17
cmd/scion-pki: add certificate template command

oncilla:
Add a `scion-pki certificate template` command that extracts the subject template from an existing certificate or CSR. This command can be used to reconstruct the subject template used to create a certificate or CSR.

### DIFF
--- a/scion-pki/certs/template.go
+++ b/scion-pki/certs/template.go
@@ -1,0 +1,114 @@
+// Copyright 2025 Anapaya Systems
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package certs
+
+import (
+	"crypto/x509"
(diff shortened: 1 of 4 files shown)


scion-pki/certs/template.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/spf13/cobra"
+)
+
+func newTemplateCmd(_ command.Pather) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "template",
+		Short: "Create subject template from a certificate or CSR",
+		Long: `'template' creates a subject template from a certificate, certificate chain, or CSR.
+
+This command allows reconstructing the subject template that was used to cerate a certificate
```
katyatitkova:
```suggestion
This command allows reconstructing the subject template that was used to create a certificate
```

katyatitkova (approved):

scion-pki/certs/template.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return pkix.Name{}, fmt.Errorf("reading file %q: %w", filename, err)
+	}
+	return loadPkixNameFromRaw(raw)
+}
+
+func loadPkixNameFromRaw(raw []byte) (pkix.Name, error) {
+	pemData, _ := pem.Decode(raw)
+	if pemData == nil {
+		return pkix.Name{}, fmt.Errorf("not valid PEM")
```
romshark, line 97: Consider returning a sentinel defined on package level:

```go
var ErrInvalidPEM = errors.New("invalid PEM")
```

romshark: actually, this isn't used anywhere but in this file so it doesn't really matter. It's fine as it is.

romshark (approved):
LGTM

PR #4822, opened 2025-08-05
pkg/scrypto/cppki: ignore trailing whitespace on PEM certificate bundles

oncilla:
Ignore trailing whitespace in PEM certificate
bundles to avoid unnecessary errors when reading
them. This is more lenient when operators are
copying files around manually. Whitespace between the PEM blocks are still not allowed.

E.g., this will allow to run `scion-pki certificate inspect` on a certificate bundle that has trailing whitespace.

### DIFF
--- a/pkg/scrypto/cppki/certs.go
+++ b/pkg/scrypto/cppki/certs.go
@@ -121,7 +121,11 @@ func ReadPEMCerts(file string) ([]*x509.Certificate, error) {
 // ParsePEMCerts parses the PEM encoded certificate blocks in raw.
 func ParsePEMCerts(raw []byte) ([]*x509.Certificate, error) {
 	var certs []*x509.Certificate
-	for len(raw) > 0 {
+	if isAllASCIIWhitespace(raw) {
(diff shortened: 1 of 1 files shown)


pkg/scrypto/cppki/certs.go
```diff
@@ -121,7 +121,11 @@ func ReadPEMCerts(file string) ([]*x509.Certificate, error) {
 // ParsePEMCerts parses the PEM encoded certificate blocks in raw.
 func ParsePEMCerts(raw []byte) ([]*x509.Certificate, error) {
 	var certs []*x509.Certificate
-	for len(raw) > 0 {
+	if isAllASCIIWhitespace(raw) {
+		return nil, serrors.New("empty")
+	}
+
+	for len(raw) > 0 && !isAllASCIIWhitespace(raw) {
```
katyatitkova, line 128: The error is already returned above if `isAllASCIIWhitespace(raw)`, so the check for it is redundant

katyatitkova (changes requested):

Issue #4787, opened 2025-06-18
doc: improve SIG description & document SGRP


nicorusti:
Based on our 2025 hackathon and [feedback received from IETF ISE reviewers](https://github.com/scionassociation/scion-cp_I-D/issues/116), the SIG doc needs a bit of an overhaul.

## [SIG page](https://docs.scion.org/en/latest/sig.html)
- It should be called something like SIG framing specification
- SGRP should also be described there (replacing `The description of that protocol is yet to be written`)
- Clarify when it is necessary in sentence below. E.g. when the path is not cached??
   > If necessary, the ingress SIG performs SCION path lookups and selects a SCION path to the egress SIG.

- Clarify which 6 tuple is used in 
  > The ingress SIG encapsulates IP packets that cannot be proven independent (e.g., with the same IP 6-tuple) in the same stream.

## Gateway [manual](https://docs.scion.org/en/latest/manuals/gateway.html) 
- It should be called SIG manual and include some deployment examples
- The two pages should be cross-linked 
- Document command `scion-ip-gateway sample config`
 
[Slack discussion](https://scionproto.slack.com/archives/C07EDUQSE6L/p1749829142506639?thread_ts=1722849539.667079&cid=C07EDUQSE6L)
[Anapaya docs on SGRP](https://docs.anapaya.net/en/latest/gate/configuration/scion-tunneling/#gate-tunneling-sgrp)

katyatitkova:
> Clarify when it is necessary in sentence below. E.g. when the path is not cached??
> 
> If necessary, the ingress SIG performs SCION path lookups and selects a SCION path to the egress SIG.

From what I see, this "If necessary" should be removed altogether since the paths are retrieved from the daemon regularly: https://github.com/scionproto/scion/blob/ea645ec789c05027a6cb1fa226e7c954119429c0/gateway/pathhealth/remotewatcher.go#L122-L138
and the path selection doesn't happen on send: https://github.com/scionproto/scion/blob/ea645ec789c05027a6cb1fa226e7c954119429c0/gateway/dataplane/session.go#L87

katyatitkova:
6-tuple is a 5-tuple: protocol, src IP, dst IP, src port, dst port: https://github.com/scionproto/scion/blob/ea645ec789c05027a6cb1fa226e7c954119429c0/gateway/dataplane/session.go#L191-L225

PR #4825, opened 2025-08-15
Add SGRP section

knmeynell:


doc/sig.rst
```diff
@@ -117,3 +119,27 @@ Following example shows three IP packets packed into three SIG frames::
   +----------------------------+------------+
   | SIG HDR Index=0xffff Seq=2 | ...payload |
   +----------------------------+------------+
+
+
+SCION Gateway Routing Protocol (SGRP)
+=====================================
+
+The SCION Gateway Routing Protocol (SGRP) enables IP-in-SCION tunneling endpoints (should this be SIGs?) to map IP prefixes to SCION ASes.
```
katyatitkova: The section in parentheses looks like it shouldn't be in the doc

knmeynell: It's a question with respect to the terminology (SIG versus tunnelling endpoint) we should use. There's inconsistent usage throughout the documentation.

katyatitkova: Maybe we should write it as "tunneling endpoints (SIGs)" here, and decide on the proper usage in the docs later?

katyatitkova (approved):

katyatitkova:
@knmeynell Can we merge it?

doc/sig.rst
```diff
@@ -25,6 +25,8 @@ This protocol is designed to:
 - provide fast detection of packet loss and subsequent recovery of decapsulation for packets that weren't lost.
```
nicorusti, line 25: There are some points from #4787 that are not yet addressed: 
* SCION-IP Gateway --> SCION-IP Gateway specification
*~They then exchange IP prefixes. The description of that protocol is yet to be written.~ this should mention SGRP

There is also some work that could be done on the SIG manual https://docs.scion.org/en/latest/manuals/gateway.html

doc/sig.rst
```diff
@@ -117,3 +112,27 @@ Following example shows three IP packets packed into three SIG frames::
   +----------------------------+------------+
   | SIG HDR Index=0xffff Seq=2 | ...payload |
   +----------------------------+------------+
+
+
+SCION Gateway Routing Protocol (SGRP)
```
nicorusti: I would put this section before the frame spec, right after introduction

doc/sig.rst
```diff
@@ -117,3 +112,27 @@ Following example shows three IP packets packed into three SIG frames::
...
+=====================================
+
+The SCION Gateway Routing Protocol (SGRP) enables SIGs to map IP prefixes to SCION ASes.
+
+A SIG participating in SGRP between two SCION ASes does the following:
+
+1. It discovers the SIGs in the remote SCION AS by periodically sending a discovery message to the Control Plane of the remote AS which replies with a list of local tunneling endpoints.
```
nicorusti: @katyatitkova @romshark  I guess this is using the discovery service. There is not too much docs about it, is there a doc page we could link?

doc/sig.rst
```diff
@@ -117,3 +112,27 @@ Following example shows three IP packets packed into three SIG frames::
...
+
+2. It periodically queries each discovered SIG in the remote AS to learn the IP prefixes that it announces. From that, the local SIG builds a mapping of IP prefix to remote SIGs.
+
+3. When queried by a remote SIG, the local SIG replies with the set of IP prefixes it wants to announce.
+
+The set of announced IP prefixes can either be statically configured or can be dynamically learned via BGP.
```
nicorusti:
```suggestion
The set of announced IP prefixes cane be statically configured.
```
This sentence should link to https://docs.scion.org/en/latest/manuals/gateway.html#routing-policy-file 
And the OS SIG does NOT support BGP

doc/sig.rst
katyatitkova: Not that I know of... I checked the issues and PRs to try to see if there is something relevant there, but couldn't find it. Didn't dig too deep though.

nicorusti: Ok let's leave it like this for now then..

oncilla (approved):

nicorusti (approved):

nicorusti (approved):

PR #4831, opened 2025-09-02
doc: update Slack link

katyatitkova:


FR4NK-W:
The link in [doc/dev/contribute.rst](https://github.com/scionproto/scion/blob/master/doc/dev/contribute.rst) should also be updated

katyatitkova:
Thanks for noticing! I had searched for the previous version of the link in the code only. Interesting that one of the links was outdated for so long...

romshark (approved):
LGTM

PR #4836, opened 2025-09-24
private/storage: improve sqlite usage

### LINKED ISSUE #4835: control: pathdb WAL can grow unchecked
When running the control service we observed the WAL file of the control service growing to 29Gb (`/share/cache/control-64-2_0_0.path.db-wal`).

One suspect is that we [insert all segments in a single db transaction](https://github.com/scionproto/scion/blob/66ca97f817fd765ef796cb8764b0a81b7e14b1d3/private/segment/seghandler/storage.go#L67-L93). 
In each loop iteration we inside of the `tx.Insert` function we try to acquire a lock. 
If many StoreSegs calls happen at the same time they might contend for the lock while they are in a transaction.

We should do each segment insert in a separate db transaction.
Each insert inserts/updates a segment to the latest state that was observed. So it shouldn't be a problem if two StoreSegs calls interleave

oncilla:
This pull request makes several improvements to the usage of sqlite in our storage layer.

1. We now only depend on modernc.org/sqlite, removing the dependency on mattn/go-sqlite3. This simplifies our dependency management and removes another cgo dependency

2. We open a read and write connection pool. The read pool is limited to one open connection. In combination with the WAL journal mode this allows for concurrent reads and writes. Furthermore, we can get rid of the mutex, as the Go sql package deals with concurrency internally.

3. We set transactions to IMMEDIATE mode. This prevents the default DEFERRED mode, which only acquires a write lock when the first write statement is executed. In IMMEDIATE mode, the write lock is acquired when the transaction begins. This prevents unexpected SQLITE_BUSY errors during a transaction.

4. We periodically run a WAL checkpoint in the background. This prevents the WAL file from growing indefinitely.

At Anapaya we have used this pattern with great success and it has shown to increase performance and stability of our sqlite usage.

Inspiration:

- https://kerkour.com/sqlite-for-servers
- https://riverqueue.com/docs/sqlite#concurrency-best-practices
- https://theitsolutions.io/blog/modernc.org-sqlite-with-go

I have also run a very unscientific stress test on our internal implementation. Note that every write does significant amount of work like creating path segments (including signatures). Nevertheless, the speed-up is considerable, even with cpu cycles being wasted on signing.

--------------

	concurrentReaders = 1000
	concurrentWriters = 500

Before Refactor:

    dbtest.go:451:  8,846.0 reads/s
    dbtest.go:452:      8.6 writes/s

After refactor:

    dbtest.go:474: 25,949.4 reads/s
    dbtest.go:475:    594.5 writes/s

--------------

	concurrentReaders = 50
	concurrentWriters = 10

Before Refactor

    dbtest.go:451:  6,661.0 reads/s
    dbtest.go:452:    134.1 writes/s

After Refactor

    dbtest.go:474: 24,309.2 reads/s
    dbtest.go:475:    535.1 writes/s

fixes #4835

### DIFF
--- a/private/revcache/mock_revcache/mock.go
+++ b/private/revcache/mock_revcache/mock.go
@@ -109,27 +109,3 @@ func (mr *MockRevCacheMockRecorder) Insert(arg0, arg1 interface{}) *gomock.Call 
 	mr.mock.ctrl.T.Helper()
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Insert", reflect.TypeOf((*MockRevCache)(nil).Insert), arg0, arg1)
 }
-
-// SetMaxIdleConns mocks base method.
-func (m *MockRevCache) SetMaxIdleConns(arg0 int) {
-	m.ctrl.T.Helper()
-	m.ctrl.Call(m, "SetMaxIdleConns", arg0)
-}
-
-// SetMaxIdleConns indicates an expected call of SetMaxIdleConns.
-func (mr *MockRevCacheMockRecorder) SetMaxIdleConns(arg0 interface{}) *gomock.Call {
-	mr.mock.ctrl.T.Helper()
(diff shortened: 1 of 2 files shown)


katyatitkova (approved):
👏

PR #4837, opened 2025-09-26
[Experimental] Add specification annotations to the `pkg/addr` package, verify it in the CI with Gobra

jcp19:


katyatitkova:
I think what could be done is making this CI check optional: so it's still being run, but if it fails for any reason, it's still possible to merge the PR. Also, is there any link for Gobra 101? It could be useful if the check fails and the person seeing it is interested enough in fixing it

jcp19:
> I think what could be done is making this CI check optional: so it's still being run, but if it fails for any reason, it's still possible to merge the PR. Also, is there any link for Gobra 101? It could be useful if the check fails and the person seeing it is interested enough in fixing it

Thank you for the idea, @katyatitkova. As suggested, I made the CI checks optional (with the option `continue-on-error: true` for each verification step) and I added a bit of documentation in the Gobra worflow configuration. There, I list two sources of documentation for Gobra.

pkg/addr/fmt.go
```diff
@@ -56,29 +62,35 @@ func ParseFormattedISD(isd string, opts ...FormatOption) (ISD, error) {
 
 // ParseFormattedAS parses an AS number that was formatted with the FormatAS
 // function. The same options must be provided to successfully parse.
-func ParseFormattedAS(as string, opts ...FormatOption) (AS, error) {
+// @ trusted
+// @ requires false
+func ParseFormattedAS(a string, opts ...FormatOption) (AS, error) {
```
katyatitkova, line 67: Why renaming `as` to `a`?

pkg/slayers/path/path.go
```diff
@@ -12,27 +12,35 @@
...
 	"github.com/scionproto/scion/pkg/private/serrors"
 )
 
 // PathType is uint8 so 256 values max.
 const maxPathType = 256
 
 var (
-	registeredPaths [maxPathType]metadata
-	strictDecoding  bool = true
+	registeredPaths/*@@@*/ [maxPathType]metadata
```
katyatitkova, line 30: There's another variable in this PR where the comment is added after a space. IMO, it looks better with a space, but I don't have a preference. But I think it's best to stick to one way to "style" this comment

.github/workflows/gobra.yml
```diff
@@ -1,3 +1,18 @@
+# This file sets up a workflow that runs Gobra (github.com/viperproject/gobra)
+# on selected packages of the SCION implementation. Gobra is a program verifier,
+# i.e., it checks that the implementation of each function behaves as expected.
+# The expected behaviour of each function is provided as a formal specification
+# written in comments above the function signature. When Gobra is not able to
+# prove that a function behaves according to its specification, it generates an
+# (hopefully) informative error message, which identifies the parts of the code
+# where the issues might lie.
+# For more information on Gobra and its error messages, check the following:
+# - The Gobra Book (WIP): viperproject.github.io/gobra-book/
```
katyatitkova, line 10: A feedback about the book itself, not this particular PR: the search there doesn't work with special characters, like `@`. Fortunately for me (a complete beginner with Gobra), I noticed that one chapter has `@` in its name. But for people like me, the search bar might be a first place to go...

katyatitkova (changes requested):

pkg/addr/fmt.go
jcp19: `as` is currently a keyword of Gobra.

pkg/slayers/path/path.go
jcp19, line 30:
```suggestion
	registeredPaths /*@@@*/ [maxPathType]metadata
```

jcp19: Good point, I agree! I added a space here :)

.github/workflows/gobra.yml
jcp19: Thank you! I added a feature request in the book repository (https://github.com/viperproject/gobra-book/issues/43).

pkg/slayers/path/path.go
jcp19: Oops, turns out that this spacing is actually enforced by go-fmt. It looks like the comments after a variable identifier have different whitespacing depending on whether they occur in a single variable declaration or in declaration of multiple variables (https://go.dev/play/p/xhmKLzNbVtV).

pkg/addr/fmt.go
katyatitkova: Well, it's unfortunate considering that our code has a lot of variables called `as`... But it's OK I guess

katyatitkova:
Looks good to me. I think we could try to merge these changes to see how it works in real life. Let's see what others will say

PR #4843, opened 2025-11-06
scion-pki: allow creating voting certificates without ISD-AS

oncilla:
Only for cp-root, cp-ca, and cp-as certificates the ISD-AS is mandatory.

### DIFF
--- a/scion-pki/certs/create.go
+++ b/scion-pki/certs/create.go
@@ -202,7 +202,7 @@ A valid example for a JSON formatted template::
 			if err != nil {
 				return serrors.Wrap("parsing profile", err)
 			}
-			requireIA := ct == cppki.Sensitive || ct == cppki.Regular
+			requireIA := ct != cppki.Sensitive && ct != cppki.Regular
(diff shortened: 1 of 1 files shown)


scion-pki/certs/create.go
```diff
@@ -202,7 +202,8 @@ A valid example for a JSON formatted template::
 			if err != nil {
 				return serrors.Wrap("parsing profile", err)
 			}
-			subject, err := createSubject(args[0], flags.commonName)
+			requireIA := ct == cppki.Sensitive || ct == cppki.Regular
```
lukedirtwalker: isn't a not missing here? Sensitive and Regular are voting certificates, but according to the title for those it is optional.

nicorusti:
JFYI in the draft we have this wording: 
>    *Note*: Voting certificates are not required to include the ISD-AS
   number attribute in their distinguished name.

From https://github.com/scionassociation/scion-cppki_I-D/issues/107

scion-pki/certs/create.go
oncilla: Fixed

PR #4844, opened 2025-11-09
doc: update SCION homepage URL

ddogfoodd:
https://scion-architecture.net/ shows a message saying it moved to https://www.scion.org/


katyatitkova:
Thank you for catching this!

PR #4854, opened 2025-12-09
beaconing: fix diversity calculation

lukedirtwalker:
Currently the beacon diversity calculation was wrong. We only considered ingress interface in construction direction. However the first ASEntry will always have a zero ingress interface. Furthermore with this approach we did not really consider the last ingress interface, which was the ingress interface into the AS, which is not yet represented by an ASEntry (as the termination/extension only happens after selection).

By using the egress interfaces interfaces instead we solve both problems.


katyatitkova (approved):

control/beacon/beacon_test.go
```diff
@@ -20,14 +20,19 @@ import (
...
 )
 
 // TestBeaconDiversity tests that diversity is calculated correctly.
 func TestBeaconDiversity(t *testing.T) {
 	var tests = []struct {
-		name      string
-		beacon    []uint16
+		name   string
+		beacon []uint16
+		// mBeacon allows to manually specify the other beacon.
```
FR4NK-W, line 34: nit: Maybe also mention in the comment that `beacon` should not be set when setting `mBeacon`?  Or simply mention that `mBeacon` overrides `beacon`?

shitz (approved):

JordiSubira (approved):

PR #4855, opened 2025-12-09
control/beacon: selection algorithm becomes pluggable

lukedirtwalker:
With this commit the selection algorithm of the beacon store can now be changed with options. This allows users of the beacon store to plug flexible selection algorithms based on their needs.

### DIFF
--- a/control/beacon/store.go
+++ b/control/beacon/store.go
@@ -28,7 +28,6 @@ type usager interface {
 }
 
 type storeOptions struct {
-	chainChecker  ChainProvider
 	selectionAlgo SelectionAlgorithm
 }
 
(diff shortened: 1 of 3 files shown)


control/beacon/store.go
```diff
@@ -47,7 +48,17 @@ func (c chainCheckerOption) apply(o *storeOptions) {
 // beacons are verifiable with cryptographic material available in the local
 // trust store.
 func WithCheckChain(p ChainProvider) StoreOption {
```
JordiSubira: `WithCheckChain()` is not used anymore, should we maintain it or can we drop it? Beside, do we want to allow the double-wrapping:
```
beacon.NewBeaconStore(policies, db,
    beacon.WithSelectionAlgorithm(beacon.NewChainsAvailableAlgo(p, algo)),
    beacon.WithCheckChain(p),
)
```
Just asking, perhaps with another selection algorithm it makes sense...

JordiSubira (changes requested):

control/beacon/store.go
lukedirtwalker: Yeah, I first thought it would be nice to be backwards compatible. But I guess the only other users of this is us (Anapaya), so not a big deal if we just remove it.

Done.

JordiSubira (approved):

PR #4860, opened 2025-12-22
fix: break loop and the not select

romshark:
@oncilla Please correct me if I'm wrong but I suppose you wanted to break out of the loop in here. Currently, this `break` breaks on the `select` so it does nothing.

Second question: shouldn't it also break if `ctx` is canceled?

### DIFF
--- a/gateway/control/aggregator.go
+++ b/gateway/control/aggregator.go
@@ -97,13 +97,12 @@ func (a *Aggregator) run(ctx context.Context) error {
 	go func() {
 		defer log.HandlePanic()
 		ticker := time.NewTicker(a.ReportingInterval)
-	LOOP:
 		for {
 			select {
 			case <-ticker.C:
 				a.report()
 			case <-a.workerBase.GetDoneChan():
-				break LOOP
+				return
 			}
 		}
 	}()


gateway/control/aggregator.go
```diff
@@ -97,12 +97,13 @@ func (a *Aggregator) run(ctx context.Context) error {
 	go func() {
 		defer log.HandlePanic()
 		ticker := time.NewTicker(a.ReportingInterval)
+	LOOP:
```
oncilla:
```suggestion
```

gateway/control/aggregator.go
```diff
@@ -97,12 +97,13 @@ func (a *Aggregator) run(ctx context.Context) error {
...
 		for {
 			select {
 			case <-ticker.C:
 				a.report()
 			case <-a.workerBase.GetDoneChan():
-				break
+				break LOOP
```
oncilla:
```suggestion
				return
```

oncilla:
> @oncilla Please correct me if I'm wrong but I suppose you wanted to break out of the loop in here. Currently, this `break` breaks on the `select` so it does nothing.

I am not the author of the code. I simply open-sourced it back in the day :) (See all the co-authors on that commit)
Looking at it, yes it should probably be a return, and not a break.

> Second question: shouldn't it also break if `ctx` is canceled?

TBH, I don't know. Likely, it should be stopped, yes.
But you would need to double check the intended semantics of the worker packge.

PR #4863, opened 2026-01-05
snet: implement STUN for NAT traversal

jdslab:
### Background

SCION packet headers contain a SRC address to which packets should be returned. This address needs to be visible/reachable by the first-hop border router, assuming that the path gets simply reversed by the peer. This address may not be easy to discover if the sender is separated from the receiver by a NAT.

A solution has been designed that uses STUN for NAT traversal, allowing clients behind NATs to query the first-hop border router to obtain their external mapped address. This external address can then be used as the SRC address in the SCION header. The relevant design document can be found [here](https://docs.scion.org/en/latest/dev/design/NAT-address-discovery.html). The implementation of a STUN server at the border router has been done [here](https://github.com/scionproto/scion/pull/4740). This PR and its implementation details were originally discussed [here](https://github.com/jdslab/scion/pull/2)

### Summary of Changes

The present PR concerns the client side implementation of the NAT travelsal mechanism in the snet library. Modifications involve adding a configurable option to `SCIONNetwork` that enables transparent NAT traversal. If enabled, SCION connections obtained by `SCIONNetwork.Dial` will automatically query the first-hop border router for the external address using a STUN request. The external address is then used as the clients SRC address instead of its local address when sending packets. 

The STUN mechanism is implemented compliant with [RFC8489](https://www.rfc-editor.org/rfc/rfc8489), which includes retransmissions on timeouts. Changes to the behavior of the SCION connection's methods are kept to a minimum. Specifically, the `ReadFrom` and `WriteTo` methods closely mimick the behavior of `net.UDPconn`. The STUN mechanism happens in the background and is transparent towards client applications.

In summary, the changes in this PR allow client applications to connect to the SCION internet even in the presence of a NAT device.

### Testing

A new integration test has been added in `/acceptance/stun`. The test simulates a SCION network topology with a NAT device present. A test client behind the NAT sends a test packet using snet's `SCIONNetwork.Dial` to a test server outside of the NAT, in a different AS. The test server, in turn, responds with a packet to the client. The test succeeds if the returning packet arrives at the client without issue. As usual, this test is part of the CD/CI pipeline.

The integration test can be run, manually, using the following instructions:
 
- Set up development environment following [these instructions](https://docs.scion.org/en/latest/dev/setup.html), replacing the SCION repository with code from this PR.
- Run `make`, then `make docker-images`
- Run `bazel test --config=integration --test-output=streamed //acceptance/stun:test`

### DIFF
--- a/acceptance/daemon_vs_standalone/BUILD.bazel
+++ b/acceptance/daemon_vs_standalone/BUILD.bazel
@@ -1,0 +1,24 @@
+load("//acceptance/common:topogen.bzl", "topogen_test")
+
+# Test with standalone daemon (embedded daemon using topology files) - default
+topogen_test(
+    name = "test_standalone",
+    src = "test.py",
+    args = [
+        "--executable=end2end_integration:$(location //tools/end2end_integration)",
+    ],
+    data = ["//tools/end2end_integration"],
+    topo = "//topology:tiny.topo",
+)
+
+# Test with remote daemon (connecting to sciond via gRPC)
+topogen_test(
+    name = "test_daemon",
+    src = "test.py",
+    args = [
+        "--executable=end2end_integration:$(location //tools/end2end_integration)",
+        "--use-daemon",
+    ],
+    data = ["//tools/end2end_integration"],
+    topo = "//topology:tiny.topo",
+)
--- a/acceptance/daemon_vs_standalone/test.py
+++ b/acceptance/daemon_vs_standalone/test.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+
+# Copyright 2025 SCION Association
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Test that compares daemon vs standalone mode for end2end connectivity.
+
+This test runs the end2end_integration test in two modes:
+- Daemon mode: Uses remote SCION daemon connector (connecting via gRPC)
+- Standalone mode: Uses embedded daemon with topology files (no sciond)
+
+The daemon mode can be selected via --use-daemon flag:
+- With --use-daemon: Uses remote daemon connector (via gRPC)
+- Without --use-daemon (default): Uses standalone daemon connector
+"""
+
+from plumbum import cli
+
+from acceptance.common import base
+
+
+class Test(base.TestTopogen):
+    """
+    Tests end2end connectivity using either daemon or standalone mode.
+    """
+
+    use_daemon = cli.Flag(
+        "--use-daemon",
+        help="Use remote SCION daemon instead of standalone daemon",
+    )
+
+    def setup_start(self):
+        super().setup_start()
+        self.await_connectivity()
+
+    def _run(self):
+        ping_test = self.get_executable("end2end_integration")
+
+        if self.use_daemon:
+            print("=== Running with remote daemon (sciond) ===")
+            ping_test["-d", "-sciond", "-outDir", self.artifacts].run_fg()
+        else:
+            print("=== Running with standalone daemon ===")
+            ping_test["-d", "-outDir", self.artifacts].run_fg()
+
+
+if __name__ == "__main__":
+    base.main(Test)
--- a/demo/drkey/main.go
+++ b/demo/drkey/main.go
@@ -100,7 +100,7 @@ func realMain() int {
 		DstHost: clientAddr.Host.IP.String(),
 	}
 
-	daemon, err := daemon.NewService(scionEnv.Daemon()).Connect(ctx)
+	daemon, err := daemon.NewAutoConnector(ctx, daemon.WithDaemon(scionEnv.Daemon()))
 	if err != nil {
 		fmt.Fprintln(os.Stderr, "Error dialing SCION Daemon:", err)
 		return 1
--- a/doc/command/scion-pki/scion-pki_certificate_renew.rst
+++ b/doc/command/scion-pki/scion-pki_certificate_renew.rst
@@ -119,6 +119,9 @@ Options
                                The CAs are tried in order until success or all of them failed.
                                --ca is mutually exclusive with --remote
       --common-name string     The common name that replaces the common name in the subject template
+      --config-dir string      Directory containing topology.json and certs/ for standalone mode.
+                               If both --sciond and --config-dir are set, --sciond takes priority.
+                               Defaults to /etc/scion on Linux.
       --curve string           The elliptic curve to use (P-256|P-384|P-521) (default "P-256")
       --expires-in string      Remaining time threshold for renewal
       --features strings       enable development features ()
@@ -140,7 +143,9 @@ Options
                                and all specified remotes are tried in order until success or all of them failed.
                                --remote is mutually exclusive with --ca.
       --reuse-key              Reuse the provided private key instead of creating a fresh private key
-      --sciond string          SCION Daemon address. (default "127.0.0.1:30255")
+      --sciond string          Connect to SCION Daemon at the specified address instead of using
+                               the local topology.json (IP:Port or "default" for 127.0.0.1:30255).
+                               If both --sciond and --config-dir are set, --sciond takes priority.
       --sequence string        Space separated list of hop predicates
       --subject string         The path to the custom subject for the CSR
       --timeout duration       The timeout for the renewal request per CA (default 10s)
--- a/doc/command/scion/scion_address.rst
+++ b/doc/command/scion/scion_address.rst
@@ -35,11 +35,16 @@ Options
 
 ::
 
-  -h, --help            help for address
-      --isd-as isd-as   The local ISD-AS to use. (default 0-0)
-      --json            Write the output as machine readable json
-  -l, --local ip        Local IP address to listen on. (default invalid IP)
-      --sciond string   SCION Daemon address. (default "127.0.0.1:30255")
+      --config-dir string   Directory containing topology.json and certs/ for standalone mode.
+                            If both --sciond and --config-dir are set, --sciond takes priority.
+                            Defaults to /etc/scion on Linux.
+  -h, --help                help for address
+      --isd-as isd-as       The local ISD-AS to use. (default 0-0)
+      --json                Write the output as machine readable json
+  -l, --local ip            Local IP address to listen on. (default invalid IP)
+      --sciond string       Connect to SCION Daemon at the specified address instead of using
+                            the local topology.json (IP:Port or "default" for 127.0.0.1:30255).
+                            If both --sciond and --config-dir are set, --sciond takes priority.
 
 SEE ALSO
 ~~~~~~~~
--- a/doc/command/scion/scion_ping.rst
+++ b/doc/command/scion/scion_ping.rst
@@ -86,6 +86,9 @@ Options
 
 ::
 
+      --config-dir string      Directory containing topology.json and certs/ for standalone mode.
+                               If both --sciond and --config-dir are set, --sciond takes priority.
+                               Defaults to /etc/scion on Linux.
   -c, --count uint16           total number of packets to send
       --epic                   Enable EPIC for path probing.
       --format string          Specify the output format (human|json|yaml) (default "human")
@@ -107,7 +110,9 @@ Options
                                the total size of the packet is still variable size due to the variable size of
                                the SCION path.
       --refresh                set refresh flag for path request
-      --sciond string          SCION Daemon address. (default "127.0.0.1:30255")
+      --sciond string          Connect to SCION Daemon at the specified address instead of using
+                               the local topology.json (IP:Port or "default" for 127.0.0.1:30255).
+                               If both --sciond and --config-dir are set, --sciond takes priority.
       --sequence string        Space separated list of hop predicates
       --timeout duration       timeout per packet (default 1s)
       --tracing.agent string   Tracing agent address
--- a/doc/command/scion/scion_showpaths.rst
+++ b/doc/command/scion/scion_showpaths.rst
@@ -90,6 +90,9 @@ Options
 
 ::
 
+      --config-dir string      Directory containing topology.json and certs/ for standalone mode.
+                               If both --sciond and --config-dir are set, --sciond takes priority.
+                               Defaults to /etc/scion on Linux.
       --epic                   Enable EPIC.
   -e, --extended               Show extended path meta data information
       --format string          Specify the output format (human|json|yaml) (default "human")
@@ -101,7 +104,9 @@ Options
       --no-color               disable colored output
       --no-probe               Do not probe the paths and print the health status
   -r, --refresh                Set refresh flag for SCION Daemon path request
-      --sciond string          SCION Daemon address. (default "127.0.0.1:30255")
+      --sciond string          Connect to SCION Daemon at the specified address instead of using
+                               the local topology.json (IP:Port or "default" for 127.0.0.1:30255).
+                               If both --sciond and --config-dir are set, --sciond takes priority.
       --sequence string        Space separated list of hop predicates
       --timeout duration       Timeout (default 5s)
       --tracing.agent string   Tracing agent address
--- a/doc/command/scion/scion_traceroute.rst
+++ b/doc/command/scion/scion_traceroute.rst
@@ -79,6 +79,9 @@ Options
 
 ::
 
+      --config-dir string      Directory containing topology.json and certs/ for standalone mode.
+                               If both --sciond and --config-dir are set, --sciond takes priority.
+                               Defaults to /etc/scion on Linux.
       --epic                   Enable EPIC.
       --format string          Specify the output format (human|json|yaml) (default "human")
   -h, --help                   help for traceroute
@@ -88,7 +91,9 @@ Options
       --log.level string       Console logging level verbosity (debug|info|error)
       --no-color               disable colored output
       --refresh                set refresh flag for path request
-      --sciond string          SCION Daemon address. (default "127.0.0.1:30255")
+      --sciond string          Connect to SCION Daemon at the specified address instead of using
+                               the local topology.json (IP:Port or "default" for 127.0.0.1:30255).
+                               If both --sciond and --config-dir are set, --sciond takes priority.
       --sequence string        Space separated list of hop predicates
       --timeout duration       timeout per packet (default 1s)
       --tracing.agent string   Tracing agent address
--- a/private/app/flag/BUILD.bazel
+++ b/private/app/flag/BUILD.bazel
@@ -29,7 +29,6 @@ go_test(
     embed = [":go_default_library"],
     deps = [
         "//pkg/addr:go_default_library",
-        "//pkg/daemon:go_default_library",
         "//private/app/env:go_default_library",
         "@com_github_spf13_pflag//:go_default_library",
         "@com_github_stretchr_testify//assert:go_default_library",
--- a/private/app/flag/env.go
+++ b/private/app/flag/env.go
@@ -20,6 +20,7 @@ import (
 	"io/fs"
 	"net/netip"
 	"os"
+	"runtime"
 	"sync"
 
 	"github.com/spf13/pflag"
@@ -34,6 +35,8 @@ const (
 	defaultDaemon = daemon.DefaultAPIAddress
 
 	defaultEnvironmentFile = "/etc/scion/environment.json"
+
+	defaultConfigDirLinux = "/etc/scion"
 )
 
 type stringVal string
@@ -77,15 +80,17 @@ func (v *ipVal) String() string { return netip.Addr(*v).String() }
 // SCIONEnvironment can be used to access the common SCION configuration values,
 // like the SCION daemon address and the local IP as well as the local ISD-AS.
 type SCIONEnvironment struct {
-	sciondFlag *pflag.Flag
-	sciondEnv  *string
-	ia         addr.IA
-	iaFlag     *pflag.Flag
-	local      netip.Addr
-	localEnv   *netip.Addr
-	localFlag  *pflag.Flag
-	file       env.SCION
-	filepath   string
+	sciondFlag    *pflag.Flag
+	sciondEnv     *string
+	ia            addr.IA
+	iaFlag        *pflag.Flag
+	local         netip.Addr
+	localEnv      *netip.Addr
+	localFlag     *pflag.Flag
+	configDir     string
+	configDirFlag *pflag.Flag
+	file          env.SCION
+	filepath      string
 
 	mtx sync.Mutex
 }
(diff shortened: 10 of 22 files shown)


pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,360 @@
...
+				c.mutex.Lock()
+				pktLen := int64(len(data))
+				if c.queuedBytes+pktLen <= c.maxQueuedBytes {
+					select {
+					case c.recvChan <- dataPacket{data: data, addr: addr}:
+						c.queuedBytes += pktLen
+					default:
+					}
+				}
+				c.mutex.Unlock()
```
JordiSubira: Do we need to lock the mutex over the channel operation as well? I guess we only need the lock over `c.quedBytes`.

pkg/snet/writer.go
JordiSubira, line 121: Related to this discussion https://github.com/jdslab/scion/pull/2#discussion_r2658129356. IIANW, I believe we can have a worst corner case 2*deadline for the first packet. This may only happen if the `makeStunRequest()` takes close to deadline (but doesn't hit it) and `c.conn.WriteTo()` hits the deadline. Since we are using the same underlay for sending STUN and application data, I believe this is extremely unlikely scenario. Just commenting on this for completeness, I do not think it is worth to change the implementation for this.

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+	defer fd.Close()
+	rcvBufSize, err := syscall.GetsockoptInt(int(fd.Fd()), syscall.SOL_SOCKET, syscall.SO_RCVBUF)
+	if err != nil {
+		return nil, err
+	}
+	maxNumPacket := rcvBufSize / 64 // assuming lower bound of per packet metadata of 64 bytes
+	if maxNumPacket < 10 {
+		maxNumPacket = 10
+	}
```
marcfrei: Instead of asserting conn to *net.UDPConn, we should use conn.SyscallConn.Control here.

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,360 @@
...
+					case ch <- stunResponse{addr: mappedAddr, err: err}:
+					default:
+					}
+				}
+			} else {
+				data := make([]byte, n)
+				copy(data, buf[:n])
+				c.mutex.Lock()
+				pktLen := int64(len(data))
+				if c.queuedBytes+pktLen <= c.maxQueuedBytes {
```
marcfrei: Only use `n` up to this check and only if the condition evaluates to `true` then allocate a new byte slice and copy the packet.

pkg/snet/stun_conn.go
jdslab: Yes but given we are accessing `c.queuedBytes` twice maybe it is better to keep holding the lock during the whole duration to reduce lock release/aquire overhead? Since the select is non-blocking we are not losing a lot of performance there either. Otherwise, we would need something like this:
```
c.mutex.Lock()
if c.queuedBytes+pktLen <= c.maxQueuedBytes {
    c.mutex.Unlock()
    select {
    case c.recvChan <- dataPacket{data: data, addr: addr}:
        c.mutex.Lock()
        c.queuedBytes += pktLen
        c.mutex.Unlock()
    default:
    }
} else {
    c.mutex.Unlock()
}
```
Which is also not very readable IMO.

pkg/snet/writer.go
jdslab: I don't think this is the case. The deadline is not a time duration but an absolute time ([src](https://pkg.go.dev/net#Conn)). So `makeStunRequest()` is not extending the deadline in any way, however long up to the deadline it may take.

JordiSubira: Yes, you are right, it is absolute time. My bad, then all good.

pkg/snet/stun_conn.go
JordiSubira: Yes, we are not losing a lot of performance because the select isn't blocking. We can leave it as it is.

jdslab: Actually I realize my code in the comment above is not correct either. We need to hold the lock during the whole operation, otherwise there is a race condition between checking the `if` condition and adding `pktLen` to `queuedBytes`, which could cause inconsistencies.

JordiSubira (approved):

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,359 @@
...
+		rcvBufSize, err = syscall.GetsockoptInt(int(fd), syscall.SOL_SOCKET, syscall.SO_RCVBUF)
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	maxNumPacket := rcvBufSize / 64 // assuming lower bound of per packet metadata of 64 bytes
+	if maxNumPacket < 10 {
+		maxNumPacket = 10
+	}
```
romshark: nit:
```suggestion
	// assuming lower bound of per packet metadata of 64 bytes
	maxNumPacket := max(rcvBufSize / 64, 10)
```

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,359 @@
...
+					data := make([]byte, n)
+					copy(data, buf[:n])
+					select {
+					case c.recvChan <- dataPacket{data: data, addr: addr}:
+						c.queuedBytes += pktLen
+					default:
+					}
+				}
+				c.mutex.Unlock()
+			}
```
romshark: This is a bit safer and less brittle. Performance should not be affected.

```suggestion
			} else {
				func() {
					c.mutex.Lock()
					defer c.mutex.Unlock()

					pktLen := int64(len(buf[:n]))
					if c.queuedBytes+pktLen <= c.maxQueuedBytes {
						data := make([]byte, n)
						copy(data, buf[:n])
						select {
						case c.recvChan <- dataPacket{data: data, addr: addr}:
							c.queuedBytes += pktLen
						default:
						}
					}
				}()
			}
```

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,359 @@
...
+		c.mutex.Lock()
+		delete(c.stunChans, txID)
+		c.mutex.Unlock()
+	}()
+
+	var mappedAddr netip.AddrPort
+	currentRTO := initialRTO
+
+STUNLoop:
+	for i := 0; i < Rc; i++ {
```
romshark: nit:
```suggestion
	for i := range Rc {
```

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,359 @@
...
+	delete(c.pendingRequests, dest)
+	c.cond.Broadcast()
+	if err != nil {
+		c.mutex.Unlock()
+		return netip.AddrPort{}, err
+	}
+	addr := mapping.mappedAddr
+	c.mutex.Unlock()
+	return addr, nil
+}
```
romshark, line 212: Again, I suggest the use of anon funcs and `defer` to make the code less brittle:

```go
func (c *stunConn) mappedAddr(dest netip.AddrPort) (addr netip.AddrPort, err error) {
	exists := func() bool {
		c.mutex.Lock()
		defer c.mutex.Unlock()

		for {
			// Check if mapping exists and is valid
			if mapping, ok := c.mappings[dest]; ok {
				if mapping.isValid() {
					mapping.touch()
					addr = mapping.mappedAddr
					return true
				}
			}
			// Check if STUN request is already happening concurrently
			if c.pendingRequests[dest] {
				// Wait() automatically releases the mutex, waits, then reacquires it before continuing
				c.cond.Wait()
				continue // Re-check mapping
			}

			c.pendingRequests[dest] = true
			return false
		}
	}()
	if exists {
		return addr, nil
	}

	mapping, err := c.makeSTUNRequest(dest)
	c.mutex.Lock()
	defer c.mutex.Unlock()

	delete(c.pendingRequests, dest)
	c.cond.Broadcast()
	if err != nil {
		return netip.AddrPort{}, err
	}
	return mapping.mappedAddr, nil
}
```

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,359 @@
...
+			var deadlineExceeded <-chan time.Time
+			if !deadline.IsZero() {
+				timeout := time.Until(deadline)
+				if timeout <= 0 {
+					return nil, os.ErrDeadlineExceeded
+				}
+				deadlineExceeded = time.After(timeout)
+			} else {
+				deadlineExceeded = nil // no timeout
+			}
```
romshark: nit: This is quite obvious IMHO, doesn't need to be so explicit.
```suggestion
			}
```

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,359 @@
...
+			deadlineChanged := c.writeDeadlineChanged
+			c.mutex.Unlock()
+
+			var deadlineExceeded <-chan time.Time
+			if !deadline.IsZero() {
+				timeout := time.Until(deadline)
+				if timeout <= 0 {
+					return nil, os.ErrDeadlineExceeded
+				}
+				deadlineExceeded = time.After(timeout)
```
romshark: `time.After` is quite expensive. Maybe we can reuse a `time.Timer`, potentially even allocated once as member of `*stunConn` (probably can't be reused for all callers though, but at least across retransmissions)? This would reduce the number of runtime allocations.

romshark:
```go
package main

import (
	"testing"
	"time"
)

func BenchmarkTimeAfter(b *testing.B) {
	for b.Loop() {
		for range 7 {
			<-time.After(1 * time.Microsecond)
		}
	}
}

func BenchmarkReuseTimer(b *testing.B) {
	t := time.NewTimer(0)
	<-t.C // wait for initial fire

	for b.Loop() {
		for range 7 {
			t.Reset(1 * time.Microsecond)
			<-t.C
		}
	}
}
```

```
goos: darwin
goarch: arm64
pkg: benchfmt
cpu: Apple M4 Pro
BenchmarkTimeAfter-14              39968             29366 ns/op            1736 B/op         21 allocs/op
BenchmarkReuseTimer-14             42736             28290 ns/op               0 B/op          0 allocs/op
PASS
ok      benchfmt        2.624s
```

Reusing a `time.Timer` results in no garbage being generated for the GC to clean up.

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,359 @@
...
+	defer c.mutex.Unlock()
+	err := c.sysPacketConn.SetWriteDeadline(t)
+	if err == nil {
+		c.readDeadline = t
+		c.writeDeadline = t
+		close(c.readDeadlineChanged)
+		close(c.writeDeadlineChanged)
+		c.readDeadlineChanged = make(chan struct{})
+		c.writeDeadlineChanged = make(chan struct{})
+	}
```
romshark, line 319: This looks rather wasteful as it probably can be done with `sync.Cond` without runtime allocations. But I can't tell straight away how complicated that would be to do here and whether it's worth it. This is just an idea, so feel free to resolve this comment without action.

pkg/stun/stun.go
```diff
@@ -4,36 +4,78 @@
...
+	var tx TxID
+	if _, err := crand.Read(tx[:]); err != nil {
+		panic(err)
+	}
+	return tx
+}
+
+// Request generates a binding request STUN packet.
+// The transaction ID, tID, should be a random sequence of bytes.
+func Request(tID TxID) []byte {
```
romshark, line 52: Instead of forcing an allocation by design, consider the `Append` pattern, like https://pkg.go.dev/strconv#AppendBool

```go
// AppendRequest generates and appends a binding request STUN packet to b.
// The transaction ID, tID, should be a random sequence of bytes.
func AppendRequest(b []byte, tID TxID) []byte {
	const sz = headerLen+lenFingerprint
	if b == nil {
		b = make([]byte, 0, sz)
	}

	// STUN header, RFC5389 Section 6.
	b = append(b, bindingRequest...)
	b = appendU16(b, uint16(lenFingerprint)) // number of bytes following header
	b = append(b, magicCookie...)
	b = append(b, tID[:]...)

	// Attribute FINGERPRINT, RFC5389 Section 15.5.
	fp := fingerPrint(b)
	b = appendU16(b, attrNumFingerprint)
	b = appendU16(b, 4)
	b = appendU32(b, fp)

	return b
}
```

romshark: Ah, this code was copied from Tailscale, I see. Well, keep as is then I guess, up to you.

acceptance/stun/test-server/main.go
```diff
@@ -0,0 +1,102 @@
...
+		if !ok {
+			log.Printf("Failed to read packet payload")
+			continue
+		}
+
+		if int(pld.DstPort) != localAddr.Host.Port {
+			continue
+		}
+
+		log.Printf("Received data: \"%v\" from %v:%v", string(pld.Payload), pkt.Source, pld.SrcPort)
```
romshark:
```suggestion
		log.Printf("Received data: %q from %v:%v", string(pld.Payload), pkt.Source, pld.SrcPort)
```
no?

pkg/snet/stun_conn.go
jdslab: I implemented the change, reusing `time.Timer` on retransmissions and deadline changes to reduce the number of timer allocations. Allocating as member on `stunConn` does not work though, since we need separate timers for concurrent calls.

pkg/snet/stun_conn.go
jdslab: We cannot wait on `sync.Cond` directly inside a `select` statement. If we use a `Cond`, we would still need to bridge it to a channel in the waiting call, e.g. like this:

```
deadlineChanged := make(chan struct{})
go func() {
	c.mutex.Lock()
	c.readDeadlineCond.Wait()
	c.mutex.Unlock()
	deadlineChanged <- struct{}{}
}()

select {
...
```
I don't know which is better idiomatically. Any thoughts?

pkg/snet/snet.go
```diff
@@ -89,6 +90,7 @@ type SCIONNetwork struct {
 	// SCMPHandler describes the network behaviour upon receiving SCMP traffic.
 	SCMPHandler       SCMPHandler
 	PacketConnMetrics SCIONPacketConnMetrics
+	STUNEnabled       bool
```
lukedirtwalker: I think it would be good to add some documentation here

pkg/snet/writer.go
```diff
@@ -82,6 +82,34 @@ func (c *scionConnWriter) WriteTo(b []byte, raddr net.Addr) (int, error) {
 	if !ok {
 		return 0, serrors.New("invalid listen host IP", "ip", c.local.Host.IP)
 	}
+	listenHostPort := uint16(c.local.Host.Port)
+
+	// Rewrite source IP if STUN is in use
```
lukedirtwalker: I would probably move this to a separate function to keep the code here easily readable:

```go
	// Rewrite source IP if STUN is in use
	var err error
	listenHostIP, listenHostPort, err = c.rewriteSourceForSTUN(raddr, nextHop, listenHostIP, listenHostPort)
	if err != nil {
		return 0, err
	}
```

and at the end of the file:
```go

// rewriteSourceForSTUN returns the source IP and port for STUN rewriting.
// If STUN is in use and the destination is in a different IA, it returns
// the mapped address; otherwise it returns the original address unchanged.
func (c *scionConnWriter) rewriteSourceForSTUN(
	raddr net.Addr,
	nextHop *net.UDPAddr,
	listenHostIP netip.Addr,
	listenHostPort uint16,
) (netip.Addr, uint16, error) {

	scionPacketConn, ok := c.conn.(*SCIONPacketConn)
	if !ok {
		return listenHostIP, listenHostPort, nil
	}

	stunConn, ok := scionPacketConn.conn.(*stunConn)
	if !ok {
		return listenHostIP, listenHostPort, nil
	}

	var sameIA bool
	switch a := raddr.(type) {
	case *UDPAddr:
		sameIA = a.IA.Equal(c.local.IA)
	case *SVCAddr:
		sameIA = a.IA.Equal(c.local.IA)
	}

	if sameIA {
		return listenHostIP, listenHostPort, nil
	}

	nextHopIP, ok := netip.AddrFromSlice(nextHop.IP)
	if !ok {
		return netip.Addr{}, 0, serrors.New("invalid next hop IP", "ip", nextHop.IP)
	}
	nextHopIP = nextHopIP.Unmap()
	nextHopAddrPort := netip.AddrPortFrom(nextHopIP, uint16(nextHop.Port))

	mappedAddr, err := stunConn.mappedAddr(nextHopAddrPort)
	if err != nil {
		return netip.Addr{}, 0, serrors.New("Error getting mapped address for STUN", "stun", err)
	}

	return mappedAddr.Addr(), mappedAddr.Port(), nil
}
```

pkg/snet/writer.go
```diff
@@ -82,6 +82,34 @@ func (c *scionConnWriter) WriteTo(b []byte, raddr net.Addr) (int, error) {
...
+	if scionPacketConn, ok := c.conn.(*SCIONPacketConn); ok {
```
lukedirtwalker: Instead of doing those casts twice I think we should cache whether it's a stun conn or not and then simply check a boolean. Otherwise we have to do a type cast for every write...

```go
// in NewCookedConn
			hasSTUN:             hasSTUNConn(pconn),

// below in the same file
// hasSTUNConn checks if the provided PacketConn has STUN enabled.
func hasSTUNConn(pc PacketConn) bool {
	scionPacketConn, ok := pc.(*SCIONPacketConn)
	if !ok {
		return false
	}
	_, ok = scionPacketConn.conn.(*stunConn)
	return ok
}
```
then the check here becomes: (if you use my function suggestion from above)
```go
	if !c.hasSTUN {
		return listenHostIP, listenHostPort, nil
	}

	scionPacketConn := c.conn.(*SCIONPacketConn)
	stunConn := scionPacketConn.conn.(*stunConn)
```

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,368 @@
+// Copyright 2025 ETH Zurich
```
lukedirtwalker, line 1: Still have to look at this again.

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,368 @@
...
+			n, addr, err := c.sysPacketConn.ReadFrom(buf)
+			if err != nil {
+				if errors.Is(err, net.ErrClosed) ||
+					errors.Is(err, syscall.EBADF) { // bad file descriptor (connection closed)
+					close(c.recvChan)
+					return
+				}
+				continue
+			}
+			if stun.Is(buf[:n]) {
```
lukedirtwalker: Alternatively you could leverage that the parse function already calls Is:
```go
			respTxID, mappedAddr, err := stun.ParseResponse(buf[:n])
			switch err {
			case nil:
				c.mutex.Lock()
				ch, ok := c.stunChans[respTxID]
				c.mutex.Unlock()
				if ok {
					select {
					case ch <- stunResponse{addr: mappedAddr, err: err}:
					default:
					}
				}
			case stun.ErrNotSTUN:
				func() {
					c.mutex.Lock()
					defer c.mutex.Unlock()

					pktLen := int64(len(buf[:n]))
					if c.queuedBytes+pktLen <= c.maxQueuedBytes {
						data := make([]byte, n)
						copy(data, buf[:n])
						select {
						case c.recvChan <- dataPacket{data: data, addr: addr}:
							c.queuedBytes += pktLen
						default:
						}
					}
				}()
			}
```

current code is also fine.

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,368 @@
...
+}
+
+func (c *stunConn) mappedAddr(dest netip.AddrPort) (addr netip.AddrPort, err error) {
+	exists := func() bool {
+		c.mutex.Lock()
+		defer c.mutex.Unlock()
+
+		for {
+			// Check if mapping exists and is valid
+			if mapping, ok := c.mappings[dest]; ok {
```
lukedirtwalker: single if would be enoug:
```go
			if mapping, ok := c.mappings[dest]; ok && mapping.isValid() {
				mapping.touch()
				addr = mapping.mappedAddr
				return true
			}
```

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,368 @@
...
+
+func (mapping *natMapping) touch() {
+	mapping.lastUsed = time.Now()
+}
+
+func (mapping *natMapping) isValid() bool {
+	return time.Since(mapping.lastUsed) < timeoutDuration
+}
+
+func (c *stunConn) mappedAddr(dest netip.AddrPort) (addr netip.AddrPort, err error) {
```
lukedirtwalker: Couldn't you simplify this much more by using a singleflight group? (https://pkg.go.dev/golang.org/x/sync@v0.19.0/singleflight)

I think something like this should do:
```go
func (c *stunConn) mappedAddr(dest netip.AddrPort) (netip.AddrPort, error) {
	addr, ok := func() (netip.AddrPort, bool) {
		c.mutex.Lock()
		defer c.mutex.Unlock()
		// Check if mapping exists and is valid
		if mapping, ok := c.mappings[dest]; ok && mapping.isValid() {
			mapping.touch()
			return mapping.mappedAddr, true
		}
		return netip.AddrPort{}, false
	}()
	if ok {
		return addr, nil
	}

	result, err, _ := c.sg.Do(dest.String(), func() (interface{}, error) {
		return c.makeSTUNRequest(dest)
	})
    if err != nil {
        return netip.AddrPort{}, err
    }
    return result.(*natMapping).mappedAddr, nil
}
```

Also I would generally not use named return values... they are usually just confusing (but maybe that's just for me personally)

where c.sg is `sg singleflight.Group`

pkg/snet/reader.go
```diff
@@ -88,21 +88,38 @@ func (c *scionConnReader) read(b []byte) (int, *UDPAddr, error) {
...
+	// If this were ever to change, we would always fall into the following if statement, then
+	// we would like to replace this logic (e.g., using IP_PKTINFO, with its caveats).
+	if c.local.Host.AddrPort() != pktAddrPort {
+
+		// If the client is behind a NAT, the SCION packet will hold the mapped external address,
+		// which is expected to be different from the local address. To handle this case, we check
+		// whether the underlying connection is a stunConn, which indicates that NAT traversal
+		// is in use.
+		// TODO: Is it necessary to check that the address matches one of the mapped addresses?
+		scionConn, ok := c.conn.(*SCIONPacketConn)
```
lukedirtwalker: similar to the other comment I have in the writer, I think having to cast twice for checking for a stun conn is not great. Maybe a bool flag could help?

pkg/snet/stun_conn.go
```diff
@@ -0,0 +1,368 @@
...
+
+	delete(c.pendingRequests, dest)
+	c.cond.Broadcast()
+	if err != nil {
+		return netip.AddrPort{}, err
+	}
+	return mapping.mappedAddr, nil
+}
+
+func (c *stunConn) makeSTUNRequest(dest netip.AddrPort) (*natMapping, error) {
```
lukedirtwalker, line 214: version without the explicit jump label:
```go
func (c *stunConn) makeSTUNRequest(dest netip.AddrPort) (*natMapping, error) {
	txID := stun.NewTxID()
	stunRequest := stun.Request(txID)

	// values according to RFC 8489 Section 6.2.1
	// TODO: make configurable?
	const Rc = 7  // Maximum number of retransmissions
	const Rm = 16 // Multiplier for final retransmission wait time
	const initialRTO = 500 * time.Millisecond

	c.mutex.Lock()
	stunChan := make(chan stunResponse, Rc*2)
	c.stunChans[txID] = stunChan
	c.mutex.Unlock()

	defer func() {
		c.mutex.Lock()
		delete(c.stunChans, txID)
		c.mutex.Unlock()
	}()

	retransmissionTimer := time.NewTimer(0)
	retransmissionTimer.Stop()

	deadlineTimer := time.NewTimer(0)
	deadlineTimer.Stop()

	var mappedAddr netip.AddrPort
	currentRTO := initialRTO

	for i := range Rc {
		_, err := c.WriteTo(stunRequest, net.UDPAddrFromAddrPort(dest))
		if err != nil {
			return nil, err
		}

		var waitDuration time.Duration
		if i < Rc-1 {
			waitDuration = currentRTO
			currentRTO *= 2
		} else {
			waitDuration = Rm * initialRTO
		}
		retransmissionTimer.Reset(waitDuration)

		var retransmit bool
		for !retransmit {
			c.mutex.Lock()
			deadline := c.writeDeadline
			deadlineChanged := c.writeDeadlineChanged
			c.mutex.Unlock()

			if !deadline.IsZero() {
				timeout := time.Until(deadline)
				if timeout <= 0 {
					return nil, os.ErrDeadlineExceeded
				}
				deadlineTimer.Reset(timeout)
			}

			select {
			case <-retransmissionTimer.C:
				if i == Rc-1 {
					return nil, errors.New("STUN request timed out")
				}
				retransmit = true
			case <-deadlineTimer.C:
				return nil, os.ErrDeadlineExceeded
			case <-deadlineChanged:
				continue // write deadline changed, re-evaluate
			case resp := <-stunChan:
				if resp.err != nil {
					return nil, resp.err
				}
				mappedAddr = resp.addr

				if !mappedAddr.IsValid() {
					return nil, errors.New("invalid mapped address")
				}

				c.mutex.Lock()
				mapping := c.mappings[dest]
				if mapping == nil {
					mapping = &natMapping{destination: dest}
					c.mappings[dest] = mapping
				}
				mapping.mappedAddr = mappedAddr
				mapping.touch()
				c.mutex.Unlock()

				return mapping, nil
			}
		}
	}

	return nil, errors.New("STUN request timed out")
}
```

pkg/snet/writer.go
```diff
@@ -121,3 +137,47 @@ func (c *scionConnWriter) SetWriteDeadline(t time.Time) error {
 func (c *scionConnWriter) isWithinRange(port int) bool {
 	return port >= int(c.dispatchedPortStart) && port <= int(c.dispatchedPortEnd)
 }
+
+// getSTUNMappedSource returns the NAT mapped address for the source if the connection is
+// using STUN and the destination is in a different IA. Otherwise, it returns the original
+// address unchanged.
+func (c *scionConnWriter) getSTUNMappedSource(
```
marcfrei: `stunMappedSource`  instead of `getSTUNMappedSource`

lukedirtwalker (approved):

PR #180, opened 2015-06-18
pcb fixes

pszal:
- Additional Signed Data (ASD) added to PCB. It allows each AD to authenticate
  some data within beacon. Can be used for Beacon-based extensions.
- more readable and flexible parsers
- removed TRCField and its derivatives 
- moved egress revocation token from every PCBMarking to ADMarking (see #155)

Few points for future PR (comments are welcome):
- consider shorter hash, for interface,path,hop lifetime 20B is absolutely fine
- consider dropping Marking class
- renaming PCBMarking as HopMarking, and ADMarking as ADBlock
[...]

PR #4866, opened 2026-01-13
router: Priority forwarding queues

juagargi:
This PR introduces priority forwarding queues at the border router.
The main changes are:
- The `Packet` data structure now contains a `QueueIndex` field, that denotes the priority of the packet. It can be set to `WithPriority` or to `WithBestEffort`.
- The existing logic in `udpip.go` that reads the packets from the packet processor output channel, and sends them via `WriteBatch` has changed, mainly inside the function `readUpTo`, where priority packets are always read before anything else.
- Both the `send` and `readUpTo` use now circular iterators (ring buffer iterators). See clarification below.

Because we are now reading from two distinct Go channels (in priority order), the amount of time needed to send N packets with the new feature is greater than before.
To improve performance, circular iterators are used in the function `udpConnection.send`.
Accompanying benchmarks are in the PR, mainly to ease the discussion on how not to impact performance (too much) while still having priority queues.

### DIFF
--- a/demo/router_priority/BUILD.bazel
+++ b/demo/router_priority/BUILD.bazel
@@ -13,6 +13,7 @@ topogen_test(
         "--executable=sender:$(location //demo/router_priority/sender)",
     ],
     data = [
+        "tc_setup.sh",
         "//demo/router_priority/sender",
     ],
     topo = "//topology:tiny.topo",
--- a/demo/router_priority/tc_setup.sh
+++ b/demo/router_priority/tc_setup.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -ex
+
+NETWORK=$1
+RATE=$2
+
+veths=$(bridge link show | awk "/$NETWORK/{print \$2}")
+for veth in $veths
+do
+    echo $veth
+    tc qdisc add dev $veth root tbf rate $RATE latency 1ms burst 50kb mtu 10000
+done
--- a/demo/router_priority/test.py
+++ b/demo/router_priority/test.py
@@ -19,6 +19,7 @@ from collections import defaultdict
 from prometheus_client.parser import text_string_to_metric_families
 from typing import Iterable, Dict, List, Tuple
 import json
+import os
 import re
 import requests
 import subprocess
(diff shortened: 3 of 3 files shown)


router/priority/priority.go
```diff
@@ -0,0 +1,121 @@
...
+	if ok {
+		// We got a value we can return.
+		return v, ok
+	}
+	// Block until any queue has a value.
+	select {
+	case v, ok := <-queue[0]:
+		return v, ok
+	case v, ok := <-queue[1]:
+		return v, ok
```
marcfrei, line 74: This might break priority order because select implements uniform pseudo-random selection in case more than one communication becomes available at the same time. I guess we are fine with this edge case, right?

router/dataplane.go
```diff
@@ -140,23 +152,47 @@ type Packet struct {
...
-	_ [1 + is32bit*28]byte
+	// The struct padding field cannot be the last field of the struct. This is because if the
+	// helper constant _pad is zero and the field is at the end, the compiler will need to avoid
+	// aliasing this field with the next struct's pointer (e.g. in an array).
+	// Since the real last field of this struct is a byte long, this does't introduce alignment
+	// issues (and thus does not modify the final size of the struct regardless of the value
+	// of _pad). See notes for the Packet struct.
+	_ [_pad]byte
+	// Priority forwarding label: packets with more priority are forwarded first
+	QueueIndex pr.PriorityLabel
```
marcfrei: Doesn't this result in packets having `QueueIndex == WithPriority` by default, if not explicitly set? This would be somewhat counterintuitive, right? As far as I can see, there's the tension between two design choices: We want to have highest priority at index 0 in router/priority/priority.go and uninitialized fields in packets should default to 'best effort' instead of highest priority here.

marcfrei:
As already discussed in the open source contributors call, we're in an unfortunate dilemma here. On the one hand, we don't want to block progress towards the implementation of Hummingbird while on the other hand intermediate steps like this priority queue mechanism introduce additional complexity with unclear benefits: at this point, differentiation only applies on egress. Ingress and packet processing use simple FIFO queueing. A high-priority packet arriving behind best-effort packets will still wait for them to be processed first, resulting in weak end-to-end QoS guarantees when bottlenecks are elsewhere (ingress saturation, processor contention).

I'd welcome additional opinions on how to proceed with this in a constructive way.

router/priority/priority.go
juagargi: For most of the cases, it should not break the priority contract. At the point where we do the `select` on the two channels (right after the comment `// Block until any queue has a value.`), there was no packet in the priority queue and we are waiting for _any_ packet to arrive. If two packets arrive simultaneously, this thread will be awaken by the first unlocked channel. This is the common case.

The less common case involves a race between this reader blocked or about to be blocked on the `select` and a sender (or more) sending two packets, one to the priority channel and one to the best-effort one.
Although possible (this reader routine could be parked for enough cycles for the writers to write two packets to the channels), I consider this case to happen seldom enough to not have a real impact.

router/dataplane.go
juagargi: This is a good catch. The original assumption is that every processor would determine the priority for the packet, and this is not happening at the moment: the `QueueIndex` field is not reset.
I will add this in `Packet.reset()` now.

juagargi: Additionally, I am also renaming `QueueIndex` to `PriorityLabel`, as I think it is a more intuitive name.

juagargi:
It is decided to merge against a Netsec local version of the border router instead of `scionproto/master`.
Closing this PR in favor of https://github.com/netsec-ethz/scion/pull/180

PR #4872, opened 2026-01-19
cmd/scion: add help message to ping

worxli:
Add a hint to use `scion address` to determine a pingable IP address on
a SCION host.


romshark (approved):
LGTM

PR #4710, opened 2025-02-24
build: use rules_distroless instead of rules_debian_packages

katyatitkova:
`rules_debian_packages` is [deprecated](https://github.com/bazel-contrib/rules_debian_packages) in favor of `rules_distroless`.

`rules_distroless` has its own symlink-related issues, and unfortunately I could make `sh` symlink to work properly in the resulting `tester` image: it worked on my Ubuntu machine (with a special [hack](https://github.com/scionproto/scion/pull/4710/files/b7cd50c8fcc831b9de054a331109a954a3043032#diff-be07e9d0572a30579120beef09c1680b684023a138c2f47a6d9a98434b295327R45) though), but didn't work in the CI with Amazon Linux. I replaced `sh` with `bash` in the code and now it works both locally and in the CI. Suggestions on how to make it work with `sh` are very welcome!

PR #4877, opened 2026-02-05
tools/dc: use bash instead of sh in exec_tester

jdslab:
In #4710 the docker tester container was modified to use `bash` instead of `sh`. However, this was not reflected in the tools/dc `exec_tester` command. This causes the command to fail with `"sh": executable file not found in $PATH`. This PR fixes the issue.


katyatitkova:
Thanks for noticing!

Issue #4834, opened 2025-09-12
proposal: SCION endhost API


shitz:
# SCION Endhost API

## Purpose

The SCION Endhost API provides SCION endhosts with all necessary information required to send and receive SCION packets using the available SCION AS infrastructure. It is the contract between a SCION AS and its internal SCION endhosts. The SCION endhost API allows endhosts to …

* … discover the available transport underlays on which they can send/receive SCION packets.
* … retrieve path segments.
* … retrieve TRCs and SCION CP-PKI certificate chains to verify path segments.
* … optionally retrieve the different DRKey key material.

So far, the [SCION daemon API](https://github.com/scionproto/scion/blob/master/proto/daemon/v1/daemon.proto) has served a similar purpose, however, there are multiple reasons why we want to evolve this API:

1. It was clear that the UDP/IP transport underlay would only be one of multiple possible transport underlays. At Anapaya, we are working on an alternative underlay based on QUIC datagrams that can provide user authorization and seamlessly deals with changes in networks between the endhost and the local SCION infrastructure. To accommodate for this new transport underlay, we need to enable endhosts to discover the different available transport underlays.
2. The path request API has two issues:

   1. It provides the first N end-to-end paths (not path segments) between a source and destination AS without any way for a requester to influence the selection of paths.
   2. It does not provide any form of pagination thus either requiring limiting the number of paths returned or returning huge path responses.
   3. The application wasn't able to verify the path information and had to trust the SCION daemon.

   Ultimately, we want to provide the endhost with a more flexible and efficient API for path discovery based on path segments, such that endhosts can discover the paths that are most useful for them while also being able to verify the path information themselves. An additional benefit of switching to a segment-based path API is that a sender can easily incorporate multiple sources of path segments in their end-to-end path retrieval logic. For example, an endhost might create end-to-end paths made up of public segments fetched via the SCION endhost API and hidden segments fetched from some different service.

3. Because of the strong trust relationship between the endhost and the SCION daemon, the daemon was usually co-located on the same host as the application. However, this is an issue especially for mobile platforms where it is not easily possible to distribute and run a separate daemon process together with the application \- and requiring the installation of another package becomes impractical and is bad user experience. Thus, the SCION endhost stack should be daemon-less which necessitates a change to the API.

4. Client authentication and authorization is necessary for other use cases, e.g., for DRKey key retrieval and possibly others in the future.

### Specification format and protocol

[...]

PR #4884, opened 2026-02-12
doc: endhost API design doc

JordiSubira:
This PR adds the endhost API design document. It was originally introduced and the content is mostly based on the discussion in #4834. Some aspects are still open to discuss/improve:

[] How should we argue that MTU and Core information, previously returns on `AS(·)` method are not necessary anymore? In other words, I do not think any application/endhost functionality will miss them, but should we argue what is the case?
[] Following the trend on the discussion issue, I have described pages as **self-contained**, mostly to favor average clients. Nonetheless, we should have a consensus on this. We can of course, evolve the document after trying things out with the beta implementations we are doing (or going to do).

### DIFF
--- a/doc/dev/design/endhost-api.rst
+++ b/doc/dev/design/endhost-api.rst
@@ -432,7 +432,7 @@ libraries and their current status:
 - **jpan** (Java) -- supports daemon and daemon-less operation.
 - **snet** (Go) -- supports daemon and daemon-less operation.
 - **scion-sdk (Rust)** -- supports new endhost API only.
-- **csnet** -- uses a direct control service interface.
+- **csnet (C)** -- uses a direct control service interface.
 
 
 API Versioning Strategy
--- a/doc/dev/design/endhost-api.rst
+++ b/doc/dev/design/endhost-api.rst
@@ -1,0 +1,493 @@
+*******************
+SCION Endhost API
+*******************
+
+- Author(s): Samuel Hitz (main), Jordi Subirà-Nieto
+- Last updated: 2026-02-11
+- Discussion at: :issue:`4834`
+- Status: **WIP**
+
+Abstract
+--------
+
+This document proposes a new SCION Endhost API that replaces the current SCION
+daemon gRPC API. The new API provides underlay discovery for multiple transport
+types, segment-based path retrieval with pagination, cryptographic material for
+client-side path verification, and optional DRKey key retrieval. It uses
+`ConnectRPC <https://connectrpc.com/>`_ over HTTP with mandatory TLS server authentication.
+
+Background
+----------
+
+The current SCION daemon exposes a ``DaemonService`` gRPC API
+(``proto/daemon/v1/daemon.proto``) with 9 RPCs:
+
+- ``Paths`` -- return pre-built end-to-end paths to a destination.
+- ``AS`` -- return information about an AS (ISD-AS, core flag, MTU).
+- ``Interfaces`` -- return underlay addresses for local interfaces.
+- ``Services`` -- return addresses for local infrastructure services.
+- ``NotifyInterfaceDown`` -- notify the daemon of a failing interface.
+- ``PortRange`` -- return the dispatched port range for the local AS.
+- ``DRKeyASHost``, ``DRKeyHostAS``, ``DRKeyHostHost`` -- DRKey retrieval.
+
+This API was designed around a co-located daemon process that caches path
+segments and serves pre-computed end-to-end paths. Several limitations motivate
+a redesign:
+
+1. **Multiple underlays.** The current API assumes a single UDP/IP underlay.
+   Future deployments may use additional transport underlays (e.g., 
+   SNAP underlay). The API must support discovering and selecting among
+   multiple underlay types.
+
+2. **Path API limitations.** Returning pre-built end-to-end paths prevents
+   clients from verifying path segments against the CP-PKI, limits flexibility
+   in path construction, and makes it difficult to compose paths with segments
+   obtained from other sources (e.g., hidden path servers).
+
+3. **Daemon-less operation.** The requirement for a co-located daemon process
+   is a significant barrier on platforms where running a shared background
+   service is impractical (e.g., mobile platforms).
+   PRs :issue:`4868`, :issue:`4869`, :issue:`4870` and :issue:`4871` already allow for
+   self-contained applications that do not require a local daemon, but the API design still
+   inherits the limitations of the original daemon-centric model.
+
+4. **Client authentication.** Certain operations (e.g., DRKey retrieval)
+   benefit from authenticated requests. Thus the API should support client authentication.
+
+Related design documents:
+
+- :doc:`router-port-dispatch` -- dispatching based on UDP/SCION ports.
+- :doc:`endhost-bootstrap` -- automated end host bootstrapping.
+
+Proposal
+--------
+
+This section specifies the SCION Endhost API. The API is defined in proto3
+and served over ConnectRPC.
+
+Specification Format and Protocol
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The API is specified in Protocol Buffers version 3 (proto3). It is served using
+`ConnectRPC <https://connectrpc.com/>`_ over HTTP.
+
+- **HTTP/1.1** support is required.
+- **HTTP/2** and **HTTP/3** support is optional. Servers and clients may use
+  ALPN to negotiate the HTTP version.
+- All RPCs are **unary** (request-response).
+- The content type is ``application/proto``.
+
+Authentication
+^^^^^^^^^^^^^^
+
+- **Server authentication** via TLS is mandatory. The server certificate may
+  be issued by a WebPKI CA or a private TLS PKI operated by the AS.
+- **Client authentication** is optional. Any HTTP-based authentication scheme
+  (e.g., mTLS, bearer tokens) may be used. Client authentication is, for
+  instance, relevant for DRKey retrieval, where the server must authorize
+  certain requests.
+
+Endhost API Discovery
+^^^^^^^^^^^^^^^^^^^^^
+
+The mechanism for discovering the endhost API endpoint is out of scope for this
+document. The minimum bootstrap information required is the **endhost API URL**
+(scheme, host, port, and path prefix).
+
+See :doc:`endhost-bootstrap` for the general bootstrapping framework. How the
+endhost API URL is integrated into the bootstrap process is left for a
+dedicated design.
+
+Underlay Discovery
+^^^^^^^^^^^^^^^^^^
+
+The ``UnderlayService`` allows endhosts to discover available transport
+underlays in the local AS, including relevant information for them, such as,
+border router addresses, interface identifiers, and dispatched port ranges.
+
+RPC path: ``/scion.endhost.v1.UnderlayService/ListUnderlays``
+
+This replaces the daemon ``Interfaces`` and ``PortRange`` RPCs. The ``AS``
+RPC is also removed: if `isd_as` is left empty or set to 0,
+it indicates that the requester wants the underlay information for all possible local SCION ASes
+
+.. code-block:: protobuf
+
+   // ListUnderlaysRequest is used to discover available transport underlays.
+  message ListUnderlaysRequest {
+    // Include the underlays for the provided ISD-AS. This can be left empty or set to 0 which
+    // indicates that the requester wants the underlay information for all possible local SCION ASes.
+    optional uint64 isd_as = 1;
+  }
+
+   // ListUnderlaysResponse contains information about available underlays. At least one underlay
+  // must be present for endhosts to use.
+  message ListUnderlaysResponse {
+    // The UDP/IP underlay of the AS.
+    optional UdpUnderlay udp = 1;
+
+    // The SNAP underlay of the AS.
+    optional SnapUnderlay snap = 2;
+
+    // Any future underlay goes here.
+  }
+
+  // The UDP/IP underlay consisting of the available SCION routers.
+  message UdpUnderlay {
+    repeated Router routers = 1;
+  }
+
+  // Router represents a SCION routers internal interface and the associated SCION interface IDs.
+  message Router {
+    // The ISD-AS identifier
+    uint64 isd_as = 1;
+
+    // The UDP endpoint (host:port) of the internal interface.
+    string address = 2;
+
+    // The list of external SCION interface IDs on that router.
+    repeated uint32 interfaces = 3;
+
+    message PortRange{
+        // The lowest port in the SCION/UDP dispatched port range.
+        uint32 dispatched_port_start = 1;
+        // The highest port in the SCION/UDP dispatched port range.
+        uint32 dispatched_port_end = 2;
+    }
+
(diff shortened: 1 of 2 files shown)


doc/dev/design/endhost-api.rst
```diff
@@ -0,0 +1,493 @@
...
+    uint64 src_isd_as = 1;
+
+    // The destination ISD-AS the final end-to-end path.
+    uint64 dst_isd_as = 2;
+
+    // The maximum total number of segments to return.
+    // The service may return fewer than this value.
+    // If unspecified, the maximum number of segments
+    // per page is returned, which is 64.
+    int32 page_size = 3;
```
tzaeschke, line 258: I assume this is the total number of segments, not per UP/CORE/DOWN?
If all directions are sufficiently available, I will get 22xUP, 21x CORE and 21xDOWN?
Do we need to specify the distribution, should we have a preference for CORE segments (if available)?

doc/dev/design/endhost-api.rst
```diff
@@ -0,0 +1,493 @@
...
+Client-side
+"""""""""""
+
+Each client library defines its own backward-compatibility policy. Existing
+libraries and their current status:
+
+- **jpan** (Java) -- supports daemon and daemon-less operation.
+- **snet** (Go) -- supports daemon and daemon-less operation.
+- **scion-sdk (Rust)** -- supports new endhost API only.
+- **csnet** -- uses a direct control service interface.
```
tzaeschke:
```suggestion
- **csnet (C)** -- uses a direct control service interface.
```

tzaeschke (approved):
LGTM % some nitpicks

doc/dev/design/endhost-api.rst
JordiSubira: Yes, but this was my understanding, we can discuss it. I wouldn't impose any distribution, my idea is to provide leeway to the server restricted "only" to this "self-sufficient" pages. Thus the phrasing with "server policy-based" ordering of the segments, if a server believes there should be a preference for CORE segments, it is possible to have it, but not necessary.

tzaeschke:
(Copied from #4834)

Should we specify error codes?
For example, the current scion-sdk appears to produce the following error codes for a HTTP/1.1 ListSegmentsRequest:

    ISD-AS = 0 -> error 400: BAD_REQUEST
    ISD-AS = unregistered-AS -> error 500: INTERNAL_SERVER_ERROR

I think it could be useful if errors were standardized so clients know how to deal with them.

It may also okay to simply return an empty set instead of throwing an error for unknown-ASes above.
And what about the rest of the API?

JordiSubira:
> Should we specify error codes?
> And what about the rest of the API?

We can specify those errors or response codes for which we have a clear idea at the moment.

>  For example, the current scion-sdk appears to produce the following error codes for a HTTP/1.1

I guess you mean that Anapaya's beta implmentation of the Endpoint API server returns this code, right? At the end the value of this exercise (creating the document, dicussing) is also to reach an initial consensus about these details. My assumption is that Anapaya's implementation is already very close to the outcome of this PR, i.e., the "official" specification, but some smalls things may change.

> It may also okay to simply return an empty set instead of throwing an error for unknown-ASes above.

Yes, I also think an empty set should suffice.

tzaeschke:
Another question: it seems the `mtu` of the local AS is currently not in the API. Can we add that?
I am using that, for example, when creating SCION path objects for communication within the local AS. 
If we don't have it then we have the weird situation that we have MTU information only for remote destinations, but not for local destinations.

doc/dev/design/endhost-api.rst
```diff
@@ -0,0 +1,493 @@
...
+DRKey requests SHOULD be authenticated. The server uses client identity to
+authorize key derivation. See the `Authentication`_ section for available
+authentication mechanisms.
+
+Obsoleted Daemon RPCs
+^^^^^^^^^^^^^^^^^^^^^
+
+The following daemon RPCs are obsoleted by the endhost API:
+
+- ``AS`` -- The local ISD-AS identifier, core flag, and MTU is omitted from the endhost API.
```
tzaeschke, line 399: I am in favor of having MTU information in the API. I think it is useful, for example, when communicating with endhosts in the local AS.

doc/dev/design/endhost-api.rst
```diff
@@ -0,0 +1,493 @@
...
+This replaces the daemon ``Paths`` RPC.
+
+.. code-block:: protobuf
+
+   message ListSegmentsRequest {
+    // The source ISD-AS of the final end-to-end path. This is most likely
+    // the requester's own ISD-AS.
+    uint64 src_isd_as = 1;
+
+    // The destination ISD-AS the final end-to-end path.
```
tzaeschke, line 251: It would be good to specify the behavior in case of ISD=0 or ISD=unknown. Do we get an empty set or an error? Which error?

doc/dev/design/endhost-api.rst
tzaeschke: Regarding the ISD/AS, I understand this is actually available from querying the underlay without ISD/AS code in the request, which will return Router information with ISD/AS attached.

JordiSubira: Yes, both UDP/IP underlay and SNAP return this information.

tzaeschke: Thinking about this a bit more, I would still find it useful to get the ISD/AS information explicitly. Imagine a local test AS without border routers, or a situation where all border routers are disabled. If that happens, having no access to the ISD/AS info precludes us from communicating with other endhosts in the same AS (they may reasonably expect the correct ISD/AS in the SCION header, but the client has no way of getting it).

doc/dev/design/endhost-api.rst
```diff
@@ -0,0 +1,493 @@
...
+    repeated uint32 interfaces = 3;
+
+    message PortRange{
+        // The lowest port in the SCION/UDP dispatched port range.
+        uint32 dispatched_port_start = 1;
+        // The highest port in the SCION/UDP dispatched port range.
+        uint32 dispatched_port_end = 2;
+    }
+
+    // The dispatched port range supported by the router. UDP/SCION packets with a destination
```
tzaeschke, line 205: I suggest changing the name to something like "direct_ports", "non_forwarded_ports" or "dynamically_assigned_ports". "Dispatch" just means to "send", I think this name is based an history and is not very descriptive.

I would suggest avoiding the term "dispatched port", it is difficult to understand, especially without knowing historical context.

doc/dev/design/endhost-api.rst
```diff
@@ -0,0 +1,493 @@
...
+    // port within this range will be sent to the same UDP destination port on the destination host.
+    // For other ports, the router will use the well-known dispatcher port (30041). If not set,
+    // the entire port range is assumed to be dispatched.
```
tzaeschke, line 208: I think it is unclear means for "the entire port range to be dispatched".
I assume "entire range" means "dispatched_ports == NONE"? 

I would actually suggest a semantic change that should allow us to phase out this property. How about we change the definition to "Not present -> dispatched_ports == ALL"?

With the topology file API, we cannot do that because old client may require dispatched ports and may not know about the setting. But with the new endhost API, we can assume that every client that uses it understands that not-present means ALL. 
We can then gradually phase out old client and eventually, once everybody uses the new API, we can deprecate this field and stop supporting it.

This semantic difference would be communicated clearer if we gave the field a different name.

doc/dev/design/endhost-api.rst
JordiSubira: There's always some complication changing the name of a field/term. This field was already present in the `topology.json` file. The `topolgy.json` file will get removed from endhosts with the new Endhost API, but infrastracture nodes, such as the BRs may continue using it. Thus, changing the field name has some "consistency" implications.

doc/dev/design/endhost-api.rst
JordiSubira: Similar as the comment above, this may introduce a consistency problem.

doc/dev/design/endhost-api.rst
JordiSubira: I would refer to ConnectRPC codes (https://connectrpc.com/docs/protocol#error-codes). For requests that are going to be always invalid like "ISD=0" I would use `invalid_argument` == 400. For queries that are correct but do not match anything, I would return an empty set only (i.e., no error).

doc/dev/design/endhost-api.rst
tzaeschke: Yes, it may. I think the bigger problem is that we either keep the dispatched_ports feature or remove it. 
If we keep it, we need to specify it, possibly in an RFC, and my guess it that reviewers won't like the name because it is not very descriptive. 
If we want to remove it, we have to change the semantics (currently non-present means everything goes to 30041). If we change the semantics, the least we should do is use a different name (see my comment further down). 

So either way, I think we have to rename it. I'd argue better rename it now than later.

doc/dev/design/endhost-api.rst
tzaeschke: Also, I am not sure why the ISD/AS number is given per BR. Has that to do with Secure Access Groups?
This is a bit off-topic now, but for the Private ISD concept, it would be useful to have multiple ISD/AS per BR (one per interface). However, it is not necessary to provide this information in the API, the endhost can get the ISD/AS, respectively interface and border router, from the segment information.

So I am in favor of having the ISD/AS list preferrably in the `ListUnderlaysResponse` or the `UdpUnderlay`, so it is available even if no router is available.

doc/dev/design/endhost-api.rst
tzaeschke: > Similar as the comment above, this may introduce a consistency problem.

Not sure how this is inconsistent. Legacy applications would anyway need either a mapping layer or keep accessing using the old control service API. These would still use the empty range as default range.

doc/dev/design/endhost-api.rst
```diff
@@ -0,0 +1,493 @@
...
+TRC and Certificate Retrieval
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``TrustService`` provides access to TRCs and certificate chains, enabling
+clients to verify path segments against the SCION CP-PKI.
+
+ListChains
+""""""""""
+
+RPC path: ``/scion.endhost.v1.TrustService/ListChains``
```
rohrerj, line 282: The design document is not specifying the return type of this RPC endpoint. My suggestion would be to use something like this:
```
message ListChainResponse {
    repeated Chains list_chain = 1;
}

message Chains {
    // List of chains that match the subject.
    repeated Chain chains = 1;
    // The corresponding subject.
    Subject subject = 2;
}

message Chain {
    // AS certificate in the chain.
    bytes as_cert = 1;
    // CA certificate in the chain.
    bytes ca_cert = 2;
}
```

shitz: So the original proposal (https://github.com/scionproto/scion/issues/4834) has this:
```protobuf
message ListChainsResponse {
  // List of chains that match the request.
  repeated Chain chains = 1;
}

message Chain {
  // The subject corresponding to this chain.
  Subject subject = 1;

  // AS certificate in the chain.
  bytes as_cert = 2;

  // CA certificate in the chain.
  bytes ca_cert = 3;
}
```

I would keep the flattened structure and not introduce a list of chains per subject.

Furthermore, I think it doesn't hurt if we also add pagination support, i.e., 

```protobuf
// Request to fetch certificate chains for a given AS certificate subject.
message ListChainsRequest {
  // The subjects for which the chains should be returned.
  repeated Subject subjects = 1;

  // Point in time at which the AS certificate must still be valid. In seconds
  // since UNIX epoch.
  uint32 at_least_valid_until = 2;

  // Point in time at which the AS certificate must be or must have been
  // valid. In seconds since UNIX epoch.
  uint32 at_least_valid_since = 3;

  // The maximum total number of chains to return.
  // The service may return fewer than this value.
  // If unspecified, the maximum number of segments
  // per page is returned, which is 64.
  int32 page_size = 4;

  // A page token, received from a previous
  // ListChainsRequest call.
  // Provide this to retrieve the subsequent page.
  //
  // When paginating, all other parameters provided in
  // ListChainsRequest must match the call that provided
  // the page token.
  string page_token = 5;
}
```
and the reponse

```protobuf
message ListChainsResponse {
  // List of chains that match the request.
  repeated Chain chains = 1;
  // The token for the next page of results.
  string next_page_token = 2;
}
```

doc/dev/design/endhost-api.rst
```diff
@@ -0,0 +1,493 @@
...
+
+Path Segment Retrieval
+^^^^^^^^^^^^^^^^^^^^^^
+
+The ``PathService`` provides paginated access to path segments (up, core, down)
+that the client can combine to construct end-to-end paths. Returning segments
+instead of pre-built paths enables client-side verification against the CP-PKI
+and composability with segments from other sources.
+
+RPC path: ``/scion.endhost.v1.PathService/ListSegments``
```
rohrerj, line 240: Suggested change: Rename `PathService` to `SegmentsService`, since it returns segments and not paths.

tzaeschke: It actually returns path segments, so to be precise it would be `PathSegmentService`?
But I don't think it's that bad to have it shortened to `PathService`. I think either is ok.

tzaeschke:
@shitz @d-s-d @lukedirtwalker We currently have @rohrerj working on an implementation for the scionproto reference implementation. It is close to being finished, so maybe we could push a bit on getting this design finalized?

I'm not sure whether you have more changes from your side?

doc/dev/design/endhost-api.rst
tzaeschke: It seems this has already been implemented. 
Nitpick: `SegmentsService` uses plural, `UnderlayService` uses singular. It should probably be `SegmentSevice` (or `PathSegmentService`).

doc/dev/design/endhost-api.rst
```diff
@@ -0,0 +1,493 @@
...
+    int32 page_size = 3;
+
+    // A page token, received from a previous
+    // ListSegmentsRequest call.
+    // Provide this to retrieve the subsequent page.
+    //
+    // When paginating, all other parameters provided in
+    // ListSegmentsRequest must match the call that provided
+    // the page token.
+    string page_token = 4;
```
rohrerj, line 267: I would suggest to specify what the expected response should be once the last page has been queried. I would suggest to return an empty string as page_token in that case, since querying using empty string would return the first page, and would therefore "close" the loop.

tzaeschke:
@JordiSubira Hi Jordi, how are you? 
Would you mind giving write access to this fork to @d-s-d ([d-s-d](https://github.com/d-s-d) = [Stefan Dietiker](https://github.com/d-s-d) from Anapaya)?
Then we could keep the discussion while updating the branch.
Or is there another solution?

JordiSubira:
> @JordiSubira Hi Jordi, how are you? Would you mind giving write access to this fork to @d-s-d ([d-s-d](https://github.com/d-s-d) = [Stefan Dietiker](https://github.com/d-s-d) from Anapaya)? Then we could keep the discussion while updating the branch. Or is there another solution?

All good Til, hope you too. I sent both of you an invitation ( I believe it should suffice with that).

tzaeschke:
@JordiSubira Good to hear, and thanks for the write access!

PR #4888, opened 2026-02-17
fix: correctly apply pr #4872

romshark:
Running `make all` deletes the change made in #4872.
CI shouldn't have allowed that to happen but it did for yet unknown reasons.


romshark:
So the reason is probably that it didn't run `make all` at all https://buildkite.com/scionproto/scion/builds/6170 and couldn't have notice that change.

romshark:
But the contributor couldn't know that this doc file is actually generated. I consider this the actual problem.

PR #4892, opened 2026-02-20
router: afxdp underlay

romshark:


jiceathome:
Whoohoo! I don’t have approval powers, but thanks to Church-Turing separation I can still use my self-bestowed blessing powers: ship it! Thanks for making good the last project of my career. 🎉

PR #4895, opened 2026-03-03
doc: document UDP underlay and ports

nicorusti:
This PR contains a description of the IP/UDP underlay. This page will be added as an informative reference in the SCION dataplane RFC.

During the IETF review process, it emerged that we need to document the behavior of default port UDP 30041, which is required to be unique so that endhosts (endpoints in the drafts) can receive SCMP informational messages. 
The UDP/IP underlay behavior in the nearly final version [draft-dekater-scion-dataplane-11](https://datatracker.ietf.org/doc/draft-dekater-scion-dataplane/11) is not well specified. 

We discussed wether to include this in the draft, and we decided that it is best to leave it out to an external reference. Mian reasons were: 
- this is currently an evolving topic (dispatcherless, STUN, etc...) and we don't want behaviors set in stone in the RFC
- this document would better fit an Internet Draft of its own  (e.g. SCION on top of IP behavior)   

For the record, discussions happened in: 
-  https://github.com/scionassociation/scion-cp_I-D/issues/213
-  slack ([SCMP](https://scionproto.slack.com/archives/C8ADA9CEP/p1770398184282579), [underlay](https://scionproto.slack.com/archives/C8A08D5FB/p1772190585855269))
- https://github.com/scionassociation/scion-dp_I-D/issues/4 (where we'll attempt to register port 30041 with IANA)

### DIFF
--- a/doc/protocols/underlay.rst
+++ b/doc/protocols/underlay.rst
@@ -107,6 +107,8 @@ SCION border routers utilize specific underlay ports to process and forward traf
 Control Plane Instances
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Control plane components of different ASes communicate with each other through RPC messages that are transported via Connect RPC. This protocol carries messages over HTTP/3, that uses a QUIC transport layer. Identification of the relevant addresses and ports for inter-domain queries is provided by `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_.
+Control plane components communicate with each other and with end-hosts through RPC messages transported via Connect RPC.
 
-For intra-domain communication of endpoints with service instances, the operator may use arbitrary ports, that have to be communicated to endpoints. For a comprehensive list of ports used by this implementation, refer to the :ref:`control-port-table`.
+For inter-AS communication, traffic is transported over ConnectRPC/HTTP3/QUIC/UDP/SCION/UDP/IP between control service instances. Identification of the relevant control service addresses and ports is provided by `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_.
+
+For intra-AS communication, traffic is transported over ConnectRPC/HTTP/TCP/IP between end-hosts and control service instances. The port used by control service instances is configured by operators, and configured on end-hosts. For a comprehensive list of ports used by this implementation, refer to the :ref:`control-port-table`.


doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
...
+
+
+Protocol Stack Summary
+----------------------
+
+A visual summary of the overall SCION protocol stack is shown below:
 
 .. figure:: fig/stack.excalidraw.png
+
+The current implementation supports an IP/UDP underlay. However, other underlay protocols (e.g., MPLS) are possible in principle.
```
tzaeschke:
```suggestion
The current implementation supports an IP/UDP underlay. However, other underlay protocols (e.g., MPLS) are possible in principle.

```

tzaeschke: Linter complains about missing final new-line.

doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
...
+
+In the modern "dispatcherless" design (see :doc:`Router Port Dispatch <../dev/design/router-port-dispatch>`), applications open a UDP/IP underlay socket directly. To ensure that traffic goes to the correct application, the ingress router at the destination AS MUST select the underlay destination port by inspecting the Layer 4 destination port from the TCP/SCION, UDP/SCION, or SCMP Error payload:
+
+* **Dispatched Ports**: If the port falls within the configured dispatched ports range, the router forwards the packet to the end host using that same port as the UDP underlay destination port.
+* **UDP underlay default port**: If the port falls outside the configured range, or if the system is handling legacy fallback traffic (i.e., traffic from older clients that do not support dynamic port mapping), the router forwards the packet to the default end-host data port: **30041**.
+* **SCMP**: For SCMP informational messages, or if the port cannot be extracted, the router forwards the packet to the default end-host port: **30041**.
```
tzaeschke:
```suggestion
* **SCMP**: For SCMP informational messages, the router forwards the packet to the default end-host port: **30041**. For SCMP error messages, the router tries to extract the end-host port and sends the packet to the endhost, subject to dispatched port settings and the default underlay port. If the port cannot be extracted, the packet is dropped.
```

doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
...
+
+.. note::
+   Historically, SCION end hosts relied on a user-space "dispatcher" process listening on the default port UDP 30041 to route incoming packets to the correct application socket. For more details, see the :doc:`Dispatcher Manual <../manuals/dispatcher>`.
+
+The SCMP Daemon (``scmpd``) MUST listen on the UDP underlay default port (30041) to process and reply to informational SCMP messages, such as echo requests (pings) and traceroutes.
```
tzaeschke:
```suggestion
The SCMP Daemon (``scmpd``) SHOULD listen on the UDP underlay default port (30041) to process and reply to informational SCMP messages, such as echo requests (pings) and traceroutes.
```

doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
...
+
+* **Internal Interfaces**: Used for intra-AS communication between routers and with end hosts. A range of ports can be used for multiple interfaces.
+* **External Interfaces**: Used for inter-AS links facing neighboring SCION ASes. These can be arbitrarily configured to match the neighboring router's requirements.
+
+Control Plane Instances
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Control plane components require service ports for topology synchronization, beaconing, and PKI tasks:
+
+* **Control Plane Intra-AS**: Control services typically communicate over TCP/IP and UDP/SCION on ports **40000-40099**.
```
JordiSubira: Control Services comunicate with whom over UDP/SCION on those ports. This is not clear to me and it also contradicts the values on the table, does not it?

doc/protocols/stack.rst
tzaeschke: I think it is not a MUST, or is it?

doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
...
+End hosts send traffic to the SCION router's Internal Interface address and port. This underlay port must be configured on the endpoints via the ``topology.json`` file.
+
+STUN Support and NAT Address Discovery
+--------------------------------------
+
+End hosts located behind a Network Address Translation (NAT) device face a unique challenge: the source address they encode must be the external IP and port visible to the first-hop border router, rather than their internal local address.
+
+To resolve this, SCION incorporates a :doc:`NAT IP/port discovery mechanism <../dev/design/NAT-address-discovery>` conceptually similar to the STUN (Session Traversal Utilities for NAT) protocol, operating directly between clients and border routers. The border router acts as a detector; when the client sends a discovery request, the border router observes the NAT-mapped IP and port and reports it back to the client.
+
+The end host can then reliably inject this public, border-router-visible IP and port into the SCION source address fields of its outbound packets. This guarantees that return traffic from the remote destination can be successfully routed back through the NAT to the client.
```
tzaeschke, line 97:
```suggestion
The end host can then reliably inject this public border-router-visible IP and port into the SCION source address fields of its outbound packets. This guarantees that return traffic from the remote destination can be successfully routed back through the NAT to the client.
```

tzaeschke: "guarantee" is a bit too strong for my taste, all kinds of thing could go wrong.
Maybe "facilitates"?

doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
...
+| UDP underlay dispatched ports           | any        | UDP 31000-32767 | AS-wide              | :doc:`topology.json <../manuals/common>`                                                                                          |
++-----------------------------------------+------------+-----------------+----------------------+-----------------------------------------------------------------------------------------------------------------------------------+
+| Router Internal Interfaces              | any        | UDP 30100-30199 | Router-wide          | :doc:`topology.json <../manuals/common>`                                                                                          |
++-----------------------------------------+------------+-----------------+----------------------+-----------------------------------------------------------------------------------------------------------------------------------+
+| Router External Interfaces              | any        | UDP 31000-39999 | Link                 | :doc:`topology.json <../manuals/common>`                                                                                          |
```
JordiSubira, line 61: What are we describing ports that are not relevant for the intra-AS UDP/IP underlay here?

doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
...
+
+The end host can then reliably inject this public, border-router-visible IP and port into the SCION source address fields of its outbound packets. This guarantees that return traffic from the remote destination can be successfully routed back through the NAT to the client.
+
+Routers
+~~~~~~~
+
+SCION border routers utilize specific underlay ports to process and forward traffic:
+
+* **Internal Interfaces**: Used for intra-AS communication between routers and with end hosts. A range of ports can be used for multiple interfaces.
+* **External Interfaces**: Used for inter-AS links facing neighboring SCION ASes. These can be arbitrarily configured to match the neighboring router's requirements.
```
tzaeschke:
```suggestion
* **External Interfaces**: Used for inter-AS links facing neighboring SCION ASes. Underlay and ports can be freely configured to match the neighboring router's requirements.
```

doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
...
+| Router Internal Interfaces              | any        | UDP 30100-30199 | Router-wide          | :doc:`topology.json <../manuals/common>`                                                                                          |
++-----------------------------------------+------------+-----------------+----------------------+-----------------------------------------------------------------------------------------------------------------------------------+
+| Router External Interfaces              | any        | UDP 31000-39999 | Link                 | :doc:`topology.json <../manuals/common>`                                                                                          |
++-----------------------------------------+------------+-----------------+----------------------+-----------------------------------------------------------------------------------------------------------------------------------+
+| Control Plane Intra-AS                  | any        | UDP/TCP 30252   | AS-wide              | `Control Port Table <../manuals/control.html#port-table>`_                                                                        |
```
JordiSubira: Same as above, how is this relevant for the UDP/IP underlay for SCION packets, the communication from client to CS is over TCP/IP (or at least as far as I remember right now).

doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
...
++-----------------------------------------+------------+-----------------+----------------------+-----------------------------------------------------------------------------------------------------------------------------------+
+| Control Plane Inter-AS                  | any        | Dynamic         | AS-wide              | `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_   |
```
JordiSubira: Again, why is this relevant? Packets to the CS are dispatched similarly than to endhost right? Or am I forgetting something?

doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
...
+* **Internal Interfaces**: Used for intra-AS communication between routers and with end hosts. A range of ports can be used for multiple interfaces.
+* **External Interfaces**: Used for inter-AS links facing neighboring SCION ASes. These can be arbitrarily configured to match the neighboring router's requirements.
+
+Control Plane Instances
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Control plane components require service ports for topology synchronization, beaconing, and PKI tasks:
+
+* **Control Plane Intra-AS**: Control services typically communicate over TCP/IP and UDP/SCION on ports **40000-40099**.
+* **Control Plane Inter-AS**: Control plane traffic across AS boundaries relies on QUIC/SCION, with the exact ports dynamically chosen by the service.
```
tzaeschke: Should we mention service addresses here?

doc/protocols/stack.rst
tzaeschke:
```suggestion
The current implementation supports an UDP/IP underlay. However, other underlay protocols (e.g., MPLS) are possible in principle.

```

doc/protocols/stack.rst
```diff
@@ -1,9 +1,109 @@
-********************
-SCION Protocol Stack
-********************
+***************
+IP/UDP underlay
+***************
 
 .. _stack:
```
tzaeschke, line 5: Why is this document called "stack"? Maybe rename it to "underlay"?

doc/protocols/stack.rst
tzaeschke: I think the idea is that inter-AS traffic may also some an underlay?

tzaeschke: But I agree with Jordi, this document should probably focus on intra AS underlays.

tzaeschke: More specifically, on communication between BR<->endhost and BR<->BR

doc/protocols/stack.rst
nicorusti: > "guarantee" is a bit too strong for my taste 

Right, especially with NATs and middleboxes, where all sort of things can happen. How about:
```suggestion
The end host can then reliably inject this public border-router-visible IP and port into the SCION source address fields of its outbound packets. This enables return traffic to traverse the NAT from the remote destination back to the client.
```

doc/protocols/stack.rst
nicorusti: You are right... I asked Gemini to create the initial version of this page, I thought I fixed most inaccuracies but I missed this one. I'll change

nicorusti: I think the AI got confused with monitoring ports from [here](https://docs.scion.org/en/latest/manuals/control.html#port-table)

doc/protocols/stack.rst
nicorusti: My idea was that this page generally describes how SCION is stacked over IP/UDP. This should IMHO include inter-AS links. I reworded to clarify that  inter-AS links are independent from intra-AS links.

doc/protocols/stack.rst
nicorusti: Absolutely, this is a missing point and we could take some text from the [draft](https://www.ietf.org/archive/id/draft-dekater-scion-controlplane-15.html#name-control-service-discovery) or from the [design docs](https://docs.scion.org/en/latest/dev/design/router-port-dispatch.html#service-addresses) to this page

doc/protocols/stack.rst
nicorusti: You are right. Let me add a paragraph mentioning this and remove them from the table.

doc/protocols/stack.rst
nicorusti: removed control plane lines

nicorusti:
@tzaeschke @JordiSubira I made some changes based on your feedback. How does it look now? 
Feel free to edit directly. 
It would be great if we can merge it by March 16 so  that I can have the link ready for the reference in the draft. We can always fine tune it later.

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,103 @@
...
+
+* **Dispatched Ports**: If the port falls within the configured dispatched ports range, the router forwards the packet to the end host using that same port as the UDP underlay destination port.
+* **UDP underlay default port**: If the port falls outside the configured range, or if the system is handling legacy fallback traffic (i.e., traffic from older clients that do not support dynamic port mapping), the router forwards the packet to the default end-host data port: **30041**.
+* **SCMP**: For SCMP informational messages, the router forwards the packet to the default end-host port: **30041**. For SCMP error messages, the router tries to extract the end-host port and sends the packet to the endhost, subject to dispatched port settings and the default underlay port. If the port cannot be extracted, the packet is dropped.
+
+.. note::
+   Historically, SCION end hosts relied on a user-space "dispatcher" process listening on the default port UDP 30041 to route incoming packets to the correct application socket. For more details, see the :doc:`Dispatcher Manual <../manuals/dispatcher>`.
```
JordiSubira, line 80: This is incomplete if it intends to refer the current behaviour. This is true for requests, for replies the ID field is inspected to dispatched the port directly to the underlay client port (if within the dispatched range)

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,103 @@
...
+STUN Support and NAT Address Discovery
+--------------------------------------
+
+End hosts located behind a Network Address Translation (NAT) device face a unique challenge: the source address they encode must be the external IP and port visible to the first-hop border router, rather than their internal local address.
+
+To resolve this, SCION incorporates a :doc:`NAT IP/port discovery mechanism <../dev/design/NAT-address-discovery>` conceptually similar to the STUN (Session Traversal Utilities for NAT) protocol, operating directly between clients and border routers. The border router acts as a detector; when the client sends a discovery request, the border router observes the NAT-mapped IP and port and reports it back to the client.
+
+The end host can then reliably inject this public, border-router-visible IP and port into the SCION source address fields of its outbound packets. This guarantees that return traffic from the remote destination can be successfully routed back through the NAT to the client.
+
+Routers
```
JordiSubira, line 99: Why is this in a subsection of STUN?

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,103 @@
...
+~~~~~~~
+
+SCION border routers utilize specific underlay ports to process and forward traffic:
+
+* **Internal Interfaces**: Used for intra-AS communication to receive traffic from end-hosts. Operators can choose the port freely. The same port must be configured on endpoints so that they can send outbound traffic. Routers with multiple internal interfaces can use a range of ports.
+* **External Interfaces**: Used for inter-AS links towards neighboring SCION ASes. Note that the choice of underlay protocol and UDP port is per link. It is independent from other links and intra-AS underlay. 
+
+Control Plane Instances
```
JordiSubira, line 107: Same, should be a separate section or not within STUN section.

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,103 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Control plane components communicate  through RPC messages that are transported via Connect RPC. This protocol carries messages over HTTP/3, that uses a QUIC transport layer.  Identification of the relevant addresses and ports for inter-domain queries is provided by  `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_. 
+
+For intra-domain control service instances, the operator may use arbitrary ports, that have to be communicated to endpoints. For a comprehensive list of ports used by this implementation, refer to the `Control Port Table <../manuals/control.html#port-table>`_.
```
JordiSubira: Again, if we removed intra-domain because it is TCP/IP communication why are we having this paragraph?

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,103 @@
...
+
+The SCMP Daemon (``scmpd``) SHOULD listen on the UDP underlay default port (30041) to process and reply to informational SCMP messages, such as echo requests (pings) and traceroutes.
+
+
+Traffic from End-hosts
+~~~~~~~~~~~~~~~~~~~~~~
+
+End hosts send traffic to the SCION router's Internal Interface address and port. This underlay port must be configured on the endpoints via the ``topology.json`` file.
+
+STUN Support and NAT Address Discovery
```
tzaeschke: Actually, this should not be called "STUN". THe mechanism is based on STUN, but it's not entirely compliant with it.

Maybe just call this section "NAT Address Discovery" and avoid the term STUN?

tzaeschke:
```suggestion
NAT Address Discovery
```

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,103 @@
...
+--------------------------------------
```
tzaeschke:
```suggestion
""""""""""""""""""""""""""""""""""""""
```
I assume this should be a sub-heading of the previous section?

doc/protocols/underlay.rst
tzaeschke: I proposed an update, what do you think?

Also, I could not find this behavior described here: https://docs.scion.org/en/latest/protocols/scmp.html, so we probably need to update the doc.

doc/protocols/underlay.rst
tzaeschke: The heading-level of NAT Address discovery has been downgraded. This is now a subsection of "Ports Overview".

doc/protocols/underlay.rst
tzaeschke: Fixed, see above.

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,103 @@
...
+
+* **Dispatched Ports**: If the port falls within the configured dispatched ports range, the router forwards the packet to the end host using that same port as the UDP underlay destination port.
+* **UDP underlay default port**: If the port falls outside the configured range, or if the system is handling legacy fallback traffic (i.e., traffic from older clients that do not support dynamic port mapping), the router forwards the packet to the default end-host data port: **30041**.
+* **SCMP**: For SCMP informational messages, the router forwards the packet to the default end-host port: **30041**. For SCMP error messages, the router tries to extract the end-host port and sends the packet to the endhost, subject to dispatched port settings and the default underlay port. If the port cannot be extracted, the packet is dropped.
```
tzaeschke:
```suggestion
* **SCMP**: 

  * For SCMP informational request messages, the router forwards the packet to the default end-host port: **30041**. 
  * For SCMP informational response messages, the router extracts the destination port from the identifier field of the SCMP message, subject to dispatched port settings and the default underlay port. 
  * For SCMP error messages, the router tries to extract the end-host port and sends the packet to the endhost, subject to dispatched port settings and the default underlay port. If the port cannot be extracted, the packet is dropped.
```

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,103 @@
...
+
+SCION border routers utilize specific underlay ports to process and forward traffic:
+
+* **Internal Interfaces**: Used for intra-AS communication to receive traffic from end-hosts. Operators can choose the port freely. The same port must be configured on endpoints so that they can send outbound traffic. Routers with multiple internal interfaces can use a range of ports.
+* **External Interfaces**: Used for inter-AS links towards neighboring SCION ASes. Note that the choice of underlay protocol and UDP port is per link. It is independent from other links and intra-AS underlay. 
+
+Control Plane Instances
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Control plane components communicate  through RPC messages that are transported via Connect RPC. This protocol carries messages over HTTP/3, that uses a QUIC transport layer.  Identification of the relevant addresses and ports for inter-domain queries is provided by  `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_.
```
tzaeschke:
```suggestion
Control plane components of different ASes communicate with each other through RPC messages that are transported via Connect RPC. This protocol carries messages over HTTP/3, that uses a QUIC transport layer.  Identification of the relevant addresses and ports for inter-domain queries is provided by  `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_. 
```

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,103 @@
...
+* **External Interfaces**: Used for inter-AS links towards neighboring SCION ASes. Note that the choice of underlay protocol and UDP port is per link. It is independent from other links and intra-AS underlay. 
+
+Control Plane Instances
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Control plane components communicate  through RPC messages that are transported via Connect RPC. This protocol carries messages over HTTP/3, that uses a QUIC transport layer.  Identification of the relevant addresses and ports for inter-domain queries is provided by  `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_. 
+
+For intra-domain control service instances, the operator may use arbitrary ports, that have to be communicated to endpoints. For a comprehensive list of ports used by this implementation, refer to the `Control Port Table <../manuals/control.html#port-table>`_.
```
tzaeschke:
```suggestion
For intra-domain communication of endpoints with service instances, the operator may use arbitrary ports, that have to be communicated to endpoints. For a comprehensive list of ports used by this implementation, refer to the `Control Port Table <../manuals/control.html#port-table>`_.
```

doc/protocols/underlay.rst
tzaeschke: Not sure I understand, where did we remove intra-domain communication.

I think this paragraph simply says that endpoints may use arbitrary ports to communicate with control services.

doc/protocols/underlay.rst
tzaeschke: I removed "control" because in future we may communicate with the end-host-service (which may or may not be a control service).

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,103 @@
...
+
+
+
+Protocol Stack Summary
```
tzaeschke: Actually, wouldn't it make sense to keep the "stack" file and have "underlay" as a separate doc?
It is weird having the full stack in the underlay...

Also, this is why the build currently fails, data-plane.rst still references stack, and underlay is not referenced anywhere.

I can do that we people agree.

doc/protocols/underlay.rst
nicorusti: I don't see the update. 

>  for replies the ID field is inspected to dispatched the port directly to the underlay client port 

~What ID? And~ how do rotuers map this ID to a port?

doc/protocols/underlay.rst
nicorusti: > Why is this in a subsection of STUN?

Because STUN is how the end host gets the right source IP. But I believe I got the wording wrong, it should be: 

> The end host can then reliably inject this public, border-router-visible IP and port into the ~SCION~ underlay source address

doc/protocols/underlay.rst
nicorusti: > why are we having this paragraph? 

So that a reared landing on this page (e.g., from the IETF reference), wondering what ports are used for CP, gets sent in the right direction. In my mind, the goal of this page is to explain: How does SCION run on top of IP?  And this as comprehensively as possible.

doc/protocols/underlay.rst
nicorusti: Right, we could rename it "NAT traversal"
We say that it is based on STUN. Are you following any RFC? (e.g., rfc8489)?

doc/protocols/underlay.rst
JordiSubira: I meant why the "Routers" section was a subsection of STUN, but Til already fixed it

doc/protocols/underlay.rst
nicorusti: @JordiSubira how is the mapping from ID to underlay address done?

doc/protocols/underlay.rst
JordiSubira: Well this is TCP/IP (without SCION on top, so technically not an underlay used by the SCION protocol here). I do not see how this feet in a document called UDP/IP underlay.

doc/protocols/underlay.rst
nicorusti: Ok sounds good. That stack image somehow overlaps with the underlay topic, this is why I thought we could keep it there. But we can also have two separate pages, and then in Stack reference IP/UDP underlay. 

It would be great if you could do it!

doc/protocols/underlay.rst
JordiSubira: Identifier field in SCMP information messages. It's a direct mapping.

doc/protocols/underlay.rst
JordiSubira: Mostly good, but I would add:
```
For SCMP informational response messages, the router extracts the destination port from the identifier field of the SCMP message, subject to dispatched port settings and the default underlay port
```

doc/protocols/underlay.rst
nicorusti: Ok, I got your point. We can make it clear in the text that control plane traffic is not directly using the underlay. I propose a change and tag you there

nicorusti: We could also change it to IP underlay.. because intra-AS traffic goes directly over IP, but this is not using the IP/UDP underlay

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,95 @@
...
+* **External Interfaces**: Used for inter-AS links towards neighboring SCION ASes. Note that the choice of underlay protocol and UDP port is per link. It is independent from other links and intra-AS underlay.
+
+Control Plane Instances
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Control plane components of different ASes communicate with each other through RPC messages that are transported via Connect RPC. This protocol carries messages over HTTP/3, that uses a QUIC transport layer.  Identification of the relevant addresses and ports for inter-domain queries is provided by  `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_.
+
+For intra-domain communication of endpoints with service instances, the operator may use arbitrary ports, that have to be communicated to endpoints. For a comprehensive list of ports used by this implementation, refer to the `Control Port Table <../manuals/control.html#port-table>`_.
```
nicorusti:
```suggestion
Control plane components communicate with each other and with end-hosts through RPC messages transported via Connect RPC. This protocol carries messages over HTTP/3, that uses a QUIC transport layer. They therefore do not communicate directly on top of the UDP/IP underlay. 

 For inter-AS communication, traffic is transported on top of UDP/SCION between control service instances.  Identification of the relevant control service addresses and ports is provided by  `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_.

For intra-AS communication, traffic is transported on top of IP between end-hosts and control service instances. The port used by control service instances is configured by operators, and configured on end-hosts. For a comprehensive list of ports used by this implementation, refer to the `Control Port Table <../manuals/control.html#port-table>`_.
```

How about this @JordiSubira ?

doc/protocols/underlay.rst
nicorusti: Does the identifier include a port directly? Or is the router supposed to stare state containing mappings of identifiers to source endpoint underlay ports?

tzaeschke: No state, the field just contains the port number.

tzaeschke: See #4901.

doc/protocols/underlay.rst
romshark: It seems we have a contradiction:
> This protocol carries messages over HTTP/3, that uses a QUIC transport layer

But port-table says:
> QUIC + gRPC with HTTP/2 

https://github.com/nicorusti/scion/blob/9fdeabdc157194ece544c56d5aa8d556d60e151b/doc/manuals/control/port-table.rst?plain=1#L6

The port-table should probably say "Connect RPC over HTTP/3"

JordiSubira: My point is that inter-AS traffic is SCION/UDP/IP and intra-AS traffic is TCP/IP (no SCION), if you still believe you have to mention it here, I would make clear this (gRPC/HTTP/TCP/IP or ConnectRPC/HTTP/TCP/IP)  (no SCION) whatever the client implements or will implement in the future.

tzaeschke: I think now I understand the problem.
If we discuss non-SCION traffic (e.g. endpoint<->control server), then we should add that to the introduction (the introduction currently says we are discussing SCION/UDP/IP traffic). 
I guess my summary:
- inter-domain traffic: SCION/UDP/IP (dataplane + control plane)
- intra-domain: 
   - SCION/UDP/IP (dataplane) (for simplicity let's assume intra-AS dataplane uses SCION)
   - TCP/IP (control plane) 

Right?

JordiSubira: This was my point, indeed.

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,95 @@
+***************
+IP/UDP underlay
+***************
+
+.. _underlay:
+
+Introduction
+------------
+
```
tzaeschke, line 44:
```suggestion
SCION allows reusing existing network fabrics to provide connectivity among SCION infrastructure services, routers, and end-hosts. To maximize compatibility with current network infrastructures, most implementations encapsulate the SCION header inside a standard UDP/IPv6 or UDP/IPv4 packet. For dataplane traffic and inter-domain control-plane traffic, SCION packets are enclosed within the standard IP/UDP protocol stack:

.. code-block:: text

    +-----------------------------+
    |                             |
    |        Payload (L4)         |
    |                             |
    +-----------------------------+
    |           SCION             |
    +-----------------------------+ <-+
    |            UDP              |   |
    |                             |   | Existing
    +-----------------------------+   | network
    |            IP               |   | fabric
    +-----------------------------+   |
    |        Link Layer           |   |
    +-----------------------------+ <-+

The exception is intra-domain control-plane traffic between end-hosts and the control-plane. This usually uses gRPC or ConnectRPC over HTTP/TCP/IP without SCION:

    +-----------------------------+
    |                             |
    |        Payload (L4)         |
    |                             |
    +-----------------------------+
    |            TCP              |
    |                             |
    +-----------------------------+
    |            IP               |
    +-----------------------------+
    |        Link Layer           |
    +-----------------------------+

```

doc/protocols/underlay.rst
tzaeschke: I updated the intro, does that help?

doc/protocols/underlay.rst
nicorusti: Super, thanks a lot for proposing it!

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,128 @@
...
+    +-----------------------------+
+    |           SCION             |
+    +-----------------------------+ <-+
+    |            UDP              |   |
+    |                             |   | Intra-domain
+    +-----------------------------+   | protocol
+    |            IP               |   |
+    +-----------------------------+   |
+    |        Link Layer           |   |
+    +-----------------------------+ <-+
```
tzaeschke:
```suggestion
```

tzaeschke: This can be removed it is a duplication of the above.

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,110 @@
...
+    |            UDP              |   |
+    |                             |   | Existing
+    +-----------------------------+   | network
+    |            IP               |   | fabric
+    +-----------------------------+   |
+    |        Link Layer           |   |
+    +-----------------------------+ <-+
+
+The exception is intra-domain control-plane traffic between end-hosts and the control-plane. This usually uses gRPC or ConnectRPC over HTTP/TCP/IP without SCION:
+
```
tzaeschke, line 29:
```suggestion

.. code-block:: text

```

doc/protocols/underlay.rst
tzaeschke: How about this:
```suggestion
Control plane components communicate with each other and with end-hosts through RPC messages transported via Connect RPC.

For inter-AS communication, traffic is transported over ConnectRPC/HTTP3/QUIC/UDP/SCION/UDP/IP between control service instances. Identification of the relevant control service addresses and ports is provided by `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_.
 
For intra-AS communication, traffic is transported over ConnectRPC/HTTP/TCP/IP between end-hosts and control service instances. The port used by control service instances is configured by operators, and configured on end-hosts. For a comprehensive list of ports used by this implementation, refer to the `Control Port Table <../manuals/control.html#port-table>`_.
```

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,112 @@
...
+* **External Interfaces**: Used for inter-AS links towards neighboring SCION ASes. Note that the choice of underlay protocol and UDP port is per link. It is independent from other links and intra-AS underlay.
+
+Control Plane Instances
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Control plane components of different ASes communicate with each other through RPC messages that are transported via Connect RPC. This protocol carries messages over HTTP/3, that uses a QUIC transport layer. Identification of the relevant addresses and ports for inter-domain queries is provided by `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_.
+
+For intra-domain communication of endpoints with service instances, the operator may use arbitrary ports, that have to be communicated to endpoints. For a comprehensive list of ports used by this implementation, refer to the :ref:`control-port-table`.
```
tzaeschke:
```suggestion
Control plane components communicate with each other and with end-hosts through RPC messages transported via Connect RPC.

For inter-AS communication, traffic is transported over ConnectRPC/HTTP3/QUIC/UDP/SCION/UDP/IP between control service instances. Identification of the relevant control service addresses and ports is provided by `Service discovery <https://datatracker.ietf.org/doc/html/draft-dekater-scion-controlplane-15#name-control-service-discovery>`_.

For intra-AS communication, traffic is transported over ConnectRPC/HTTP/TCP/IP between end-hosts and control service instances. The port used by control service instances is configured by operators, and configured on end-hosts. For a comprehensive list of ports used by this implementation, refer to the :ref:`control-port-table`.
```
How about this? 
This is an alternative to Nicola's proposal (which got removed during one of the recent changes).

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,112 @@
...
+
+The SCMP Daemon (``scmpd``) SHOULD listen on the UDP underlay default port (30041) to process and reply to informational SCMP messages, such as echo requests (pings) and traceroutes.
+
+
+Traffic from End-hosts
+~~~~~~~~~~~~~~~~~~~~~~
+
+End hosts send traffic to the SCION router's Internal Interface address and port. This underlay port must be configured on the endpoints via the ``topology.json`` file.
+
+NAT Address Discovery
```
nicorusti:
```suggestion
NAT Traversal
```

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,112 @@
...
+""""""""""""""""""""""""""""""""""""""
+
+End hosts located behind a Network Address Translation (NAT) device face a unique challenge: the source address they encode must be the external IP and port visible to the first-hop border router, rather than their internal local address.
+
+To resolve this, SCION incorporates a :doc:`NAT IP/port discovery mechanism <../dev/design/NAT-address-discovery>` conceptually similar to the STUN (Session Traversal Utilities for NAT) protocol, operating directly between clients and border routers. The border router acts as a detector; when the client sends a discovery request, the border router observes the NAT-mapped IP and port and reports it back to the client.
```
nicorusti:
```suggestion
To resolve this, this SCION implementation incorporates a :doc:`NAT IP/port discovery mechanism <../dev/design/NAT-address-discovery>` conceptually similar to the STUN (Session Traversal Utilities for NAT) protocol, operating directly between clients and border routers. The border router acts as a detector; when the client sends a discovery request, the border router observes the NAT-mapped IP and port and reports it back to the client.
```

doc/protocols/underlay.rst
```diff
@@ -0,0 +1,112 @@
...
+
+The end host can then reliably inject this public, border-router-visible IP and port into the SCION source address fields of its outbound packets. This guarantees that return traffic from the remote destination can be successfully routed back through the NAT to the client.
```
nicorusti, line 97:
```suggestion
The end host can then reliably inject this public, border-router-visible IP and port into the SCION underlay source address field of its outbound packets. This guarantees that return traffic from the remote destination can be successfully routed back through the NAT to the client.
```
I believe the NAT-traversal address is used in the underlay this source address, not inside the SCION address header, right?

doc/protocols/underlay.rst
nicorusti: sounds good for me!

nicorusti:
Thank you all for contributing! I think it looks great and it can be merged.
@tzaeschke I left some minor edits on NAT traversal, if you agree could you add them? 
If not, could you merge anyways and then we polish it in another PR? 
I need the link up on Monday so that I can also upload the last DP draft version :)

doc/protocols/underlay.rst
tzaeschke: The endhost injects it into the SCION header src field. This is required so that for return packets, the BR can use the address in the underlay to reach the NAT router.
The endhost cannot inject it into the underlay, the NAT router does that.

nicorusti: ok thanks for clarifying

romshark (approved):
LGTM

PR #4898, opened 2026-03-06
ping: allow pinging svc addresses

uniquefine:
Extend the `scion ping` tool to allow pinging SCION control plane services by svc address.

This is useful if you want to ping some host in a remote ISD-AS but don't know a specific IP.

### DIFF
--- a/pkg/snet/addrutil/addrutil.go
+++ b/pkg/snet/addrutil/addrutil.go
@@ -188,12 +188,9 @@ func ExtractDestinationServiceAddress(a addr.SVC, path snet.Path) net.Addr {
 	}
 }
 
-// ExtractDstSvcUdpAddr extracts the destination service UDP address
-// from the provided path. If the path contains discovery information, it will
-// use a random available control or discovery service address based on the
-// provided service type (addr.SvcCS or addr.SvcDS). If no discovery information
-// is available, it will return a SVC address with the destination IA and the
-// path's underlay next hop.
+// ExtractDstSvcUdpAddr attempts to extract service UDP address
+// from the provided path metadata and reports whether the extraction was successful.
+// If successful one of the available addresses is selected at random.
 // The caller must ensure that the path is not nil.
 func ExtractDstSvcUdpAddr(a addr.SVC, path snet.Path) (net.UDPAddr, bool) {
 	if path == nil {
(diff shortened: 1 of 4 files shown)


scion/cmd/scion/ping.go
```diff
@@ -99,6 +100,10 @@ func newPing(pather CommandPather) *cobra.Command {
 		Long: fmt.Sprintf(`'ping' test connectivity to a remote SCION host using SCMP echo packets.
 Use 'scion address' on the remote SCION host to determine the ISD-AS and pingable IP address.
 
+You can also ping SCION control plane services in the remote AS by using the SVC address of the
+service, e.g., '1-ff00:0:110,CS' to ping the control service.
```
tzaeschke:
```suggestion
service, e.g., '1-ff00:0:110,CS' to ping a control service.
```
I guess it randomly picks one?

uniquefine: Yes, it picks the first one.

scion/cmd/scion/ping.go
```diff
@@ -205,15 +210,54 @@ On other errors, ping will exit with code 2.
...
+					}
+					target = remote.Host.IP().AsSlice()
 				}
 				if localIP, err = addrutil.ResolveLocal(target); err != nil {
 					return serrors.Wrap("resolving local address", err)
 				}
 				printf("Resolved local address:\n  %s\n", localIP)
 			}
+
+			// Resolve remote Addr if it's an SVC address
```
lukedirtwalker, line 232: you can also use addrutil.ExtractDestinationServiceAddress that might already resolve the SVC address without extra roundtrip

JordiSubira: I was just writing the same message :)

uniquefine: Nice, I didn't know about that function.
I refactored it a bit to fit both usecases.

uniquefine: Done

scion/cmd/scion/ping.go
uniquefine: Done

pkg/snet/addrutil/addrutil.go
```diff
@@ -171,43 +171,47 @@ func ExtractDestinationServiceAddress(a addr.SVC, path snet.Path) net.Addr {
...
+	}
+}
 
+// ExtractDstSvcUdpAddr extracts the destination service UDP address
+// from the provided path. If the path contains discovery information, it will
+// use a random available control or discovery service address based on the
+// provided service type (addr.SvcCS or addr.SvcDS). If no discovery information
+// is available, it will return a SVC address with the destination IA and the
+// path's underlay next hop.
+// The caller must ensure that the path is not nil.
```
lukedirtwalker, line 194: This documentation string doesn't match the behavior...

lukedirtwalker: it seems to be a plain copy of the function doc above.

PR #4899, opened 2026-03-09
scion-pki: support additional extended key usage

lukedirtwalker:
Add the ability to specify additional extended key usage (EKU) in the certificate creation process. This applies to `scion-pki certificate create` and `scion-pki certificate sign` commands.

When setting the `--eku-any` flag is set, the ExtKeyUsageAny is set on the certificate or CSR. Furthermore, the `--eku` flag allows the user to provide a comma-separated list of ASN.1 OIDs for additional EKUs that are unknown to scion-pki.


oncilla (approved):

oncilla (approved):

private/options/options.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2025 Anapaya Systems
```
romshark: 2026, right?

scion-pki/certs/create.go
```diff
@@ -655,3 +777,17 @@ func keyUsage() pkix.Extension {
 		Value:    val,
 	}
 }
+
+func parseOID(s string) (asn1.ObjectIdentifier, error) {
```
romshark, line 783: I'm not an expert on this topic, but it looks like `parseOID` is too permissive according to [RFC 4512 - 1.4](https://datatracker.ietf.org/doc/html/rfc4512#section-1.4)

Seems to violate:
- minimum two arcs
- no leading zeros 
- no negative arcs 

```
func TestParseOIDRejectsInvalidRFC4512Numericoids(t *testing.T) {
	// https://datatracker.ietf.org/doc/html/rfc4512#section-1.4
	testCases := map[string]string{
		// "numericoid = number 1*( DOT number )"
		// A numericoid must contain at least two arcs.
		"single arc": "1",

		// "number = DIGIT / ( LDIGIT 1*DIGIT )"
		// A numericoid arc is a decimal number, so '-' is not allowed.
		"negative arc": "1.-2",

		// "number = DIGIT / ( LDIGIT 1*DIGIT )"
		// Multi-digit arcs must start with LDIGIT (1-9), not 0.
		"leading zero in first arc": "01.2",

		// "number = DIGIT / ( LDIGIT 1*DIGIT )"
		// The same no-leading-zero rule applies to every arc.
		"leading zero in later arc": "1.02",
	}
	for name, input := range testCases {
		t.Run(name, func(t *testing.T) {
			_, err := parseOID(input)
			require.Error(t, err)
		})
	}
}
```

Is strict compliance not necessary in this particular context?
Please correct me if I'm wrong, but as far as I undertood the consequence would be a nonsensical OID baked in to the cert, which isn't really a security problem but a minor inconvenience?

nicorusti:
@lukedirtwalker is this in any way affecting the PKI draft (that we can still change)
JFYI we were discussing EKU recently in https://github.com/scionassociation/scion-cppki_I-D/issues/79

private/options/options.go
lukedirtwalker: Done.

scion-pki/certs/create.go
lukedirtwalker: Adapted.

lukedirtwalker:
>@lukedirtwalker is this in any way affecting the PKI draft (that we can still change)
JFYI we were discussing EKU recently in https://github.com/scionassociation/scion-cppki_I-D/issues/79

@nicorusti I think not. This is only about tooling to make it possible to set additional EKUs on certs.

romshark (approved):
LGTM

PR #4902, opened 2026-03-10
control: fix beacon grouping

lukedirtwalker:
Instead of grouping the beacons when loading them from the database we only group them before we send them off to the registration plugins. This makes sure that the registration policy can use the "full" path including the ingress interface added in the extension.

### DIFF
--- a/control/beaconing/writer.go
+++ b/control/beaconing/writer.go
@@ -46,9 +46,9 @@ type Pather interface {
 // SegmentProvider provides segments to register for the specified type.
 type SegmentProvider interface {
 	// SegmentsToRegister returns the segments that should be registered for the
-	// given segment type as GroupedBeacons.
-	// The returned GroupedBeacons must not be nil if the returned error is nil.
-	SegmentsToRegister(ctx context.Context, segType seg.Type) (beacon.GroupedBeacons, error)
+	// given segment type.
+	// The returned slice must not be nil if the returned error is nil.
+	SegmentsToRegister(ctx context.Context, segType seg.Type) ([]beacon.Beacon, []beacon.RegistrationPolicy, error)
 }
 
 // SegmentStore stores segments in the path database.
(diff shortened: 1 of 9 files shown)


oncilla (approved):

oncilla (approved):

control/beaconing/writer.go
```diff
@@ -556,3 +564,30 @@ func (w *GroupWriter) Write(
 	}
 	return writeStats, nil
 }
+
+// groupBeacons groups beacons according to the registration policies defined in the policy.
+// The beacons that do not match any registration policy are dropped.
+// If the policy defines no registration policy, all the beacons will be put into
+// the default group.
+func (w *GroupWriter) groupBeacons(
```
romshark: nit: Does this need to be a method? can't it just be an unexported standalone func?

```go
func groupBeacons(
	beacons []beacon.Beacon,
	policies []beacon.RegistrationPolicy,
) beacon.GroupedBeacons {
```

I don't see `w` being used anywhere.

control/beaconing/writer.go
```diff
@@ -556,3 +564,30 @@ func (w *GroupWriter) Write(
...
+	beacons []beacon.Beacon,
+	policies []beacon.RegistrationPolicy,
+) beacon.GroupedBeacons {
+	if len(policies) == 0 {
+		return map[string][]beacon.Beacon{
```
romshark:
```suggestion
		return beacon.GroupedBeacons{
```

control/beaconing/writer.go
```diff
@@ -449,62 +457,60 @@ type GroupWriter struct {
...
+	for i, b := range beacons {
+		// If the beacon does not have a valid interface ID, skip it.
+		if w.Intfs != nil && w.Intfs.Get(b.InIfID) == nil {
+			continue
+		}
+		// Try to terminate the segment if an extender is configured.
+		if w.Extender != nil {
+			err := w.Extender.Extend(ctx, b.Segment, b.InIfID, 0, peers)
+			var signerGenError trust.SignerGenError
+			if errors.As(err, &signerGenError) {
```
romshark:
```suggestion
			if signerGenError, ok := errors.AsType[trust.SignerGenError](err); ok {
```

romshark (approved):
LGTM besides the few nitpicks

control/beaconing/writer.go
lukedirtwalker: Done.

control/beaconing/writer.go
lukedirtwalker: Done.

control/beaconing/writer.go
lukedirtwalker: Nice. Haven't used this before :)

romshark (approved):
LGTM

PR #4908, opened 2026-03-19
docs: improve SIG docs

katyatitkova:
Fixes https://github.com/scionproto/scion/issues/4787

### DIFF
--- a/doc/manuals/gateway.rst
+++ b/doc/manuals/gateway.rst
@@ -73,7 +73,7 @@ Each SIG requires two configuration files, with an optional third:
 
 The SIG in AS ``1-ff00:0:111`` needs a matching traffic policy pointing back to AS ``1-ff00:0:110`` and listing the prefixes reachable through it.
 
-Once both SIGs are running, IP traffic matching the configured prefixes is automatically tunneled over SCION. Each SIG independently discovers remote SIGs via :ref:`SIG Discovery <sig-discovery>`, fetches their prefix announcements via :ref:`SGRP <sgrp>`, and creates tunneling sessions.
+Once both SIGs are running, IP traffic matching the configured prefixes is automatically tunneled over SCION. Each SIG independently discovers remote SIGs via :ref:`SIG Discovery <sig-discovery>`, fetches their prefix announcements via :ref:`SGRP <sgrp>`, and creates tunneling sessions. Note that path policies for SIGs are not configurable now.
 
 IP traffic must be routed to the SIG using standard IP routing. The SIG automatically installs routes on the machine it runs on, but other hosts in the local network require static routes pointing to the SIG. Dynamic routing mechanisms are not supported.
 


doc/index.rst
```diff
@@ -36,13 +36,14 @@ The ideas and concepts behind SCION.
...
    Cryptography <cryptography/index>
-   sig
+   SIG Protocol Specification <sig>
    glossary
 
 * **Overview**:
   :doc:`SCION <overview>` |
   :doc:`Control Plane <control-plane>` |
   :doc:`Data Plane <data-plane>` |
+  :doc:`SIG <sig>` |
```
nicorusti: I'd in general expand SIG (unless you are in a SIG-specific page)
```suggestion
  :doc:`SCION-IP Gateway <sig>` |
```

doc/manuals/gateway.rst
```diff
@@ -1,6 +1,147 @@
...
+   Run :option:`gateway help completion [shell] <gateway help>` for usage information on the
+   autocomplete script for a particular shell.
+
+Deployment Examples
+===================
+
+Basic SIG Pair
+--------------
+
+The simplest deployment uses one SIG per AS to tunnel IP traffic between two SCION ASes. In this example, AS ``1-ff00:0:110`` and AS ``1-ff00:0:111`` each run a SIG.
```
nicorusti: With this wording, it feels like the concept of SIG is somehow bound to the concept of AS, while it is not directly.. How about something like:
```suggestion
The simplest deployment uses a pair of SIGs to tunnel IP traffic between them. In this example, the two SIGs are respectively in AS ``1-ff00:0:110`` and AS ``1-ff00:0:111``.
```

doc/sig.rst
```diff
@@ -1,6 +1,6 @@
-****************
-SCION-IP Gateway
-****************
+**************************
+SIG Protocol Specification
```
nicorusti: How about 
```suggestion
IP in SCION Tunneling
```

This way we distinguish between the manual and this page

doc/sig.rst
```diff
@@ -28,28 +28,31 @@ This protocol is designed to:
...
 -------------
 
-https://github.com/scionproto/scion/blob/master/proto/gateway/v1/prefix.proto
+The prefix exchange protocol is defined in `proto/gateway/v1/prefix.proto <https://github.com/scionproto/scion/blob/master/proto/gateway/v1/prefix.proto>`_.
+
+Server - `gateway/control/grpc/prefix_server.go <https://github.com/scionproto/scion/blob/master/gateway/control/grpc/prefix_server.go>`_
 
-Server - https://github.com/scionproto/scion/blob/master/gateway/control/grpc/prefix_server.go
-Client - https://github.com/scionproto/scion/blob/master/gateway/control/grpc/prefix_fetcher.go
+Client - `gateway/control/grpc/prefix_fetcher.go <https://github.com/scionproto/scion/blob/master/gateway/control/grpc/prefix_fetcher.go>`_
```
nicorusti: Could you derive a textual specification from the code?

doc/manuals/gateway.rst
nicorusti: To be more precise, I believe this is using some sort of service discovery to discover remote SIGs. This should be clarified, and and distinguished from the case when SIGs are preconfigured (including SIG addresses).

doc/manuals/gateway.rst
```diff
@@ -1,6 +1,147 @@
...
+       },
+       "ConfigVersion": 1
+     }
+
+3. **Routing policy** (optional): controls which IP prefixes to accept and advertise via :ref:`SGRP <sgrp>`. See `Routing Policy File`_ for the full syntax.
+
+The SIG in AS ``1-ff00:0:111`` needs a matching traffic policy pointing back to AS ``1-ff00:0:110`` and listing the prefixes reachable through it.
+
+Once both SIGs are running, IP traffic matching the configured prefixes is automatically tunneled over SCION. Each SIG independently discovers remote SIGs via :ref:`SGRP <sgrp>`, fetches their prefix announcements, and creates tunneling sessions.
+
```
nicorusti, line 77: Should we also mention something about path policies? I'm not sure if they can be configured at all, perhaps a allow/denylist of ASes?

doc/manuals/gateway.rst
```diff
@@ -1,6 +1,147 @@
...
+Multi-SIG Deployment
```
nicorusti, line 80: I think you can also have multi-SIG with SIG discovery. I'd distinguish the two cases on wether discovery is used or not

doc/manuals/gateway.rst
```diff
@@ -1,6 +1,147 @@
...
+
+   # The bind address for encapsulated data traffic.
+   # (default ":30056")
+   data_addr = ":30056"
+
+   # The bind address for path probes.
+   # (default ":30856")
+   probe_addr = ":30856"
+
+- ``id``: Identifier for this gateway instance. Relevant when running multiple SIGs.
```
nicorusti, line 126: How is the ID configured?

doc/sig.rst
```diff
@@ -28,28 +28,31 @@ This protocol is designed to:
...
 
 
+.. _sgrp:
+
 SCION Gateway Routing Protocol (SGRP)
 =====================================
 
 The SCION Gateway Routing Protocol (SGRP) enables SIGs to map IP prefixes to SCION ASes.
 
 A SIG participating in SGRP between two SCION ASes does the following:
```
nicorusti: I think this is what it does:
```suggestion
The SCION Gateway Routing Protocol (SGRP) enables SIGs to mutually exchange IP prefixes through  RPC queries.

A SIG participating in SGRP does the following:
```

doc/sig.rst
```diff
@@ -28,28 +28,31 @@ This protocol is designed to:
...
 
-1. It discovers the SIGs in the remote SCION AS by periodically sending a discovery message to the Control Plane of the remote AS which replies with a list of local tunneling endpoints.
+1. It discovers the SIGs in the remote SCION AS by periodically sending a ``DiscoveryService.Gateways`` gRPC request to the remote AS. The remote AS replies with a list of gateways, each described by a control address, a data address, a probe address, and an optional set of allowed AS interfaces. The discovery service is defined in `proto/discovery/v1/discovery.proto <https://github.com/scionproto/scion/blob/master/proto/discovery/v1/discovery.proto>`_.
```
nicorusti: This is not SGRP. This should be a separate section, something like "Remote SIG discovery"

nicorusti: > a ``DiscoveryService.Gateways`` gRPC request 

Is the SIG still using gRPC? Or did it get also upgraded to Connect RPC?

nicorusti: And it should link to the discovery mechanisms.
Is it perhaps the same as https://www.ietf.org/archive/id/draft-dekater-scion-controlplane-16.html#name-control-service-discovery ?

nicorusti: And is this affected by this: 
https://github.com/scionproto/scion/pull/4789 ?

nicorusti: The section should also describe how a SIG makes itself discoverable to other ASes

doc/manuals/gateway.rst
```diff
@@ -1,6 +1,147 @@
...
+.. option:: completion [shell]
+
+   Generate the autocompletion script for :program:`gateway` for the specified shell.
+
+   Run :option:`gateway help completion <gateway help>` for a list of the available shells.
+
+   Run :option:`gateway help completion [shell] <gateway help>` for usage information on the
+   autocomplete script for a particular shell.
+
+Deployment Examples
```
nicorusti, line 49: We should clarify somewhere in this section how packets make it to the SIG: AFAIK currently it only supports static routes. 
Dynamic routing mechanisms are not supported (well, they are in Anapaya's stack)

doc/sig.rst
nicorusti, line 130: In this section, we should (optionally) add some considerations on tunneling aspects of RFC8085. E.g.: 
- Do SISGs tunnel multicast traffic
- State that unicast IP traffic is assumed to be congestion controlled

See: 
https://github.com/scionassociation/scion-cp_I-D/issues/228

doc/manuals/gateway.rst
katyatitkova: Can we preconfigure SIG addresses? I don't see where in the code it can be preconfigured, it looks to me that it's always dynamic discovery

doc/manuals/gateway.rst
katyatitkova: Path policy is not configurable now. Does it make sense to mention it? Usually the manual has the info only about what is supported, unless something standard is not supported

doc/manuals/gateway.rst
katyatitkova: It is mentioned above, in the sample config:
```
# ID of the gateway (default "gateway")
id = "gateway"
```

doc/sig.rst
katyatitkova: Fixed in [a2fdfae](https://github.com/scionproto/scion/pull/4908/commits/a2fdfae43d3eb583676ee8c0cd34de56436835eb)

doc/sig.rst
katyatitkova: Fixed in [4e2e1df](https://github.com/scionproto/scion/pull/4908/commits/4e2e1df2bb04fa6571ee73bc2b26e73cfaf81da7)

doc/sig.rst
katyatitkova: #4789 didn't affect the doc here. Other feedback is now fixed in [14302ac](https://github.com/scionproto/scion/pull/4908/commits/14302ac9ce31ecf82e59787dd4e72d8992137aa7) and [415e360](https://github.com/scionproto/scion/pull/4908/commits/415e360612e084e0a2486e7ca00e54847fb359a9)

doc/sig.rst
katyatitkova: Fixed in [d711491](https://github.com/scionproto/scion/pull/4908/commits/d711491d5943e55b8268f5ef7bde441d88e09a9c)

doc/manuals/gateway.rst
katyatitkova: Fixed in [6618ba9](https://github.com/scionproto/scion/pull/4908/commits/6618ba9141c124f50073e7eda1e97061b6896059)

doc/sig.rst
katyatitkova: Fixed in [6d5ddad](https://github.com/scionproto/scion/pull/4908/commits/6d5ddadf5a72fb365ba8cbb64cfd44514f0af957)

doc/manuals/gateway.rst
katyatitkova: Same as in https://github.com/scionproto/scion/pull/4908#discussion_r2994424135 - can we actually bypass discovery?

doc/manuals/gateway.rst
nicorusti: Hmm, in my own experience, I remember I was looking for them for a long time while I was deploying some SIGs in the past.. I still think it would be valuable to mention it somewhere

doc/sig.rst
nicorusti: Super, thank you!

doc/manuals/gateway.rst
nicorusti: I'm not too familiar with this, I think you may be right that the current implementation only relies on service discovery. Also here it still leverage discovery: https://docs.scion.org/en/latest/manuals/gateway.html#network-prefix-pinning

katyatitkova:
@nicorusti I added info on path policies. However, now the reapproval is needed :)

PR #4922, opened 2026-04-15
docs: update ISD and AS Numbering section

katyatitkova:


doc/overview.rst
```diff
@@ -52,14 +52,92 @@ ISD and AS Numbering
...
-The 48-bit AS numbers are globally unique, and use a superset of the existing BGP AS numbering
-scheme.
-Formatting rules and allocations are currently described in `wiki page "ISD and AS numbering" <https://github.com/scionproto/scion/wiki/ISD-and-AS-numbering>`_.
-
-The endpoint local address is not used for inter-domain routing or forwarding, does not need to be
-globally unique, and can thus be an IPv4, IPv6, or MAC address, for example.
-A SCION endpoint address is the ``ISD-AS,local address`` 3-tuple.
+An ISD-AS number is 64 bits, with the top 16 bits indicating the ISD, and the bottom 48 bits indicating the AS. The text representation uses a ``-`` separator between the ISD and AS numbers. For example, ``4-ff00:1:f``.
+
+ISD and AS numbering is described in Section 1.5 "Addressing" of the `Control Plane Internet Draft <https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.html>`_. SCION ISD and AS numbers are assigned and registered by the `SCION Registry <https://www.scion.org/registry>`_, operated by the `SCION Association <https://www.scion.org>`_.
```
nicorusti, line 57:
```suggestion
ISD and AS numbering is described in Section 1.5 "Addressing" of the `Control Plane Internet Draft <https://www.ietf.org/archive/id/draft-dekater-scion-controlplane-18.html#name-addressing>`_. SCION ISD and AS numbers are assigned and registered by the `SCION Registry <https://www.scion.org/registry>`_, operated by the `SCION Association <https://www.scion.org>`_.
```

doc/overview.rst
```diff
@@ -52,14 +52,92 @@ ISD and AS Numbering
...
+     - Reserved for documentation and sample code (analogous to
+       `RFC 5398 <http://www.iana.org/go/rfc5398>`_).
+   * - 16 - 63
+     - Private use (analogous to `RFC 6996 <http://www.iana.org/go/rfc6996>`_) -- can be used
+       for testing and private deployments.
+   * - 64 - 4094
+     - Public ISDs. They must be globally unique.
+   * - 4095
+     - Reserved.
+   * - 4096 - 65535
```
nicorusti, line 82:
```suggestion
   * - 4095 - 65535
```
4095  was merged to the following unallocated block.  discussed and unified during IETF draft editing

doc/overview.rst
```diff
@@ -52,14 +52,92 @@ ISD and AS Numbering
...
+     - ~4.3 billion
+     - Unallocated.
+   * - ``2:0:0`` - ``2:ffff:ffff``
+     - ~4.3 billion
+     - Public SCION AS numbers -- ASes created for SCION that are not existing BGP ASes.
+   * - ``3:0:0`` - ``feff:ffff:ffff``
+     - ~280 trillion
+     - Unallocated.
+   * - ``ff00:0:0/32``
+     - 65535
```
nicorusti, line 123:
```suggestion
     - 65536
```

nicorusti (changes requested):
Hi Katya, I proposed some changes to some numbers to align with 
https://www.ietf.org/archive/id/draft-dekater-scion-controlplane-18.html#name-addressing

However, could we perhaps just directly point to the draft without repeating all the info? I feel that the text description in the draft is more precise/accurate than the one in this PR.

For the text representation, you could point to: 
https://www.ietf.org/archive/id/draft-dekater-scion-controlplane-18.html#name-text-representation

Perhaps you could then extract key text from the RFC, and only if necessary add some text expanding on it. For example, this text on folders in win/mac is an implementation detail that of course is not in the draft, so it should stay: 

> As : is not a legal filename character on Windows and macOS, file or directory names containing an AS number should use _ as a separator instead. For example, a directory for 1-11:bc:1c could be ISD1-AS11_bc_1c.

PR #4925, opened 2026-04-28
docs: add NOTICE for patents

nicorusti:
This PR is to clarify licensing in relation to the [patent pledge](https://learn.anapaya.net/docs/resources/patent-pledge/#pledged-patents ) by Anapaya 

I added a top-level NOTICE file linking Anapaya Systems AG's patent pledge, the list of known patents, the disclosure process for contributions implementing a patented technique (including the PATENT-NOTICE source header convention), and an FAQ. The reason for using this file, is that this is mentioned explicitly in the Apache license. 

The contribution also proposes guidelines for committing patented claims. There are two cases: 

1. Contributor owns the patent (e.g. Anapaya) --> A patent license is automatically granted according to Apache 2.0 section 3
2. Contributor does not own the patent --> contributor must rely on a patent pledge

In both cases, I propose that the relevant patent is disclosed in the relevant file header


NOTICE
```diff
@@ -0,0 +1,86 @@
...
+----------------------------------------   -------------------------------------------------
+EP3756317B1, CN111771359B, US11240140B2    SCION-to-IP Gateway (SIG) including traffic
+                                            classification, path evaluation, high-availability
+                                            deployment.
+
+EP3941006B1, US11362930B2                  Hides a SCION network behind a single BGP AS, with
+                                           gateways exchanging routes with legacy networks via
+                                            BGP.
+
+EP3941003B1, US11362932B2                  High availability for SCION AS control services via
```
tzaeschke, line 55: This patent is not listed on the Anapaya page (https://learn.anapaya.net/docs/resources/patent-pledge/#pledged-patents) , but I assume that is an error on their part, they have [EP3941003A1](https://patents.google.com/patent/EP3941003A1/) listed twice.

NOTICE
```diff
@@ -0,0 +1,86 @@
...
+     according to the Apache License 2.0:
+
+        // PATENT-NOTICE: EP1234567B1
+        // This file implements a technique described in the patent(s) above, owned by
+        // <you/your organization>. As the patent owner and contributor of this code, you
+        // license these claims to all users under the Apache License 2.0, Section 3.
+
+
+
+Known Patents
```
tzaeschke, line 42: Just briefly looking at this, I don't think any of these technologies are implemented in scionproto. So they should probably not be listed here?

NOTICE
```diff
@@ -0,0 +1,86 @@
...
+        // Use of this file relies on an Open Source Patent Pledge. See NOTICE.
+
+
+    - If you are contributing code that implements a technique covered by a patent that
+     you or your employer own, note that patent rights come from you as the contributor
+     according to the Apache License 2.0:
+
+        // PATENT-NOTICE: EP1234567B1
+        // This file implements a technique described in the patent(s) above, owned by
+        // <you/your organization>. As the patent owner and contributor of this code, you
```
tzaeschke:
```suggestion
        // <you/your organization>. As the patent owner and contributor of this code,  <you/your organization>
```

NOTICE
```diff
@@ -0,0 +1,86 @@
...
+        // license these claims to all users under the Apache License 2.0, Section 3.
```
tzaeschke, line 38:
```suggestion
        // PATENT-NOTICE: EP1234567B1
        // This file implements a technique described in the patent(s) above, owned by
        // <you/your organization>. As the patent owner and contributor of this code,
        // <you/your organization> license these claims to all users under the 
        // Apache License 2.0, Section 3.
```
Alternatively, replace "you" with "I" or "we".

doc/dev/contribute.rst
```diff
@@ -243,3 +243,18 @@ Learning resources
...
+.. _licensing:
+
+Licensing and Intellectual Property
+===================================
+
+The SCION open-source implementation is released under the `Apache License, Version 2.0 <https://github.com/scionproto/scion/blob/master/LICENSE>`_.
+
+You may use, reproduce, and distribute the software in compliance with the License.
+
+Anapaya Systems AG holds patents related to certain SCION technologies, for which it has issued an open-source patent pledge.
```
tzaeschke, line 257: This sentence should probably be moved to the list in NOTICE,
Otherwise we need to add instructions for other patent holders to add their name here.

tzaeschke (changes requested):

NOTICE
```diff
@@ -0,0 +1,86 @@
+Additional IP Rights Grant (Patents)
```
shitz: The pledge is a promise not to sue, not a grant, and its scope is narrower than a license (it excludes hardware, non-FOSS software, and FOSS combined with either, except internal-only use). Calling it a grant will mislead readers. I suggest you change the title to "Patent Notice"

NOTICE
```diff
@@ -0,0 +1,86 @@
...
+====================================
+
+Anapaya Systems AG holds patents for which it has issued a formal
```
shitz, line 4: It would be good to explicitly state the pledge's scope limitation, e.g., "The pledge applies to Free or Open Source Software and internal-only use, not to hardware or to distribution in combination with proprietary software."

NOTICE
```diff
@@ -0,0 +1,86 @@
...
+open-source patent pledge.  Its terms and patents covered are available
+at:
+
+    https://learn.anapaya.net/docs/resources/patent-pledge/
```
shitz, line 8: I would at least add when it was accessed (the same way research papers cite a URL)

NOTICE
```diff
@@ -0,0 +1,86 @@
...
+
+
+
+Patent-Covered Contributions
```
shitz, line 17: This asks contributors to assess whether their code infringes patents and to record the result in the repo. I would make that invitational rather than a duty. Also, the process text should not be in NOTICE.

NOTICE
```diff
@@ -0,0 +1,86 @@
...
+
+FAQ
+===
+
+
+1. Can I use this implementation because of the patents?
+
+  Should this implementation use any of the patents above, 
+  the Apache License 2.0, Section 3 grants a patent
+  license for any contributions from patent owners.
```
shitz: I'm not sure this correct. Section 3 is contribution-scoped. It doesn't grant anything to a user for a contribution made by a third-party. That's what the pledge is for in the case of Anapaya not Apache-2.0 Section 3.

I suggest replacing with a direct answer:

FOSS development, distribution, and internal-only use are covered by the pledge; contributors also grant patent licenses for their own contributions under Apache-2.0 Section 3. Combining with proprietary software or hardware for distribution is outside the pledge.

NOTICE
shitz: I agree, you doesn't work. I would simply state the patent number, the owner, and that the code was contributed under Apache-2.0.

shitz (changes requested):

NOTICE
nicorusti: Yes, I think this may be a typo,  @shitz is informed

NOTICE
nicorusti:
```suggestion
Patent Notice
```

NOTICE
nicorusti: It is now listed

NOTICE
nicorusti: My point for listing known patents here is that contributors should be informed of any patents, so that when they contribute, they can be aware of they are using the pledge

NOTICE
nicorusti: Good point, how about:
```suggestion
        // PATENT-NOTICE:
        // This file implements a technique described in the patent(s) < EP1234567B1>, 
        // owned by <owner>. This code was contributed by the patent owner under the
        // Apache License 2.0, Section 3.
```

NOTICE
nicorusti:
```suggestion
FOSS development, distribution, and internal-only use are covered by the pledge; contributors also grant patent licenses for their own contributions under Apache-2.0 Section 3. Combining with proprietary software or hardware for distribution is outside the pledge.
```

NOTICE
```diff
@@ -0,0 +1,86 @@
...
+
+
+    - If you are contributing code that implements a technique covered by a patent that
+     you or your employer own, note that patent rights come from you as the contributor
+     according to the Apache License 2.0:
+
+        // PATENT-NOTICE: EP1234567B1
+        // This file implements a technique described in the patent(s) above, owned by
+        // <you/your organization>. As the patent owner and contributor of this code, you
+        // license these claims to all users under the Apache License 2.0, Section 3.
```
tzaeschke:
```suggestion
        // PATENT-NOTICE: <patent ID>
        // This file implements a technique described in the patent(s) <patent ID>, owned by
        // <owner>. The implementation is contributed by the patent owner under the
        // Apache License 2.0, Section 3.
```
- Use `patent ID` i.o. of actual number in example
- `The implementation` is more precise than `this code`: This code can be anything in the file, there may be multiple patent implementations.
- Use `is` instead of `was`, because at the time this will be written it should be present tense?

NOTICE
tzaeschke: What is the definition of "known patents"? There are millions, so why include some patents but not others? Sorry for being nitpicky :)

NOTICE
tzaeschke: I think we need to be careful mentioning the pledge here. Is the paragraph still valid if a different company pledges their patents on their own pledge terms? Or is there a general definition of "pledge" that defines terms?

NOTICE
nicorusti: I mean Known patents on the SCIOn technology: so any patent that relates to SCION

nicorusti: I will clarify this in the text

doc/dev/contribute.rst
nicorusti: As proposed by Sam, I moved the disclosure process to the contribution guidelines, so I believe this sentence now is a fit here

NOTICE
nicorusti: Added your wording

NOTICE
nicorusti: > I would make that invitational rather than a duty. 

Ok, sounds good. Changed to: 

> If you are contributing code that you know or believe implements a technique covered by a
patent, you are **encouraged** to follow these disclosure steps

>  Also, the process text should not be in NOTICE. 

One reason why I thought about using  NOTICE is that this file is explicitly linked by Apache 2.0, and being in the repo, it gets "propagated" together with the SW. However, fine to move the process, I moved it

NOTICE
nicorusti: done

NOTICE
nicorusti: > Known patents relating to the SCION technology are:

NOTICE
nicorusti: I rewrote this section to make it clearer, what do you both think about the current text:

```
1. Can I use this implementation because of the patents?

  There are currently no known contributions from non-patent owners.
  Once they come, such contributions should be disclosed according to 
  the process linked above. Their use would be dependent on a patent
  pledge. 

  The Anapaya pledge linked above covers FOSS development, distribution, and internal-only use, but not distribution combined with proprietary
  software or hardware. The pledge itself is the ultimate source of truth; this is only a summary and does not modify or supersede it. 

  Note that contributors grant patent licenses for their own contributions 
  under Apache-2.0 Section 3.
  ```

File: doc/dev/contribute.rst [hunk 12]
```diff
@@ -243,3 +243,49 @@ Learning resources
...
+===================================
+
+The SCION open-source implementation is released under the `Apache License, Version 2.0 <https://github.com/scionproto/scion/blob/master/LICENSE>`_.
+
+You may use, reproduce, and distribute the software in compliance with the License.
+
+Patent Disclosure Process
+=========================
+
+Anapaya Systems AG holds patents related to certain SCION technologies, for which it has issued an open-source patent pledge.
```
tzaeschke, line 260:
```suggestion
There may be patents related to certain SCION technologies, and these patents may by subject to an open-source patent pledge.
```
It seems my comment disappeared. I would avoid mentioning Anapaya here. This is not about Anapaya, it is about patents and pledges.

NOTICE
```diff
@@ -0,0 +1,74 @@
...
+============================
+
+If you are contributing code that you know or believe implements a technique covered by a
+patent, you are encouraged to follow the disclosure steps described in
+https://docs.scion.org/en/latest/dev/contribute.html
+
+Known Patents
+=============
+
+Known patents relating to the SCION technology are:
```
tzaeschke, line 27:
```suggestion
This is a purely informative non-comprehensive list of patents directly relating to the SCION technology. Note that these patent are not necessarily implemented in the code of this repository. Conversely, there may be implementations of patents that are not listed here. Some patents related to SCION technology are are: 
```
If we list patents here that are unrelated to the reference implementation, maybe we should clarify that. Also, we should make clear that the list is not meant to be comprehensive, there may be patents that are used here but not listed here; these could be patents related to SCION or unrelated to SCION.

NOTICE
```diff
@@ -0,0 +1,74 @@
...
+                                           gateways exchanging routes with legacy networks via
+                                            BGP.
+
+EP3941003B1, US11362932B2                  High availability for SCION AS control services via
+                                            sharding into independent shards, so control and
+                                            data planes keep working through failures; also
+                                            covers interoperation between SCION and BGP.
+
+                                            
+Please contact the maintainers or submit a PR if you become aware of
```
tzaeschke, line 45: Maybe clarify under what circumstances a patent should be mentioned here?
Should this be a list of patents related to SCION but potentially unrelated to this implementation?
Or, as is usually the case for NOTICE files, should this include patents related to this implementation but potentially unrelated to SCION?
Maybe this file should not be called NOTICE: NOTICE usually relates to the code in the project. This list appears to be about the SCION project, not about the implementation in this repository...

Maybe move the list to a separate file?

NOTICE
```diff
@@ -0,0 +1,74 @@
...
+
+
+1. Can I use this implementation because of the patents?
+
+  There are currently no known contributions from non-patent owners.
+  Once they come, such contributions should be disclosed according to 
+  the process linked above. Their use would be dependent on a patent
+  pledge. 
+
+  The Anapaya pledge linked above covers FOSS development, distribution,
```
tzaeschke, line 59: I don't think this answer should be specific to Anapay, but to anyone who holds relevant patents?

NOTICE
```diff
@@ -0,0 +1,74 @@
...
+
+                                            
+Please contact the maintainers or submit a PR if you become aware of
+any other patents that should be mentioned here.
+
+FAQ
+===
+
+
+1. Can I use this implementation because of the patents?
```
tzaeschke, line 52:
```suggestion
1. Can I use this implementation despite of the patents?
```
This sentence is unclear to me. Did you mean "despite"?

NOTICE
```diff
@@ -0,0 +1,74 @@
...
+
+  There are currently no known contributions from non-patent owners.
```
tzaeschke, line 54:
```suggestion

```
What is a contribution from a non-patent owner? ETH Zurich and none of the master students own patents...?
And how does it change the answer if there are contributions of non-patent owners?

I would just remove this sentence, otherwise we have to remove it as soon as we get contributions from non-patent owners (such as ETH Zurich?).

tzaeschke (changes requested):

doc/dev/contribute.rst
nicorusti: Reworded to: 

There may be patents related to certain SCION technologies for which
an open-source patent pledge was issued.

Known pledges are listed below. Their terms and patents covered are available
at the respective link: 

- Anapaya Systems AG:
    https://learn.anapaya.net/docs/resources/patent-pledge/
    Last accessed: 9 September 2026
    The pledge applies to Free or Open Source Software and internal-only use,
    not to hardware or to distribution in combination with proprietary software.

NOTICE
nicorusti: Added your exact wording below the table

NOTICE
nicorusti: > 1. Can I use this implementation because of the patents?
 
 Th answer to this question totally depends on what patents and related pledges exist, and wether they are implemented in the repo, and what conditions are given in each pledge. Because of that,  this should not be a generic answer IMHO, rather it should refer to the Anapaya pledge. I would not change that text.

NOTICE
nicorusti: > Maybe clarify under what circumstances a patent should be mentioned here?

Changed text to: 
> Please contact the maintainers or submit a PR if you become aware of
any other patents _relating to the SCION technology or to this
implementation_ that should be mentioned here.

The idea is to list patents related to this implementation, which is the SCION reference implementation. The patents listed are IMHO very related. This list should stay in the notice: 

1.  so new contributors are aware of what potential contributions would be covered by a pledge (e.g., someone wants to commit some HA functionality for the CP)
2. When components covered by the patents are contributed, so that users are aware

NOTICE
nicorusti: 1. Can I use this implementation in spite of the patents?

NOTICE
nicorusti: Right now, they are contributions from anyone except Anapaya, this is why the current sentence is valid. 

However, this may change. I reworded to: 
>   Contributions of code covered by patents from non-patent owners
  should be disclosed according to the process linked above.
  Their use would be dependent on a patent pledge. 
  
 Once we get any contribution covered, we could add a small list of components likely covered

NOTICE
tzaeschke: 1) I was told there is actually a SIG in SCION proto, so the first row of patents should be listed here.
2) As for other patents, what I mean is why list SCION patents in NOTICE that are not infringed here, but not list any other possibly related patents? 

Regarding 2): 
- I think a "NOTICE" file should only contain patents that are actually infringed. I think the purpose and expectation of the NOTICE file (I am not aware of a clear definition though) is to inform users of the library any legal "baggage". Patents that are not infringed by the library should no be here. I expect people find this confusing, having a patent listed in NOTICE suggests it is infringed, when in fact it is not.
- If we want to list other patents (SCION, AF-XDP, HTTP, protobuf, QUIC, or algorithms for routing, queueing, graph processing, ...) we should list them somewhere else, either in a separate file or probably outside the repository on the association's website. But I would actually recommend not to list them at all. Listing patents that may be possibly infringed in the future feels weird/dangerous: everybody whose patent may be infringed at some point in the future may then ask to have their patents advertised in the repository. Again, if you want to list them, I propose a separate file/website.

NOTICE
tzaeschke: > The patents listed are IMHO very related. This list should stay in the notice:

If a developer or a legal person looks at the repository, it may be difficult to explain that they are related to the code in the repository. I'd argue these patents are related to SCION, but not to the library in the repository.
(actually, since we have a SIG, some patents *are* clearly related).

NOTICE
tzaeschke: What I mean here is that this is a general question, but the answer is not general.

The answer is wrong as soon as someone else has a patent or if Anapaya adds a patent with a different pledge. That means we would have to update the answer every time a new patent holder shows up.

I propose to phrase this in more general terms.

If you really want to mention Anapaya, maybe write something like: 

```
At the time of writing (September 2026), only Anapaya holds patents on SCION, and all their patents are under a pledge (linked above) that covers FOSS development, distribution, ...
```

nicorusti: Thank you, changed to: 

>   At the time of writing (September 2026), only Anapaya Systems is known
  to hold patents on SCION, and their patents fall under the pledge
  linked above, which covers FOSS development, distribution, ...

NOTICE
nicorusti: How about we add this wording: 

> Components of this repository that may leverage a patent pledge are:

- scion/gateway (EP3756317B1, CN111771359B, US11240140B2)

nicorusti: > I was told there is actually a SIG in SCION proto, so the first row of patents should be listed here.
 
That code was mostly contributed by Anapaya, so article 3 of the apache 2.0 license applies and it grants any patent rights.  Should that code be touched again by anyone who is not Anapaya, it will have to bear the patent notice in file headers and it s usage would depend on the pledge.


> As for other patents, what I mean is why list SCION patents in NOTICE that are not infringed here, but not list any other possibly related patents?

Hi TIl, as I pointed out in another comment, so that new contributors are aware of what potential contributions would be covered by a pledge (e.g., someone wants to commit some HA functionality for the CP)
I think it is important to have them somewhere..

tzaeschke (approved):
LGTM

Issue #4909, opened 2026-03-20
doc: how are peering segments stitched when core is involved


tzaeschke:
As far as I understand, peering can happen between CORE AS and child AS and even between two core ASes.

However, I could not find any documentation on what exactly is possible and how this is done.

Looking at the docs:
- The scenario overview does not show peering with core ASes, see bottom figure: https://docs.scion.org/en/latest/control-plane.html#path-segment-combination
- Likewise, https://docs.scion.org/en/latest/protocols/scion-header.html#peering-links and https://docs.scion.org/en/latest/protocols/scion-header.html#path-calculation only mention UP/DOWN peering.

Some of the text can probably easy be adapted for peering between core ASes and child ASes. However, that leaves some open questions:
- Is CORE-CORE peering possible? If yes, are there any restriction?
- CORE-CORE means that we have two peering ASes in a CORE segment. How does the cutting/stitching work?
- It seems this is currently implemented by adding a UP or DOWN segment? If so what happens if there are already UP/DOWN segments?

Separately, it may be useful to give advice how to deal when multiple (mutually exclusive) stitching options are available.

See also #4828, #4890, #4905

oncilla:
Relevant slack discussion: https://scionproto.slack.com/archives/C8ADA9CEP/p1758826648132419

tzaeschke:
For my understanding, based on the slack discussion, I have some more questions. Let's use the following example:

Example with UP, CORE and DOWN segments: `(X->A)->(A->B->C)->(C->Y)`
```
(A) --- (B) --- (C)
 |               |
 | UP       DOWN |
 |               |
(X)             (Y)
```

1) In the example above, core peering means that `(X->A)->(A->B->C)->(C->Y)` could become `(X->A)--peering-link-->(C->Y)` (assuming A and C have a peering link). Correct?
2) In the example above, can (B) announce peering links, and can we realize a peering link between (B) and any of the other ASes: (X), (A), (C), or (Y)?
3) The discussion mentions "requesting" 1-AS segments. How is that done? If I request a 1-AS segment from a CS it seems to return an empty list.

katyatitkova:
> In the example above, core peering means that `(X->A)->(A->B->C)->(C->Y)` could become `(X->A)--peering-link-->(C->Y)` (assuming A and C have a peering link). Correct?

Yes

> In the example above, can (B) announce peering links, and can we realize a peering link between (B) and any of the other ASes: (X), (A), (C), or (Y)?

Yes

[...]

PR #4926, opened 2026-05-15
doc: core peering

tzaeschke:
This PR proposes documentation that describes how peering shortcuts with core ASes work. See also: #4909

This PR does not add any diagrams. New/additional diagrams may be useful, in that case I would ask for access to the originals of the existing diagrams. It could generally be useful if diagram sources would be added to the repository.

### DIFF
--- a/doc/control-plane.rst
+++ b/doc/control-plane.rst
@@ -312,7 +312,7 @@ shortcuts are only allowed (and announced) between ASes that are on up- segment 
 This trivially allows peering shortcuts between non-core ASes and ASes.
 
 Whenever an AS is queried for up- or down-segments that terminate in that same AS,
-and if the AS has any peering links, it will return a "zero"-hop segment that contains only
+and if the AS has any peering links, it will return a single-AS segment that contains only
 the AS itself and its peering links.
 
 
--- a/doc/control-plane.rst
+++ b/doc/control-plane.rst
@@ -171,7 +171,7 @@ Peering Links
 
 PCBs do not traverse peering links.
 Instead, available peering links are announced along with a regular path in the individual AS
-entries of PCBs during *intra-AS* beaconing. Peering links are not announced in *inter-AS* beaconing.
+entries of PCBs during *intra-ISD* beaconing. Peering links are not announced in *inter-ISD* beaconing.
 If both ASes at either end of a peering link have registered path segments that include a specific
 peering link, then it can be used to during segment combination to create an end-to-end path.
 
--- a/doc/control-plane.rst
+++ b/doc/control-plane.rst
@@ -171,7 +171,7 @@ Peering Links
 
 PCBs do not traverse peering links.
 Instead, available peering links are announced along with a regular path in the individual AS
-entries of PCBs.
+entries of PCBs during *intra-AS* beaconing. Peering links are not announced in *inter-AS* beaconing.
 If both ASes at either end of a peering link have registered path segments that include a specific
 peering link, then it can be used to during segment combination to create an end-to-end path.
 
@@ -195,7 +195,7 @@ core AS that originated the PCB.
 Intra-ISD Path-Segment Registration
 -----------------------------------
 
-Every *registration period* (determined by each AS), the AS's control service selects of
+Every *registration period* (determined by each AS), the AS's control service selects
 PCBs to transform into path segments:
 
 - Up-segments, which allow the infrastructure entities and endpoints in this AS to communicate with
@@ -206,6 +206,9 @@ PCBs to transform into path segments:
   core AS that originated the PCB.
   As a result, a core AS's path database contains all down-segments registered by their
   direct or indirect customer ASes.
+
+In addition, if the AS is a core AS and if peering links are present, the control service adds
+an UP and a DOWN segment that contains only the AS itself along with all peering link metadata.
 
 Core Path-Segment Registration
 ------------------------------
@@ -225,13 +228,13 @@ Path Lookup
 An endpoint (source) that wants to start communication with another endpoint (destination), needs
 up to three path segments:
 
-- An up-path segment to reach the core of the source ISD
+- An up-path segment to reach the core of the source ISD or to get peering metadata for the core AS,
 - a core-path segment to reach
 
   - another core AS in the source ISD, in case the destination AS is in the same source ISD, or
   - a core AS in a remote ISD, if the destination AS is in another ISD, and
 
-- a down-path segment to reach the destination AS.
+- a down-path segment to reach the destination AS or to get peering metadata for the core AS.
 
 The process to look up and fetch path segments consists of the following steps:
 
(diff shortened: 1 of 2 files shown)


doc/control-plane.rst
```diff
@@ -171,7 +171,7 @@ Peering Links
 
 PCBs do not traverse peering links.
 Instead, available peering links are announced along with a regular path in the individual AS
-entries of PCBs.
+entries of PCBs during *intra-AS* beaconing. Peering links are not announced in *inter-AS* beaconing.
```
oncilla:
```suggestion
entries of PCBs during *intra-ISD* beaconing. Peering links are not announced in *inter-ISD* beaconing.
```

doc/control-plane.rst
```diff
@@ -296,6 +299,42 @@ end-to-end paths are encoded in the packet header.
...
+
+All ASes may have physical peering links. However, for segment combinations, peering
+shortcuts are only allowed (and announced) between ASes that are on UP segment or DOWN segments.
+This trivially allows peering shortcuts between non-core ASes and ASes.
+
+Peering shortcuts However, for segment combination,
+peering links with core ASes can only be used *indirectly* and only for the first and last AS in a core segment.
+Indirect means:
+
+- Any core AS that has peering links returns, when queried for an UP or DOWN segment,
```
oncilla: This is imprecise. A core as returns these segments when queried for an up/down segment with the source/destination being the core AS themselves.

doc/control-plane.rst
```diff
@@ -296,6 +299,42 @@ end-to-end paths are encoded in the packet header.
...
+This section distinguished peering links and peering shortcuts:
+
+- Peering link: A physical link between two ASes.
+- Peering shortcut: A peering link that is used when constructing a full path.
+
+All ASes may have physical peering links. However, for segment combinations, peering
+shortcuts are only allowed (and announced) between ASes that are on UP segment or DOWN segments.
+This trivially allows peering shortcuts between non-core ASes and ASes.
+
+Peering shortcuts However, for segment combination,
```
oncilla: I am unsure if this explanation will not further confuse the uninitiated reader.

It might be easier to understand if we just describe what the role of the core segment is in such scenarios. In the end, it is simply a mechanism to discover the source ASes for the candidate down segments.

doc/control-plane.rst
```diff
@@ -296,6 +299,42 @@ end-to-end paths are encoded in the packet header.
...
+peering links with core ASes can only be used *indirectly* and only for the first and last AS in a core segment.
+Indirect means:
+
+- Any core AS that has peering links returns, when queried for an UP or DOWN segment,
+  a "zero"-hop segment that contains only itself and its peering links.
+- Client must always query for UP and DOWN links, even when they are located in a core AS.
```
oncilla: The client must do so, if the source or destination of the path is core.
Where the client is located, is irrelevant.

doc/control-plane.rst
```diff
@@ -296,6 +299,42 @@ end-to-end paths are encoded in the packet header.
...
+
+As a result, if the core ASes have peering links, any client will have at least one UP and one DOWN
+segment (even if source and destination AS are core ASes).
+These UP and DOWN segments, or any other UP or DOWN segments if they are available, can be used to
+create a peering shortcut.
+
+FAQ: Why do core segments not contain peering metadata?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+AS may use peering link instead of a public link if they
```
oncilla: That sentence doesn't read well. How about this?

> Peering links enable two AS to route traffic between their respective downstream ASes while explicitly restricting global transit. Embedding this peering data within a core segment, which inherently facilitates transit, would directly contradict the non-transit nature of peering.

oncilla (changes requested):

doc/control-plane.rst
```diff
@@ -296,6 +299,42 @@ end-to-end paths are encoded in the packet header.
...
+does not want the link to be used for traffic that transits the ISD.
+The peering link allows usage to be limited to traffic that starts or ends in the local AS or one of its child ASes.
+If peering links were announced withe core segment metadata, they could be used on
+core segments that *transit* the local AS.
```
tzaeschke:
```suggestion
Peering links enable two ASes to route traffic between each other and their respective downstream ASes while restricting global transit. Embedding this peering data within a core segment, which inherently facilitates transit, would directly contradict the non-transit nature of peering.
```
@oncilla Thanks for the suggestion, how about this?

doc/control-plane.rst
```diff
@@ -296,6 +299,39 @@ end-to-end paths are encoded in the packet header.
...
+- Peering link: A physical link between two ASes.
+- Peering shortcut: A peering link that is used when constructing a full path.
+
+All ASes may have physical peering links. However, for segment combinations, peering
+shortcuts are only allowed (and announced) between ASes that are on up- segment or down-segments.
+This trivially allows peering shortcuts between non-core ASes and ASes.
+
+Whenever an AS is queried for up- or down-segments that terminate in that same AS,
+and if the AS has any peering links, it will return a "zero"-hop segment that contains only
+the AS itself and its peering links.
```
tzaeschke: @oncilla My change seems to have dropped your comments. I rewrote this section, do you think it is clearer now?

oncilla: I think this is more understandable.

I am unsure about calling them "zero"-hop segments though, because technically, they have exactly one hop.

To bad one-hop path is a thing (and a misnomer, given it has two hops 🤦), which would make calling these segments "one-hop segments" even more confusing.

Any chance we can find a better name?
Maybe "single-AS segment"?

doc/control-plane.rst
oncilla: Sounds good to me!

doc/control-plane.rst
```diff
@@ -296,6 +299,35 @@ end-to-end paths are encoded in the packet header.
...
+
+- Peering link: A physical link between two ASes.
+- Peering shortcut: A peering link that is used when constructing a full path.
+
+All ASes may have physical peering links. However, for segment combinations, peering
+shortcuts are only allowed (and announced) between ASes that are on up- segment or down-segments.
+This trivially allows peering shortcuts between non-core ASes and ASes.
+
+Whenever an AS is queried for up- or down-segments that terminate in that same AS,
+and if the AS has any peering links, it will return a "zero"-hop segment that contains only
```
tzaeschke:
```suggestion
and if the AS has any peering links, it will return a single-AS segment that contains only
```

doc/control-plane.rst
```diff
@@ -296,6 +299,35 @@ end-to-end paths are encoded in the packet header.
...
+the AS itself and its peering links.
+
+
+This ensures that, if the core ASes have peering links, any client will have at least one up- and one
+down-segment (even if source and destination AS are core ASes).
+The peering metadata from these zero-hop-up/down-segments, or any other up- or down-segments
```
tzaeschke:
```suggestion
The peering metadata from these single-AS up/down-segments, or any other up- or down-segments
```

doc/control-plane.rst
tzaeschke: Good point, yes, I chose zero-hop because OneHop was already taken :-)
I like "single-AS".

doc/control-plane.rst
tzaeschke: Fixed.

doc/control-plane.rst
tzaeschke: Fixed

doc/control-plane.rst
tzaeschke: Done.

doc/control-plane.rst
tzaeschke: @oncilla I'm not sure what you mean by "role of the core segment"?

Anyway, I rephrased the whole section, maybe it is better now?

PR #4928, opened 2026-05-20
hummingbird: quic integration test

juagargi:
For the purpose of checking that QUIC can indeed run with Hummingbird, create test artifacts that asses it.

- [ ] Unit test that runs manually using the tiny topology.
- [ ] Integration test.


juagargi:
Bad target branch, closing.

PR #4940, opened 2026-06-25
graph: fix how topogen's random interface IDs get replaced with the graph package's fixed ones

katyatitkova:
Apparently, the files got sorted slightly differently occasionally, so we can't rely on the line order anymore.


lukedirtwalker (approved):
TBH no clue what this does. but if it fixes things I'm happy.

PR #4958, opened 2026-07-28
deps: make Dependabot more usable

katyatitkova:
There's a problem in this repo: updates created by Dependabot are often non-mergeable for various reasons:

- Go dependencies PRs only update the versions, so if protobuf or license file regeneration is required, someone has to do it manually;
- Python dependencies are updated only in `requirements.txt` files, not touching `requirements.in` at all;
- JS vulnerabilities are mostly coming from dependencies of (non-vulnerable) redocly, so just a version update of redocly is needed, but Dependabot doesn't get it;
- the PR titles don't pass our PR title check.

We have regular Dependabot updates turned off, but it would still be nice to use Dependabot's capabilities to update dependencies that become vulnerable (which I try to do regularly). Right now it's still a lot of manual work. So, this PR tries to solve the Dependabot problems for most cases.

I added a workflow that does the following:

1. It renames the PR.
2. If Go deps were updated, it runs `make protobuf mocks gazelle write_all_source_files licenses`.
3. If Python deps were updated, it adds the dependency from `requirements.txt` to `requirements.in` and runs Bazel command to re-generate `requirements.txt`.
4. If JS deps were updated, lockfile is re-generated.
5. All the changes are committed into the branch.

Since JS vuls come mostly from redocly deps, I also added a separate config that will update JS deps weekly. If a vulnerability appears and it can't be fixed by Dependabot, we still can trigger the JS deps update manually (from Insights -> Dependency graph -> Dependabot).

There are two caveats:

1. If a commit was created by the workflow, someone needs to let the checks re-run (the same thing as when non-maintainers create PRs).
2. If a commit was not created, the PR title is still changed, but somehow it doesn't trigger the PR title check 🫠 If a human changes the PR title afterwards, the check is triggered. It's annoying, but still better than before, when 100% of PR titles by Dependabot didn't pass the check.

Examples of the PRs that are being created manually from the "Security" tab and then modified automatically can be found here: https://github.com/katyatitkova/scion-dependabot-test/pulls

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+# Scheduled version updates: npm only. Most JS vulnerabilities arrive as
+# transitive dependencies of @redocly/cli, so keeping the mgmtapi tools
+# current eliminates the alert churn in this ecosystem. Go and Python stay
+# security-updates-only: ecosystems absent from this file get no version
+# updates.
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: "/private/mgmtapi/tools"
+    schedule:
+      interval: weekly
+    commit-message:
+      prefix: "deps"
+    groups:
+      npm-tools:
+        patterns:
+          - "*"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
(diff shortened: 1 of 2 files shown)


romshark (approved):
LGTM

